PR #138, opened 2025-04-28
Update package.json

i5d6:
Introduced via: docusaurus-plugin-internaldocs-fb@1.19.1 DOMPurify@3.2.4

The vulnerability in DOMPurify could allow an attacker to inject a malicious script into the application if user input is not correctly sanitized. In particular, this issue manifests in the handling of certain HTML elements or attributes, which can be exploited through a crafted payload that bypasses the sanitization process.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -34,7 +34,7 @@
     "@stylexjs/webpack-plugin": "^0.11.1",
     "@svgr/webpack": "8.1.0",
     "clsx": "^1.2.1",
-    "docusaurus-plugin-internaldocs-fb": "^1.19.1",
+    "docusaurus-plugin-internaldocs-fb": "^3.2.4",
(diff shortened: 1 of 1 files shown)


i5d6 (commented):
The Cross-Site Scripting (XSS) vulnerability found in DOMPurify (introduced via docusaurus-plugin-internaldocs-fb@1.19.1) can have severe consequences, including session hijacking and credential theft. This issue has been addressed and fixed in DOMPurify@3.2.4.

We strongly recommend updating to the fixed version of DOMPurify and ensuring that proper sanitization is applied to all user inputs.

ndmitchell:
This changes the wrong package. Was this human authored? Can you explain how it solves the issue?

i5d6:
@ndmitchell yes

PR #148, opened 2025-05-01
Updates Docs page with example typo fix in configuration.mdx

sifex:
## Context

This PR updates Website Docs page with example typo fix in `configuration.mdx`.

The example provided had referenced `[pyrefly]` where the TOML stanza should be `[tool.pyrefly]` instead. Confirmed that Pyrefly catches that configuration once the prefix is updated. 

**Python Version:** Python 3.11.11
**Poetry Version:** Poetry (version 2.1.1)
**Pyrefly Version**: pyrefly 0.13.0

I couldn't find which PEP lays out the standard for prefixing `tool.` with tooling in `pyproject.toml`, but it is mentioned slightly in here. https://peps.python.org/pep-0621/

### DIFF
--- a/website/docs/configuration.mdx
+++ b/website/docs/configuration.mdx
@@ -438,7 +438,7 @@ invalid-argument = false
 ...
 
 # Pyrefly header
-[pyrefly]
+[tool.pyrefly]
 
 #### configuring what to type check and where to import from
 project_includes = ["src"]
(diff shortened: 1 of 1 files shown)


connernilsen:
Hey @sifex, thanks for making this change! Didn't realize I missed that when copying things around 😅

yangdanny97:
@Solumin do we also need to update the `[error]` config in the examples for https://pyrefly.org/en/docs/migrating-from-mypy/ and https://pyrefly.org/en/docs/migrating-from-pyright/ ?

connernilsen:
@yangdanny97 it should be fine as is. `[errors]` is the correct top-level table, but this was adjusting our `pyconfig.toml` example.

Issue #55, opened 2025-03-28
Docs: Setup Pyrefly with Neovim


akmalsoliev:
Nice project, wanted to add some information on setting up Pyrefly with Neovim 0.11+ or `lspconfig`.

**`lspconfig`**
```lua
local lspconfig = require("lspconfig")
local configs = require("lspconfig.configs")

if not configs.pyrefly then
  configs.pyrefly = {
      cmd = { "uv", "run", "pyrefly", "lsp" },
      filetypes = { "python" },
      root_dir = function(fname)
        return lspconfig.util.find_git_ancestor(fname) or vim.loop.os_homedir()
      end,
      settings = {},
    },
  }
end

lspconfig.pyrefly.setup({})
```

**For Neovim 0.11+**
Place lua file under `~/.config/nvim/lsp/`, for simplicity sake let's name it `pyrefly.lua`
```lua
return {
  cmd = { "pyrefly", "lsp" },
  filetypes = { "python" },
  settings = {},
}
```

then enable it in `init.lua`
```lua
vim.lsp.enable({"pyrefly"})
```

connernilsen:
Hey @akmalsoliev, thanks so much for creating this! I'm a Neovim user myself, but haven't had the time to put anything together for it yet. This is going to be super helpful, and at some point, once our docs are ready, it would be awesome if we can add this in.

At the moment, we're still figuring out how we want to do docs, but someone will reach out or pull this in once we're ready.

akmalsoliev:
> Hey [@akmalsoliev](https://github.com/akmalsoliev), thanks so much for creating this! I'm a Neovim user myself, but haven't had the time to put anything together for it yet. This is going to be super helpful, and at some point, once our docs are ready, it would be awesome if we can add this in.
> 
> At the moment, we're still figuring out how we want to do docs, but someone will reach out or pull this in once we're ready.

Hey @connernilsen, thank you for the kind words and happy to hear that there a fellow neovim user. v0.11 is super easy and super nice addition, highly recommended. I'm currently working on modifying my config to solve the absurd setup of lsp and other services, I'd also recommend to add Pyrefly to https://github.com/williamboman/mason.nvim registry.

mrsobakin:
Hi @akmalsoliev! Thank you for sharing your configuration.

I'm using Pyrefly configured with the new LSP API introduced in v0.11, and I've been running into some issues.

[...]

PR #185, opened 2025-05-11
Fix LSP's hanging threads, refactor connection

mrsobakin:
Fixes a deadlock where `wait_on_connection()` would hang indefinitely because the sending side of the `Connection` wasn’t properly closed. You can find the full context [here](https://github.com/facebook/pyrefly/issues/55#issuecomment-2868667781).

I’ve added an explicit drop of the `Server` to ensure the connection is properly shut down before waiting.

I've also taken the liberty to clean up the LSP connection code a bit: clunky ad-hoc `send` closure was replaced with a `ServerConnection` wrapper around `Arc<Connection>`, which groups message sending logic into its methods.

### DIFF
--- a/pyrefly/lib/commands/lsp.rs
+++ b/pyrefly/lib/commands/lsp.rs
@@ -221,8 +221,45 @@ enum ServerEvent {
     LspRequest(Request),
 }
 
+#[derive(Clone, Dupe)]
+struct ServerConnection(Arc<Connection>);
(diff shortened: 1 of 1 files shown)


SamChou19815:
Thanks for the contribution, especially for making us aware of the problem with non-vscode integration. The cleanup certainly makes the code look nicer.

PR #202, opened 2025-05-15
chore: bump to 2024 edition

gvozdvmozgu:


ndmitchell:
Thanks for the diff! There are two things here - moving to the 2024 edition, and running the formatter. I'm splitting these apart and will land the 2024 edition.

Internally we are using rustfmt 2.0, with patches, that is incomplete. I'm discussing internally, but don't think this gives a very nice open source experience. I'm going to either move us to real cargo fmt, OR write the code such that it passes in both variants.

ndmitchell:
I've made the code format identically with both versions. I'm just validating the change to apply the cargo fmt open source too #230

ndmitchell:
We now have that landed, and are using standard cargo fmt on our open source.

PR #220, opened 2025-05-16
Build aarch64-unknown-linux-gnu wheels

amartani:
Build wheels for aarch64 (ie. arm64) linux.


ndmitchell:
Thanks for the code. I happened to discover this independently, so am putting up a pull request with the same contents, although if you would prefer to mark your code as ready for review and click through the CLA I'd be happy to merge yours instead.

ndmitchell:
Done as #309 - thanks for the effort fixing this.

PR #282, opened 2025-05-19
Fix minor typo in the README.md

ricardoleal20:
I found a minor typo issue in the README.md file.

Nothing too fancy, but in the part of the inspiration, I've found that it's currently written as **_We are take inspiration_**, and I suppose that you're thinking of writing **We are taking inspiration**


ndmitchell:
Thanks very much - appreciate it!

PR #283, opened 2025-05-19
Re-sync with Meta

bigfootjon:


pyrefly/lib/test/lsp/hover_type.rs
```diff
@@ -257,50 +257,6 @@ Hover Result: `Literal[5] | int`
     );
 }
 
-#[test]
-fn synthesized_vars_for_awaitable_any_test() {
```
yangdanny97, line 261: I think this is the one we need to keep

pyrefly/lib/test/lsp/hover_type.rs
```diff
@@ -257,50 +257,6 @@ Hover Result: `Literal[5] | int`
...
-        ^
-Hover Result: `@_`
-"#
-        .trim(),
-        report.trim(),
-    );
-}
-
 #[test]
 fn synthesized_vars_for_awaitable_any_test() {
```
yangdanny97, line 261: this one can be deleted

bigfootjon:
I really recommend that we merge this as-is, as it will allow ShipIt to actually sync from a clean state. Otherwise we're going to have to do another resync loop

PR #303, opened 2025-05-20
fix: assert_type test for indexing into byte literal

jchanke:
continuing the work from https://github.com/facebook/pyrefly/commit/03084134d88ae15f5e1943850e39a473fb19316c

fixing the test case when indexing into a bytestring: if `x: bytes`, then `x[0]: int`.
see [PEP 358](https://peps.python.org/pep-0358/#specification) for more info!

### DIFF
--- a/pyrefly/lib/test/literal.rs
+++ b/pyrefly/lib/test/literal.rs
@@ -181,7 +181,7 @@ testcase!(
     r#"
 from typing import assert_type, Literal
 x = b"far"
-assert_type(x[0], int)
+assert_type(x[0], Literal[102])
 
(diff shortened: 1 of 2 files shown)


pyrefly/lib/test/literal.rs
```diff
@@ -181,7 +181,7 @@ testcase!(
     r#"
 from typing import assert_type, Literal
 x = b"far"
-assert_type(x[0], Literal[b"f"])
+assert_type(x[0], Literal[102])
```
stroxler, line 184: looks good! Let's just add another test case for when the literal is not an int

PR #323, opened 2025-05-21
Add a root Cargo.toml

ndmitchell:
Summary: Means users can cargo from the root, rather than doing a `cd`.

Reviewed By: rchen152

Differential Revision: D75097360


rchen152 (approved):

PR #325, opened 2025-05-21
Fix invalid code in error supression demo

Zeko369:
Before:
CleanShot 2025-05-21 at 11 08 13

After:
CleanShot 2025-05-21 at 11 08 32


yangdanny97:
Thanks!

PR #327, opened 2025-05-21
Fix singular and plural in summary

hugovk:
Computers are really bad at some things, but one thing they're really good at is counting :)

So let's replace:

```
 INFO 0 errors shown, 0 errors ignored, 1 modules, 65 transitive dependencies, 18,208 lines, took 0.10s, peak memory physical 76 MiB
```

With:

```
 INFO 0 errors shown, 0 errors ignored, 1 module, 65 transitive dependencies, 18,208 lines, took 0.57s, peak memory physical 89.3 MiB
```

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -575,7 +575,7 @@ impl Args {
 
         if !self.no_summary {
             info!(
-                "{} errors shown, {} errors ignored, {} modules, {} transitive dependencies, {} lines, took {timings}, peak memory {}",
+                "errors shown: {}, errors ignored: {}, modules: {}, transitive dependencies: {}, lines: {}, took {timings}, peak memory {}",
                 number_thousands(shown_errors_count),
                 number_thousands(errors.disabled.len() + errors.suppressed.len()),
                 number_thousands(handles.len()),


pyrefly/lib/commands/check.rs
```diff
@@ -570,15 +570,30 @@ impl Args {
...
+        let ignored_errors_count = errors.disabled.len() + errors.suppressed.len();
+        let transitive_dependencies_count = transaction.module_count() - handles.len();
         timings.report_errors = report_errors_start.elapsed();
 
         info!(
-            "{} errors shown, {} errors ignored, {} modules, {} transitive dependencies, {} lines, took {timings}, peak memory {}",
+            "{} error{} shown, {} error{} ignored, {} module{}, {} transitive dependenc{}, {} line{}, took {timings}, peak memory {}",
             number_thousands(shown_errors_count),
-            number_thousands(errors.disabled.len() + errors.suppressed.len()),
+            if shown_errors_count == 1 { "" } else { "s" },
```
yangdanny97: let's use the `count` function we already wrote, for the simple cases where we just add an s: https://github.com/facebook/pyrefly/blob/d24d371c02729d410147e96eefadab11996edc00/pyrefly/lib/util/display.rs#L141

yangdanny97:
I'm mixed on this overall, I generally prefer these types of outputs to have exact same text and only the numbers differ, because then it's easier to do some find/replace or regex to turn it into a CSV.

These days you could probably feed it to an LLM to extract the numbers, so maybe it's no big deal.

hugovk:
This is human-readable output, intended for humans to read, not machine-readable JSON or other format. If someone does want to parse these, it's not too difficult either.

By the same token, you may at some point want to change this summary for other reasons, like adding or removing more info. This could also cause problems with parsing the output, but that shouldn't stop you.

Here's a couple of other tools that do human output:

```console
# git
 1 file changed, 1 insertion(+), 2 deletions(-)
...
 55 files changed, 1364 insertions(+), 866 deletions(-)

# mypy
...
Success: no issues found in 6 source files
```

Alternatively, it could be rewritten:
```
INFO errors shown: 0, errors ignored: 0, modules: 1, transitive dependencies: 65, lines: 18,208, took 0.10s, peak memory physical 76 MiB
```

pyrefly/lib/commands/check.rs
hugovk: We're currently showing the numbers with thousands commas (using `number_thousands`) but `count(18208, "lines")` returns "18208 lines" and we want "18,208 lines".

Shall we add `number_thousands` into `count`, something like this?

```diff
 /// Format the number and descriptor word with the right pluralization.
 /// E.g., count(2, "cat") -> "2 cats". Pluralizes by tacking on an "s".
 pub fn count(n: usize, descriptor: &str) -> String {
-    format!("{n} {descriptor}{}", if n == 1 { "" } else { "s" })
+    format!("{} {descriptor}{}", number_thousands(n), if n == 1 { "" } else { "s" })
 }
```

yangdanny97: Sure, that could work. Maybe we could also make `count()` handle the `y`/`ies` pattern

but let's also see if anyone else on the team has opinions on this approach v.s. the alternative you suggested

stroxler:
I like the idea of flipping the order - then the benefits @yangdanny97 mentioned are there without the grammatical error. It also keeps the code more concise

hugovk:
Thanks, order flipped.

yangdanny97:
Thanks for your contribution @hugovk :)

PR #328, opened 2025-05-21
Fix typos

singiamtel:


website/docs/python-typing-for-beginners.mdx
```diff
@@ -59,7 +59,7 @@ class Rectangle:
 
 rect = Rectangle(width=100, height=50)
 
-area = rect.width * rect.hieght
+area = rect.width * rect.height
```
yangdanny97: this is intentional

website/docs/python-typing-for-beginners.mdx
```diff
@@ -79,7 +79,7 @@ class Rectangle:
...
```
yangdanny97: I think this is also intentional

website/docs/python-typing-for-beginners.mdx
singiamtel: Oops, my bad. Reverted

website/docs/python-typing-for-beginners.mdx
singiamtel: Same as above

rchen152 (approved):

Issue #301, opened 2025-05-20
Last summary line should be an option.


amar-laksh:
### Describe the Bug

On successive runs, the following line might change because of time taken differences and memory utilisation:
```
INFO 108 errors shown, 0 errors ignored, 37 modules, 897 transitive dependencies, 275,024 lines, took 0.69s (listing 0.49s; checking 0.19s), peak memory physical 209.8 MiB
```
this should be an optional line since users might want to use a diffing function on the output and this creates unnecessary noise that needs to be filtered outside the tool.

Thanks for the great tool! :)

rish-singhal:
can I pick this up?

yangdanny97:
@rish-singhal of course :)

yangdanny97:
completed in https://github.com/facebook/pyrefly/pull/358

Thanks again @arnav-jain1

PR #358, opened 2025-05-25
added -t flag to stop time and memory message from being shown in the…

arnav-jain1:
Adds feature requested in #301. The flag is -t (or --no-timer) and just removes the part of the output that has memory usage and time taken. Simple enough.
To be honest I am not sure how to add it to the printed information when running --help. If shown how to, I can add it and update the commit.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -116,6 +116,10 @@ pub struct Args {
     )]
     summarize_errors: Option<usize>,
 
+    /// Flag that will remove time and memory usage in output.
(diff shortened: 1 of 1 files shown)


arnav-jain1:
Signed

yangdanny97:
Hi, thanks for making a PR!

I think the linked issue is asking, "could we make the entire last line optional, and have the output just be the list of errors". While the output is deterministic with your change, there are some use cases where the last line would still need to be filtered out. I think it would be cleaner to have some sort of `--no-summary` flag that hides that line entirely.

arnav-jain1:
Made the change

stroxler (approved):

arnav-jain1:
Yippee

PR #368, opened 2025-05-27
Handling @overload with a docstring declaration

### LINKED ISSUE #359: Don't check return type for @overload and docstring
### Describe the Bug

`@overload` with a docstring declaration is very handy - the ellipsis is handled correctly. However, @overload combined with a docstring is not handled properly.

```python
from typing import *

@overload
def foo(a: str) -> str:      # <-- Function declared to return `str` but is missing an explicit `return`
    """Docstring"""

@overload
def foo(a: int) -> int: ...  # This is OK

def foo(*args, **kwargs) -> Any:
    pass
```

[Playground link](https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQAJgBuApiADZgCGAJkU1cFMGGABQMBcmFDACUUALQA+ITEEA6BVCgBiKABUAFkgDOmXQHkA0iXLU6jFmw5deAqNpggxU+48FKPqgDzjxUAGIArigAxjBIYChQbCG0DCBUTLBgUAkwgSBRAAYOIFlQAEaB+DpQEDraqBgMUVQAHgi0SCFI+FlpGShZRB5QAEQDACJgIblVA30kVpzcPATxaNoANIQEANYA7gvaztIAgihw-D0eCAza2kA).

Pyright [doesn't report any errors](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQAJgBuApiADZgCGAJkU1cFMGGABQMBcmFDACUUALQA%2BITEEA6BVCgBiKABUAFkgDOmXQHkA0iXLU6jFmw5deAqNpggxU%2B48FKPqgDzjxUAGIArigAxjBIYChQbCG0DCBUTLBgUAkwgSBRAAYOIFlQAEaB%2BDpQEDraqBgMUVQAHgi0SCFI%2BFlpGShZRB5QAEQDACJgIblVA30kVpzcPATxaNoANIQEANYA7gvaztIAgihw-D0eCAza2kA) in the code above.

### Sandbox Link

_No response_

_No response_

arnav-jain1:
fixes #359. Added a check to make sure the body was not a docstring. Also added a test in the the third_party folder to show that the change is working. Let me know if anything needs to be changed.


yangdanny97:
Thanks! 

The `third_party/conformance` tests are pulled from the typing spec conformance test suite, and shouldn't be modified directly here: https://github.com/python/typing/tree/main/conformance

For this type of feature, could you please add an integration test here? https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/overload.rs

Also, I think we should only allow docstrings for function bodies if there is an `@overload` decorator

For example, a program like this should still be invalid:

```
def foo() -> int:
    """hello"""
```

arnav-jain1:
Should be fixed. Added 3 test cases, first one is the basic one. Second one was to make sure that stuff after a docstring would still be typechecked properly and if the body of the function had more info, it wouldn't deem it a stub. The last one is for the case you mentioned of the nonoverloaded function with a docstring.

arnav-jain1:
My commits are a bit messy because I am a little new to this but the code should be right

arnav-jain1:
Also if you have others that you want me to fix, let me know! I am a bit swamped right now but I can try and get to it soon.

yangdanny97:
Sorry for the delay @arnav-jain1 and thanks for your contribution

We'll get this merged soon!

My only feedback would be that it's more readable to just use a loop and set the is_overload/has_no_type_checking inside the loop v.s. using a reduce. I've applied that change to the patch that we're merging, so there's no action required from you.

arnav-jain1:
All good on the delay. Been a bit busy so I forgot. But yeah I'll remember that for the future. Im tryna learn haskell or functional programming in general and kind of used it as a fun exercise.

yangdanny97:
> Im tryna learn haskell or functional programming in general and kind of used it as a fun exercise.

Functional programming is fun! Pyre being written in OCaml was one of the reasons I joined this team in the first place haha.

The simple reduce ops like sum/any/all are already their own functions, so it's pretty rare to actually use reduce directly I think. These days I only really use reduce if 1) it's chained after a map or filter 2) I'm reducing to a single value 3) the reduce operation is simple.

PR #395, opened 2025-05-28
Fix InitVar parameter type checking in dataclass constructors

### LINKED ISSUE #390: Support dataclasses.InitVar
### Describe the Bug

Defining a dataclass with an `InitVar` field causes pyrefly to insist every value assigned to the field to also be `InitVar`. Instead, pyrefly should behave as if `InitVar` were not there, i.e. pyrefly should interpret `mode: InitVar[str]` as `mode: str`.

```
from dataclasses import dataclass, field, InitVar

@dataclass
class Data:
    value: int = field(init=False)
    mode: InitVar[str]

    def __post_init__(self, mode: str):
        if mode == "number":
            self.value = 10
        else:
            self.value = 0

d = Data("number")
```

Result:
```
ERROR 14:10-18: Argument `Literal['number']` is not assignable to parameter `mode` with type `InitVar[str]` in function `Data.__init__` [[bad-argument-type](https://pyrefly.org/en/docs/error-kinds/#bad-argument-type)]
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBAJghgFzgYwDZwM4YKYagSwgAcwQFZEV0sAaKYfbVGOgSQDt8EA1OEAKH4ABeEjSYM-cVigARSgC5+UFVABucVAFdsCgu3IBeeo2YAKfJwSGAYppwBKZaohgYuqBy68QAbQwIIAC6gqqw2MBQAPpRJAFRllwxZjiowHSu7noBIA5KYWH4kZnYUIbGAETsWhAARtggFfkFLanAAHQa2qXGAIwADM4tTDjNLWFtnZo6ZVCD-DCz8khmVTX1jU5EIJYIZjBT3U5AA

_No response_

Adist319:
## Fix InitVar parameter type checking in dataclass constructors

Fixes dataclass constructors incorrectly expecting `InitVar[T]` arguments instead of `T` arguments, violating PEP 557 specification. fixes #390.

### Changes
- Modified `ClassField::as_param()` to detect `InitVar` types and extract inner type for constructor parameters  
- Added conformance test case `DC5` to verify `InitVar[str]` accepts `str` arguments

### Example
```python
@dataclass
class Data:
    mode: InitVar[str]
    def __post_init__(self, mode: str): ...

Data("number")  # Now works correctly ✅
```

### DIFF
--- a/conformance/third_party/dataclasses_postinit.py
+++ b/conformance/third_party/dataclasses_postinit.py
@@ -53,3 +53,19 @@ class DC4(DC3):
 
     def __post_init__(self, _name: str, _age: int):
         ...
+
+
+@dataclass
+class DC5:
+    value: int = field(init=False)
+    mode: InitVar[str]
+
+    def __post_init__(self, mode: str):
+        if mode == "number":
+            self.value = 10
+        else:
+            self.value = 0
+
+
+# This should work - passing str to InitVar[str] parameter
+dc5 = DC5("number")  # Should not error, but Pyrefly incorrectly expects InitVar[str]
--- a/conformance/third_party/dataclasses_postinit.py
+++ b/conformance/third_party/dataclasses_postinit.py
@@ -53,19 +53,3 @@ class DC4(DC3):
 
     def __post_init__(self, _name: str, _age: int):
         ...
-
-
-@dataclass
-class DC5:
-    value: int = field(init=False)
-    mode: InitVar[str]
-
-    def __post_init__(self, mode: str):
-        if mode == "number":
-            self.value = 10
-        else:
-            self.value = 0
-
-
-# This should work - passing str to InitVar[str] parameter
-dc5 = DC5("number")  # Should not error, but Pyrefly incorrectly expects InitVar[str]
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -210,26 +210,15 @@ impl ClassField {
     }
 
     pub fn as_param(self, name: &Name, default: bool, kw_only: bool) -> Param {
-        let ClassField(ClassFieldInner::Simple { ty, annotation, .. }) = self;
-        
-        // For InitVar fields, use the inner type instead of InitVar[T]
-        let param_ty = if let Some(annotation) = &annotation
-            && annotation.qualifiers.contains(&Qualifier::InitVar)
-        {
-            // The type is already unwrapped by the qualifier processing
-            ty
-        } else {
-            ty
-        };
-        
+        let ClassField(ClassFieldInner::Simple { ty, .. }) = self;
         let required = match default {
             true => Required::Optional,
             false => Required::Required,
         };
         if kw_only {
-            Param::KwOnly(name.clone(), param_ty, required)
+            Param::KwOnly(name.clone(), ty, required)
         } else {
-            Param::Pos(name.clone(), param_ty, required)
+            Param::Pos(name.clone(), ty, required)
         }
     }
 
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -211,14 +211,25 @@ impl ClassField {
 
     pub fn as_param(self, name: &Name, default: bool, kw_only: bool) -> Param {
         let ClassField(ClassFieldInner::Simple { ty, .. }) = self;
+
+        // For InitVar fields, use the inner type instead of InitVar[T]
+        let param_ty = if let Type::ClassType(cls) = &ty
+            && cls.has_qname("dataclasses", "InitVar")
+            && let [inner_ty] = cls.targs().as_slice()
+        {
+            inner_ty.clone()
+        } else {
+            ty
+        };
+
         let required = match default {
             true => Required::Optional,
             false => Required::Required,
         };
         if kw_only {
-            Param::KwOnly(name.clone(), ty, required)
+            Param::KwOnly(name.clone(), param_ty, required)
         } else {
-            Param::Pos(name.clone(), ty, required)
+            Param::Pos(name.clone(), param_ty, required)
         }
     }
 
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -210,18 +210,18 @@ impl ClassField {
     }
 
     pub fn as_param(self, name: &Name, default: bool, kw_only: bool) -> Param {
-        let ClassField(ClassFieldInner::Simple { ty, .. }) = self;
-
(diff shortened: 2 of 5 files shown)


conformance/third_party/dataclasses_postinit.py
rchen152, line 1: The tests in conformance/ are a copy of the conformance test suite from https://github.com/python/typing/tree/main/conformance and shouldn't be manually modified. Could you add your new test to https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/dataclasses.rs instead?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -211,14 +211,25 @@ impl ClassField {
 
     pub fn as_param(self, name: &Name, default: bool, kw_only: bool) -> Param {
         let ClassField(ClassFieldInner::Simple { ty, .. }) = self;
+
+        // For InitVar fields, use the inner type instead of InitVar[T]
```
rchen152: InitVar is a type qualifier (https://typing.python.org/en/latest/spec/annotations.html#type-and-annotation-expressions), so I'd prefer to modify `expr_qualifier` here to detect InitVar as a qualifier: https://github.com/facebook/pyrefly/blob/509d35b65f6b06a43af874024a5d13dea58e10bc/pyrefly/lib/alt/solve.rs#L255

There already exists a Qualifier::InitVar variant (https://github.com/facebook/pyrefly/blob/509d35b65f6b06a43af874024a5d13dea58e10bc/pyrefly/lib/types/annotation.rs#L78) ready to use.

rchen152 (commented):
Thanks! Although this works, I think it would be cleaner to mark InitVar as a type qualifier (see in-line comment).

yangdanny97:
Probably out of scope for this PR, but I think InitVar also means that it's not an actual field, it's just a parameter in the constructor.

So in a followup we should make sure that the dataclass doesn't actually have any InitVars as members, and it's only used for generating init/post_init

https://docs.python.org/3/library/dataclasses.html#dataclasses-init-only-variables

pyrefly/lib/alt/class/class_field.rs
Adist319: Thanks for the feedback! I agree that handling `InitVar` as a qualifier would be cleaner architecturally. However, since `InitVar` is not a `_SpecialForm` in Python (it's just a regular generic class), the existing qualifier detection mechanism won't work for it without some changes to accomodate InitVar. Correct me if my understanding is wrong.

Below is also my understanding on the qualifier system:
## How Qualifier Detection Currently Works

The qualifier system currently only handles `SpecialForm` types. Here's the relevant code from `expr_qualifier()`:

```rust
// In expr_qualifier() function
if let Some(Type::Type(box Type::SpecialForm(special))) = ty {
    let qualifier = special.to_qualifier();
    // ... validation logic ...
}
```

And in `special_form.rs`, only specific `SpecialForm` variants can become qualifiers:

```rust
pub fn to_qualifier(self) -> Option<Qualifier> {
    match self {
        Self::Annotated => Some(Qualifier::Annotated),
        Self::ClassVar => Some(Qualifier::ClassVar),
        Self::Final => Some(Qualifier::Final),
        Self::NotRequired => Some(Qualifier::NotRequired),
        Self::ReadOnly => Some(Qualifier::ReadOnly),
        Self::Required => Some(Qualifier::Required),
        Self::TypeAlias => Some(Qualifier::TypeAlias),
        _ => None,
    }
}
```

So, my question now is would it be appropriate to mark InitVar as SpecialForm too to work with the current qualifier system?

For now, I've moved the test to https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/dataclasses.rs as suggested in my latest commit. Do you know of any other approach that would be more architecturally sound?

---

yangdanny97: I think you could modify expr_qualifier to match on the class type, something similar to this: https://github.com/facebook/pyrefly/blob/509d35b65f6b06a43af874024a5d13dea58e10bc/pyrefly/lib/alt/expr.rs#L802

we could check `has_qname("dataclasses", "InitVar")` or something like that. similar to what you're doing here but detecting ClassDef instead of ClassType

rchen152: Thanks for moving the test! +1 to Danny's suggestion; you should be able to add a check for InitVar to expr_qualifier without touching any of the SpecialForm logic.

Adist319: Thanks for the help! I added a check for `InitVar` in `expr_qualifier` as suggested, matching on `Type::ClassDef(cls)` and using `cls.has_qname("dataclasses", "InitVar")`. This avoids touching any of the `SpecialForm` logic. Let me know if you have any further suggestions!

Edit: Added test case in https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/dataclasses.rs to test the new multiple arguments in InitVar error message.

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -210,15 +210,26 @@ impl ClassField {
     }
 
     pub fn as_param(self, name: &Name, default: bool, kw_only: bool) -> Param {
-        let ClassField(ClassFieldInner::Simple { ty, .. }) = self;
+        let ClassField(ClassFieldInner::Simple { ty, annotation, .. }) = self;
+        
+        // For InitVar fields, use the inner type instead of InitVar[T]
+        let param_ty = if let Some(annotation) = &annotation
```
rchen152: I don't think you need the change to this file anymore; the if and else branches here are identical.

pyrefly/lib/alt/expr.rs
```diff
@@ -935,6 +935,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     // (https://typing.readthedocs.io/en/latest/spec/annotations.html#type-and-annotation-expressions)
                     Type::type_form(Type::type_form(targ))
                 }
+                // Handle dataclasses.InitVar[T] - should behave as just T
```
rchen152: Pyrefly already catches this issue: https://pyrefly.org/sandbox/?code=GYJw9gtgBAJghgFzgYwDZwM4YKYagSwgAcwQEoBJAO3wQDU4QAoJ62hkAbXyoQBooGBCAC6QA. I don't think this change is needed.

pyrefly/lib/alt/class/class_field.rs
rchen152: The check in expr_qualifier and the new test look good to me. I don't think the edits in class_field.rs and expr.rs are needed; could you revert those?

Adist319:
@rchen152 Good catch, I did feel that I was overcomplicating things. I've reverted the changes in `class_field.rs` and `expr.rs`. Thanks!

PR #408, opened 2025-05-30
Move dotfile exclusion from project_excludes to core globbing logic

### LINKED ISSUE #321: Dotfiles should be included in project excludes always
### Describe the Bug

A file that starts with `.` can never be importable by Python. We should probably make this a part of `project_excludes`, even when overwriting from a config.

### Sandbox Link

_No response_

_No response_

Adist319:
Move dotfile exclusion from `project_excludes` to `Glob::files()`

Files starting with '.' can never be importable by Python, so this
exclusion should be built into the core globbing logic rather than
relying on configuration. This ensures dotfiles are always excluded
even when project_excludes is overridden.

Resolves #321

### DIFF
--- a/pyrefly_util/src/globs.rs
+++ b/pyrefly_util/src/globs.rs
@@ -911,7 +911,9 @@ mod tests {
 
         // Helper function to assert that a glob pattern returns no files
         let assert_empty_glob = |pattern_str: &str, description: &str| {
-            let found_files = Glob::new_with_root(root, pattern_str.to_owned()).files().unwrap_or_else(|_| Vec::new());
+            let found_files = Glob::new_with_root(root, pattern_str.to_owned())
+                .files()
+                .unwrap_or_else(|_| Vec::new());
             assert!(
                 found_files.is_empty(),
                 "{} should be excluded, found: {:?}",
--- a/pyrefly_util/src/globs.rs
+++ b/pyrefly_util/src/globs.rs
@@ -32,6 +32,12 @@ use crate::prelude::SliceExt;
 use crate::prelude::VecExt;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Default)]
+
+/// A glob pattern for matching files.
+///
+/// Only matches Python files (.py, .pyi) and automatically excludes:
+/// - Files that don't have .py or .pyi extensions
+/// - Files whose names start with '.' (dot files)
 pub struct Glob(PathBuf);
 
 impl Glob {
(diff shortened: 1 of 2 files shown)


yangdanny97:
Thanks for the PR, we'll get this reviewed soon.

pyrefly_util/src/globs.rs
```diff
@@ -32,6 +32,12 @@ use crate::prelude::SliceExt;
 use crate::prelude::VecExt;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Default)]
+
+/// A glob pattern for matching files.
+///
+/// Only matches Python files (.py, .pyi) and automatically excludes:
+/// - Files that don't have .py or .pyi extensions
+/// - Files whose names start with '.' (dot files)
```
connernilsen, line 40: Awesome! Thanks for updating these docs!

pyrefly_util/src/globs.rs
```diff
@@ -862,12 +879,107 @@ mod tests {
...
-                "a/.dotfile.py",
                 "a/__pycache__/g.py",
                 "a/__pycache__/h.pyi",
                 "c/j/k.py",
             ],
         )
         .unwrap();
     }
+
+    #[test]
```
connernilsen, line 890: Thanks for adding tests!

pyrefly_util/src/globs.rs
```diff
@@ -862,12 +879,107 @@ mod tests {
...
+        // Test explicit dot file exclusion
+        let dotfile_direct = Glob::new_with_root(root, "a/.dotfile.py".to_owned()).files();
+        match dotfile_direct {
+            Err(_) => {} // Expected: no files found
+            Ok(files) => assert!(
+                files.is_empty(),
+                "Direct dot file path should be excluded, found: {:?}",
+                files
+            ),
+        }
```
connernilsen: It seems like the next couple of lines are following this logic almost exactly. Can we turn this into a local function? Something like

```rust
let assert_empty_glob = |pattern_str: &str, name: &str| {
  let pattern = Glob::new_with_root(root, pattern_str);
  ...
    assert!(
      files.is_empty(),
      "Expected empty for {name}, got ...", // or however you want this to be worded
      files,
    ),
};
```

Or, since we're in a test, you can also unwrap the pattern directly since a panic would be okay here, which would also shorten this enough that you wouldn't really need a helper function
```rust
let found_files = Glob::new_with_root(root, <pattern>.to_owned()).files().unwrap();
assert!(found_files.is_empty(), <message>, ...);
```

pyrefly_util/src/globs.rs
```diff
@@ -100,10 +106,28 @@ impl Glob {
...
+    /// Returns true if the given file should be included in results.
+    /// Filters out non-Python files and dot files.
+    fn should_include_file(path: &Path) -> bool {
+        // Check if it's a Python file
+        if !Self::is_python_extension(path.extension()) {
+            return false;
+        }
+
+        // Check if it's a dot file
+        if let Some(file_name) = path.file_name().and_then(OsStr::to_str) {
```
connernilsen, line 118: We could do something like`Glob::contains_glob_char` here to avoid the possibly fallible operation of converting to a `&str`, but I don't feel super strongly about this though, so up to you if you want to change it over to that.

Adist319:
Thanks for the comments @connernilsen! Will work on this again later this week.

Adist319:
Pushed my changes, I decided to go with the local function approach.

connernilsen (approved):
Looks great, thanks for doing this!

PR #416, opened 2025-06-01
Error if function with @deprecated decorator is called

### LINKED ISSUE #402: [feature] warn on calling deprecated functions
### Describe the Bug

https://typing.python.org/en/latest/spec/directives.html#deprecated

See parent issue for context.

### Sandbox Link

_No response_

_No response_

Alex-Aron:
Resolves #402 

Added a check for if a function deprecated. Currently throws the error without the string in `@deprecate("string")`. Still throws the error if a string is given, just doesn't show the string.


website/src/sandbox/pyrefly_wasm.js
Alex-Aron, line 1: I assume I was not supposed to commit these changes I just noticed they were there too my apologies.

pyrefly/lib/error/kind.rs
```diff
@@ -79,6 +79,8 @@ pub enum ErrorKind {
     DeleteError,
     /// An error related to the import machinery.
     /// e.g. failed to import a module.
+    Deprecated,
+    /// An error raised if a function that is deprecated is used.
```
yangdanny97: you've inserted this in between the comment & definition for `ImportError`; the comment should go above

website/src/sandbox/pyrefly_wasm.js
yangdanny97: yes, please revert the changes to this file

pyrefly/lib/alt/function.rs
```diff
@@ -64,6 +64,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let skip_implementation = self.module_info().path().style() == ModuleStyle::Interface
             || class_metadata.is_some_and(|idx| self.get_idx(*idx).is_protocol());
         let def = self.get_idx(idx);
+        if def.metadata.flags.is_deprecated {
```
yangdanny97: hmm, doesn't this emit the error where the function is defined, not where it is called?

If the deprecated function exists and nobody calls it, there shouldn't be an error.

yangdanny97:
Hi! Thanks for putting up this PR. I left a few comments.

Additionally, could you please add an integration test in this file? https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/calls.rs

pyrefly/lib/alt/function.rs
Alex-Aron: yes you are right, I will fix this

pyrefly/lib/error/kind.rs
```diff
@@ -77,6 +77,8 @@ pub enum ErrorKind {
     BadUnpacking,
     /// Attempting to `del` something that cannot be deleted
     DeleteError,
+    /// An error rasied if a function that is deprecated is called.
```
srchilukoori: @Alex-Aron #nit if you don't mind please fix the spelling of raised here.

yangdanny97: "Calling a function marked with @deprecated"

is probably more concise. We can edit the comment later once we add the import check

pyrefly/lib/alt/function.rs
```diff
@@ -173,6 +174,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .iter()
             .filter(|k| {
                 let decorator = self.get_idx(**k);
+                if decorator.ty().to_string() == *"deprecated".to_owned() {
```
yangdanny97: are we able to use the CalleeKind mechanism to detect this decorator? I thought we could just add a new CalleeKind; we probably don't want to be matching on stringified types.

(see the other decorator logic for overload, member, final, etc. right below what you added)

yangdanny97 (changes requested):
Thanks for the edits, one more change + a typo and after that it should be good.

pyrefly/lib/alt/function.rs
Alex-Aron: I tried updating the logic to handle "deprecated".to_owned() using callee.kind but ran into issues due to how @deprecated("text") is implemented as a decorator factory.

Type(deprecated("text")) returns warnings.deprecated, not function, unlike other decorators like @final, which are directly functions. This makes @deprecated behave differently (more like a callable class ) and I suspect it’s being handled differently in our current callee resolution logic.

I also tried treating deprecated as a ClassKind, but that didn’t work either , possibly due to missing logic in how we handle factory-style decorators. I'm not entirely sure where in the pipeline factories like this are resolved (maybe in the type resolution phase?), so I left the .to_owned() as-is for now to keep the code compiling.

Would appreciate any pointers on where exactly decorator factories should be handled . Is there a way we detect and type check those? Sorry for the trouble, the project is huge and I am still getting up to speed on type checking /  rust as a whole

yangdanny97:
thanks @Alex-Aron !

Issue #360, opened 2025-05-26
pyrefly gets really slow when run in environment with a lot of dependencies


hauntsaninja:
### Describe the Bug

I have a small 3k line project with not many dependencies

If I install its dependencies in a fresh environment, pyrefly runs in about 200ms:
```
 INFO 17 errors shown, 4 errors ignored, 19 modules, 485 transitive dependencies, 143,623 lines, took 0.20s, peak memory physical 130.6 MiB
```

However, if I run in a large monolithic Python virtual environment with over a thousand installs, pyrefly runs in ~30s:
```
 INFO 17 errors shown, 4 errors ignored, 19 modules, 17,920 transitive dependencies, 8,128,708 lines, took 26.33s, peak memory physical 1.9 GiB
```

The project is in a closed source repo, but I will try to see if I can make an open source reproduction.

In the meantime, is there anything I can do to figure out what the extra transitive dependencies pyrefly is checking are? `--verbose` doesn't seem to make a difference

pyrefly 0.16.3

lolpack:
Hi @hauntsaninja thanks for reaching. I wonder if the issue is ignoring the correct files? We have tasks for https://github.com/facebook/pyrefly/issues/319 and https://github.com/facebook/pyrefly/issues/321. I wonder if you have generated artifacts or binaries that are getting pulled into to analysis that should not be.

hauntsaninja:
By "environment" I mean Python virtual environment. It was the same project checkout. From the output attached, pyrefly is discovering the same number of modules to check, just the number of transitive dependencies explodes. 

Is there a way to get insight into what pyrefly is checking? --verbose was not revealing

lolpack:
Can you try excluding the venv directory with project_excludes https://pyrefly.org/en/docs/configuration/#project_excludes ? @connernilsen is there a way to output all of the search paths that we pull in? I don't think there is a way to show the transitive dependency graph today.

hauntsaninja:
I'll try and report back! But the venv location in my case is out of tree

connernilsen:
@lolpack we have `pyrefly dump-config` which will tell us what we analyze for a given config. It enumerates `project_includes`, and lets us know the `site_package_path` (along with a few other things).

@hauntsaninja, is the project you're working on open source? Would we be able to take a look/try to replicate it ourselves or is that not possible?

Also, do you have a `pyrefly.toml` or `[tool.pyrefly]` section in a `pyproject.toml`? Can you upload that here if so?

[...]

Issue #319, opened 2025-05-20
Automatically include `.gitignore` in `project_excludes`


connernilsen:
We want to be able to include `.gitignore` globs in `project_excludes` by default to reduce the amount of manual configuration that needs to be done.

Right now, we have some [basic, hard-coded defaults](https://github.com/facebook/pyrefly/blob/2363b48a0ad520336a0e86e93746cb4d28a5fa0d/pyrefly/lib/config/config.rs#L345) to ignore certain files/patterns we notice aren't useful to search for type checking. We should expand this behavior to include everything we can find in a `.gitignore`. 

This will likely involve three steps:
1. Performing an [upward search](https://github.com/facebook/pyrefly/blob/main/pyrefly_util/src/upward_search.rs) for a `.gitignore` file
2. Parsing the results of the `.gitignore` file and turning them into two `Globs`, one for includes and another for excludes
3. Wiring it into the config, most likely in [`Config::configure()`](https://github.com/facebook/pyrefly/blob/2363b48a0ad520336a0e86e93746cb4d28a5fa0d/pyrefly/lib/config/config.rs#L463)

BaoNguyen09:
i'm a new contributor, and want to work on this. Do i need to assign myself or just start?

yangdanny97:
@BaoNguyen09 I can assign you, go ahead and start, and let us know if you have any questions

Thanks!

connernilsen:
Hey @BaoNguyen09, thanks for picking this up! Please let me know if you have any questions with this, and feel free to [join our Discord server](https://discord.gg/Cf7mFQtW7W) for quicker feedback (you can DM me directly if you need).

BaoNguyen09:
For sure, i'm setting up the project and will ask you as soon as i need anything. Thank you for the support!

connernilsen:
Hey @BaoNguyen09, just wanted to check in and see how it's going. Have you been able to take a look yet? Don't forget that we're here if you have any questions about getting it set up/implementation approaches/....

BaoNguyen09:
yeah i'm working on it right now, last week was busy but i finished setting up and understanding the issue now

connernilsen:
Awesome, thanks for the update @BaoNguyen09!

connernilsen:
[...]

PR #440, opened 2025-06-06
Don't enumerate files and directories that are excluded by a FilteredGlob

connernilsen:
Summary:
[An issue was reported](https://github.com/facebook/pyrefly/issues/360) that mentioned installing many dependencies slows down Pyrefly by a lot.

While we have [a task](https://github.com/facebook/pyrefly/issues/319) to address that by adding `.gitignore` entries into `project-excludes`, I realized that the way I implemented `FilteredGlobs` won't handle this well. We'll be ignoring them, but **we still have to enumerate everything first**, which means the `.gitignore` task will only help with reducing errors, not speeding up type checking.

This diff addresses that problem by threading our `project-excludes` as a `Globs` filter through our enumeration logic, and checking if a given glob or path matches, before iterating through the glob or path (if it's a directory).

This does cause us to lose messaging around "all found files were filtered by project excludes", but I think this is worth the trade. We can always add something back later if no files are returned that performs the same logic at the cost of slowing down the runtime massively.

---

Test on [Pyp](https://github.com/hauntsaninja/pyp) with [these dependencies](https://github.com/facebook/pyrefly/issues/360#issuecomment-2946645042):

Before
```
INFO ... transitive dependencies: 498, ... time: 20.89s (listing 19.01s; checking 1.49s) ...
```
(I've also seen durations up to as long as 50s!)

After
```
INFO ... transitive dependencies: 498, ... time: 1.26s ...
```

Differential Revision: D76149499

### DIFF
--- a/pyrefly_util/src/globs.rs
+++ b/pyrefly_util/src/globs.rs
@@ -124,31 +124,38 @@ impl Glob {
         true
     }
 
-    fn resolve_path(path: PathBuf, results: &mut Vec<PathBuf>) -> anyhow::Result<()> {
+    fn resolve_path(
+        path: PathBuf,
+        results: &mut Vec<PathBuf>,
+        filter: &Globs,
+    ) -> anyhow::Result<()> {
+        if filter.matches(&path)? {
+            return Ok(());
+        }
         if path.is_dir() {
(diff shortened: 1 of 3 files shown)


rchen152 (approved):

Issue #430, opened 2025-06-04
[unclear error message] overloads


kinto0:
### Describe the Bug

example described in sandbox comments

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMpgBuApiADZgCGAJgDRQAySMplZ9AysWcPSMSTYB9eAmIAocQGIoxAB6VsZYlACM02AAsV-NlATgARsujb+mAM5QARBeIwAYmDBRqxYKmLUoh97hVEpBQ0FtbibsBQdo7OABR2PPSEbACuxPSoZJ4AlABc4lCFmJGZnvlFFVD8MCkgKFHcwAVF3HbllYXVtfUAcmAoEuIAAoHkVNTh7g0xYPGNSanEuZgoMBkoWQPLTCwgbADaDmx2ALrZUAC0AHxQXDzLAHRPkiMkYzSTkdFOswm8UMkyGllqg1itNktGMxWGR9gAVEBpM6XG59LZQJ4PSQyeSKBDKKAAJmGo2CEwi0wAQpQQHNEgDFiDVusIdtoXtYUcyKdztdbo1Hs8SW8yZ8qTS6f9AcCVmDSuidjD4YjiMi+WjIZjsVodMQ9AYwMZiKZSCokFYYJpzVBXkFxq5iABjXCUGC4KKaMApMjeFBgfC+KBmCS297kqbRam0v4LIHpcE5dqFJAlDZlZodLp1Bo8DOyblLPMVLO9foSIA

yangdanny97:
Something like: "Outside of a stub file, overloaded functions must have a concrete implementation not decorated with `@overload`"

ZPyrolink:
I think the error `Overloaded function needs at least two signatures` could also be clearer in some use cases

## Sandbox
[...]

PR #443, opened 2025-06-07
FEAT: Func implementation must come after all @overload declarations

iwakitakuma33:
https://github.com/facebook/pyrefly/issues/430

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -65,27 +65,86 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             || class_metadata.is_some_and(|idx| self.get_idx(*idx).is_protocol());
         let def = self.get_idx(idx);
         if def.metadata.flags.is_overload {
+            if !skip_implementation && def.stub_or_impl == FunctionStubOrImpl::Impl {
+                self.error(
+                    errors,
+                    def.id_range,
+                    ErrorKind::InvalidOverload,
+                    None,
+                    "@overload decorator should not be used on function implementations.".to_owned(),
+                );
+            }
+
             // This function is decorated with @overload. We should warn if this function is actually called anywhere.
             let successor = self.bindings().get(idx).successor;
             let ty = def.ty.clone();
             if successor.is_none() {
                 // This is the last definition in the chain. We should produce an overload type.
                 let mut acc = Vec1::new((def.id_range, ty));
-                let mut first = def;
-                while let Some(def) = self.step_overload_pred(predecessor) {
-                    acc.push((def.id_range, def.ty.clone()));
-                    first = def;
+                let mut has_overload_after = false;
+                let mut has_implementation_before_overload = false;
+                let mut has_any_implementation = false;
+                let mut temp_pred = *predecessor;
+                while let Some(current_pred_idx) = temp_pred {
+                    let mut current_binding = self.bindings().get(current_pred_idx);
+                    while let Binding::Forward(forward_key) = current_binding {
+                        current_binding = self.bindings().get(*forward_key);
+                    }
+                    if let Binding::Function(func_idx, next_predecessor, _) = current_binding {
+                        let func_def = self.get_idx(*func_idx);
+                        
+                        if func_def.metadata.flags.is_overload {
+                            has_overload_after = true;
+                        }
+                        if func_def.stub_or_impl == FunctionStubOrImpl::Impl {
+                            has_any_implementation = true;
+                            if !func_def.metadata.flags.is_overload {
+                                has_implementation_before_overload = true;
+                            }
+                        }
+                        if has_overload_after && has_any_implementation && has_implementation_before_overload {
+                            break;
+                        }
+                        temp_pred = *next_predecessor;
+                    } else {
+                        break;
+                    }
+                }
+
+                let mut first = def.clone();
+                while let Some(predecessor_def) = self.step_overload_pred(predecessor) {
+                    acc.push((predecessor_def.id_range, predecessor_def.ty.clone()));
+                    first = predecessor_def;
                 }
                 if !skip_implementation {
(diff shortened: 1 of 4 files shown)


kinto0:
Thank you!!! I'll let @rchen152 review this. Just FYI #430 included [both](https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMpgBuApiADZgCGAJgDRQAySMplZ9AysWcPSMSTYB9eAmIAocQGIoxAB6VsZYlACM02AAsV-NlATgARsujb+mAM5QARBeIwAYmDBRqxYKmLUoh97hVEpBQ0FtbibsBQdo7OABR2PPSEbACuxPSoZJ4AlABc4lCFmJGZnvlFFVD8MCkgKFHcwAVF3HbllYXVtfUAcmAoEuIAAoHkVNTh7g0xYPGNSanEuZgoMBkoWQPLTCwgbADaDmx2ALrZUAC0AHxQXDzLAHRPkiMkYzSTkdFOswm8UMkyGllqg1itNktGMxWGR9gAVEBpM6XG59LZQJ4PSQyeSKBDKKAAJmGo2CEwi0wAQpQQHNEgDFiDVusIdtoXtYUcyKdztdbo1Hs8SW8yZ8qTS6f9AcCVmDSuidjD4YjiMi+WjIZjsVodMQ9AYwMZiKZSCokFYYJpzVBXkFxq5iABjXCUGC4KKaMApMjeFBgfC+KBmCS297kqbRam0v4LIHpcE5dqFJAlDZlZodLp1Bo8DOyblLPMVLO9foSIA) this issue and another (likely more difficult to implement) one

```
@overload
def setBar(self, value: int, inline: Literal[False]) -> Self: ... # ERROR: Overloaded function must have an implementation

@overload
def setBar(self, value: int, inline: Literal[True]) -> None: ...

# the real problem here is this @overload decorator should not be here
@overload
def setBar(self, value, inline):
    if inline:
        return self
    else:
        return None
```

iwakitakuma33:
@kinto0 
Thank you for your reply!
Yes, I will look into issues that seem more difficult problem.

This is my first time contributing to this repository, so I started small.

yangdanny97:
@iwakitakuma33 

Thanks! Could you please add an integration test case in this file?

https://github.com/facebook/pyrefly/blob/fab3bdec8aa597a5e9c10436921d28f617e54f48/pyrefly/lib/test/overload.rs

iwakitakuma33:
@yangdanny97 
I added test.

I did fix to show error if concrete function has overload.
If the overload function has ... Is it OK .
If the overload function has pass , is it ok that show error??

pyrefly/lib/alt/function.rs
```diff
@@ -234,6 +256,16 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             self_type = self_type.map(Type::type_form);
         }
 
+        if is_overload && stub_or_impl == FunctionStubOrImpl::Impl {
+            self.error(
+                errors,
+                def.name.range,
+                ErrorKind::InvalidOverload,
+                None,
+                "@overload decorator should not be used on function implementations.".to_owned(),
```
rchen152: Similar to my other comment - this is a good error message, but we should report either this error or "Overloaded function must have an implementation," not both.

If you add a new `stub_or_impl` field to DecoratedFunction (https://github.com/facebook/pyrefly/blob/4a08127ee842d006904558dca4baecf9a3b12251/pyrefly/lib/alt/types/decorated_function.rs#L29), then you can move this check to solve_function_binding and do something like:
```
if !skip_implementation {
    if def.stub_or_impl == FunctionStubOrImpl::Impl {
        // report "@overload decorator should not be used on function implementations."
    else {
        // report "Overloaded function must have an implementation"
    }
}
```

pyrefly/lib/alt/function.rs
```diff
@@ -110,7 +110,29 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                }
+                current_successor = self.bindings().get(succ_key_idx).successor;
+            }
+            if has_overload_successor && !skip_implementation {
+                self.error(
+                    errors,
+                    implementation_range,
+                    ErrorKind::InvalidOverload,
+                    None,
+                    "Function implementation must come after all @overload declarations. ".to_owned(),
```
rchen152: I like this error message, but I think it's confusing that this error and "Overloaded function must have an implementation" are both reported; we should report one or the other.

What I'd suggest doing is: when we encounter an `@overload`-decorated signature with no successors, we loop through the predecessors to see if there's an implementation among them. If there is, then we report this error rather than the "Overloaded function must have an implementation" error. Here's an example of how to loop through predecessors: https://github.com/facebook/pyrefly/blob/4a08127ee842d006904558dca4baecf9a3b12251/pyrefly/lib/alt/function.rs#L485-L502

rchen152 (changes requested):
Thank you for the PR! It would be great if the new error messages could be reported instead of the confusing "Overloaded function must have an implementation" messages, rather than both messages being shown. I left in-line comments with suggestions for how to do that.

pyrefly/lib/alt/function.rs
iwakitakuma33: @rchen152 
Thank you for your feedback. I have made revisions based on your comments.
I have also added tests.

When I revised the code to prevent the message “Overloaded function needs at least two signatures” from appearing unnecessarily, evaluating get_idx in the successor direction caused lazy evaluation to occur, causing tests other than overload to fail. Therefore, in this PR,I have limited the scope to cases where successor.is_none is true.

rchen152: Sorry for the delay! I somehow missed a bunch of GitHub notifications this past week. I'll review this tomorrow.

rchen152:
Thanks @iwakitakuma33!

PR #447, opened 2025-06-08
Prevent Required and NotRequired qualifiers together in TypedDict fields

### LINKED ISSUE #414: [feature] don't allow Required & NotRequired to appear at the same time as a qualifier for TypedDict fields
### Describe the Bug

Something like this should not be allowed:

```
class TD(TypedDict):
    x: Required[NotRequired[int]]
    y: NotRequired[Required[int]]
```

The check should probably go here: https://github.com/facebook/pyrefly/blob/c6a5b667065373cdad63e3fbadaba1baa0675998/pyrefly/lib/alt/solve.rs#L469

Once this is done, please add a test to: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/typed_dict.rs

### Sandbox Link

_No response_

_No response_

ajaymiranda:
### **Summary:**

This PR resolves #414  by adding a check and testcase to prevent `Required` and `NotRequired` from being used together as qualifiers in TypedDict fields. 

### **Changes:**  

- Implemented a validation check in `solve.rs` to reject conflicting qualifiers.
- Added a testcase in `typed_dict.rs` to cover this case.


pyrefly/lib/alt/solve.rs
```diff
@@ -467,6 +467,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         "`ClassVar` may not be nested inside `Final`".to_owned(),
                     );
                 }
+
+                let has_required = ann.qualifiers.contains(&Qualifier::Required);
+                let has_not_required = ann.qualifiers.contains(&Qualifier::NotRequired);
+                if (qualifier == Qualifier::Required && has_not_required)
```
ndmitchell: This code first checks if the `required` is present (I think using a linear scan), but only uses that result if `qualifier == Qualifier::Required`. It would be more efficient to inline `has_required` and `has_not_required`, then you only do the scan if you are on a relevant qualifier.

ndmitchell (commented):
This looks great! I suggested a minor performance tweak, but definitely a good change.

yangdanny97:
I'd expect `test.py` to generate some changes to the conformance outputs - was this not the case?

pyrefly/lib/alt/solve.rs
ajaymiranda: Certainly! Thanks, I’ll update the code.

ajaymiranda:
> I'd expect `test.py` to generate some changes to the conformance outputs - was this not the case?

Apologies for the oversight. The outputs were generated, but I missed committing them after running `test.py`. I'll add them, thanks.

Issue #413, opened 2025-05-31
[feature] disallow non-frozen dataclass from extending a frozen dataclass, and vice versa


yangdanny97:
### Describe the Bug

To check whether a class is a frozen dataclass based on metadata, do something like:

```
metadata.dataclass_metadata().is_some_and(|dataclass| dataclass.kws.is_set(&DataclassKeywords::FROZEN))
```

The check should probably go somewhere in this function or some helper called by it:

https://github.com/facebook/pyrefly/blob/c6a5b667065373cdad63e3fbadaba1baa0675998/pyrefly/lib/alt/class/class_metadata.rs#L150

Once this is done, please add a test to: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/dataclasses.rs

PR #466, opened 2025-06-11
Validate frozen dataclass inheritance rules

pt2302:
## Description

This PR adds validation to frozen dataclass inheritance rules. Specifically, frozen dataclasses are not allowed to extend non-frozen dataclasses, and non-frozen datataclasses are not allowed to extend frozen dataclasses.

## Relevant Issue
Closes https://github.com/facebook/pyrefly/issues/413.

## Testing Instructions

```

@dataclass(frozen=<True/False>)
class ParentClass:
    x: int

@dataclass(frozen=<True/False>)
class ChildClass(ParentClass):
    y: str
```
and run `pyrefly check` on the 4 possible combinations of frozen/non-frozen. Verify that errors are produced in the non-frozen/frozen and frozen/non-frozen cases, and that no errors are produced in the other two cases.


pyrefly/lib/alt/types/class_metadata.rs
```diff
@@ -96,6 +109,46 @@ impl ClassMetadata {
...
+    }
+
+    fn validate_frozen_dataclass_inheritance(
+        cls: &Class,
+        is_current_frozen: bool,
+        bases_with_metadata: &[(ClassType, Arc<ClassMetadata>)],
+        errors: &ErrorCollector,
+    ) {
+        for (base_type, base_metadata) in bases_with_metadata {
+            if base_metadata.dataclass_metadata().is_some() {
```
kinto0, line 123:
```suggestion
            if let Some(dataclass_metadata) = base_metadata.dataclass_metadata() {
```

this syntax lets you refer to it as non-None below

pyrefly/lib/alt/types/class_metadata.rs
```diff
@@ -48,6 +49,7 @@ pub struct ClassMetadata {
     /// Is it possible for this class to have type parameters that we don't know about?
     /// This can happen if, e.g., a class inherits from Any.
     has_unknown_tparams: bool,
+    is_frozen_dataclass: bool,
```
kinto0, line 52: this is already stored in `dataclass_metadata` so this new entry is unnecessary. with the other suggestion below, it's not an issue to grab the information from dataclass_metadata

pyrefly/lib/alt/types/class_metadata.rs
```diff
@@ -96,6 +109,46 @@ impl ClassMetadata {
...
+                    } else {
+                        "non-frozen"
+                    };
+
+                    errors.add(
+                        cls.range(),
+                        ErrorKind::InvalidInheritance,
+                        None,
+                        vec1![format!(
+                            "Cannot inherit {} dataclass `{}` from {} dataclass `{}`",
```
kinto0, line 143: maybe we should say why we can't? I will let @rchen152 decide how best to word it

kinto0 (approved):
thank you for the contribution!! I left a few minor notes which were resolved when I imported it.

pyrefly/lib/alt/types/class_metadata.rs
```diff
@@ -82,6 +84,17 @@ impl ClassMetadata {
         errors: &ErrorCollector,
     ) -> ClassMetadata {
         let mro = Mro::new(cls, &bases_with_metadata, errors);
+        let is_frozen_dataclass = dataclass_metadata
+            .as_ref()
+            .is_some_and(|dataclass| dataclass.kws.is_set(&DataclassKeywords::FROZEN));
```
stroxler, line 89: I think it would be slightly more readable to pass `dataclass_metadata` down to `validate_frozen_dataclass_inheritance` and compute `is_frozen_dataclass` there.

That keeps the responsibility for dealing with frozen-ness all in one function

stroxler (approved):

pyrefly/lib/alt/types/class_metadata.rs
kinto0: nvm, this is actually pretty clear

PR #477, opened 2025-06-13
Add a test to keep error-kinds.mdx in sync with the ErrorKind enum

rchen152:
Summary: It's much too easy for to add a new ErrorKind variant and forget to update the docs. Add a test to catch this.

Differential Revision: D76565757


connernilsen (approved):
Ooh, this is a cool addition. I might add something like this for config files too soon

Issue #366, opened 2025-05-26
Pandas imports


ZPyrolink:
#### Description
Depending on the import path used, `pyrefly` infers different type hints for the same objects in `pandas`.

For example:
```py
import pandas as pd

df = pd.DataFrame([1,2,3]) # pyrefly detect Unknown
columns = df.columns # pyrefly detect Unknown
```
or
```py
import pandas.core.api as pd

df = pd.DataFrame([1,2,3]) # pyrefly correctly detect DataFrame
columns = df.columns # but pyrefly detect Unknown
```

To have the correct typing:
```py
import pandas.core.frame as pd

df = pd.DataFrame([1,2,3]) # pyrefly correctly detect DataFrame
columns = df.columns # pyrefly correctly detect Index
```

#### Commentary
I understand that internally we have imports of imports:
```py
# On pandas
from pandas.core.api import DataFrame

# On pandas.core.api
from pandas.core.frame import DataFrame

# On pandas.core.frame
class DataFrame(...): ...`
```
The issue might come from here

#### Environment
- `pandas` 2.2.3
- `pandas-stubs` 2.2.3.250308

ndmitchell:
Using the development version of Pyrefly it seems that these all now give identical answers, correctly inferring `Index[str]`. Do you still see the same bug with Pyrefly 0.20.0?

ZPyrolink:
I just tested it now with pyrefly 0.20.1 and indeed the first and the last example works perfectly. However, I still have the same result with the second one (importing from `pandas.core.api`).

I also notice another errors (`Unexpected keyword argument 'columns' in function 'pandas.core.generic.NDFrame.__init__'` and 'Missing argument 'data' in function `pandas.core.generic.NDFrame.__init__'`). These errors appear only on the last example (importing from `pandas.core.frame`). I don't know if the problem is related or deserves a separate issue.

connernilsen:
Fun fact, type checking 
```python
import pandas.core.api as pd

df = pd.DataFrame([1,2,3]) # pyrefly correctly detect DataFrame
columns = df.columns # but pyrefly detect Unknown
```

With just `pandas-stubs` installed works fine, but it stops working once you install `pandas`. Trying to help @yangdanny97 with a minimal repro right now.

connernilsen:
[...]

PR #478, opened 2025-06-13
Add fallback alias resolution for pandas.DataFrame via trace_alias_chain

myselfRaifMondal:
Overview
This PR addresses [Issue #366](https://github.com/facebook/pyrefly/issues/366), where Pyrefly fails to correctly infer the type of pd.DataFrame when imported via the public API (import pandas as pd). Currently, inference only works reliably with deeper internal imports like pandas.core.frame.

What This PR Does
Adds a new fallback resolution mechanism inside attr_infer() for unresolved attributes.
Introduces a helper function trace_alias_chain() which attempts to trace known re-exported aliases.
Implements a temporary hardcoded resolution for:
import pandas as pd
df = pd.DataFrame([1, 2, 3])  # ✅ now correctly inferred as pandas.core.frame.DataFrame
Implementation Details
Hooked into the attr_infer_for_type flow within pyrefly/lib/alt/expr.rs.
Added Type::from_known(path: &str) in types.rs for building known types.
Enhanced TypeInfo::of_ty() to build minimal TypeInfo with required fields.
Introduced a new test in lib/test/attributes.rs to verify alias resolution.
Future Work
Generalize trace_alias_chain() to dynamically read from pandas-stubs or module re-export maps.
Add broader support for other common aliases like pd.Series, pd.Index, etc.
Motivation
Improving Pyrefly’s public-API type inference unlocks better static analysis for real-world Python code, particularly in data science, where pandas is ubiquitous.

🧪 Test Coverage

 Added test: test_resolve_dataframe_alias
 Verified cargo test passes with patch applied


stroxler:
Thanks for putting up a PR! Before reviewing it in detail I want to start a conversation about whether this is the right approach, or whether there's a bug in Pyrefly's import logic we need to fix.

I'm not opposed to hardcoding pandas support if we absolutely need it, but ideally this ought to work out of the box, and it seems like it is pointing us at something not working properly in the core system.

@ndmitchell I think at one point you were looking at Pandas support, but I don't recall details of what you found. Did you determine what was causing #366?

ndmitchell:
Thanks for the fix, appreciate it! In this case I would have expected our alias stuff to work automatically, but it doesn't seem to have done so. Let me take a brief look at this before we add any specific overrides (probably tomorrow morning)

ndmitchell:
Discussing on the underlying #366

myselfRaifMondal:
I will get it fixed wait a second

myselfRaifMondal:
@stroxler @grievejia @rchen152 I would like to know that my updated code works or not or there need to be any changes further.

yangdanny97:
@connernilsen is working on a fix from the import resolution/search path side, so I don't think this PR will be necessary. Thanks!

Issue #464, opened 2025-06-11
Add CLI flag for `ignore-errors-in-generated-code`


connernilsen:
Add a CLI flag and environment variable overriding [`ignore-errors-in-generated-code`](https://github.com/facebook/pyrefly/blob/086fe6b2799f18dcfd3ccc3afcfff856a00d6d04/pyrefly/lib/config/base.rs#L61). See #458 for more details.

ducdetronquito:
Hello !

I would like to work on this issue, is it OK ? :)

connernilsen:
Hey @ducdetronquito, thanks for offering to take a look at this issue! Let us know if you run into any issues.

Issue #458, opened 2025-06-11
Flags for all `ConfigFile` options


connernilsen:
[This is `ConfigFile`](https://github.com/facebook/pyrefly/blob/bac2bae04df1fe738793f66027f0a0aa3ed2628d/pyrefly/lib/config/config.rs#L164-L285). `ConfigFile` configures the settings in Pyrefly, and it's what a `pyrefly.toml` or `pyproject.toml` is read into and passed around.

Currently, a few of the config options (`python_interpreter`, `project_includes`, ...) are configurable by command flags and environment variables, but it would be nice if everything could be configurable. We read in CLI flags (especially those for config overrides) with [`ConfigOverrideArgs`](https://github.com/facebook/pyrefly/blob/086fe6b2799f18dcfd3ccc3afcfff856a00d6d04/pyrefly/lib/commands/check.rs#L157-L176), and override the `ConfigFile` in [`check::Args::override_config()`](https://github.com/facebook/pyrefly/blob/086fe6b2799f18dcfd3ccc3afcfff856a00d6d04/pyrefly/lib/commands/check.rs#L524-L544).

Let's add the missing config overrides so they can be configured by CLI flags or env vars, and make sure to override them in the config file. We'll also need to update our [configuration documentation](https://github.com/facebook/pyrefly/blob/main/website/docs/configuration.mdx) with the changes that are made.

Rules for new config override flags that are added:
- Pretty much just follow the existing format 😃 
- If a config option has `serde(skip)`, it is not meant to be serialized or deserialized. Please don't add it.
- All configs should have a `clap_env()` env. This creates a `PYREFLY_`-prefixed env override that clap checks for.

If you'd like to add one of these config overrides, just add yourself to the sub-task and get started!

Let us know if you have any questions about how you plan to implement this, or if you run into issues along the way! You can do that directly on the sub-task or in our [Discord server](https://discord.gg/Cf7mFQtW7W)!

TheRustyPickle:
@connernilsen Thank you for creating a detailed guide for the issue. I wanted to ask if it is allowed to create a single PR tackling two or more of the sub-issues? If yes, I would like to go ahead with it otherwise, get assigned on the next sub-issue on #460.

connernilsen:
Hey @TheRustyPickle, these PRs will probably be small enough that it shouldn't be an issue. Go ahead and comment on the tasks you want to pick up (I can't assign you to them otherwise) and I'll add you.

connernilsen:
[...]

PR #481, opened 2025-06-13
Add CLI flag for `ignore-errors-in-generated-code` (#464)

ducdetronquito:
Add a CLI flag `ignore-errors-in-generated-code` to override the value from the config.

Resolves: #464

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -187,6 +187,9 @@ struct ConfigOverrideArgs {
     /// Ignore missing source packages when only type stubs are available, allowing imports to proceed without source validation.
     #[arg(long, env = clap_env("IGNORE_MISSING_SOURCE"))]
     ignore_missing_source: Option<bool>,
+    // Whether to ignore type errors in generated code.
+    #[arg(long, env = clap_env("IGNORE_ERRORS_IN_GENERATED_CODE"))]
+    ignore_errors_in_generated_code: Option<bool>,
 }
 
 impl OutputFormat {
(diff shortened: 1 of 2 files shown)


pyrefly/lib/config/config.rs
```diff
@@ -802,7 +816,7 @@ mod tests {
                         (ErrorKind::AssertType, true),
                         (ErrorKind::BadReturn, false)
                     ]))),
-                    ignore_errors_in_generated_code: Some(true),
+                    ignore_errors_in_generated_code: None,
```
ducdetronquito: I'm not sure about this change as I don't understand why the field  `ignore_errors_in_generated_code` of the root config was `Some(true)` in the first place: [by looking at ConfigBase definition](https://github.com/facebook/pyrefly/blob/27f4e8a0f2103094ede6ba8ba3b558bb7a1892e1/pyrefly/lib/config/base.rs#L61) I would have expected it to be `None` already.

I would appreciate if someone could enlighten me on this matter :)

ducdetronquito: My bad, I did not understand that I had to set the ` ignore_errors_in_generated_code` field of the root config instead of create a new `ignore_errors_in_generated_code` field on ConfigFile.

connernilsen (commented):
Looks great, thanks for doing this. I'll pull it in and get it merged

connernilsen (approved):

ducdetronquito:
Thank you for the detailed issue #458; it was really helpful to get started !

rchen152 (approved):

PR #484, opened 2025-06-13
Allow blank github issues

### LINKED ISSUE #475: Create other issue templates
We should have a blank issue template because I'm getting annoyed with the additional formatting I have to remove when adding feature specs.

Maybe another `Feature` template too?

I'm mostly using this here as a reminder to do it

connernilsen:
Summary:
I get annoyed having to fill out the bug form and clean it up after creating an issue when it's something unrelated to a bug.

We should add non-bug issues back. I don't think we need much more structure than just a blank issue for things other than bugs.

Reviewed By: ndmitchell

Differential Revision: D76603522

Resolves #475

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,1 +1,1 @@
-blank_issues_enabled: false
+blank_issues_enabled: true


connernilsen:
Resolves #475

PR #485, opened 2025-06-14
Handle ReadOnly TypedDict subtyping

### LINKED ISSUE #333: Non-ReadOnly TypedDict fields should behave invariantly in subtyping
### Describe the Bug

```python
from typing import Any, Dict

from typing_extensions import ReadOnly, TypedDict

class TD(TypedDict):
    a: int

class TD2(TypedDict):
    a: ReadOnly[int]

class TD3(TypedDict):
    a: bool

def want_td(td: TD) -> None:
    pass

def capybara(
    td: TD, td2: TD2, td3: TD3
) -> None:
    want_td(td2)  # E: incompatible_argument
    want_td(td3)  # E: incompatible_argument

```

Both of the `want_td` calls should be rejected. A TypedDict with a ReadOnly[X] key is not assignable to one with an equivalent X key, because the latter allows assignment. For similar reasons, all (non-ReadOnly) items should be treated as invariant in assignability checks.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFACJIDGMAULaJLIqmgPoCmAHjBygM5IwAzNlz4AShwCGAEwDyKADakoAFUQdZlGvSpLp-fuvIAKDQi06YASgBctKE6jS7mFHVr7DxteQBM5pra1LYOzi5uUnKKKgDaqDAAunoGRiYAzEGWITT2js6uUABGYGBK9LIcwFAA7tIebDCyps1ufjZQALQAfFAAcsIc4c4IPpXVUFTSCHDF0iDSpgVObSZkzf7tARuyGdsZtJ29A0MjTvWNza2y-p1QAMRQAKJuqFSQYzBIxUocbAs0ABXCB8OgRS4wJotZoZe5PV7uD7YaTfX7-QEgsG0IA

_No response_

xl4624:
Fixes #333 and handles:
* ReadOnly[X] assigned to equivalent X key
* non-ReadOnly items treated as invariants

I also referenced
> Read-only items behave covariantly, as they cannot be mutated. This is similar to container types such as Sequence, and different from non-read-only items, which behave invariantly.

from https://typing.python.org/en/latest/spec/typeddict.html#id4

### DIFF
--- a/pyrefly/lib/solver/subset.rs
+++ b/pyrefly/lib/solver/subset.rs
@@ -729,9 +729,16 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                 let want_fields = self.type_order.typed_dict_fields(want);
 
                 want_fields.iter().all(|(k, want_v)| {
-                    got_fields
-                        .get(k)
-                        .is_some_and(|got_v| self.is_subset_eq(&got_v.ty, &want_v.ty))
+                    got_fields.get(k).is_some_and(|got_v| {
+                        match (got_v.read_only, want_v.read_only) {
+                            // ReadOnly cannot be assigned to Non-ReadOnly
+                            (true, false) => false,
+                            // Non-ReadOnly fields are invariant
+                            (false, false) => self.is_equal(&got_v.ty, &want_v.ty),
+                            // ReadOnly `want` fields are covariant
+                            (_, true) => self.is_subset_eq(&got_v.ty, &want_v.ty),
+                        }
+                    })
                 }) && got_fields.iter().all(|(k, got_v)| {
                     want_fields
                         .get(k)
--- a/pyrefly/lib/test/typed_dict.rs
+++ b/pyrefly/lib/test/typed_dict.rs
@@ -238,9 +238,27 @@ def foo(a: Coord, b: Coord3D, c: Pair):
     coord: Coord = b
     coord2: Coord3D = a  # E: `TypedDict[Coord]` is not assignable to `TypedDict[Coord3D]`
     coord3: Coord = c  # E: `TypedDict[Pair]` is not assignable to `TypedDict[Coord]`
-    coord4: Pair = a
+    coord4: Pair = a  # E: `TypedDict[Coord]` is not assignable to `TypedDict[Pair]`
     "#,
 );
+
+testcase!(
+    test_typed_dict_readonly_subtype,
+    r#"
+from typing import ReadOnly, TypedDict
+
+class TD(TypedDict):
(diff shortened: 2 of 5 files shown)


pyrefly/lib/test/typed_dict.rs
```diff
@@ -238,7 +238,7 @@ def foo(a: Coord, b: Coord3D, c: Pair):
     coord: Coord = b
     coord2: Coord3D = a  # E: `TypedDict[Coord]` is not assignable to `TypedDict[Coord3D]`
     coord3: Coord = c  # E: `TypedDict[Pair]` is not assignable to `TypedDict[Coord]`
-    coord4: Pair = a
+    coord4: Pair = a  # E: `TypedDict[Coord]` is not assignable to `TypedDict[Pair]`
```
yangdanny97, line 241: could you please add a case checking covariance for readonly fields?

pyrefly/lib/solver/subset.rs
```diff
@@ -729,9 +729,16 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
...
                 want_fields.iter().all(|(k, want_v)| {
-                    got_fields
-                        .get(k)
-                        .is_some_and(|got_v| self.is_subset_eq(&got_v.ty, &want_v.ty))
+                    got_fields.get(k).is_some_and(|got_v| {
+                        match (got_v.read_only, want_v.read_only) {
+                            // ReadOnly cannot be assigned to Non-ReadOnly
+                            (true, false) => false,
+                            // Non-ReadOnly fields are invariant
+                            (false, _) => self.is_equal(&got_v.ty, &want_v.ty),
```
yangdanny97: I think if want is ReadOnly here it should be covariant, like so:

```
from typing import TypedDict, ReadOnly

class RO(TypedDict):
    x: ReadOnly[object]

class RW(TypedDict):
    x: int

def test(rw:RW):
    _ro: RO = rw  # OK
```

so the cases would be
- (_, true) -> covariant
- (true, false) -> not allowed
- (false, false) -> invariant

pyrefly/lib/alt/call.rs
```diff
@@ -6,6 +6,7 @@
  */
 
 use dupe::Dupe;
+use regex_syntax::hir::print;
```
yangdanny97: I think this is unnecessary

xl4624: Oops, thanks for catching 💀

rchen152 (approved):

PR #488, opened 2025-06-14
Add CLI flag to override typeshed

### LINKED ISSUE #483: Add CLI flag to override typeshed
I forgot to add a `--typeshed`-like flag to `pyrefly check`, which will let us override typeshed. This is useful for testing typeshed upgrades, testing how changes to typeshed will affect a codebase, and other typeshed-related changes.

This will involve a couple of steps. For whomever takes this on, I can turn this into more sub-tasks if that would make it easier for you!
1. Add a CLI flag somewhere around 
https://github.com/facebook/pyrefly/blob/cfb9561edef5787b0ffbdc98ed953b25e4398dd1/pyrefly/lib/commands/check.rs#L161
2. Add a `ConfigFile` option for `typeshed` path
https://github.com/facebook/pyrefly/blob/cfb9561edef5787b0ffbdc98ed953b25e4398dd1/pyrefly/lib/config/config.rs#L164
3. Wire it into our `find_import()` logic
https://github.com/facebook/pyrefly/blob/cfb9561edef5787b0ffbdc98ed953b25e4398dd1/pyrefly/lib/config/config.rs#L342-L346
4. You'l likely have to update our typeshed functionality (all in the file below) to handle loading a custom typeshed
https://github.com/facebook/pyrefly/blob/cfb9561edef5787b0ffbdc98ed953b25e4398dd1/pyrefly/lib/module/bundled.rs#L124-L129

krikera:
Add --typeshed-path flag to pyrefly check command for testing typeshed upgrades and custom typeshed modifications.

- Add typeshed_path CLI argument to ConfigOverrideArgs
- Add typeshed_path field to ConfigFile struct
- Wire custom typeshed lookup into find_import() logic
- Custom typeshed takes precedence over bundled typeshed

Fixes #483

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -174,6 +174,9 @@ struct ConfigOverrideArgs {
     /// after first checking `search_path` and `typeshed`.
     #[arg(long, env = clap_env("SITE_PACKAGE_PATH"))]
     site_package_path: Option<Vec<PathBuf>>,
+    /// Override the bundled typeshed with a custom path.
+    #[arg(long = "typeshed-path", env = clap_env("TYPESHED_PATH"))]
+    typeshed_path: Option<PathBuf>,
     /// The Python executable that will be queried for `python_version`
     /// `python_platform`, or `site_package_path` if any of the values are missing.
     #[arg(long, env = clap_env("PYTHON_INTERPRETER"), value_name = "EXE_PATH")]
@@ -554,6 +557,9 @@ impl Args {
         if let Some(x) = &self.config_override.site_package_path {
             config.python_environment.site_package_path = Some(x.clone());
             config.python_environment.site_package_path_source = SitePackagePathSource::CommandLine;
+        }
+        if let Some(x) = &self.config_override.typeshed_path {
+            config.typeshed_path = Some(x.clone());
         }
         if let Some(x) = &self.config_override.python_interpreter {
             config.python_interpreter = Some(x.clone());
--- a/pyrefly/lib/config/config.rs
+++ b/pyrefly/lib/config/config.rs
@@ -223,6 +223,10 @@ pub struct ConfigFile {
     )]
     pub fallback_search_path: Vec<PathBuf>,
 
+    /// Override the bundled typeshed with a custom path.
+    #[serde(skip)]
+    pub typeshed_path: Option<PathBuf>,
+
     // TODO(connernilsen): make this mutually exclusive with venv/conda env
     /// The python executable that will be queried for `python_version`,
     /// `python_platform`, or `site_package_path` if any of the values are missing.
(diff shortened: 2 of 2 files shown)


pyrefly/lib/config/config.rs
```diff
@@ -354,14 +359,38 @@ impl ConfigFile {
...
+                self.ignore_missing_source,
+            )? {
+                Ok(path)
+            } else {
+                Err(FindError::import_lookup_path(
+                    self.structured_import_lookup_path(),
+                    module,
+                    &self.source,
+                ))
+            }
```
connernilsen: Oops, it looks like this part might have been duplicated from below. Can you remove this?

pyrefly/lib/config/config.rs
```diff
@@ -354,14 +359,38 @@ impl ConfigFile {
...
         } else if let Some(path) = typeshed()
```
connernilsen: We should only use the builtin typeshed if one isn't provided. Can we add an extra condition for that here?

```
} else if self.typeshed_path.is_none() && let Some(path) = ...
```

pyrefly/lib/commands/check.rs
```diff
@@ -174,6 +174,9 @@ struct ConfigOverrideArgs {
     /// after first checking `search_path` and `typeshed`.
     #[arg(long, env = clap_env("SITE_PACKAGE_PATH"))]
     site_package_path: Option<Vec<PathBuf>>,
+    /// Override the bundled typeshed with a custom path.
+    #[arg(long = "typeshed-path", env = clap_env("TYPESHED_PATH"))]
```
connernilsen: nit: `arg(long, ...)` does the same thing as `arg(long = "typeshed-path", ...)`, but specifying `"typeshed-path"` is more prone to typos/less resilient to refactors. Can we just leave this as `arg(long, ...)`?

pyrefly/lib/config/config.rs
```diff
@@ -223,6 +223,10 @@ pub struct ConfigFile {
     )]
     pub fallback_search_path: Vec<PathBuf>,
 
+    /// Override the bundled typeshed with a custom path.
+    #[serde(skip)]
```
connernilsen: Can we make this like the other serde macros above? Probably instead of regular `skip`, do
```
#[serde(
  default,
  skip_serializing_if = "Vec::is_empty",
)]
```

pyrefly/lib/config/config.rs
```diff
@@ -354,14 +359,33 @@ impl ConfigFile {
...
+                self.ignore_missing_source,
+            )? {
+                Ok(path)
+            } else {
+                Err(FindError::import_lookup_path(
+                    self.structured_import_lookup_path(),
+                    module,
+                    &self.source,
+                ))
+            }
```
connernilsen: Hey @krikera, it looks like this part is still duplicated here. Can you delete this before I pull it in?

It's doing the same thing as lines 388-403 as far as I can tell.

stroxler (approved):

yangdanny97:
Thanks for your contribution @krikera !

Issue #468, opened 2025-06-11
[feature] ban TypedDict special form outside of base class list


yangdanny97:
### Describe the Bug

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQCpQC8UpiApgGoCGIAFAESlsA0UARmAFcUAEwo0EtYQBEkAYxgBKEsNrBYtAM4wWADwBc1OtLmK9RKBagJGGjSX2GJx+SRJA

When TypedDict is used outside of the base class list, it should give an invalid annotation error.

The task:
- change the code here https://github.com/facebook/pyrefly/blob/c351c566f0cd53ea2b06fbc7fd7f28d6c605c4e0/pyrefly/lib/types/special_form.rs#L90 to check the TypeFormContext instead of always  returning true
- add the cases from the above example as integration tests for pyrefly/lib/test/typed_dict.rs
- run test.py and commit any generated conformance changes

pt2302:
I would like to work on this issue.

PR #508, opened 2025-06-17
Disallow TypedDict special form outside of base class context

pt2302:
## Description

This PR disallows TypedDicts from being used outside of the list of base classes, generating invalid annotation errors if this is the case.

## Relevant Issue
Closes https://github.com/facebook/pyrefly/issues/468.

## Testing Instructions

```
from typing import TypedDict, TypeVar

T = TypeVar("T", bound=TypedDict)

def test(x: TypedDict):
    pass

x: TypedDict

```
and run `pyrefly check`. Verify that errors are generated.

### DIFF
--- a/pyrefly/lib/test/typed_dict.rs
+++ b/pyrefly/lib/test/typed_dict.rs
@@ -627,7 +627,7 @@ class TD(TypedDict):
 testcase!(
     test_typed_dict_isinstance_issubclass_not_allowed,
     r#"
-from typing import *
+from typing import TypedDict
 
 class D(TypedDict):
     x: int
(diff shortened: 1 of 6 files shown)


pyrefly/lib/test/typed_dict.rs
```diff
@@ -637,3 +637,28 @@ isinstance(x, D)  # E: TypedDict `D` not allowed as second argument to isinstanc
...
+    age: int
+    "#,
+);
+
+testcase!(
+    test_typeddict_invalid_annotations,
+    r#"
+from typing import TypedDict, TypeVar
+
+T = TypeVar("T", bound=TypedDict)  # E: Expected a type argument for `TypedDict`
```
yangdanny97, line 657: This error message is confusing, but fixing it may require some more refactoring so I think it's out of scope for this PR.

ndmitchell:
Thanks for the fix here!

Issue #456, opened 2025-06-10
Treat `.pyc` imports as `typing.Any` when no stubs can be found


connernilsen:
When a project uses `.pyc` imports, the source may not exist. In that case, when there are no analyzable source or stubs available, we probably shouldn't return an error on imports that can't be resolved, since we can't do the usual trick of inferring type annotations from unannotated code. There is no decompiled code for us to use!

Instead, we should use this logic: when we see a `<file>.pyc` import, but no non-`pyc` candidate, we should automatically perform the work of `replace_import_with_any`.

We'll likely want to do that by updating the [`FindResult`](https://github.com/facebook/pyrefly/blob/fffb8477b0ce8d4cb4eff311db2589fa3564c90d/pyrefly/lib/module/finder.rs#L33) to add a new variant for `.pyc` results, which when searching are lower precedence than `.pyi`, but higher than `NamespacePackages`.

We should also update the [configuration documentation](https://github.com/facebook/pyrefly/blob/fffb8477b0ce8d4cb4eff311db2589fa3564c90d/website/docs/configuration.mdx?plain=1#L265) to mention that we do this, since this behavior might be unexpected.

---

You can test this yourself with the following: 
1. `uv venv venv && source venv/bin/activate && pip install requests && cd $(python3 -c "import site; print(site.gesitepackages()[0])")/requests`
    - Note: this attempts to cd into the site directory that `requests` was installed into. If it was the wrong site package directory, then inspect other `site.getsitepackages()` directories to find the `requests` directory.
2. `python3 -m compileall *`
    - Note: you can also have the same effect as this by passing in `--compile` to `pip install requests`
4. `rm *.py && mv __pycache__/* . && rm -rf __pycache__`
5. `python3 -c 'import os; [os.rename(file, ".".join([part for part in file.split(".") if "cpython" not in part])) for file in os.listdir()]'`

The above should set up requests to look like a compiled library.

melvinhe:
Thanks for outlining the issue. I'm currently working on an internal fix to handle .pyc imports when no stubs are available. I'll keep you updated on the progress and ensure our tests and configuration docs reflect this new preferred behavior.

melvinhe:
Implementing an internal fix for this across multiple diff changes.

yangdanny97:
I think we have the pyc-as-Any change merged, so this issue should be good to close. Feel free to reopen if I'm wrong.

connernilsen:
[...]

PR #512, opened 2025-06-17
Treat .pyc imports as typing.Any when no stubs can be found

melvinhe:
Fixes https://github.com/facebook/pyrefly/issues/456

Summary:
Updates the module resolution logic to treat .pyc files as typing.Any when no source or stub files are available. Ensures that compiled Python files are handled gracefully without causing import errors due to missing type information.

Adds doc updates and unit test cases to verify the correct identification and handling of .pyc files in the module search process.

Differential Revision: D76854224


connernilsen:
Hey @melvinhe, I'm going to close this, since we're splitting this up and implementing it internally.

PR #525, opened 2025-06-18
Store `reveal_type` and other errors separately in `CollectedErrors`

### LINKED ISSUE #506: Error summary line counts `reveal_type` as an error
### Describe the Bug

When I run pyrefly over this file:

```py

reveal_type(42)
```

I get this output:

```
 INFO revealed type: Literal[42] [reveal-type]
 --> /.../blank.py:3:12
  |
3 | reveal_type(42)
  |            ----
  |
 INFO errors shown: 1, errors ignored: 0, modules: 1, transitive dependencies: 58, lines: 17,355, time: 0.19s, peak memory: physical 68.8 MiB
```

Note the final line saying "errors shown: 1".

### Sandbox Link

_No response_

_No response_

fangyi-zhou:
Per suggestion in https://github.com/facebook/pyrefly/issues/506#issuecomment-2980865884

This commit separates errors of kind `RevealType` from all shown errors
in `CollectedErrors`. By doing so, we can separate actual type errors
and pseudo-errors generated by calls to `reveal_type`. In the meantime,
this commit adds a function `shown_errors_and_reveal_types` to obtain
all diagnostics for displaying and reporting.

Closes #506.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -639,14 +639,15 @@ impl Args {
         }
 
         let errors = loads.collect_errors();
+        let shown_errors_and_reveal_types = errors.shown_errors_and_reveal_types();
         if let Some(path) = &self.output.output {
             self.output
                 .output_format
-                .write_errors_to_file(path, &errors.shown)?;
+                .write_errors_to_file(path, &shown_errors_and_reveal_types)?;
         } else {
             self.output
                 .output_format
-                .write_errors_to_console(&errors.shown)?;
+                .write_errors_to_console(&shown_errors_and_reveal_types)?;
         }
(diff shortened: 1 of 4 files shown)


yangdanny97 (changes requested):
Thanks for taking a stab at this!

There's a few CI issues that make me think the approach suggested by Conner may be trickier than anticipated - reveal-type no longer shows up in the sandbox with this change, and the diagnostics get reordered so all the reveal_types are at the end.

I think I have a simpler solution for this, so let's hold off for a bit.

fangyi-zhou:
I see, I was thinking that we can add some getter APIs to `CollectedErrors` for "real" errors only, but decided to go with the recommended approach in the issue thread.

PR #532, opened 2025-06-19
Enforce total as the only valid keyword for TypedDict

### LINKED ISSUE #476: [feature] TypedDict base class list should only allow the keyword "total"
### Describe the Bug

Emit a `BadTypedDict` error for any other keyword

Example: https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjAFBUA2AhgM4uHkCMAFEQmZRoAaWGBhMGAXgIgArsQCUALigJW7KAGIoAeQDS9Zmw4AmXiQrUYI4GDCTOy1eqhaoAOTG69QA

The code change probably goes here:
https://github.com/facebook/pyrefly/blob/27f4e8a0f2103094ede6ba8ba3b558bb7a1892e1/pyrefly/lib/alt/class/class_metadata.rs#L317

Add an integration test to pyrefly/lib/tests/typed_dict.rs

Run test.py and commit any generated changes.

### Sandbox Link

_No response_

_No response_

Adist319:
Fixes #476.
This PR makes sure that only the total keyword is allowed when defining a TypedDict. Before, Pyrefly would silently ignore other keywords, but now it will throw a BadTypedDict error, which matches the PEP 589 spec.

You'll probably notice the conformance test diff is pretty big. It looks like my local environment was out of sync with the main branch, so after fixing my build with a cargo clean and a rebase, the new test run corrected a lot of other unrelated test outputs. The final diff includes my TypedDict fix plus all the other corrections from the healthier build.


pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -315,6 +315,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 _ => Either::Right((n.clone(), self.expr_infer(x, errors))),
             });
         let typed_dict_metadata = if is_typed_dict {
+            // Validate that only 'total' keyword is allowed for TypedDict
+            for (name, _) in &keywords {
```
yangdanny97: we could probably set is_total in this loop to avoid iterating over it twice

yangdanny97:
The large conformance diff isn't expected. Does `test.py` work for you? That should generate the conformance test outputs.

Not a blocker for review though, since I can re-generate the conformance outputs locally + we have a merge driver that regens it automatically during a merge.

Adist319:
@yangdanny97 Thanks for the feedback! I've updated the logic to use a single loop for the is_total check as you suggested.
On a separate note, I wanted to let you know about a recurring issue I'm having with the test suite.

When I run python3 test.py, the unit tests all pass successfully. However, the script then fails during the conformance test step with the following error:
Failed to get conformance output
 INFO errors shown: 1,038, errors ignored: 6, modules: 136, transitive dependencies: 143, lines: 43,475, time: 2.45s (checking 1.91s), peak memory: physical 55 MiB
Expectations failed for my_local_path\pyrefly\conformance\third_party\aliases_explicit.py: expected 8 errors, but got 23


Failed to get conformance output for directory conformance/third_party

This happens consistently, even after running cargo clean.
I'd appreciate any guidance you might have.

yangdanny97:
@Adist319 is that the whole output? 

The "expectations failed" for that line just means that there were 8 `# E:` comments in the file but we emitted 23 errors - that's expected for any conformance test case we don't fully pass, but the script should be able to continue.

Adist319:
@yangdanny97 Yup, that was the whole output I received. In my local environment, the test.py script treats the Expectations failed message as a fatal error and exits immediately, so it never continues on to regenerate the conformance files.

yangdanny97:
That's very weird; what OS and version of python do you have?

Adist319:
I'm on WSL Ubuntu 24.04.1, and I'm on Python 3.12.3.

zsol:
I can repro this on Windows (not WSL). Let me make a new issue

PR #537, opened 2025-06-20
Implement a basic model for `functools.total_ordering`

### LINKED ISSUE #491: [pyrefly] special case handling for `functools.total_ordering` decorator
### Describe the Bug

When running type checking with pyrefly, and I have a class that's decorated by total_ordering (imported from functools), I am only defining \_\_eq__ and \_\_lt__, but I expect <= to work without issues. Instead I'm getting `<=` is not supported between `A` and `A`

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBMCuB2BjALmMAbAzlAlhADmCMlKsgIboD6RAJgKYg7wDmAUGwAJmU0gNNWbROnKZsAQQBcbKHKgNgUKlWY5kKgBSZ66YABooADym54yAJRQAtAD4oAOTDx6M+e6g69AOiNQAvMay8orKVPQAjlpeBlBgyAAWjKYARBIpVnZQAEZo6G4eciD0yLAg8J66wL4BgfFJIL7BcqEq6BpU2lWG9clQaRk29rkYBYXFpeWVPn4APHGJjE1s5AFQEppG-gCMFmzZaxtbAEx7K1CzgdlsQA

_No response_

fangyi-zhou:
This PR adds modelling of `functools.total_ordering` (closes #491), which allows a class with (at least) one rich comparison method to automatically have all other rich comparison methods.

At a high level, this PR incorporates the following changes:
1. Capture `is_total_ordering` in class metadata if the decorator is present.
2. Synthesize missing rich comparison methods for those decorated classes.
3. Report an error if the decorator is present but no rich comparison method is present.

I have some open issues that can be fixed either in this PR after receiving some code pointers, or in subsequent PRs.
Open issues (marked in FIXMEs):
- [x] Error is currently raised on class declaration, ideally it should be on the decorator
- [x] Type information for synthesized rich comparison methods are not following the comparison method they're synthesizing from: e.g. if `__lt__` is synthesized from `__gt__`, then they should have the same type. Currently we model the synthesizing order, but we are not giving the same type.
- [x] Classes that have synthetic methods via multiple pathways will not have all the synthetic methods. For example, it's possible to decorate a class with `@dataclass` and `@total_ordering` at the same time, and they should have synthesized methods from both decorators.

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -405,6 +405,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
         }
         let mut is_final = false;
+        let mut is_total_ordering = false;
         for decorator in decorators {
             let decorator = self.get_idx(*decorator);
             match decorator.ty().callee_kind() {
@@ -430,6 +431,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                             "@runtime_checkable can only be applied to Protocol classes".to_owned(),
                         );
                     }
+                }
+                Some(CalleeKind::Function(FunctionKind::TotalOrdering)) => {
+                    is_total_ordering = true;
                 }
                 _ => {}
             }
@@ -477,6 +481,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             is_new_type,
             is_final,
             has_unknown_tparams,
+            is_total_ordering,
             errors,
         )
     }
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -49,7 +49,7 @@ use crate::types::types::Type;
 
 /// Private helper type used to share part of the logic needed for the
 /// binding-level work of finding legacy type parameters versus the type-level
-/// work of computing inherticance information and the MRO.
+/// work of computing inheritance information and the MRO.
 #[derive(Debug, Clone)]
 pub enum BaseClass {
     TypedDict,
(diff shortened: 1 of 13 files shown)


yangdanny97:
> Error is currently raised on class declaration, ideally it should be on the decorator

For this, instead of using a boolean flag in the class metadata you can have a `TotalOrderingMetadata` that holds a range instead (or just an `Option<TextRange>` to keep it simple, with `is_total_ordering` being a method that checks that it's not None).

Then you can emit the error according to that range.

yangdanny97:
> Classes that have synthetic methods via multiple pathways will not have all the synthetic methods. For example, it's possible to decorate a class with `@dataclass` and `@total_ordering` at the same time, and they should have synthesized methods from both decorators.

We could probably add a merge method to ClassSynthesizedFields and calculate + merge total_ordering separately from the other stuff.

https://github.com/facebook/pyrefly/blob/39f10ddef76a9d8a135b8ba4ccd688cddd7b001c/pyrefly/lib/alt/types/class_metadata.rs#L281

fangyi-zhou:
> > Error is currently raised on class declaration, ideally it should be on the decorator
> 
> For this, instead of using a boolean flag in the class metadata you can have a `TotalOrderingMetadata` that holds a range instead (or just an `Option<TextRange>` to keep it simple, with `is_total_ordering` being a method that checks that it's not None).
> 
> Then you can emit the error according to that range.

Is there a way to get a text range of the decorator in `class_metadata_of`? I don't seem to find a way to find the text range from the decorator.

yangdanny97:
BindingClassMetadata doesn't give this information, but we could change its decorator representation from `pub decorators: Box<[Idx<Key>]>,` to `pub decorators: Box<[(Idx<Key>, TextRange)]>,` and populate the range when we create the binding.

`decorator_list` of the parsed node will have the ranges, so maybe we could edit this function to return the range along with the key

https://github.com/facebook/pyrefly/blob/231ac8d432416208ff10faf5d67ce57f935710f8/pyrefly/lib/binding/class.rs#L118

fangyi-zhou:
Thanks so much for the code pointers, @yangdanny97! Really appreciate your help!

Issue #513, opened 2025-06-18
`Enum[some_string]` is a type error


bigfootjon:
### Describe the Bug

This is a generalized version of https://github.com/facebook/pyrefly/issues/212

AFAIK this is the correct way to convert a string into an enum, but pyrefly rejects it

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBApgdgV2gSwgBzCALlAooiAKEIGMAbAQwGcqoBZAT3yQApmIBKALkKj6gBiAeSFQAvFABEwMGEnEAJjGBQA5jCwB9CA03xW+iJqpYQXKCZAcoAWgB89JgR78oIDQhBxH7ANqHjUwBdIA

yangdanny97:
The task is to extend the solution in https://github.com/facebook/pyrefly/commit/650ae0456335c1134b47b2844720d894a4d67fc9 to handle arbitrary strings instead of just literals.

Test cases can be added to the same place. Once tests pass, run test.py and commit any generated changes.

PR #545, opened 2025-06-22
Enable typechecking of Enum[str]

bigfootjon:
Fixes: #513

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -862,6 +862,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         )
                     }
                 }
+                Type::ClassDef(ref cls) if self.get_enum_from_class(cls).is_some() => {
+                    if self.is_subset_eq(
+                        &self.expr(slice, None, errors),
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/expr.rs
```diff
@@ -862,6 +862,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         )
                     }
                 }
+                Type::ClassDef(ref cls) if self.get_enum_from_class(cls).is_some() => {
+                    Type::ClassType(cls.as_class_type())
```
yangdanny97: Is this intended to work with arbitrary slices? Or do we want to check that the slice's type is a string?

bigfootjon: Yeah I think that would be better, but I can’t figure out how to do that without generating spurious type errors on other tests :(

any ideas on what to try?

yangdanny97: You can get the type of the slice with `self.expr(<expr>, None, errors)`, then use `is_subset_eq` to check that it's a subtype of `str` (`self.stdlib.str().clone().to_type()`)

Issue #550, opened 2025-06-23
Recognize attributes defined in Initialization Helper Methods


kimasplund:
### Describe the Bug

# Pyrefly Bug Report: False Positive for Initialization Helper Methods

## Summary
Pyrefly 0.20.2 incorrectly reports "implicitly-defined-attribute" errors when attributes are set in helper methods called from `__init__`. This is a common pattern for organizing complex initialization logic.

## Environment
- **Pyrefly version**: 0.20.2
- **Python version**: 3.12
- **OS**: Linux

## Minimal Reproducible Example

```python
from prometheus_client import Counter, Histogram, CollectorRegistry

class MetricsCollector:
    def __init__(self, registry=None):
        self.registry = registry or CollectorRegistry()
        # Delegate initialization to helper methods for organization
        self._init_performance_metrics()
        self._init_security_metrics()
    
    def _init_performance_metrics(self):
        """Initialize performance-related metrics."""
        # Pyrefly reports: ERROR Attribute `request_duration` is implicitly defined
        self.request_duration = Histogram(
            "app_request_duration_seconds",
            "Time spent processing requests",
            registry=self.registry
        )
        
        # Pyrefly reports: ERROR Attribute `request_count` is implicitly defined
        self.request_count = Counter(
            "app_requests_total", 
            "Total number of requests",
            registry=self.registry
        )
    
    def _init_security_metrics(self):
        """Initialize security-related metrics."""
        # Pyrefly reports: ERROR Attribute `auth_attempts` is implicitly defined
        self.auth_attempts = Counter(
            "app_auth_attempts_total",
            "Total authentication attempts",
            registry=self.registry
        )
    
    def track_request(self, duration: float):
        # These attributes ARE defined during construction
        self.request_duration.observe(duration)
        self.request_count.inc()
```

## Expected Behavior
- Pyrefly should recognize that helper methods called from `__init__` are part of the construction process
- No "implicitly-defined-attribute" errors should be reported for attributes set in these methods
- This is equivalent to setting the attributes directly in `__init__`

## Actual Behavior
Pyrefly reports errors for every attribute defined in the helper methods:
```
ERROR Attribute `request_duration` is implicitly defined by assignment in method `_init_performance_metrics`, which is not a constructor [implicitly-defined-attribute]
ERROR Attribute `request_count` is implicitly defined by assignment in method `_init_performance_metrics`, which is not a constructor [implicitly-defined-attribute]
[...]

Issue #549, opened 2025-06-23
False Positive "missing-attribute" on Union Types


kimasplund:
### Describe the Bug

# Pyrefly Bug Report: False Positive "missing-attribute" on Union Types

## Summary
Pyrefly 0.20.2 incorrectly reports "missing-attribute" errors when accessing methods on dictionary values within union types, even when proper type narrowing or initialization guarantees the correct type.

## Environment
- **Pyrefly version**: 0.20.2
- **Python version**: 3.12
- **OS**: Linux

## Minimal Reproducible Example

```python
from typing import Dict, Any, List, Union

def process_sla_result() -> Dict[str, Any]:
    # Initialize with explicit structure
    sla_result = {
        "sla_id": "sla-001",
        "sla_name": "Response Time SLA",
        "compliant": True,
        "violations": [],  # Clearly a list
        "metrics_checked": {}
    }
    
    # Pyrefly reports these as errors:
    # ERROR Object of class `bool` has no attribute `append` [missing-attribute]
    # ERROR Object of class `dict` has no attribute `append` [missing-attribute]
    sla_result["violations"].append({
        "type": "response_time",
        "severity": "high"
    })
    
    return sla_result

# Another example with TypedDict
from typing import TypedDict, List

class SLAResult(TypedDict):
    violations: List[Dict[str, Any]]
    compliant: bool

def process_typed() -> SLAResult:
    result: SLAResult = {
        "violations": [],
        "compliant": True
    }
    
    # Still reports error even with TypedDict
    result["violations"].append({"error": "timeout"})
    
    return result
```

## Expected Behavior
- Pyrefly should recognize that `sla_result["violations"]` is initialized as a list
- When using TypedDict, the type of `violations` is explicitly `List[Dict[str, Any]]`
- No "missing-attribute" error should be reported for the `append` method

## Actual Behavior
Pyrefly reports multiple errors:
```
ERROR Object of class `bool` has no attribute `append` [missing-attribute]
ERROR Object of class `dict` has no attribute `append` [missing-attribute]
ERROR Object of class `int` has no attribute `append` [missing-attribute]
ERROR Object of class `str` has no attribute `append` [missing-attribute]
```

## Root Cause Analysis
1. When accessing `dict[key]`, pyrefly considers all possible types in the union `Dict[str, Any]`
2. The `Any` type expands to a union including `bool | dict | int | str | list | ...`
3. Pyrefly checks each type in the union and reports errors for types that don't have `append`
[...]

Issue #548, opened 2025-06-23
Pydantic field overload resolution fails when using positional default values with constraint parameters


kimasplund:
### Describe the Bug

# Pyrefly Bug Report: Pydantic Field Overload Resolution Failure

## Summary
Pyrefly 0.20.2 fails to correctly resolve Pydantic v2 `Field()` function overloads when using positional default values with constraint parameters (ge, le, etc.).

## Environment
- **Pyrefly version**: 0.20.2
- **Pydantic version**: 2.8.0 - 2.11.5
- **Python version**: 3.12
- **OS**: Linux

## Minimal Reproducible Example

```python
from pydantic import BaseModel, Field

class DatabaseConfig(BaseModel):
    # This triggers "no-matching-overload" error in pyrefly
    port: int = Field(5432, ge=1, le=65535, description="Database port")
    
    # This also fails
    min_connections: int = Field(5, ge=1, le=100)
    
    # Workaround: Using explicit 'default' keyword works
    max_connections: int = Field(default=20, ge=1, le=1000)
```

## Expected Behavior
Pyrefly should recognize that `Field(5432, ge=1, le=65535)` matches the second overload of Pydantic's Field function:
```python
@overload
def Field(
    default: _T,  # First positional argument
    *,
    ge: SupportsGe | None = _Unset,
    le: SupportsLe | None = _Unset,
    ...
) -> _T: ...
```

## Actual Behavior
Pyrefly reports:
```
ERROR No matching overload found for function `pydantic.fields.Field` [no-matching-overload]
  Possible overloads:
  (default: EllipsisType, *, ...) -> Any
  (default: int, *, ...) -> int [closest match]
  ...
```

## Root Cause Analysis
1. Pydantic's Field function uses TypeVar `_T` in its overload: `def Field(default: _T, ...) -> _T`
2. Pyrefly appears to have difficulty inferring the TypeVar when the default value is passed as a positional argument
3. The error message shows pyrefly is looking for `default: int` specifically rather than understanding the generic TypeVar

## Impact
- False positive errors on valid Pydantic code
- Forces developers to use less idiomatic syntax (explicit `default=` keyword)
- Affects all Pydantic models using Field with constraints

## Suggested Fix
Improve TypeVar inference in overloaded functions, particularly when:
1. The TypeVar appears in both parameter and return type
2. Positional arguments are used with keyword-only parameters
3. The actual type can be inferred from the argument value

## Workarounds
[...]

PR #551, opened 2025-06-23
Add tests and partial fixes for issues #547, #548, #549, #550

### LINKED ISSUE #547: Cannot Handle None Checks Before Dictionary Access
### Describe the Bug

# Pyrefly Bug Report: Cannot Handle None Checks Before Dictionary Access

## Summary
Pyrefly 0.20.2 reports false positive errors when accessing dictionary methods after explicit None checks, failing to recognize that the type has been narrowed.

## Environment
- **Pyrefly version**: 0.20.2
- **Python version**: 3.12
- **OS**: Linux

```python
from typing import Optional, Dict, Any

class ErrorContext:
    def __init__(self):
        self.system_context: Optional[Dict[str, Any]] = None
    
    def update_context(self, data: Dict[str, Any]) -> None:
        # Explicit None check
        if self.system_context is not None:
            # Pyrefly reports: ERROR Cannot set item on type `Dict[str, Any] | None`
            self.system_context["updated"] = True
            
            # Also reports: ERROR Object of class `NoneType` has no attribute `get`
            value = self.system_context.get("key", "default")
        
        # Another pattern with explicit else
        if self.system_context:
            self.system_context["status"] = "active"
        else:
            self.system_context = {"status": "active"}
            
        # Even after initialization in else branch, still reports errors
        self.system_context["timestamp"] = "2024-01-01"

# Function example
def process_config(config: Optional[Dict[str, Any]]) -> str:
    # Pattern 1: Early return
    if config is None:
        return "no config"
    
    # At this point, config cannot be None
    # But pyrefly still reports: ERROR Object of class `NoneType` has no attribute `get`
    return config.get("setting", "default")

# Pattern 2: Ternary narrowing
def get_value(data: Optional[Dict[str, str]]) -> str:
    # Pyrefly doesn't recognize the narrowing here either
    return data.get("key") if data is not None else "none"
```

## Expected Behavior
- After `if x is not None:`, the type of `x` should be narrowed from `Optional[T]` to `T`
- After `if x:`, the type should be narrowed (though less precisely due to falsy values)
- No NoneType errors should be reported in code paths where None is impossible

## Actual Behavior
Pyrefly reports errors as if the None check doesn't exist:
```
ERROR Cannot set item on type `Dict[str, Any] | None` [cannot-set-item]
ERROR Object of class `NoneType` has no attribute `get` [missing-attribute]
ERROR Union type `Dict[str, Any] | None` cannot be indexed [index-error]
```

## Root Cause Analysis
1. Pyrefly's control flow analysis doesn't properly narrow Optional types after None checks
2. Type narrowing from isinstance(), `is None`, and truthiness checks isn't propagated
3. The type remains as `Optional[Dict]` throughout the entire scope

## Impact
- 28+ false positives for NoneType access in our codebase
- Forces unnecessary code changes or type: ignore comments
- Reduces confidence in pyrefly's type checking

## Suggested Fix
Implement proper type narrowing for:
1. `if x is None:` / `if x is not None:` checks
2. `if x:` / `if not x:` truthiness checks (with appropriate warnings for falsy values)
3. `isinstance(x, type)` checks
4. Early returns that eliminate None possibility

This is a fundamental feature of modern Python type checkers (mypy, pyright, etc.).

## Workarounds
1. Use assertion to narrow type:
   ```python
   assert self.system_context is not None
   self.system_context["key"] = value
   ```

2. Create new variable with narrowed type:
   ```python
   if self.system_context is not None:
       ctx: Dict[str, Any] = self.system_context
       ctx["key"] = value
   ```

3. Use type: ignore comments:
   ```python
   if self.system_context is not None:
       self.system_context["key"] = value  # type: ignore[index]
   ```

## Related Issues
This is likely related to the union type attribute access issue, but specifically affects Optional types which are extremely common in Python code.

## Additional Context
Type narrowing is a crucial feature for practical type checking. Without it, Optional types become very difficult to use, leading to either:
1. Excessive type: ignore comments
2. Unnecessary runtime assertions
3. Developers avoiding Optional types altogether

This significantly impacts pyrefly's usability for production Python code.

### Sandbox Link

_No response_

_No response_

kimasplund:
## Summary

This PR adds comprehensive test cases for four reported Pyrefly issues and includes a partial fix for issue #549.

### Issue #547: Dict narrowing after None checks
- ✅ **Status**: Already fixed in current version
- Added comprehensive tests in `issue_547_dict_narrowing.rs`
- Tests confirm proper type narrowing for `Optional[Dict]` after `is not None` checks
- Covers attribute chains, early returns, and nested narrowing

### Issue #548: Pydantic Field overload resolution  
- ⚠️ **Status**: Root cause identified, workaround documented
- Added tests in `issue_548_pydantic_field.rs` showing expected behavior
- Issue stems from Pydantic's use of sentinel values (`_Unset`) in type stubs
- **Workaround**: Use keyword arguments (`Field(default=value, ...)` instead of `Field(value, ...)`)

### Issue #549: False positives on union types with Any
- 🔧 **Status**: Partial fix implemented
- Added `type_contains_any()` helper in `attr.rs`
- When `Any` is present in a union, allows any attribute access without errors
- Tests in `issue_549_any_union.rs` verify the fix
- Note: `Dict[str, Any]` subscript still needs separate handling

### Issue #550: Init helper methods
- 📝 **Status**: Tests added, full fix requires flow analysis
- Added tests in `issue_550_init_helpers.rs` documenting the limitation
- Helper methods called from `__init__` aren't recognized as constructors
- Full fix would require tracking method calls from `__init__`

## Test Results

All new tests pass:
```
test test::issue_547_dict_narrowing::test_dict_none_narrow_attribute_chain ... ok
test test::issue_547_dict_narrowing::test_dict_none_narrow_early_return ... ok
test test::issue_547_dict_narrowing::test_dict_none_narrow_nested ... ok
test test::issue_548_pydantic_field::test_pydantic_field_positional ... ok
test test::issue_548_pydantic_field::test_pydantic_field_type_inference ... ok
test test::issue_549_any_union::test_any_union_attribute_access ... ok
test test::issue_549_any_union::test_any_union_method_call ... ok
test test::issue_549_any_union::test_non_any_union_attribute_access ... ok
```

## Benefits

- Improves test coverage for edge cases
- Documents both working features and known limitations
- Provides foundation for future improvements
- Partially fixes union+Any attribute access issues

## Related Issues

Fixes #547 (tests confirm it's already fixed)
Partially addresses #549 (union+Any case)
Documents workarounds for #548 and #550


yangdanny97:
Hi, thanks for the PR! Since the issues aren't related, would you be able to split this up? That would make it easier to review.

kimasplund:
sure. il do that :)

yangdanny97:
Thanks! I'll close this one and review the others individually.

Issue #324, opened 2025-05-21
[Behavior] Noisy errors when not explicitly defining types for **kwargs


cscanlin:
### Describe the Bug

This probably falls under "working as intended", but I'm curious if I'm missing an obvious solution here.

The main issue I see is that when I use a type like `dict[str, str | int]` and pass this through as kwargs using `**`, this actually causes more failures than when I use `dict[str, Any]`, even when the types I give match the types of the function (see sandbox link).

The clear solution to this is to use a TypeDict for args like this to define the shape/interface, but it's clunky to specify both this and the annotations for the function itself.

I would appreciate this level of explicitness much more if there was an easier way to reference the interfaces for these functions like in my 4th example. I'm curious if there's an obvious reason this hasn't been implemented, but I'm not sure if this is the right forum for that conversation.

Really awesome tool though, nice job team :)

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFACqICmAJgCJIDGMAUC4wDYCGAzj4SDQ8AFJQS0GzAJQAuFlAVQuMzClaKoAIxU8YINjSrAoEOAH1gAVxSNhy1TDLaoukFKgBaAHwu9cjSBUMJYgKFDAAET2AN5cAL5OKtGacRFsLADEUADuuADW-NkAFlRhVAAeelwU1PRM+BUIHExIMBxwUIbAqLQsXII8KgQDUAC8UNEA5FyTKgCMZJOas1CTJRwcYJNxLIE8lhww-OOmFta2AFQX-UJSbFnAXEgcVDxkVABupTklYfDi-FKlggVBAXBgvRuPAATCoaPUANquMiuKAAHwcAF0xhNpisFqtlio1lQNlsdnsDkdoTjTlYbMIrlDoXdMrAikh+DQwK8cvlYGAwOywJY0EU+gMAMxwxHIwjEbHjKYzeaLImrdabba7V5UmG08z0y7XKWsrLkDmFEUcGhaKhQRiCjhsqEAFhUdPOADoAJJqUGPRj2pV41WElYksnaoA

yangdanny97:
Hi, thanks for reporting this. The behavior you describe is not part of the typing spec, but we are aware that it's a common pattern and plan to implement some sort of support for it.

[...]

PR #557, opened 2025-06-24
fix: Allow attribute access on union types containing Any

### LINKED ISSUE #549: False Positive "missing-attribute" on Union Types

kimasplund:
## Summary

This PR fixes false positive errors when accessing attributes on union types that contain `Any`.

## Issue #549: False Positive "missing-attribute" on Union Types

When a union type contains `Any` (e.g., `A  < /dev/null |  Any`), Pyrefly was incorrectly reporting missing attribute errors for the non-Any members of the union. According to gradual typing principles, `Any` should allow any operation.

## Changes

1. Added `type_contains_any()` helper method that recursively checks if a type contains `Any`
2. Modified attribute access, assignment, and deletion to:
   - Check if the base type contains `Any`
   - If yes, return `Any` without generating errors
   - If no, proceed with normal type checking

## Tests

Added comprehensive tests in `issue_549_any_union.rs`:
- ✅ Attribute access on `A | Any` unions (no errors)
- ✅ Method calls on `A | Any` unions (no errors)  
- ✅ Attribute access on `A | B` unions without `Any` (still errors as expected)

## Limitations

This fixes the issue for explicit unions like `A | Any`. The case of `Dict[str, Any]` subscript access expanding to unions needs separate handling and is not addressed in this PR.

Partially fixes #549


erictraut:
The change implemented in this PR looks incorrect to me from the standpoint of type theory and the Python typing spec. 

I think the PR is based on an incorrect or incomplete understanding of unions and gradual types. For more details about how these work in the Python type system, refer to [this section of the typing spec](https://typing.python.org/en/latest/spec/concepts.html#union-types).

When accessing an attribute or a method on a union, that method or attribute must be accessible on _all_ subtypes within the union. If one of those subtypes is `Any`, a type checker still needs to verify that the attribute or method is accessible on the _other_ type in the union. This PR appears to disable all such checks if the union contains an `Any`. This will result in false negatives.

You can see that pyrefly's current behavior matches that of other type checkers including [pyright](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcAUCQCYCmwUwArigMYAUAHgFyYr4A%2BhxASnYkooqAhCoYbAHTAwYAWSA) and [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=824cca5d13642c6bb837bcc4e1eed980).

yangdanny97:
For the reasons @erictraut mentioned, I don't think we can merge this PR. I'm going to close this, since TypedDicts appear to work as expected and we can track the plan to handle regular dictionaries in https://github.com/facebook/pyrefly/issues/324

PR #558, opened 2025-06-24
test: Add tests documenting Pydantic Field behavior

kimasplund:
## Summary  

This PR adds tests documenting the expected behavior for Pydantic Field overload resolution.

## Issue #548: Pydantic Field Overload Resolution Failure

The issue reports that Pyrefly fails to match Pydantic Field overloads when using positional default arguments.

## Root Cause

The issue stems from how Pydantic uses sentinel values (`_Unset`/`PydanticUndefined`) in its type stubs. These sentinel values make parameters appear as having defaults when they should be treated as required from a caller's perspective.

## Tests Added

The test file `issue_548_pydantic_field.rs` includes:
- Tests showing expected usage with positional arguments
- Tests demonstrating TypeVar inference with different types
- Documentation of the workaround (using keyword arguments)

## Workaround

Users can work around this issue by using keyword arguments:
```python
# Instead of: Field(5432, ge=1)
# Use: Field(default=5432, ge=1)
```

Related to #548 - This PR documents the current behavior and workaround rather than fixing the underlying issue.


kimasplund:
Related to #548

yangdanny97 (changes requested):
Same here, this PR includes the changes from the other PRs.

yangdanny97:
@kimasplund I'm not seeing any overload errors show up when I run the test cases

PR #559, opened 2025-06-24
test: Add tests documenting init helper method limitation

kimasplund:
## Summary

This PR adds tests documenting the current limitation where helper methods called from `__init__` are not recognized as valid attribute-defining methods.

## Issue #550: False Positive for Initialization Helper Methods  

Pyrefly reports "implicitly-defined-attribute" errors when attributes are set in helper methods called from `__init__`, which is a common pattern for organizing initialization logic.

## Current Behavior

Pyrefly only recognizes these methods as valid for attribute definition:
- `__init__`, `__init_subclass__`, `__new__`
- For test classes: `setUp`, `setUpClass`, and related methods

Helper methods called from these recognized methods are not tracked.

## Tests Added

The test file `issue_550_init_helpers.rs` includes:
- Basic init helper pattern
- Conditional init helpers  
- Nested init helpers
- Test showing errors when helper is NOT called from init

## Future Enhancement

A full fix would require flow analysis to:
1. Track which methods are called from `__init__`
2. Mark those methods as part of the construction process
3. Allow attribute assignments in those methods

Related to #550 - This PR documents the current limitation to guide future implementation.


kimasplund:
Related to #550

PR #560, opened 2025-06-24
Fix bugs in attribute narrowing

### LINKED ISSUE #547: Cannot Handle None Checks Before Dictionary Access

yangdanny97:
Summary:
fixes https://github.com/facebook/pyrefly/issues/547

BUG 1: In the linked issue, we observed that the narrowing for an attribute was invalidated by assigning to a subscript of that attribute. This is because we accidentally check the narrowing information for `next_facet` instead of `facet` in `NarrowedFacets::update_for_assignment`, causing the program to take a different branch and overwriting the existing narrow.

BUG 2: Once I fixed the first bug, the attribute narrowing was no longer being wiped, but the subscript narrowing did not get applied. This is because the new branch the program takes hit another bug in `NarrowedFacet::update_for_assignment`, which stops execution if the current narrowed facet is a leaf (even if we want to narrow a subtree). I ended up inlining the fixed logic directly since the function is only called once.

Differential Revision: D77175069

### DIFF
--- a/pyrefly/lib/test/subscript_narrow.rs
+++ b/pyrefly/lib/test/subscript_narrow.rs
@@ -154,3 +154,39 @@ def test(x: dict[str, object], c1: C1, c2s: dict[str, C2], key: str, s: str):
     assert_type(c2s["key1"].x, object)
  "#,
 );
+
+testcase!(
+    test_subscript_narrow_does_not_invalidate_attribute,
+    r#"
+from typing import Optional, Dict, Any, assert_type, Literal
+
+class ErrorContext:
+    def __init__(self):
+        self.system_context: dict[str, Any] | None = None
+    
+    def update_context(self, data: dict[str, Any]) -> None:
+        # Explicit None check
+        if self.system_context is not None:
+            assert_type(self.system_context, dict[str, Any])
+
+            self.system_context["updated"] = True
+
+            assert_type(self.system_context, dict[str, Any])
+            assert_type(self.system_context["updated"], Literal[True])
+
+            value = self.system_context.get("key", "default")
+
+    def update_context_2(self, data: dict[str, Any]) -> None:
+        if self.system_context:
+            self.system_context["status"] = "active"
+        else:
+            self.system_context = {"status": "active"}
+
+        assert_type(self.system_context, dict[str, Any])
+            
+        self.system_context["timestamp"] = "2024-01-01"
+
+        assert_type(self.system_context, dict[str, Any])
+        assert_type(self.system_context["timestamp"], Literal["2024-01-01"])
+"#,
+);
--- a/pyrefly/lib/types/type_info.rs
+++ b/pyrefly/lib/types/type_info.rs
@@ -269,8 +269,17 @@ impl NarrowedFacets {
                 }
             }
             [next_facet, remaining_facets @ ..] => {
-                if let Some(narrowed_facet) = self.0.get_mut(next_facet) {
-                    narrowed_facet.update_for_assignment(next_facet, remaining_facets, ty);
+                if let Some(narrowed_facet) = self.0.get_mut(facet) {
+                    match narrowed_facet {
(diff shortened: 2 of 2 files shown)


stroxler (approved):

Issue #494, opened 2025-06-16
[feature] handle dataclass `field(kw_only=False)`


yangdanny97:
If the dataclass is decorated with `@dataclass(kw_only=True)` but the field is defined with `field(kw_only=False)`, the field should not be keyword-only. Right now the logic only checks if the flag is set to true.

Code pointers:
https://github.com/facebook/pyrefly/blob/5397b47061187b1a99bc170972a1bfc7736c0491/pyrefly/lib/alt/class/dataclass.rs#L139

After you make the change, add a test case to pyrefly/lib/test/dataclasses.rs, and once that is passing, run test.py and commit any generated changes.

pt2302:
I would like to work on this.

ndmitchell:
We've assigned it to you, go for it :)

Issue #493, opened 2025-06-16
[feature] dataclass kw_only field ordering in constructors


yangdanny97:
### Describe the Bug

Dataclasses should re-order kw-only fields to be after non-kw-only fields in the constructor.

If a child class defines a new non-kw-only field, it should go before any inherited kw-only fields.

Possibly a good first issue, will fill out details later. If you'd like to pick this up, LMK and I can dig for some code pointers.

pt2302:
I would like to work on this.

yangdanny97:
Thanks!

I think the relevant ordering is in the dataclass metadata (and the places that populate it): https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/alt/types/class_metadata.rs#L367

The data structure we use to store the fields is ordered.

Right now, when we inherit the fields from a parent class, we just copy them over and add any newly defined fields after that. 

One possible solution is to keep track of keyword-only fields in a separate set from non-keyword-only fields, and append to the correct one depending on the field type.

Let me know if that makes sense, and if you have any more questions.

darricklaidin:
I was playing around with the sandbox and noticed that the code below was being reported as invalid by Pyrefly, even though it runs with no issues: 
- Missing argument `x` in function `C.__init__`
- Expected 0 positional arguments, got 1 in function `C.__init__`

```python
from dataclasses import dataclass, field

@dataclass(kw_only=False)
class C:
    y: int = field(kw_only=True)
    x: int
    
C(2, y=1)
```
Is the fix for this issue aiming to resolve this behaviour, or is it specifically for inheritance cases?

yangdanny97:
Oh, good catch - my hunch is that if we store positional and kw-only fields separately, it should fix that behavior as well.

rchen152:
@yangdanny97 Is this fixed by https://github.com/facebook/pyrefly/commit/e6aba0d648b1a0c1afc70dbcfab060c476e91090? Or is there more to do?

yangdanny97:
I think this is good to close

PR #566, opened 2025-06-24
Handle dataclass field kw_only overrides

pt2302:
## Description

This PR allows for field-level overrides to set `kw_only=False` for a dataclass when the class has `kw_only` set to `True`. 

## Relevant Issue
Closes https://github.com/facebook/pyrefly/issues/494.

## Testing Instructions

```

@dataclass(kw_only=True)
class SomeClass:
    x: int = field(kw_only=False)

SomeClass(x=1)
SomeClass(1)

```
and run `pyrefly check`. Verify that no errors are generated.

Note: this PR does not handle cases where there are mixed positional and keyword-only arguments; that will be handled in a follow-up PR (related issue: https://github.com/facebook/pyrefly/issues/493).


yangdanny97:
Nice, thanks!

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -141,11 +141,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-                params.push(field.as_param(
-                    &name,
-                    field_flags.is_set(&DataclassKeywords::DEFAULT),
-                    kw_only || field_flags.is_set(&DataclassKeywords::KW_ONLY),
-                ));
+                params.push(
+                    field.as_param(
+                        &name,
+                        field_flags.is_set(&DataclassKeywords::DEFAULT),
+                        field_flags
```
rchen152, line 148: You can change the default value passed to `is_set` like so:
```
field_flags.is_set(&(DataclassKeywords::KW_ONLY.0, kw_only)),
```
No need for a new `get_explicit` method.

yangdanny97: I'll address this in the internal version of the PR

Issue #521, opened 2025-06-18
[feature] include a reason for why a field is read-only


yangdanny97:
Ie: frozen dataclass, namedtuple element, Final, etc.

I think we can make a ReadOnlyReason enum and attach it to a class field (or perhaps put it inside Visibility::ReadOnly) so that we can generate better error messages.

Adist319:
I can work on this!

Adist319:
Hey @yangdanny97,

I've been making some progress on adding the `ReadOnlyReason` enum. That said, I’ve got one last integration test failing: `test_final_attribute_assigned_in_init`.

My new logic catches fields marked as `Final` and blocks edits, but I think it's a bit too aggressive because it’s also flagging assignments *inside* the class’s own `__init__`, which should be allowed.

So I thought, before raising a read-only error in `solve.rs`, I want to add a check like:

* Is the `ReadOnlyReason` specifically `Final`? Basically seeing if the reason something is ReadOnly is due to the Final keyword.
* And are we currently in that class’s `__init__`?

I added a stub method in `answers.rs` called `is_in_init_of`, and I’d love advice on the cleanest way to implement it. Specifically, I'm wondering whether there's a clean, idiomatic way within to determine if the current TextRange is inside the __init__ method of a specific Class?

I started poking around with `self.bindings`, but I’m not sure what the cleanest way is to trace from a `TextRange` to its containing function + class.

Here’s the stub I have now:

```rust
/// Check if we are currently inside the __init__ method of the specified class
pub fn is_in_init_of(&self, cls: &Class, range: TextRange) -> bool {
    // ??? Need to walk up scope tree using self.bindings?
    // Should return true if we're inside the `__init__` method of `cls`.
    // or maybe take a different approach altogether
    false
}
```

Any pointers or examples would be awesome. Thanks in advance! I can make a PR too if you wanted to see the diffs so far.

yangdanny97:
So I think the current behavior is that Final attrs only get marked as read-only if they're initialized on the class (not in a method), because of the problem with `__init__` that you described.

Fixing that will require some refactoring (during the bindings step, we already collect which methods assign to which attributes, and if we plumb that through we should be able to avoid read-only errors in `__init__`). But I think that refactoring is out of scope for this task, and I'd rather we add the ReadOnlyReason system first without modifying current behavior, instead of bundling both changes together.

[...]

PR #570, opened 2025-06-25
Implement ReadOnlyReason for specific error messages

Adist319:
This PR completes the architectural refactoring to provide more descriptive errors for assignments to read-only fields, addressing issue #521. The core change replaces the simple readonly: bool flag with a more expressive readonly: Option<ReadOnlyReason> type across the checker.


Adist319:
Just a note: I ran python test.py on a Windows environment, which gave me a conformance output with Windows-style backslashes. Let me know if this is expected.

Other than that, the expected error messages for several tests have been updated from the generic Cannot assign to read-only attribute... to the new, more specific messages, for example:
`Cannot assign to field 'x'; it is part of a frozen dataclass`
`Cannot assign to field 'x'; it is part of a NamedTuple`

And we could address https://github.com/facebook/pyrefly/issues/521#issuecomment-3000551239 in a separate issue/PR.

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -354,6 +369,41 @@ impl ClassField {
...
+    /// Determine the readonly reason from annotation and other factors
+    fn determine_readonly_reason<Ans: LookupAnswer>(
+        cls: &Class,
+        name: &Name,
+        annotation: &Option<Annotation>,
+        solver: &AnswersSolver<'_, Ans>,
+    ) -> Option<ReadOnlyReason> {
+        // 1. Check annotations first (highest precedence)
+        if let Some(ann) = annotation {
+            // if ann.is_final() {
```
yangdanny97: why's this commented out?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -504,8 +554,12 @@ fn bind_instance_attribute(
             Some(make_bound_method(instance, attr).into_inner()),
             instance.class.dupe(),
         ),
-        attr if is_class_var || readonly => {
-            Attribute::read_only(make_bound_method(instance, attr).into_inner())
+        attr if is_class_var => Attribute::read_only(
+            make_bound_method(instance, attr).into_inner(),
+            ReadOnlyReason::Inherited,
```
yangdanny97, line 574: I think the guard is saying, this attribute is a ClassVar, so it can't be set from the instance

see https://typing.python.org/en/latest/spec/class-compat.html#classvar

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -109,7 +110,7 @@ enum ClassFieldInner {
         ty: Type,
         annotation: Option<Annotation>,
         initialization: ClassFieldInitialization,
-        readonly: bool,
+        readonly: Option<ReadOnlyReason>,
```
yangdanny97: Nit: could you add a comment here explaining that None means it's read-write?

Also, it might make sense to rename to readonly_reason

pyrefly/lib/alt/solve.rs
```diff
@@ -1733,7 +1733,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             (Type::TypedDict(typed_dict), Type::Literal(Lit::Str(field_name))) => {
                 let field_name = Name::new(field_name);
                 if let Some(field) = self.typed_dict_field(typed_dict, &field_name) {
-                    if field.read_only {
+                    if field.read_only.is_some() {
```
yangdanny97: Would prefer a is_read_only() helper method to get the boolean value, to avoid exposing the implementation unnecessarily. The read_only field can remain public, but maybe rename to readonly_reason as well.

pyrefly/lib/solver/subset.rs
```diff
@@ -730,13 +730,13 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
 
                 want_fields.iter().all(|(k, want_v)| {
                     got_fields.get(k).is_some_and(|got_v| {
-                        match (got_v.read_only, want_v.read_only) {
```
yangdanny97, line 757: After you address https://github.com/facebook/pyrefly/pull/570/files#r2166837276, these Some/None checks can go back to being true/false checks

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -678,24 +732,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
 
-        let is_namedtuple_member = metadata
-            .named_tuple_metadata()
-            .is_some_and(|named_tuple| named_tuple.elements.contains(name));
-        let is_frozen_dataclass_field = metadata.dataclass_metadata().is_some_and(|dataclass| {
-            dataclass.kws.is_set(&DataclassKeywords::FROZEN) && dataclass.fields.contains(name)
-        });
-
-        // Read-onlyness
-        let readonly = is_namedtuple_member
```
yangdanny97, line 689: the logic here re: final attrs initialized on the class doesn't seem to be preserved in `determine_readonly_reason `

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -354,6 +369,41 @@ impl ClassField {
...
+            if ann.has_qualifier(&Qualifier::ReadOnly) {
+                return Some(ReadOnlyReason::ReadOnlyAnnotation);
+            }
+        }
+
+        // 2. Check class-level properties
+        let metadata = solver.get_metadata_for_class(cls);
+
+        if metadata.named_tuple_metadata().is_some() {
+            return Some(ReadOnlyReason::NamedTupleField);
```
yangdanny97, line 408: NamedTuples can have attributes that are not elements. The spec says:

> A named tuple class can be subclassed, but any fields added by the subclass are not considered part of the named tuple type.

I think the added fields should not be marked as read-only

The previous logic handled this, so it can probably be copied over.

yangdanny97 (changes requested):
Thanks! Back to you with some comments.

DW about the line endings, they'll be automatically regenerated when this is merged. If it's causing a lot of noise I can look into it.

pyrefly/lib/alt/class/class_field.rs
Adist319: I commented this out because we're deferring the Final + __init__ logic to a separate PR, but let me know if you want this removed instead.

pyrefly/lib/alt/class/class_field.rs
Adist319: Was actually debating on renaming to readonly_reason, but didn't get around to it. Will do!

pyrefly/lib/alt/class/class_field.rs
Adist319: Got it, didn't actually know this about NamedTuples.

pyrefly/lib/alt/class/class_field.rs
Adist319: Good catch. Will add this logic back.

Adist319:
Pushed my new changes. Thanks for the detailed feedback @yangdanny97!

Edit: I noticed that a few tests in the suite — like test_mypy_demo, test_module_type, and test_nested_func have been running for over 60 seconds on CI. Just wanted to check if that’s expected behavior or if it might be a hang or a recent regression.

For what it's worth, cargo test --release completes normally on my local machine.

pyrefly/lib/alt/class/class_field.rs
yangdanny97: We can leave the existing behavior of marking Final attrs that are initialized on the class as read-only, but leave Final attrs that are initialized in the constructor alone.

Adist319: Yup, thanks for the catch, I believe the existing behavior should already be back in the latest changes.

yangdanny97:
@Adist319 I think we broke those tests in the main branch yesterday but that was reverted last night, let me know if it's still broken after a rebase.

Adist319:
Will get back to this, the rebase turned out messier than expected and I likely messed something up.

Edit: I’ll probably open a new branch and PR to keep the commit history clean.

rchen152 (approved):

yangdanny97:
@Adist319 no worries, I can rebase and clean it up for you. Thanks for your work on this!

PR #575, opened 2025-06-26
Add option to specify error type with suppression comment

### LINKED ISSUE #538: Target specific errors by adding slugs to suppression comments
### Describe the Bug

From [the docs](https://pyrefly.org/en/docs/error-suppressions/), we can suppress errors using `# pyrefly: ignore`, but we should also be able to suppress specific errors using `# pyrefly: ignore[no-matching-overload]`, similar to other type checkers.

### Sandbox Link

_No response_

_No response_

TheRustyPickle:
Summary:

* Added support for error type with suppression comment such as `# pyrefly: ignore[error-type]`
* Tests for the new suppression type
* Modified mdx files to include information about the new suppression option

Resolves #538

### DIFF
--- a/pyrefly/lib/error/kind.rs
+++ b/pyrefly/lib/error/kind.rs
@@ -189,6 +189,19 @@ pub enum ErrorKind {
     UnsupportedOperand,
 }
 
+impl std::str::FromStr for ErrorKind {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        ERROR_KIND_CACHE
+            .iter()
+            .zip(enum_iterator::all::<ErrorKind>())
+            .find(|(cached, _)| cached == &s)
+            .map(|(_, kind)| kind)
+            .ok_or(())
+    }
+}
+
 /// Computing the error kinds is disturbingly expensive, so cache the results.
 /// Also means we can grab error code names without allocation, which is nice.
 static ERROR_KIND_CACHE: LazyLock<Vec<String>> = LazyLock::new(ErrorKind::cache);
--- a/pyrefly/lib/error/kind.rs
+++ b/pyrefly/lib/error/kind.rs
@@ -14,6 +14,7 @@ use enum_iterator::Sequence;
 use parse_display::Display;
 use serde::Deserialize;
 use serde::Serialize;
+use starlark_map::small_map::SmallMap;
 use yansi::Paint;
 use yansi::Painted;
 
@@ -193,28 +194,32 @@ impl std::str::FromStr for ErrorKind {
     type Err = ();
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        ERROR_KIND_CACHE
-            .iter()
-            .zip(enum_iterator::all::<ErrorKind>())
-            .find(|(cached, _)| cached == &s)
-            .map(|(_, kind)| kind)
-            .ok_or(())
+        ERROR_KIND_CACHE.get(s).copied().ok_or(())
     }
 }
 
 /// Computing the error kinds is disturbingly expensive, so cache the results.
 /// Also means we can grab error code names without allocation, which is nice.
-static ERROR_KIND_CACHE: LazyLock<Vec<String>> = LazyLock::new(ErrorKind::cache);
+static ERROR_KIND_CACHE: LazyLock<SmallMap<String, ErrorKind>> = LazyLock::new(ErrorKind::cache);
 
 impl ErrorKind {
-    fn cache() -> Vec<String> {
-        enum_iterator::all::<ErrorKind>()
-            .map(|x| x.to_string().to_case(Case::Kebab))
-            .collect()
+    fn cache() -> SmallMap<String, ErrorKind> {
+        let mut map = SmallMap::new();
+
+        for kind in enum_iterator::all::<ErrorKind>() {
+            let key = kind.to_string().to_case(Case::Kebab);
+            map.insert(key, kind);
+        }
+
+        map
     }
 
     pub fn to_name(self) -> &'static str {
-        ERROR_KIND_CACHE[self as usize].as_str()
+        ERROR_KIND_CACHE
+            .get_index(self as usize)
+            .unwrap()
+            .0
+            .as_str()
     }
 
     pub fn severity(self) -> Severity {
--- a/pyrefly/lib/module/ignore.rs
+++ b/pyrefly/lib/module/ignore.rs
@@ -154,6 +154,15 @@ mod tests {
         assert!(Ignore::get_suppression_kind("# ignore: pyrefly").is_none());
         assert!(Ignore::get_suppression_kind(" pyrefly: ignore").is_none());
         assert!(Ignore::get_suppression_kind("normal line").is_none());
+        assert!(
+            Ignore::get_suppression_kind("# pyrefly: ignore") == Some(SuppressionKind::Pyrefly)
+        );
+        assert!(
+            Ignore::get_suppression_kind("# pyrefly: ignore[bad-return]")
+                == Some(SuppressionKind::TypedPyrefly(ErrorKind::BadReturn))
+        );
+        assert!(Ignore::get_suppression_kind("# pyrefly: ignore[]").is_none());
+        assert!(Ignore::get_suppression_kind("# pyrefly: ignore[bad-]").is_none());
     }
 
     #[test]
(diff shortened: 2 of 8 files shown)


yangdanny97:
Thanks! I think the general direction is good, but I'd like to check on two things:

1. should we just make SuppressionKind::Pyrefly take an Option<ErrorKind> instead of making a new case?
2. should we emit an error if the thing in the brackets does not match a known error slug? if that happens, should it behave like a suppress-all-errors-on-the-next-line, or should it behave like the suppression comment wasn't there? I think the former but I'm not sure what everyone else thinks.

LMK if you have opinions on these @TheRustyPickle

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
 3 + "3" # E:
 "#,
 );
+
```
connernilsen, line 65: great tests!

TheRustyPickle:
Here are my thoughts on them:

1. I lean toward having a new SuppressionKind case, it makes the intent clearer and what to expect from it 
2. If the error slug doesn’t match, my opinion is that we should ignore the suppression. If someone (or some tool) has made the effort to target a specific error, then silently suppressing all might hide real issues.

pyrefly/lib/error/kind.rs
```diff
@@ -189,6 +189,19 @@ pub enum ErrorKind {
     UnsupportedOperand,
 }
 
+impl std::str::FromStr for ErrorKind {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        ERROR_KIND_CACHE
+            .iter()
+            .zip(enum_iterator::all::<ErrorKind>())
```
connernilsen: instead of zipping here, could we make the `ERROR_KIND_CACHE` a `SmallMap<String, ErrorKind>`?

The reason we have a cache for this is because `enum_iterator` is apparently pretty slow, so we only want to do this once.

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
...
+  return 1  # pyrefly: ignore[bad-return]
+"#,
+);
+
+testcase!(
+    test_pyrefly_suppression_typed_wrong_type,
+    r#"
+def foo() -> str:
+  # pyrefly: ignore[bad-assignment]
+  return 1 # E:
```
connernilsen: Is it possible to put the error message after the `E:` so that we can verify the right error message is coming up? Also it serves as good documentation about what an error from this would look like.

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
...
+"#,
+);
+
+testcase!(
+    test_pyrefly_suppression_typed_inline_wrong_type,
+    r#"
+def foo() -> str:
+  return 1  # pyrefly: ignore[bad-assignment]  # E:
```
connernilsen: Same with error message

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
...
+"#,
+);
+
+testcase!(
+    test_pyrefly_suppression_typed_bad_type,
+    r#"
+def foo() -> str:
+  # pyrefly: ignore[bad-]
+  return 1 # E:
```
connernilsen: Same with error message

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
...
+"#,
+);
+
+testcase!(
+    test_pyrefly_suppression_typed_empty,
+    r#"
+def foo() -> str:
+  # pyrefly: ignore[]
+  return 1 # E:
```
connernilsen: Same with error message

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
...
+  return 1
+"#,
+);
+
+testcase!(
+    test_pyrefly_suppression_typed_multiple_codes,
+    r#"
+def foo() -> str:
+  # pyrefly: ignore[bad-return, bad-assignment]
+  return 1  # E:
```
connernilsen: Same with error message

pyrefly/lib/test/suppression.rs
```diff
@@ -62,3 +62,73 @@ testcase!(
...
+  #   pyrefly:    ignore   [  bad-return  ]
+  return 1
+"#,
+);
+
+testcase!(
+    test_pyrefly_suppression_typed_multiple_codes,
+    r#"
+def foo() -> str:
+  # pyrefly: ignore[bad-return, bad-assignment]
```
connernilsen, line 131: Good idea for this test!

If you're open to it, could you write another test suppressing multiple error kinds?

Something like this would be great!
```python
def foo() -> str:
  # pyrefly: ignore[bad-return]
  # pyrefly: ignore[bad-argument-type]
  return len(1)
```

pyrefly/lib/module/ignore.rs
```diff
@@ -71,36 +76,61 @@ impl Ignore {
...
-                words.next() == Some("ignore")
-            } else {
-                false
+                if let Some(word) = words.next() {
+                    if word == "ignore" {
+                        return Some(SuppressionKind::Pyrefly);
+                    }
+
+                    if word.starts_with("ignore[") && word.ends_with(']') {
+                        if let Ok(kind) = ErrorKind::from_str(&word[7..word.len() - 1]) {
```
connernilsen: minor nit, feel free to ignore if you disagree, but can we make the index this instead?

```
&word["ignore[".len()..word.len() - 1]
```

connernilsen (changes requested):
Looks great so far, love the implementation.

Just a few small things I have comments on. Re-request me when you're ready for another review and I'll pull it in!

nth10sd:
Drive-by question for @TheRustyPickle: I'm assuming this will work well with `--remove-unused-ignores`?

e.g. if `# pyrefly: ignore[bad-return,bad-assignment]` is listed in the source code, `bad-return` is no longer applicable so `--remove-unused-ignores` removes `bad-return`, leaving `# pyrefly: ignore[bad-assignment]` in the source code.

Or if `# pyrefly: ignore[bad-return]` is listed, and the entire message gets removed.

pyrefly/lib/test/suppression.rs
TheRustyPickle: I have a question on this. The current implementation does not track multi-line ignores so the test case you mentioned would fail. Did you mean I should add support for this or just add the test case with the expected error message?

pyrefly/lib/module/ignore.rs
ndmitchell: Pretty much any string indexing is a poor plan. Better to do `let Some(word) = word.strip_prefix("ignore[") && let Some(word) = word.strip_suffix(']')` and then you don't have to muck around with string offsets.

TheRustyPickle:
@nth10sd Good question. I just checked. Currently, typed suppressions are being ignored by the `--remove-unused-ignores` flag. Also, the current implementation only supports a single slug inside `ignore[]`

If @connernilsen is open to it, I can look into adding support for both in this PR. Otherwise, I’ll wait for guidance on how they should be addressed.

yangdanny97:
Why don't we support the remove-unused-ignores part in a followup PR, since this one is getting a bit large.

I'd OK with only allowing a single error inside the brackets, assuming that you can still suppress multiple errors by adding more `pyrefly-ignore[...]` comments later in the line. If that's not the case then we could discuss how to support multiple specific suppressions (also probably best saved for a followup PR)

pyrefly/lib/test/suppression.rs
connernilsen: Oh, didn't realize it would fail. Ignore this one then, just thought it would be an easy thing to test/demo.

ndmitchell:
We should not allow `type: ignore[a, b]` as that isn't something any other checker does, and we should not be over-complicating this unless we have to.

nth10sd:
> If @connernilsen is open to it, I can look into adding support for both in this PR.

@TheRustyPickle Thanks for the fix! Would you like to open the follow-up?

ndmitchell:
I'm doing a bunch of refactoring in this module right now, to make it do some other things I need to do. Any pending list?

ndmitchell:
I think I did everything that was outlined a bug but please open follow-ups for anything still missing.

nth10sd:
> @TheRustyPickle Thanks for the fix! Would you like to open the follow-up?

I filed [a follow-up issue](https://github.com/facebook/pyrefly/issues/674) for this.

Issue #492, opened 2025-06-16
[feature] error when dataclass field without default follows dataclass field with default


yangdanny97:
### Describe the Bug

Similar to https://github.com/facebook/pyrefly/issues/410 but for dataclasses.

The solution will be very similar to what was implemented in https://github.com/facebook/pyrefly/commit/8c4b8e587ff7cf65ba2ac4c811e91761cb6ce59b

Code pointer: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/alt/class/dataclass.rs

Please add integration tests to https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/dataclasses.rs, run test.py, and commit any generated changes.

darricklaidin:
I'm interested in picking this up.

yangdanny97:
Thanks, I've assigned it to you. Let me know if you have any questions!

darricklaidin:
I'm runnign

darricklaidin:
I'm trying to run `python3 test.py` on my local environment in the `main` branch, but it seems to be hanging on a couple of them. Looking at this [workflow job](https://github.com/facebook/pyrefly/actions/runs/15887189919/job/44801941555?pr=570), it seems like there's something funky going on with the test pipeline.

I'm running:
- OS: Kubuntu 25.04
- Node: 24.2.0
- Python: 3.13.5
- Rust: rustc 1.87.0-nightly (920d95eaf 2025-03-28)

yangdanny97:
Tests were broken on the main branch yesterday due to a bad commit, but
it's been reverted now. Does it reproduce after a rebase?

darricklaidin:
These 2 test cases still fail for me. 😕

```
failures:

---- module::finder::tests::test_find_module_prefixes_nested_namespaces stdout ----

thread 'module::finder::tests::test_find_module_prefixes_nested_namespaces' panicked at pyrefly/lib/module/finder.rs:978:9:
assertion `left == right` failed
  left: [ModuleName("foo2"), ModuleName("foo")]
 right: [ModuleName("foo"), ModuleName("foo2")]

---- module::finder::tests::test_find_module_prefixes_namespaces stdout ----

thread 'module::finder::tests::test_find_module_prefixes_namespaces' panicked at pyrefly/lib/module/finder.rs:995:9:
assertion `left == right` failed
  left: [ModuleName("foo2"), ModuleName("foo")]
 right: [ModuleName("foo"), ModuleName("foo2")]
[...]

Issue #410, opened 2025-05-31
[feature] error when NamedTuple field without a default appears after any NamedTuple field with default


yangdanny97:
### Describe the Bug

From the typing spec: 

> The runtime implementation of NamedTuple enforces that fields with default values must come after fields without default values. Type checkers should likewise enforce this restriction

Here is the code that needs to change:

https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/alt/class/named_tuple.rs#L30

Once this is done, please add a test to: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/named_tuple.rs

Run test.py and commit any generated changes.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoByAhhAKYAmAKgK4IA2pAUIwMZ1EDOHUAYmGAApiZKrQYBKAFyMosqAA9JmFPgC8UAIwy5cJanxQAxFEoALJNw6mw1OuSikQ4EEA

adsi7698:
Hi Yang,

I am interested in trying out this task.

Thanks,

PR #583, opened 2025-06-27
Add check for dataclass field default ordering

darricklaidin:
This PR introduces a validation step to raise an error when a dataclass field without a default follows a field with a default, addressing issue #492.

This ensures dataclasses are declared in the correct field order, aligning with Python's rules for `@dataclass`.

The core change involves modifying the `get_dataclass_init` function when getting the dataclass synthesized fields to perform an ordering validation check (similar to #410). Due to the additional properties of a dataclass, additional checks for `init`, `kw_only`, and `default` has to be made.

Separately, certain test cases that I've added rely on issue #493 being fixed. For now, I have commented those out - and will uncomment and verify them once the aforementioned issue has been resolved.


darricklaidin:
Let me know if there's a better place to put the validation logic, e.g., between lines [66-67](https://github.com/facebook/pyrefly/pull/583/files#diff-d2a2a2a622f8ae1eeda175db314e514ac3a31215a0bd4b894fcf27ed3972bca0L60).

pyrefly/lib/test/dataclasses.rs
```diff
@@ -708,3 +710,333 @@ SomeClass(x=1) # OK
...
+from dataclasses import dataclass
+@dataclass
+class C:
+    x: int = 1
+    class_var: ClassVar[str] = "ignored"  # OK - ClassVar fields bypass ordering validation
+    y: int  # E: DataClass field 'y' without a default may not follow DataClass field with a default
+    "#,
+);
+
+/*
```
rchen152: Why are these tests commented out? If they're broken, our convention is to add `# E:` comments matching the current behavior, plus a `bug = ...` line describing why the current behavior is buggy. Here's an example: https://github.com/facebook/pyrefly/blob/747a4bce327b0b022da552f6e830327904e17e06/pyrefly/lib/test/simple.rs#L1494-L1503

pyrefly/lib/test/dataclasses.rs
rchen152, line 1: This is a really thorough set of test cases, thank you!

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -137,17 +144,39 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                if !is_kw_only {
+                    if !has_default && has_seen_default {
+                        if let Some(range) = cls.field_decl_range(&name) {
+                            self.error(
+                                errors,
+                                range,
+                                error::kind::ErrorKind::BadClassDefinition,
+                                None,
+                                format!(
+                                    "DataClass field '{}' without a default may not follow DataClass field with a default", name
```
rchen152: nit: dataclass is usually written all lower-case.
```suggestion
                                    "Dataclass field '{}' without a default may not follow dataclass field with a default", name
```

rchen152 (commented):
Thanks, this looks great overall! Just a couple minor comments.

pyrefly/lib/test/dataclasses.rs
darricklaidin: Thanks for letting me know and providing the example. Those were commented out because I was planning to keep the PR open until issue #493 was fixed and then re-visit these test cases, but if that's not necessary, it's even more convenient for me to follow the convention so I'll do just that. 🙂

Issue #530, opened 2025-06-18
`type[tuple[Any, ...]]` should be considered callable


rchen152:
### Describe the Bug

```
from typing import Any
def f(x: type[tuple[Any, ...]]):
    return x() # Expected a callable, got type[tuple[Any, ...]]
```
I'd expect pyrefly to report no error in this code.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcAUCQCYCmwUwAFAB4BcsilA2jAK4IA2HROABooAOnEBdCQEomJKAqghK3ECigM60oA

yangdanny97:
This is probably caused by our special representation of tuples

The fix is to change https://github.com/facebook/pyrefly/blob/87ec2dc28e7c08beb81d3b83574bbb9871450849/pyrefly/lib/alt/call.rs#L169 to handle `Type::Type(Type::Tuple)` and generate a `Target::Class`.

Test can be added to 
https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/tuple.rs

krathul:
Hi,
I am interested in working on a fix for this.

yangdanny97:
Assigned, let me know if you have any questions. Thanks!

krathul:
I think I need to generate a class::ClassType from a tuple::Tuple. How can I do this ?

yangdanny97:
@krathul here are some examples: https://github.com/search?q=repo%3Afacebook%2Fpyrefly%20stdlib.tuple(&type=code

krathul:
Thanks :)

PR #588, opened 2025-06-27
Handle `type[tuple]` correctly

krathul:
This fix addresses #530.


krathul:
Some of the test cases in `test/simple.rs` do not run to completion locally, it just says `has been running for over 60 seconds`. What could be the reason for this?

yangdanny97:
> Some of the test cases in test/simple.rs do not run to completion locally, it just says has been running for over 60 seconds. What could be the reason for this?

@krathul are you rebased on the latest revision? There were some tests that looped infinitely a day or two ago, caused by a bad commit that has been reverted.

pyrefly/lib/alt/call.rs
```diff
@@ -182,6 +182,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             Type::Type(box Type::ClassType(cls)) | Type::Type(box Type::SelfType(cls)) => {
                 Some(CallTarget::new(Target::Class(cls)))
             }
+            Type::Type(box Type::Tuple(tup)) => {
+                let cls = self.stdlib.tuple(Type::Tuple(tup));
```
yangdanny97: I don't think this is right, 

calling `stdlib.tuple()` with `Type::Tuple` is making a tuple of tuples - the same thing as writing `tuple[tuple[...]]`

This example

https://github.com/facebook/pyrefly/blob/d186028ed276f33dfd3c713d67bf8c52da3d39b7/pyrefly/lib/alt/attr.rs#L1578

Shows how to match on `tup` to create the appropriate types to pass to `stdlib.tuple()`

yangdanny97:
Could you include the original example from the issue as a test case? Thanks!

krathul:
> @krathul are you rebased on the latest revision? There were some tests that looped infinitely a day or two ago, caused by a bad commit that has been reverted
I remember I had done it.
Yup, it works now, thanks.

pyrefly/lib/alt/call.rs
krathul: Ah, now I get it. Thanks :)

Issue #599, opened 2025-06-28
Assert False is not cutting branches


stolati:
### Describe the Bug

When places in the code should never arrive, I do put `assert False` to cut the branch, but it's not taken into account after switching to pyrefly.

### Sandbox Link

https://pyrefly.org/sandbox/?code=FAEwpgZgBBCWA2YD6B7ADmAdkgti8AFLAFxQDOALgE4CUUAtAHznXHBQdSzSxQC8fKAHIqQtpwlQqYCgFcqmLu07cu-QUIDuY5ZKkz5i2LqgBDMmTBUKUAGKn4lgDRQARADkUFABaxMAcxgUWUwQVygAYigAERQwMkwhG00UKgBrYEzwaDhEVAxsPEISFloGZkoqcRUedWFRar1pOQUlCVVeAWFtRslmwzbOKlNYSygANQdZMABRKipUglcKLxRXOiiU9LIoNAWMKngAT0ygA

yangdanny97:
Thanks for reporting! This one should be fairly easy so I'll describe the fix if anyone wants to pick it up.

The fix is probably to check if the [assert condition](https://github.com/facebook/pyrefly/blob/c247d452349c10958a44a3842540b12f7634e220/pyrefly/lib/binding/stmt.rs#L765) is [trivially false](https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_python/src/sys_info.rs#L275), and if so, [mark the control flow](https://github.com/facebook/pyrefly/blob/c247d452349c10958a44a3842540b12f7634e220/pyrefly/lib/binding/stmt.rs#L719) similar to how we do for `raise`.

Then, we would probably want to add an integration test case to pyrefly/lib/tests/flow.rs

PR #603, opened 2025-06-29
binding: treat `assert False` as terminating control flow

Abel981:
This change makes Pyrefly recognize `assert False` as a control flow terminator, similar to `raise`. It adds a call to `mark_flow_termination` when the assert condition is statically known to be `False`.

closes https://github.com/facebook/pyrefly/issues/599

### DIFF
--- a/pyrefly/lib/test/flow.rs
+++ b/pyrefly/lib/test/flow.rs
@@ -1134,6 +1134,36 @@ assert_type(vari, Literal["test"])
 );
 
 testcase!(
+    test_assert_false_terminates_flow,
+    r#"
+from typing import assert_type, Literal
+
+def test_unreachable_code_after_assert_false() -> None:
+    # This test verifies that code immediately following 'assert False'
+    # is considered unreachable by Pyrefly's type checker.
+
+    x = 10
+    assert False, "This assertion should cause control flow to terminate in this branch."
+    # This line should be unreachable.
+    raise Exception("Something unexpected happened here!")
+    "#,
+);
+testcase!(
+    test_assert_true_does_not_terminate_flow,
+    r#"
+from typing import assert_type, Literal, reveal_type
+
+def test_assert_true_normal_flow():
+    x = 10
+
+    assert True, "This assertion always passes"
+    # This line should be reachable and its operations should be evaluated
+    y = x + 5
+    assert_type(y, int)
+    "#,
+);
+
+testcase!(
     bug = "Merge flow is lax about possibly-undefined locals, so we don't catch that `z` may be uninitialized.",
     test_named_inside_boolean_op,
     r#"
--- a/pyrefly/lib/binding/stmt.rs
+++ b/pyrefly/lib/binding/stmt.rs
@@ -765,6 +765,9 @@ impl<'a> BindingsBuilder<'a> {
             Stmt::Assert(mut x) => {
                 self.ensure_expr(&mut x.test, &mut Usage::Narrowing);
                 self.bind_narrow_ops(&NarrowOps::from_expr(self, Some(&x.test)), x.range);
+                if let Some(false) = self.sys_info.evaluate_bool(&x.test) {
+                    self.scopes.mark_flow_termination();
+                }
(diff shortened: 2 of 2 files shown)


pyrefly/lib/test/flow.rs
```diff
@@ -1133,6 +1133,36 @@ assert_type(vari, Literal["test"])
...
+from typing import assert_type, Literal
+
+def test_unreachable_code_after_assert_false() -> None:
+    # This test verifies that code immediately following 'assert False'
+    # is considered unreachable by Pyrefly's type checker.
+
+    x = 10
+    assert False, "This assertion should cause control flow to terminate in this branch."
+    # This line should be unreachable.
+    raise Exception("Something unexpected happened here!")
```
yangdanny97, line 1148: I don't understand how this test demonstrates the difference the code change makes.

Based on the sandbox, Pyrefly currently behaves the same way for this test https://pyrefly.org/sandbox/?code=MQAg6gpgNgxg9gWwiALnVALZAFAngJwgDMpcQBlAQwDsATAIzgA8AaTASwGcQuRKQADpQDmEAFCgA7lkIhccAK4gYNEJPzsUENpwyVZNWiCgR91PowUpUuAe2rCe5vCgxxqAOgkSQACQgGBsqKAiYgRKYoCoTccEQc3JyGjEwAXD4AtCAA4hDWhJRQGSjsSDYCyDBYMADW9o4REAyUtZkgAKqcyIQAbqZQAPoothAAFACUqOj2nBUw1vYR+IRGwxWcbRBMFRpI1NaSmhiYyACSACIAouGR0RDcrpQHyJwKAgJw+NajEB7CHnwrHB4AhQnltCBhHBitDaMR7Jp2O42G4+vg2HkYOM2mANFo+HQQPgFM5cK53ME4SBaOxCPNSE45Ip8CB6Pg4JIuvgxGIiOyEOV6jxQZ9rAAqHlw+JaTgoAYkgpVSj0EwDeBwgaUIhafCazhcuVEQpdCYgDIAPhAADl3BB0iAHSBQAAVDC8GXWNHsIjse6YJ6U5ClJA0p7QMhEOBQKAcoUAcko+oC1gAYsaIHGxI6nTxuPBqJx2HCViAFaYlSrkPQyHhCCRcHGHiNlNUagEvNmmCAALwgACMAAYs47EwaQGmoF02AAiV28UfJpHmXSKKBGFQKLrBfbsqDhGOSKaoAIIexhxmuXhsmhVDzT4cOl1u7hQewvNwKNes5Bllp6Ssdo6+CUFwyCXEwMAQAIJTuKM07kIgeRug4pbUFscxaEYejvBAaFYQEEAAITTuMQA

I would suggest marking the test function to return something like `int`, and seeing if Pyrefly complains about implicit None return.

For example, here:

https://pyrefly.org/sandbox/?code=MQAg6gpgNgxg9gWwiALnVALZAFAngJwgDMpcQBlAQwDsATAIzgA8AaTASwGcQuRKQADpQDmEAFCgA7lkIhccAK4gYNEJPzsUENpwyVZNWiCgR91PowUpUuAe2rCe5vCgxxqAOgkSQACQgGBsqKAiYgRKYoCoTccEQc3JyGjEwAXD4AtCAA4hDWhJRQGSjsSDYCyDBYMADW9o4REAyUtZkgAKqcyIQAbqZQAPoothAAFACUqOj2nBUw1vYR+IRGwxWcbRBMFRpI1NaSmhiYyACSACIAouGR0RDcrpQHyJwKAgJw+NajEB7CHnwrHB4AhQnltCBhHBitDaMR7Jp2O42G4+vg2HkYOM2mANFo+HQQPgFM5cK53ME4SBaOxCPNSE45Ip8CB6Pg4JIuvgxGIiOyEOV6jxQZ9rAAqHlw+JaTgoACMExAGQAfE4UOkQJqiZQuMhLkwYBABCV3BNJcRUPcUAAmRUqtUarWUThc6wAMUKXSAA

With your code change, the error in the second function will not be there.

yangdanny97: I've patched the test and added handling in another place that was missed

PR #608, opened 2025-06-30
shared helper for erasing tuples

yangdanny97:
Summary: We duplicated this logic in a few places. This diff aligns the implementations. Small conformance error changes (0 delta).

Differential Revision: D77545278


rchen152 (approved):

PR #610, opened 2025-06-30
Emit error on uncalled function for if statment

K1T3K1:


pyrefly/lib/alt/expr.rs
```diff
@@ -187,7 +189,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     fn ifs_infer(&self, comps: &[Comprehension], errors: &ErrorCollector) {
         for comp in comps {
             for if_clause in comp.ifs.iter() {
-                self.expr_infer(if_clause, errors);
+                let if_wrapper = Expr::If(ExprIf {
+                    test: Box::new(if_clause.clone()),
+                    body: Box::new(Expr::NoneLiteral(Default::default())),
+                    orelse: Box::new(Expr::NoneLiteral(Default::default())),
+                    range: if_clause.range(),
```
grievejia: I think we could avoid this wrapper here by restructuring the code a little bit. 

I think we could have a separate helper method which takes a type and performs the truthy check on it. Then we could simply pass the return of `self.expr_infer(if_clause, errors)` to that method here, as opposed to passing the wrapper to `expr_infer()`. 

The same helper method could then be reused below in `expr_infer_type_no_trace()` and `solve_expectations()`.

grievejia (changes requested):
Thank you so much for the contribution, and apologies for the delayed response!

There's one thing I think we could probably clean up here, after which I'll import & merge it.

K1T3K1:
@grievejia 
Hey, I noticed that due to recent changes working with the project became a little cumbersome.
- since moving more packages to `crates` LSP is working extremely slow. 
I work with lazyvim with autosave. Everytime I leave input mode to normal my file is saved and `cargo check` is run. This freezes vim for 2-15sec. Which makes working with the project now almost impossible. Ik that my workflow is kinda heavy, but it used to be perfect with pyrefly project
- were some breaking changes introduced to `website`? I used to test with it. Now I get **_'Sandbox isn't currently available Internally, please use the [public sandbox](https://pyrefly.org/sandbox)'_**

yangdanny97:
cc @ndmitchell for the crates, @dluo for the sandbox

K1T3K1:
@grievejia hey, I applied Your suggestions and resolved the conflicts

PR #611, opened 2025-07-01
sort keyword-only and regular dataclass fields

yangdanny97:
Summary:
The diff below this one adds a few bug test cases + has a small regression due to an interaction w/ dataclass field ordering.

The TLDR is that dataclass fields are always grouped s.t. positional ones are before kw-only ones, and the kw_only=True is not inherited from parent dataclasses. So if you have a child class w/o kw_only=True, your fields will go before the parent class's fields in the init function. Overriding a field w/ a different kw_only-ness will reorder it.

This diff implements those semantics, and additionally:
- cleans up some of the synthesis helpers to get rid of boolean flags
- make InitVars obey kw_only=True, KW_ONLY sentinel, etc.

Reviewed By: rchen152

Differential Revision: D77561604


rchen152 (approved):

PR #619, opened 2025-07-02
Limit `unexpected-keyword` error range to just the name

### LINKED ISSUE #601: Unexpected keyword argument has too large a range
### Describe the Bug

Given:

```python
def test(): pass

test(foo=1)
```

We raise the error:

```
ERROR 14:6-11: Unexpected keyword argument `foo` in function `test`
```

But the error range encompasses `foo=1`, not just `foo`. If you have a large multiline expression after `foo` you end up with a lot of red squiggly when just highlighting `foo` would be sufficient.

### Sandbox Link

https://pyrefly.org/sandbox/?code=CYUwZgBALiDOUAoCUAuCAHAhrWAoG8CYA9sQLwCMSQA

_No response_

InSyncWithFoo:
Resolves #601.

Before this change:

```python
def test(): pass

test(foo=1)
#    ^^^^^ unexpected-keyword
```

After this change:

```python
def test(): pass

test(foo=1)
#    ^^^ unexpected-keyword
```

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -791,7 +791,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         hint = Some(ty.clone());
                         has_matching_param = true;
                     } else if kwargs.is_none() {
-                        unexpected_keyword_error(&id.id, kw.range);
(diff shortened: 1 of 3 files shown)


ndmitchell:
Thanks, this looks great!

PR #631, opened 2025-07-04
Add support for checking code snippets

### LINKED ISSUE #503: Feature request: allow checking snippets, equivalent to `mypy -c`
I was testing out `pyrefly` on some things and it would be very nice to try some small snippet without having to add it to a file! Let me know if I just missed some flag to `pyrefly check`.

darricklaidin:
This PR aims to resolve #503. It adds support for checking Python code snippets directly from the command line using the `snippet` command, equivalent to `mypy -c`. This enables quick type checking of code without creating temporary files.

The implementation creates a virtual module system where snippets are treated as <string> files stored in memory. This allows reuse of the entire existing type checking pipeline while maintaining clean separation between file-based and snippet-based workflows. The virtual module approach also ensures snippets integrate seamlessly with import resolution and configuration systems.

# Core Implementation

- CLI Interface: Added `snippet <CODE>` option to accept Python code strings
- Module System: Extended `ModulePathDetails` with `CommandSnippet` variant for virtual modules
- Memory Storage: Implemented in-memory code storage using `<string>` as virtual file path
- Type Checking: Reused existing `run_inner()` infrastructure for consistent behavior

# Key Files Modified

- pyrefly/bin/main.rs: Added command-line argument parsing, a new `SnippetCheckArgs` (to exclude `behavior`), and `run_snippet_check()`
- pyrefly/lib/module/module_path.rs: Added `CommandSnippet` variant and display logic
- pyrefly/lib/commands/check.rs: Added `run_once_with_snippet()` method
- pyrefly/lib/state/load.rs: Added memory lookup for snippet code loading

# Usage Examples

```bash
pyrefly snippet "x: int = 'hello'"
pyrefly snippet  "import requests; r = requests.get('url')"
pyrefly snippet  "def foo(): pass" --config strict.toml
pyrefly snippet  "bad_code" --output-format json
```

Scrut tests have been added to `test/snippet.md`.

### DIFF
--- a/pyrefly/bin/main.rs
+++ b/pyrefly/bin/main.rs
@@ -57,7 +57,7 @@ struct Args {
 #[deny(clippy::missing_docs_in_private_items)]
 #[derive(Debug, Clone, Parser)]
 struct FullCheckArgs {
-    /// Type check a string of Python code directly (equivalent to mypy -c)
+    /// Type check a string of Python code directly
     #[arg(long, value_name = "CODE", conflicts_with = "files")]
     command: Option<String>,
     /// Files to check (glob supported).
--- a/pyrefly/bin/main.rs
+++ b/pyrefly/bin/main.rs
@@ -20,6 +20,7 @@ use library::run::CommandExitStatus;
 use library::run::CommonGlobalArgs;
 use library::run::InitArgs;
 use library::run::LspArgs;
+use library::run::SnippetCheckArgs;
 use library::run::dump_config;
 use pyrefly::library::library::library::library;
 use pyrefly_util::args::get_args_expanded;
@@ -57,9 +58,6 @@ struct Args {
 #[deny(clippy::missing_docs_in_private_items)]
 #[derive(Debug, Clone, Parser)]
 struct FullCheckArgs {
-    /// Type check a string of Python code directly
-    #[arg(long, value_name = "CODE", conflicts_with = "files")]
-    command: Option<String>,
     /// Files to check (glob supported).
     /// If no file is specified, switch to project-checking mode where the files to
     /// check are determined from the closest configuration file.
@@ -88,12 +86,33 @@ struct FullCheckArgs {
     args: CheckArgs,
 }
 
+#[deny(clippy::missing_docs_in_private_items)]
+#[derive(Debug, Clone, Parser)]
+struct SnippetArgs {
+    /// Python code to type check
+    code: String,
+
+    /// Explicitly set the Pyrefly configuration to use when type checking.
+    /// When not set, Pyrefly will perform an upward-filesystem-walk approach to find the nearest
+    /// pyrefly.toml or pyproject.toml with `tool.pyrefly` section'. If no config is found, Pyrefly exits with error.
+    /// If both a pyrefly.toml and valid pyproject.toml are found, pyrefly.toml takes precedence.
+    #[arg(long, short, value_name = "FILE")]
+    config: Option<PathBuf>,
+
+    /// Type checking arguments and configuration
+    #[command(flatten)]
+    args: SnippetCheckArgs,
+}
+
 /// Subcommands to run Pyrefly with.
 #[deny(clippy::missing_docs_in_private_items)]
 #[derive(Debug, Clone, Subcommand)]
 enum Command {
     /// Full type checking on a file or a project
     Check(FullCheckArgs),
+
+    /// Full type checking on a Python code snippet
+    Snippet(SnippetArgs),
 
     /// Dump info about pyrefly's configuration. Use by replacing `check` with `dump-config` in your pyrefly invocation.
     DumpConfig(FullCheckArgs),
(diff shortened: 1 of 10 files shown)


darricklaidin:
I have stuck with `--command` as the flag name for now. Happy to change this if/when we finalize on something.

yangdanny97:
urgh, the import is failing. I'll debug this later. might need a rebase from you @darricklaidin

yangdanny97:
Thanks for the detailed test suite! 

cc @connernilsen @ndmitchell for a review on some of the module/path related plumbing for snippets

pyrefly/lib/commands/check.rs
```diff
@@ -487,6 +487,63 @@ impl Args {
         )
     }
 
+    pub fn run_once_with_snippet(
+        self,
+        code: String,
+        _files_to_check: FilteredGlobs,
```
rchen152: Since this parameter is unused, it seems simpler to just remove it and adjust the function call in main.rs accordingly.

pyrefly/lib/commands/check.rs
```diff
@@ -487,6 +487,63 @@ impl Args {
...
+        config_finder: ConfigFinder,
+        allow_forget: bool,
+    ) -> anyhow::Result<(CommandExitStatus, usize)> {
+        use std::path::PathBuf;
```
rchen152: nit: we usually put imports at the top of the file.

pyrefly/lib/commands/check.rs
```diff
@@ -487,6 +487,63 @@ impl Args {
...
+        use std::sync::Arc;
+
+        use pyrefly_python::module_name::ModuleName;
+
+        use crate::module::module_path::ModulePath;
+
+        let mut timings = Timings::new();
```
rchen152: I don't think we need to time the setup here; we do so in `run_once` because listing files can take a while, but here there's nothing time-consuming happening.

pyrefly/lib/commands/check.rs
```diff
@@ -487,6 +487,63 @@ impl Args {
...
+            holder
+                .as_ref()
+                .new_transaction(require_levels.default, None),
+            allow_forget,
+        );
+
+        // Add the snippet source to the transaction's memory
+        transaction
+            .as_mut()
+            .set_memory(vec![(PathBuf::from("<string>"), Some(Arc::new(code)))]);
```
rchen152: It would be nice to minimize the number of places where the `"<string>"` path is hard-coded. I believe that there's a `ModulePath` available in most of the places where you've hard-coded `"<string>"`, so it should be possible to instead use `ModulePath::as_path()`.

rchen152 (changes requested):
Sorry for the delayed review! I made a few minor suggestions, but overall this looks pretty good to me.

pyrefly/bin/main.rs
```diff
@@ -62,6 +62,9 @@ struct Args {
 
 #[derive(Debug, Clone, Parser)]
 struct FullCheckArgs {
+    /// Type check a string of Python code directly
+    #[arg(long, value_name = "CODE", conflicts_with = "files")]
+    command: Option<String>,
```
connernilsen: I'm wondering if it might be worth making a separate [`Command` variant](https://github.com/facebook/pyrefly/blob/48f12936ccf7ce20d6f74e3ee4dc843df81e0182/pyrefly/bin/main.rs#L101) taking in a new struct, since a lot of the `FullCheckArgs` are mutually exclusive with `--command`.

It might be useful to allow a [`config`](https://github.com/facebook/pyrefly/blob/48f12936ccf7ce20d6f74e3ee4dc843df81e0182/pyrefly/bin/main.rs#L91) to be passed. We also probably don't want [`behavior`, but should `output` and `config_override`](https://github.com/facebook/pyrefly/blob/48f12936ccf7ce20d6f74e3ee4dc843df81e0182/pyrefly/lib/commands/check.rs#L93-L99) errors in this.

test/command_snippet.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+## Help text shows command option
+
+```scrut
+$ $PYREFLY check --help | grep -A 1 "command"
+      --command <CODE>
+          Type check a string of Python code directly
+[0]
+```
```
connernilsen: Can we add one more test trying to import something from a local file?

Probably something like `cat "x: int = 5 > test.py && pyrefly <check snippet>  "import test; reveal_type(test.x)"`

pyrefly/bin/main.rs
darricklaidin: Great suggestion and I think it's definitely worth doing! I've implemented a separate `Command::Snippet` variant that addresses those concerns.

pyrefly/lib/commands/check.rs
darricklaidin: I've replaced the instances where I used the hardcoded `"<string>"` path as per your suggestion.
Additionally, I've created a `const` variable for `"<string>"` itself so that it can be more easily changed in the future if needed.

darricklaidin:
I've done a rebase on top of the main branch @yangdanny97. ☺️

rchen152 (approved):
Looks good to me, thank you!

PR #634, opened 2025-07-06
Detect redundant cast usage and emit warning

### LINKED ISSUE #593: [feature] warn on redundant cast
### Describe the Bug

If we have `cast(T, x)` and x is already assignable to type `T`, we should give a redundant cast warning. This requires a new ErrorKind that is always a warning, like deprecated function calls.

Code pointer: https://github.com/facebook/pyrefly/blob/bd52a9d93125b6c0d3d4722ea8fe81ebbdde6402/pyrefly/lib/alt/expr.rs#L1368

Please add an integration test case in `pyrefly/lib/tests`, and document the new ErrorKind

### Sandbox Link

_No response_

_No response_

sanma613:
## Summary

This PR adds detection for redundant usage of `typing.cast` where the value being cast already matches the target type. In such cases, a warning is now emitted to inform users of unnecessary casting.

## Implementation Details

- A check was added before applying the cast to compare the inferred type of the value against the target type.
- If both types match, an `ErrorKind::RedundantCast` warning is emitted instead of applying the cast.
- The cast still proceeds (non-fatal) to maintain soundness.
- Includes tests under `redundant_cast.rs`.

## Example

```python
from typing import cast

def f(x: int) -> int:
    return cast(int, x)  # warning: Redundant cast
```
## Testing

- Added test case: `test_redundant_cast_simple`
- Added test case: `test_redundant_cast_with_annotation`
- Added test case: `test_non_redundant_cast`

Closes #593

### DIFF
--- a/pyrefly/lib/alt/special_calls.rs
+++ b/pyrefly/lib/alt/special_calls.rs
@@ -234,6 +234,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 "`typing.cast` missing required argument `val`".to_owned(),
             );
         }
+        if let Some(val_expr) = val {
+            let val_type = self.expr_infer(val_expr, errors);
+        
+            if self.subtype(&val_type, &ret) {
+                self.error(
+                    errors,
+                    range,
+                    ErrorKind::RedundantCast,
+                    None,
+                    "`typing.cast` is unnecessary because the value already has the target type".to_owned(),
+                );
+            }
+        }        
         ret
     }
 
--- a/pyrefly/lib/test/redundant_cast.rs
+++ b/pyrefly/lib/test/redundant_cast.rs
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+ use crate::testcase;
+
+
+testcase!(
+    test_redundant_cast_simple,
+    r#"
+from typing import cast
+
+x = cast(int, 5)  # E: Redundant cast: expression already has type `int`
+y = cast(str, "hello")  # E: Redundant cast: expression already has type `str`
+"#,
+);
+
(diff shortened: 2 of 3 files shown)


pyrefly/lib/alt/special_calls.rs
```diff
@@ -234,6 +234,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 "`typing.cast` missing required argument `val`".to_owned(),
             );
         }
+        if let Some(val_expr) = val {
+            let val_type = self.expr_infer(val_expr, errors);
+        
+            if self.subtype(&val_type, &ret) {
```
yangdanny97, line 240: I don't think `subtype` is a function?

Try `is_subset_eq`

pyrefly/lib/alt/special_calls.rs
```diff
@@ -234,6 +234,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                self.error(
+                    errors,
+                    range,
+                    ErrorKind::RedundantCast,
+                    None,
+                    "`typing.cast` is unnecessary because the value already has the target type".to_owned(),
```
yangdanny97, line 246: this message is slightly misleading if the thing being cast is `Any`

why not something simpler like "redundant cast to `{}`"? note that when we pass the type we should call `.deterministic_printing()` on it

pyrefly/lib/test/redundant_cast.rs
```diff
@@ -0,0 +1,38 @@
...
+
+ use crate::testcase;
+
+
+testcase!(
+    test_redundant_cast_simple,
+    r#"
+from typing import cast
+
+x = cast(int, 5)  # E: Redundant cast: expression already has type `int`
```
yangdanny97, line 16: these messages don't seem to match the implementaiton in special_calls.rs

yangdanny97:
this PR is almost there, but I'm going to close it because it's very stale. feel free to reopen w/ updates, or if anyone else wants to take this over LMK

PR #635, opened 2025-07-06
Add cli flag for errors, issue #461

### LINKED ISSUE #461: Add CLI flag for `errors`
Add a CLI flag and environment variable overriding [`errors`](https://github.com/facebook/pyrefly/blob/086fe6b2799f18dcfd3ccc3afcfff856a00d6d04/pyrefly/lib/config/base.rs#L29). See #458 for more details.

This one will likely require two new flags: an `--enable-error` and `--disable-error` flag (or whatever you think it should be named)!

Alex-Aron:
resolves #461


yangdanny97:
I think this looks reasonable but I'll let @connernilsen take a look

pyrefly/lib/commands/check.rs
```diff
@@ -202,6 +204,12 @@ struct ConfigOverrideArgs {
     /// Whether Pyrefly will respect ignore statements for other tools, e.g. `# mypy: ignore`.
     #[arg(long, env = clap_env("PERMISSIVE_IGNORES"))]
     permissive_ignores: Option<bool>,
+    /// Enable specific error types. Overrides any disabled errors of the same type.
+    #[arg(long, env = clap_env("ENABLE_ERROR"))]
+    enable_error: Option<Vec<String>>,
+    /// Disable specific error types.
+    #[arg(long, env = clap_env("DISABLE_ERROR"))]
+    disable_error: Option<Vec<String>>,
```
connernilsen: Can we make these `Vec<ErrorKind>`s instead? 

We shouldn't need `Option` here since the enable/disable flags let us know how we want to change the default instead of fully overwrite the default.

Also, parsing them directly into `ErrorKind` gives us validation on the error kinds passed in and handles the parsing for us. If you run into trouble with this at all, just let me know and we can land with the string error kind and parsing as-is.

pyrefly/lib/commands/check.rs
```diff
@@ -555,6 +563,24 @@ impl Args {
             self.config_override.site_package_path.as_deref(),
         )?;
         validate_arg("--search-path", self.config_override.search_path.as_deref())?;
+
+        // Check for conflicting error types in enable_error and disable_error
+        if let (Some(enable_errors), Some(disable_errors)) = (
```
connernilsen: This would require both `enable_errors` and `disable_errors` to be provided by a user for us to change the way we're reporting errors. If we switch to a regular `Vec<_>` instead of `Option<Vec<_>>`, then we can just attempt to convert to some kind of set here and not have to worry about it being present or not.

pyrefly/lib/commands/check.rs
```diff
@@ -202,6 +204,12 @@ struct ConfigOverrideArgs {
     /// Whether Pyrefly will respect ignore statements for other tools, e.g. `# mypy: ignore`.
     #[arg(long, env = clap_env("PERMISSIVE_IGNORES"))]
     permissive_ignores: Option<bool>,
+    /// Enable specific error types. Overrides any disabled errors of the same type.
```
connernilsen: Down below, it looks like you return an error if the same `kind` is mentioned in both enable and disable, so adding a message about that here seems inconsistent.
```suggestion
    /// Enable specific error kinds.
```

pyrefly/lib/commands/check.rs
```diff
@@ -202,6 +204,12 @@ struct ConfigOverrideArgs {
...
+    #[arg(long, env = clap_env("ENABLE_ERROR"))]
+    enable_error: Option<Vec<String>>,
+    /// Disable specific error types.
```
connernilsen:
```suggestion
    /// Disable specific error kinds.
```

pyrefly/lib/commands/check.rs
```diff
@@ -555,6 +563,24 @@ impl Args {
...
+            &self.config_override.disable_error,
+        ) {
+            let enable_set: HashSet<_> = enable_errors.iter().collect();
+            let disable_set: HashSet<_> = disable_errors.iter().collect();
+            let conflicts: Vec<_> = enable_set.intersection(&disable_set).collect();
+
+            if !conflicts.is_empty() {
+                return Err(anyhow::anyhow!(
+                    "Error types cannot be both enabled and disabled: [{}]",
+                    display::commas_iter(|| conflicts.iter().map(|&&s| s))
```
connernilsen: Thanks for checking this!

pyrefly/lib/commands/check.rs
```diff
@@ -603,6 +629,28 @@ impl Args {
...
+        }
+
+        if let Some(disable_errors) = &self.config_override.disable_error {
+            let mut error_config = config.root.errors.clone().unwrap_or_default();
+            for error_name in disable_errors {
+                if let Ok(error_kind) = error_name.parse::<ErrorKind>() {
+                    error_config.with_error_setting(error_kind, false);
+                }
+            }
+            config.root.errors = Some(error_config);
```
connernilsen: For `enable_errors` and `disable_errors`, it's great that you're setting the `config.root.errors`! Can we also do that for any `root.subconfigs` that are there too?

connernilsen (changes requested):
Hey @Alex-Aron, thanks so much for doing this! And also sorry for the rush with the bot comment. It was something we just turned on and are still tweaking the messaging for.

Things look great so far, the direction you're developing this in looks right to me! Just a few changes, mostly small things I'd like to take a look at before merging this in. If you have questions about any of it, let me know and I can help out. 

Also, when you're ready for another review, please request me again on your PR!

pyrefly/lib/commands/check.rs
```diff
@@ -577,6 +586,26 @@ impl Args {
...
+            .iter()
+            .collect::<HashSet<_>>()
+            .intersection(
+                &self
+                    .config_override
+                    .disable_error
+                    .iter()
+                    .collect::<HashSet<_>>(),
+            )
+            .copied()
```
yangdanny97, line 601: can save a copy here if you make separate variables for disabled_errors and enabled_errors and then have conflicts be just the intersection operation.

```
        let disabled_errors = &self
            .config_override
            .disable_error
            .iter()
            .collect::<HashSet<_>>();
        let enabled_errors = self
            .config_override
            .enable_error
            .iter()
            .collect::<HashSet<_>>();
        let conflicts: Vec<_> = enabled_errors.intersection(disabled_errors).collect();
```

pyrefly/lib/commands/check.rs
```diff
@@ -635,6 +664,23 @@ impl Args {
...
+        let apply_error_settings = |mut error_config: ErrorDisplayConfig| {
+            for error_kind in &self.config_override.enable_error {
+                error_config.with_error_setting(*error_kind, true);
+            }
+            for error_kind in &self.config_override.disable_error {
+                error_config.with_error_setting(*error_kind, false);
+            }
+            error_config
+        };
+        config.root.errors = Some(apply_error_settings(
```
yangdanny97, line 676: you can do something like config.root.errors.get_or_insert_default to insert the default value if it's None, and get a mutable reference to the ErrorDisplayConfig.

Since apply_error_settings is mutating, we don't need to clone and return a new one.

```
        let apply_error_settings = |error_config: &mut ErrorDisplayConfig| {
            for error_kind in &self.config_override.enable_error {
                error_config.with_error_setting(*error_kind, true);
            }
            for error_kind in &self.config_override.disable_error {
                error_config.with_error_setting(*error_kind, false);
            }
        };
        let root_errors = config.root.errors.get_or_insert_default();
        apply_error_settings(root_errors);
        for sub_config in config.sub_configs.iter_mut() {
            let sub_config_errors = sub_config.settings.errors.get_or_insert_default();
            apply_error_settings(sub_config_errors);
        }
```

yangdanny97:
These comments are just for future reference, no changes needed on your end as I've already applied them as patches.

This PR will be merged shortly. Thanks for your contribution!

PR #656, opened 2025-07-10
Check positional argument count for type guard functions

### LINKED ISSUE #596: [feature] TypeGuard/TypeIs functions must have at least 1 positional parameter. If it's a method, it must have at least 2 positional parameters.
### Describe the Bug

See the sandbox link for examples of the cases we should cover.

We should add the check to this function: https://github.com/facebook/pyrefly/blob/c247d452349c10958a44a3842540b12f7634e220/pyrefly/lib/alt/function.rs#L166

The return type is calculated here: https://github.com/facebook/pyrefly/blob/c247d452349c10958a44a3842540b12f7634e220/pyrefly/lib/alt/function.rs#L376

Once we get the return type, we can see if we need to have 1 or 2 params based on what kind of function it is, and emit a BadFunctionDefinition error if it doesn't have the required number.

Once this task is done, add an integration test to pyrefly/lib/tests/narrow.rs, run test.py, and commit any generated changes.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQMRQB8UAksFAIayICmUaArvSACZTDsoAxjCRgUmAM6ZsAG2YRmKGM170p9cagkwNg1gpgALML1xRBMtRLKUoB49wA0sQ62BIQ2qDglIRY+hkGEA4FJTt6BCkYMBdWCWZBMV4ELnoDZhAbKgAKemBlEChAHAIEmWBAXAIoM2KLCQqASgA6EgsrKABhAC4iKD6obmY6GDQAfQBGHLLgBqgAWioAFRYAcU4eAG1UGABdLr7yADkwfAB5AGle-uuQZhh2EHEAMUCEkmuAATaJCXsTK76g2GYwATDk6rMFlBlghmGsuNwtko9gcoMczpdrjc7g9nq9mCRyItXAlgm5UASgbAxgBmHLAMBgSFLVbrRHbPYAqC3e6PKAvGRvKkjUYAFnpjOcACMuMzoayEUjdj1sby8YLKUNqaMAKwSsDOAD0cphcLZSs5qtx-PxJCIwrGADYciaFZsOfsoEcTlALlyedaBUKtSKAOw5AjOBlM+Ys2Hw93Iz3ejFRxmSKAAa2YcAA7rhuHMxDI4P6cXygwSgA

_No response_

fangyi-zhou:
Closes #596.

According to the Python spec, type guard functions (i.e. with return
types `TypeGuard[_]` or `TypeIs[_]`) must have at least one positional
argument, or at least two positional arguments in the case of instance
method or class method.
This diff implements the required check when typechecking the function
definition by validating the number of positional arguments; and adds
test cases for both kind of type guard functions.


yangdanny97:
Thanks for the detailed tests!

PR #659, opened 2025-07-10
Validate subtype relation for `TypeIs`

fangyi-zhou:
(This diff depends on https://github.com/facebook/pyrefly/pull/656)

According to the spec for `TypeIs`
https://typing.python.org/en/latest/spec/narrowing.html#typeis,
the return type in `TypeIs` must be a subtype of the input type being
narrowed. This diff adds a check to that effect for the required
subtyping relation, and adds appropriate test cases.

After this change, all the conformance tests for typeis should pass.


yangdanny97:
Thanks for your contribution! Both PRs should be merged shortly.

fangyi-zhou:
Brilliant! Thank you and have a nice weekend!

Issue #248, opened 2025-05-18
Pyrefly should not allow forward reference variable names in function scope


grievejia:
### Describe the Bug

Minmal repro:
```python
def test() -> None:
    x = y
    y = 1
    reveal_type(x)
```
Expected: Error on first assignment. Revealed `Any`
Actual: No error on first assignment. Revealed `Literal[1]`

### Sandbox Link

https://pyrefly.org/sandbox/?code=CYUwZgBALiDOUAoCUEC0A+CA5A9gOxAC4AoCMiADwgF4IBPU8umiARkbIGIIAnEANxABDADYB9KHQAOIBBSRA

fangyi-zhou:
This seems have a similar root cause to #15.

I've tried to tackle this issue but it turns out to be more complicated that I thought, I'll nonetheless write down what I tried here, in hope that someone will validate my attempt and potentially give additional pointers.

When a successful name lookup is being validated in `BindingBuilder::ensure_name`:
https://github.com/facebook/pyrefly/blob/9c3cf0be61ea75e85bc65615648fb78d0f10492e/pyrefly/lib/binding/expr.rs#L230-L242

The flow style is checked for potential uninitialisation, this is done by inspecting the flow style. How the flow style is obtained is via `Scope::get_flow_style`:
https://github.com/facebook/pyrefly/blob/9c3cf0be61ea75e85bc65615648fb78d0f10492e/pyrefly/lib/binding/scope.rs#L784-L789

It's important to note, if the flow info lookup is unsuccessful (i.e. in this case when we look at `y` at the scope of `test`), the flow style will be defaulting to `Other`.

I've attempted to return the flow style as an `Option`, and then raise an error for unbound name whenever `None` is raised. This however becomes an overkill. As it would falsely raise an error for valid forward defined globals:

```python
def test():
  x = y

y = 1
```

It appears that the correct scope valiation would require the outer scopes to be resolved first, and then we can validate whether we're using any forward declared names if they are present at a later time in the same scope.

samwgoldman:
@fangyi-zhou thanks for your effort! I think modeling "uninitialized" in the flow state is a reasonable approach to solve this issue and address the linked #15 as well.

> I've attempted to return the flow style as an Option, and then raise an error for unbound name whenever None is raised. This however becomes an overkill. As it would falsely raise an error for valid forward defined globals:

[...]

PR #666, opened 2025-07-12
Raise error on variable use before write

### LINKED ISSUE #15: Uninitialized locals check
```
def foo():
    x  # <- This should error
    x = 5
```

https://pyrefly.org/sandbox/?code=CYUwZgBGD20BQEoBcAoC6IA90GIIB4BaCAFQAsBLAZwirOgFcAbYCEAJ3enbQ2wF4IAViA

fangyi-zhou:
Closes #15, #248

This change adds additional checks when building bindings, so that
variables that are used before initialised are correctly flagged as
errors.

Currently, we error on uninitialised variables by checking their flow
style. If the flow style is uninitialised or partially uninitialised, an
error is generated. When no flow style information is obtainable (which
is the case for use before write), a default "other" flow style value is
returned.

In this diff, we perform additional checks before returning a default
"other" value. By looking up the scopes for static information for
variables, we can correctly detect if a variable is declared in an upper
scope or in the current scope, and return "uninitialised" flow style
when the variable is used before write.

When doing so, we need to consider the special case for type usages,
so that name resolutions in types do not fail.

Help needed:
There's a failing test that I need some help: `test::flow::test_assert_not_in_flow`

```
testcase!(
    test_assert_not_in_flow,
    r#"
from typing import assert_type, Literal
if 0.1:
    vari = "test"
    raise SystemExit
assert_type(vari, Literal["test"])   <-- Shall we raise an error here?
"#,
);
```


pyrefly/lib/binding/scope.rs
```diff
@@ -807,10 +808,42 @@ impl Scopes {
...
+        }
+        for scope in iter {
+            if let Some(info) = scope.stat.0.get_hashed(name) {
+                return Some(info);
+            }
+        }
+        None
+    }
+
+    pub fn get_flow_style(&self, name: &Name, usage: &mut Usage) -> &FlowStyle {
```
stroxler: We probably should put a doc comment here, and explicitly call out that it's not really just getting the flow style anymore, it's doing additional checks of static info (with the appropriate semantic information) to track initialization except in type annotations

pyrefly/lib/binding/expr.rs
```diff
@@ -216,6 +216,7 @@ impl<'a> BindingsBuilder<'a> {
         &mut self,
         name: &Identifier,
         value: Result<Binding, LookupError>,
+        usage: &mut Usage,
```
stroxler: I'd prefer to use a boolean flag `used_in_static_type` here and everywhere further down the stack.

That makes the architecture clearer - `Usage` can potentially be used for a lot of things, whereas a boolean flag with a clear name does something pretty obvious

fangyi-zhou:
Rebased and addressed your inline comments. There should be only one test failing (which is the one I linked in the PR description)

yangdanny97:
Thanks!

For the failing test case `test_assert_not_in_flow`, I think raising an error there would be correct.

fangyi-zhou:
> Thanks!
> 
> 
> 
> For the failing test case `test_assert_not_in_flow`, I think raising an error there would be correct.

If it's not too inconvenient, could you fix that for me during the import? Thanks!

stroxler:
This may have to wait a day or two, we have some internal infra issues and I can't import the PR right now

stroxler (approved):
Looks good to me, I'll work on getting it merged. Thanks!

fangyi-zhou:
Thanks a lot! @stroxler

Issue #204, opened 2025-05-15
[Bug] No error on missing `__all__` implementation


kinto0:
Pyright errors in the following module:
```
__all__ = ["test"] # E: "test" is specified in __all__ but is not present in module  (reportUnsupportedDunderAll)
```

but not if you add `test`:
```
__all__ = ["test"]
test = 5
```

Pyrefly should implement this behavior.

kinto0:
this is more difficult than I thought. Naive implementation [here](https://github.com/facebook/pyrefly/pull/685).

It doesn't work on `dunderallentry::remove` at all. To solve that, we need to invoke the export's `exports()` like the following change in `step_exports`: 
```
        let exports = Exports::new(&ast.body, &load.module_info, ctx.sys_info);
        exports.exports(ctx.lookup);
        exports
```
but this causes a deadlock

stroxler:
I suspect this cannot be done in `definitions.rs`, which by design cannot know about all names imported using wildcards.

We likely need to do it in the bindings layer instead

PR #685, opened 2025-07-16
error on missing __all__ implementation

kinto0:
Summary:
if an item is present in __all__ but not implemented, pyre should raise an error.

See [this github issue](https://github.com/facebook/pyrefly/issues/204).

Rollback Plan:

Differential Revision: D77612155


yangdanny97:
closing because stale, we can reopen if someone wants to pick this up again later

PR #696, opened 2025-07-17
Move all the bundled code into pyrefly_bundled

ndmitchell:
Summary:
The main pyrefly crate had a lot of complexity and settings to deal with the fact it was bundling typeshed. Push that out into a separate crate.

I'll update the docs on regenerated typeshed after this lands, to make this diff as small as possible.

Reviewed By: stroxler

Differential Revision: D78481039


stroxler (approved):

Issue #403, opened 2025-05-30
[feature] warn on importing a deprecated class or function


yangdanny97:
### Describe the Bug

https://typing.python.org/en/latest/spec/directives.html#deprecated

See parent issue for context.

fangyi-zhou:
We can close this issue now that #791 got merged.

rchen152:
Thank you!

PR #699, opened 2025-07-17
Warn on `@deprecated` method calls

fangyi-zhou:
Fast follow from #416, adding deprecation warnings for method calls.

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -3962,6 +3962,66 @@
     },
     {
       "code": -2,
+      "column": 5,
+      "concise_description": "`+` is not supported between `Spam` and `Literal[1]`",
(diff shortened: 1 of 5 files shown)


fangyi-zhou:
Inspecting the conformance testsuite, we're missing the cases for handling `@deprecated`:
- importing deprecated names #403 https://github.com/facebook/pyrefly/blob/856bfbff3c381cc8434bd5d8e05890bb9d66dd55/conformance/third_party/directives_deprecated.py#L18
- function references https://github.com/facebook/pyrefly/blob/856bfbff3c381cc8434bd5d8e05890bb9d66dd55/conformance/third_party/directives_deprecated.py#L25
- function overloading https://github.com/facebook/pyrefly/blob/856bfbff3c381cc8434bd5d8e05890bb9d66dd55/conformance/third_party/directives_deprecated.py#L30

We probably should file separate tracking issues for the function references, and overloading

migeed-z (approved):

Issue #708, opened 2025-07-18
Irregularities in Pattern Matching `T | Tuple[T,T]`


anabelle2001:
### Describe the Bug

I often use this pattern when writing a variadic `__getitem__`. It works at runtime, but Pyrefly does not like it:
```py
from typing import *

class T: ...

def my_func(x: T | tuple[T,T]):
    match x:
        case T(),T():  # Type `T` is not iterable
            print("Two T")
        case T():
            print("One T")

my_func(T())
my_func((T(),T()))


```
### Sandbox **Link**

https://pyrefly.org/sandbox/?code=MQAg6gpgNgxg9gWwiALnVALZAFAngJwgDMpcQBlAQwDsATAIzgA8AaTASwGcQuRKQADpQDmEAFCgA7lkIhccAK4gYNEJPzsUENpwyVZNWiCgR91PowUpUuAe2rCe5vCgxxqAOgkSQACQgGBsqKAiYgRKYoCoTccEQc3JyGjEwAXD4AtCAA4hDWhJRQGSjsSDYCyDBYMADW9o4REAyUtZkgAKqcyIQAbqZQAPoothAAFACUqOj2nBUw1vYR+IRGwxWcbRBMFRpI1NaSmhiYyACSACIAouGR0RDcrpQHyJwKAgJw+NajEB7CHnwrHB4AhQnltCBhHBitDaMR7Jp2O42G4+vg2HkYOM2mANFo+HQQPgFM5cK53ME4SBaOxCPNSE45Ip8CB6Pg4JIuvgxGIiOyEOV6jxQZ9rAAqHkwKCUTjcAAqqRAHmVPLh8QQuAGRBJMFGaRAcpAAB9UG8TABtOUsOUAXXG6RAjpACCeVRAaTETq9yhlyDlE2tE0VIFAcpGIAABnKIzxuNQ4AstPhKPQTJ7vRmBBp9qMAERyyToOW57EZx0qLoGoPpstOrP2FB5gDy1D9JZ5Gq1OtG-vG2M72uoup7Ad7faAA

Tejas-Ketkar:
I noticed this issue while looking at another bug. I'll try a fix.

randolf-scholz:
[...]

PR #710, opened 2025-07-18
Bug Fix 708 - Union Iterability Checker

Tejas-Ketkar:
This PR tries to solve bug #708. The problem with the current implementation was that given a union type, solve.rs would iterate over each of the two types without considering whether both types were iterable. The fix tries to check if each option in the union is iterable, checking for explicit `__iter__` or seeing if it behaves like a tuple or is a tuple. The original bug was added as a test to pattern_match.rs.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -560,6 +560,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    fn is_iterable(&self, ty: &Type) -> bool {
+        self.unwrap_iterable(ty).is_some()
+            || matches!(ty, Type::Tuple(_))
+            || matches!(ty, Type::ClassType(cls) if self.as_tuple(cls).is_some())
+    }
+
     /// Given an `iterable` type, determine the iteration type; this is the type
     /// of `x` if we were to loop using `for x in iterable`.
     ///
@@ -585,11 +591,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             Type::Union(ts) => ts
                 .iter()
                 .filter_map(|t| {
-                    let is_iterable = self.unwrap_iterable(t).is_some()
-                        || matches!(t, Type::Tuple(_))
-                        || matches!(t, Type::ClassType(cls) if self.as_tuple(cls).is_some());
-
-                    if is_iterable {
+                    if self.is_iterable(t) {
                         Some(self.iterate(t, range, errors))
                     } else {
                         None
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -584,7 +584,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
             Type::Union(ts) => ts
                 .iter()
-                .flat_map(|t| self.iterate(t, range, errors))
+                .filter_map(|t| {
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/solve.rs
```diff
@@ -584,7 +584,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
             Type::Union(ts) => ts
                 .iter()
-                .flat_map(|t| self.iterate(t, range, errors))
+                .filter_map(|t| {
+                    let is_iterable = self.unwrap_iterable(t).is_some()
```
migeed-z: can we add is_iterable be added to types.rs and invoked from there instead?

migeed-z (changes requested):

pyrefly/lib/alt/solve.rs
Tejas-Ketkar: Moving the function out is a good idea, but I think placing the function types.rs won't work since pyrefly_types is in a different crate. From the way it's defined right now we'd end up with circular dependencies?

stroxler: I think that's correct, it needs access to `self`.

Let's just pull it into a helper method?

We probably should make an `AnswersSolver` module for "type utilities" that require access to a solver and therefore cannot be in `types.rs` at some point.

stroxler:
This looks great, thanks for the contribution!

I think it would be nice to extract the helper method, after that I could work on getting it merged.

Tejas-Ketkar:
Done! Let me know if there's any other changes needed.

stroxler (approved):

stroxler:
LGTM, I'll aim to get it merged soon

ndmitchell:
I think this is wrong, as given the example:

```


def f(x: int | tuple[int, int]):
    for y in x:
        reveal_type(y)
```

Before this diff we get an error `Type int is not iterable`, and after we succeed. My best guess is that for pattern matching, where we are asking "does this type match", we should be making a suitable check. But for other uses of iterate, we should require it to match. But I don't know enough about how pattern matching works to know what to do there.

stroxler:
Oh good point - we probably need something like this logic, but only used in match.

The idea is that if we see a tuple form in a match, we should take all the iterable cases of a union as potential hits (so if we have `x: float | tuple[int, int] | list[bytes]` then matching `x` against `a, b` should give us `tuple[int, int]` and `list[bytes]` as the potential union cases that match, but the `float` is eliminated)

It might actually need to be done as a narrowing operation

cc @yangdanny97  who has the most context on matching

yangdanny97:
@stroxler We do have narrowing that checks sequence length, but it doesn't exclude types that can't be sequences. I don't think we can modify it directly because it's used for stuff like `if len(x) == 2`, so maybe we could introduce a new NarrowOp and generate a compound narrow op for match statements.

yangdanny97 (changes requested):
@Tejas-Ketkar there was a bit of back and forth discussion on this PR earlier, but it sounds like the next step is to update the PR it s.t. the checker only works when pattern matching.

Tejas-Ketkar:
Thanks, I'll look into making the fix.

samwgoldman:
Hi @Tejas-Ketkar, are you still planning to work on this?

rchen152:
Hey @Tejas-Ketkar, I'm going to close this PR since it's been inactive for a while. Thanks for tackling this issue, and please open a new PR if you're interested in taking another stab at it!

PR #711, opened 2025-07-19
Warn on calling `@deprecated` overload

fangyi-zhou:
WIP: need to tidy up the code a little bit before marking this PR ready to review.


rchen152:
This looks great to me, thank you! I'll work on getting it merged.

samwgoldman:
Would it be better to change `OverloadType::Callable(Callable)` to `OverloadType::Callable(Function)`, so we preserve the function metadata, which includes `is_deprecated`?

(Edit: on second look, we store the function metadata on the side with overloads -- can we use that metadata instead of introducing a new struct OverloadSignature?)

fangyi-zhou:
> Would it be better to change `OverloadType::Callable(Callable)` to `OverloadType::Callable(Function)`, so we preserve the function metadata, which includes `is_deprecated`?
> 
> 
> 
> (Edit: on second look, we store the function metadata on the side with overloads -- can we use that metadata instead of introducing a new struct OverloadSignature?)

I'm not against the idea. I was thinking that since only the deprecation flag is needed, I might as well drop the unnecessary information. If we're going to find uses of the other metadata, we might as well keep the entire metadata struct.

samwgoldman:
The metadata is already part of the CallTarget, so I think we can avoid cloning the metadata. If creating this struct avoids a bunch of allocations, then I'm more supportive, but at a glance it looks like additional allocations can be avoided.

fangyi-zhou:
Rebased and implemented the required change :)

fangyi-zhou:
Do you think I shall also replace `OverloadTarget` with `Function` as well?

stroxler (approved):

samwgoldman (approved):
Thank you!

PR #715, opened 2025-07-20
Fix caught exception leaking through `except` scope

### LINKED ISSUE #154: Del the name of the exception at the end of the exception handler
Minimal repro:
```
try:
    1 / 0
except Exception as e:
    pass
print(e)
```
Expected: Name `e` is uninitialized at the print statement
Actual: No error

fangyi-zhou:
As reported in #154, if a caught exception is bound to a name `e` in a `except` block, the caught name `e` should be out of scope after the corresponding `except` scope.

This is reflected in the spec as follows
https://docs.python.org/3/reference/compound_stmts.html#except-clause:

> When an exception has been assigned using as target, it is cleared at the end of the except clause.

This PR aims to model this effect by marking the caught exception name as uninitialised at the end of the current `except` block.

Closes #154.

### DIFF
--- a/pyrefly/lib/test/scope.rs
+++ b/pyrefly/lib/test/scope.rs
@@ -506,6 +506,8 @@ def try_except_else():
     else:
         e6 # E: `e6` is uninitialized
 
+    e6 # E: `e6` is uninitialized
+
 def try_except_else_finally():
     try:
         1 / 0
@@ -517,5 +519,7 @@ def try_except_else_finally():
         e7 # E: `e7` is uninitialized
     finally:
         e7 # E: `e7` is uninitialized
+
+    e7 # E: `e7` is uninitialized
 "#,
 );


pyrefly/lib/binding/stmt.rs
```diff
@@ -755,12 +782,39 @@ impl<'a> BindingsBuilder<'a> {
                             Binding::ExceptionHandler(type_, x.is_star),
                         );
                     }
+
+                    if let Some(ref name) = last_caught_exception_name {
```
yangdanny97: I'm confused why this block & the one on 816 are necessary - doesn't each exception get marked as uninitialized on 799?

fangyi-zhou: ~~I'll try to give a more detailed answer later. In short, this has something to do with how we error on uninit variable (cf. PR #666).~~

~~If we only error at the end of the except scope, then the uninit does nothing in the next scope since silent lookup failures won't error.
So we inject an uninit to make it error.~~

~~On the other hand, we get slightly confusing and error messages of "maybe uninit" if we only inject uninit in the next scope due to how flow merging works.~~

~~I concede this is a somewhat hacky solution, let me know if you have a better idea. Sorry for the concise explanation.~~

Edit: Please disregard what I said previously. I agree with you that this is unnecessary

fangyi-zhou:
Sorry for the earlier confusion. I was under the impression that adding another uninit in the next scope was necessary, but I seem to got confused by myself earlier during debugging. I've validated my first commit and concluded that adding an uninit in the end of the scope should be sufficient.

It looks like I was chasing a non-existent bug earlier; but on the positive side, the solution to the issue is straightforward.

stroxler (approved):

PR #728, opened 2025-07-22
Return Result for all absolutize calls

connernilsen:
Summary:
This diff rewrites some of our logic to return a `Result` instead of just calling `unwrap()` on a `Path::absolutize()` call. `Path::absolutize()` and `absolutize_from()` have signatures returning `Result`, but in practice, don't ever actually return one (`Absolutize::absolutize_virtually()` does return real results though).

We should probably handle them properly, just in case an update ever comes out or something changes.

Differential Revision: D78754538


connernilsen:
doing a different approach

PR #729, opened 2025-07-22
Absolutize in config path rewriting

connernilsen:
Summary:
Use absolutize calls in config path rewriting.

I'm fairly confident these two are the same:
```
let x = config_path.clone();
x.push(config.other_path);
config.other_path = x
```
```
config.other_path.absolutize_from(config_path);
```

So we should do that. Also, we should make sure any paths that we use here are absolutized. I saw in a test I was doing that `pyrefly dump-config` was still showing relative paths in a way that we really don't want to be using them.

Differential Revision: D78764725


connernilsen:
doing a different approach

PR #731, opened 2025-07-22
Fix some clippy issues on main

fangyi-zhou:
Ran `cargo clippy --fix` and also applied fixes manually for preferring `matches!` macros


migeed-z:
Thanks! will work on merging

Issue #730, opened 2025-07-22
Removal of `--no-summary` flag breaks mypy-primer workflow


fangyi-zhou:
### Describe the Bug

https://github.com/facebook/pyrefly/commit/199458e801b271a3c869c5867ba79b222f87859c removes the `--no-summary` flag, and proposes `--summary=none` as a replacement. This unfortunately break the mypy-primer workflow on GitHub, which is run for Pull Requests.

The removed flag is used in mypy-primer here: https://github.com/hauntsaninja/mypy_primer/blob/092737ed53f28191cf7a7760853319634d288b16/mypy_primer/model.py#L354

fangyi-zhou:
Example of stuck workflow: https://github.com/facebook/pyrefly/actions/runs/16456719641/job/46515449116?pr=727#step:6:80

yangdanny97:
I'll make a PR to mypy primer after our next release. 0.25.0 from yesterday still has `--no-summary`

yangdanny97:
cc @ndmitchell

ndmitchell:
Oops. I'll reinstate `--no-summary`, then after the next release we can adjust mypy primer, and then after the release after that remove `--no-summary` again.

fangyi-zhou:
FYI: there are some reference in the current codebase that uses `--no-summary`, probably should be cleaned up in the future as well

https://github.com/facebook/pyrefly/blob/59a0fef777e7df0b9615a669139b10efe0299b22/pyrefly/lib/commands/init.rs#L142
https://github.com/facebook/pyrefly/blob/59a0fef777e7df0b9615a669139b10efe0299b22/pyrefly/lib/commands/init.rs#L299

PR #736, opened 2025-07-23
Migrate `--no-summary` to `--summary=none` in init command

fangyi-zhou:
As noted in #730, the deprecated flag is used in the `init` command for suppressing errors. This is an easy change to migrate the flag to `--summary=none`.

Tested `pyrefly init` on a library manually and saw no regression


ndmitchell:
Thanks!

PR #781, opened 2025-07-29
Point unittests to site-package-path

migeed-z:
Summary:
My goal is to set unittests that can see pydantic definitions.

We will use stubs for now for finer granularity tests and also setup integration tests for coarser granularity testing against the real library.

For this testcase, I'm following the structure of pydantic by having a main and an init file.

I'm also hacking the Testcase macro and writing some simple stubs and linking them to the unittests via setting the site package path.

Risks
Stubs are written manually so far, which can always be error prone as we scale. But this will me mitigated by integration tests. Manual stubs can also be time consuming to write out manually.

 The pros to the approach though is that we keep the dependencies minimal, instead of relying on the entire library.

Differential Revision: D79131498


rchen152 (changes requested):

rchen152 (approved):

PR #791, opened 2025-07-31
Warn on importing deprecated symbols

fangyi-zhou:
This commit implements warning for importing deprecated symbols (#403).

We collect deprecated symbols when building the exports, by checking the decorators when processing function definitions and class definitions. The deprecated symbols are then stored as part of the export information.

When importing new symbols, we check whether the exported name is deprecated and generate a warning accordingly.

Limitations/Discussions points:
- The last piece of work for deprecation warning is to warn on access a deprecated symbol (e.g. passing function reference, or accessing a deprecated property). It would probably be beneficial if we provide functions to check whether a symbol/name is deprecated, and then integrate that with import checks
- Checking decorators with strings in definition builder is a bit hacky and I don't like it; but otherwise processing the deprecation information inside a binding builder would require some design for threading the required information across components.
- ~~I haven't tested re-exports of deprecated symbols, probably they don't work.~~

### DIFF
--- a/pyrefly/lib/test/decorators.rs
+++ b/pyrefly/lib/test/decorators.rs
@@ -209,7 +209,7 @@ testcase!(
     test_invalid_top_level_function_decorators,
     r#"
 from typing import *
-from abc import abstractstaticmethod, abstractmethod
+from abc import abstractstaticmethod, abstractmethod # E: `abstractstaticmethod` is deprecated
 from enum import member, nonmember
 
 @member  # E: can only be used on methods
@@ -497,13 +497,13 @@ testcase!(
     r#"
 from typing import overload, Any
 @overload
-def foo(a: int) -> int: ...  
+def foo(a: int) -> int: ...
 @overload
 def foo(a: str) -> str:
     """Docstring"""
 def foo(*args, **kwargs) -> Any:
     pass
-    
+
     "#,
 );
 
@@ -512,14 +512,14 @@ testcase!(
     r#"
 from typing import overload, Any
 @overload
-def foo(a: int) -> int: ...  
+def foo(a: int) -> int: ...
 @overload
-def foo(a: str) -> str: 
+def foo(a: str) -> str:
     """Docstring"""
     return 123             # E: Returned type `Literal[123]` is not assignable to declared return type `str`
 def foo(*args, **kwargs) -> Any:
     pass
-    
+
     "#,
 );
 


fangyi-zhou:
Let me know of your thoughts and we can work together to iterate on this implementation a bit

yangdanny97:
Thanks for making this PR! We'll get this reviewed soon

pyrefly/lib/binding/stmt.rs
```diff
@@ -901,6 +903,18 @@ impl<'a> BindingsBuilder<'a> {
                                 if &x.name == "*" {
                                     for name in module_exports.wildcard(self.lookup).iter_hashed() {
                                         let key = Key::Import(name.into_key().clone(), x.range);
+                                        if let Some(ExportLocation::ThisModule(Export {
```
migeed-z, line 906: I am not super familiar with deprecated, but this does makes sense. A few questions though. Why did we choose to handle the import this way instead of as a special export?

pyrefly/lib/export/exports.rs
```diff
@@ -39,6 +39,7 @@ pub struct Export {
     pub location: TextRange,
     pub symbol_kind: Option<SymbolKind>,
     pub docstring_range: Option<TextRange>,
+    pub is_deprecated: bool,
```
migeed-z, line 42: Is adding a new field here necessary/do we need to attach deprecated to the definition?

migeed-z:
If it's possible to rebase, that would be great!

pyrefly/lib/binding/stmt.rs
fangyi-zhou: > Why did we choose to handle the import this way instead of as a special export?

I considered modelling the decorator as a special export. Correct me if I'm wrong, I'm trying to check for the decorator in the definition builder, and there is no simple way to check whether a decorator is a special export. Hence my hacky solution. If there's a better way please let me know!

pyrefly/lib/export/exports.rs
fangyi-zhou: Good question. I did it this way because it was the easiest, as in the struct `Export` is already being used when importing a name, so adding the flag there would be sufficient.

I suppose it is also possible to create another method for `Exports` struct and return a set of deprecated names. Do you think that would be more appropriate?

fangyi-zhou:
> If it's possible to rebase, that would be great!

Thanks for the review! Rebased the PR.

pyrefly/lib/binding/stmt.rs
migeed-z: I agree that we should not do this for the decorator part. I was just curious about part where we detect a depreciated import itself and model it as a special export/why we chose to do it this way.

pyrefly/lib/export/exports.rs
migeed-z: Tagging @rchen152  to weigh in on this part. Everything else LGTM.

migeed-z:
@fangyi-zhou there are a couple of unittests that are failing. Does this need rebasing again? :) 

I plan to export and work on this afterwards

fangyi-zhou:
Okay I probably need to generate the conformance test results again, but this PR shouldn't break any unit test.

fangyi-zhou:
I've added more unit tests and fixed the issue with handling overload decorator. This should be ready to review now.

PR #801, opened 2025-07-31
Pypi -> PyPI

jack-mcivor:
That's how PyPI styles themselves (eg https://docs.pypi.org/)

> PyPI is the official repository of packages for Python


jack-mcivor:
check CLA again?

Issue #511, opened 2025-06-17
Error when defining `@final` classes that have unimplemented abstract methods


grievejia:
### Describe the Bug

Minimal repro:
```python
from abc import ABC, abstractmethod
from typing import final

class B(ABC):
    @abstractmethod
    def foo(cls) -> None: ...

@final
class D(B):
    pass
```
Expected: Error on class `D()` since `foo()` has not been implemented yet
Actual: No error

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBAhgRgYygSwgBzCALlAggIQGEAaWOAZyxBgSwgFMsALMAEwChRIosBPNZADsA5inSYcwITAA27dghkxy5KPgAUBQgEoAXOyiGoAAXiVqtBszYGjresCjAwYdYvLaoAWgB8UAHJggvS6UAB0EfLGUoKyCkoqUAAi6vh6toZoyuRAA

grievejia:
Related issue: #510

yangdanny97:
The infrastructure for this has been set up so it should be ready for someone to pick up

https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/alt/class/class_metadata.rs#L1123

Tests can be added here

https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/abstract_class_instantiation.rs

PR #807, opened 2025-08-01
Throws Bad Instantation Error when Creating an Instance of an Abstract Class

### LINKED ISSUE #592: [feature] emit an error when trying to call an abstract method

Adist319:
Resolves #592.

We now only throw an error when creating an instance of an abstract class, instead of when an abstract method is called.

**Behavior:**
  ```python
  # Before: No error thrown

  # After: BadInstantiation error
  class Shape(ABC):
      @abstractmethod
      def area(self): pass

  Shape()  # Error: Cannot instantiate abstract class `Shape`

  # Concrete implementations work fine
  class Circle(Shape):
      def area(self): return 3.14

  Circle()  # works

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -408,7 +408,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             );
         }
 
-        let mut overall_ret = None;
         if let Some(hint) = hint {
             self.solver()
                 .freshen_class_targs(cls.targs_mut(), self.uniques);


yangdanny97:
Hmm, mypy primer runs are panicking. We'll get this reviewed and see if we can figure out what's going on

pyrefly/lib/alt/call.rs
```diff
@@ -396,6 +396,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         hint: Option<&Type>,
     ) -> Type {
         // Based on https://typing.readthedocs.io/en/latest/spec/constructors.html.
+        let instance_ty = Type::ClassType(cls.clone());
```
yangdanny97, line 399: this appears unused?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1782,4 +1782,117 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+    fn inherits_from_abc(&self, cls: &ClassType) -> bool {
+        let class_obj = cls.class_object();
+
+        // Check if this class is abc.ABC itself
+        if class_obj.module_name().as_str() == "abc" && class_obj.name().as_str() == "ABC" {
+            return true;
+        }
+
+        // Recursively check base classes
+        let metadata = self.get_metadata_for_class(class_obj);
```
yangdanny97, line 1800: I think we can pre-calculate `inherits_from_abc` and store it in class metadata. cc @grievejia

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1782,4 +1782,117 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        // Then collect abstract methods from this class
+        for field_name in class_obj.fields() {
+            if let Some(field) = self.get_from_class(
+                class_obj,
+                &KeyClassField(class_obj.index(), field_name.clone()),
+            ) {
+                match &field.0 {
+                    ClassFieldInner::Simple { ty, .. } => {
+                        if let Type::Function(function) = ty {
+                            if function.metadata.flags.is_abstract_method {
```
yangdanny97, line 1867: we might need to handle overloaded functions too

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1782,4 +1782,117 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            if !self.class_provides_concrete_implementation(cls, method_name) {
+                return true;
+            }
+        }
+
+        false
+    }
+
+    /// Recursively collect abstract methods from a class and its bases
+    fn collect_abstract_methods_from_class(
```
yangdanny97, line 1845: I'm not sure we need to traverse all the base classes like this. If we look up each field the class has like so https://github.com/facebook/pyrefly/blob/b35b7ab29e154af6cb7acaad495afd855e1d33ef/pyrefly/lib/alt/class/class_field.rs#L1589 it would handle the lookup, and we can check if the field it returns is abstract or not.

Of course, this relies on the MRO being correct and able to handle cases when one base class declares something as abstract and another base class provides the impl. @stroxler would know more about this

Adist319:
@yangdanny97 Thanks for the feedback! I just wanted to check if the panics have been root-caused yet, or if that’s still being investigated, so I know when I can continue work on this PR.

pyrefly/lib/alt/class/class_field.rs
Adist319: I like this idea, we could do something like: 
```
pub struct ClassMetadata {
       // ... existing fields ...
       inherits_from_abc: bool,  // New field
}
```

And then we could populate ABC Inheritance during class metadata creation.

yangdanny97:
@Adist319 

The panics are caused by the changes this PR, probably to handling descendants of the `PathLike` class (that's what declares the `__fspath__` we see in the panic messages.

You can repro this by checking a simple snippet.
```
from pathlib import Path
p = Path("x")
```

This didn't show up in the unit tests, because we emit a DEBUG message instead of panic: https://github.com/facebook/pyrefly/blob/29aa8114214cfbd697ab46b2de120b4049f185e3/pyrefly/lib/state/state.rs#L1517

cc @ndmitchell I'm a bit confused about what this does, or why we don't always panic

Adist319:
@yangdanny97 thanks for looking into it , will get back to this soon.

Adist319:
@yangdanny97 Apologies for the extended delay, but I think I've made good progress here.
I've implemented abstract class instantiation checking for #592 and #511 - abstract classes can't be instantiated and @final validation works. But I ran into a problem:
3 pre-existing tests are now failing:
- `test::decorators::test_abstract_classmethod`
- `test::decorators::test_abstract_staticmethod`
- `test::descriptors::test_abstract_property`

These test deprecated Python decorators (`@abstractclassmethod`, `@abstractstaticmethod`, `@abstractproperty` - deprecated 10+ years ago I believe?).

**Current behavior:**
- Abstract checking works (can't instantiate)
- Type resolution returns `Any` instead of proper type

```python
class C(abc.ABC):
    @abc.abstractclassmethod
    def f(cls) -> int:
        return 42

assert_type(C.f(), int)  # FAILS: Got Any, expected int
```

## Question
Should I fix type resolution for these deprecated decorators or can we skip these 3 tests?

I can create a new pull request once all tests pass.

yangdanny97:
Thanks for continuing to work on this! Do you mind updating the PR so I can pull the changes and dig a bit into what's happening?

Given that we're returning `Any` instead of emitting a spurious error, I'm tentatively OK with skipping those tests as long as the new version decorated w/ both `@abstract` `@classmethod` behaves correctly.

Adist319:
New PR #1185 for this, trying to resolve a delimiter issue after resolving some merge conflicts. But feel free to take a look at the changes first.

yangdanny97:
we're working on #1192 now so I'll close this one

Issue #784, opened 2025-07-30
Pyrefly VSCode extension doesn't respect conda environment for selecting the interpreter


Dogacel:
### Describe the Bug

When using the pyrefly vscode extension, the interpreter is not choosen correctly if the environment is managed with Conda. For example, I have a named environment called `mor` and it uses python 3.12.11. However, pyrefly chooses to use the main interpreter for my conda installation.

I am seeing imports fail with the following hint,

```
 Site package path queried from interpreter: ["/home/dogac/miniconda3/lib/python3.13", "/home/dogac/miniconda3/lib/python3.13/lib-dynload", "/home/dogac/miniconda3/lib/python3.13/site-packages"]Pyrefly(import-error)
```
However my environment uses the conda env `mor` with python 3.12 so the interpreter path should be `/home/dogac/miniconda3/envs/mor/bin/python`.

I suspsect the bug exists here:
https://github.com/facebook/pyrefly/blob/5e4e4d409b75cb76da60912be9f6f6418eabb7c1/lsp/src/extension.ts#L62

It should be fairly easy to reproduce by installing conda, creating a named environment with python 3.12 and install some packages.

connernilsen:
Hey @Dogacel, thanks for reporting this. Would you be able to paste your `pyrefly.toml` or the Pyrefly section of your `pyproject.toml` here if you have one? Also, can you run `pyrefly dump-config` from the root of your project if you have it runnable from CLI?

There's a design choice I made a while ago to force Pyrefly to only use the interpreter or conda environment specified in a config file, if one is specified, and ignore a vscode-supplied interpreter, but I've been thinking about changing it for a while now. It may be related to this, but just want to double check.

Dogacel:
> Hey [@Dogacel](https://github.com/Dogacel), thanks for reporting this. Would you be able to paste your `pyrefly.toml` or the Pyrefly section of your `pyproject.toml` here if you have one? Also, can you run `pyrefly dump-config` from the root of your project if you have it runnable from CLI?
> 
> There's a design choice I made a while ago to force Pyrefly to only use the interpreter or conda environment specified in a config file, if one is specified, and ignore a vscode-supplied interpreter, but I've been thinking about changing it for a while now. It may be related to this, but just want to double check.

Sure, I don't have neither `pyrefly.toml` nor `pyproject.toml`. Let me dump my vscode project settings and the output of pyrefly.

This is my workspace vscode settings,

```json
{
[...]

PR #830, opened 2025-08-02
Move precedence of configuration of LSP/IDE provided interpreters be higher

Dogacel:
Relates to 
- #784

When running pyrefly as an IDE plugin for VSCode, the `CONDA_ENV` environment variable set, however it points to the wrong environment. I'm not exactly sure why it points to the wrong environment, so I want to override the interpreter path. However it is not possible to override the python interpreter path for `pyrefly`, because configuration values in `pyrefly.toml` has less precedence than environment variables.

I believe IDE / LSP configuration should take precedence over environment variables, otherwise VSCode extension chooses the wrong environment.

### DIFF
--- a/crates/pyrefly_config/src/environment/interpreters.rs
+++ b/crates/pyrefly_config/src/environment/interpreters.rs
@@ -78,8 +78,8 @@ impl Interpreters {
         self.python_interpreter.is_none() && self.conda_environment.is_none()
     }
 
-    pub fn set_python_interpreter(&mut self, interpreter: PathBuf) {
-        self.python_interpreter = Some(ConfigOrigin::auto(interpreter));
+    pub fn set_lsp_python_interpreter(&mut self, interpreter: PathBuf) {
+        self.python_interpreter = Some(ConfigOrigin::lsp(interpreter));
     }
 
     /// Finds interpreters by searching in prioritized locations for the given project
@@ -87,17 +87,22 @@ impl Interpreters {
     ///
     /// The priorities are:
     /// 1. Check for an overridden `--python-interpreter` or `--conda-environment`
-    /// 2. Check for an active venv or Conda environment
-    /// 3. Check for a configured `python-interpreter`
-    /// 4. Check for a configured `conda-environment`
-    /// 5. Check for a `venv` in the current project
-    /// 6. Use an interpreter we can find on the `$PATH`
-    /// 7. Give up and return an error
+    /// 2. Check for an IDE / LSP provided `python-interpreter`.
+    /// 3. Check for an active venv or Conda environment
+    /// 4. Check for a configured `python-interpreter`
+    /// 5. Check for a configured `conda-environment`
+    /// 6. Check for a `venv` in the current project
+    /// 7. Use an interpreter we can find on the `$PATH`
+    /// 8. Give up and return an error
     pub(crate) fn find_interpreter(
         &self,
         path: Option<&Path>,
     ) -> anyhow::Result<ConfigOrigin<PathBuf>> {
         if let Some(interpreter @ ConfigOrigin::CommandLine(_)) = &self.python_interpreter {
+            return Ok(interpreter.clone());
+        }
+
+        if let Some(interpreter @ ConfigOrigin::Lsp(_)) = &self.python_interpreter {
             return Ok(interpreter.clone());
         }
 
@@ -196,6 +201,23 @@ mod test {
         let tempdir = setup_test_dir();
 
         let python_interpreter = ConfigOrigin::cli(PathBuf::from("asdf"));
+
+        let interpreters = Interpreters {
+            python_interpreter: Some(python_interpreter.clone()),
+            ..Default::default()
+        };
+
+        assert_eq!(
+            interpreters.find_interpreter(Some(tempdir.path())).unwrap(),
+            python_interpreter
+        );
+    }
+
+    #[test]
+    fn test_find_interpreter_precedence_lsp_highest_priority() {
+        let tempdir = setup_test_dir();
+
+        let python_interpreter = ConfigOrigin::lsp(PathBuf::from("asdf"));
 
         let interpreters = Interpreters {
             python_interpreter: Some(python_interpreter.clone()),
--- a/crates/pyrefly_config/src/environment/interpreters.rs
+++ b/crates/pyrefly_config/src/environment/interpreters.rs
@@ -87,9 +87,9 @@ impl Interpreters {
     ///
     /// The priorities are:
     /// 1. Check for an overridden `--python-interpreter` or `--conda-environment`
-    /// 2. Check for a configured `python-interpreter`
-    /// 3. Check for a configured `conda-environment`
-    /// 4. Check for an active venv or Conda environment
+    /// 2. Check for an active venv or Conda environment
+    /// 3. Check for a configured `python-interpreter`
+    /// 4. Check for a configured `conda-environment`
     /// 5. Check for a `venv` in the current project
     /// 6. Use an interpreter we can find on the `$PATH`
     /// 7. Give up and return an error
@@ -108,6 +108,10 @@ impl Interpreters {
                 .transpose_err();
         }
 
+        if let Some(active_env) = ActiveEnvironment::find() {
+            return Ok(ConfigOrigin::auto(active_env));
+        }
+
         if let Some(interpreter) = &self.python_interpreter {
             return Ok(interpreter.clone());
         }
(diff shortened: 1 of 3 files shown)


samwgoldman:
Thanks for the thoughtful analysis! @connernilsen is the expert on this stuff. What you’re saying makes good sense to me, but Conner might know of some downside to this approach. Mind taking a look, Conner?

connernilsen:
Hey @Dogacel, thanks for your interest in improving Pyrefly and for looking into it yourself! This looks like it's definitely moving in the right direction.

I'm hesitant to change the behavior of selecting environment variables over everything else we currently have, since it's the easiest way to test out using a different venv/enviornment with a project (you don't have to mess around with a config, and you just say `venv/bin/activate` or `conda activate <env>`).

I think what we should do instead is create a new `ConfigOrigin` enum variant (like `VSCode`) or rename `CommandLine` to `ExplicitOverride` or something, and have that be the first condition we hit. Would you be open to making that change, or would you like us to do it?

samwgoldman:
Thanks, Conner! I am curious about something from the original report:

> When running pyrefly as an IDE plugin for VSCode, the CONDA_ENV environment variable set, however it points to the wrong environment. I'm not exactly sure why it points to the wrong environment

I'd love to know what is happening here. If something is simply misconfigured, maybe we don't need to change any behavior, but we can make the misconfiguration more debuggable? Or maybe it's common/expected to have CONDA_ENV set like this and we really need to work around?

connernilsen:
@samwgoldman sure! The logic I have implemented right now prefers an "active environment" over a configured one. This is because the assumption might be that you want to test out a different environment than the one you have configured. Otherwise, we'll fall back to what's in a config file, a venv in the current directory, then any `python3`/`python` executable we can find. 

Most of the time, activating an environment is an explicit step that has to be taken, so that's why it's preferred right now. For Conda, I think it might be that some setups might try to auto activate it so that the conda Python executable is available on the path. I assume in this case, when starting VSCode, the non-interactive startup script is sourcing conda's setup, and we're finding the "active environment" before what's set by our LSP logic.

The solution that I'm thinking of here is that when we're setting the interpreter we want to use in the config, we should say that VSCode's interpreter is the highest priority. We do this with a transparent wrapper around the values that tell us where they came from, and in my opinion, it's a bug (that I missed 😅) where the LSP priority was set too low (at the same priority as us auto finding something). 

I think another part of the solution could be to update the `ConfigOrigin` format function to say "<interpreter path> (from <origin>)`, where origin might be one of `VSCode`, `CLI`, `Config`, or `Autoconfiguration`.

Dogacel:
> Hey @Dogacel, thanks for your interest in improving Pyrefly and for looking into it yourself! This looks like it's definitely moving in the right direction.
> 
> I'm hesitant to change the behavior of selecting environment variables over everything else we currently have, since it's the easiest way to test out using a different venv/enviornment with a project (you don't have to mess around with a config, and you just say `venv/bin/activate` or `conda activate <env>`).
> 
> I think what we should do instead is create a new `ConfigOrigin` enum variant (like `VSCode`) or rename `CommandLine` to `ExplicitOverride` or something, and have that be the first condition we hit. Would you be open to making that change, or would you like us to do it?

Hi @connernilsen, what you said makes sense, as I am less experienced with different setups. Do you think this new `Origin` will be a VSCode configuration setting rather than toml override or command line parameter?

@samwgoldman so, my issue only exists with pyrefly, when I disable it the default type checker finds the imports fine. My issue here is precedence of `CONDA_ENV` is so high that I can't override it, therefore I can't even workaround. I had some hopes that this line would override the interpreter somehow,

https://github.com/facebook/pyrefly/blob/5e4e4d409b75cb76da60912be9f6f6418eabb7c1/lsp/src/extension.ts#L108

But it seems like this serves a different purpose.

Dogacel:
So couple more observations,

1. I have increased precedence of `workspace.configuration.python_interpreter` to be higher than the environment variable. I have done it by adding a new config origin.
2. However it didn't fix my issue, as pyrefly picks up the right environment after some time.

Here are the logs

```
starting generic LSP server
Reading messages
Finding Python interpreter...
Interpreters: <none found successfully>
Finding Python interpreter...
Interpreters: <none found successfully>
Finding Python interpreter...
Interpreters: <none found successfully>
Using interpreter site package path: ["/home/dogac/miniconda3/lib/python3.13", "/home/dogac/miniconda3/lib/python3.13/lib-dynload", "/home/dogac/miniconda3/lib/python3.13/site-packages"]
Using interpreter site package path: ["/home/dogac/miniconda3/lib/python3.13", "/home/dogac/miniconda3/lib/python3.13/lib-dynload", "/home/dogac/miniconda3/lib/python3.13/site-packages"]
Finding Python interpreter...
Interpreters: <none found successfully>
Using interpreter site package path: ["/home/dogac/miniconda3/lib/python3.13", "/home/dogac/miniconda3/lib/python3.13/lib-dynload", "/home/dogac/miniconda3/lib/python3.13/site-packages"]
Using interpreter site package path: ["/home/dogac/miniconda3/lib/python3.13", "/home/dogac/miniconda3/lib/python3.13/lib-dynload", "/home/dogac/miniconda3/lib/python3.13/site-packages"]
We should cancel request RequestId(I32(1))
We should cancel request RequestId(I32(2))
We should cancel request RequestId(I32(5))
We should cancel request RequestId(I32(10))
We should cancel request RequestId(I32(11))
We should cancel request RequestId(I32(9))
Request textDocument/documentSymbol (1) is canceled due to explicit cancellation
Request textDocument/codeAction (2) is canceled due to explicit cancellation
Request textDocument/inlayHint (3) is canceled due to subsequent mutation
Request textDocument/documentSymbol (4) is canceled due to subsequent mutation
Request textDocument/codeAction (5) is canceled due to explicit cancellation
Site package path: None
Finding Python interpreter...
Interpreters: /home/dogac/miniconda3/envs/mor/bin/python
Using Python interpreter from LSP: /home/dogac/miniconda3/envs/mor/bin/python
Site package path: None
Finding Python interpreter...
Interpreters: /home/dogac/miniconda3/envs/mor/bin/python
Using Python interpreter from LSP: /home/dogac/miniconda3/envs/mor/bin/python
Site package path: None
Finding Python interpreter...
Interpreters: /home/dogac/miniconda3/envs/mor/bin/python
Using Python interpreter from LSP: /home/dogac/miniconda3/envs/mor/bin/python
Request textDocument/documentSymbol (6) is canceled due to subsequent mutation
Request textDocument/semanticTokens/range (7) is canceled due to subsequent mutation
Request textDocument/semanticTokens/full (8) is canceled due to subsequent mutation
Request textDocument/documentHighlight (9) is canceled due to explicit cancellation
Request textDocument/codeAction (10) is canceled due to explicit cancellation
Request textDocument/codeAction (11) is canceled due to explicit cancellation
Request textDocument/codeAction (12) is canceled due to subsequent mutation
Site package path: None
Finding Python interpreter...
Interpreters: /home/dogac/miniconda3/envs/mor/bin/python
Using Python interpreter from LSP: /home/dogac/miniconda3/envs/mor/bin/python
Using interpreter site package path: ["/home/dogac/miniconda3/envs/mor/lib/python3.12", "/home/dogac/miniconda3/envs/mor/lib/python3.12/lib-dynload", "/home/dogac/miniconda3/envs/mor/lib/python3.12/site-packages"]
Using interpreter site package path: ["/home/dogac/miniconda3/envs/mor/lib/python3.12", "/home/dogac/miniconda3/envs/mor/lib/python3.12/lib-dynload", "/home/dogac/miniconda3/envs/mor/lib/python3.12/site-packages"]
Site package path: None
Finding Python interpreter...
Interpreters: /home/dogac/miniconda3/envs/mor/bin/python
Using Python interpreter from LSP: /home/dogac/miniconda3/envs/mor/bin/python
Using interpreter site package path: ["/home/dogac/miniconda3/envs/mor/lib/python3.12", "/home/dogac/miniconda3/envs/mor/lib/python3.12/lib-dynload", "/home/dogac/miniconda3/envs/mor/lib/python3.12/site-packages"]
Using interpreter site package path: ["/home/dogac/miniconda3/envs/mor/lib/python3.12", "/home/dogac/miniconda3/envs/mor/lib/python3.12/lib-dynload", "/home/dogac/miniconda3/envs/mor/lib/python3.12/site-packages"]
```

As you can see, it finds the right interpreter now because I have changed the precedence. However it still can't find it initially, probably a race-condition causes imports to be resolved using the initial interpreter configuration rather than the VSCode configured workspace configuration. I'm not sure why that race condition exists, I can keep debugging when I find free time.

Edit: I suspect the later interpreter not being able to fix import errors might be due to the stack overflow I have mentioned in the original issue. 

I have tested it in another project where pyrefly didn't crash, it succesfully updated the python interpreter to be the right one after some time. I think some improvements can be made regarding the initialization behavior, I will keep looking on why it happens. Maybe I will focus on the crash first.

cc: @connernilsen please let me know if there are any future steps I should take.

connernilsen:
I believe VSCode takes time to find interpreters initially, and sends them as a later request. Because of that, Pyrefly starts up and runs with whatever interpreter it can find before it's overridden by VSCode. c.c. @kinto0 does that sound right?

Dogacel:
> @Dogacel I think this looks good! I'll pull it in and get it merged soon

TYSM! Let me know if there is anything else I can do.

Issue #241, opened 2025-05-17
[cli] Separate lines of source code checked & lines of dependencies checked in summary


rosmur:
### Describe the Bug

When checking with `uvx pyrefly check`, the reported number of lines (eg: 180400) is far larger than number of lines in my source code. It is likely including in this number everything in the venv.

ndmitchell:
It is indeed reporting the entire venv, since it is checking everything in the venv. This makes sense in terms of how much work Pyrefly did, which is why we were initially displaying. I guess you were hoping to see how many lines are in your code?

rosmur:
> I guess you were hoping to see how many lines are in your code?

Yep correct

> I guess you were hoping to see how many lines are in your code?

Hmm, isn't it ignoring the venv by default? 

Also, the modules reported are correct (dont include the venv) but the lines aren't, thus there's an internal inconsistency (example attached).
![Image](https://github.com/user-attachments/assets/cdd13b97-a298-4dbe-bcb7-a764039d7e77)
Reporting the N lines of your project code that pyrefly went over is what the user would care about more?

grievejia:
Would it be better if we clarify the message of "XXX lines" to "XXX lines in transitive dependencies"? Or do you want us to report "XXX lines in my project" as well?

yangdanny97:
Maybe the format could be like:

`7 modules (XXX lines), 492 transitive dependencies (180,400 lines)`

rosmur:
yeah that looks good! I'd propose a minor modification for readability like: 

```
SOURCE CODE: x modules, y lines
TRANSITIVE DEPENDENCIES: x models, y lines
```

zriser:
Hello, new contributor here, I’d like to attempt this issue!

yangdanny97:
Assigned, thanks! Let us know if you have questions or need any assistance.

zriser:
Will do, thank you!

zriser:
I see the log line has changed a bit since the initial issue was created, and would like some feedback on the preferred verbiage before I begin prepping a PR. At the moment I have it generating:
` INFO 4 modules (269 dependent modules); 82,204 lines (17 lines in your project, 82,187 lines in dependencies); took 1.84s; memory (physical 173.7 MiB)`

PR #867, opened 2025-08-07
Separate lines of source code & lines of dependencies checked in summary

zriser:
Resolves facebook/pyrefly#241

Adds a new method split_line_count that updates the info line in the cli's full summary to include both user and dependency line counts.

Example: `INFO 4 modules (269 dependent modules); 82,204 lines (17 lines in your project, 82,187 lines in dependencies); took 1.98s; memory (physical 169.6 MiB)`

A warning is now thrown during build (warning: method `line_count` is never used) but didn't want to modify or remove this method to avoid breaking any external usage.

Let me know if you'd like any changes to the formatting, output wording, or implementation approach.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -717,7 +717,7 @@ impl CheckArgs {
             };
         }
         if self.output.summary == Summary::Full {
-            let user_handles: HashSet<Handle> = handles.iter().map(|(h, _)| h.clone()).collect();
+            let user_handles: HashSet<&Handle> = handles.iter().map(|(h, _)| h).collect();
             let (user_lines, dep_lines) = transaction.split_line_count(&user_handles);
             info!(
                 "{} ({}); {} ({} in your project, {} in dependencies); \
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -6,6 +6,7 @@
  */
 
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fmt;
 use std::fmt::Display;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


migeed-z:
That change LGTM. Although what's going to happen if we remove the method? (we can also always silence that specific warning with a header)

pyrefly/lib/commands/check.rs
```diff
@@ -716,14 +717,19 @@ impl CheckArgs {
             };
         }
         if self.output.summary == Summary::Full {
+            let user_handles: HashSet<Handle> = handles.iter().map(|(h, _)| h.clone()).collect();
```
rchen152: Nit: could you change this to be `HashSet<&Handle>` to avoid the clone? (You'll also have to update the usage and the type signature of `split_user_count`.)

pyrefly/lib/state/state.rs
```diff
@@ -539,6 +539,46 @@ impl<'a> Transaction<'a> {
         res
     }
 
+    /// Computes line count split between user-owned and dependency modules.
+    /// Returns (user_lines, dependency_lines).
+    pub fn split_line_count(&self, user_handles: &HashSet<Handle>) -> (usize, usize) {
```
rchen152: I'm fairly certain `line_count` has no external usages. Would you mind deleting it?

rchen152 (changes requested):
Thank you! This looks pretty good to me, back to you with a couple minor comments.

zriser:
@rchen152 will do!

zriser:
@rchen152 Requested changes submitted

rchen152:
Thank you! I'll import this and get it landed.

PR #870, opened 2025-08-07
Install scrut via cargo on CI workflows

fangyi-zhou:
The installation shell script seems to cause random failures when running CI workflow on macos-latest. Try to see if using cargo-installed version of scrut is more robust.


fangyi-zhou:
I see retries are added in #876, maybe we don't need this PR any more.

Issue #859, opened 2025-08-06
Support recognizing `Return: tuple[TypeA]` when handling hover documentation


michaelfortunato:
### Describe the Bug

Small thing. 
In the `Returns` part of the hover documentation, Pyrefly recognizes `Dataloader` as a type but doesn't highlight `tuple` properly. 
<img width="950" height="258" alt="Image" src="https://github.com/user-attachments/assets/4b30885c-1a62-47ef-8265-816a245a886f" />
Here is the function source. 
<img width="692" height="254" alt="Image" src="https://github.com/user-attachments/assets/549451fa-ccab-4222-9d47-004ab15819c9" />

yangdanny97:
Thanks for reporting, I didn't even know there were colors in the hover docs wow

I remember there being an issue with improper highlighting of special typing forms, but tuple is declared as a regular class in the stdlib stubs, so I'm not sure what's going on here

cc @kinto0

kinto0:
Thanks for the report! I think this is an editor-specific coloring. What is your vim setup?

michaelfortunato:
# Reproduction Steps
*Step 1*
  Create empty folder with venv activated that has Pyrefly.
```sh
mkdir foo
cd foo
uv venv
uv pip install pyrefly
```

*Step 2*
Create a file `repro.lua` with the contents
  ```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"--single-branch",
		"https://github.com/folke/lazy.nvim.git",
		lazypath,
	})
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
	-- do not remove the colorscheme!
	"folke/tokyonight.nvim",
	"saghen/blink.cmp",
	"neovim/nvim-lspconfig",
	-- add any other pugins here
}
require("lazy").setup(plugins, {
	root = root .. "/plugins",
})

vim.lsp.enable("pyrefly")

-- add anything else here
vim.opt.termguicolors = true
-- do not remove the colorscheme!
[...]

PR #877, opened 2025-08-08
escape brackets

kinto0:
Summary: #859 reports UI issues around brackets in docstrings. escaping them makes no difference in vscode and likely helps in other editors that do custom formatting.

Differential Revision: D79903803


michaelfortunato:
I think this PR needs to be rolled back. I do not think brackets should be escaped. it should be up to the lsp client to render the markdown or
Plaintext, depending hon how the lsp server (pyrefly) responds to the hover request. 

And see: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover

And then see: See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContent

PR #888, opened 2025-08-12
Fix comment and docstring typos

IDrokin117:
# Summary

This PR fixes **only** typos in comments and docstings. The [typos-cli](https://docs.rs/crate/typos-cli/latest) used for this purpose.

# Proposals

1. Should I correct typos in variables/methods/structs, etc.? Examples
```
commitable -> committable
Overriden ->  Overridden
further -> further
```
2.  Add a typos check to the CI pipeline. If this seems like a good idea to you, I can implement it.


ndmitchell:
This is great! To answer your questions:

1. Yes! Arguably these are more important in variable names.
2. Sounds good to me. Ideally there is a pre-canned action we can just slot in quite easily.

PR #902, opened 2025-08-13
Fix False-Positive when trying to access a `@property` of a generic parameter with union type bound

### LINKED ISSUE #869: False-Positive When Trying to Access a `@property` of a Generic Parameter with Union Type Bound
### Describe the Bug

Pyrefly version: 0.27.0

I expect the following code to be okay:
```python
from collections import defaultdict

class Reader:
    def __init__(self, source: str):
        self._source = source

    @property
    def source(self) -> str:
        return self._source

class Listener:
    def __init__(self, source: str):
        self._source = source

    @property
    def source(self) -> str:
        return self._source

def group_by_source[T: Reader | Listener](x: list[T]) -> dict[str, list[T]]:
    result: defaultdict[str, list[T]] = defaultdict(list)
    for item in x:
        result[item.source].append(item)
    return result
```
While both Pyright and MyPy don't find any issues in this code, Pyrefly complains:
```
ERROR Argument `T` is not assignable to parameter `self` with type `Self@Listener` in function `Listener.source` [bad-argument-type]
  --> /home/coder2k/dev/generics-test/main.py:34:16
   |
34 |         result[item.source].append(item)
   |                ^^^^^^^^^^^
   |
ERROR Argument `T` is not assignable to parameter `self` with type `Self@Reader` in function `Reader.source` [bad-argument-type]
  --> /home/coder2k/dev/generics-test/main.py:34:16
   |
34 |         result[item.source].append(item)
   |                ^^^^^^^^^^^
   |
 INFO 2 errors
```
However, when the `source` members in  `Reader` and `Listener` are implemented as regular instance attributes instead of `@property`s, Pyrefly is satisfied.

So, the following version *does* pass Pyrefly's check:
```python
from collections import defaultdict

class Reader:
    def __init__(self, source: str):
        self.source = source

class Listener:
    def __init__(self, source: str):
        self.source = source

def group_by_source[T: Reader | Listener](x: list[T]) -> dict[str, list[T]]:
    result: defaultdict[str, list[T]] = defaultdict(list)
    for item in x:
        result[item.source].append(item)
    return result
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBAxmA28CmMAuBLMA7AzldEADmCKlACZLACGArvKuemgFAszzU54BKS1lEAC4WUMRSpQA+lPRZ0qGQAocSeMAA0UHGFogYSIdtQgAlCPGXta4ADopOvQagBebbv1I2lgAKFwhEikAJ6i4pTA7k5IKjamUAC0AHzGwmFWUCBIqHpY1ur2jp5sHFx4ADLoOKhIWEEW4ZIycgrKqupaRQZG1WYNGe12Dh7Obl1e6VB+AUGooZYRUZ6x6vHJqf1WWTkgeYOFIxOLAObgtIRSAEbBw9EA2gAqRnwCQVAAPlCV1bVBALpKAAeRngVVQjz+axSTDQd16WlB1Qhf02WRw9FQRgidAYMPB8KgiPBDz+f1cEhoGLxSiJpkmwBI+Bq0DkUGBk0saIxdwUSAgtnGf1s1EIgSw5CUvIgdM52VymSQ6IYLCAA

_No response_

IDrokin117:
# Summary

Resolves #869

This fix ensures that when accessing properties on generic type parameters with union constraints (e.g., T: A | B), the type checker correctly resolves the property return type instead of treating it as the TypeVar itself.

## Explanation

To access TypeVar methods with union bounds, we must ensure that each union member type has the specified method. Currently, Pyrefly bound union types to self when accessing methods, rather than to concrete types. This fix should resolve that issue.

## Test Plan

- Added several test cases to ensure correctness of method calling on TypeVar with union bound and restrictions.

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -1496,7 +1496,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             },
             AttributeBase1::Quantified(q, bound) => {
                 if let Some(upper_bound) = bound {
-                    match self.get_bounded_quantified_attribute(q.clone(), upper_bound, attr_name) {
+                    let specific_q = match q.restriction() {
+                        Restriction::Constraints(_) => {
+                            q.clone().with_restriction(Restriction::Constraints(vec![
+                                upper_bound.clone().to_type(),
+                            ]))
+                        }
+                        Restriction::Bound(_) => q
+                            .clone()
+                            .with_restriction(Restriction::Bound(upper_bound.clone().to_type())),
+                        Restriction::Unrestricted => q.clone(),
+                    };
+                    match self.get_bounded_quantified_attribute(specific_q, upper_bound, attr_name)
+                    {
                         Some(attr) => acc.found(attr, base),
                         None => acc.not_found(NotFoundOn::ClassInstance(
                             upper_bound.class_object().dupe(),
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -658,15 +658,8 @@ fn bind_instance_attribute(
 ) -> Attribute {
     // Decorated objects are methods, so they can't be ClassVars
     if attr.is_property_getter() {
-        // For property getters on TypeVar instances, use the bound type instead of the TypeVar
-        let obj_type = match &instance.kind {
-            InstanceKind::TypeVar(_) => {
-                ClassType::new(instance.class.dupe(), instance.targs.clone()).to_type()
-            }
-            _ => instance.to_type(),
-        };
         Attribute::property(
-            make_bound_method(obj_type, attr).into_inner(),
+            make_bound_method(instance.to_type(), attr).into_inner(),
             None,
             instance.class.dupe(),
         )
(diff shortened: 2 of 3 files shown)


yangdanny97:
I don't think this is quite correct

If we instantiate based on the bound, then properties returning `Self` give an incorrect type.

```
from typing import Self
class A:
    @property
    def attr(self) -> Self: ...

class B:
    @property
    def attr(self) -> Self: ...

def foo[T: A | B](x: T) -> T:
    return x.attr # error: Returned type `A | B` is not assignable to declared return type `T`
```

That said, there are a few other spots in the codebase that have the same mistake, and are marked as such:

https://github.com/facebook/pyrefly/blob/f792282a4b3811072630fdcaf2cfca0b4c6f3e56/pyrefly/lib/alt/call.rs#L187

https://github.com/facebook/pyrefly/blob/f792282a4b3811072630fdcaf2cfca0b4c6f3e56/pyrefly/lib/alt/attr.rs#L1734

IDrokin117:
@yangdanny97 I've reworked the changes and updated the description. I've included your test case in the tests. Could you please review the changes? Let me know if I'm missing other nuances about TypeVar.

yangdanny97:
Thanks for making the changes @IDrokin117!

This overall looks good to me, the only thing I'd say is that we can probably move the specific_q stuff inside get_bounded_quantified_attribute.

 I'll apply that change and resolve the merge conflicts once the PR is imported.

yangdanny97 (approved):

PR #928, opened 2025-08-15
Implementing feature #593, warn on redundant cast

### LINKED ISSUE #593: [feature] warn on redundant cast

prasannavenkateshgit:
Summary
This PR adds detection for redundant usage of typing.cast where the value being cast already matches the target type. In such cases, a warning is now emitted to inform users of unnecessary casting.

Implementation Details
A check was added before applying the cast to compare the inferred type of the value against the target type.
If both types match, an ErrorKind::RedundantCast warning is emitted instead of applying the cast.
Includes tests under redundant_cast.rs.

Closes #593

### DIFF
--- a/pyrefly/lib/alt/special_calls.rs
+++ b/pyrefly/lib/alt/special_calls.rs
@@ -252,6 +252,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 "`typing.cast` missing required argument `val`".to_owned(),
             );
         }
+        if let Some(val_expr) = val {
+            let val_type = self.expr_infer(val_expr, errors);
+            if !val_type.is_any() && self.is_subset_eq(&val_type, &ret) {
+                self.error(
+                    errors,
+                    range,
+                    ErrorInfo::Kind(ErrorKind::RedundantCast),
+                    format!(
+                        "Redundant cast: `{}` is already assignable to type `{}`",
+                        val_type.deterministic_printing(),
+                        ret.clone().deterministic_printing()
+                    ),
+                );
+            }
+        }
         ret
     }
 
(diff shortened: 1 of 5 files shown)


migeed-z:
Hello! thanks for the PR. 

It seems some unittests are failing. Could you double check by running ./test.py?

conformance/third_party/conformance.result
```diff
@@ -196,7 +196,14 @@
     "Line 30: Expected 1 errors"
   ],
   "generics_paramspec_semantics.py": [],
-  "generics_paramspec_specialization.py": [],
+  "generics_paramspec_specialization.py": [
```
migeed-z: There are unexpected behaviors, so we should suspect there is a bug in the implementation. Specifically, there should be no warning here.

pyrefly/lib/alt/special_calls.rs
```diff
@@ -252,6 +252,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 "`typing.cast` missing required argument `val`".to_owned(),
             );
         }
+        if let Some(val_expr) = val {
+            let val_type = self.expr_infer(val_expr, errors);
+            if !val_type.is_any() && self.is_subset_eq(&val_type, &ret) {
```
migeed-z: can we change this check and error message to only emit a warning when the types are equal? not just assignable to each other?

conformance/third_party/conformance.result
prasannavenkateshgit: Updated tests

pyrefly/lib/alt/special_calls.rs
prasannavenkateshgit: Thanks for the comment, I have updated it to warn only when the types are exactly equal

PR #945, opened 2025-08-15
Implement the Type Server Protocol for Pyrefly

rchiodo:
# Type Server Protocol

This PR adds support for the [Type Server Protocol](https://github.com/microsoft/pylance-release/discussions/7180) to Pyrefly. That link describes why this might be useful for Pyrefly but I'll reiterate some of them here:

### Get pyrefly in front of more customers. 

Pylance is the default Python language server in VS code. (with about 10 million active users per month).

Pylance is changing such that it's splitting into two parts. A language server part and what's called a type server. The type server provides an API like a compiler and inspects types for Python source. 

This separation would allow Pylance to use different type servers to provide language services in VS code. That should make it possible for everybody using Pylance to also use Pyrefly as the underlying Type Server. With hopefully getting Pyrefly in front of more people.

### Not have to reimplement LS features

Pylance has some extra features over a base LS that are outlined in the link above. By supporting the TSP, Pyrefly can then get these features on top of its own LS.

## Explanation of changes

I've added a few things to Pyrefly:

### tsp_types crate (crates\tsp_types)

This new crate exports a protocol definition for the Type Server Protocol in rust. This should allow other rust based language serves also implement the Type Server Protocol if they wish. 

The protocol is generated using a python script that uses something called [lsprotocol](https://github.com/microsoft/lsprotocol) that already generates a protocol for rust and other languages for LSP. I modified the output a bit to work better with pyrefly.

The source for the generation is a tsp.json and tsp.schema.json that describe the TSP. I plan on copying this tsp.json and tsp.schema.json into Pylance after this.

### Changes to base Pyrefly 

I've added a couple things to the base pyrefly.

#### State 

The state module was modified to track a couple more things. 
- A `snapshot` that is updated whenever the state changes. TSP requires this indicate that the server is no longer matching the client. 
- A `version` (or diagnostics version) for every module. TSP uses this to determine if analysis of a file (like say for semantic tokens) matches the version that the type server thinks a file is.
- A registry of type and declaration handles that are used to identify internal types and data when communicating with the TSP client. Some of the TSP methods generate 'handles' for types that can then be passed back into a TSP server so that the TSP server can lookup the relevant information. This handle registry is being stored on the State object and is cleared when the snapshot changes (as that's the contract that TSP specifies).

#### LSP

Some of the parts of the lsp module were made public so I could reuse them in the TSP logic

#### Commands

A new command was added to the main application. The command is "tsp" and it starts a TspServer that sits and listens to LSP and TSP messages. TSP messages are basically and addition to LSP, so the TspServer delegates LSP messages to the LSP Server class.

### TSP Implementation

As described above TSP is implemented by delegating LSP messages to the original LSP Server class but then also has handlers in lib\tsp\requests for each of the TSP messages. These for the most part use a transaction to find a definition or a type and then convert the internal information into the TSP response.

Testing for TSP is in lib\tests\tsp. There are unit tests and interaction tests.

## Other testing and future plans

I tested with the 100 or so tests I added but also by using Pylance. The latest version of Pylance will work with this Pyrefly by setting these options in the settings.json:

```jsonc
  "python.analysis.typeServerExecutable": "c:\\Users\\rchiodo\\source\\repos\\pyrefly\\lsp\\bin\\pyrefly.exe",
  "python.analysis.typeServerArguments": [
    "tsp"
  ],
  "python.analysis.enableAsyncProgram": true
```

Granted everything doesn't work. I added a TODO.md file that explains some of the problems I found. Pylance should ignore these for the most part (you just don't get any results).

### Future Plans

I have a couple of ideas:

1. I hope to maintain this code here in Pyrefly if this change is accepted. As the TSP grows, I'll update Pyrefly to continue to implement it.
2. I hope once this goes through to add a nightly test to Pylance that continually verifies that Pylance can work with Pyrefly. It would pull the official build of Pyrefly and test it against the latest Pylance.

## Apologies :)

I'm sorry this PR is so big. I could potentially just add the pieces in different steps. I thought maybe if this is too big it could be split up into:

1. Just adding the tsp_types crate
4. Add the TspServer and command with the handlers
5. Add each TSP handler one at a time.

Please let me know if you'd like it split instead. And thanks for your input and feedback on this.


maggiemoss:
@kinto0 might have the most context here

kinto0:
Thanks for submitting this! I think you mentioned splitting it up on discord - I'll wait to review until that happens unless you want me to take a look first. 

An order I came up with that mirrors how we would submit it internally:
- formatting (if you meant to do this, there's a lot of json being formatted with no changes)
- tsp protocol crate itself
- add empty tsp server to pyrefly with command 
- each request as one pr with the request and any tests associated

rchiodo:
> Thanks for submitting this! I think you mentioned splitting it up on discord - I'll wait to review until that happens unless you want me to take a look first.
> 
> An order I came up with that mirrors how we would submit it internally:
> 
> * formatting (if you meant to do this, there's a lot of json being formatted with no changes)
> * tsp protocol crate itself
> * add empty tsp server to pyrefly with command
> * each request as one pr with the request and any tests associated

Okay that's how I'll split it up. The json files should have had changes in them (I added a build task and some new launch targets, but I can remove the auto formatting changes).

rchen152:
(Added Kyle as reviewer just so our automation stops complaining about this PR needing triage.)

rchiodo:
I should close this one anyway. I'm working on splitting it into smaller parts.

Issue #675, opened 2025-07-15
Feature request: Add codeactions (ignore error)


michaelfortunato:
It would be great if pyrefly can add [codeactions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) to, for example, annotate a piece of flagged code with `# pyrefly: ignore[error-code]`. 

Thanks!

PR #956, opened 2025-08-18
feat(code-action): add ignore comment to suppress error

michaelfortunato:
This PR address #675.

It works by
1. By gathering all the errors whose range contains the current cursor.
2. Finding the column position of the first non-whitespace character of the line the cursor is on. Let us call this `column offset`.
3. Inserting an ignore comment on the line above the cursor; the comment is preceded by `column offset` single-spaces to match the prior indentation.

In my tests I have found this simple scheme to be surprisingly robust. My first attempt used the ast exclusively, but the ast does not contain comments, nor a way to insert _comments_ into the tree, meaning, just for this one use case of programmatically adding ignore comments, I cannot update the document using the ast, and I found myself having to calculate whitespace offsets in my first ast only attempt.

It is also worth pointing out that some LSPs might not _want_ to provide this kind of code action--they want to encourage the user to address the underlying cause, not ignore it. But because Pyrefly is new, there are many false positives and so I think this will be a welcome addition. Plus, the suppression comments in this code action reference specific error codes, so the user in the future can address the lints precisely.

Future work is needed to address different types of ignore comments and a code action to

- Add all ignore comments to a module
- Add a top-level ignore comment to a module

Both should be relatively straightforward to do by using the utility functions I added in `pyrefly/lib/state/ide.rs`

Let me know your thoughts on this.

Thanks!

### DIFF
--- a/crates/pyrefly_python/src/ignore.rs
+++ b/crates/pyrefly_python/src/ignore.rs
@@ -283,15 +283,24 @@ impl Ignore {
         false
     }
 
-    /// Get all pyrefly ignores.
+    #[inline]
+    /// Get pyrefly ignores that are global, and do not have an error code.
+    pub fn get_pyrefly_all_ignores(&self) -> SmallSet<LineNumber> {
+        self._get_pyrefly_ignores(true)
+    }
+
+    #[inline]
     pub fn get_pyrefly_ignores(&self) -> SmallSet<LineNumber> {
+        self._get_pyrefly_ignores(false)
+    }
+    fn _get_pyrefly_ignores(&self, only_all: bool) -> SmallSet<LineNumber> {
         self.ignores
             .iter()
             .filter(|ignore| {
                 ignore
                     .1
                     .iter()
-                    .any(|s| s.tool == Tool::Pyrefly && s.kind.is_empty())
+                    .any(|s| s.tool == Tool::Pyrefly && (!only_all || s.kind.is_empty()))
             })
             .map(|(line, _)| *line)
             .collect()
@@ -306,6 +315,8 @@ mod tests {
 
     #[test]
     fn test_parse_ignores() {
+        // TODO: The second component of `expect` (the error code)
+        // is not actually tested.
         fn f(x: &str, expect: &[(Tool, u32)]) {
             assert_eq!(
                 &Ignore::parse_ignores(x)
@@ -325,6 +336,7 @@ mod tests {
             "code # mypy: ignore\n# pyre-fixme\nmore code",
             &[(Tool::Mypy, 1), (Tool::Pyre, 3)],
         );
+        f("# pyrefly: ignore[bad-]\n", &[(Tool::Pyrefly, 2)]);
         f(
             "# type: ignore\n# mypy: ignore\n# bad\n\ncode",
             &[(Tool::Any, 4), (Tool::Mypy, 4)],
--- a/pyrefly/lib/test/lsp/code_actions.rs
+++ b/pyrefly/lib/test/lsp/code_actions.rs
@@ -25,12 +25,44 @@ fn apply_patch(info: &ModuleInfo, range: TextRange, patch: String) -> (String, S
     (before, after)
 }
 
-fn get_test_report(state: &State, handle: &Handle, position: TextSize) -> String {
+fn get_test_report_insert_import(state: &State, handle: &Handle, position: TextSize) -> String {
     let mut report = "Code Actions Results:\n".to_owned();
     let transaction = state.transaction();
     for (title, info, range, patch) in transaction
         .local_quickfix_code_actions(handle, TextRange::new(position, position))
-        .unwrap_or_default()
+        .into_iter()
+        .flatten()
+        // NOTE: We do NOT patch buffer with ignore comment code action that is
+        // tested separately
+        .filter(|(title, _, _, _)| !title.contains("ignore"))
+    {
+        let (before, after) = apply_patch(&info, range, patch);
+        report.push_str("# Title: ");
+        report.push_str(&title);
+        report.push('\n');
+        report.push_str("\n## Before:\n");
+        report.push_str(&before);
+        report.push_str("\n## After:\n");
+        report.push_str(&after);
+        report.push('\n');
+    }
+    report
+}
+
+fn get_test_report_ignore_code_action(
+    state: &State,
+    handle: &Handle,
+    position: TextSize,
+) -> String {
+    let mut report = "Code Actions Results:\n".to_owned();
+    let transaction = state.transaction();
+    for (title, info, range, patch) in transaction
+        .local_quickfix_code_actions(handle, TextRange::new(position, position))
+        .into_iter()
+        .flatten()
+        // NOTE: We ONLY patch buffer with ignore comment code action so
+        // we can test it separately
+        .filter(|(title, _, _, _)| title.contains("ignore"))
     {
         let (before, after) = apply_patch(&info, range, patch);
         report.push_str("# Title: ");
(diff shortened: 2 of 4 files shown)


michaelfortunato:
FYI Ruff's wording for their equivalent action is 
`Ruff (F100): Disable for this line`

connernilsen:
Thanks for doing this @michaelfortunato! @kinto0 thoughts?

crates/pyrefly_python/src/ignore.rs
```diff
@@ -283,15 +283,24 @@ impl Ignore {
...
+    /// Get pyrefly ignores that are global, and do not have an error code.
+    pub fn get_pyrefly_all_ignores(&self) -> SmallSet<LineNumber> {
+        self._get_pyrefly_ignores(true)
+    }
+
+    #[inline]
     pub fn get_pyrefly_ignores(&self) -> SmallSet<LineNumber> {
+        self._get_pyrefly_ignores(false)
+    }
+    fn _get_pyrefly_ignores(&self, only_all: bool) -> SmallSet<LineNumber> {
```
kinto0, line 296: this argument name is kinda confusing: could we name it something else?

and could you add a comment for why we ned to include ones that are non-empty in the call?

crates/pyrefly_python/src/ignore.rs
```diff
@@ -306,6 +315,8 @@ mod tests {
 
     #[test]
     fn test_parse_ignores() {
+        // TODO: The second component of `expect` (the error code)
```
kinto0, line 318: could you add a test for the difference in only_all true vs false? (I'm still confused as to why it's necessary)

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -45,6 +49,128 @@ fn get_test_report(state: &State, handle: &Handle, position: TextSize) -> String
...
+        report.push_str(&before);
+        report.push_str("\n## After:\n");
+        report.push_str(&after);
+        report.push('\n');
+    }
+    report
+}
+
+#[test]
+fn add_ignore_comment() {
```
kinto0, line 81: could this test instead be multiple smaller tests that each test something different? I don't know if you need multiple files either for any of these

- add_ignore_comment_basic
- add_ignore_comment_multiline
- add_ignore_indentation (what happens when the indentation changes?)
- add_to_existing_ignore_basic
- add_to_existing_ignore_multiline
- add_to_existing_ignore_indentation (what happens here when indentation is different?)

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -25,12 +25,16 @@ fn apply_patch(info: &ModuleInfo, range: TextRange, patch: String) -> (String, S
     (before, after)
 }
 
-fn get_test_report(state: &State, handle: &Handle, position: TextSize) -> String {
+fn get_test_report_insert_import(state: &State, handle: &Handle, position: TextSize) -> String {
```
kinto0, line 28: why do we need insert_import? if we need it for all of these below, could you add a comment clarifying why?

kinto0 (changes requested):
thanks for the contribution! this looks great - a few suggestions

kinto0:
> In my tests I have found this simple scheme to be surprisingly robust. My first attempt used the ast exclusively, but the ast does not contain comments, nor a way to insert comments into the tree, meaning, just for this one use case of programmatically adding ignore comments, I cannot update the document using the ast, and I found myself having to calculate whitespace offsets in my first ast only attempt.

I wonder if we have any way to get access to the CST... @SamChou19815 did you use this at all for the auto import work or do you have any suggestions?

SamChou19815:
> > In my tests I have found this simple scheme to be surprisingly robust. My first attempt used the ast exclusively, but the ast does not contain comments, nor a way to insert comments into the tree, meaning, just for this one use case of programmatically adding ignore comments, I cannot update the document using the ast, and I found myself having to calculate whitespace offsets in my first ast only attempt.
> 
> I wonder if we have any way to get access to the CST... @SamChou19815 did you use this at all for the auto import work or do you have any suggestions?

Auto-import does use AST, but only used it to figure out where to insert the import.

SamChou19815:
Also note that we do have error suppression infra: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/error/suppress.rs

so we should try to reuse the code over there.

michaelfortunato:
Hey everyone thanks for taking the time to look and comment on this. I'll address the comments individually and update the pr when I can.

michaelfortunato:
> Also note that we do have error suppression infra: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/error/suppress.rs
> 
> so we should try to reuse the code over there.

Yes, I had looked into that. The issue is the error suppression infra works on files themselves at the moment--so its server side. In contrast, this code action sends the required range and content for the lsp client (IDE) to update. We should probably update the existing suppression system to handle this use case.

michaelfortunato:
> > In my tests I have found this simple scheme to be surprisingly robust. My first attempt used the ast exclusively, but the ast does not contain comments, nor a way to insert comments into the tree, meaning, just for this one use case of programmatically adding ignore comments, I cannot update the document using the ast, and I found myself having to calculate whitespace offsets in my first ast only attempt.
> 
> I wonder if we have any way to get access to the CST... @SamChou19815 did you use this at all for the auto import work or do you have any suggestions?

So originally I had used `pyrefly_python::ast::Ast::locate_node` to get the node at the bottom of the parse tree. I then iterated through the list moving up until I found the first statement node, and used that to get the offset and line number. Problem is I found statement to be too large a container--the ignore comment was being applied too far away from the location of the error itself.
```rust
  use pyrefly_python::ast::Ast;
  let nodes = Ast::locate_node(module, position);
  let node = nodes.iter().find(|node| node.is_statement()).unwrap();
  let loc = node.range();
  // etc...
```
Plus, since each `error` in `let errors = self.get_errors(vec![handle]).collect_errors().shown;` gives contains its location, I figured it probably best to use that. 
Then comes the question of given the precise error location (again, we have that for free with `let errors = self.get_errors(vec![handle]).collect_errors().shown;`), where do we insert the comment? If we insert the comment right at at this point, we can get some issues. 

For instance
```python 
X = 4 + Y
#      ^ Error point is here
```
After insertion (assume we append a carriage return to the comment) becomes
```python
X = 4 + # pyrefly: ignore[code]
Y
```
Which is not ideal. Hence, instead of inserting the comment right at the location of the error, I found it better to _insert_ it a line above. I emphasize _insert_ as the key point is the comment will be on its own line, so its less likely it'll interfere with surrounding code. 

Of course doing this requires us knowing what level of indentation we are at, so at best we case insert the comment right at that level or beyond it so as not to cause a parse error (and let an autoformatter take care of aligning).  

Thanks

pyrefly/lib/test/lsp/code_actions.rs
michaelfortunato: The `get_test_report` function in upstream right now applies _all_ code actions returned by `local_quickfix_code_actions` to the module in question. So for future code actions and the code action in this PR, it would be hard to predict the form of the resulting module if all possible code actions were applied (ignore comment, add import). So I figured its better to have a dedicated report function for each possible code action type. But really it might be best to paramterized get_test_report with a filter function and have the caller determine which code actions it wants applied on the module referenced by the given `Handle`.
 
Upshot is I do think get_test_report needs updating to allow us to selectively apply code actions to the associated module.

pyrefly/lib/test/lsp/code_actions.rs
michaelfortunato: Certainly!

samwgoldman:
Maybe we can learn from how Ruff adds noqa comments to the end of lines? Especially since we are using their parser and their AST.

michaelfortunato:
> Maybe we can learn from how Ruff adds noqa comments to the end of lines? Especially since we are using their parser and their AST.

I personally think that we should add a noqa comment **on its own line above the line of the lowest (leafiest) syntax node responsible for the error (Pyrefly gives us this node already)**. That way we prevent the case where a noqa comment is at the end of a line, which spans >80 (eg) columns, and an autoformatter, upon enforcing the line length constraint, moves the noqa comment to a place it is no longer effective--usually the line _below_ the _statement_ node which contains the error. This has happened to me a bunch. Hopefully that makes sense.

rchen152:
Closing this pull request since it's gone stale. @michaelfortunato please feel free to open an new pull request if you're still interested in working on this.

PR #960, opened 2025-08-18
Type Server Protocol tsp_types crate creation

rchiodo:
As descibed in https://github.com/facebook/pyrefly/pull/945, this PR adds a new crate for the types in the [Type Server Protocol](https://github.com/microsoft/pylance-release/discussions/7180)

This crate will be used in future PRs to receive and send TSP messages.

The protocol.rs file is generated by running the `crates\tsp_types\protocol_generator\generate_protocol.py`. It injests the tsp.json file to generate the protocol.

Testing is done with unit tests for each request.


.vscode/tasks.json
```diff
@@ -111,6 +111,34 @@
         }
       }
     },
+    {
+      "type": "shell",
+      "command": "cargo watch -x 'build --all-features --artifact-dir ../lsp/bin -Z unstable-options'",
+      "group": "build",
+      "isBackground": true,
+      "label": "cargo: watch debug pyrefly for extension",
```
kinto0, line 119: thanks for adding this! I usually develop by building the LS somewhere and reloading the window every time I make a change. but this will make it much easier.

crates/tsp_types/protocol_generator/generate_protocol.py
```diff
@@ -0,0 +1,510 @@
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "generator @ git+https://github.com/microsoft/lsprotocol.git",
+# ]
+# ///
+"""
```
kinto0: hmm... I don't love the implicit dependency. Opening this project and setting it up with cargo will not include lsprotocol and make onboarding an extra hurdle. 
- does it have to be python? can we write this part in rust instead?
- can we vendor that codebase? it doesn't seem that big
- or I can try adding it in buck when I import this and seeing if it can generate a cargo rule that imports it in cargo?

crates/tsp_types/protocol_generator/generate_protocol.py
```diff
@@ -0,0 +1,510 @@
...
+def convert_json_to_model(tsp_json: Dict[str, Any]) -> model.LSPModel:
+    """
+    Convert our TSP JSON format to the lsprotocol internal model format.
+    
+    This function maps our JSON structure to the Python dataclasses used
+    by the lsprotocol generator.
+    """
+    
+    # Extract metadata
+    metadata = tsp_json["metaData"]
```
kinto0, line 46: I wish https://github.com/microsoft/lsprotocol didn't capitalize the d in `metadata` (it's one word)...

crates/tsp_types/protocol_generator/generate_protocol.py
kinto0: side question: how often will we need to regenerate this protocol? will it only be you? should / could we have red CI if the protocol has updated but not regenerated?

rchiodo: I'm not sure what buck is but running `uv run <this python file>` will automatically download the dependency and run it. 

It's part of https://peps.python.org/pep-0723/, the ability to specify dependencies. 

Rewriting this in rust seems like it would be easier to just submit the output and drop the python file. It's reusing the logic in the generator package and that would have to be reimplemented. 

I'd likely be the only one to rerun this. Or anybody adding new methods to pyrefly for tsp. When the tsp.json file changes, the generator needs to be rerun.

rchiodo: I think vendoring the code base is possible. Let me know if you'd rather do that.

kinto0: I actually think it's ok. we can suppress any CI failures

PR #965, opened 2025-08-20
Add a command to run a TSP server

rchiodo:
Continuing the split of https://github.com/facebook/pyrefly/pull/945, this PR adds a new command to Pyrefly for running a [Type Server Protocol](https://github.com/microsoft/pylance-release/discussions/7180) server. 

This PR doesn't handle any extra messages, those will come one at a time in subsequent PRs.

### DIFF
--- a/pyrefly/lib/lsp/server.rs
+++ b/pyrefly/lib/lsp/server.rs
@@ -190,8 +190,8 @@ use crate::state::semantic_tokens::SemanticTokensLegends;
 use crate::state::state::State;
 use crate::state::state::Transaction;
 
-/// Interface exposed for outside modules to interact with the LSP server
-pub trait LspServerInterface {
+/// Interface exposed for TSP to interact with the LSP server
+pub trait TspInterface {
     /// Send a response back to the LSP client
     fn send_response(&self, response: Response);
 
@@ -203,28 +203,6 @@ pub trait LspServerInterface {
         subsequent_mutation: bool,
         event: LspEvent,
     ) -> anyhow::Result<ProcessEvent>;
-}
-
-/// Factory for creating LSP server instances that implement LspServerInterface
-pub struct LspServerFactory;
-
-impl LspServerFactory {
-    /// Create a new LSP server instance
-    pub fn create(
-        connection: Arc<Connection>,
-        lsp_queue: LspQueue,
-        initialization_params: InitializeParams,
-        indexing_mode: IndexingMode,
-        workspace_indexing_limit: usize,
-    ) -> Box<dyn LspServerInterface> {
-        Box::new(Server::new(
-            connection,
-            lsp_queue,
-            initialization_params,
-            indexing_mode,
-            workspace_indexing_limit,
-        ))
-    }
 }
 
 #[derive(Clone, Dupe)]
@@ -258,7 +236,7 @@ impl ServerConnection {
     }
 }
 
-struct Server {
+pub struct Server {
     connection: ServerConnection,
     /// A thread pool of size one for heavy read operations on the State
     async_state_read_threads: ThreadPool,
@@ -845,7 +823,7 @@ impl Server {
         Ok(ProcessEvent::Continue)
     }
 
-    fn new(
+    pub fn new(
         connection: Arc<Connection>,
         lsp_queue: LspQueue,
         initialize_params: InitializeParams,
@@ -1936,7 +1914,7 @@ impl Server {
     }
 }
 
-impl LspServerInterface for Server {
+impl TspInterface for Server {
     fn send_response(&self, response: Response) {
         self.send_response(response)
     }
--- a/pyrefly/lib/tsp/server.rs
+++ b/pyrefly/lib/tsp/server.rs
@@ -18,34 +18,20 @@ use lsp_types::ServerCapabilities;
 use crate::commands::lsp::IndexingMode;
 use crate::lsp::queue::LspEvent;
 use crate::lsp::queue::LspQueue;
-use crate::lsp::server::LspServerFactory;
-use crate::lsp::server::LspServerInterface;
 use crate::lsp::server::ProcessEvent;
+use crate::lsp::server::TspInterface;
 use crate::lsp::server::capabilities;
 use crate::lsp::server::dispatch_lsp_events;
 use crate::lsp::transaction_manager::TransactionManager;
 
 /// TSP server that delegates to LSP server infrastructure while handling only TSP requests
-pub struct TspServer<T: LspServerInterface> {
-    pub inner: T,
+pub struct TspServer {
+    pub inner: Box<dyn TspInterface>,
 }
 
 impl TspServer {
-    pub fn new(
-        connection: Arc<Connection>,
-        lsp_queue: LspQueue,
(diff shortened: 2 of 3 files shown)


pyrefly/lib/tsp/server.rs
```diff
@@ -0,0 +1,131 @@
...
+            event,
+        )
+    }
+
+    fn handle_tsp_request<'a>(
+        &'a self,
+        _ide_transaction_manager: &mut TransactionManager<'a>,
+        _request: &Request,
+    ) -> anyhow::Result<bool> {
+        // TODO: Add handling for TSP requests
```
rchiodo, line 72: This is where I'll add new handlers in subsequent PRs

pyrefly/lib/lsp/server.rs
```diff
@@ -219,19 +219,19 @@ impl ServerConnection {
     }
 }
 
-struct Server {
```
kinto0, line 224: do you think it's possible to make a new struct for the generic server things that are shared between tsp / lsp? then we use composition to reuse that generic stuff in both tsp / lsp servers?

I think that'd be much preferred than making all this state public.

rchiodo: Maybe? Have a common Server shared by both? Then make an LspServer and a TspServer?

That common Server would be public but the Lsp specific stuff wouldn't? I'm not sure that would work though for everything. Well because LspServer handles doc updates, but TspServer needs those updates for the diagnostics version and the snapshot tracking. 

So the snapshot has to be updated in the LspServer, but looked at in the TspServer.

kinto0: > but TspServer needs those updates for the diagnostics version and the snapshot tracking

diagnostics version / snapshot tracking is inherently a TSP concept, right? lsp server shouldn't need to keep track of this

kinto0: I think doc updates _should_ be a tsp-server-only thing if possible

rchiodo: I wonder if in the TspServer I can listen to the same messages but not mark them as handled? That might make it possible to update the snapshot without having to add code to the lsp handlers but still let the lsp server handle the document requests.

The problem is it's tied right now into the Server::invalidate method. When the transaction is updated, the snapshot is incremented. Maybe there's a way to add an event for that instead.

But that's not in this PR. That's when I add the get_snapshot handler. 

This PR is only making public the things to allow request handling. Meaning the public parts are necessary for the delegation

pyrefly/lib/lsp/server.rs
```diff
@@ -221,16 +221,16 @@ impl ServerConnection {
...
     connection: ServerConnection,
     /// A thread pool of size one for heavy read operations on the State
     async_state_read_threads: ThreadPool,
     lsp_queue: LspQueue,
     initialize_params: InitializeParams,
     indexing_mode: IndexingMode,
     workspace_indexing_limit: usize,
-    state: Arc<State>,
-    open_files: Arc<RwLock<HashMap<PathBuf, Arc<String>>>>,
+    pub state: Arc<State>,
```
rchiodo: Do you think it would be better to add functions to access the necessary information?

rchiodo: This data members aren't actually accessed just yet, but will be in future request handlers.

pyrefly/lib/lsp/server.rs
```diff
@@ -255,7 +292,7 @@ struct Server {
 /// - priority_events includes those that should be handled as soon as possible (e.g. know that a
 ///   request is cancelled)
 /// - queued_events includes most of the other events.
-fn dispatch_lsp_events(connection: &Connection, lsp_queue: LspQueue) {
+pub fn dispatch_lsp_events(connection: &Connection, lsp_queue: LspQueue) {
```
kinto0, line 273: at some point I'd like this to be encapsulated in some better API that lets tsp_loop and lsp_loop reuse logic but it doesn't have to be right now

pyrefly/lib/lsp/server.rs
```diff
@@ -1898,3 +1935,24 @@ impl Server {
         self.invalidate(|t| t.invalidate_config());
     }
 }
+
+impl LspServerInterface for Server {
```
kinto0: I would even call this something specific to TSP since we don't really want anything else to use this interface yet. the benefit of this interface vs `pub` methods is we know / control the fanout of these APIs. I'd like to keep this exposure as minimal as possible.

pyrefly/lib/tsp/server.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::lsp::server::LspServerInterface;
+use crate::lsp::server::ProcessEvent;
+use crate::lsp::server::capabilities;
+use crate::lsp::server::dispatch_lsp_events;
+use crate::lsp::transaction_manager::TransactionManager;
+
+/// TSP server that delegates to LSP server infrastructure while handling only TSP requests
+pub struct TspServer {
+    pub inner: Box<dyn LspServerInterface>,
+}
```
kinto0: I think you can do this to avoid the heap allocation:
```suggestion
pub struct TspServer<T: LspServerInterface> {
    pub inner: T,
}
```

pyrefly/lib/lsp/server.rs
```diff
@@ -190,6 +190,43 @@ use crate::state::semantic_tokens::SemanticTokensLegends;
...
+        Box::new(Server::new(
+            connection,
+            lsp_queue,
+            initialization_params,
+            indexing_mode,
+            workspace_indexing_limit,
+        ))
+    }
+}
+
```
kinto0: I think you can skip the box if you skip stack allocation below and likely don't need the factory

edit: I meant heap allocation

kinto0 (commented):
Thanks for making the change! I think the heap allocation can be avoided making this interface make more sense. Let me know if that doesn't work

pyrefly/lib/lsp/server.rs
rchiodo: Should I rename it to TspInterface?

rchiodo:
@kinto0 is this what you meant? I'm not sure how I'd create an LSP server for use in the TSP server without making it public or making a factory that returns the interface.

rchiodo:
I have another idea that might be what you mean. If I refactor the lsp_loop into a common portion which allocates the lsp_server, I can pass this lsp_server out of the common portion and use it in the tsp_loop to create the TspServer without having the TspServer allocate the LSP server. This common portion would then be the public part.

kinto0:
> @kinto0 is this what you meant? I'm not sure how I'd create an LSP server for use in the TSP server without making it public or making a factory that returns the interface.

This is what I meant, the public constructor is necessary but I think worthwhile since it avoids needing to make a factory

kinto0:
> I have another idea that might be what you mean. If I refactor the lsp_loop into a common portion which allocates the lsp_server, I can pass this lsp_server out of the common portion and use it in the tsp_loop to create the TspServer without having the TspServer allocate the LSP server. This common portion would then be the public part.

I think it's really good for now. I don't want to slow you down too much -this extra stuff can happen later if at all.

PR #1000, opened 2025-08-26
Create lambda scope with barrier false

### LINKED ISSUE #959: Lambda scoping issue
### Describe the Bug

```

from typing import *

# Taken from pytorch example:
#   if isinstance(new, torch.Tensor):
        # assert isinstance(old, torch.Tensor)
        # torch._check(
        #     old.dim() == new.dim(), lambda: f"{old.shape} != {new.shape} (old != new)"
        # )


_T = TypeVar("_T")

class A:
    shape: int = 0


def check(new: _T, old: _T) -> None:
    if isinstance(new, A):
        assert isinstance(old, A)
        lambda: f"{old.shape}" # Object of class `object` has no attribute `shape` [missing-attribute]
```

the `old` variable in the nested if statments is typed as `_T`

### Sandbox Link

https://pyrefly.org/sandbox/?code=FAMwTg9gtgBALgTwA4EsB2BzGKpImOGAKmGAGIYAVAQwGsBTNGcaGJBOfAYwAsZ6AHtVwAbegC5yMaShDYAzunlxqaLvQAUaegHcANPG48AdJUbz8ASknTbtitXnz6BBUpVrNEEQBMDnMF5TcytgOzsKAKCAfV56LloNMPDpCjtvH2MfHA1LGABefJhtHSycywMRYQAjH2pxZgAiAG8M43keaiR6AF8YAEIi5pL2zu6%2BjQyBopLLRuSUiktSYGjKAqpkegA1ajANRrXG5eAuKqcYAEEbWw6uiWw0QiKABhWfejk4hK1dBrWDBl-pQ8gBaAB8MAAchBtDcZHIUIo0MpVOpfvortYFuFHM5XEj3GivL4DJdlilpFUoLV6k1Wr5Rvceo0YBQAPLVABW8UIEC%2B53kMAABhBubzhTBOkK0BAYNQ4HAwChqgBXOD0EV3bqSgDaUCRyIwoIVSpV6voAF1gEA&version=3.12

_No response_

IDrokin117:
# Summary
Resolves #959

Lambda expressions were incorrectly using the same scope creation method as regular functions (Scope::function), which sets barrier = true. This was semantically incorrect because:

 1. Lambdas cannot assign variables - Python syntax prohibits assignments in lambda bodies
 2. No UnboundLocalError risk - The barrier mechanism protects against reading variables before local assignment, but lambdas can't create local assignments

## Test plan

I've added a simple test case `test_lambda_captures_narrowed_variable` to ensure that lambda body has access to outer scope flow (not only static info).

### DIFF
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -325,7 +325,7 @@ impl<'a> BindingsBuilder<'a> {
     }
 
     pub fn bind_lambda(&mut self, lambda: &mut ExprLambda, usage: &mut Usage) {
-        self.scopes.push(Scope::function(lambda.range, false));
+        self.scopes.push(Scope::lambda(lambda.range, false));
         if let Some(parameters) = &lambda.parameters {
             for x in parameters {
                 self.bind_lambda_param(x.name());
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -582,6 +582,16 @@ impl Scope {
             }),
         )
     }
+    pub fn lambda(range: TextRange, is_async: bool) -> Self {
+        Self::new(
+            range,
+            false,
+            ScopeKind::Function(ScopeFunction {
+                yields_and_returns: Default::default(),
+                is_async,
+            }),
+        )
+    }
 
     pub fn method(range: TextRange, name: Identifier, is_async: bool) -> Self {
         Self::new(
--- a/pyrefly/lib/test/scope.rs
+++ b/pyrefly/lib/test/scope.rs
@@ -612,3 +612,17 @@ def try_except_else_finally():
     e7 # E: `e7` is uninitialized
 "#,
 );
+
+// https://github.com/facebook/pyrefly/issues/959
+testcase!(
+    test_lambda_captures_narrowed_variable,
+    r#"
+class A:
+    shape: int = 0
+
+def check[T](new: T, old: T) -> None:
+    if isinstance(new, A):
+        assert isinstance(old, A)
+        lambda: old.shape
+"#,
+);


rchen152 (approved):
This looks great to me, thank you! I'll work on getting it merged.

stroxler:
I think this is a reasonable change and will make Pyrefly better, but I do want to note that it is potentially controversial.

The `barrier = true` scope logic is intended for any situation where we can capture non-local variables (globals and nonlocals - by this I mean any capture, not necessarily mutable captures using the `global` and `nonlocal` keywords) in a deferred execution environment, which primarily means lambda and function bodies.

The reason is because the lifetime of the lambda may well be longer than the current flow. Consider:
```
x = 5
f = lambda: x
x = "foo"
z = f()
```

The actual type of `z` at runtime will be `str`, but a type checker can't know that unless we inline the analysis (which we probably wouldn't do because that could scale very badly).

The correct type for `f` in a conservative system is probably `Callable[[], int | str]`, which is exactly what `barrier = true` gives us: the barrier causes us to use `Anywhere` bindings for `x`, and the `Anywhere` node is a Phi of int and string.

If we did a more complex syntactical analysis, there would be some cases of lambdas (and nested function definitions) where we could conservatively do better - for example, in
```
def foo(x: int | str):
    if isinstance(x, int):
        f = lambda: x
        z = f()
    else:
        z = 5
```
we could in principle note that no reference to `f` has a lifetime longer than the `if` block.

I believe @samwgoldman was suggesting we implement something like this, maybe with a reaching-definition analysis. My initial instinct is that it might be somewhat hard to do that, and even if we did it soundly it would probably produce a lot of `Anywhere` bindings anyway because most uses are going to pass the lambda to some other function, and we have no way of guaranteeing that the lifetime of the lambda is short in that case (it could get assigned to a global, or a class attribute, or anything else). In particular, I don't think even a very clever sound analysis would work for simple cases like
```
def foo(x: int | str):
    if isinstance(x, int):
         # this is *probably* okay, but we don't know enough about lifetimes to really prove it using local information
         takes_callback_returning_int(lambda: x)
 ```

Because I think the example above probably needs to be considered okay, I'm pretty sympathetic to the idea that we should indeed just set `barrier = false` for lambdas, and also for nested functions (I can't see a good argument for why one and not the other - their scope behavior is pretty similar).

So my vote is to merge this, but in doing so to make an intentional decision that we are going to trade some soundness for better usability in the many scenarios that aren't easily statically analyzable

**Comment by samwgoldman** [comment]
@stroxler is correct that capturing the flow-sensitive type in a lambda is unsafe in general, because the lambda can execute at a different time. Pyrefly is already intentionally unsound w.r.t. narrowing which could potentially be invalidated, so I think it's reasonable to consider this behavior an extension of that existing unsoundness.

That is to say, I don't mind being unsound here. Making this sound and precise is too complex. Making this sound and imprecise (current behavior) could easily be too restrictive. Unfortunately, this is a place where mypy and pyright disagree:

```
x = 5
f = lambda: x
x = "foo"
z = f()
reveal_type(z) # mypy: int, pyright: int | str
```

(Note that pyright is being conservative here, and also gives `int | str` when `x = "foo"` comes after `z = f()`, similar to Pyrefly before this change.)

PR #1007, opened 2025-08-27
Prevent kwargs `Unpacked[TypedDict]` fields overlapping with pos arg

### LINKED ISSUE #595: [feature] don't allow Unpacked TypedDict kwargs to overlap with any named parameter
### Describe the Bug

Refer to the linked sandbox for an example.

I think the place to change is https://github.com/facebook/pyrefly/blob/c247d452349c10958a44a3842540b12f7634e220/pyrefly/lib/alt/function.rs#L374

We should check if kwargs is an unpacked TypedDict, and if so, make sure it does not overlap with any existing named params.

When completing this task, please add an integration test with the sandbox example to pyrefly/lib/tests/typed_dict.rs, run test.py, and commit any generated changes.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQMYA2AhgM7VQDSA7pSGtQBQAqiApgCYARJKRgBKAFxEo0qAA9xmFDBIBiWAAskdLVADy9KACMepSgFdqPDVeBIQ1fLMx0c1JDCRgUlcgFov5HBEfDzAsDwO7PKKMAA0UAD08QQEANbMrNQKAKooCJSkqQDaTCxsALoSUAB0taoaOtTqYGbkfFAoYPjGevTBoeGR0ahxhGkZbDl5BcWlmZUKtdVAA

_No response_

IDrokin117:
# Summary
Resolves #595

Implemented check to prevent `Unpack[TypedDict]` field names from overlapping with positional parameter names

# Test Plan

- Added test case `test_typed_dict_kwargs_overlap_pos_arg` based on the original issue sandbox example


yangdanny97:
I would expect a conformance change from this - did running test.py generate any changes?

IDrokin117:
@yangdanny97 After running test.py, it added prefix path `conformance\\third_party\\` to file names. It doesn't feel right, but I have no idea how to fix it. I just run `python test.py` from root dir. What should I do?

yangdanny97:
Oh, that's weird. Are you on a windows machine? For now the workaround would be a manual find & replace, but it's not a big deal since the files get regenerated automatically when a merge conflict happens.

In any case, things are merged now. Thanks for your work!

IDrokin117:
@yangdanny97 Great! Yeah, I am on a windows machine. Good to know, thanks.

PR #1010, opened 2025-08-28
Update TSP to version 0.2.0

rchiodo:
This PR updates TSP to match the newest internal version that Pylance is using. 

It eliminates the following requests:

- `SearchForTypeAttributes`
- `GetTypeAttributes`
- `GetSymbolsForFile`
- `GetSymbol`
- `GetFunctionParts`

It eliminates the `Attribute` class and replaces it with the already existing `Symbol` idea.

It adds two new requests

- `GetSymbolsForType` that can find the symbols for a type.
- `GetSymbolsForNode` that can find the symbols scoped by a node.

These two new requests handle the situations required by the 4 of the original functions. `GetFunctionParts` was eliminated entirely as it was printing out parameters and such and could be reimplemented with just the `GetRepr` request.

This PR is a prerequisite for adding the handlers for these requests.

### DIFF
--- a/crates/tsp_types/protocol_generator/generate_protocol.py
+++ b/crates/tsp_types/protocol_generator/generate_protocol.py
@@ -535,14 +535,7 @@ def extract_flag_enums_from_tsp(tsp_json: Dict[str, Any]) -> Dict[str, Dict[str,
 
         if "Flags" in enum_name:
             is_flag_enum = True
-        # Heuristic: most non-zero values are powers of 2
-
-        elif (
-            total_values > 1 and power_of_two_count >= (total_values - zero_count) * 0.7
-        ):
-            is_flag_enum = True
         # Special case: all values are powers of 2 or 0
-
         elif total_values > 1 and (power_of_two_count + zero_count) == total_values:
             is_flag_enum = True
         if is_flag_enum:


crates/tsp_types/protocol_generator/generate_protocol.py
```diff
@@ -472,6 +490,66 @@ def find_block_end(idx: int) -> Optional[int]:
     return content[:doc_start] + replacement + content[de_end:]
 
 
+def extract_flag_enums_from_tsp(tsp_json: Dict[str, Any]) -> Dict[str, Dict[str, int]]:
```
rchiodo, line 493: I added this because the original generator had hardcoded a bunch of flags to fixup. This searches for them instead. Should work better if flags change in the future.

crates/tsp_types/protocol_generator/tsp.json
```diff
@@ -987,54 +838,52 @@
...
-                            "name": "Declaration"
-                        }
+                        "kind": "reference",
+                        "name": "Type"
                     },
-                    "documentation": "The declarations for the symbol. This can include multiple declarations for the same symbol."
+                    "documentation": "The type of the symbol found."
                 },
                 {
-                    "name": "synthesizedTypes",
```
rchiodo, line 1001: Got rid of this idea in Symbols as it was very Pyright specific.

crates/tsp_types/protocol_generator/generate_protocol.py
kinto0: I think I'm missing something but why do some flags allow values that aren't powers of two? can we guarantee that all bitflags are powers of two?

rchiodo: Well we can't guarantee, but I agree the heuristic is weird. We can't guarantee it because there's nothing in the json that can enforce AFAIK.

rchiodo: I'm just going to remove the heuristic, the final output from generation is the same anyway.

kinto0: I'm wondering if we need to check for flags in the name too. Not urgent though, we can always make changes to it as we go. I can import it now

kinto0 (approved):

stroxler (approved):

Issue #376, opened 2025-05-27
[Feature] warn on @deprecated


yangdanny97:
### Describe the Bug

https://typing.python.org/en/latest/spec/directives.html#deprecated

warn on the following:

1. import deprecated class/function
2. call deprecated function

For now, this can be a regular type error since we don't support different error levels yet. It should have its own error code(s) and be disabled by default to avoid causing false positives. Once we support warnings, we'll turn it on.

Alex-Aron:
Hey, I am happy to do 1 or 2, which one are you looking into doing?

yangdanny97:
@Alex-Aron I haven't started working on either of them so you're free to take your pick. I've split this into sub-issues. Please let us know which one you'd like to look at and if you have any questions :)

Alex-Aron:
I can work on #402 . What files would you recommend looking at first? I was looking into modifying lib/binding/function.rs and binding.rs first. Just want to make sure I am looking in the right direction. Thanks :)

yangdanny97:
@Alex-Aron 

Function decoration info & other metadata is stored in this FuncFlags struct: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/types/callable.rs#L172

It would be a good starting point to see how that's populated & add a new is_deprecated field, then check the field and raise an error whenever it's called.

There's a few places that implicitly call functions under the hood and there's some nuances for overloads and such, but I was thinking maybe we can just start with the simple case where it's straightforward call expression: https://github.com/facebook/pyrefly/blob/bcc4ac7e7be208dd5234ee59b06a024115431c8f/pyrefly/lib/alt/expr.rs#L1388

Alex-Aron:
Hey! I got an error up if `@deprecated` is above a function. Let me know how the pull request looks + any other general criticisms so I can improve future open source contributions.

fangyi-zhou:
All the warnings should now be implemented. We might want to consider whether to show the deprecation message in the warning as the next step, or maybe just call it done.

yangdanny97:
Thanks for your work on this! I think if it's easy to show the message we could add it, but otherwise I'm happy to consider this done until someone specifically requests it

fangyi-zhou:
Are we good to close this issue?

PR #1011, opened 2025-08-29
[RFC] Warn if deprecated function is referred

fangyi-zhou:
This diff implements the last missing feature for deprecation warnings (#376): to warn when the deprecated function is referred (e.g. passed as function reference).

The implementation checks when typechecking an expression, whether the type of an expression is a deprecated function/bound method. If that's the case, a deprecated warning is generated.

If we only add this check, we might run into the issue that a deprecated function call would generate 2 deprecation warnings, one when the function is typechecked, and another when the call is typechecked.

To avoid the duplication, we avoid emitting the deprecation warning when typechecking a function call, instead we emit the deprecation warning when generating the call target.

As a consequence, the error message of deprecation warning is changed from "Calling" a deprecated function to "Reference to" a deprecated function. Suggestions are welcome for how to improve clarify of this error message.


yangdanny97:
Sorry for the delay here,

Maybe `{function name} is deprecated` for simplicity?

Mypy and Pyright's error messages, for reference: 

https://github.com/python/typing/blob/main/conformance/results/pyright/directives_deprecated.toml

https://github.com/python/typing/blob/main/conformance/results/mypy/directives_deprecated.toml

The rest of it LGTM

yangdanny97:
@fangyi-zhou  I'll patch the error message in the imported version

PR #1023, opened 2025-09-01
merge same type

### LINKED ISSUE #996: Merge function types for display when they have the same type signature
### Describe the Bug

```py
def f1(x: int) -> int: ...
def f2(x: int) -> int: ...
def f3(x: int) -> int: ...

DICT = {
    1: f1,
    2: f2,
    3: f3
}
```

I get

```py
(constant) DICT: dict[int, ((x: int) -> int) | ((x: int) -> int) | ((x: int) -> int)]
```

but seems they are the same (at least from the string form), I want to see

```py
(constant) DICT: dict[int, ((x: int) -> int)]
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #996


yangdanny97:
awesome, thanks @asukaminato0721 !

PR #1024, opened 2025-09-01
Handle scoping correctly for class fields in body

### LINKED ISSUE #264: Class definition bindings should not be visible in nested scopes
### Describe the Bug

Minimal repro:
```python
x = 1

class A:
    x = 4
    [reveal_type(x) for a in range(1)]

    class B:
        reveal_type(x)
```
Expected: Both reveals produce `Literal[2]`
Actual: Both reveals produce`Literal[1]`

### Sandbox Link

https://pyrefly.org/sandbox/?code=B4AgvCCMBQ0MYBsCGBnFICCAuaI8lAgBZd8BtAJwFMA3KpBAfQBcBPAByoApgBKEAGYB7CiCQgAlgDsQFJFIDm3SLwC6sfCESp0AIRybN1OgxYdufaEA

_No response_

fangyi-zhou:
Fixes #264.

This diff addresses the scoping issues of class fields in nested scopes.

When a class field is defined in a class body, the scope of the field should only be available in the immediate class block, unless for annotations. This is described in the Python doc at https://docs.python.org/3/reference/executionmodel.html#resolution-of-names

To address this issue, we check the flow style when looking up a variable. If the variable is defined as a class field, we apply appropriate checks to avoid the incorrect scoping rule from being applied.

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -788,9 +788,24 @@ impl<'a> BindingsBuilder<'a> {
     ) -> Result<(Idx<Key>, Option<Idx<Key>>), LookupError> {
         let mut barrier = false;
         let ok_no_usage = |idx| Ok((idx, None));
+        let is_static_type_information = matches!(usage, Usage::StaticTypeInformation);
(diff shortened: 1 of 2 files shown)


migeed-z (approved):

PR #1025, opened 2025-09-01
Fix some clippy issues

fangyi-zhou:
Ran `cargo clippy --fix`


maggiemoss (approved):

PR #1029, opened 2025-09-02
Add handler for the TSP "typeServer/getSupportedProtocol" request

rchiodo:
This is the next stage in adding support for [Type Server Protocol](https://github.com/microsoft/pylance-release/discussions/7180) to Pyrefly

This is a very simple request that just returns the version number supported. It also includes the infrastructure for wiring up this request and testing it.


pyrefly/lib/lsp/server.rs
```diff
@@ -195,6 +195,9 @@ pub trait TspInterface {
     /// Send a response back to the LSP client
     fn send_response(&self, response: Response);
 
+    /// Get access to the state for creating transactions
+    fn state(&self) -> &Arc<State>;
```
rchiodo, line 222: This is used so I can do this in a request handler:

```rust
                let transaction =
                    ide_transaction_manager.non_committable_transaction(self.inner.state());
```

pyrefly/lib/test/tsp/tsp_interaction/util.rs
```diff
@@ -0,0 +1,285 @@
...
+use crate::commands::lsp::IndexingMode;
+use crate::commands::tsp::TspArgs;
+use crate::commands::tsp::run_tsp;
+// Re-export necessary types and functions for TSP tests
+pub use crate::test::lsp::lsp_interaction::util::TestCase;
+use crate::test::lsp::lsp_interaction::util::get_initialize_messages;
+use crate::test::util::init_test;
+
+/// Simple TSP test runner that follows the same pattern as LSP tests but uses TSP server
+pub fn run_test_tsp(test_case: TestCase) {
```
kinto0: if you've seen any of the newer tests, I actually wrote them in a much much easier way. unfortunately I haven't had time to get rid of run_test_lsp, but it is marked deprecated. you should look at any of the tests that use the [object model](https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/lsp/lsp_interaction/object_model.rs). I actually think you'll save a lot of time writing them like that - sorry you already wrote them this way.

pyrefly/lib/test/lsp/lsp_interaction/util.rs
```diff
@@ -281,7 +281,7 @@ fn get_initialize_messages(
     ]
 }
 
-fn get_initialize_responses(find_refs: bool) -> Vec<Message> {
+pub fn get_initialize_responses(find_refs: bool) -> Vec<Message> {
```
kinto0: I think with the redo of lsp_interaction yuou won't have to expose this

kinto0 (changes requested):
looks great. just some suggestions around testing. i'm sorry you implemented the test this way!! I think the other way will be much easier long-term. if you think it's too annoying, though, we can do it later

pyrefly/lib/test/tsp/tsp_interaction/util.rs
rchiodo: Thanks, I'll have a look. Most of the test code was generated by Claude anyway, so it shouldn't take long to switch.

rchiodo: Okay should be following the same pattern as the new lsp_interaction object_model test framework

pyrefly/lib/test/tsp/tsp_interaction/get_supported_protocol_version.rs
```diff
@@ -0,0 +1,90 @@
...
+fn test_tsp_get_supported_protocol_version_interaction_empty_params() {
+    // Test protocol version retrieval with empty object params
+    let temp_dir = TempDir::new().unwrap();
+    let test_file_path = temp_dir.path().join("version_test_2.py");
+
+    let test_content = r#"# Another test file
+x = 42
+"#;
+
+    std::fs::write(&test_file_path, test_content).unwrap();
```
kinto0, line 67: for the more complicated tests, I might suggest putting these in version control entirely. for now, you don't need to since this method does not require contents

kinto0 (approved):

rchiodo:
@kinto0 is there some way I can run the linter and tests that are internal?

grievejia (approved):

Issue #985, opened 2025-08-22
Feature Request: support multiple files in the sandbox


rchen152:
### Describe the Bug

It would be really nice if the sandbox allowed creating multiple files. This would make it possible to create reproducers of import bugs and to add fake stubs to demonstrate things involving third-party libraries.

hashiranhar:
Does this still need working on?

yangdanny97:
Nobody has picked it up yet, are you interested in working on it?

hashiranhar:
> Nobody has picked it up yet, are you interested in working on it?

Hey! Yeah I'd be interested in working on this. Any context and constraints I need to be aware of would be helpful.

Also any preferences on the UX? Was thinking maybe a file tree sidebar or tabs, but lmk if there's an existing design system I need to follow.

yangdanny97:
> Also any preferences on the UX? Was thinking maybe a file tree sidebar or tabs, but lmk if there's an existing design system I need to follow.

No preferences on the UX, though I think I prefer tabs over sidebar since it takes up less screen space.

I know Ty's sandbox has tabs and it's implemented in WASM, so perhaps that's the closest reference I can think of.

rchen152:
I believe this is done now. Thank you, @hashiranhar!

PR #1037, opened 2025-09-04
Feature:  Multiple files now supported in the sandbox #985

hashiranhar:
## Addresses
Issue #985 - Enables creating multiple files in the sandbox, making it possible to create reproducers of import bugs and add fake stubs to demonstrate things involving third-party libraries.

## Summary
- Implements  multi-file Python support with tab-based file management
- Enables cross-file imports  in Type Checking and runtime execution
- Adds real-time type checking across all files in sandbox session
- Users can create, edit, rename, and delete files.

#### Core Architecture
- Extended Playground struct with multi-file state management (handles: HashMap<String, Handle>)
- Implemented PlaygroundSourceDatabase for proper module resolution and import handling (custom_module_paths deprecated)
- Added incremental update optimization using Require::Everything vs Require::Exports strategy

#### WASM Interface
- Added new WASM methods: updateSandboxFiles, updateSingleFile, setActiveFile
- Extended TypeScript interfaces with multi-file support methods
- Enhanced usePythonWorker with runMultiFilePython for runtime execution

#### Frontend Multi-File UI
- Built tab-based file management system with Monaco editor model switching
- Implemented file operations: create, rename, delete with proper validation
- Added real-time tab synchronization with backend file state

#### Python Runtime Execution
- Integrated Pyodide filesystem (FS.writeFile) for multi-file module support
- Enhanced Python worker protocol to handle both single-file and multi-file payloads
- Added sys.path configuration and importlib.invalidate_caches() for module discovery

## Testing
- ✅ `test_cross_file_import()` - Verifies imports work between files
- ✅ `test_multi_file_errors_with_filenames()` - Confirms error attribution to correct files  
- ✅ `test_incremental_update_with_cross_file_errors()` - Validates optimized single-file updates
- ✅ Manual testing of file management UI, tab switching, and Python execution with imports

## Checklist
- [x] Code builds and passes all tests
- [x] Maintains backward compatibility with single-file mode
- [x] WASM bindings updated with proper TypeScript interfaces
- [x] Default multi-file setup demonstrates working imports
- [x] Performance optimized with incremental updates solutiontics to specific files across all tabs


migeed-z:
That's a great feature!

yangdanny97:
Thanks for doing this, given it's a fairly large PR be aware that it might take a while to fully review.

pyrefly/lib/playground.rs
```diff
@@ -177,31 +223,131 @@ impl Playground {
...
-                    message_details: e.msg_details().unwrap_or("").to_owned(),
-                    kind: e.error_kind().to_name().to_owned(),
-                    // Severity values defined here: https://microsoft.github.io/monaco-editor/typedoc/enums/MarkerSeverity.html
-                    severity: match e.error_kind().default_severity() {
-                        Severity::Error => 8,
-                        Severity::Warn => 4,
-                        Severity::Info => 2,
-                        Severity::Ignore => 1, // Ignored errors shouldn't be in `CollectedErrors.shown`
-                    },
+            .run_with_committing_transaction(transaction, &handle_reqs);
```
yangdanny97: need to rebase, we changed the shape of this recently.

Instead of handle_reqs being a list of pairs of handle + requires, we now pass a list of handles, and a single requires that applies to all of them.

pyrefly/lib/playground.rs
```diff
@@ -131,6 +164,9 @@ pub struct InlayHint {
 pub struct Playground {
     state: State,
     handle: Handle,
+    handles: HashMap<String, Handle>,
```
yangdanny97, line 166: why do we have handle & handles? 

I thought we could get rid of the old single handle and just store everything in the map.

pyrefly/lib/playground.rs
```diff
@@ -177,31 +223,131 @@ impl Playground {
...
+        config.source_db = Some(ArcId::new(Box::new(source_db) as Box<dyn SourceDatabase>));
+
+        config.configure();
+        let config = ArcId::new(config);
+        let new_config_finder = ConfigFinder::new_constant(config.dupe());
+
+        self.state = State::new(new_config_finder);
+        self.config_finder = ConfigFinder::new_constant(config.dupe());
+
+        if let Some(sandbox_handle) = self.handles.get("sandbox.py") {
```
yangdanny97: why is this sandbox.py but it's test.py elsewhere?

I suggest we refactor the magic strings to a single constant somewhere so we don't mix things up.

pyrefly/lib/playground.rs
hashiranhar: The handle points to the active file in the playground.

In the current flow the most methods will continue to use self.handle except get_errors which iterates over all files in the map ( want all the typing checking errors to show up in console at all times )

I chose this approach because it allows most methods to remain unchanged and continue using the legacy self.handle . I also wasn't completely sure if the playground pointed to something else aside from the sandbox hence wanted legacy methods intact.

We could refactor the code so every method could extract the active file as needed inside rather than separate storage.

yangdanny97: understood. In that case could we name the fields in a way that makes this more obvious, and add some explanatory comments in the definition of the playground struct? Something like `active_handle`.

We could store the active file as a name and just look up from the map as needed, but there would need to be an invariant that they're always kept in sync when the active file is deleted - but if that makes the code more complicated then we can save that change for another time.

Playground.rs is only used for the web playground, so no need to worry about affecting anything else.

pyrefly/lib/playground.rs
hashiranhar: Legacy method where there was only one python file had the file saved as test.py. Again similar answer to the prev query. Wherever you see test.py is just the old method preserved in case the playground points to something other than the sandbox.

We could remove the legacy single file methods if needed

yangdanny97: Yeah, let's clean those up to reduce confusion.

pyrefly/lib/playground.rs
hashiranhar: Are you perhaps suggesting having a str instead of a handle and modifying the set_active_file so that methods search up the file that set_active_file scraped? 

I was thinking of this^

yangdanny97: Yes, to avoid duplicating the handle.

hashiranhar:
> Review automatically exported from Phabricator review in Meta.

Thanks for reviewing this!!

There seemed to be a small conflict that was exported, I have sorted it out.

PR #1053, opened 2025-09-07
Fix class override check with contravariant Self

### LINKED ISSUE #1052: Broken return type list[Self] while Self is working (regression from 0.31.0 -> 0.31.1)
### Describe the Bug

```python
from typing import Self


class Vehicle:
    def generateOne(self, uniqueId: int) -> Self:
        return Vehicle()

    def generateSome(self, uniqueId: int) -> list[Self]:
        return [Vehicle(), Vehicle()]


class Car(Vehicle):
    def generateOne(self, uniqueId: int) -> Self:
        return Car()

    def generateSome(self, uniqueId: int) -> list[Self]:
        return [Car()]
```

The last `generateSome` will produce the following error in 0.31.1:

```
`Car.generateSome` has type `BoundMethod[Car, (self: Car, uniqueId: int) -> list[Car]]`, which is not assignable to `BoundMethod[Vehicle, (self: Vehicle, uniqueId: int) -> list[Vehicle]]`, the type of `Vehicle.generateSome`
```

`Self` is handled correctly, but `list[Self]` seems to be resolved into `list[Vehicle]` and not kept as `list[Self]` in the inherited class.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDKApgDbABQlAxiQIYDO9UAakQBZI1EBc5U-UACZFgUNERREQtGEQDykgBT1SwADRQAriiQBHTUQCSg7phQwAlFAC0APkKreA51BBEYmkChbtOJIooWlM7CouKS0rIEkAEqZBraegbGpqiWNvYkSPQwANrEZAC6Ti78bh5eULmsHFyBGjV%2BARaFlOQ0DEwAwrQgio1cFiX8oWISUjLySnHqWjr6RiZm6XYOZMMu5Z7ePX1BfAKj4RNRMcqqCfPJS2lWq1k5%2BarFB6Vblbm7ga1AA&version=3.12

_No response_

samwgoldman:
Summary:
When we check class consistency, we use the same technique as protocol matching: we substitute Self using the derived class.

Fixes https://github.com/facebook/pyrefly/issues/1052

Differential Revision: D81885918

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1809,8 +1809,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 // keys but not regular fields.
                 continue;
             }
-            let want_attribute =
-                self.as_instance_attribute(&want_class_field, &Instance::of_class(parent));
+            // Substitute `Self` with derived class to support contravariant occurrences of `Self`
+            let want_attribute = self.as_instance_attribute(
+                &want_class_field,
+                &Instance::of_protocol(parent, self.instantiate(cls)),
+            );
             if got_attribute.is_none() {
                 // Optimisation: Only compute the `got_attr` once, and only if we actually need it.
                 got_attribute = Some(self.as_instance_attribute(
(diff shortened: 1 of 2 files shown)


stroxler (approved):

Issue #1061, opened 2025-09-09
Pyrefly Sandbox multi-file go-to-def support


connernilsen:
Attempting a go-to-def across files goes to random location at the top of the same file in the Pyrefly Sandbox web editor

hashiranhar:
I’ve made a PR on this for reference I think it didn’t get flagged properly. 
I was planning to do another PR for GitHub gists implementation. Thought it would be useful to merge this before I make the next PR to avoid conflicts

kinto0:
> I’ve made a PR on this for reference I think it didn’t get flagged properly. I was planning to do another PR for GitHub gists implementation. Thought it would be useful to merge this before I make the next PR to avoid conflicts

so sorry this went under the radar!! thank you very much for your contribution. I'm reviewing ASAP.

connernilsen:
Hey @hashiranhar, just added you as the owner of this task for now. If you no longer want to do this, or if you have other questions, please let us know here or in the Discord server. Thanks!

kinto0:
> Hey [@hashiranhar](https://github.com/hashiranhar), just added you as the owner of this task for now. If you no longer want to do this, or if you have other questions, please let us know here or in the Discord server. Thanks!

it's already code-complete [here](https://github.com/facebook/pyrefly/pull/1070), merging now

PR #1070, opened 2025-09-10
Fix cross-file go-to-definition navigation in Pyrefly Sandbox - Issue #1061

hashiranhar:
### Problem
 Issue: #1061  Go-to-definition across files in the web editor was jumping to random locations in the current file
  instead of navigating to the correct location in the target file.

  ### Solution
  Enhanced the go-to-definition system to return target file information:

  1. **Backend**: Updated `goto_definition()` to return `DefinitionResult` struct with both range and
  target filename
  2. **Frontend**: Modified Monaco definition provider to use correct target file URI instead of always
  using current file URI
  3. **Navigation**: Added automatic file switching when definition is in a different file

  ### Key Changes
  - `playground.rs`: New `DefinitionResult` struct with filename field
  - `configured-monaco.ts`: Fixed definition provider to use target file URI
  - `Sandbox.tsx`: Added cross-file navigation support
  - Added test `test_cross_file_goto_definition` for cross-file goto-definition functionality

  ### Testing
  - ✅ Cross-file navigation now works correctly
  - ✅ Same-file navigation unchanged
  - ✅ Added test coverage for new functionality


kinto0 (approved):
Thank you so much!! 

This change is great. Awesome test coverage.

stroxler (approved):

pyrefly/lib/playground.rs
```diff
@@ -522,6 +541,40 @@ mod tests {
...
+            result.is_some(),
+            "Should find definition for cross-file function call"
+        );
+        let def_result = result.unwrap();
+        assert_eq!(
+            def_result.filename, "utils.py",
+            "Definition should be in utils.py"
+        );
+        assert_eq!(
+            def_result.range.start_line, 1,
```
kinto0: please ensure the test case has different ranges in each file. this will avoid bugs where we end up at the range from the file clicked, not the destination file

kinto0 (changes requested):
manual testing reveals a few issues:
- hovering the function and pressing cmd ends up performing the go-to def (this isn't the case on the prod sandbox so I wonder if it was introduced here?)
- navigating across files does not seem to end up on the correct line

would you be able to do more manual testing to ensure this feature works correctly? sorry I didn't catch these earlier, I am not too familiar with the sandbox

hashiranhar:
> manual testing reveals a few issues:
> 
>     * hovering the function and pressing cmd ends up performing the go-to def (this isn't the case on the prod sandbox so I wonder if it was introduced here?)
> 
>     * navigating across files does not seem to end up on the correct line
> 
> 
> would you be able to do more manual testing to ensure this feature works correctly? sorry I didn't catch these earlier, I am not too familiar with the sandbox

1. This is standard monaco editor behaviour I believe, did not code anything on my side, just that now the goto definition works hence so does the shortcut for it
2. Let me check this, my current logic just lands you on the file where the definition is stored I believe. Its a bit short sighted but I was just assuming at that time people would be using small imports to check out a sandbox. I can refine this tho

kinto0 (changes requested):
Sorry about the delayed review!! I was at a conference and this dropped off my radar. 

> This is standard monaco editor behaviour I believe, did not code anything on my side, just that now the goto definition works hence so does the shortcut for it

VSCode has two apis: "get definition" and "go to definition". Get definition happens on cmd+hover in order to preview the location, but we only want to "go" on click. I imagine monaco is similar, and we have hooked into the "get" instead of the "go to". Do you think you could attempt this?

> Let me check this, my current logic just lands you on the file where the definition is stored I believe. Its a bit short sighted but I was just assuming at that time people would be using small imports to check out a sandbox. I can refine this tho

I think it's necessary to bring them to the line/column if that would work. otherwise, finding the import is just as good since the user likely only has a few tabs open in the sandbox

hashiranhar:
> Sorry about the delayed review!! I was at a conference and this dropped off my radar.
> 
> > This is standard monaco editor behaviour I believe, did not code anything on my side, just that now the goto definition works hence so does the shortcut for it
> 
> VSCode has two apis: "get definition" and "go to definition". Get definition happens on cmd+hover in order to preview the location, but we only want to "go" on click. I imagine monaco is similar, and we have hooked into the "get" instead of the "go to". Do you think you could attempt this?
> 
> > Let me check this, my current logic just lands you on the file where the definition is stored I believe. Its a bit short sighted but I was just assuming at that time people would be using small imports to check out a sandbox. I can refine this tho
> 
> I think it's necessary to bring them to the line/column if that would work. otherwise, finding the import is just as good since the user likely only has a few tabs open in the sandbox

I will have a look at the first one, second one is up and running

hashiranhar:
@kinto0 

Hope you are doing well! I tried to solve the ctrl+hover bug but I have not found anything so far. 
I believe the monaco editor has one api that does both go and get (unlike vscode), when you attach this manually its not able to recognise the cross files by default and does that random jump thing ( the current problem ). I might be wrong tho maybe you could explore on your side. 

I've resolved the conflicts coming from upstream so the solution should still work if you import the request. Let me know what would be the ideal next steps. 

Apologies!

kinto0:
> @kinto0
> 
> Hope you are doing well! I tried to solve the ctrl+hover bug but I have not found anything so far. I believe the monaco editor has one api that does both go and get (unlike vscode), when you attach this manually its not able to recognise the cross files by default and does that random jump thing ( the current problem ). I might be wrong tho maybe you could explore on your side.
> 
> I've resolved the conflicts coming from upstream so the solution should still work if you import the request. Let me know what would be the ideal next steps.
> 
> Apologies!

hey, thanks for all the work. I will leave this PR open until someone has time to figure out this last step. We're very close!!

NathanTempest:
Hi @hashiranhar closing this PR as it has not seen activity in quite a while, but since you already covered a lot of road, we will be willing to reopen if you are willing to walk the last mile?

Issue #1074, opened 2025-09-10
Could not find name ... - when using a constant as a default value in a class method


kilroy42:
### Describe the Bug

```python
class Object:
    BOOK = 1
    CHAIR = 2
    HOUSE = 3

    def __init__(self, typ: int = BOOK):
        self.typ = typ

print(Object().typ)
```

This is correct Python code, but it fails with the following error in pyrefly:

`ERROR 6:35-39: Could not find name BOOK [[unknown-name](https://pyrefly.org/en/docs/error-kinds/#unknown-name)]`

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQDy2AVjJQC6IA669-9AEKNGAaXoBeegEZeA%2BgGEAEgEEAkgCVJ9AExyBSxgFUAygFFtAZl77%2BmGGHoB9JxHQQOLgBRwYUMAA09BykxIj0bhzawmIAlDx88gK%2B-oQhxNrpNui8xABOkV7MbJxesWmhsbwgASAArhzQcCTkiCAAxPRGjVAepPRgdeicELjocNn2jmC4eQC2qJ7odXPYMHle%2BOGRsfQAtAB89HAceQnyeTAcdXl8YNwgAHIra2f0wPgAvg-Vn7WoIwAbjAAGLQGAUNBYPBEMggT5AA&version=3.12

yangdanny97:
this is related to #1073, I think both should be solved together.

stroxler:
Oh this is now fixed! :)

Issue #1073, opened 2025-09-10
"Could not find name" when referencing class level variable from another class level var


lost-theory:
### Describe the Bug

This code:

```python

class Foo:
    COLORS = ["red", "green", "blue"]
    COLORS_UPPER = [c.upper() for c in COLORS]

print(Foo.COLORS_UPPER)
```

Worked in pyrefly-0.31.0 but started failing in 0.31.1 with:

```bash
$ ~/sand2-env/bin/pyrefly check --python-interpreter ~/sand2-env/bin/python x.py
ERROR Could not find name `COLORS` [unknown-name]
 --> ./x.py:4:40
  |
4 |     COLORS_UPPER = [c.upper() for c in COLORS]
  |                                        ^^^^^^
  |
 INFO 1 error
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuiAOuvV-QMIDyAGT4AlAMr0AvPQDabEACcYmOQBp6cgOaKY6VepDYoAVxhyAuh269BI0QH0AqgAUnAUWGSZlQkeLEY8gAUAJT0YLjy9JT0EJz8QmIWuujE8rEALoHMuITxto4u7sEgKiBG6dBwJOSIIADE9A4VUBDppGFG6JQVuOhwHByYMGBhEQC2qOl26EZj2AGB%2BIgx6OmhALQAfPRw6fLsnNyK6UbynGByAHKz8-v0wPgAvnIcII%2BlqN0QAG4wjNAwChoLB4IhkN5AA&version=3.12

yangdanny97:
Thanks for reporting this regression, we'll try to have a fix soon.

PR #1079, opened 2025-09-11
Handle scoping correctly for class fields in body (reland)

fangyi-zhou:
This is a reland of PR #1024 / D81681451 for addressing #264

The original change was reverted in
2747227a88409096d68d0a07223105024227d4f9 / D82227101 due to regression (#1073 and #1074).

This reland addresses the regression:
- allows annotation scope to access class fields (instead of static type information which was incorrect)
- allows comprehension scopre to access class fields (this leaves the test still broken --- since comprehension scoping is incorrect for iterator expressions in a comprehension)


fangyi-zhou:
cc @stroxler

stroxler:
@fangyi-zhou just a heads-up: I'm going ahead and fixing the comprehension scope issue on top of your change because I realized that some refactors to scope I need to make for other reasons are mixed in with this code

So don't worry about handling that :)

PR #1081, opened 2025-09-11
Add unix only tag for unix-only import in LSP test

fangyi-zhou:
This diff fixes open source windows build.

In L78-L83 of the file, the code block accessing unix permissions are already tagged `#[cfg(unix)]`. This diff adds the same config tag on the import so that windows build can be fixed.

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/configuration.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/configuration.rs
@@ -6,6 +6,7 @@
  */
 
 use std::fs;
+#[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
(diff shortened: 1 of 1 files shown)


yangdanny97:
I think we reverted that change

fangyi-zhou:
Open source build are still failing for windows on main though

fangyi-zhou:
See https://github.com/facebook/pyrefly/actions/workflows/pyrefly.yml

fangyi-zhou:
@yangdanny97 Do you mind reopening this PR?

rchen152:
Yeah, looks like the Windows build is still broken even after the revert. @fangyi-zhou thanks for the fix!

Issue #1038, opened 2025-09-04
Set `deprecated` tag on LSP completion items


vanillajonathan:
### Describe the Bug

I expect Pyrefly to set the `tags` property to `CompletionItemTag.Deprecated` on completion items which are deprecated.

yangdanny97:
cc @fangyi-zhou this might be an interesting followup for you since it's related to https://github.com/facebook/pyrefly/issues/376, but I can also take this if you don't want to

I assume setting the flag makes the item crossed out in the list of completions.

kinto0:
[#1093](https://github.com/facebook/pyrefly/pull/1093) implements this for functions. it is still not working for properties or classes. leaving this task open

PR #1093, opened 2025-09-14
Attach deprecated flags during auto complete

fangyi-zhou:
We know whether a method or a function is deprecated in Pyrefly, but the information is not propagated to LSP autocomplete (#1038). This diff passes along the deprecated tags in LSP auto completion. 

Caveat: deprecated properties in classes are currently not tagged with deprecated tag correctly.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -13,6 +13,7 @@ use fuzzy_matcher::skim::SkimMatcherV2;
 use itertools::Itertools;
 use lsp_types::CompletionItem;
 use lsp_types::CompletionItemKind;
+use lsp_types::CompletionItemTag;
 use lsp_types::DocumentSymbol;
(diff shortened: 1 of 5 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -2005,11 +2020,17 @@ impl<'a> Transaction<'a> {
                         })
                     }
                     let exports = self.get_exports(&handle);
-                    for name in exports.keys() {
+                    for (name, export) in exports.iter() {
+                        let is_deprecated = matches!(export, ExportLocation::ThisModule(export) if export.is_deprecated);
```
kinto0: nit: could we replace this with an explicit match? if we add or change cases it would make it easier to catch bugs

fangyi-zhou:
Screenshot 2025-09-14 at 21 45 09

stroxler (approved):

Issue #1095, opened 2025-09-14
Literal autocomplete should escape quote marks and other escapable strings


fangyi-zhou:
> If there's a quote mark in the string (or other escapeable characters like `\n`), we should probably escape this quote marks (and/or other things that possibly needs escaping)?
> 
> ```python
> def foo(quote_style: Literal["'", '"']):
>     ...
> 
> foo('<>')
> ```
> 
> Ideally, the autocomplete option inside `<>` should be `"` or `\'`
> 
> and similarly
> 
> ```python
> def foo(quote_style: Literal["'", '"']):
>     ...
> 
> foo("<>")
> ```
> 
> Ideally, the autocomplete option inside `<>` should be `\"` or `'`
> 
> (This should be filed under a different issue, but possibly can be fixed alongside)
>  

 _Originally posted by @fangyi-zhou in [#1086](https://github.com/facebook/pyrefly/issues/1086#issuecomment-3288934202)_

kinto0:
resolved with #1096, thank you!

Issue #1086, opened 2025-09-12
Literal autocomplete should take care of existing quote marks


fangyi-zhou:
### Describe the Bug

```python
from typing import Literal

def foo(fruit: Literal["apple", "pear"]):
    ...

foo(''
```

Python: 3.12

When the cursor is in between the quotes, the auto-complete populates the literals correctly. Then if I select one of the options, the post-completion state becomes

```
foo(''apple''
```

which didn't take into account the existing quote marks and is incorrect.

On a side note, the current auto-complete doesn't handle strings with quotes correctly --- which is an orthogonal issue with lower priority.

Relevant code pointer: https://github.com/facebook/pyrefly/blob/4ffa2e037a105ee284acc2b127ac454944ea96e1/pyrefly/lib/state/lsp.rs#L1934

Presumably this call wraps the string into another layer of single quotes
https://github.com/facebook/pyrefly/blob/4ffa2e037a105ee284acc2b127ac454944ea96e1/pyrefly/lib/state/lsp.rs#L1938

https://github.com/facebook/pyrefly/blob/4ffa2e037a105ee284acc2b127ac454944ea96e1/crates/pyrefly_types/src/literal.rs#L57

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AZCAxiVUUADropmGGDphcuABRhKAV2GJBw0eIDaEkKmLFYRgDR0jxGKkpGAugEpEUuh7qFvUqYpUA5AEgFiDqDNBwJOSIIADEdACqEVDCpArq6ADGEbjocL5Ycgq8NKgMAPro6jTYosr42qwMznQAtAB8dHAMlG7onnSUMAzqlANgRgByNXV9dMD4AL5GUiBLoag5EABuMABi0DAUaFh4RGTrQA&version=3.12

fangyi-zhou:
cc @kinto0

kinto0:
Thanks so much for the report! I'll try to get to this on monday. But feel free to take a stab at it in the meantime if you want since you seem to have found the root cause!

First issue: double quotes on completion, makes sense. 
Second issue: 

> On a side note, the current auto-complete doesn't handle strings with quotes correctly --- which is an orthogonal issue with lower priority.

What do you mean by this? Do you mean this example doesn't complete options?

> foo: Literal["apple"] = "<>"

[...]

PR #1096, opened 2025-09-14
Escape strings when auto-completing literals

fangyi-zhou:
Implementing escaping for single quotes to address #1095.

For a more proper fix, we should detect whether the string being completed is using single/double quotes (depends on fix for #1086), and auto-complete accordingly. Maybe, also a user-preference setting to allow the user to choose as well?

### DIFF
--- a/crates/pyrefly_types/src/literal.rs
+++ b/crates/pyrefly_types/src/literal.rs
@@ -54,7 +54,7 @@ pub struct LitEnum {
 impl Display for Lit {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Lit::Str(x) => write!(f, "'{x}'"),
+            Lit::Str(_) => write!(f, "{}", self.to_string_escaped(true)),
             Lit::Int(x) => write!(f, "{x}"),
             Lit::Bool(x) => {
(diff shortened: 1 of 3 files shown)


kinto0:
Sorry I didn't see this change yesterday and already landed a test case!

My proposed fix was [this](https://github.com/facebook/pyrefly/pull/1097?fbclid=IwY2xjawM1DxxleHRuA2FlbQIxMQBicmlkETFMT1hROGdQV2tTaEg4clBtAR4iTm4b7fIgSt1_Ywtu9M0gX2cpWQ21hTxbELcXECL4_FAGp6T7gOgDCvXeJw_aem_X7DKiIm_MIl4sKRR9DVRWQ). I was not sure what the side effects were from the rest of the typechecker and wanted to get some other thoughts on it first. @samwgoldman do you have opinions on the side effects of this change here? I wonder if it's easier to make an IDE-specific display. 

If we go with this approach, I think rust's `escape_default` might be easier than escaping them all manually.

samwgoldman:
I prefer to keep the data in the type unprocessed as much as possible, and if we need to escape for display, doing that at display time.

changing the text data in the literal type could interfere with narrowing, since it might change the nature of equality between two strings (two different strings become equal, eg)

PR #1097, opened 2025-09-15
escape escape characters

kinto0:
Differential Revision: D82452765

### DIFF
--- a/crates/pyrefly_types/src/literal.rs
+++ b/crates/pyrefly_types/src/literal.rs
@@ -54,7 +54,7 @@ pub struct LitEnum {
 impl Display for Lit {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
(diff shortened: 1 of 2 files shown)


fangyi-zhou:
Could you grab the testcase in #1096 and incorporate it in this diff? The main reason I wrote my own escaping function was due to (1) not escaping single quotes in double quotes and vice versa, (2) Python and Rust escaping sequences might be differing (e.g. Python allows `\a` `\b` `\v` that is not accepted in Rust).

kinto0:
> Could you grab the testcase in #1096 and incorporate it in this diff? The main reason I wrote my own escaping function was due to (1) not escaping single quotes in double quotes and vice versa, (2) Python and Rust escaping sequences might be differing (e.g. Python allows `\a` `\b` `\v` that is not accepted in Rust).

good call. I did not consider that different languages had different escape sequences. I'll close this and keep yours. thanks!

PR #1101, opened 2025-09-16
Add django stubs to test env

migeed-z:
Summary:
Added stubs for models.

This directory contains many useful stubs which I noticed in the testsuite, including field definitions, enums, as well as the Model definition.

There were some subdirectories in the original stubs which I removed until I know for sure if we need them.

Also silenced all linting errors related to missing license.

Differential Revision: D82494385


samwgoldman (approved):

PR #1105, opened 2025-09-16
Add license file to pydantic stubs

migeed-z:
Summary:
We should also rename the directory from stubs to third-party so that in the future, we don't have to silence any lints.

Rollback Plan:

Differential Revision: D82552558


samwgoldman (approved):

PR #1112, opened 2025-09-18
Sandbox: add `pyrefly.toml` support + config-driven version precedence

### LINKED ISSUE #1062: Sandbox: pyrefly.toml file
### Describe the Bug

Now that we have multi-file support in the sandbox, we can add a config file so that users can experiment with different configurations

### Sandbox Link

_No response_

_No response_

AryanBagade:
Adds first-class `pyrefly.toml` support to the website sandbox so users can experiment with Pyrefly configuration in-browser. When a config is present, it becomes the source of truth for Python version and error rules.

### What’s included
- UI (website)
  - Allow creating/editing `pyrefly.toml` in the sandbox (TOML syntax highlighting).
  - Persist `pyrefly.toml` in the shareable URL (`?project=...` payload).
  - When `pyrefly.toml` exists, the URL omits `?version=...`.
  - Version selector is effectively disabled when a config is present (to avoid conflicts).
  
- WASM
  - Only `.py` files are treated as modules; non-`.py` files are ignored.
  - Parse and apply `pyrefly.toml` (TOML) in `Playground.update_sandbox_files`.
  - Surface config parse/validation errors as diagnostics attached to `pyrefly.toml`.
  
Files touched
- Sandbox UI: [`website/src/sandbox/Sandbox.tsx`](https://github.com/AryanBagade/pyrefly/blob/sandbox-pyrefly-toml/website/src/sandbox/Sandbox.tsx)
- Playground engine: [`pyrefly/lib/playground.rs`](https://github.com/AryanBagade/pyrefly/blob/sandbox-pyrefly-toml/pyrefly/lib/playground.rs)
- Tests (snapshot): [`website/src/__tests__/__snapshots__/sandbox.test.tsx.snap`](https://github.com/AryanBagade/pyrefly/blob/sandbox-pyrefly-toml/website/src/__tests__/__snapshots__/sandbox.test.tsx.snap)

Closes #1062

### DIFF
--- a/website/src/__tests__/__snapshots__/sandbox.test.tsx.snap
+++ b/website/src/__tests__/__snapshots__/sandbox.test.tsx.snap
@@ -17,6 +17,11 @@ exports[`Sandbox Component render sandbox correctly 1`] = `
         </button>
       </div>
       <div>
+        <button>
+          pyrefly.toml
+        </button>
+      </div>
+      <div>
         <button
           disabled=""
           title="Cannot delete sandbox.py"
@@ -118,54 +123,6 @@ reveal_type(test(42))
             ⚠️ Report Issue
           </span>
         </button>
-        <div>
-          <select
-            aria-label="Select Python version"
-          >
-            <option
-              value="3.8"
-            >
-              Python 
-              3.8
-            </option>
-            <option
-              value="3.9"
-            >
-              Python 
-              3.9
-            </option>
-            <option
-              value="3.10"
-            >
-              Python 
-              3.10
-            </option>
-            <option
-              value="3.11"
-            >
-              Python 
-              3.11
-            </option>
-            <option
-              value="3.12"
-            >
-              Python 
-              3.12
-            </option>
-            <option
-              value="3.13"
-            >
-              Python 
-              3.13
-            </option>
-            <option
-              value="3.14"
-            >
-              Python 
-              3.14
-            </option>
-          </select>
-        </div>
       </div>
     </div>
     <div


rchen152:
@AryanBagade Thank you so much for doing this! Some thoughts:
* Would it be possible to have the `pyrefly.toml` file in the default sandbox? That would make it easier to play around with the configuration (and more obvious that the sandbox has this capability).
* The "cfg" button doesn't seem necessary, since clicking the regular "+" lets you add a `pyrefly.toml`.
* The version selector is redundant now that `python_version` can be set in the configuration file. Would be nice to clean that up and regain a bit of screen real estate.

AryanBagade:
@rchen152 
### Update

- Added default `pyrefly.toml` in the sandbox.
- Removed “cfg” button and version selector (config is the source of truth).
- URL omits `?version=` when config exists; share/restore via `?project=…` only.
- WASM applies `pyrefly.toml`; config errors surface as diagnostics; only `.py` files are modules.

website/src/__tests__/__snapshots__/sandbox.test.tsx.snap
```diff
@@ -28,6 +28,11 @@ exports[`Sandbox Component render sandbox correctly 1`] = `
         >
           +
         </button>
+        <button
```
rchen152: I think this change should be reverted? There's no longer a cfg button.

website/src/sandbox/Sandbox.tsx
```diff
@@ -165,17 +165,18 @@ export default function Sandbox({
     const renameFile = useCallback((oldName: string, newName: string) => {
         if (!newName.trim() || models.has(newName)) return false;
 
-        // Ensure .py extension
-        const finalName = newName.endsWith('.py') ? newName : `${newName}.py`;
+        // Allow .py and .toml; default to .py if no known extension
+        const finalName = (newName.endsWith('.py') || newName.endsWith('.toml')) ? newName : `${newName}.py`;
```
rchen152: Since playground.rs only recognizes `pyrefly.toml`, I don't think we should allow other `.toml` files. They'd just be ignored, which I think would be confusing. Instead, what about allowing only `pyrefly.toml` and otherwise always adding `.py`?

website/src/sandbox/Sandbox.tsx
```diff
@@ -213,9 +214,10 @@ export default function Sandbox({
             return false;
         }
 
-        const finalName = inputValue.trim().endsWith('.py')
-            ? inputValue.trim()
-            : `${inputValue.trim()}.py`;
+        const trimmed = inputValue.trim();
+        const finalName = (trimmed.endsWith('.py') || trimmed.endsWith('.toml'))
```
rchen152: Same comment as on line 169; if we only support `pyrefly.toml`, we shouldn't allow other `.toml` files to be created.

website/src/sandbox/Sandbox.tsx
```diff
@@ -383,19 +385,22 @@ export default function Sandbox({
...
                     createNewFile('sandbox.py', DEFAULT_SANDBOX_PROGRAM);
                     createNewFile('utils.py', DEFAULT_UTILS_PROGRAM);
+                    // Add a default pyrefly.toml so users can immediately tweak configuration
+                    const defaultConfig = [
+                        '# Pyrefly sandbox configuration',
+                        'python-version = "3.12"',
+                        '',
+                        '[errors]',
+                        'reveal-type = false',
+                        ''
```
rchen152: Hmm, I don't think it's a good idea to disable reveal-type; that makes the sandbox look broken. How about this instead? I deleted the errors section but added a link to the configuration docs so that users can go there to see what error configuration looks like.

```suggestion
                        '# Pyrefly sandbox configuration.',
                        '# See https://pyrefly.org/en/docs/configuration/ for available configuration options."
                        'python-version = "3.12"',
                        '',
```

website/src/sandbox/Sandbox.tsx
```diff
@@ -383,19 +385,22 @@ export default function Sandbox({
...
+                    ].join('\n');
+                    createNewFile('pyrefly.toml', defaultConfig);
                     setActiveFileName('sandbox.py');
                 }
             }
         }
     }, [createNewFile, models.size, isCodeSnippet, sampleFilename, codeSample]);
 
-    // Initialize Python version from URL on component mount
```
rchen152, line 392: Would it be possible to remove the version selector but still keep the code that sets the version based on the url? That way old sandbox urls can continue to work.

website/src/__tests__/__snapshots__/sandbox.test.tsx.snap
AryanBagade: Good catch, agreed. The cfg button is removed now, so the snapshot was updated to reflect its absence.

website/src/sandbox/Sandbox.tsx
AryanBagade: Good point —> implemented.
- Only `pyrefly.toml` is allowed now; any other `.toml` is rejected in rename/validation.
- Filenames without an extension default to `.py`.

website/src/sandbox/Sandbox.tsx
AryanBagade: Thanks for the comment, applied!!

website/src/sandbox/Sandbox.tsx
AryanBagade: Version Selector is removed

`?version=` behavior:
- `?version=` and no `?project=`: on first load we auto-create `pyrefly.toml` with `python-version` set to that value (becomes the source of truth).
- `?project=` present but it has no `pyrefly.toml`: we use `?version=` only to initialize the runtime for that session; we do not create `pyrefly.toml`. (old sandbox urls will work without any changes)
- `?project=` contains `pyrefly.toml`: we ignore `?version` (config wins).
- After load, any edit/share rewrites the URL to only `?project=…` — we never write `?version=` back.

rchen152:
Hey @AryanBagade, I didn't get time to review this today, but I'll do so as soon as I can. Sorry for the wait!

AryanBagade:
No Worries, take your time

AryanBagade:
Hi @rchen152, no rush, just following up to see if you might get a chance to review these changes sometime soon. Thanks a lot!

rchen152 (approved):
Thanks for your patience! I'm working on getting this merged internally.

Issue #706, opened 2025-07-18
[bug] `C` should not be assignable to `Self@C` unless `C` is final


yangdanny97:
### Describe the Bug

Right now a concrete instance of the current class is always treated as assignable to `Self`, but this is not true if the current class is not final. See sandbox example.

Code pointer: https://github.com/facebook/pyrefly/blob/fac4501/pyrefly/lib/solver/subset.rs#L975

Please add integration tests in pyrefly/lib/tests and once the tests pass, run test.py and commit any generated changes.

### Sandbox Link

- https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDKApgDbABQ5AxiQIYDO9UAggFzlSdQAmRwsRejAAU9UsACUUALQA+QuPZdlUEERgBXEChbCpAYigBREOBBA

connernilsen:
@yangdanny97 can you add a new pointer here? It looks like [this one](https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/solver/subset.rs#L878) wasn't pointing to any specific commit, so it's not clear what the link was pointing to.

yangdanny97:
Updated the code pointer to: https://github.com/facebook/pyrefly/blob/fac4501/pyrefly/lib/solver/subset.rs#L975

samwgoldman:
I’m not sure the final special case is really necessary. If a class is final, there’s not much point in using the Self annotation anyway.

connernilsen:
Hey @edmondop, thanks for volunteering! I'll add you to this, and let us know if you run into any problems or have questions here or in the Discord server!

fangyi-zhou:
I suppose we can finally close this after https://github.com/facebook/pyrefly/commit/b037b1e739e51a1a6d3106ebee1a08b15928a51c

yangdanny97:
nice! whew, what a journey

yangdanny97:
thanks for all your hard work here @fangyi-zhou

PR #1117, opened 2025-09-20
C should not be assignable to Self@C unless C is final #706

edmondop:
This PR addresses https://github.com/facebook/pyrefly/issues/706 by making  the `subset.rs` checks  on Self return type more strict


stroxler (commented):
Hi @edmondop, thanks for the PR!

This looks plausible to me, but there are several failing tests and it would be nice to see the behavior changes in the diff view in case this is interacting badly with other parts of Pyrefly (in particular, I see some incompatible-override errors popping up and it would be nice to make sure those make sense).

cc @rchen152 and @samwgoldman  who are probably our experts on the Self type (and for that matter function subtyping as well!)

edmondop:
Thanks, will look into the failing tests. I didn’t catch them locally
because I think they were all passing until the new one was failing and
once I fixed the new one, I forgot to run again the others. Let me look
into the failing tests before anyone spend time reviewing broken code

edmondop:
I confirm the change has broken some existing funtionality, and some tests that should be passing are not. I haven't investigated the other tests, but this one is a good example:

```rust
testcase!(
    test_call_instance_method_from_classmethod,
    r#"
class A:
    def f(self):
        pass

class B(A):
    @classmethod
    def g(cls):
        super().f(cls())
    "#,
);
```

I need to dive deeper into `subset.rs` and understand if I have the context, I think the difference in the good case is that Self is in the return type, while here is an argument to f. According to Liskov Substitution principle, this test case should pass for any subclass of B, so the fact that Bm is not final is not a problem

edmondop:
I haven't being able to find a solution that affect only `subset.rs` that wouldn't break previous tests, because within solver it's unclear whether got/want are in parameter or return type position, while in the solve.rs I extended the return type handling code

pyrefly/lib/alt/solve.rs
```diff
@@ -2679,6 +2679,27 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                             &|| TypeCheckContext::of_kind(TypeCheckKind::TypeGuardReturn);
                         self.expr(expr, hint.as_ref().map(|t| (t, tcc)), errors)
                     } else {
+                        if let Some(Type::SelfType(want_class)) = hint.as_ref() {
+                            if !self.type_order().is_final(want_class.class_object()) {
```
asukaminato0721: how about https://rust-lang.github.io/rfcs/2497-if-let-chains.html

edmondop: Thanks, I thought I had run cargo clippy. Fixed now

stroxler:
Yeah this is tricky and tied into a few very difficult questions:
- what the `Self` type actually means (PEP 673 actually specified it as a *method* level type parameter, but I don't think most type checkers treat it that way anymore)
- whether it is allowed to appear in parameter position, except by being bound as the receiver object (this is unsound, but I think most type checkers allow it anyway because best-effort unsound analysis is sometimes helpful for stating intent and catching a subset of bugs)

If we actually solved `Self` as a bound type variable in the call as PEP 673 specifies, I'm pretty certain that the error would go away but at the moment our handling of `Self` is entirely adhoc and I don't think we ever actually solve it as a `Quantified`.

The adhoc handling has benefits because it allows us to do things like handle an attribute specified as `Self` (which is again unsound if the attribute is mutable, but we allow this), so I'm unsure how easy it would be to change the behavior and match PEP 673 better. We might be able to detect the direct use of an unbound method and convert it to a ForAll type where `Self` behaves like a normal function-level type variable.

Let's see if @rchen152 and @samwgoldman have thoughts

samwgoldman:
Thanks for contributing! I don't think the fix is in the right place. This diff adds some logic to the code handling returns with explicit annotations, but assignability happens in more places, like `x: Self = C()` within a method in `C`. It would also apply to the cases in returns that are not handled by this change, async and generated functions.

The right place for this fix is to remove the invalid rules in subset.rs, here: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/solver/subset.rs#L1005-L1013

These rules say it is OK to assign C to Self@C, which is exactly false, so the right thing to do is remove them.

However, there is a snag -- we rely on this rule in a few places. Usually this happens when we fail to preserve "Self"-ness. I recently fixed a couple examples of this, in 968a99c3a997ca4c198ce0897500668bc338e8e3 and in fc7592c1bc92558c79b5c025e5b45835db0760b1.

I know of at least one other place we get this wrong -- calling a constructor on the call target `type[Self]` [[Sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BlGFDAAaOqjhwYfAPpNiMADroVAYyiS4dAMKIVdQ3QACGrTRgMAFrkwGjmGGDo0AFBrgBKfeiN%2BJUjIM8sww7lBwrp7iQiKeIKIgAK4M0HAk5IggAMR0AKqpUBBMdGBJ6GqpuOhwKiqOzmC8NKjB6Ek02DKu%2BIic6AyedAC0AHx0cAyUPn6UlkmUvmBKIAByHV3TdMD4AL4rKiC7iaiVEABuMABi0DAUaFh4RGRHQA&version=3.12)].

samwgoldman (changes requested):
I think the original issue is mislabeled as "good first issue" -- but you are welcome to keep going if you'd like. Try removing the invalid rules first and see what tests fail. Hopefully you will see an example like the ctor call on `type[Self]`, but maybe some others. Those issues would need to be fixed first, preferably in separate PRs.

edmondop:
Thanks, I tried to originally make that change and it was causing several test to fail. I was trying to restrict my condition in the `subset.rs` based on the location of the type (return type) but I wasn't able to find a way through the `type_order` to access this information. Was I looking in the wrong place?

samwgoldman:
I see -- in subset.rs it would not be possible to restrict to just return types, but I don't think such a restriction is worthwhile, since the bug exists everywhere. The fact that other tests start to fail is likely because of other latent issues that need to be fixed first, per my comment above. For example, the `test_call_instance_method_from_classmethod` fails because calling `cls()` _should_ return `Self@C` but currently incorrectly returns `C` -- constructing `type[Self]` should return `Self`.

edmondop:
I think I found the root cause of why I wasn't able to solve the problem in `subset.rs`.  Doing the check on final would fix the problems indicated in the issue #706 but would introduce test failures somewhere else.

The problem was that in `call.rs`

```rust
     Type::Type(box Type::ClassType(cls)) | Type::Type(box Type::SelfType(cls)) => {
                Some(CallTarget::Class(cls))
            }
```

this handling means that later in the `call_infer`, the type[Self] becomes a type[ConcreteClass],

In practice this mean that would end up being processed on a different branch of subset.rs rather than the one we expect


```python
class B:
    @classmethod
    def factory(cls) -> Self:
        return cls()
```


The solution seems to be to introduce a new enum type for the CallTarget of type Self, and produce a Type::SelfType later for subset.rs. This ensures that `subset.rs` pattern matching receive the right type. (I haven't figure out how to handle type parameters yet)

pyrefly/lib/alt/call.rs
```diff
@@ -633,6 +634,47 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 };
                 self.construct_class(cls, args, keywords, range, errors, context, hint)
             }
+            CallTarget::SelfClass(cls) => {
+                if cls.has_qname("typing", "Any") {
```
edmondop, line 638: maybe this should be factor out so we can share it between ClassTarget::Class and ClassTarget::SelfClass since this code is identical

pyrefly/lib/alt/call.rs
```diff
@@ -633,6 +634,47 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                    self.construct_class(cls.clone(), args, keywords, range, errors, context, hint);
+                // Handle custom __new__
+                match &result {
+                    Type::ClassType(result_cls)
+                        if result_cls.class_object() == cls.class_object() =>
+                    {
+                        Type::SelfType(result_cls.clone())
+                    }
+                    _ => result,
+                }
```
samwgoldman, line 676: This part seems a bit off to me. If we have `type[Self]` where in `class C`, then we will call `construct_class` using `C`, then check to see if that method returned `C`, and if it does assume that the constructor actually returned `Self`.

In order to return `Self`, we need to know that for `C` _and any of its subclasses_, the constructor call will return the class itself. Here we are only checking that calling it for `C` returns `C`, so the implication feels incomplete.

Consider this program, which fails at runtime. Because of the `__new__` implementation, which always returns a `C`, when we construct `cls(0)` when called from a subclass, we actually get a `C` not a `D`.

```
from typing import Self

class C:
    def __new__(cls, x: int) -> 'C': # Note C instead of Self here
        return C(x)

    @classmethod
    def m(cls: type[Self]) -> Self:
        return cls(0) # should fail

class D(C):
    d_only: int = 0

D.m().d_only # runtime error
```

My feeling is that the solution here needs a bit more work. We would need to push the "self"-ness into `construct_class`, and be a bit careful to check that metaclass `__call__` and `__new__` preserve "self"-ness as well, so we can infer that `cls(0)` on `cls: type[Self]` actually returns `C` and not `Self`.

That said, looking at the behavior of mypy and Pyright, both seem to accept the program above. I couldn't find any explanation for that behavior in the typing spec, which talks a lot about metaclass `__call__` and `__new__`, but not `type[Self]`.

I'm hesitant to request changes on this bit, because I think those changes might be pretty intricate. I'd really like to understand why mypy/Pyright seem to accept the program above.

pyrefly/lib/alt/call.rs
samwgoldman: I think we can skip all of these checks for `CallTarget::SelfClass`, actually. 

1. The only place we could have `type[Self]` where `Self` is `Any` is the implementation of `Any`, which doesn't actually exist.

2. For `type[P]` where `P` is a protocol, we actually should allow calling a constructor. This also applies to `type[Self]` where `Self` is a protocol.

3. Similar to (1), the only place this check comes up is in the implementation of `bool` itself, since `type[Self]` where `Self` is `typing.bool` can only exist inside that class.

pyrefly/lib/solver/subset.rs
```diff
@@ -1002,11 +1002,24 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
...
-            ),
+            (Type::ClassType(got), Type::SelfType(want)) => {
+                let has_superclass = self
+                    .type_order
+                    .has_superclass(got.class_object(), want.class_object());
+                // Only apply the finality check when:
+                // 1. The classes are the same
+                // 2. Both have no type arguments (i.e., they're the exact same bare type)
+                // This ensures NonFinalClass is not assignable to Self@NonFinalClass,
+                // but allows partial[int, str] to be assignable to Self@partial
```
samwgoldman, line 1013: Do you have a test case for this? I'm not sure I understand what this is referring to.

pyrefly/lib/solver/subset.rs
```diff
@@ -1002,11 +1002,24 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
...
-                    .has_superclass(got.class_object(), want.class_object()),
-                SubsetError::Other,
-            ),
+            (Type::ClassType(got), Type::SelfType(want)) => {
+                let has_superclass = self
+                    .type_order
+                    .has_superclass(got.class_object(), want.class_object());
+                // Only apply the finality check when:
+                // 1. The classes are the same
+                // 2. Both have no type arguments (i.e., they're the exact same bare type)
```
samwgoldman, line 1011: A test case with an expected error in this case, which also explains why it's not safe, would be great.

pyrefly/lib/alt/call.rs
edmondop: Thanks, do we need to log a separate issue for a deeper dive in the self handling in pyright and mypy?

Issue #1091, opened 2025-09-13
Incorrect return path identification with try/finally


rwarren:
### Describe the Bug

This code:

```python
def foo() -> int:
    try:
        return 1
    finally:
        print("done")
```

Results in:
```
Function declared to return `int`, but one or more paths are missing an explicit `return
```

That `finally` return path should not be considered as a return path.

I checked both `mypy` and `pyright` and neither trip up on this.

### Sandbox Link

⚠️ I can't post a sandbox link right now because the sandbox is down!

Zaczero:
https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGLrgBQCUNAtAHw0ToAuiAHXQ0RNXgCdSg4aNniYvAK7jhARiGzI6VFChSNs0cXE9ejAZVzoYF5iAA0IRb2hwS5RCADENAKouoCF5SOkV0AGMXKzghISpaenEAW1ReAH10RSTsGHFGfERuPlZOGjgJaTkFZWEwCwA5LJzxQuB8AF8LIRB2x1RIiAA3GAAxaBgKNCw8IjIeoA&version=3.12

Sandbox link

PR #1118, opened 2025-09-21
Fix: Incorrect return path identification with try/finally #1091

sandeshbhusal:
This patch fixes an error in incorrect return patch identification with try/finally.
When a finally block is seen:
- Previously: only process the finally block for returns
- Now: check if finally block has a "return" statement. If not, process other blocks.

E.g. failing case:

```python
try:
    return 1
finally:
    pass
```


ndmitchell (approved):

PR #1153, opened 2025-09-24
Add django metadata to enum metadata

migeed-z:
Summary:
Enums are responsible for the majority of false positives in the django stubs' assert_type testsuite.

Out of 71  total false positives, 59 are coming from enums and 41 of them are assert type failures.

In this diff, we start by adding a flag to identify IntegerChoices, which is an enum class defined in django stubs.

Reviewed By: rchen152

Differential Revision: D83096242


rchen152 (approved):

PR #1183, opened 2025-09-25
Do not require an optional pydantic field

migeed-z:
Summary: We should inspect field annotations and if they are optional, we should not require them.

Reviewed By: rchen152

Differential Revision: D83273115


rchen152 (approved):

PR #1185, opened 2025-09-26
Throws Bad Instantation Error when Creating an Instance of an Abstract Class

Adist319:
Continuation of PR #807. Addresses issue #511 and #592.

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -28,6 +28,7 @@ use starlark_map::small_map::SmallMap;
 
 use crate::alt::answers::LookupAnswer;
 use crate::alt::answers_solver::AnswersSolver;
+use crate::alt::class::class_field::ClassField;
 use crate::alt::solve::TypeFormContext;
 use crate::alt::types::class_metadata::ClassMetadata;
 use crate::alt::types::class_metadata::ClassMro;
@@ -274,6 +275,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             self.validate_frozen_dataclass_inheritance(cls, dm, &bases_with_metadata, errors);
         }
 
+        // Compute abstract class metadata before consuming bases_with_metadata
+        let inherits_from_abc = self.check_inherits_from_abc(cls, &bases_with_metadata);
+        let abstract_methods = if inherits_from_abc {
+            self.collect_unimplemented_abstract_methods(cls, &bases_with_metadata)
+        } else {
+            Vec::new()
+        };
+
         // Compute final base class list.
         let bases = if is_typed_dict && bases_with_metadata.is_empty() {
             // This is a "fallback" class that contains attributes that are available on all TypedDict subclasses.
@@ -298,6 +307,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .as_ref()
             .map(|m| m.pydantic_model_kind.clone());
 
+        // Check if @final class has abstract methods (Issue #957)
+        if is_final && !abstract_methods.is_empty() {
+            let methods_str = abstract_methods
+                .iter()
+                .map(|n| format!("`{}`", n))
+                .collect::<Vec<_>>()
+                .join(", ");
+            self.error(
+                errors,
+                cls.range(),
+                ErrorInfo::Kind(ErrorKind::InvalidInheritance),
+                format!(
+                    "Final class `{}` has unimplemented abstract methods: {}",
+                    cls.name(),
+                    methods_str
+                ),
+            );
+        }
+
         ClassMetadata::new(
             bases,
             metaclass,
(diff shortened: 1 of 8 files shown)


Adist319:
I’m not seeing any issues locally (cargo fmt / cargo check both pass), but CI continues to fail. To keep things clean, I’ll open a new PR with the changes rebased on main and with feedback addressed as it comes in.

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -1242,4 +1259,142 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        bases_with_metadata: &Vec<(Class, Arc<ClassMetadata>)>,
+    ) -> bool {
+        // Check if any base is abc.ABC
+        for (base, _) in bases_with_metadata {
+            if base.module_name().as_str() == "abc" && base.name().as_str() == "ABC" {
+                return true;
+            }
+        }
+
+        // Check if any base inherits from abc.ABC
```
yangdanny97, line 1276: could we include this check in the previous loop since it's iterating over the same thing?

if base.has_qname("abc", "ABC") || metadata.inherits_from_abc() { return true }

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -1242,4 +1259,142 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+
+    /// Check if a class inherits from abc.ABC
+    fn check_inherits_from_abc(
+        &self,
+        _cls: &Class,
+        bases_with_metadata: &Vec<(Class, Arc<ClassMetadata>)>,
+    ) -> bool {
+        // Check if any base is abc.ABC
+        for (base, _) in bases_with_metadata {
+            if base.module_name().as_str() == "abc" && base.name().as_str() == "ABC" {
```
yangdanny97, line 1271: ` base.has_qname("abc", "ABC")`

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -399,7 +426,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         bases_with_metadata
             .iter()
             .find_map(|(base_class_object, metadata)| {
-                if base_class_object.has_toplevel_qname(
```
yangdanny97, line 402: these diffs are fishy, looks like there was a conflict & might need to revert some of those changes

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -1242,4 +1259,142 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .collect();
         ClassMro::new(cls, bases_with_mros, errors)
     }
+
+    /// Check if a class inherits from abc.ABC
```
yangdanny97, line 1263: i think the docstring is redundant

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -1242,4 +1259,142 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        bases_with_metadata: &Vec<(Class, Arc<ClassMetadata>)>,
+    ) -> Vec<Name> {
+        use starlark_map::small_set::SmallSet;
+
+        let mut abstract_methods = SmallSet::new();
+
+        // Collect abstract methods from all base classes
+        for (_, metadata) in bases_with_metadata {
+            for method in metadata.abstract_methods() {
+                abstract_methods.insert(method.clone());
```
yangdanny97, line 1299: q here: is it possible that one base class declares a method as abstract but another base class provides the concrete impl? if so, idt we handle that case.

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -1242,4 +1259,142 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                }
+            }
+        }
+
+        // Also check for any new abstract methods defined in this class
+        for field_name in cls.fields() {
+            if let Some(field) =
+                self.get_from_class(cls, &KeyClassField(cls.index(), field_name.clone()))
+            {
+                if self.is_abstract_field(&field, cls, field_name) {
```
yangdanny97, line 1322: the implementation for is_abstract/is_concrete are fairly complex, instead of looking thru the bindings, I think we could calculate an is_abstract flag and store it on the ClassField. When ClassField is calculated we should have access to the original method declarations along w/ any decorators.

This would be similar to what I did in https://github.com/facebook/pyrefly/commit/66ecc58d068e2dfc55bc11d3b208f167867115ce, though in that case I didn't store it in ClassField I put it in another struct.

yangdanny97 (changes requested):
back to you with some comments, thanks

Adist319:
Thanks for the feedback @yangdanny97, will also create a new PR to more cleanly resolve the conflicts.

Adist319:
New PR at #1192, resolved merge conflicts.

PR #1192, opened 2025-09-26
WIP: Throws Bad Instantation Error when Creating an Instance of an Abstract Class

Adist319:
# Update on PR #1185, Need Guidance on Metadata Issue

Hey @yangdanny97,

I've addressed all your feedback from the PR, including storing the `is_abstract` flag in ClassField (similar to your `is_deprecated` in 66ecc58). I believe we did talk about storing this flag in ClassField before, but now I'm hitting a metadata propagation issue:

The flag gets set properly during decorator processing in [`function.rs:437`](I can see it in debug output), but by the time we check it in [`class_field.rs:1386-1391`] after `deep_force`, it's always false.

I suspect this is why in my original implementation, I had that complex check directly in the class field creation logic to work around this metadata propagation issue. The type seems to lose its FuncMetadata when going through `solve_binding` → `deep_force`.

The issue with the workflow currently: When a method with `@abstractmethod` is processed, the `is_abstract_method` flag is set in the FuncFlags during decorator processing, but this metadata doesn't survive to the final Type that gets stored in ClassField. The type resolution pipeline (`solve_binding` → `deep_force`) seems to reconstruct the Type without preserving the FuncMetadata.

Should I:
1. **Fix the root cause** - Ensure FuncMetadata is preserved through the type resolution pipeline (might affect other metadata like `is_deprecated` too), or
2. **Work around it** - Move the abstract method detection logic directly into class field creation, checking decorators there instead of relying on FuncMetadata

Which approach aligns better with the codebase architecture? Happy to hear your guidance.

Thanks!


yangdanny97:
When we solve bindings and hit some recursive case where we need the value of the thing being solved, we use a recursive placeholder, which has default values for certain things. I wonder if that's happening here. I'll take a look this week

yangdanny97:
Yep, it's the recursive case as described above.

I think a working approach would be something like this:
1. in class metadata store whether or not the class inherits from ABC or has metaclass=ABCMeta + the list of fields names (without checking if those fields are abstract or not, this is just a list of fields to check later)
2. in the class field store whether the field is abstract
3. on class instantiation look through the list of fields in the class metadata, and check whether the field returned in the MRO is abstract

I can experiment with this on top of your PR, this is a great start and I think the lift to make it fully working is small.

For the "error on declaration" part for final classes, I'll think a bit more about how to do that. We might need to make a new binding for the class for checks that happen after the class fields are solved.

yangdanny97:
@Adist319 

I've patched this PR according to the previous comments and will merge those changes soon. To keep things simple the merged changes will not include the checks for `@final` classes in the original PR; those can be implemented as described below once this is merged.

https://github.com/facebook/pyrefly/issues/1200 and https://github.com/facebook/pyrefly/issues/511 can be implemented by creating a new kind of [BindingExpect](https://github.com/facebook/pyrefly/blob/2bc073a2d9c0352c35a501e22e2c05c6f1ee90ae/pyrefly/lib/binding/binding.rs#L504), which can contain validations that happen after the class metadata & all the fields are calculated.

We'd want an error that's on-by-default for #511 and an error that's off-by-default for #1200, due to the latter's risk of false positives.

Adist319:
Got it, thanks for taking a look and patching this! Let me know if you want me to follow up with the @final class checks in a separate PR using the BindingExpect approach you described, once the changes are merged.

grievejia (approved):

Issue #874, opened 2025-08-08
Relative module autocomplete


lolpack:
### Describe the Bug

implement relative module autocomplete on imports

SamChou19815:
@lolpack Do you have specific examples?

lolpack:
- Import autocomplete for relative imports
- Have it work without a config needing to be present

Maybe both of these exist today?

kinto0:
reopening: relative import autocomplete is still not done

PR #1193, opened 2025-09-26
relative modules on import

kinto0:
Summary:
completion for the <> in from .<> imports currently don't work.

in this diff, I implement basic support for them:

- if we have a config and this file is in that config, use the absolute import path for them
- if this fails, use the fallback_search_path at the level of the dots

we still don't have support for go-to definition / hover on these relative modules.

fixes https://github.com/facebook/pyrefly/issues/874

Differential Revision: D83351301


stroxler:
Closing this to clean up the PR backlog since it looks like a separate fix has landed - feel free to reopen if I'm mistaken

PR #1194, opened 2025-09-26
Add support for TSP snapshot request

rchiodo:
This PR adds another request handler for TSP = "typeServer/getSnapshot". The snapshot request is how the client side of a TSP connection indicates which 'state of the world' it's currently querying about. The server side keeps a 'snapshot' to indicate when the state of the world has changed. 

Internally I used the 'epoch' and the 'RecheckFinished' event to indicate what the snapshot is and when it's changed.

### DIFF
--- a/pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
+++ b/pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//! Tests for TSP getSnapshot request
+
+use lsp_server::RequestId;
+use lsp_server::Response;
+use tempfile::TempDir;
+
+use crate::test::tsp::tsp_interaction::object_model::TspInteraction;
+
+#[test]
+fn test_tsp_get_snapshot() {
+    // Test retrieval of TSP snapshot version
+    let temp_dir = TempDir::new().unwrap();
+    let test_file_path = temp_dir.path().join("test.py");
+
+    let test_content = r#"# Simple test file
+print("Hello, World!")
+"#;
+
+    std::fs::write(&test_file_path, test_content).unwrap();
+
+    let mut tsp = TspInteraction::new();
+    tsp.set_root(temp_dir.path().to_path_buf());
+    tsp.initialize(Default::default());
+
+    // Open the test file
+    tsp.server.did_open("test.py");
+
+    // Wait for any diagnostics/RecheckFinished events
+    tsp.client.expect_any_message();
+
+    // Get snapshot
+    tsp.server.get_snapshot();
+
+    // Expect snapshot response with integer
+    tsp.client.expect_response(Response {
+        id: RequestId::from(2),
+        result: Some(serde_json::json!(0)), // Should start at epoch 0
+        error: None,
+    });
+
+    tsp.shutdown();
+}
+
+#[test]
+fn test_tsp_snapshot_updates_on_file_change() {
+    // Test that snapshot increments when files change
+    let temp_dir = TempDir::new().unwrap();
+    let test_file_path = temp_dir.path().join("changing_test.py");
+
+    let initial_content = r#"# Initial content
+x = 1
+"#;
+
+    std::fs::write(&test_file_path, initial_content).unwrap();
+
+    let mut tsp = TspInteraction::new();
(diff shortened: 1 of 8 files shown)


connernilsen:
Hey @rchiodo, thanks for doing this. Kyle's out until Friday, and I don't know as much of what's going on, so I'm going to assign to Kyle for now and let him review when he's back. Let me know if there's anything else you need in the meantime :)

rchiodo:
> Hey @rchiodo, thanks for doing this. Kyle's out until Friday, and I don't know as much of what's going on, so I'm going to assign to Kyle for now and let him review when he's back. Let me know if there's anything else you need in the meantime :)

Thanks for letting me know :). No rush but I was worried Meta was no longer interested in TSP.

pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    std::fs::write(&test_file_path, updated_content).unwrap();
+
+    // Get updated snapshot - snapshot tracking works automatically via RecheckFinished
+    tsp.server.get_snapshot();
+
+    // Expect second snapshot response (value may be same since file watching is complex)
+    tsp.client.expect_response(Response {
+        id: RequestId::from(3),
+        result: Some(serde_json::json!(0)), // May still be 0 without proper file watching
```
kinto0: this doesn't seem like it tests anything. can we make this test guaranteed to update the epoch?

pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
```diff
@@ -0,0 +1,102 @@
...
+        id: RequestId::from(2),
+        result: Some(serde_json::json!(0)), // Should start at epoch 0
+        error: None,
+    });
+
+    tsp.shutdown();
+}
+
+#[test]
+fn test_tsp_snapshot_updates_on_file_change() {
```
kinto0, line 63: can we add a test for a new epoch on in-memory-content changes (textdocument/didChange)?

pyrefly/lib/tsp/server.rs
```diff
@@ -43,6 +49,21 @@ impl TspServer {
         subsequent_mutation: bool,
         event: LspEvent,
     ) -> anyhow::Result<ProcessEvent> {
+        // Handle TSP-specific logic for RecheckFinished to track snapshot updates
+        if let LspEvent::RecheckFinished = event {
+            // Update our snapshot when global state changes
+            let transaction =
+                ide_transaction_manager.non_committable_transaction(self.inner.state());
+            let new_snapshot = transaction.current_epoch().as_u32() as i32;
```
kinto0: could it be cleaner to make a `Snapshot` impl with `increment` methods so we don't have to store the raw value?

rchiodo: Yeah that works. Then it's doesn't need to know about the epoch value. It just needs to track the event.

pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
rchiodo: Will do

pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
rchiodo: Yeah it seems like it's not working. The update should have caused a refreshevent, no? I'll investigate.

pyrefly/lib/lsp/server.rs
```diff
@@ -224,6 +224,9 @@ pub trait TspInterface {
     /// Get access to the state for creating transactions
     fn state(&self) -> &Arc<State>;
 
+    /// Get access to the recheck queue for async task processing
```
rchiodo, line 230: Needed this to process so that async events are processed in the tsp server (sometimes RecheckEvent is async). Seemed like DidChangeWatchedFiles notification causes an async event.

pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
rchiodo: Thanks. This actually found a bug in the snapshot tracking because RefreshEvent isn't sent on text document change.

pyrefly/lib/test/tsp/tsp_interaction/get_snapshot.rs
rchiodo: Yeah it wasn't working. File watching doesn't actually watch files on disk, it watches the didChangeWatchedFiles notification from lsp. Thanks for pointing out that this test did nothing.

PR #1195, opened 2025-09-27
Fix output stream duplicating

### LINKED ISSUE #1088: vscode "Restart Pyrefly Client" generates multiple output streams and should not
### Describe the Bug

Using the `pyrefly: Restart Pyrefly Client` oprion in vscode results in a brand new "Output" stream for every run, instead of reusing the same one. This is probably not intended behaviour. It is confusing to try and find the output to debug when you have problems.

Here is a sample of my Output window options:

Image

And again after I restart the pyrefly client one more time:

Image

Note that there were 4 instance, and then 5.

vscode extension version `0.32.0`

### Sandbox Link

_No response_

vscode extension version `0.32.0`

mohesham88:
## Description
This PR fixes an issue where using the "Pyrefly: Restart Client" command would create a new Output channel for every restart instead of reusing the existing one. This led to multiple "Pyrefly" channels accumulating in the Output panel, making it confusing for users to track and debug issues.

## Changes
- Added a global `outputChannel` variable to maintain a single instance of the Output channel
- Initialized the output channel during extension activation
- Modified client options to use the shared output channel
- Updated the restart client command to clear and reuse the existing channel instead of creating a new one
- Added proper cleanup in the deactivate function to dispose of the output channel

## Related Issue
closes #1088 - Multiple output channels created when restarting Pyrefly client

### DIFF
--- a/lsp/src/extension.ts
+++ b/lsp/src/extension.ts
@@ -25,6 +25,7 @@ import {PythonExtension} from '@vscode/python-extension';
 
 let client: LanguageClient;
 let statusBarItem: vscode.StatusBarItem;
+let outputChannel: vscode.OutputChannel;
 
 /// Get a setting at the path, or throw an error if it's not set.
 function requireSetting<T>(path: string): T {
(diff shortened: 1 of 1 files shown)


connernilsen:
Hey @mohesham88, thanks for working on this! I'm going to pull it in and test it out. If things look good, I'll merge it in, otherwise, I'll let you know what I find.

connernilsen:
It worked for me. I just need an approval from someone internally, then I'll land.

Issue #1141, opened 2025-09-23
special-case truthyness of known builtins


KotlinIsland:
### Describe the Bug

```py
def f(a: int):
    assert not a
    reveal_type(a)  # int, expected Literal[0]
```

yangdanny97:
I think if it's stuff like int/str it should be OK to do, and I think we already do this for bool. We can't do this for tuples yet because we don't support https://github.com/facebook/pyrefly/issues/273

If anyone wants to take this on, the code pointer is https://github.com/facebook/pyrefly/blob/63f0fc9b973ecdc99b9595e9aa978bfdb43e9f44/pyrefly/lib/alt/narrow.rs#L620

airvzxf:
Hi, I'd like to work on this for Hacktoberfest!

yangdanny97:
@airvzxf assigned, and welcome!

airvzxf:
Hi, again. The pull request [PR 1249](https://github.com/facebook/pyrefly/pull/1249) is merged and closed. What are the next steps to close this issue?

rchen152:
There's nothing else you need to do; there's just a problem with auto-closing issues right now so we have to manually close them. Thanks for implementing this!

PR #1204, opened 2025-09-29
fix: https://github.com/facebook/pyrefly/issues/1141

fatelei:
fix issue #1141

### DIFF
--- a/crates/pyrefly_python/src/sys_info.rs
+++ b/crates/pyrefly_python/src/sys_info.rs
@@ -431,6 +431,48 @@ impl SysInfo {
                 }
                 _ => None,
             },
+            // Special-case truthiness for known builtins and simple calls we can evaluate statically.
+            Expr::Call(ExprCall {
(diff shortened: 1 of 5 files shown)


yangdanny97 (changes requested):
hmm, I don't think this solves the referenced issue.

the issue and code pointer provided are related to narrowing, s.t. if we have `x: int` and `assert not x` we refine the type of x to be `Literal[0]`. This can also be done for `str` (refine to `Literal[""]`), and maybe a few others. I think this already happens for `bool`.

The code you're modifying is related to statically evaluating expressions, which is a separate thing that happens at the binding step to determine whether some branches are statically known to be true or false, like for `sys.version` or `sys.platform` checks.

airvzxf:
It seems that issue #1141 was resolved in [PR 1249](https://github.com/facebook/pyrefly/pull/1249). I believe this pull request can now be closed.

PR #1205, opened 2025-09-29
Add NumPy typing blogpost

MarcoGorelli:


website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,111 @@
+---
+title: Bringing NumPy's type-completeness score to nearly 90%
```
rec, line 2: "How we doubled type completeness in NumPy with a one-line fix" 

I think it will excite more curiosity, and reduce duplication with the `description` field right below.

website/blog/2025-09-29-blog-post-title.md
```diff
@@ -0,0 +1,111 @@
+---
+title: Bringing NumPy's type-completeness score to nearly 90%
+description: We tell the story of how we brought NumPy's type-completeness score from ~33% to nearly 90%
+slug: numpy-type-completeness
+authors: [marcogorelli]
+tags: [typechecking, news]
+hide_table_of_contents: false
+---
+
+[NumPy](https://numpy.org/) is one of the most downloaded packages in the Python ecosystem. Therefore, any incremental improvement can have a large impact on the data science ecosystem. In particular, improvements related to static typing can improve developer experience and help downstream libraries write safer code. We'll tell the story about how we helped bring its type-completeness score to nearly 90% from an initial 33%.
```
rec: "Because [NumPy](https://numpy.org/) is one of the most downloaded packages in the Python ecosystem, even incremental improvements can have a large impact on the data science ecosystem. In particular, enhancing static typing improves developer experience and helps downstream libraries write safer code. [...]"

website/blog/2025-09-29-blog-post-title.md
```diff
@@ -0,0 +1,111 @@
...
+**TL;DR**:
+
+- NumPy's type-completeness score was ~33%.
+- A one-line fix doubled coverage to over 80%.
+- Fully typing MaskedArray pushed the score to nearly 90%.
+- What's left? Top-level numpy.ma functions, more precise overloads, and adding a type-checker to NumPy's CI.
+
+## Wait, what's type completeness?
+
+Modern IDEs use type annotations to help developers by showing them helpful suggestions about which methods may be available on their variables, as well as to facilitate syntax highlighting. [Pyright](https://github.com/microsoft/pyright) is a popular type-checker which, as well as checking for correctness and consistency, can also measure what percentage of a library's public API has type annotations. We call the percentage of fully-typed symbols exported by a library the _type-completeness score_.
```
rec: "Modern IDEs use type annotations to help developers by showing them helpful suggestions and highlighting syntax. [...]"

(If you were going to go into detail, you'd need to list more features than just "method discovery", but I think the shorter statement is better.)

website/blog/2025-09-29-typing-numpy.md
MarcoGorelli: thanks for the suggestion, but this covers more than the one-line fix so I feel like it would be reductive to have that be the title

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,111 @@
...
+description: We tell the story of how we brought NumPy's type-completeness score from ~33% to nearly 90%
+slug: numpy-type-completeness
+authors: [marcogorelli]
+tags: [typechecking, news]
+hide_table_of_contents: false
+---
+
+Because [NumPy](https://numpy.org/) is one of the most downloaded packages in the Python ecosystem, any incremental improvement can have a large impact on the data science ecosystem. In particular, improvements related to static typing can improve developer experience and help downstream libraries write safer code. We'll tell the story about how we helped bring its type-completeness score to nearly 90% from an initial 33%.
```
javabster:
```suggestion
Because [NumPy](https://numpy.org/) is one of the most downloaded packages in the Python ecosystem, any incremental improvement can have a large impact on the data science ecosystem. In particular, improvements related to static typing can improve developer experience and help downstream libraries write safer code. We'll tell the story about how we helped bring its type-completeness score to nearly 90% from an initial 33%.

<!-- truncate -->
```

we need to add this truncate line so that on the blog landing page we only show a short preview, not the whole blog

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,111 @@
...
+
+**TL;DR**:
+
+- NumPy's type-completeness score was ~33%.
+- A one-line fix doubled coverage to over 80%.
+- Fully typing MaskedArray pushed the score to nearly 90%.
+- What's left? Top-level numpy.ma functions, more precise overloads, and adding a type-checker to NumPy's CI.
```
javabster:
```suggestion
- What's left? Top-level `numpy.ma` functions, more precise overloads, and adding a type-checker to NumPy's CI.
```

website/blog/authors.yml
```diff
@@ -10,6 +10,10 @@ kylei:
   name: Kyle Into
   url: https://www.kylei.dev/
 
+marcogorelli:
+  name: Marco Gorelli
```
javabster: I think it would be good to make it clearer here what your affiliation is, something like `Marco Gorelli, Quansight`

website/blog/2025-09-29-typing-numpy.md
javabster: Do we need to be a bit clearer about who "we" is referring to here? As this is going on the pyrefly website I'm wondering if some folks might assume this is purely the pyrefly team that has been doing this work. Maybe somewhere in the intro it should be more clear that "we" means "Quansight with support from Meta's pyrefly team" or similar. Curious what others think though. maybe @lolpack cares to weigh in?

javabster (commented):
Thanks Marco! Left a few minor comments

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+
+But...we pulled through it, and thanks to some very timely and constructive reviews from the amazing [Joren Hammudoglu](https://github.com/jorenham), MaskedArray is now reported as 100% type-complete! As for the overall type-completeness, that's now at 88%. So...what's left?
+
+## What's missing from NumPy?
+
+In the MaskedArray module, there's still some untyped top-level functions, such as `numpy.ma.count`. Overloads could be made more precise and shape-preserving (e.g. if the input is 2D, then make sure to preserve this fact in the output where possible). There's some missing defaults in the stubs. There's no shortage of work here.
+
+The biggest missing bit, however, is the elephant in the room: NumPy doesn't run a type-checker over its codebase in CI. It has some typing tests, sure, but that's different from running a type-checker. If any motivated reader is interested in making a significant open source contribution, then getting NumPy's typing into a state such that a type checker (and possibly even [stubtest](https://mypy.readthedocs.io/en/stable/stubtest.html)) can be run over it could be a great use of your time.
```
jorenham, line 108: Note that I've done this already in [numtype](https://github.com/numpy/numtype), where pyright, mypy, and stubtest report 0 errors in strict mode. But the only way I was able to get there, was by making several backwards-incompatible changes. So doing the same in numpy would be rather problematic. That's actually one of the main reasons for having numtype in the first place.

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+
+- `ma.count()` returns an integer.
+- `ma.count(axis=0)` returns an array.
+- `ma.count(keepdims=True)` also returns an array, of the same shape as the input array.
+
+We can type this by having a different overload of each of these different cases. This is a relatively simple example, but there are others where the number of necessary overloads was as high as 9! This isn't something which is easy to automate, and requires careful reading of the documentation and of the source code. It's a non-trivial amount of work.
+
+But...we pulled through it, and thanks to some very timely and constructive reviews from the amazing [Joren Hammudoglu](https://github.com/jorenham), MaskedArray is now reported as 100% type-complete! As for the overall type-completeness, that's now at 88%. So...what's left?
```
jorenham, line 102: ❤️

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+would have a 50% type-completeness score, because:
+
+- `foo` is partially unknown, as it's missing a return annotation.
+- `bar` is type-complete.
+
+By changing the `foo` signature to be `def foo(a: int) -> None:`, the type-completeness score would jump to 100%. The more type-complete a library is, the more helpful the suggestions an IDE can show to the user.
+
+Note that type-completeness only measures how much of the public API is covered by types. If you want to verify that those types are correct and self-consistent, you'll also need to run a type checker. The most used type checkers currently are [mypy](https://github.com/python/mypy) and Pyright, but [Pyrefly](https://github.com/facebook/pyrefly) and [ty](https://github.com/astral-sh/ty) are also attracting a lot of attention due to their impressive performance characteristics (note however that neither yet describes itself as production-ready, so temper your expectations accordingly if you try them out!).
```
jorenham: > Note that type-completeness only measures how much of the public API is covered by types.

The part of the public API that Pyright is aware of, that is. Because I don't think that Pyright is "aware" of any C extension modules that (in some way) could be part of the public API.

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+So, `ndarray` was reported as "partially unknown", but 98% of its methods had known types. It shouldn't be much effort to bring that number to 100%! In fact, all it took was a [one-line change to fix a typo in a type annotation](https://github.com/numpy/numpy/pull/28908): 
+
+```diff
+- def setfield(self, /, val: ArrayLike, dtype: DTypeLike, offset: CanIndex = 0) -> None: ...
++ def setfield(self, /, val: ArrayLike, dtype: DTypeLike, offset: SupportsIndex = 0) -> None: ...
+```
+
+That's it! `CanIndex` was an unknown symbol and was probably mistyped, and replacing it with the correct `SupportsIndex` one brought NumPy's overall type-completeness to over 80%! We then started examining other NumPy classes to see if there was anywhere else where we could make an impact, and hopefully a much larger one.
```
jorenham, line 79: So 20% of the type completeness score can depend on a single parameter type of an arbitrary `ndarray` method 🤔. Some readers might conclude that the type-completeness score isn't really all that useful as a metric here ;)

But don't listen to me, because I'm clearly biased: I'm to blame for this bug, and about to blog about it 🤭

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Enter MaskedArray
+
+When we looked at the percentage of typed symbols from the MaskedArray class, we noticed something interesting:
+
+```python
+In [14]: np.mean([x['isTypeKnown'] for x in exported if x['name'].startswith('numpy.ma.core.MaskedArray.')])
+Out[14]: np.float64(0.2)
```
jorenham: Does ipython show-up properly in a `python` code  block? Otherwise you could try <code>```pycon</code> for the vanilla `>>> ` python console stuff

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Some objects, such as `DTypeLike`, were reported to be "partially unknown", even though they had type annotations.
+- The type-completeness report included test modules such as `numpy.tests.test_matlib`.
+
+The first issue was caused by an import from the standard library `decimal` module, which itself is partially untyped. Given that this is outside of NumPy's direct control, we decided to exclude it from the coverage report by using `--ignoreexternal`, [as suggested by Eric Traut](https://github.com/microsoft/pyright/discussions/9911).
+
+For the second issue, PyRight gives us an option to export the coverage report to json (`--outputjson`). We could then parse the json and exclude `numpy.tests`. Given that NumPy users wouldn't ordinarily interact with NumPy's internal test suite but that PyRight considers it public, we decided that it made sense for us to exclude tests in order to focus our efforts on what would make the biggest user-facing impact.
```
jorenham:
```suggestion
For the second issue, Pyright gives us an option to export the coverage report to json (`--outputjson`). We could then parse the json and exclude `numpy.tests`. Given that NumPy users wouldn't ordinarily interact with NumPy's internal test suite but that Pyright considers it public, we decided that it made sense for us to exclude tests in order to focus our efforts on what would make the biggest user-facing impact.
```

website/blog/2025-09-29-typing-numpy.md
```diff
@@ -0,0 +1,113 @@
...
+
+Once we'd addressed the two steps above, the baseline type-completeness score became 33%. This was our starting point. We could then focus our efforts on the remaining 67%!
+
+## The one-line change which doubled type-completeness
+
+Pyright's report includes classes, methods, functions, type aliases, and more. A lot of scientific Python code centres around some central classes such as `ndarray`. `ndarray` was reported as "partially unknown", but eye-balling the exported symbols related to that class revealed something interesting:
```
jorenham:
```suggestion
Pyright's report includes classes, methods, functions, type aliases, and more. A lot of scientific Python code centres around some central classes such as `numpy.ndarray`. `ndarray` was reported as "partially unknown", but eye-balling the exported symbols related to that class revealed something interesting:
```

jorenham (commented):
Great blogpost!

website/blog/2025-09-29-typing-numpy.md
jorenham: I like every title that isn't "How @jorenham halved type completeness in NumPy with a simple typo" 🤣

website/blog/2025-09-29-typing-numpy.md
MarcoGorelli: true, the metric is sadly a bit too binary (`isTypeKnown` can only be `False` or `True`, and a single parameter type can be all it takes to flip the switch). maybe we need to define a new measure later 😄

jorenham: > maybe we need to define a new measure later 😄

Hn yea, perhaps that's not such a bad idea 🤔

jorenham:
Oh I just remembered that it might be nice to also mention this comment: https://github.com/microsoft/pyright/issues/10414#issuecomment-2853219421

javabster:
Ah sorry @jorenham just saw your comment and we've merged already. If you want to open another PR to update the blog with that extra piece please go ahead, if we can merge it before EOD today it should make it in before the website updates overnight. Also not a big deal if we don't merge in time, we can still update after the blog is published if need be

website/blog/2025-09-29-typing-numpy.md
jorenham: It turns out out was wrong: I managed to get stubtest and mypy running on NumPy's stubs without breaking backwards-compatibility, after all. So sorry about that.

PR #1215, opened 2025-09-30
collect type params at bindings phase for legacy explicit type alias

### LINKED ISSUE #1162: A `TypeAlias` of a `TypeAliasType` loses parameter order
### Describe the Bug

```py
from typing import TypeAlias, TypeAliasType, TypeVar

T1 = TypeVar("T1")
T2 = TypeVar("T2", bound=str | bytes)

Spam1 = TypeAliasType("Spam1", T2 | type[T1], type_params=(T1, T2))
Spam2: TypeAlias = Spam1[T1, T2]

x1: Spam1[int, str] = int  # ✔️ 
x2: Spam2[int, str] = int  # ❌ 
```

The reported errors show that the typar order "gets lost" in the `Spam2` alias:

```
ERROR sandbox.py:10:5-20: Type `int` is not assignable to upper bound `bytes | str` of type variable `T2` [[bad-specialization](https://pyrefly.org/en/docs/error-kinds/#bad-specialization)]
ERROR sandbox.py:10:23-26: `type[int]` is not assignable to `int | type[str]` [[bad-assignment](https://pyrefly.org/en/docs/error-kinds/#bad-assignment)]
```

---

For context: This pattern is used in [optype](https://github.com/jorenham/optype), where e.g. `optype.numpy.ToArray1D` ([src](https://github.com/jorenham/optype/blob/bee20fc36be20b6e30af76e332f27aa881fc530b/optype/numpy/_to.py#L262C1-L262C10)) is a `TypeAlias` to a `TypeAliasType` called `_To1D2` ([src](https://github.com/jorenham/optype/blob/bee20fc36be20b6e30af76e332f27aa881fc530b/optype/numpy/_to.py#L154)). This leads to a bunch of false positives and incorrect inferences when running pyrefly on [`scipy-stubs`](https://github.com/scipy/scipy-stubs), which heavily relies on these `optype.numpy` types.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjACCUCKjgAaQcLES4Q4jBlKYANVSUAOul0CAjHQC8s5ZsoAKbSEM2AlPoBMJsxq3XbTmzLwBXLGM4Bko6AB86bFIGeEc9dABlYlQaI1M1eUk1TySUgx9BFwimZQBtQwBdGRKYAH1kyhS4Y0tDVSd7ONyaJ0Q3TLhXboNyg3aK3V18Az7h0tYGGWDKCtcFqd66bqd59EW6ZdXTdfQQKRA-Bmg4EnJEEABiOgBVK-EmOjAAgGMr3HQ4JMsDAwJ9eDRUAxaug-DRsDArPg%2Bgt7HQALQAPgOIUQujo%2BLolBgDD8lHQnxsADlYfDKH1gPgAL42XQgRnnVC-CAANxgADFoDAKGgsHgiGQ2UA&version=3.13

_No response_

yangdanny97:
Summary:
fixes https://github.com/facebook/pyrefly/issues/1162

This is related to our (problematic) handling of legacy type aliases as well as our internal representation of type aliases.

For generic legacy type aliases, we traverse the resolved type of the RHS to determine the type parameters. However, when the RHS is a specialization of another type alias, the type we see is the aliased type because `untype()` eagerly resolves aliases. The order of the type arguments during specialization may not match the order that they appear in the aliased type, resulting in this bug.

https://github.com/facebook/pyrefly/issues/1162#issuecomment-3353329339 explains a bit more.

The solution in this diff is to use the legacy type parameter collector (also used in function and class definitions) to collect type parameters at the bindings phase when we know for sure we're dealing with a type alias (the assignment is annotated with TypeAlias or the RHS is a UnionType, for example). If we're not sure, then we fall back to the previous behavior of collecting type params from the resolved type.

One long term fix is to have a type that represents a specialized type alias, like we do for classes using`ClassType`.

Differential Revision: D83597568

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -4980,8 +4980,8 @@
     {
       "code": -2,
       "column": 18,
-      "concise_description": "`ParamSpec[P]` is not allowed in this context",
-      "description": "`ParamSpec[P]` is not allowed in this context",
+      "concise_description": "`ParamSpec` is not allowed in this context",
+      "description": "`ParamSpec` is not allowed in this context",
       "line": 15,
       "name": "invalid-annotation",
       "stop_column": 19,
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -6,6 +6,7 @@
  */
 
 use std::iter;
+use std::ops::Deref;
 use std::sync::Arc;
 
 use dupe::Dupe;
@@ -73,6 +74,7 @@ use crate::binding::binding::FunctionStubOrImpl;
 use crate::binding::binding::IsAsync;
 use crate::binding::binding::Key;
 use crate::binding::binding::KeyExport;
+use crate::binding::binding::KeyLegacyTypeParam;
 use crate::binding::binding::KeyUndecoratedFunction;
 use crate::binding::binding::LastStmt;
 use crate::binding::binding::LinkedKey;
@@ -92,6 +94,7 @@ use crate::error::context::ErrorInfo;
 use crate::error::context::TypeCheckContext;
 use crate::error::context::TypeCheckKind;
 use crate::error::style::ErrorStyle;
+use crate::graph::index::Idx;
 use crate::solver::solver::SubsetError;
 use crate::types::annotation::Annotation;
 use crate::types::annotation::Qualifier;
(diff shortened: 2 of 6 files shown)


stroxler (approved):

PR #1233, opened 2025-10-02
improve hover display for overloads

### LINKED ISSUE #969: format BoundMethod parameters on new lines
### Describe the Bug

I guess it's because the `Overload[ ... ]`

Image

ty solution https://github.com/astral-sh/ruff/pull/19979#issuecomment-3198702378

### Sandbox Link

_No response_

_No response_

yangdanny97:
Summary:
Changes for hover:
- show each overload on a new line (if overloads have multiple parameters, each parameter is also on a new line)
- add `def ...` prefix for syntax highlighting, like we do for named functions and bound methods
- add `overload` before the first signature, to mark it as a group of overloads. to reduce repetition, we don't add it to any subsequent signatures

fixes https://github.com/facebook/pyrefly/issues/969

Differential Revision: D83771464

### DIFF
--- a/crates/pyrefly_types/src/display.rs
+++ b/crates/pyrefly_types/src/display.rs
@@ -311,15 +311,28 @@ impl<'a> TypeDisplayContext<'a> {
                 }
             }
             Type::Overload(overload) => {
-                write!(
-                    f,
-                    "Overload[{}",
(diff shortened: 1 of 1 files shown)


stroxler (approved):

PR #1236, opened 2025-10-03
Fix sandbox browser history pollution on every keystroke   (#1198)

### LINKED ISSUE #1198: every keystroke in the sandbox adds an entry in the browser history

AryanBagade:
## Summary
Closes #1198

This PR prevents browser history pollution caused by updating the URL on every keystroke in the sandbox editor.

## Changes
- Removed `updateURL()` calls from the editor's `onChange` handler
- Removed `updateURL()` call from Reset button
- Modified Share button to call `updateURL()` before copying
  URL to clipboard

## Behavior
URL only updates when user explicitly clicks the Share button

## Testing
- ✅ All Jest tests pass 
- ✅ Manually tested in browser - typing doesn't update URL
- ✅ Clicking Share button updates URL and copies to clipboard
- ✅ Shared URLs correctly restore sandbox state

### DIFF
--- a/website/src/sandbox/Sandbox.tsx
+++ b/website/src/sandbox/Sandbox.tsx
@@ -880,20 +880,7 @@ function getPyreflyEditor(
                 defaultValue={codeSample}
                 defaultLanguage="python"
                 theme={editorTheme}
-                onChange={(value) => {
-                    forceRecheck();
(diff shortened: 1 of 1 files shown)


yangdanny97:
woohoo, thanks!

Issue #48, opened 2025-03-27
special-case join for LiteralString


yangdanny97:
https://typing.python.org/en/latest/spec/literal.html#inferring-literalstring

Adist319:
I can work on this!

PR #1241, opened 2025-10-03
Add LiteralString support for format(), join(), and replace() methods

Adist319:
Add LiteralString return types for `format()`, `join()`, and `replace()` when all arguments are literal strings, per PEP 675.

Addresses #48

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -692,13 +692,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
 
                 // Only promote for user-defined methods
                 if let FunctionKind::Def(func_id) = &metadata.kind {
-                    result = self.promote_literal_string_method_return(
+                    if self.should_return_literal_string(
                         &func_id.func,
                         &obj_for_helper,
                         args,
                         keywords,
-                        result,
-                    );
+                        &result,
+                    ) {
+                        result = Type::LiteralString;
+                    } else if result.is_literal_string() {
+                        // Demote from LiteralString to str (e.g., when starred args present)
+                        result = Type::ClassType(self.stdlib.str().clone());
+                    }
                 }
 
                 result
@@ -761,13 +766,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
 
                 // Only promote for user-defined methods
                 if let FunctionKind::Def(func_id) = &meta.kind {
-                    result = self.promote_literal_string_method_return(
+                    if self.should_return_literal_string(
                         &func_id.func,
                         &obj_for_helper,
                         args,
                         keywords,
-                        result,
-                    );
+                        &result,
+                    ) {
+                        result = Type::LiteralString;
+                    } else if result.is_literal_string() {
+                        // Demote from LiteralString to str (e.g., when starred args present)
+                        result = Type::ClassType(self.stdlib.str().clone());
+                    }
                 }
 
                 result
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/call.rs
```diff
@@ -645,20 +648,36 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                    errors,
+                    errors,
+                    context,
+                    hint,
+                    ctor_targs,
+                );
+
+                // Only promote for user-defined methods
+                if let FunctionKind::Def(func_id) = &metadata.kind {
+                    result = self.promote_literal_string_method_return(
```
yangdanny97: At the risk of being pedantic, I think converting `str` -> `LiteralString` is technically a narrowing/demotion, since LiteralString is narrower than str.

pyrefly/lib/alt/call.rs
```diff
@@ -1101,4 +1133,186 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        }
+        if !matches!(return_ty, Type::ClassType(ref cls) if cls.is_builtin("str")) {
+            return return_ty;
+        }
+
+        // Error collector for re-inferring (never extended to main errors)
+        let error_swallower = self.error_collector();
+
+        match method_name.as_str() {
+            "format" => {
```
yangdanny97: My hunch is that maybe we don’t need to be super specific here with the argument count, and we can just have a single block handle format/replace that checks all the args/kwargs are LiteralString.

Since LiteralString is special-cased in the typechecker, I don’t think there's a risk of conflating these methods with different methods w/ the same name.

It might be good to put the starred arg/kwarg checks down here too, jsut the arg/kwarg checks are grouped together.

pyrefly/lib/alt/call.rs
```diff
@@ -712,8 +732,20 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     context,
                     hint,
                     ctor_targs,
-                )
-                .0
+                );
+
+                // Only promote for user-defined methods
+                if let FunctionKind::Def(func_id) = &meta.kind {
+                    result = self.promote_literal_string_method_return(
```
yangdanny97: I think I'd prefer it if this was a `should_return_literal_string` that returns a bool, and then we did the result = `Type::LiteralString` here.

`promote_literal_string_method_return` is a bit oddly named since it doesn't always change the type

yangdanny97 (changes requested):
Thanks for contributing this change! The overall approach & test cases look good, back to you with a few stylistic/code structure things

Adist319:
Thanks for the feedback @yangdanny97! Let me know if there's anything else I can improve on.

yangdanny97:
Thanks for making these changes, the code looks good! Although...

As I was reviewing I realized that we could avoid this special-casing by having the attribute lookup bind methods to `LiteralString` instead of `str`. Since `str` contains overloads for `LiteralString` operations (see https://github.com/python/typeshed/blob/9c26ab12e280c3703d0bd27fe5074294f44703cb/stdlib/builtins.pyi#L524 for example), fixing the attribute lookup will cause overload selection to automatically give the correct behavior.

So it seems like the special-casing wasn't necessary in the end. Really sorry about this! I should have realized it sooner and saved you the effort.

This PR will still be merged, I've amended it internally with the changes described above. Fwiw the test cases you wrote were super helpful for verifying that the behavior is equivalent, so thanks for that and thanks for prototyping this solution!

yangdanny97 (approved):

Adist319:
@yangdanny97 All good! I learned something new too about str overloads. Thanks for reviewing this PR!

ndmitchell (approved):

PR #1242, opened 2025-10-03
Fix namespace package finding

connernilsen:
Summary:
This diff mostly includes updates to an interface and the affected tests.

This diff stores found namespace results, and returns them  yfrom the end of `find_import_filtered` if no higher-priority `FindResult` is found sooner. This is because results in `search-path`, `typeshed`, and `site-package-path` can override a previously found namespace package if they contain a file or regular package match, and we shouldn't just early return from a single find component if we find a namespace result.

Differential Revision: D83847654


stroxler (approved):

Issue #1227, opened 2025-10-02
Pydantic: setting `model_config` to a plain dict doesn't work


rchen152:
### Describe the Bug

```
class Model1(BaseModel, extra='forbid'):
    a: int

reveal_type(Model1.__init__)  # (self: Model1, *, a: Any) -> None


class Model2(BaseModel):
    a: int

    model_config = ConfigDict(extra='forbid')

reveal_type(Model2.__init__)  # (self: Model2, *, a: Any) -> None


class Model3(BaseModel):
    a: int

    model_config = {'extra': 'forbid'}


# ❌ not inferred as ConfigDict
reveal_type(Model3.__init__)  # (self: Model2, *, a: Any, **Unknown) -> None
```

PR #1247, opened 2025-10-04
Allow using dict expressions for Pydantic config

Viicos:
Fixes https://github.com/facebook/pyrefly/issues/1227.

To avoid duplicating logic, I abstracted the logic away, which simplifies the configuration extraction logic.

cc @rchen152, ~~I just need to fix conflicts with https://github.com/facebook/pyrefly/commit/7f376a689f30f0760310bc2c26aafb65be6357d2~~ done.

### DIFF
--- a/pyrefly/lib/test/pydantic/config.rs
+++ b/pyrefly/lib/test/pydantic/config.rs
@@ -37,6 +37,36 @@ m.x = 10 # E: Cannot set field `x`
 );
 
 pydantic_testcase!(
+    test_model_config_dict_func,
+    r#"
+from pydantic import BaseModel
+
+class Model(BaseModel):
+    model_config = dict(frozen=True)
+
+    x: int = 42
+
+m = Model()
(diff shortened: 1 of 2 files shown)


pyrefly/lib/test/pydantic/config.rs
```diff
@@ -36,6 +36,36 @@ m.x = 10 # E: Cannot set field `x`
 "#,
 );
 
+pydantic_testcase!(
+    test_model_config_dict_func,
+    r#"
+from pydantic import BaseModel
+
+class Model(BaseModel):
+    model_config = dict(frozen=True)
```
rchen152: Looks like this test is failing due to some missing TypedDict functionality in Pyrefly. I created https://github.com/facebook/pyrefly/issues/1248. Could you add an `# E:` comment here and link to that issue? I'll look into fixing it on Monday.

ndmitchell (approved):

PR #1259, opened 2025-10-07
minimal provide-type impl using get_type_at

### LINKED ISSUE #1181: experiment with a `types/provide-type` lsp method to return a fully qualified type annotation
similar to tsp, we can support pyrefly as a type provider/type engine in PyCharm

```py
def f[T: int](x: T) -> str: ...
f
```

it would accept `textDocument` and a `ranges` (ranges for multiple expressions) for the expression `f`, and returns a list of strings like `"[T: builtins.int](x: T) -> builtins.str"`

everything needs to be fully qualified, even `Literal` and `Any`

ideally typevars would include their scope somehow if they are not included in the definition `T@mod.SomeClass`

yangdanny97:
Summary:
closes https://github.com/facebook/pyrefly/issues/1181

This is just a hover that sets `always_display_module_name` to true. That functionality is used in the query module, and I assume that it's adequate for getting fully qualified names. In any case, if it has issues we can iterate on this prototype.

Differential Revision: D84072683

### DIFF
--- a/pyrefly/lib/lsp/features/mod.rs
+++ b/pyrefly/lib/lsp/features/mod.rs
@@ -6,3 +6,4 @@
  */
 
 pub mod hover;
+pub mod provide_type;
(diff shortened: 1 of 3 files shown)


kinto0 (approved):

PR #1260, opened 2025-10-07
display qualified names from typing module

yangdanny97:
Summary:
We qualify classes, but not things that have their own variant in the Type enum like Any, Literal, Never, etc.

This affects the query module, so probably needs a review from folks using that to make sure it's OK.

Differential Revision: D84072684


grievejia (approved):

Issue #890, opened 2025-08-12
Hover over type ignores should show the error message


ndmitchell:
### Describe the Bug

Given:

```
a: int = "test" # pyrefly: ignore
```

I expect this to print out no errors. But in the IDE, it would be nice to see the error. One way to achieve that would be that hovering over the `ignore` could show the error that was ignored.

Adist319:
I can try this out!

yangdanny97:
assigned, thanks! 

i have some vague ideas for this, i think the first part is detecting whether you’re hovering over a suppression comment, and the second part is matching it with a suppressed error in the ErrorCollector for that module

Adist319:
Yup that's what I was thinking as well. And maybe some helper functions:
- is_hovering_over_ignore_keyword() helper to detect position in ignore comment
- add find_ignore_keyword_in_comment() to find the keyword's TextRange within a comment

And the main file changes should be in in pyrefly/lib/lsp/features/hover.rs, tests in pyrefly/lib/test/lsp/lsp_interaction/hover.rs

yangdanny97:
I think if the comment is recognized as an ignore comment, we can just have the error appear on the whole comment. Saves you a bit of work slicing the string. but not a huge deal either way.

yangdanny97:
Hm, I got around to trying this for myself, but I don't think it works for me on the latest build. Did it work for you during manual testing @Adist319 ?

PR #1290, opened 2025-10-14
Hovering over type ignore shows error message

Adist319:
Addresses #890. Implements hover functionality that shows suppressed errors when hovering over # pyrefly: ignore or # type: ignore comments.

### DIFF
--- a/crates/pyrefly_python/src/ignore.rs
+++ b/crates/pyrefly_python/src/ignore.rs
@@ -28,8 +28,53 @@
 
 use dupe::Dupe;
 use pyrefly_util::lined_buffer::LineNumber;
+use ruff_text_size::TextRange;
+use ruff_text_size::TextSize;
 use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
+
+/// Finds the byte offset of the first '#' character that starts a comment.
+/// Returns None if no comment is found or if all '#' are inside strings.
+/// Handles escape sequences and single/double quotes.
+///
+/// This is string-aware parsing that avoids treating '#' inside strings as comments.
+/// For example: `x = "hello # world"  # real comment` correctly identifies the second '#'.
+fn find_comment_start_in_line(line: &str) -> Option<usize> {
+    let mut in_string = false;
+    let mut string_char = '\0';
+    let mut escaped = false;
+    let mut byte_offset = 0;
+
+    for ch in line.chars() {
+        if escaped {
+            escaped = false;
+            byte_offset += ch.len_utf8();
+            continue;
+        }
+
+        if ch == '\\' && in_string {
+            escaped = true;
+            byte_offset += ch.len_utf8();
+            continue;
+        }
+
+        if ch == '"' || ch == '\'' {
+            if !in_string {
+                in_string = true;
+                string_char = ch;
+            } else if ch == string_char {
+                in_string = false;
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


crates/pyrefly_python/src/ignore.rs
kinto0, line 1: I think it would be easier if we added something like a comment_range to the Suppression struct [here](https://github.com/facebook/pyrefly/pull/1290/files#diff-2da5b3df3e051b29e92c2c8cf61462ddb86e1c6c18298990b1dd3a1e033e1549R126)

then we can keep the parsing logic within `parse_ignores`

pyrefly/lib/lsp/features/hover.rs
```diff
@@ -17,10 +17,150 @@ use pyrefly_types::types::Type;
...
 use crate::state::lsp::FindDefinitionItemWithDocstring;
 use crate::state::lsp::FindPreference;
 use crate::state::state::Transaction;
 
+/// Finds the byte offset of the first '#' character that starts a comment.
+/// Returns None if no comment is found or if all '#' are inside strings.
+/// Handles escape sequences and single/double quotes.
+///
+/// TODO: could handle triple-quoted strings on the same line in the future?
+fn find_comment_start_in_line(line: &str) -> Option<usize> {
```
kinto0: we don't need this function if we store comment range in ignores

pyrefly/lib/lsp/features/hover.rs
```diff
@@ -17,10 +17,150 @@ use pyrefly_types::types::Type;
...
+        }
+
+        byte_offset += ch.len_utf8();
+    }
+    None
+}
+
+/// Detects if the given position is hovering over an ignore comment.
+/// Returns the comment range and the line number where suppressions apply.
+fn find_ignore_comment_at_position(
```
kinto0: this will be much simpler: 
```
    let suppression_line = module.ignore().get_suppression_target_line(hover_line)?;
    
    let comment_range = module.ignore().get_comment_range(hover_line)?; # new API in ignore implement this as "check current line, then check next line"
    if !comment_range.contains(position) {
        return None; // Hovering over code, not comment
    }
```

pyrefly/lib/test/lsp/hover.rs
```diff
@@ -93,3 +93,101 @@ from lib import foo_renamed
...
+    assert!(
+        report.contains("int"),
+        "Expected type hover, got: {}",
+        report
+    );
+    assert!(
+        !report.contains("Suppressed"),
+        "Should not show suppressed error when hovering over code"
+    );
+}
```
kinto0, line 193: what about a test for ignores on different lines?

Adist319:
@kinto0 Thanks for the feedback! Refactored to use the comment range suppression API

  - Moved comment parsing logic into parse_ignores() so ranges are computed once during initialization
  - Added TextRange field to Suppression struct to store pre-computed comment positions
  - Simplified hover.rs by using the new get_comment_range() API
  - Fixed pending suppression handling to properly clear on empty lines
  - I added a unit/smoke test for above-line ignores in ignore.rs. I was having issues adding integration tests 
  for above-line ignores in the hover test file due to the cursor marker (`#^`) interfering with the suppression 
  logic. The marker line sits between the comment and code, breaking the above-line detection. Let me know if this is fine.

yangdanny97 (approved):

PR #1308, opened 2025-10-15
Error on defining attributes via assignment to self on a Protocol

### LINKED ISSUE #1305: Error on defining attributes via assignment to `self` on a `Protocol`
### Describe the Bug

```
from typing import Protocol

class Template(Protocol):
    name: str
    value: int = 0

    def method(self) -> None:
        self.name = "name"  # OK, this was defined on the class
        self.temp: list[int] = []  # Not OK, this was not defined on the class
```

1. In `calculate_class_field`, if the definition type is `ClassFieldDefinition::AssignedInBody`, we should check if the class is a `Protocol` (check `is_protocol()` in the `ClassMetadata`)
2. Add test cases to `pyrefly/lib/test/protocol.rs`. Include a test case for when a class extends a protocol but does not extend `Protocol` directly - classes that don't directly have `Protocol` in their base class list are concrete and not have this restriction.
3. Run test.py and commit any generated files

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AFag1wBjXFAA66aaKio4cOgBUY3eQxgAKIbhHioASkTS6ZuulQ0YiOnAaVT5gG6ooAVxud0-ALx0ABmknM0wYMDprBgALXEwtOBgoMEM6AFoAPjoAOVx0GxDzM0TkwktrOn9JEHKYarMAYjoAeQBpABpGaIglAHcFOjDIfMw6PK6YOjkFOEKikrBCTW5bKB6GAG1WBgBdSroNvbom3P42zpieun6ldD1B8NYYUfGYyenFEHaQMkpwqFIS1oUAoTQEpD%2BYABdgwOAIUzykDY7koqAYEDyhGkTQAyjBJtEGAxiHBEAB6Mm-f6A3hsMkwdBkzBiOBk8ToJEotEYxl0MC8OioVzQVDYWAIjkQZGo9HjXDEWXoOBY9BkGJ5NLOGCUOA8-bVADMhAAjAAmarSEAAX2%2BqFE6K1ADFoDAKGgsHgiGRrUA

_No response_

asukaminato0721:
fix #1305

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -137,8 +137,6 @@ pub enum ErrorKind {
     /// do not recognize as always executing (we recognize constructors and some test setup
     /// methods).
     ImplicitlyDefinedAttribute,
-    /// A protocol attribute was first defined inside a method instead of the class body.
-    ProtocolImplicitlyDefinedAttribute,
     /// An error related to the import machinery.
     /// e.g. failed to import a module.
     ImportError,
@@ -207,6 +205,8 @@ pub enum ErrorKind {
     NotIterable,
     /// An error related to parsing or syntax.
     ParseError,
+    /// A protocol attribute was first defined inside a method instead of the class body.
+    ProtocolImplicitlyDefinedAttribute,
     /// The attribute exists but cannot be modified.
     ReadOnly,
     /// Warning when casting a value to a type it is already compatible with.


pyrefly/lib/test/protocol.rs
```diff
@@ -364,7 +364,23 @@ from typing import Protocol
 class P(Protocol):
     x: int
     def f(self, y: str):
-        self.y = y  # E: Attribute `y` is implicitly defined by assignment in method `f`, which is not a constructor
+        self.y = y  # E: Protocol `P` must declare attribute `y` in the class body; assignment in method `f` does not
+    "#,
+);
+
+testcase!(
+    bug = "pyrefly#1305 Ensure protocols reject attributes first defined inside methods without requiring configuration toggles.",
```
yangdanny97, line 372: Note for future reference, our internal convention is that `bug=` indicates the test is passing but the current behavior is wrong. Fixing the behavior will make the test fail, so that we remember to update the test and remove the `bug=` marker.

I'll patch this internally

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1373,14 +1369,31 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
             }
-            if !defined_in_parent {
-                self.error(
-                errors,
-                range,
-                ErrorInfo::Kind(ErrorKind::ImplicitlyDefinedAttribute,
-                ),
-                format!("Attribute `{}` is implicitly defined by assignment in method `{method_name}`, which is not a constructor", &name),
-            );
+            if !defined_in_parent && !method.recognized_attribute_defining_method {
```
yangdanny97, line 1372: some of the conditions here are a bit funky - not matching on `recognized_attribute_defining_method: false`  is correct, because Protocols should not define attributes even in `__init__`, but the way they are checked here means that the behavior is the same as before.

I'll patch this internally

stroxler (approved):

PR #1313, opened 2025-10-16
Error on NamedTuple field name starting with underscore

### LINKED ISSUE #1304: Error on NamedTuple field name starting with underscore
### Describe the Bug

```
from typing import *
class Point(NamedTuple):
    x: int
    _y: int  # illegal field name
```

This has several parts.

1. In `solve_class_metadata`, check the binding (`self.bindings().get(...)`) of the class key to see if it's a functional class `BindingClass::FunctionalClassDef`. If so, plumb the data into `class_metadata_of`.
2. in `class_metadata_of`, once the `NamedTupleMetadata` is calculated, if the definition was NOT functional, check each field in the `NamedTupleMetadata` and if the name starts with an underscore, emit an error at the position of that field, which can be retrieved with `cls.field_decl_range`.
3. Add test cases to `pyrefly/lib/tests/named_tuple.rs`, run test.py and commit any generated changes

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroAxlFRw4dAAq5WDABQA5VDRiYAKgFdisAJSIRdE3XyJO6BsdMB9UhYUmAxJyiw2qKHUgwomOnQ1GBAAGhAyShgwKFJCBlooClcZUiiY0jo0LDx8OlFcdEg2bUpUBghCwhFXAGUYGDoACwYGYjhEAHpOyOjYwl42Tph0TsxcUThOgqKIErKKws6fXjpUADdUaFRsWHzC4tLyyvQ6XGJF9Dhq9DIGJsKAWnWYSjgTugBeOiEQAGZCABGABMvxEIAAvuFUKIKi8AGLQEKIEDZHAEEjkCFAA

_No response_

asukaminato0721:
fix #1304

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1085,6 +1085,23 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         };
         let metadata = self.get_metadata_for_class(class);
 
+        if metadata.named_tuple_metadata().is_some()
+            && class.is_field_annotated(name)
+            && name.as_str().starts_with('_')
+            && matches!(
+                field_definition,
+                ClassFieldDefinition::DeclaredByAnnotation { .. }
+                    | ClassFieldDefinition::AssignedInBody { .. }
+            )
+        {
+            self.error(
+                errors,
+                range,
+                ErrorInfo::Kind(ErrorKind::BadClassDefinition),
+                format!("NamedTuple field name may not start with an underscore: `{name}`"),
+            );
+        }
+
         let magically_initialized = {
             // We consider fields to be always-initialized if it's defined within stub files.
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1085,6 +1085,23 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         };
         let metadata = self.get_metadata_for_class(class);
 
+        if metadata.named_tuple_metadata().is_some()
+            && class.is_field_annotated(name)
```
yangdanny97, line 1089: `NamedTupleMetadata` contains a set of which fields count as elements, when that is constructed we already check these conditions, so I think we can get away with just checking that the field name is in that set.

I will patch this internally.

PR #1317, opened 2025-10-16
fix report unsafe `__get__` return types

### LINKED ISSUE #1175: report unsafe `__get__` return types
### Describe the Bug

descriptors can be unsafe:

```py
class A: 
    pass

class B(A):
    def __get__(self, *_) -> 1:
        return 1

class C:
    a: A = B()

C.a  # static: A, runtime: 1
```

we can avoid this by reporting an error on the signature of `__get__`: "`__get__`s return type must be assignable to the super classes `__get__` (defaulting to the type of the superclass in it's absense)"

the same applies to `__set__`

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1175

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -31,6 +31,7 @@ use ruff_python_ast::name::Name;
 use ruff_text_size::TextRange;
 use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
+use vec1::Vec1;
 use vec1::vec1;
 
 use crate::alt::answers::LookupAnswer;
@@ -1264,6 +1265,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         } else {
             value_ty
         };
+        let descriptor_value_ty = value_ty.clone();
 
         // Types provided in annotations shadow inferred types
         let ty = if let Some(ann) = annotation {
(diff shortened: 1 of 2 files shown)


stroxler (changes requested):
Hi @asukaminato0721 -

Thanks for the PR! I'm sorry to say I think we may have mis-marked this issue as a good first issue.

It's actually pretty tricky and I'm not sure we're ready for an implementation. Your code looks right to me in the sense that it implements the requested behavior, but I don't think we're at a point where we can actually commit to much here yet.

I've commented on the original issue, but there are a host of questions we need to answer before we can implement and ship a solution:
- First off, do we want to diverge from other type checkers at all? At the moment, no other type checker implements the proposed check
- What do we do if the annotated type is actually the descriptor type itself? I think this code as written would complain, but the conformance tests explicitly say we are expected to allow this (see https://github.com/python/typing/blob/main/conformance/tests/dataclasses_descriptors.py)
- How do we want to handle built-in types that run into the same kinds of behaviors, e.g. a `Callable` that gets assigned a function def (which has method binding descriptors)

While we almost certainly could implement some kind of best-effort check here, I think we may just not be ready to do that yet.

Apologies for the churn, the change itself looks like it could be a great starting point whenever we're ready to put in (and document the behavior of) best-effort checks for incompatible descriptor behaviors

rchen152:
Closing this PR since it's stale, and it sounds like we need more discussion on #1175 before implementing.

PR #1318, opened 2025-10-16
fix error on definition for implicit abstract classes

### LINKED ISSUE #1200: error when a class does not implement abstractmethods
i think it should be an error to define any class that has abstract methods

basedpyright and PyCharm both report an error here

```py

class A(ABC):
    @abstractmethod
    def f(self): ...

class B(A):
    pass
```

https://basedpyright.com/?typeCheckingMode=all&reportUnannotatedClassAttribute=false&code=GYJw9gtgBAhgRgYygSwgBzCALlAggIQGEAaWOAZyxBgSwgFMsALMAEwChOEAbGc8vAAoChAJQAudlGlQAAvErVaDZmykzW9YFGCDy9bsAlQAdGa69%2BUfMInrpaPuXZA

related #510, #511

asukaminato0721:
fix #1200

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -146,7 +146,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 "metaclass" => Either::Left(x),
                 _ => Either::Right((n.clone(), self.expr_class_keyword(x, errors))),
             });
-        let has_explicit_metaclass = !metaclasses.is_empty();
 
         let base_metaclasses = bases_with_metadata
             .iter()
@@ -285,7 +284,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         {
             self.validate_frozen_dataclass_inheritance(cls, dm, &bases_with_metadata, errors);
         }
-        let extends_abc = self.extends_abc(&bases_with_metadata);
+        let extends_abc = self.extends_abc(&bases_with_metadata, metaclass.as_ref());
 
         // Compute final base class list.
         let bases = if is_typed_dict && bases_with_metadata.is_empty() {
@@ -314,7 +313,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         ClassMetadata::new(
             bases,
             metaclass,
-            has_explicit_metaclass,
             keywords,
             typed_dict_metadata,
             named_tuple_metadata,
@@ -1160,7 +1158,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         AbstractClassMembers::new(abstract_members)
     }
 
-    fn extends_abc(&self, bases_with_metadata: &Vec<(Class, Arc<ClassMetadata>)>) -> bool {
+    fn extends_abc(
+        &self,
+        bases_with_metadata: &Vec<(Class, Arc<ClassMetadata>)>,
+        metaclass: Option<&ClassType>,
+    ) -> bool {
         for (base, base_metadata) in bases_with_metadata {
             if base.has_toplevel_qname("abc", "ABC") {
                 return true;
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -337,19 +337,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     ),
                 );
             } else if !metadata.is_protocol() && !metadata.is_new_type() {
-                let has_direct_abc_base = metadata
-                    .base_class_objects()
-                    .iter()
-                    .any(|base| base.has_toplevel_qname("abc", "ABC"));
-                let has_explicit_abc_metaclass = metadata.has_explicit_metaclass()
-                    && metadata.custom_metaclass().is_some_and(|meta| {
-                        meta.class_object().has_toplevel_qname("abc", "ABCMeta")
-                    });
+                let extends_abc = metadata.extends_abc();
                 let defines_abstract_member = cls.fields().any(|name| {
                     self.get_field_from_current_class_only(cls, name)
                         .is_some_and(|field| field.is_abstract())
                 });
-                if !has_direct_abc_base && !has_explicit_abc_metaclass && !defines_abstract_member {
+                if !extends_abc && !defines_abstract_member {
                     self.error(
                         errors,
                         cls.range(),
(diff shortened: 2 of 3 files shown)


pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -971,16 +973,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
                         }
-                    } else if metadata.is_new_type() {
-                        self.error(
-                            errors,
-                            range,
-                            ErrorInfo::Kind(ErrorKind::InvalidInheritance),
-                            "Subclassing a NewType not allowed".to_owned(),
-                        );
+                    } else {
+                        if metadata.is_new_type() {
```
yangdanny97: this conditional can be collapsed

pyrefly/lib/alt/solve.rs
```diff
@@ -336,6 +336,31 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         members
                     ),
                 );
+            } else if !metadata.is_protocol() {
+                let has_direct_abc_base = metadata
+                    .base_class_objects()
```
yangdanny97: I belive you can just check `extends_abc` on the class metadata do do the same thing as has_direct_abc_base + has_explicit_abc_metaclass

https://github.com/facebook/pyrefly/blob/ed6224fd4783eabaeb8e07ee063ffa8c96f2d1e5/pyrefly/lib/alt/class/class_metadata.rs#L1158

I'd expect the logic change here to be fairly minimal

basically just

```
if is_final:
   old error
else if !is_protocol && !extends_abc
   new error
```

pyrefly/lib/alt/solve.rs
```diff
@@ -336,6 +336,31 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                    });
+                let defines_abstract_member = cls.fields().any(|name| {
+                    self.get_field_from_current_class_only(cls, name)
+                        .is_some_and(|field| field.is_abstract())
+                });
+                if !has_direct_abc_base && !has_explicit_abc_metaclass && !defines_abstract_member {
+                    self.error(
+                        errors,
+                        cls.range(),
+                        ErrorInfo::Kind(ErrorKind::BadClassDefinition),
```
yangdanny97: this should be a unique error code `implicit-abstract-class`, with default severity = Ignore

pyrefly/lib/alt/types/class_metadata.rs
```diff
@@ -331,11 +336,20 @@ impl Display for ClassSynthesizedFields {
 /// A struct representing a class's metaclass. A value of `None` indicates
 /// no explicit metaclass, in which case the default metaclass is `type`.
 #[derive(Clone, Debug, TypeEq, PartialEq, Eq, Default)]
-struct Metaclass(Option<ClassType>);
+struct Metaclass {
+    value: Option<ClassType>,
+    explicit: bool,
```
yangdanny97: i believe this change is not necessary if you check extends_abc. For future reference (not relevant ot this PR)`custom_metaclass` on the class metadata will only return something if the metaclass is not the default.

yangdanny97 (changes requested):
Back to you with some comments

asukaminato0721:
Seems that https://github.com/facebook/pyrefly/pull/1318/commits/6f6634dfbd17fbefa79d236b5f060dd5e8320129 will fail...

extends_abc() returns true even when the class inherits an ABC only indirectly, because the flag is propagated through base metadata.

Child(Base) now satisfies `extends_abc() == true`, so the solver skips emitting ErrorKind::ImplicitAbstractClass.

```rs
failures:

---- test::abstract_methods::test_abstract_async_iterator stdout ----
DEBUG Running with 3 threads (5 MiB stack size)
DEBUG Running epoch 1 of run 0
DEBUG Committing transaction
ERROR Abstract methods for async generators should use `def`, not `async def` [bad-function-definition]
   --> pyrefly/lib/test/abstract_methods.rs:240:15
    |
240 |     async def foo(self) -> AsyncIterator[int]:  # E: Abstract methods for async generators should use `def`, not `async def`
    |               ^^^
    |
Error: Expectations failed for in-memory pyrefly/lib/test/abstract_methods.rs: expected 2 errors, but got 1

---- test::abstract_methods::test_abstract_property stdout ----
DEBUG Running with 3 threads (5 MiB stack size)
DEBUG Running epoch 1 of run 0
DEBUG Committing transaction
ERROR Cannot instantiate `Child` because the following members are abstract: `processor` [bad-instantiation]
   --> pyrefly/lib/test/abstract_methods.rs:225:10
    |
225 | x = Child()  # E: Cannot instantiate `Child`
    |          ^^
    |
Error: Expectations failed for in-memory pyrefly/lib/test/abstract_methods.rs: expected 2 errors, but got 1

---- test::abstract_methods::test_inherited_abstract_method stdout ----
DEBUG Running with 3 threads (5 MiB stack size)
DEBUG Running epoch 1 of run 0
DEBUG Committing transaction
ERROR Cannot instantiate `Child` because the following members are abstract: `method` [bad-instantiation]
   --> pyrefly/lib/test/abstract_methods.rs:102:14
    |
102 | child = Child()  # E: Cannot instantiate `Child`
    |              ^^
    |
Error: Expectations failed for in-memory pyrefly/lib/test/abstract_methods.rs: expected 2 errors, but got 1

---- test::abstract_methods::test_partial_implementation stdout ----
DEBUG Running with 3 threads (5 MiB stack size)
DEBUG Running epoch 1 of run 0
DEBUG Committing transaction
ERROR Cannot instantiate `Partial` because the following members are abstract: `method2` [bad-instantiation]
   --> pyrefly/lib/test/abstract_methods.rs:144:12
    |
144 | p = Partial()  # E: Cannot instantiate `Partial`
    |            ^^
    |
Error: Expectations failed for in-memory pyrefly/lib/test/abstract_methods.rs: expected 2 errors, but got 1


failures:
    test::abstract_methods::test_abstract_async_iterator
    test::abstract_methods::test_abstract_property
    test::abstract_methods::test_inherited_abstract_method
    test::abstract_methods::test_partial_implementation
```

yangdanny97:
Let me import and investigate

yangdanny97:
You're right, extends_abc will return true if it's transitive, we'd need a similar check for a class directly extending ABC. I'll see if I can patch this

yangdanny97:
I believe I've managed to patch it internally. Instead of storing the metaclass as Option<ClassType>, I made an enum that encodes its origin.

PR #1319, opened 2025-10-16
fix reports the wrong number of removed ignores

### LINKED ISSUE #1309: --remove-unused-ignores reports the wrong number of removed ignores
### Describe the Bug

Test file (named `test1.py`):
```
"""Test."""
1 + 1  # pyrefly: ignore
```
Expected behavior: pyrefly should report that it removed 1 unused ignore
Actual behavior:
```
$ pyrefly check test1.py --remove-unused-ignores
 INFO 0 errors
 INFO Removed 2 unused error suppression(s) in 1 file(s)
```
(Pyrefly does correctly remove the 1 ignore, but it claims to have removed 2.)

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1309

### DIFF
--- a/pyrefly/lib/error/suppress.rs
+++ b/pyrefly/lib/error/suppress.rs
@@ -162,6 +162,15 @@ pub fn find_unused_ignores<'a>(
 }
 
 pub fn remove_unused_ignores(loads: &Errors, all: bool) {
+    let removed_ignores = remove_unused_ignores_internal(loads, all);
+    info!(
+        "Removed {} unused error suppression(s) in {} file(s)",
+        removed_ignores.values().sum::<usize>(),
+        removed_ignores.len(),
+    );
+}
+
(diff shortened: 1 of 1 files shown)


yangdanny97:
Hi, do you mind writing up a quick summary of the nature of the bug and how you fixed it? It would make the logic a bit easier to follow. I can see that it fixes the issue, but it would be useful to know why certain things happen.

asukaminato0721:
old: incremented the counter for every candidate line flagged by find_unused_ignores, even if the line no longer contained a suppression marker.

new: split the work into' remove_unused_ignores_internal', returning a `SmallMap<PathBuf, usize>` of actual removals. This allows the public to log the correct totals and provides tests with direct access to the counts.

In the inner loop, now check the regex.is_match(line) before counting and mutating the line; untouched lines at the target positions are copied through without bumping the tally, so empty or whitespace-only rows don’t inflate the numbers.

Issue #52, opened 2025-03-28
Feature: override checks for typed dictionaries & named tuples


yangdanny97:
These have special override semantics

krikera:
Hey! I would like to work on this.

yangdanny97:
@krikera great! let me know if you get stuck or have any questions :)

yangdanny97:
Marking this as up for grabs since there hasn't been any progress. If you're still working on it I could assign this back

yangdanny97:
I believe this is complete based on the latest conformance results https://github.com/facebook/pyrefly/blob/main/conformance/third_party/results.json

PR #1346, opened 2025-10-21
fix Feature: inheritance checks for typed dictionaries

asukaminato0721:
part of #52

### DIFF
--- a/pyrefly/lib/test/typed_dict.rs
+++ b/pyrefly/lib/test/typed_dict.rs
@@ -174,6 +174,31 @@ def test(a: A) -> None:
 );
 
 testcase!(
+    test_typed_dict_inheritance_field_qualifiers,
+    r#"
+from typing import NotRequired, ReadOnly, Required, TypedDict
+
+class ParentRequired(TypedDict):
+    x: int
+
+class ChildOptional(ParentRequired):
+    x: NotRequired[int]  # E: TypedDict field `x` in `ChildOptional` must remain required because parent TypedDict `ParentRequired` defines it as required
+
+class ParentOptional(TypedDict, total=False):
+    x: int
+
+class ChildRequired(ParentOptional):
+    x: Required[int]  # E: TypedDict field `x` in `ChildRequired` cannot be made required; parent TypedDict `ParentOptional` defines it as non-required
+
+class ParentMutable(TypedDict):
+    x: int
+
+class ChildReadOnly(ParentMutable):
+    x: ReadOnly[int]  # E: TypedDict field `x` in `ChildReadOnly` cannot be marked read-only; parent TypedDict `ParentMutable` defines it as mutable
+"#,
+);
+
+testcase!(
     test_typed_dict_contextual,
     r#"
 from typing import TypedDict


yangdanny97:
Running test.py should generate some changes, could you please include them in the commit?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1780,6 +1780,116 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .is_some_and(|metadata| metadata.fields.contains_key(field_name))
     }
 
+    fn check_typed_dict_override_rules(
```
yangdanny97: This feels like we duplicated some of the checks from the regular class field override checks - is this necessary, or can we reuse the regular override checks & add on some additional checks just for typed dicts?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1780,6 +1780,116 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            return;
+        }
+
+        if parent_info.required && !child_info.required {
+            self.error(
+                errors,
+                range,
+                ErrorInfo::Kind(ErrorKind::BadTypedDict),
+                format!(
+                    "TypedDict field `{}.{}` cannot be marked as optional because parent `{}` requires it",
```
yangdanny97: the required/optional checks seem like the ones that were added, the read-only and assignability checks are the same as regular classes, if i'm not mistaken.

conformance/third_party/conformance.result
```diff
@@ -332,8 +332,6 @@
   "typeddicts_readonly.py": [],
   "typeddicts_readonly_consistency.py": [],
   "typeddicts_readonly_inheritance.py": [
-    "Line 98: Expected 1 errors",
-    "Line 106: Expected 1 errors",
```
yangdanny97, line 335: also, I don't think this fully fixes #52, we also have to apply these rules in multiple inheritance.

The code pointer for that is https://github.com/facebook/pyrefly/blob/41db847dcc7772470e386561b06b572a63ca5331/pyrefly/lib/alt/class/class_field.rs#L2007

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -2005,38 +2114,53 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
                 }
             }
         }
 
-        for (field_name, class_and_types) in inherited_fields.iter() {
-            if class_and_types.len() > 1 {
-                let types: Vec<Type> = class_and_types.iter().map(|(_, ty)| ty.clone()).collect();
+        let child_metadata = self.get_metadata_for_class(cls);
+
+        for (field_name, infos) in inherited_fields.iter() {
```
stroxler: Might be worth naming `infos` something more descriptive like `inherited_field_infos_by_ancestor`

The name is more verbose, but it's only used a few times, and in my opinion it's easy to lose track of what "infos" means by the time you get here - in particular, it's easy to get confused about why there is more than one info for a single field

This is pretty subjective - up to you, feel free to ignore me if you think the name is too long

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -2005,38 +2114,53 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
 
     /// For classes with multiple inheritance, check that fields inherited from multiple base classes are consistent.
     pub fn check_consistent_multiple_inheritance(&self, cls: &Class, errors: &ErrorCollector) {
-        // Maps field from inherited class
+        struct InheritedFieldInfo {
```
stroxler, line 2117: 😮 I didn't fully realize you can nest type defs inside functions, that's nifty!

conformance/third_party/conformance.result
```diff
@@ -331,10 +331,7 @@
...
-    "Line 119: Expected 1 errors",
-    "Line 132: Expected 1 errors"
+    "Line 119: Expected 1 errors"
```
stroxler, line 334: This seems like great progress!

Could we also get a unit test showing the improvements though? We try not to rely on the conformance tests as the only line of testing for features, because they are run late in the development cycle - we want as many regression tests as possible in the faster and more-frequently-run unit test suite.

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -2046,8 +2173,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                             | Type::Overload(_)
                     )
                 }) {
-                    // TODO(fangyizhou): Handle bound methods and functions properly.
```
fangyi-zhou, line 2049: Don't think this should be removed --- I still plan to come back to this

PR #1350, opened 2025-10-21
fix test_new_type_tuple

asukaminato0721:
fix test_new_type_tuple

### DIFF
--- a/pyrefly/lib/alt/class/new_type.rs
+++ b/pyrefly/lib/alt/class/new_type.rs
@@ -23,10 +23,21 @@ use crate::types::class::Class;
 use crate::types::types::Type;
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
+    fn new_type_base_type(&self, cls: &Class, base: &Class) -> Type {
+        let class_bases = self.get_base_types_for_class(cls);
+        if let Some(tuple_base) = class_bases.tuple_base() {
+            return Type::Tuple(tuple_base.clone());
+        }
+        let mut base_iter = class_bases.iter();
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/class/new_type.rs
```diff
@@ -23,10 +23,21 @@ use crate::types::class::Class;
 use crate::types::types::Type;
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
+    fn new_type_base_type(&self, cls: &Class, base: &Class) -> Type {
+        let class_bases = self.get_base_types_for_class(cls);
+        if let Some(tuple_base) = class_bases.tuple_base() {
+            return Type::Tuple(tuple_base.clone());
+        }
+        let mut base_iter = class_bases.iter();
```
yangdanny97, line 31: q: what's the purpose of lines 31-34? 

```
let mut base_iter = class_bases.iter();
if let Some(base_class_type) = base_iter.next().cloned() {
    return Type::ClassType(base_class_type);
}
```

This is behavior-altering, but doesn't appear to be exercised in the tests

rchen152 (changes requested):

pyrefly/lib/alt/class/new_type.rs
asukaminato0721: Those lines are the fallback when the base isn’t a tuple. They reuse the first fully resolved base type from class_bases, so the synthesized constructors keep whatever type arguments the NewType base had, instead of reverting to a default-instantiated class type.

rchen152 (approved):

PR #1354, opened 2025-10-21
Show Argument types in Overload Message Errors

### LINKED ISSUE #1302: Error messages: Show argument types in overload error
### Describe the Bug

Otherwise, users don't know why no overloads matched unless you hover over each argument.

```python
from typing import *

def test() -> int | None:
    return 5


a = int(test())

```

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRmGGEbwGACgCUdALQA%2BTun4AfOgDlc6GIhF1jdSjAYBXSujoBWEQ-So6AXjWyG0%2BXJEgANCBkZmBQpIQMtFAUAMR0AAqkwaF0aFh4%2BHQAxnqQbFaoDBB6hCKxAMowMHQAFgwMxHCIAPRNQRKhhLxsTTDoTZi4mXBN2ei5%2BYV6TXRgvHSoAG6o0KjYsFk5EHmUBUU2uMST6HAl6GQM1XqKCzCUcHuudEIgAMyEAIwATM%2B%2BAL4BqEyhRuADFoDAKKkcAQSORfkA

_No response_

Adist319:
Resolves #1302. With these changes, it's clearer to the user why the overload error was thrown. When no overload matches a function call, Pyrefly now displays the actual argument types that were provided. This eliminates the need to hover over each argument to understand why the call failed.

From Sandbox link:
https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRmGGEbwGACgCUdALQA%2BTun4AfOgDlc6GIhF1jdSjAYBXSujoBWEQ-So6AXjWyG0%2BXJEgANCBkZmBQpIQMtFAUAMR0AAqkwaF0aFh4%2BHQAxnqQbFaoDBB6hCKxAMowMHQAFgwMxHCIAPRNQRKhhLxsTTDoTZi4mXBN2ei5%2BYV6TXRgvHSoAG6o0KjYsFk5EHmUBUU2uMST6HAl6GQM1XqKCzCUcHuudEIgAMyEAIwATM%2B%2BAL4BqEyhRuADFoDAKKkcAQSORfkA

Before:
  ERROR No matching overload found for function `int.__new__`
    Possible overloads:
    (cls: type[int], x: Buffer | SupportsIndex | ..., /) -> int

  After:
  ERROR No matching overload found for function `int.__new__`
    Called with arguments: (type[int], int | None)
    Possible overloads:
    (cls: type[int], x: Buffer | SupportsIndex | ..., /) -> int

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -216,31 +216,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let args = call.vec_call_arg(args, self, errors);
         let keywords = call.vec_call_keyword(keywords, self, errors);
 
-        // Build a string showing the argument types for error messages
-        let mut arg_type_strs = Vec::new();
-        for arg in &args {
-            let (ty, prefix) = match arg {
-                CallArg::Arg(value) => (value.infer(self, errors), ""),
-                CallArg::Star(value, _) => (value.infer(self, errors), "*"),
-            };
-            let ty_display = self.solver().for_display(ty);
-            arg_type_strs.push(format!("{}{}", prefix, ty_display));
-        }
-        for kw in &keywords {
-            let ty = kw.value.infer(self, errors);
-            let ty_display = self.solver().for_display(ty);
-            if let Some(arg_name) = kw.arg {
-                arg_type_strs.push(format!("{}={}", arg_name.as_str(), ty_display));
-            } else {
-                arg_type_strs.push(format!("**{}", ty_display));
-            }
-        }
-        let args_display = if arg_type_strs.is_empty() {
-            "()".to_owned()
-        } else {
-            format!("({})", arg_type_strs.join(", "))
-        };
-
         // Evaluate the call following https://typing.python.org/en/latest/spec/overload.html#overload-call-evaluation.
 
         // Step 1: eliminate overloads that accept an incompatible number of arguments.
@@ -286,7 +261,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 );
 
                 // Step 3: perform argument type expansion.
-                let mut args_expander = ArgsExpander::new(args, keywords);
+                let mut args_expander = ArgsExpander::new(args.clone(), keywords.clone());
                 let owner = Owner::new();
                 'outer: while !matched
                     && let Some(arg_lists) = args_expander.expand(self, errors, &owner)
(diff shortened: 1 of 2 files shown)


yangdanny97:
> ERROR No matching overload found for function int.__new__
Called with arguments: (type[int], int | None)

Curious why this is on a separate line, rather than

```
ERROR No matching overload found for int.__new__(type[int], int | None)
```

Will review in more detail later.

pyrefly/lib/alt/overload.rs
```diff
@@ -343,6 +368,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     "No matching overload found for function `{}`",
                     metadata.kind.as_func_id().format(self.module().name())
                 ),
+                format!("Called with arguments: {}", args_display),
```
rchen152: We should only build `args_display` when the call actually errors; inferring types can be expensive. Could you move the `args_display` logic above to be right above `let mut msg = ...` instead?

pyrefly/lib/alt/overload.rs
```diff
@@ -216,6 +216,31 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
         let keywords = call.vec_call_keyword(keywords, self, errors);
 
+        // Build a string showing the argument types for error messages
+        let mut arg_type_strs = Vec::new();
+        for arg in &args {
+            let (ty, prefix) = match arg {
+                CallArg::Arg(value) => (value.infer(self, errors), ""),
+                CallArg::Star(value, _) => (value.infer(self, errors), "*"),
+            };
+            let ty_display = self.solver().for_display(ty);
```
rchen152: nit: we have a convenient AnswersSolver wrapper for for_display, so you can just do `self.for_display(ty)`.

rchen152 (changes requested):
Thanks! The argument printing looks good to me, but I think we need to change where it happens, or it'll be too expensive.

I also agree with Danny that it would be nice to put the argument types on the first line rather than adding another line.

Adist319:
@rchen152 Thanks for the feedback! I missed some of the points you brought up. Will work on the changes soon.

Adist319:
Moved the args_display logic down. Also addressed the style nit and moved the types to the first line.

PR #1361, opened 2025-10-22
fix report unsafe operator type on subclasses #1176

### LINKED ISSUE #1176: report unsafe operator type on subclasses
### Describe the Bug

```py
class A:
    def __add__(self, other: A) -> int:
        return 1


class B(A):
    def __radd__(self, other: A) -> str:
        return "B"


def f(a1: A, a2: A):
    x: int = a1 + a2
    print(x)  # static: int, runtime: "B"

f(A(), B())
```

here, the right hand subtype rule will be used, and `B.__radd__` will be called first, resulting in `"B"`, not `1`

if the supertype additionally defines a `__radd__`, then it can be incompatible with it's `__add__`, as all the information about the operation will be known

to ensure safety we must report when the subtypes `__radd__` in incompatable with the supertypes `__add__`, when the super type does not define a `__radd__`. but the right hand subtype rule complicates things...

so we should always form a union when the lower bound is not known:
```py
class A:
     def __add__(self, other: object) -> int: ...
     def __radd__(self, other: object) -> str: ...

     def __sub__(self, other: object) -> int: ...

     def __or__(self, other: object) -> int: ...

class B(A):
     def __radd__(self, other: object) -> str: ... # no error, valid override of A.__radd__

     def __rsub__(self, other: object) -> Literal[1]: ... # not error, compatiable with A.__add__

     def __ror__(self, other: object) -> str: ...  # error, incompatible with A.__or__

def f(a1: A = A(), a2: A = B()):
    # when we don't know what `a2` is
    a1 + a2  # static: int | str, runtime: str

    # we definitely know that rhs is a subtype of lhs
    A() + B()  # static: str, runtime: str

    # we definitely know that rhs is not a subtype of lhs
    a1 + A()  # static: int, runtime: int
```

# this also applies to inplace operators
```py
class A:
    def __add__(self, other: A) -> int:
        return 1
class B(A):
    def __iadd__(self, other: A) -> str:   # expect error: `__iadd__` must be compatable with `A.__add__`, as `A` does not define `__iadd__`
        return "a"

def f(a1: A, a2: A):
    a = a1
    a += a2
    print(a)  # static: int, runtime: "a"
f(B(), A())
```
note that the righthand subtype rule applies after `__iadd__`, so this case is much more straightforward to support

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1176

### DIFF
--- a/pyrefly/lib/alt/operators.rs
+++ b/pyrefly/lib/alt/operators.rs
@@ -44,6 +44,7 @@ use crate::error::context::ErrorContext;
 use crate::error::context::ErrorInfo;
 use crate::error::context::TypeCheckContext;
 use crate::error::context::TypeCheckKind;
+use crate::types::class::ClassType;
 use crate::types::literal::Lit;
 use crate::types::tuple::Tuple;
 use crate::types::types::Type;
@@ -350,6 +351,32 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         Some(self.heap.mk_concrete_tuple(repeated))
     }
 
+    fn binop_call_type(
+        &self,
+        target: &Type,
+        method_name: &Name,
+        arg: &Type,
+        range: TextRange,
+    ) -> Option<Type> {
+        let errors = self.error_collector();
+        let arg_ty = arg.clone();
+        let call_args = [CallArg::ty(&arg_ty, range)];
+        let ret = self.call_magic_dunder_method(
+            target,
+            method_name,
+            range,
+            &call_args,
+            &[],
+            &errors,
+            None,
+        );
+        if errors.is_empty() {
+            ret.filter(|ty| !ty.is_error())
+        } else {
+            None
+        }
+    }
+
     pub fn binop_infer(
         &self,
         x: &ExprBinOp,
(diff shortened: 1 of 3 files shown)


pyrefly/lib/test/operators.rs
```diff
@@ -581,6 +614,6 @@ class A:
 class B:
     def __radd__(self, other) -> Self:
         return self
-assert_type(A() + B(), A)  # E: `A.__add__` is deprecated
+assert_type(A() + B(), A | B)  # E: `A.__add__` is deprecated
```
kinto0, line 831: does this A | B imply we choose A | B in [this ](https://github.com/facebook/pyrefly/issues/1176#issuecomment-3432529675)example?


is it worth adding a test for that, and should it be just A?

pyrefly/lib/alt/operators.rs
```diff
@@ -224,14 +251,62 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-            // Reflected operator implementation: This deviates from the runtime semantics by calling the reflected dunder if the regular dunder call errors.
-            // At runtime, the reflected dunder is called only if the regular dunder method doesn't exist or if it returns NotImplemented.
-            // This deviation is necessary, given that the typeshed stubs don't record when NotImplemented is returned
-            let calls_to_try = [
-                (&Name::new_static(op.dunder()), lhs, rhs),
-                (&Name::new_static(op.reflected_dunder()), rhs, lhs),
-            ];
-            self.try_binop_calls(&calls_to_try, range, errors, &context)
+            let class_type_of = |ty: &Type| -> Option<ClassType> {
+                match ty {
```
kinto0, line 462: mapping over type like this is almost never what we want: it's hard to maintain and prone to errors

pyrefly/lib/test/operators.rs
kinto0: The current behavior for these reflected operators is:
- Try the "forward" operator, if that works, use the return type from that
- If the forward operator fails, try the reflected operator, using it's return type if it succeeds
- If both the forward and reflected operators fail, error and use the return type from the forward operator

is this level of strictness practical? might be worth a discussion

kinto0 (changes requested):
thanks for the contribution! we tested this change against instagram and it introduced a lot of unrelated type errors. not entirely sure we want to merge it just yet. it might be worth discussing it more at an office hours or on discord if you're still interested in this issue

yangdanny97 (changes requested):

kinto0:
closing old draft PRs. please reopen and mark "ready for review" if you are ready!

PR #1363, opened 2025-10-22
fix Error when calling abstract method on super() #1358

### LINKED ISSUE #1358: Error when calling abstract method on `super()`
### Describe the Bug

```python
from typing import Protocol
from abc import abstractmethod

class PColor(Protocol):
    @abstractmethod
    def draw(self) -> str:
        ...

class BadColor(PColor):
    def draw(self) -> str:
        return super().draw() # this should be an error
```

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AFag1wBjXFAA66MNXqpsozt178FcBpVSiGNGAwAWuTNOmioqOHEEBhCbwAUQ3CPFQAlIml0fdAALqmtq6%2BkYm6L50mDBgUVoA7g5wMFBg7nQAtAB8dBqUXhGRvoQlpujmltYAQqiYdlCOAvW8nt6%2B0bGYCUkpaZk5eQVFvpT6AK6UEXBjxDCUDu6EXaiJ6QDEjAYQ1nBGY1CYdNgwdBh0c9SUIAA0IGSjYFCkhCI0UBQbAqQPT7kYOAQ6OIZBA2BNUAwILh0IRpBsAMowE4GBgMYhwRAAekx9xiT0IvDYmJg6ExmDEcExwMgYK0kOhmLoYF4pwAbqhoApYEDoTTwfSIrhiAK4LD0GRDNCMqy5nAoREALx0SQgADMhAAjAAmFXSEAAX1uwQgMoAYtAYBQ0Fg8EQyAagA

_No response_

asukaminato0721:
fix #1358


pyrefly/lib/alt/attr.rs
```diff
@@ -239,6 +239,8 @@ pub enum NoAccessReason {
     /// We do not allow class-level mutation of descriptors (this is conservative,
     /// it is unspecified whether monkey-patching descriptors should be permitted).
     SettingDescriptorOnClass(Class),
+    /// Calling a protocol method via `super()` when the protocol provides no implementation.
+    SuperProtocolStubMethod(Class),
```
yangdanny97: I think this is a bit too specific, we'd probably also want this error if the method is a regular method that's marked as abstract.

crates/pyrefly_types/src/callable.rs
```diff
@@ -334,6 +334,8 @@ pub struct FuncFlags {
     pub has_final_decoration: bool,
     /// A function decorated with `@abc.abstractmethod`
     pub is_abstract_method: bool,
+    /// Function body is treated as a stub (e.g. body is `...` or absent in a stub file)
+    pub is_stub: bool,
```
stroxler: Could we change this to `lacks_implementation`?

The current name isn't terrible, but the problem problem with `is_stub` is that given standard "a function is a stub" naively means "is it defined in a `.pyi` file, which is not what this flag represents

(Note: I realized the name `Stub` in `FunctionStubOrImpl` is preexisting, but I remain concerned - see other review comments)

pyrefly/lib/test/class_super.rs
```diff
@@ -113,6 +113,36 @@ class C:
     "#,
 );
 
+testcase!(
+    test_super_protocol_stub_method,
```
stroxler: Similar to my other rename request - could we name this `test_super_protocol_unimplemented_method`?

pyrefly/lib/alt/function.rs
```diff
@@ -259,6 +259,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 .map(|(idx, range)| (self.get_idx(*idx).arc_clone_ty(), *range)),
         );
 
+        if stub_or_impl == FunctionStubOrImpl::Stub {
```
stroxler, line 266: Note: the name `FunctionStubOrImpl::Stub` is preexisting and goes against my request, but we can probably live with it for now - I may try to rename it later.

But related: @yangdanny97  do you know what happens if this is actually a "stub" in the sense of coming from a `.pyi` file?

In particular - will an actual stub from a `.pyi` file always have `FunctionStubOrImpl::Impl` rather than `FunctionStubOrImpl::Stub`?

It would be wrong for a stub to ever use `Stub` (that sentence explains exactly why the name `Stub` is probably the wrong name haha!), since for stubs we have to assume they imply an implementation exists, typically in native code.

We might want to add a test for that if we're not sure - this could throw a lot of false positives on actual `.pyi` files if we get it wrong

stroxler (commented):
Mostly this looks good to me, but I'm a little concerned that
(a) the preexisting use of the word "stub" is misleading, and
(b) we need to double-check the behavior on actual stubs (as in, .pyi files)

pyrefly/lib/alt/function.rs
yangdanny97: https://github.com/facebook/pyrefly/blob/01f0a2971e9fba5b6568f14e681e1e3671c177c2/pyrefly/lib/binding/function.rs#L477

This is where it's set, I don't think it considers the type of module, only whether it has a proper body or not.

asukaminato0721: Is there something I need to update?

yangdanny97:
I'll try to get things patched and merged this week

yangdanny97:
Thanks for your patience, I patched this internally so that we only raise the error if the protocol method is not defined in a .pyi file (for those, we have to assume the implementation is there since they are typically not included in stubs).

Issue #361, opened 2025-05-26
Never/NoReturn narrowing


Zaczero:
### Describe the Bug

```py
from typing import *

def raise_error() -> NoReturn:
    raise Exception()

a = cast(int | None, 1)
if a is None:
    raise_error()

reveal_type(a)  # revealed type: int | None
```

I expect `a` to be narrowed to `int`.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQCYCmwUIAhkgM7kD65I4IAFAJRQC0AfFAByYAErkYAVxAoAXESiLqdRlACiADwDG5BDCRgU3EjSgBeKFpr0YHVPgA+ww+QA0UAIxciSKqYbOKOTySsoMzKzsxkQg5ABu5DQANkzwCOQcNFxAA

yangdanny97:
I believe this is a gap in our control flow handling, where calls to `NoReturn` functions are not handled properly.

fangyi-zhou:
I've taken a brief look at the issue.

Currently, we are having control flow in binding builder, and we mark a flow as terminated using `self.scopes.mark_flow_termination();` in the following scenarios:
- `return`
- `raise`
- `assert` with a false condition
- calling `sys.exit()`, `exit()`, `quit()` or `os.exit()`

The tricky part of implementing control flow termination for functions with return types `Never` and `NoReturn` is that typing information is not available during binding building.

stroxler:
We have some ideas about an approach that might work

I'll tentatively claim this: I'm looking at unrelated flow merging changes related to narrowing, and this would be a logical thing to do soon after. I don't think there's much chance I'll get to it for at least a week, but it probably doesn't make sense for anyone else to work on it right now

yangdanny97:
[...]

PR #1365, opened 2025-10-22
mark flow as terminated when calling pytest.fail, pytest.xfail, pytest.skip

### LINKED ISSUE #1337: unbound-name error produced on variable set in 'try' despite 'except' has unconditional raise
### Describe the Bug

Python version: 3.14.0

Consider this Python code, using pytest 8.4.2:
```
import pytest

def test_oops() -> None:
    try:
        val = True
    except:
        pytest.fail("execution stops here")
        # raise

    assert val, "oops"
```

This is a simplified variant of a pattern in use more broadly, just to illustrate the situation clearly.

Pyrefly flags an issue:
```
ERROR `val` may be uninitialized [unbound-name]
  --> example.py:10:12
   |
10 |     assert val, "oops"
   |            ^^^
   |
 INFO 1 error
```

If the local `raise` is uncommented, there's no problem as the unconditional raise make it obvious that the execution stops and there's no chance of reaching the reference to val while it's unbound.

But, even without the local raise, the pytest.fail() call _itself_ performs an unconditional raise and I would have expected that on its own should be sufficient.

With the latest Pytest 8.4.2, the "fail" function as defined in /opt/venv/lib/python3.14/site-packages/_pytest/outcomes.py, is simply:
```
@_with_exception(Failed)
def fail(reason: str = "", pytrace: bool = True) -> NoReturn:
    """Explicitly fail an executing test with the given message.

    :param reason:
        The message to show the user as reason for the failure.

    :param pytrace:
        If False, msg represents the full failure information and no
        python traceback will be reported.

    :raises pytest.fail.Exception:
        The exception that is raised.
    """
    __tracebackhide__ = True
    raise Failed(msg=reason, pytrace=pytrace)
```

Both mypy and pyright in strict mode don't flag an issue even when the local raise is removed. The versions in use for mypy, pyright and pyrefly are:
```
# mypy --version
mypy 1.18.2 (compiled: yes)

# pyright --version
pyright 1.1.406

# pyrefly --version
pyrefly 0.37.0
```

### Sandbox Link

_No response_

_No response_

yangdanny97:
Summary:
fixes https://github.com/facebook/pyrefly/issues/1337

full resolution needs a general solution to https://github.com/facebook/pyrefly/issues/361 but we can special-case some common APIs for now.

Differential Revision: D85164900

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -97,6 +97,7 @@ use crate::error::context::ErrorInfo;
 use crate::error::context::TypeCheckContext;
 use crate::error::context::TypeCheckKind;
 use crate::error::style::ErrorStyle;
+use crate::export::special::SpecialExport;
 use crate::graph::index::Idx;
 use crate::solver::solver::SubsetError;
 use crate::types::annotation::Annotation;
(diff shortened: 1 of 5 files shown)


stroxler (approved):

PR #1376, opened 2025-10-23
fix Don't use parent class attribute type for method-declared __private attributes #1294

### LINKED ISSUE #1294: Don't use parent class attribute type for method-declared __private attributes
### Describe the Bug

Attributes starting with double-underscore have their names "mangled" at runtime. They are already exempt from override checks (declaring the attributes in the class body works today), but we don't special-case them when they are declared implicitly in a method by assigning to `self`.

```
class Config:
    pass


class DerivedConfig(Config):
    def foo(self) -> None:
        print("hello")


class B():
    def __init__(self, config: Config) -> None:
        self.__config = config


class C(B):
    def __init__(self, config: DerivedConfig) -> None:
        self.__config = config


    def bar(self) -> None:
        self.__config.foo() # should be ok
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQDCu6kA5ogDrr2-3G043YemqD6AERgAnCADcYmZqwhsAFMvYBKLjz6YYYemFy41cGFDBb6AWgB89AHIsYuvh-6z0AFzWcQAAtLKFwArRFuMTp6ACE1HW4PAyMAfVSIdAgfdPNLMAAaekoWdkQmUtUbB2dXd09eCytCdJKVNnoAXmLKtkjRGhjGNVjEvV4U%2BnTM7Nymwp728qlZBSVe6scXdDckhvp5ltS29i7F9n7kw3psVGk8q03anfqGw9bewhMzGwBiA8CuAArlBMDcYPRcABrbggAogMjSQxQUiEHy4AC2UAo-wACqQkWAUQcMDgCOdVEDpKgfBAWIRuP8AMowCGBHw%2BYhwRAAeh5iORqNw0jYPJg6B5mFwlDgPJOlOptJYPOMwvoqDkqGgqGwsApbCpNLpPFwxCV6DgDPQZB8gPQtgU0jgxrOAQAzIQAIwAJgCsIAvvDUJRaQoAGLQGAUNBYPBEMggf1AA

_No response_

asukaminato0721:
fix #1294

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1051,7 +1051,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
 
         let (value_ty, inherited_annotation) = match value {
             ExprOrBinding::Expr(e) => {
+                let is_private_attr = name.starts_with("__") && !name.ends_with("__");
                 let (inherited_ty, inherited_annot) = if direct_annotation.is_some() {
+                    (None, None)
+                } else if is_private_attr {
+                    // Private (double-underscore) attributes are effectively class-scoped and should not
+                    // inherit types or annotations from parent classes.
+                    is_inherited = IsInherited::No;
                     (None, None)
                 } else {
                     let (inherited_ty, annotation) =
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1063,6 +1069,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 };
                 let mut ty = if let Some(inherited_ty) = inherited_ty
                     && matches!(initial_value, RawClassFieldInitialization::Method(_))
+                    && !is_private_attr
```
rchen152, line 1072: This check isn't necessary, since you're already setting `inherited_ty` to `None` above. (No need to change this, I've patched it internally.)

PR #1386, opened 2025-10-25
impl docstring fold

### LINKED ISSUE #1335: Add fold/unfold all docstrings functionality
### Describe the Bug

A feature I find very useful in Pylance (which hopefully will be replaced by the blazingly fast pyrefly in time) is the ability to fold/unfold all docstrings within relevant `.py` files. An example of my suggested feature can be seen [here](https://devblogs.microsoft.com/python/python-in-visual-studio-code-november-2024-release/).

I'm not sure this feature falls under the expectations of what pyrefly is meant to be, but my colleagues and I would appreciate this ability.

asukaminato0721:
fix #1335
[Screencast_20251025_220013.webm](https://github.com/user-attachments/assets/4efb77a3-1205-492a-912d-ba13c6c6f03e)

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/basic.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/basic.rs
@@ -49,6 +49,7 @@ fn test_initialize_basic() {
                 "hoverProvider": true,
                 "inlayHintProvider": true,
                 "documentSymbolProvider": true,
+                "foldingRangeProvider":true,
                 "workspaceSymbolProvider": true,
                 "workspace": {
                     "workspaceFolders": {


lsp/src/extension.ts
```diff
@@ -253,6 +329,51 @@ export async function activate(context: ExtensionContext) {
...
+  );
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand('pyrefly.unfoldAllDocstrings', async () => {
+      await runDocstringFoldingCommand('editor.unfold');
+    }),
+  );
+
+  context.subscriptions.push(
+    vscode.languages.registerFoldingRangeProvider(
```
kinto0: I'm not convinced this part is working correctly: I'd expect this change to appear in the IDE as foldable > icons (I don't see them in your video or my local testing

we should instead put this logic (the folding range provider part) in the language server so editors other than VSCode can use them

kinto0 (changes requested):
awesome!! just a question + small nit about the location of foldingrangeprovider. if you want to take that out, the provider work could be done as a follow-up (then make sure this PR doesn't close #1335)

pyrefly/lib/state/state.rs
```diff
@@ -1597,6 +1608,49 @@ impl<'a> Transaction<'a> {
     }
 }
 
+fn collect_docstring_ranges(body: &[Stmt], ranges: &mut Vec<TextRange>) {
```
kinto0, line 1611: one small change I want to make is to redo this collector into a visitor so it doesn't miss nested cases:

```
fn collect_docstring_ranges(body: &[Stmt]) -> Vec<TextRange> {
    struct DocstringCollector {
        ranges: Vec<TextRange>,
    }

    impl Visitor<'_> for DocstringCollector {
        fn visit_body(&mut self, body: &[Stmt]) {
            if let Some(range) = Docstring::range_from_stmts(body) {
                self.ranges.push(range);
            }
            walk_body(self, body);
        }
    }

    let mut collector = DocstringCollector { ranges: Vec::new() };

    if let Some(range) = Docstring::range_from_stmts(body) {
        collector.ranges.push(range);
    }

    for stmt in body {
        Visitor::visit_stmt(&mut collector, stmt);
    }

    collector.ranges
}
```

i'll just do that in a follow-up

kinto0 (approved):
looks great! thanks for all the hard work :)

connernilsen (approved):

PR #1387, opened 2025-10-25
fix VScode extension does not highlight problems in files with symlinked paths under linux #1104

### LINKED ISSUE #1104: VScode extension does not highlight problems in files with symlinked paths under linux
### Describe the Bug

While trying out the pyrefly vscode extension I noticed that it does not display squiggly lines for type errors in the workspace I am working on. All other features (inlay hints, suggestions, etc.) are working fine and in the bottom pane "Problems" tab the errors are listed. If I click on one of the listed errors, vscode opens the file again, but this time with its absolute path (as displayed in the problem pane, with resolved symlink portions) and in that file the errors are highlighted correctly.

## Steps to reproduce
- Create a file with obviously incorrectly typed content.
- Symlink the file or the directory it lives in to a different location.
- Open the symlinked file with vscode.
- See no squiggly lines in the editor.
- See errors listed in the "Problems" pane.

Proof of concept script:
```bash
echo 'x: int = "A"' > test.py
ln -s test.py sym.py
code sym.py
```

Note that it does ot matter where in the file path the symlink occurs and whether the target is an absolute or a relative path. The behaviour is always the same.

### Sandbox Link

_No response_

- VSCode - OSS v1.104.0
- Pyrefly extension v0.33.0
- No pylance installed

asukaminato0721:
fix #1104

image


samwgoldman:
Please add an automated test. Why does this change fix the issue? Is it because canonicalize invokes realpath and absolutize doesn’t?

kinto0:
> Please add an automated test. Why does this change fix the issue? Is it because canonicalize invokes realpath and absolutize doesn’t?

I just tested it before/after and it seems to fix it well. yeah, canonicalize actually looks on the file system and resolves symlinks while absolutize does not.

PR #1390, opened 2025-10-26
fix narrowing on dict (and typed dict) .get

### LINKED ISSUE #238: [mypy parity] We should look at narrowing on dict (and typed dict) .get
### Describe the Bug

We ran into an issue trying to migrate someone where they wanted to narrow the `.get(<key>)` access on a dictionary (where the key was a literal string) from `Any | None` to `Any`

This would be a good thing to support, both for migration and general usability

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #238

Taught the narrowing logic to treat dict.get("literal") calls as key facets, so truthiness and is not None checks now refine the corresponding dictionary entry

Added regression coverage for both explicit is not None and plain truthiness checks on literal-key .get calls to confirm the dictionary entry narrows as expected

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -26,6 +26,8 @@ use crate::alt::answers_solver::AnswersSolver;
 use crate::alt::callable::CallArg;
 use crate::alt::callable::CallKeyword;
 use crate::binding::narrow::AtomicNarrowOp;
+use crate::binding::narrow::FacetOrigin;
+use crate::binding::narrow::FacetSubject;
 use crate::binding::narrow::NarrowOp;
 use crate::error::collector::ErrorCollector;
 use crate::error::style::ErrorStyle;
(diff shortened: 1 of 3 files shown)


yangdanny97:
Awesome, thanks!

yangdanny97:
The concern I have here is that this would apply to `get` methods on other types as well, not just `dict` and `TypedDict`. We might need a different type of Facet or Subject that somehow checks if we're calling a method on `dict` when we solve the narrow - if it's not a dict or TypedDict then we should not apply any narrowing.

asukaminato0721:
Introduced FacetOrigin/FacetSubject to track whether facet chains came from direct subscripts or .get calls, and only reuse them when the base is a dict-like source

Added runtime checks so .get-derived narrows only fire when the underlying type is a builtin dict or (partial) TypedDict

yangdanny97:
Thanks!

stroxler (approved):

yangdanny97:
Ok, I think this is in a mergeable state. There were two gaps in the implementation that I had to patch:
- firstly, when we have `if x.get("key"):` we can narrow the key to be truthy
- secondly, when we do a `.get()` or subscript narrow, we should be able to access the narrowed type via subscripting or calling `get` w/ the same key

PR #1391, opened 2025-10-26
impl disabled/unreachable code in gray

### LINKED ISSUE #229: Show disabled code in gray
### Describe the Bug

Given:

```
if sys.version > (3, 9):
   a
else:
   b
```

One of these branches should be in gray.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #229

image

Semantic token builder now tracks statically unreachable ranges, tagging any token inside them with the LSP deprecated modifier and synthesizing tokens for disabled class/function headers and simple assignments, so gray styling shows up even when bindings are skipped

LSP semantic token generation pulls those disabled ranges from the module AST before walking bindings, so the VS Code extension receives the extra modifier information

Added a regression test covering a version-gated branch to ensure unreachable definitions surface as deprecated tokens

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
@@ -82,57 +82,6 @@ fn test_unexpected_keyword_range() {
 }
 
 #[test]
-fn test_redundant_cast_uses_unnecessary_tag() {
-    let test_files_root = get_test_files_root();
-    let mut interaction = LspInteraction::new();
-    interaction.set_root(test_files_root.path().to_path_buf());
-    interaction.initialize(InitializeSettings {
-        configuration: Some(None),
-        ..Default::default()
-    });
-
-    interaction.server.did_change_configuration();
-
-    interaction.client.expect_configuration_request(2, None);
-    interaction.server.send_configuration_response(
-        2,
-        serde_json::json!([
-            {"pyrefly": {"displayTypeErrors": "force-on"}},
-            {"pyrefly": {"displayTypeErrors": "force-on"}}
-        ]),
-    );
-
-    interaction.server.did_open("redundant_cast.py");
-    interaction.server.diagnostic("redundant_cast.py");
-
-    interaction.client.expect_response(Response {
-        id: RequestId::from(2),
-        result: Some(serde_json::json!({
-            "items": [
-                {
-                    "code": "redundant-cast",
-                    "codeDescription": {
-                        "href": "https://pyrefly.org/en/docs/error-kinds/#redundant-cast"
-                    },
-                    "message": "Redundant cast: `int` is the same type as `int`",
-                    "range": {
-                        "end": {"character": 27, "line": 4},
-                        "start": {"character": 15, "line": 4}
-                    },
-                    "severity": 2,
-                    "source": "Pyrefly",
-                    "tags": [1]
-                }
-            ],
-            "kind": "full"
-        })),
-        error: None,
-    });
-
-    interaction.shutdown();
-}
-
-#[test]
 fn test_unreachable_branch_diagnostic() {
     let test_files_root = get_test_files_root();
     let mut interaction = LspInteraction::new();
@@ -164,8 +113,8 @@ fn test_unreachable_branch_diagnostic() {
                     "code": "unreachable-code",
                     "message": "This code is unreachable for the current configuration",
                     "range": {
-                        "end": {"character": 25, "line": 4},
-                        "start": {"character": 4, "line": 4}
+                        "end": {"character": 12, "line": 1},
+                        "start": {"character": 4, "line": 1}
                     },
                     "severity": 4,
                     "source": "Pyrefly",
--- a/pyrefly/lib/test/lsp/lsp_interaction/test_files/unreachable_branch.py
+++ b/pyrefly/lib/test/lsp/lsp_interaction/test_files/unreachable_branch.py
@@ -1,0 +1,4 @@
+if False:
+    print(1)
+else:
+    1 + 1


kinto0:
I wonder if we should instead mark these with the [`Unnecessary` tag](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag) so editors grey them out instead of just hiding semantic tokens. to me, removing semantic token information doesn't really seem like a feature

edit:
I thought about it more. I think we should use the unnecessary tag for this paths in the IDE but ignore them on the command line. is there a reason I'm not understanding for using the deprecated tag? does pylance do it?

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -302,3 +358,86 @@ impl SemanticTokenBuilder {
...
+            for handler in &try_stmt.handlers {
+                let ExceptHandler::ExceptHandler(handler) = handler;
+                collect_disabled_ranges_from_block(&handler.body, sys_info, reachable, ranges);
+            }
+            collect_disabled_ranges_from_block(&try_stmt.orelse, sys_info, reachable, ranges);
+            collect_disabled_ranges_from_block(&try_stmt.finalbody, sys_info, reachable, ranges);
+        }
+        Stmt::Match(match_stmt) => {
+            for case in &match_stmt.cases {
+                collect_disabled_ranges_from_block(&case.body, sys_info, reachable, ranges);
```
stroxler, line 425: Eventually we might want an exhaustiveness check here - a wildcard case makes everything below it unreachable. Doesn't need to be in this PR

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -302,3 +358,86 @@ impl SemanticTokenBuilder {
...
+            }
+        }
+        _ => {}
```
stroxler, line 428: Eventually I suspect we would want to handle `raise` and `return` here - statements coming after them are unreachable; doesn't need to be in this PR

asukaminato0721:
updated:

I find the proper way.

---

> I wonder if we should instead mark these with the [`Unnecessary` tag](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag) so editors grey them out instead of just hiding semantic tokens. to me, removing semantic token information doesn't really seem like a feature
> 
> edit: I thought about it more. I think we should use the unnecessary tag for this paths in the IDE but ignore them on the command line. is there a reason I'm not understanding for using the deprecated tag? does pylance do it?

There are two things that diagnostics do support, both Unnecessary and Deprecated.

But Semantic tokens, however, only provided `deprecated`. ref: https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#standard-token-types-and-modifiers

asukaminato0721:
Oh, I checked how basedpyright did..

asukaminato0721:
image

image

pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
```diff
@@ -82,7 +82,7 @@ fn test_unexpected_keyword_range() {
 }
 
 #[test]
-fn test_error_documentation_links() {
```
kinto0, line 85: was there a reason you removed this other test?

stroxler (approved):

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -247,6 +262,40 @@ impl SemanticTokenBuilder {
...
+                            self.push_if_in_range(
+                                name.range,
+                                SemanticTokenType::VARIABLE,
+                                Vec::new(),
+                            );
+                        }
+                    }
+                }
+                x.recurse(&mut |x| self.process_stmt(x));
+            }
```
grievejia, line 298: @asukaminato0721 Would you mind clarifying why this block of code was added? Trying to retroactively understand the logic here.

pyrefly/lib/test/lsp/semantic_tokens.rs
```diff
@@ -174,6 +174,47 @@ token-type: method
     );
 }
 
+#[test]
+fn deprecated_token_for_disabled_branch() {
```
grievejia, line 178: @asukaminato0721 I'm also having a hard time understanding what behavior is being tested here. Any chance you could help clarifying?

pyrefly/lib/state/semantic_tokens.rs
kinto0: - remove semantic tokens for unreachable code
- [apparently based pyright does it](https://github.com/facebook/pyrefly/pull/1391#issuecomment-3460153362)
- there's no deprecated semantic token so this is what was done

I am not sure how useful it is to remove semantic tokens. I think the useful part is the diagnostic.

pyrefly/lib/test/lsp/semantic_tokens.rs
kinto0: test should probably be renamed. it's testing the lack of semantic tokens on the disabled branch

pyrefly/lib/state/semantic_tokens.rs
grievejia: I'm still confused why this would "remove" semantic tokens -- quite the contrary we seems to be adding semantic tokens when these AST nodes falls into disabled ranges.

Is the condition here supposed to be `if !self.is_disabled(...)` (instead of `if self.is_disabled(...)`)? Even then I don't think it'll work as intended given that the binding-based scan would always push these names. If this functionality is not important perhaps we could just remove this part of the logic?

pyrefly/lib/test/lsp/semantic_tokens.rs
grievejia: In that case aren't `Legacy` on line 4 and `value` on line 5 expected to be absent from the full token list?

pyrefly/lib/state/semantic_tokens.rs
asukaminato0721: The binding pass skips bodies that sys_info proves unreachable (e.g., version guards, if False).

Because semantic tokens for definitions normally come from bindings (SemanticTokenBuilder::process_key), every class/function/assignment that lives in one of these abandoned branches stopped producing tokens entirely.

To restore highlighting for that unreachable code, SemanticTokenBuilder::process_stmt now walks the AST and, only when self.is_disabled is true, pushes ad-hoc tokens for the statement’s defining name

The disabled_ranges feeding is_disabled are computed once from the AST using the same SysInfo evaluation logic, so we only synthesize tokens for ranges the binder skipped. Reachable statements still get their tokens from the bindings path, so using !self.is_disabled would double‐emit and we’d have to deduplicate everything.

pyrefly/lib/test/lsp/semantic_tokens.rs
asukaminato0721: It just ensures the tokens exist so editors can still style the dead code (the diagnostics system already marks the whole range as “unreachable”

pyrefly/lib/state/semantic_tokens.rs
grievejia: Thanks for the detailed explanation! I hadn’t realized that bindings aren’t populated for unreachable branches, which makes the logic much clearer now.

However, I’m noticing that this approach only allows us to "recover" certain types of semantic tokens from unreachable branches. For instance, if names are imported or defined within those branches, or if we reference those names, we still won’t be able to apply semantic coloring to them, right?

Rather than just masking the symptoms, I wonder if it would be better to address the root cause directly—for example, by not skipping bindings for unreachable branches in the first place. I realize this would be a much more involved change, so I’m not suggesting we tackle it immediately.

For context, I’m currently exploring changes to the semantic token logic to avoid relying on the binding walk, and instead use a pure AST walk combined with binding lookups. This investigation is what led me to dig deeper into this part of the code.

pyrefly/lib/test/lsp/semantic_tokens.rs
grievejia: I get it now with the explanation above. I'll have to think about how to handle this test -- I may preserve it if it's easy to keep the behavior during the refactoring, but if it gets annoying to maintain I may also consider removing this temporarily since the behavior is less about semantic token logic itself but more about unreachable branch handling.

Issue #1173, opened 2025-09-25
VSCode extention import auto complete suggestions include hidden dot .directories


gmetzker-4c:
## Issue 

If I request auto completions on a class where there is an import missing I generally get good list of options from Pylance.  With pyrefly I've noticed several issues:

- The list will sometimes show weird suggestions from hidden directories such as `.tox` or `.serverless`
- Secondary Issue: It shows imports for the deepest module (typically where the class is implemented) rather than the top level module that exposes it.  Example:  For pydantic types, if I request import auto-complete on `BeforeValidator` it recommends `from pydantic.functional_validators import BeforeValidator` instead of `from pydantic import BeforeValidator`

I'm wondering if I have something configured wrong, or if this is a bug.

Note that I've tried adding these to `project-excludes` and it doesn't fix the issue.  Here is what I've tried in my `pyproject.toml`.

```toml
[tool.pyrefly]
project-includes = ["**/*"]

# project-includes = [
#     "foresee/**/*"
# ]

project-excludes = [
    "**/node_modules",
    "**/__pycache__",
    "**/*venv/**/*",
    "**/.tox",
    "**/.tox/**/*",
    "**/.serverless",
    "**/.serverless/**/*",
    "**/.ruff_cache",
    "**/builtins.pyi",
    # ".tox/**/*",
    # ".serverless/**/*",
]


```

Here are what the auto-suggested imports look like:

<img width="1517" height="1021" alt="Image" src="https://github.com/user-attachments/assets/0cd98973-f05c-49b0-9715-6eb15b39cd61" />

<img width="1481" height="650" alt="Image" src="https://github.com/user-attachments/assets/f59f0d4b-4fab-4b8c-ad2f-1a624ddb7fcd" />

## Other attempted workarounds
Here are several work arounds I've tried

1. Adding explicit project-excludes for `.tox` and `.serverless`
- With configuration in root `pyrefly.toml`
- With configuration in root `pyproject.toml`
**Effect**: Issue persists

2. Adding more restrictive includes to `project-includes`
**Effect**: Issue persists

3. Remove all pyrefly configuration.
**Effect**: `pydantic` imports are at the top in some cases, but I still get a long list of `.tox...` suggestions in the list

4. Set the `Disable Language Services` option in Pyrefly
**Effect**: In this case I no longer get any import suggestions.  This is strange as I would expect the default Pylance suggestions to kick back in.
[...]

PR #1430, opened 2025-10-30
test VSCode extention import auto complete suggestions include hidden dot .directories

asukaminato0721:
check #1173 is fixed.

~~Filtered hidden filesystem entries when gathering module prefixes so .venv-style folders no longer surface in import completions~~ fixed by other commit.

Added an LSP regression test

### DIFF
--- a/pyrefly/lib/module/finder.rs
+++ b/pyrefly/lib/module/finder.rs
@@ -213,12 +213,6 @@ fn find_one_part_prefix<'a>(
                 let file_name = path.file_name().and_then(|n| n.to_str());
 
                 if let Some(name) = file_name {
-                    // Skip hidden filesystem entries (e.g. `.venv`, `.git`) so they don't
-                    // surface in import completion lists. These entries are not valid module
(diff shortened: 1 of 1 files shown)


rchen152:
Hey @asukaminato0721! [f44e8ce](https://github.com/facebook/pyrefly/commit/f44e8ce96b686172e85f8f044ae216d312866e7f) filtered out dot directories when populating workspace files, which I was expecting to prevent them from showing up in autocomplete suggestions as well. Are you still seeing these suggestions even after that commit?

asukaminato0721:
> Hey @asukaminato0721! [f44e8ce](https://github.com/facebook/pyrefly/commit/f44e8ce96b686172e85f8f044ae216d312866e7f) filtered out dot directories when populating workspace files, which I was expecting to prevent them from showing up in autocomplete suggestions as well. Are you still seeing these suggestions even after that commit?

Yeah, seems fixed now. I only leave the test for sure.

stroxler (approved):

PR #1435, opened 2025-10-30
refresh tokens on invalidations

kinto0:
Summary:
There have been reports of semantic tokens not appearing sometimes. I have trouble reproducing this entirely, but I do notice the following:
- open file
- semantic tokens appear
- interpreter change takes effect
- imports that now are found do not have tokens until i switch windows and return

this is because we invalidated the file from a config change, but clients have no idea the content is different. we must tell the editor that the content has changed with [this message](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_refreshRequest). If a client supports it, we send it. it takes no parameters and will only re-query the minimum it needs.

to fix other parts of semantic tokens not appearing, it might be necessary to do something like this (D85099658). but I haven't been able to reproduce this issue so I'm not sure it's needed.

Reviewed By: yangdanny97

Differential Revision: D85100320


yangdanny97 (approved):

PR #1437, opened 2025-10-31
not expand TypeAlias in reveal_type

asukaminato0721:
part of #1219.

Added TypeAlias::body_type/runtime_value and used them across the solver so we can keep alias names for printing while still acting on their underlying types when needed (attribute access, class bases, isinstance/issubclass, subset checks). Scoped type aliases still behave as TypeAliasType at runtime; legacy aliases now behave like the types they wrap.

Refined isinstance handling so union aliases (e.g., X = B | C) are accepted as class info, and ensured scoped aliases still error.

Allowed type-alias bases when they’re legacy aliases and kept the explicit error for scoped aliases.


yangdanny97:
Can you elaborate on your proposed fix? I don't think this PR addresses the root cause of #1219 - the type aliases are expanded in `untype_opt` & the information those types came from an alias is lost.

asukaminato0721:
image

at least less noisy now.

yangdanny97:
I'll import this and see how we can merge it with my in-progress implementation. I'd also like to run it on mypy primer and see what the delta is.

yangdanny97:
I have some doubts on the overall approach, in particular the part where we mutate the type - this type of side effect could lead to non-determinism depending on the order that bindings are solved in.

yangdanny97:
i'm going to close the PR in favor of the other solution discussed in the original issue. @ndmitchell mentioned that after we move to a type heap a more general solution will be easier to do.

PR #1443, opened 2025-10-31
Unify docs on `displayTypeErrors`

kv9898:
Currently there are consistencies between the website doc and the extension doc (I think the website doc is more "correct", as the setting is now `displayTypeErrors` not `disableTypeErrors`). This PR fix these inconsistencies.


ndmitchell (approved):

PR #1446, opened 2025-11-01
fix docstrings on functions written in C

### LINKED ISSUE #1076: docstrings on functions written in C
### Describe the Bug

`str.isupper()` is written in C and there is not a python implementation to find the docstring in. for these functions, we should ask the python interpreter for `str.__doc__`

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1076

image

use https://github.com/DetachHead/basedpyright/blob/main/build/py3_8/generate_docstubs.py to gen code during build and bundle into lsp.

### DIFF
--- a/crates/pyrefly_bundled/generate_docstubs.py
+++ b/crates/pyrefly_bundled/generate_docstubs.py
@@ -16,7 +16,7 @@ from docify import main as docify  # type: ignore
 def main(*, overwrite: bool):
     """
     adds docstrings to the stubs located in `../stubdocs` for any modules that are compiled,
-    because otherwise pyright would have no way to see the docstrings because there's no source
+    because otherwise pyrefly would have no way to see the docstrings because there's no source
     code.
 
     note that it only generates stubs for modules and objects that exist on your current python


asukaminato0721:
The CI now takes too long...

crates/pyrefly_bundled/generate_docstubs.py
```diff
@@ -0,0 +1,42 @@
...
+from pathlib import Path
+from shutil import copytree, rmtree
+
+from docify import main as docify  # type: ignore
+
+
+def main(*, overwrite: bool):
+    """
+    adds docstrings to the stubs located in `../stubdocs` for any modules that are compiled,
+    because otherwise pyright would have no way to see the docstrings because there's no source
```
kinto0:
```suggestion
    because otherwise pyrefly would have no way to see the docstrings because there's no source
```

crates/pyrefly_bundled/generate_docstubs.py
```diff
@@ -0,0 +1,42 @@
...
+    code.
+
+    note that it only generates stubs for modules and objects that exist on your current python
+    version and OS.
```
kinto0, line 23: or CI's version/OS?

crates/pyrefly_bundled/generate_docstubs.py
```diff
@@ -0,0 +1,42 @@
...
+#   "docify",
+# ]
+# ///
+
+from __future__ import annotations
+
+from pathlib import Path
+from shutil import copytree, rmtree
+
+from docify import main as docify  # type: ignore
```
kinto0, line 13: which type checker are you ignoring diagnostics on here? is it pyrefly? can we vendor docify so we don't need to pip install on the client?

crates/pyrefly_bundled/build.rs
```diff
@@ -40,6 +41,15 @@ fn main() -> Result<(), std::io::Error> {
     // changes in the entire typeshed dir.
     println!("cargo::rerun-if-changed=third_party/typeshed_metadata.json");
 
+    println!("cargo::rerun-if-changed=crates/pyrefly_bundled/generate_docstubs.py");
+
+    let status = Command::new("uv")
```
kinto0, line 46: if we bundle docify, can we avoid needing UV on the host?

asukaminato0721:
How this script works:

It only runs on CI, writes the doc to stub, then bundles with the Pyrefly binary.

The type ignore is ... because this script needs `docify`, but only requires it at build time, so I use uv to run it, but my type checker will warn about it.

docify is a single script that relies on libcst.

https://github.com/AThePeanut4/docify/blob/main/docify.py

crates/pyrefly_bundled/generate_docstubs.py
asukaminato0721: yes

crates/pyrefly_bundled/build.rs
asukaminato0721: yes, uv is for the dependency.

crates/pyrefly_bundled/generate_docstubs.py
asukaminato0721: Actually we don't need it on the client.

kinto0:
awesome! hit ready for review when you're ready, I'm excited to get this in. 

one thing I will change is add a buck target for the build script so we can run it through buck (and maybe remove the lint suppression?)

yangdanny97 (changes requested):
IMO better to run this bundling process whenever we pull in updates for the typeshed stubs, as opposed to every time we build. I believe that is what ty does as well.

That already happens using a Python script, so you could add your docstring copying logic to that directly.

https://github.com/facebook/pyrefly/blob/fdf8f221042cfc1e543d17d39a87ac481e5d546f/crates/pyrefly_bundled/update.py#L4

kinto0:
thank you for finding this solution! since this is high priority for the beta on monday, @grievejia will move this to upgrade on typeshed upgrade + merge

Issue #955, opened 2025-08-17
Top Level Issue: Make long error codes more readable in hover cards and CLI


asukaminato0721:
### Describe the Bug

Too long, can't read.

```py
ERROR No matching overload found for function `pandas.io.parsers.readers.read_csv` [no-matching-overload]
   --> /home/runner/work/dify/dify/api/services/annotation_service.py:331:29
    |
331 |             df = pd.read_csv(file, dtype=str)
    |                             ^^^^^^^^^^^^^^^^^
    |
  Possible overloads:
[...]

PR #1447, opened 2025-11-01
fix emit unused parameter #955

asukaminato0721:
part of #955

callable_infer now records not only the expected parameter type for each call argument, but also which parameter index the argument matched.

Overload “No matching overload” diagnostics use that real argument-to-parameter mapping instead of reconstructing it from positional counts / keyword names.

If the failure is missing required args, show the full required signature so users can see what to pass, e.g. z / w.
If the failure is wrong argument type, truncate unrelated optional params with ... so the relevant parameters stand out.

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -14,12 +14,14 @@ use pyrefly_types::callable::ArgCount;
 use pyrefly_types::callable::ArgCounts;
 use pyrefly_types::callable::Param;
 use pyrefly_types::meta_shape_dsl::ShapeTransform;
+use pyrefly_types::callable::ParamList;
 use pyrefly_types::tuple::Tuple;
 use pyrefly_types::types::TArgs;
 use pyrefly_util::gas::Gas;
 use pyrefly_util::owner::Owner;
 use pyrefly_util::prelude::SliceExt;
 use pyrefly_util::prelude::VecExt;
+use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
 use vec1::Vec1;
@@ -461,9 +463,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         .unwrap_or_else(|| overload.1.signature.clone()),
                     None => overload.1.signature.clone(),
                 };
-                let signature = self
-                    .solver()
-                    .for_display(self.heap.mk_callable_from(signature));
+                let signature =
+                    self.format_overload_signature_for_call(&signature, &args, &keywords);
                 details.push(format!("{signature}{suffix}"));
             }
             // We intentionally discard closest_overload.call_errors. When no overload matches,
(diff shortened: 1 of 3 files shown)


rchen152:
I'm a little worried about how long this would make the [no-matching-overload] error message. I agree that the line breaks make the individual signatures more readable, but this could easily turn, say, a 10-line message into a 100-line one. And I assume the syntax highlighting doesn't show up unless you do an extra step of piping to bat, so it wouldn't be there by default to help with parsing the output?

I wonder if it would be better to instead implement the other suggestion in https://github.com/facebook/pyrefly/issues/955, hiding unrelated fields. That could make these messages more readable without increasing their length.

rchen152 (changes requested):
Marking this as "Changes requested", since I don't think we can merge it as-is. My main concern is about error message length, but looking more closely at the PR, it also puts the `# [closest match]` comment in an odd place, so that would need to be fixed as well.

asukaminato0721:
this will be much shorter

```
 --> /tmp/tmp/main.py:9:2
  |
9 | f(1, 2)
  |  ^^^^^^
  |
  Possible overloads:
  (x: int, y: int, ...) -> int [closest match]
  (x: str, y: str, ...) -> str
 INFO 1 error
```

yangdanny97:
i'm gonna re-trigger CI

pyrefly/lib/test/overload.rs
```diff
@@ -1403,6 +1403,24 @@ g(x=1, y="hello")  # E: No matching overload found for function `g` called with
...
+    r#"
+from typing import overload
+
+@overload
+def f(x: int, y: int, z: int) -> int: ...
+@overload
+def f(x: str, y: str, z: str, w: str) -> str: ...
+def f(x, y, z=None, w=None): return x
+
+f(1, 2)  # E: (x: int, y: int, ...) -> int [missing required arguments] # !E: z: # !E: w:
```
rchen152: Omitting `z` and `w` here doesn't seem like an improvement. Since the error is that they're missing, we should show them so that the user can see what arguments they need to pass in and in what order.

IMO the obvious case in which it would be useful to omit parameters would a wrong argument *type* error, in which the user has passed in the right number of arguments but one or more has the wrong type, so we want the error message to highlight those.

pyrefly/lib/alt/overload.rs
```diff
@@ -778,4 +782,163 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                "{}",
+                self.solver()
+                    .for_display(self.heap.mk_callable_from(signature.clone()))
+            )
+        };
+        let Params::List(params) = &signature.params else {
+            return full_signature();
+        };
+
+        let mut positional_remaining = args
```
rchen152: Reconstructing the argument->parameter mapping for error reporting seems inefficient and error-prone. We already return the mapping from `callable_infer`, so it would be nice if we could reuse it: https://github.com/facebook/pyrefly/blob/c12e9d60a04416abfb71c41c0db7b857983036e0/pyrefly/lib/alt/callable.rs#L1238

rchen152 (changes requested):

kinto0:
is this still relevant with our current error messages?

rchen152:
Sorry for the delayed response here! I ended up taking a different approach to this issue, in https://github.com/facebook/pyrefly/pull/3782. The key differences are that that PR has a narrower focus (just omits irrelevant parameter types, doesn't try to do any handling of other kinds of call errors), reuses the existing signature display code, tracks parameters by name instead of index, and filters all overload signatures rather than just the closest one. I'm going to close this PR now, but thank you for all of your contributions. Apologies again for sitting on this for so long.

Issue #995, opened 2025-08-25
support show parameter name


asukaminato0721:
### Describe the Bug

As the title says. Pylance has this.

yangdanny97:
Sorry, could you please clarify what this means? Showing parameter names where?

asukaminato0721:
<img width="524" height="200" alt="Image" src="https://github.com/user-attachments/assets/f1aac8bd-eb6c-4def-99ca-bf367c635afa" />

yangdanny97:
understood, thanks!

kinto0:
duplicate of #789

samwgoldman:
I think this is different. This appears to be a request for inlay hints for positional arguments to named parameters. The linked issue is for hover type on keyword arguments.

asukaminato0721:
<img width="567" height="92" alt="Image" src="https://github.com/user-attachments/assets/e704e2ec-679d-4926-8059-b2aa272798d5" />

need this line to make it work.

```json
"python.analysis.inlayHints.callArgumentNames": "all", // off / partial / all
```

or

```
    "python.pyrefly.analysis.inlayHints": {
        "callArgumentNames": "all"
    },
```

PR #1449, opened 2025-11-01
Added a schema entry for inlay hint

asukaminato0721:
fix https://github.com/facebook/pyrefly/issues/995#issuecomment-3476948890

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -168,6 +168,38 @@
                             "default": false
                         }
                     }
+                },
+                "python.pyrefly.analysis.inlayHints": {
+                    "type": "object",
+                    "default": {},
+                    "description": "Configure inlay hints provided by Pyrefly in the editor.",
+                    "properties": {
+                        "callArgumentNames": {
+                            "type": "string",
+                            "enum": [
+                                "off",
+                                "partial",
+                                "all"
+                            ],
+                            "default": "off",
+                            "description": "Show parameter name hints for positional arguments (\"off\", \"partial\", or \"all\")."
+                        },
+                        "functionReturnTypes": {
+                            "type": "boolean",
+                            "default": true,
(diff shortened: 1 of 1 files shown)


lsp/package.json
```diff
@@ -168,6 +168,38 @@
                             "default": false
                         }
                     }
+                },
+                "python.pyrefly.analysis.inlayHints": {
```
kinto0, line 172: does it work like this? according to the [docs](https://github.com/facebook/pyrefly/blob/f3d247769ecf28bb83f9466610f7390f4001a543/website/docs/IDE-features.mdx#L209), it needs `python.pyright.analysis.inlayHints`. could we make sure the documentation is up to date?

kinto0 (changes requested):

asukaminato0721:
did some tests

```
    "python.pyrefly.analysis.inlayHints.callArgumentNames": "all",
```

image

```
"python.pyrefly.analysis.inlayHints.callArgumentNames": "off",
```

no inlay hint

```
"python.pyright.analysis.inlayHints.callArgumentNames": "all",
```

no

So the doc is a bit outdated?

kinto0:
hmm, I think this is a bug. it seems like analysis will always be handled since the addition of [pyrefly.analysis.disabledLanguageService](https://github.com/facebook/pyrefly/blob/9514c56cabb2bfc0e058cce57af4cf0a28342716/lsp/package.json#L110)

but we don't use the [default analysis config ](https://github.com/facebook/pyrefly/blob/9514c56cabb2bfc0e058cce57af4cf0a28342716/pyrefly/lib/lsp/non_wasm/workspace.rs#L362)if we have handled one

asukaminato0721:
Thanks for the clarification.

PR #1450, opened 2025-11-02
fix [Feature Request] Support reportUnusedParameter from pyright #1403

### LINKED ISSUE #1403: [Feature Request] Support reportUnusedParameter from pyright
### Describe the Bug

For my usecase, specifically when using the LSP in VSCode, it would be nice to see the unused variable greyed out :)

Pyright:

Image

Pyrefly:

Image

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1403

Introduced per-function tracking of declared parameters and their usage so we can determine which ones are never read

Marked parameters as “allowed unused” for common conventions (leading _, self, cls, *args/**kwargs), registered them with the scope, and flagged reads via lookup to flip the usage bit

Returned the collected unused parameters from function binding, emitted unused-parameter diagnostics, and ensured they surface as DiagnosticTag::UNNECESSARY in the LSP payload

Added the new ErrorKind::UnusedParameter, wired Pyright migration support for reportUnusedParameter, and documented the error kind

Covered the behavior with an LSP interaction test and fixture that assert the diagnostic content and tag

### DIFF
--- a/website/docs/error-kinds.mdx
+++ b/website/docs/error-kinds.mdx
@@ -1034,14 +1034,3 @@ async def bar():
     await foo()  # ok
     x = foo()  # ok
 ```
-
-## unused-parameter
-
-This diagnostic is emitted when a function parameter's value is never read inside the
-function body. Prefix the parameter name with `_` (for example, `_unused`) or use
-conventions such as `self`/`cls` to indicate intentional non-use.
-
-```python
-def greet(name: str, unused: int) -> str:
-    return f"Hello {name}"
-```
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -244,8 +244,6 @@ pub enum ErrorKind {
     UntypedImport,
     /// Result of async function call is never used or awaited
     UnusedCoroutine,
-    /// A function parameter value is never used within the function body.
-    UnusedParameter,
 }
 
 impl std::str::FromStr for ErrorKind {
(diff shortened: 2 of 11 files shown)


asukaminato0721:
problem: should

```py
def f(x: int):
    ...
```

be flagged as unused if `...` is used in the body?

yangdanny97:
No, I think if the body is something like `...` then we shouldn't flag it. If we wanted to be more sophisticated, we could skip this for protocol methods & abstract methods.

asukaminato0721:
another interesting pattern

```
def f(x: int):
    x = 1
    print(x)
```

x is also not "used".

asukaminato0721:
Make the default configuration ignore so the test can pass...

Also if the function body only has `pass`, ignore it like `...`.

yangdanny97:
```
def f(x: int):
    x = 1
    print(x)
```

For this, I think it's because Pyrefly knows that `x` is immediately assigned to, so the original value passed via the parameter is not used.

I'm not sure if we should flag it or not, this behavior isn't really specified anywhere. There's an argument to be made for aligning with Pyright here and we should be careful about false-positives, but in this case our analysis is arguably more correct.

website/docs/error-kinds.mdx
```diff
@@ -1034,3 +1034,14 @@ async def bar():
     await foo()  # ok
     x = foo()  # ok
 ```
+
+## unused-parameter
+
+This diagnostic is emitted when a function parameter's value is never read inside the
+function body. Prefix the parameter name with `_` (for example, `_unused`) or use
```
yangdanny97: If a named-parameter is renamed, that could cause incompatible override errors.

Pyright actually does not report this as a diagnostic, because of the number of false-positives it gives. see https://github.com/microsoft/pyright/issues/3473#issuecomment-1130164553

yangdanny97:
I don't think we should have a dedicated error kind and emit this in the CLI, due to number of false positives it could give.

I would prefer for it to be a "fake" diagnostic that only gets emitted by the language server, like we do for unreachable code: https://github.com/facebook/pyrefly/blob/ffd27c1e19f4e1a67ff06c6c6ba970df3932cd2b/pyrefly/lib/lsp/non_wasm/server.rs#L2166

That way, it doesn't cause CI checks to fail but it will still give the greyed-out text in the IDE.

Does that sound reasonable to you?

stroxler (approved):

yangdanny97:
Thanks! I patched this internally to ignore a few more kinds of functions:
- abstractmethod
- overload
- override
- return NotImplemented
- raise NotImplementedError(...)

PR #1452, opened 2025-11-02
Implement multiple inheritance checks for instance methods

### LINKED ISSUE #943: Incompatible multiple inheritance checks

fangyi-zhou:
Follow up from #1196. Closes #943.

When getting the field type from the parent field, we now check for instance method types first before using the raw types.

When using the raw types, the type parameters are not initialised, hence causing issues with functions that use type variables during subtyping checks. This diff adds a regression test for these cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2081,7 +2081,7 @@ dependencies = [
 
 [[package]]
 name = "pyrefly"
-version = "0.39.4"
+version = "0.40.0"
 dependencies = [
  "anstream",
  "anyhow",
@@ -2133,7 +2133,7 @@ dependencies = [
 
 [[package]]
 name = "pyrefly_build"
-version = "0.39.4"
+version = "0.40.0"
 dependencies = [
  "anyhow",
  "dupe",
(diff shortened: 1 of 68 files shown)


fangyi-zhou:
cc @yangdanny97 who reviewed previous diffs on this

PR #1456, opened 2025-11-03
Verify overloaded method name displays in hover trace

### LINKED ISSUE #1416: hover on operator missing the magic method name and highlight
### Describe the Bug

continue to #994

current

Image

I want it can be

```py
def __matmul__(
    self: ndarray[tuple[Any, ...], dtype[_NumberT]],
    other: _NestedSequence[_SupportsArray[dtype[numpy.bool[builtins.bool]]]] | _NestedSequence[builtins.bool] | _SupportsArray[dtype[numpy.bool[builtins.bool]]] | builtins.bool,
    /
) -> ndarray[tuple[Any, ...], dtype[_NumberT]]
```

```py
import numpy as np

a = np.array([1])
a @ a
```

### Sandbox Link

_No response_

v0.39.2

aaron-ang:
Close #1416.

Apparently, the issue of the method name not displaying has been fixed. I added a test case to verify this, and also refactored `OverloadedCallee` to an enum to better distinguish between resolved (show one) vs candidate (show all) traces.

Screenshot 2025-11-10 at 7 04 03 PM

### DIFF
--- a/pyrefly/lib/alt/answers.rs
+++ b/pyrefly/lib/alt/answers.rs
@@ -76,10 +76,15 @@ pub struct Index {
 }
 
 #[derive(Debug)]
-struct OverloadedCallee {
-    all_overloads: Vec<Callable>,
-    closest_overload: Callable,
-    is_closest_overload_chosen: bool,
+enum OverloadedCallee {
+    Resolved {
+        callable: Callable,
+    },
+    Candidates {
+        all: Vec<Callable>,
+        closest: Callable,
+        is_closest_chosen: bool,
+    },
 }
 
 #[derive(Debug, Default)]
@@ -570,13 +575,18 @@ impl Answers {
 
     pub fn get_chosen_overload_trace(&self, range: TextRange) -> Option<Type> {
         let lock = self.trace.as_ref()?.lock();
-        let overloaded_callee = lock.overloaded_callees.get(&range)?;
-        if overloaded_callee.is_closest_overload_chosen {
-            Some(self.deep_force(Type::Callable(Box::new(
-                overloaded_callee.closest_overload.clone(),
-            ))))
-        } else {
-            None
+        match lock.overloaded_callees.get(&range)? {
+            OverloadedCallee::Resolved { callable } => {
+                Some(self.deep_force(Type::Callable(Box::new(callable.clone()))))
+            }
+            OverloadedCallee::Candidates {
+                closest,
+                is_closest_chosen,
+                ..
+            } if *is_closest_chosen => {
+                Some(self.deep_force(Type::Callable(Box::new(closest.clone()))))
+            }
+            _ => None,
         }
     }
 
@@ -586,23 +596,13 @@ impl Answers {
         range: TextRange,
     ) -> Option<(Vec<Callable>, Option<usize>)> {
         let lock = self.trace.as_ref()?.lock();
-        let overloaded_callee = lock.overloaded_callees.get(&range)?;
-        let chosen_overload_index =
-            overloaded_callee
-                .all_overloads
-                .iter()
-                .enumerate()
-                .find_map(|(index, signature)| {
-                    if signature == &overloaded_callee.closest_overload {
-                        Some(index)
-                    } else {
-                        None
-                    }
-                });
-        Some((
-            overloaded_callee.all_overloads.clone(),
-            chosen_overload_index,
-        ))
+        match lock.overloaded_callees.get(&range)? {
+            OverloadedCallee::Resolved { callable } => Some((vec![callable.clone()], Some(0))),
+            OverloadedCallee::Candidates { all, closest, .. } => {
+                let chosen_index = all.iter().position(|signature| signature == closest);
+                Some((all.clone(), chosen_index))
+            }
+        }
     }
 }
 
@@ -628,18 +628,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         &self.current().solver
     }
 
-    pub fn record_overload_trace_from_type(&self, loc: TextRange, ty: Type) {
+    pub fn record_resolved_trace(&self, loc: TextRange, ty: Type) {
         if let Some(trace) = &self.current().trace
             && let Some(callable) = ty.to_callable()
         {
-            trace.lock().overloaded_callees.insert(
-                loc,
-                OverloadedCallee {
-                    all_overloads: vec![callable.clone()],
-                    closest_overload: callable,
-                    is_closest_overload_chosen: true,
-                },
-            );
+            trace
+                .lock()
+                .overloaded_callees
+                .insert(loc, OverloadedCallee::Resolved { callable });
         }
     }
 
@@ -655,13 +651,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         if let Some(trace) = &self.current().trace {
             trace.lock().overloaded_callees.insert(
                 loc,
-                OverloadedCallee {
-                    all_overloads: all_overloads
-                        .into_iter()
-                        .map(|func| (*func).clone())
-                        .collect(),
-                    closest_overload: closest_overload.clone(),
-                    is_closest_overload_chosen,
+                OverloadedCallee::Candidates {
+                    all: all_overloads.into_iter().cloned().collect(),
+                    closest: closest_overload.clone(),
+                    is_closest_chosen: is_closest_overload_chosen,
                 },
             );
         }
--- a/pyrefly/lib/test/lsp/hover.rs
+++ b/pyrefly/lib/test/lsp/hover.rs
@@ -168,6 +168,41 @@ x: int = 5  # pyrefly: ignore[bad-return]
 }
 
 #[test]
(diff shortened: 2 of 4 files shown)


pyrefly/lib/alt/answers.rs
```diff
@@ -79,6 +79,7 @@ pub struct Index {
 struct OverloadedCallee {
     all_overloads: Vec<Callable>,
     closest_overload: Callable,
```
kinto0: after you change, do we still need this? should we only keep one?

pyrefly/lib/alt/answers.rs
```diff
@@ -571,7 +572,9 @@ impl Answers {
     pub fn get_chosen_overload_trace(&self, range: TextRange) -> Option<Type> {
         let lock = self.trace.as_ref()?.lock();
         let overloaded_callee = lock.overloaded_callees.get(&range)?;
-        if overloaded_callee.is_closest_overload_chosen {
+        if let Some(ty) = &overloaded_callee.closest_overload_type {
```
kinto0: do we want to return the overload type if not `is_closest_overload_chosen`?

pyrefly/lib/alt/answers.rs
```diff
@@ -79,6 +79,7 @@ pub struct Index {
 struct OverloadedCallee {
     all_overloads: Vec<Callable>,
     closest_overload: Callable,
+    closest_overload_type: Option<Arc<Type>>,
```
kinto0: I wonder if the reason we didn't store this the whole time was the memory consequence. but I guess the arc mitigates a lot of this

pyrefly/lib/alt/answers.rs
```diff
@@ -661,6 +665,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         .map(|func| (*func).clone())
                         .collect(),
                     closest_overload: closest_overload.clone(),
+                    closest_overload_type: None,
```
kinto0: what downsides does skipping it here have?

pyrefly/lib/test/lsp/hover_type.rs
```diff
@@ -192,7 +192,7 @@ result = a == b
 # main.py
 8 | result = a == b
                 ^
-Hover Result: `(self: My, other: object) -> bool`
+Hover Result: `BoundMethod[My, (self: My, other: object) -> bool]`
```
kinto0: it doesn't seem like it displays overloaded method name for everything. i'd expect to see `__eq__` here

aaron-ang: The `hover_type` test suite reports the type only, so the code changes made this diff.
https://github.com/facebook/pyrefly/blob/f3d247769ecf28bb83f9466610f7390f4001a543/pyrefly/lib/test/lsp/hover_type.rs#L16-L22

aaron-ang: The test cases in `hover.rs` verify the display.

pyrefly/lib/alt/answers.rs
aaron-ang: The issue is that these objects are created from two methods: one that supplies the Callable directly, and another that supplies the Type. I could try to refactor up the call chain to save only the type.

pyrefly/lib/alt/answers.rs
aaron-ang: from the initialization in `record_overload_trace_from_type`, `is_closest_overload_chosen` will always be true if overload type is Some.

kinto0: let's remove is_closest_overload_chosen then and keep closest_overload_type

pyrefly/lib/alt/answers.rs
kinto0: how feasible would it be to do that? removing duplication like this is important for memory use / code cleanliness

pyrefly/lib/alt/answers.rs
kinto0: why'd this an arc? I only see it used in deep_force once. I don't think this helps performance at all?

aaron-ang: Mainly to reduce memory usage, since I think `Type` can be quite large. I see that other usages of `Type` in the file are also wrapped in `Arc`, and are used in `deep_force` as well.

pyrefly/lib/alt/answers.rs
aaron-ang: `closest_overload_chosen` is used to filter in `get_chosen_overload_trace`; while in `get_all_overload_trace`, we search for the closest overload even if it is not chosen.

aaron-ang:
@kinto0  I've changed `OverloadCallee` into an enum to decouple the two cases. Tracing the call stack, `record_overload_trace_from_type` is used to insert resolved dunder methods, so I named one of the variants as such.

pyrefly/lib/alt/answers.rs
```diff
@@ -76,10 +75,15 @@ pub struct Index {
 }
 
 #[derive(Debug)]
-struct OverloadedCallee {
-    all_overloads: Vec<Callable>,
-    closest_overload: Callable,
-    is_closest_overload_chosen: bool,
+enum OverloadedCallee {
+    Resolved {
```
kinto0, line 80: so an overloaded callee is one of
- Overload
- Resolved

if it's resolved is it not an overload? this is a little confusing

pyrefly/lib/alt/answers.rs
kinto0: an arc is only more efficient the contents are cloned, but I don't think I see any clones of this (I could be missing them?)

edit: it looks like in the new code we do clone it when we return it from overloaded_callees

pyrefly/lib/alt/answers.rs
```diff
@@ -76,10 +75,15 @@ pub struct Index {
...
+        ty: Arc<Type>,
```
kinto0: in general, every time we export this `Type` type, there's a lot of overhead. we probably want to only export the necessary part. in this case, you only need the callable + the name. in another change, we might also want "type" to make the method change but I see that as another PR to "make hover for methods say (method) instead of (attribute). 

thoughts @stroxler ?

pyrefly/lib/alt/answers.rs
aaron-ang: It is indeed confusing; perhaps I could name it better. In places where the struct is constructed using `Type`, I see these comments:
https://github.com/facebook/pyrefly/blob/d22383b743802996f5abbfc339fd2de1c768e6ae/pyrefly/lib/alt/call.rs#L473-L475
https://github.com/facebook/pyrefly/blob/d22383b743802996f5abbfc339fd2de1c768e6ae/pyrefly/lib/alt/call.rs#L506-L507
So for now, dunders / resolved methods are considered overloads.

aaron-ang:
rebased

kinto0:
looks like some tests are still failing

aaron-ang:
> looks like some tests are still failing

@kinto0 I have fixed my test cases.

pyrefly/lib/test/lsp/hover.rs
```diff
@@ -167,6 +167,41 @@ x: int = 5  # pyrefly: ignore[bad-return]
     assert!(report.contains("No errors suppressed"));
 }
 
+#[test]
+fn hover_over_overloaded_binary_operator_shows_dunder_name() {
```
kinto0, line 171: this test actually passes before your change.... maybe this feature is done in the meantime or the bug report was incorrect

what if you split this into two PRs:
- just this test which closes #1416
-  follow-up + new issue for the rest of what this diff does (expose type, show (method) different from (attribute)) on hover

pyrefly/lib/test/lsp/hover_type.rs
kinto0: I still think this is probably a regression. users don't want to see BoundMethod in the hover type

aaron-ang: https://github.com/facebook/pyrefly/blob/ffa1a38f576646ad9bac76e38084a98d210f2473/pyrefly/lib/test/lsp/hover_type.rs#L455-L489

@kinto0 an existing test expects `BoundMethod` on hover when the method does not match any of the overloaded function signatures. I may have misinterpreted when it should be shown. Can I get some clarification here?

pyrefly/lib/test/lsp/hover.rs
aaron-ang: After I rebased this branch onto main, the test failed the assertion:
 ```python
<(method) __matmul__: def __matmul__(
>(attribute) __matmul__: (
     self: Matrix,
     other: Matrix
<) -> Matrix: ...
>) -> Matrix
 ```

aaron-ang: If it's ok for the method to be named as an attribute for now, I will refactor the PR to reflect this.

aaron-ang: I also want to add that I was able to reproduce the issue on my machine while working on it initially. So the bug report is legit, and the issue may have been fixed elsewhere afterwards.

kinto0:
thanks for the split. I like your refactor too

PR #1457, opened 2025-11-03
Fixes signature help active parameter highlighting ,issue num 1417

### LINKED ISSUE #1417: input suggestion will jump to another parameter
### Describe the Bug

now is ok.

Image

But when I finish `dtype=`, it becomes

Image

The highlight jumps to the next one.

```py
import numpy as np

a = np.array([1], dtype=)
a @ a
```

### Sandbox Link

_No response_

v0.39.2

kshitijgetsac:
## Summary
Fixes signature help active parameter highlighting issue where the wrong parameter was being highlighted during function call completion.

## Problem
When typing inside a function parameter (e.g., `np.array([1], dtype=`), the signature help incorrectly highlighted the next parameter (`copy`) instead of the current one being typed (`dtype`).

## Root Cause
The bug was caused by setting the `active_parameter` field at both:
1. The `SignatureHelp` level (correct - indicates which parameter globally)
2. The individual `SignatureInformation` level (incorrect - should be None)

According to the LSP specification, `active_parameter` should only be set at the `SignatureHelp` level to indicate which parameter is currently active across all signature overloads.

## Changes Made
- Removed the `active_parameter` calculation from `create_signature_information` function in `lsp.rs`
- Removed unused `arg_index` parameter from `create_signature_information`
- Updated the call site in `get_signature_help_at` to not pass `arg_index`
- Updated test expectations in `signature_help.rs` to check `active_parameter` at the correct level

## Testing
- [x] Code compiles successfully
- [x] All existing tests pass (including 6 updated signature help tests)
- [x] Code follows the LSP specification for signature help

Fixes #1417

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -906,7 +906,7 @@ impl<'a> Transaction<'a> {
             |(callables, chosen_overload_index, arg_index)| SignatureHelp {
                 signatures: callables
                     .into_iter()
-                    .map(|t| Self::create_signature_information(t, arg_index))
+                    .map(|t| Self::create_signature_information(t)) // Remove arg_index parameter
                     .collect_vec(),
                 active_signature: Some(chosen_overload_index as u32),
                 active_parameter: Some(arg_index as u32),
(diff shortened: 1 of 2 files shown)


aaron-ang:
Hi @kshitijgetsac, I don't think your patch fixes the issue. I checked out your branch and built the LSP locally. It still fails with the example provided in the issue

Screenshot 2025-11-02 at 10 31 58 PM

kshitijgetsac:
@aaron-ang thanks for letting me know , appreciate the help.

kshitijgetsac:
The issue was fixed by @aaron-ang and my implementation did not work

PR #1458, opened 2025-11-03
fix Support types like pydantic.types.PositiveInt #1451

### LINKED ISSUE #1451: Support types like `pydantic.types.PositiveInt`
### Describe the Bug

https://docs.pydantic.dev/2.3/api/types/#pydantic.types.PositiveInt

There's no dependent types in Python, so I'm not sure how to avoid promoting it to `int` whenever we try to do something w/ it

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1451

Leveraged new RangeConstraints plumbing so type aliases and annotations keep the Gt/Ge/Lt/Le metadata 

Propagated those constraints into Pydantic field metadata and enforced them for class-body defaults, including support for the previously missing le keyword

Added le handling to Pydantic binding constants plus range merging helpers

### DIFF
--- a/pyrefly/lib/test/named_tuple.rs
+++ b/pyrefly/lib/test/named_tuple.rs
@@ -98,8 +98,8 @@ testcase!(
     r#"
 from typing import NamedTuple, Required, NotRequired, ReadOnly, ClassVar, Final
 class MyTuple(NamedTuple):
-    v: ClassVar[int]  # E: `ClassVar` may not be used for NamedTuple members
-    w: Final[int]  # E: `Final` may not be used for NamedTuple members
+    v: ClassVar[int]  # E: `ClassVar` may not be used for TypedDict or NamedTuple members
+    w: Final[int]  # E: `Final` may not be used for TypedDict or NamedTuple members
     x: NotRequired[int]  # E: `NotRequired` may only be used for TypedDict members
     y: Required[int]  # E: `Required` may only be used for TypedDict members
     z: ReadOnly[int]  # E: `ReadOnly` may only be used for TypedDict members


migeed-z:
Thank you for this! this would be a great change to support.  I will take a more detailed look soon. 
One thing that would be really helpful is to add comments on the helper functions on what they do and how they relate to the overall solution. 

Also, would that solution also check the "gt"/"lt" etc. field constraints?  if we're adding actual range constraints to the solver then we could also leverage this for the "gt" and "lt" testcases? right now, they do not check ranges. They just check against the annotation.

pyrefly/lib/test/pydantic/field.rs
```diff
@@ -76,6 +76,56 @@ A()  # E: Missing argument `x`
...
+);
+
+pydantic_testcase!(
+    test_field_default_gt_ok,
+    r#"
+from pydantic import BaseModel, Field
+
+class Model(BaseModel):
+    value: int = Field(1, gt=0)
+    "#,
```
migeed-z, line 158: I would have expected this testcase in the same file:

```
class Model(BaseModel):
   x: int = Field(gt=0, lt=10)

Model(x=5) 
Model(x=0)  
Model(x=15)
```

to now have a different result. Would that diff not support that?

crates/pyrefly_types/src/keywords.rs
```diff
@@ -74,6 +74,22 @@ impl KwCall {
     }
 }
 
+/// Numeric range constraints extracted from metadata such as `annotated_types.Gt`.
+#[derive(Debug, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
migeed-z: Could we also add a few top level; notes about how there are certain cases that this may not resolve in static time, and clarify which cases this will cover vs the ones that have to be deferred to runtime?

pyrefly/lib/test/pydantic/field.rs
asukaminato0721: Right now, the new logic only kicks in when we analyze class-body defaults.

If we want to catch those call-site violations too, we’d need a larger follow-up that threads the range constraints through the call analysis (so the solver can reject Model(x=0)

crates/pyrefly_types/src/annotation.rs
```diff
@@ -24,6 +25,7 @@ use crate::types::Type;
 pub struct Annotation {
     pub qualifiers: Vec<Qualifier>,
     pub ty: Option<Type>,
+    pub range_constraints: RangeConstraints,
```
rchen152: The `Annotation` struct is supposed to represent an "annotation expression" as defined in the typing specification: https://typing.python.org/en/latest/spec/annotations.html#type-and-annotation-expressions. We shouldn't add a `range_constraints` field to it. Is there a way to implement this feature without modifying `Annotation`?

rchen152:
Thanks for updating the PR! I'll review this as soon as I can.

rchen152:
Thanks again! I'll work on getting this merged internally. Just FYI (no action required on your part), I reverted what looked like some accidental changes caused by a merge conflict and moved `check_pydantic_range_default` from `class_field.rs` to `pydantic.rs`.

migeed-z (approved):

pyrefly/lib/test/pydantic/field.rs
asukaminato0721: fixed in #1902

Youngboy22tgomori (commented):
class Model(BaseModel):
   x: int = Field(gt=0, lt=10)

Model(x=5) 
Model(x=0)  
Model(x=15)

PR #1459, opened 2025-11-03
Track positional and keyword arguments

### LINKED ISSUE #1417: input suggestion will jump to another parameter

aaron-ang:
Close #1417.

We need to save the keyword identifier during signature range search to find the correct position when creating signature information.

I added some test cases to `signature_help.rs`, and manually verified the fix by building and running LSP locally:
Screenshot 2025-11-02 at 10 51 53 PM

There is also a bug in positional arguments, as the newly-added `positional_arguments_test` fails on the `main` branch. To fix this, we need to count the number of separators (commas) and adjust the index accordingly.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -477,8 +477,11 @@ pub struct FindDefinitionItem {
 /// The currently active argument in a function call for signature help.
 #[derive(Debug)]
 enum ActiveArgument {
+    /// The cursor is within an existing positional argument at the given index.
     Positional(usize),
+    /// The cursor is within a keyword argument whose name is provided.
     Keyword(Name),
+    /// The cursor is in the argument list but not inside any argument expression yet.
     Next(usize),
 }
 
@@ -857,36 +860,11 @@ impl<'a> Transaction<'a> {
         if let Expr::Call(call) = x
             && call.arguments.range.contains_inclusive(find)
         {
-            // Check positional arguments
-            for (i, arg) in call.arguments.args.as_ref().iter().enumerate() {
-                if arg.range().contains_inclusive(find) {
-                    Self::visit_finding_signature_range(arg, find, res);
-                    if res.is_some() {
-                        return;
-                    }
-                    *res = Some((
-                        call.func.range(),
-                        call.arguments.range,
-                        ActiveArgument::Positional(i),
-                    ));
-                    return;
-                }
+            if Self::visit_positional_signature_args(call, find, res) {
+                return;
             }
-            // Check keyword arguments
-            let kwarg_start_idx = call.arguments.args.len();
-            for (j, kw) in call.arguments.keywords.iter().enumerate() {
-                if kw.range.contains_inclusive(find) {
-                    Self::visit_finding_signature_range(&kw.value, find, res);
-                    if res.is_some() {
-                        return;
-                    }
-                    let active_argument = match kw.arg.as_ref() {
-                        Some(identifier) => ActiveArgument::Keyword(identifier.id.clone()),
-                        None => ActiveArgument::Positional(kwarg_start_idx + j),
-                    };
-                    *res = Some((call.func.range(), call.arguments.range, active_argument));
-                    return;
-                }
+            if Self::visit_keyword_signature_args(call, find, res) {
+                return;
             }
             if res.is_none() {
                 *res = Some((
@@ -900,6 +878,51 @@ impl<'a> Transaction<'a> {
         }
     }
 
+    fn visit_positional_signature_args(
+        call: &ExprCall,
+        find: TextSize,
+        res: &mut Option<(TextRange, TextRange, ActiveArgument)>,
+    ) -> bool {
+        for (i, arg) in call.arguments.args.as_ref().iter().enumerate() {
+            if arg.range().contains_inclusive(find) {
+                Self::visit_finding_signature_range(arg, find, res);
+                if res.is_some() {
+                    return true;
+                }
+                *res = Some((
+                    call.func.range(),
+                    call.arguments.range,
+                    ActiveArgument::Positional(i),
+                ));
+                return true;
+            }
+        }
+        false
+    }
+
+    fn visit_keyword_signature_args(
+        call: &ExprCall,
+        find: TextSize,
+        res: &mut Option<(TextRange, TextRange, ActiveArgument)>,
+    ) -> bool {
+        let kwarg_start_idx = call.arguments.args.len();
+        for (j, kw) in call.arguments.keywords.iter().enumerate() {
+            if kw.range.contains_inclusive(find) {
+                Self::visit_finding_signature_range(&kw.value, find, res);
+                if res.is_some() {
+                    return true;
+                }
+                let active_argument = match kw.arg.as_ref() {
+                    Some(identifier) => ActiveArgument::Keyword(identifier.id.clone()),
+                    None => ActiveArgument::Positional(kwarg_start_idx + j),
+                };
+                *res = Some((call.func.range(), call.arguments.range, active_argument));
+                return true;
+            }
+        }
+        false
+    }
+
     /// Finds the callable(s) (multiple if overloads exist) at position in document, returning them, chosen overload index, and arg index
     fn get_callables_from_call(
         &self,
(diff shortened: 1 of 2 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -86,6 +86,13 @@ use crate::types::callable::Params;
 use crate::types::module::ModuleType;
 use crate::types::types::Type;
 
+#[derive(Debug, Clone)]
```
stroxler: Could we add a doc comment to this? The most important thing is to just note that it's used for powering signature help this module is big enough to make that not necessarily obvious to someone skimming without context.

stroxler:
Thanks for the PR!

This looks pretty good to me but we should get a review from @kinto0 as well

pyrefly/lib/state/lsp.rs
aaron-ang: added and moved the enum def closer to where it is used.

aaron-ang:
rebased

pyrefly/lib/test/lsp/signature_help.rs
```diff
@@ -128,6 +128,99 @@ Signature Help Result: active=0
...
+    assert_eq!(
+        r#"
+# main.py
+4 | f(a)
+      ^
+Signature Help Result: active=0
+- def f(
+    a: str,
+    b: int
+) -> None, parameters=[a: str, b: int], active parameter = 0
```
stroxler:
```suggestion
) -> None: ..., parameters=[a: str, b: int], active parameter = 0
```

pyrefly/lib/test/lsp/signature_help.rs
```diff
@@ -128,6 +128,99 @@ Signature Help Result: active=0
...
+
+6 | f(a=)
+       ^
+Signature Help Result: active=0
+- def f(
+    a: str,
+    b: int
+) -> None, parameters=[a: str, b: int], active parameter = 0
```
stroxler:
```suggestion
) -> None: ..., parameters=[a: str, b: int], active parameter = 0
```

pyrefly/lib/test/lsp/signature_help.rs
```diff
@@ -128,6 +128,99 @@ Signature Help Result: active=0
...
+
+8 | f(b=)
+       ^
+Signature Help Result: active=0
+- def f(
+    a: str,
+    b: int
+) -> None, parameters=[a: str, b: int], active parameter = 1
```
stroxler:
```suggestion
) -> None: ..., parameters=[a: str, b: int], active parameter = 1
```

pyrefly/lib/test/lsp/signature_help.rs
```diff
@@ -128,6 +128,99 @@ Signature Help Result: active=0
...
+        r#"
+# main.py
+4 | f(1,,)
+        ^
+Signature Help Result: active=0
+- def f(
+    x: int,
+    y: int,
+    z: int
+) -> None, parameters=[x: int, y: int, z: int], active parameter = 1
```
stroxler:
```suggestion
) -> None: ..., parameters=[x: int, y: int, z: int], active parameter = 1
```

pyrefly/lib/test/lsp/signature_help.rs
```diff
@@ -128,6 +128,99 @@ Signature Help Result: active=0
...
+
+6 | f(1,,)
+         ^
+Signature Help Result: active=0
+- def f(
+    x: int,
+    y: int,
+    z: int
+) -> None, parameters=[x: int, y: int, z: int], active parameter = 2
```
stroxler:
```suggestion
) -> None: ..., parameters=[x: int, y: int, z: int], active parameter = 2
```

pyrefly/lib/test/lsp/signature_help.rs
```diff
@@ -128,6 +128,99 @@ Signature Help Result: active=0
...
+
+8 | f(1,,3)
+        ^
+Signature Help Result: active=0
+- def f(
+    x: int,
+    y: int,
+    z: int
+) -> None, parameters=[x: int, y: int, z: int], active parameter = 1
```
stroxler:
```suggestion
) -> None: ..., parameters=[x: int, y: int, z: int], active parameter = 1
```

stroxler (commented):
I think you rebased past a change in the hover result, hence the test failures

pyrefly/lib/state/lsp.rs
```diff
@@ -844,25 +852,47 @@ impl<'a> Transaction<'a> {
...
         x: &Expr,
         find: TextSize,
-        res: &mut Option<(TextRange, TextRange, usize)>,
+        res: &mut Option<(TextRange, TextRange, ActiveArgument)>,
     ) {
         if let Expr::Call(call) = x
             && call.arguments.range.contains_inclusive(find)
         {
+            // Check positional arguments
             for (i, arg) in call.arguments.args.as_ref().iter().enumerate() {
```
kinto0: nit: if this was extracted into a helper function, we wouldn't need the comments since the function name would explain it. it also makes visit_finding_signature_range a lot easier to read

pyrefly/lib/state/lsp.rs
```diff
@@ -844,25 +852,47 @@ impl<'a> Transaction<'a> {
...
+                    *res = Some((
+                        call.func.range(),
+                        call.arguments.range,
+                        ActiveArgument::Positional(i),
+                    ));
+                    return;
+                }
+            }
+            // Check keyword arguments
+            let kwarg_start_idx = call.arguments.args.len();
```
kinto0: same nit: another helper function? this function is getting big

pyrefly/lib/state/lsp.rs
```diff
@@ -474,6 +474,14 @@ pub struct FindDefinitionItem {
     pub module: Module,
 }
 
+/// The currently active argument in a function call for signature help.
+#[derive(Debug)]
+enum ActiveArgument {
+    Positional(usize),
+    Keyword(Name),
+    Next(usize),
```
kinto0, line 485: nit: could you add a doc-comment to these? it's unclear what Next is

pyrefly/lib/state/lsp.rs
```diff
@@ -942,6 +984,46 @@ impl<'a> Transaction<'a> {
...
+    }
+
+    fn parameter_name(param: &Param) -> Option<&Name> {
+        match param {
+            Param::PosOnly(Some(name), ..)
+            | Param::Pos(name, ..)
+            | Param::VarArg(Some(name), ..)
+            | Param::KwOnly(name, ..)
+            | Param::Kwargs(Some(name), ..) => Some(name),
+            _ => None,
```
kinto0: looks like this covers all cases so please exclude this wildcard (makes it less bug-prone when we add a new Param)

pyrefly/lib/state/lsp.rs
```diff
@@ -942,6 +984,46 @@ impl<'a> Transaction<'a> {
...
+            ActiveArgument::Positional(index) | ActiveArgument::Next(index) => {
+                (*index < params.len()).then_some(*index)
+            }
+            ActiveArgument::Keyword(name) => params.iter().position(|param| {
+                Self::parameter_name(param).is_some_and(|param_name| param_name == name)
+            }),
+        }
+    }
+
+    fn parameter_name(param: &Param) -> Option<&Name> {
```
kinto0: this function should probably go in pyrefly_types/src/callable.rs as an impl for Param. maybe fmt_with_type should use it?

pyrefly/lib/state/lsp.rs
```diff
@@ -875,11 +905,17 @@ impl<'a> Transaction<'a> {
...
         handle: &Handle,
         position: TextSize,
-    ) -> Option<(Vec<Type>, usize, usize)> {
+    ) -> Option<(Vec<Type>, usize, ActiveArgument)> {
         let mod_module = self.get_ast(handle)?;
         let mut res = None;
         mod_module.visit(&mut |x| Self::visit_finding_signature_range(x, position, &mut res));
-        let (callee_range, call_args_range, arg_index) = res?;
+        let (callee_range, call_args_range, mut active_argument) = res?;
+        if let ActiveArgument::Next(index) = &mut active_argument
```
kinto0, line 939: please add a comment here for what this part does and why it's necessary

kinto0 (changes requested):
thanks for the contribution! I'm a little unclear as to why we need Next and why we need to count the separators now. could you add more information to the PR?

kinto0 (approved):

stroxler (approved):

PR #1467, opened 2025-11-04
Fix broken wasm test

stroxler:
Summary:
I think a PR yesterday missed this, which is understandable although I'm a little
surprised internal CI didn't catch it; this should fix a failing github CI test.

Differential Revision: D86196577


ndmitchell (approved):

Issue #105, opened 2025-04-16
Function invocation pins type of arguments too early


grievejia:
Minimal repro:
```
def foo[T](a: T, b: T) -> T:
    ...

def test(x: int | None, y: int | None) -> int | None:
    if x is None:
        return foo(x, y)
    return None
```
Expected: No error
Actual: Argument `int | None` is not assignable to parameter `b` with type `None` in function `foo` [bad-argument-type]

lolpack:
Sandbox URL: https://pyrefly.org/sandbox/?code=CYUwZgBGD20NoBUC6AKAhgLgggNBARlggJQQC0AfNhgFAT0QB0zNNokALiAM4coAeWAJYA7DhAA+EAHLQRIPAE9hYyTLkhSlCKPFTZ82gx2R+O7usN1jDAE4gOAV1siosAUuLX69py8sgQA

martindemello:
i don't understand the initial example - `if x is None` coerces the type of `x` to `None` and then `foo(x, y)` could indeed fail typechecking if `y` is `int`, so pyrefly is doing the right thing there.

grievejia:
@martindemello If the typevar `T` for `foo` gets instantiated into `int | None` then type checking won't fail. i.e.
```python
def foo(a: int | None, b: int | None) -> int | None:
    ...

def test(x: int | None, y: int | None) -> int | None:
    if x is None:
        return foo(x, y)  # This should type check
    return None
```

We don't really want to assume the call `foo(x, y)` forces `T` to be `None` just because its first argument `x` is `None` -- the point here is that if we wait longer and get to see what the type of `y` is, we could have guessed better type for `T`.

samwgoldman:
If I understand, the following is a even simpler example where the eager pinning of generics can go awry:

```
def foo[T](a: T, b: T) -> T: ...
foo(0, 1.)   # Argument `float` is not assignable to parameter `b` with type `int` in function `foo`
```

samwgoldman:
Still working on this :)

asukaminato0721:
Now, no errors.

rchen152:
Huh, thanks for checking! I'm not sure why the original example doesn't error anymore (maybe it was fixed by one of the recent improvements that @stroxler made?), but it's easy to modify the example to error again:
```
def foo[T](a: T, b: T) -> T:
    ...

def test(x: int | None, y: int | None) -> int | None:
    if x is None:
[...]

PR #1468, opened 2025-11-04
fix preserve Literal through generic function calls #1138

### LINKED ISSUE #1138: preserve `Literal` through generic function calls
### Describe the Bug

```py
def f[T](t: T) -> T:
    return t

reveal_type(1)  # Literal[1]
a = 1
reveal_type(a)  # Literal[1]
reveal_type(f(a))  # int, expect Literal[1]
reveal_type(f(1))  # int, expect Literal[1]
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1138

Preserved literal instantiations for quantified type variables, so generic calls now keep Literal[...] precision. Let quantified variables keep their precise literal answer only until a later constraint rejects it, then automatically widen to the promoted class type so multi-argument inference succeeds without regressing literal-preserving cases

stores the literal answer first, only promoting to the base type when the literal fails its bound, and continues to record specialization errors if both literal and promoted forms violate the bound.

After TypedDict construction, promote inferred type arguments from literals to their general class types so C(x=0) now infers C[int] instead of C[Literal[0]]

### DIFF
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -2010,8 +2010,9 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                         match self.is_subset_eq(&t1, t2) {
                             Ok(()) => Ok(()),
                             Err(err) => {
-                                let t1_promoted =
-                                    t1.clone().promote_literals(self.type_order.stdlib());
+                                let t1_promoted = t1
+                                    .clone()
+                                    .promote_implicit_literals(self.type_order.stdlib());
                                 if t1_promoted != t1 {
                                     self.solver
                                         .variables
@@ -2156,8 +2157,9 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                         match self.is_subset_eq(t1, &t2) {
                             Ok(()) => Ok(()),
                             Err(err) => {
-                                let t2_promoted =
-                                    t2.clone().promote_literals(self.type_order.stdlib());
+                                let t2_promoted = t2
+                                    .clone()
+                                    .promote_implicit_literals(self.type_order.stdlib());
                                 if t2_promoted != t2 {
                                     self.solver
                                         .variables
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -2007,7 +2007,22 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                         let t1 = t1.clone();
                         drop(v1_ref);
                         drop(variables);
-                        self.is_subset_eq(&t1, t2)
+                        match self.is_subset_eq(&t1, t2) {
+                            Ok(()) => Ok(()),
+                            Err(err) => {
+                                let t1_promoted =
+                                    t1.clone().promote_literals(self.type_order.stdlib());
+                                if t1_promoted != t1 {
+                                    self.solver
+                                        .variables
+                                        .lock()
+                                        .update(*v1, Variable::Answer(t1_promoted.clone()));
+                                    self.is_subset_eq(&t1_promoted, t2)
+                                } else {
+                                    Err(err)
+                                }
+                            }
+                        }
                     }
                     Variable::Quantified {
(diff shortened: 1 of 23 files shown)


pyrefly/lib/solver/solver.rs
```diff
@@ -1238,24 +1253,38 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                         let t2 = t2.clone();
                         drop(v2_ref);
                         drop(variables);
-                        self.is_subset_eq(t1, &t2)
+                        match self.is_subset_eq(t1, &t2) {
```
stroxler, line 2157: (same thing here - I don't think we would ever want to change an answer after it has been decided already)

pyrefly/lib/solver/solver.rs
```diff
@@ -1196,7 +1196,22 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                         let t1 = t1.clone();
                         drop(v1_ref);
                         drop(variables);
-                        self.is_subset_eq(&t1, t2)
+                        match self.is_subset_eq(&t1, t2) {
```
stroxler, line 2010: This section looks unsafe to me, I don't see how it could be okay to change an `Answer` after it's already been decided.

Do you remember why this was needed? It may be a hint that propagating literals isn't going to work well at least at this time.

stroxler:
The core part of the change in `Type::Quantified` looks about right to me in the sense that it does what the issue says it wants done.

But that issue is controversial, I'm not convinced this is a good idea for two reasons:
- First, issue https://github.com/facebook/pyrefly/issues/1138 is marked as needs-discussion because we intentionally decided to promote in generic function calls, partly inspired by what Pyright and mypy do. I think any change needs to be made intentionally
- Second, it worries me that this implementation relied on changing `Answer`s after the fact. I'm fairly certain we cannot do that safely at all; if that was necessary to make this work, then I think it may be a clear indication that we should just continue promoting literals at least for now.

I believe the discussion of #1138 also suggested that we might also have to promote sometimes depending on variance to make this work well because, for example if `f[T](x: T) -> list[T]: return [x]` then we would really want `f(1)` to have type `list[int]` rather than `list[Literal[1]]` to avoid lots of noise do to invariance.

I'm curious what @samwgoldman thinks, I'd also be curious to hear more about why the lines rewriting `Variable::Answer` seemed necessary.

samwgoldman (commented):
So, I think the overall premise here is OK in theory.

When inferring an instantiation for a type variable in a generic call, we promote literals specifically to accommodate later writes. However, if the type variable appears only in positive positions, then we don't need to worry about any writes. In this case, preserving literals yields a more precise inferred type.

Whether that additional precision is _useful_ is still not necessary clear to me. Neither mypy nor pyright implement this, for example.

That said, I think the implementation is a bit premature, and we need to wait for at least the resolution to #105. To see why, consider this example:

```
def f(x: T, y: T) -> T: ...
f(0, 1)
```

When we see the call `f(0, 1)`, we can determine that `T` appears covariantly in the return type, meaning we can avoid literal promotion. However, because of our eager solver, this would cause us to solve `T=Literal[0]` when checking the first argument against parameter`x`, causing an error when we go to check the second argument against parameter `y`.

Once we have the lazier solver, we will have a better place to make a choice about instantiations. However, we still would need to correctly determine if the type variable appears in an input position, which doesn't seem to be done here. There might be some subtleties here, like:

```
def f(x: T, cb: Callable[[list[T], None]) -> None: ...
def g(xs: list[int]): ...
f(0, lambda xs: g(xs))
```

Even though `T` doesn't appear in the return type at all, we still want to promote here, so we need to look at where `T` appears carefully.

Lastly, we should be thoughtful about creating unions of literal values in general. Unions can lead to poor type checker performance and overly verbose inlay hints / hover cards. If the promoted type is just as good, the additional precision is not worthwhile.

So, my suggestion is that we table this idea until later, but still meet it with some skepticism. I think the original issue should still be marked "needs discussion."

kinto0:
closing old draft PRs. please reopen and mark "ready for review" if you want a review!

PR #1500, opened 2025-11-05
publish diagnostics + test

yangdanny97:
Summary: Previously, publishDiagnostics were coming with malformed URIs due to Url::from_file_path, which prepends file://... to all URLs, even those starting with vscode-notebook-cell://...

Differential Revision: D86271655


kinto0 (approved):

PR #1501, opened 2025-11-05
fix class magic method missing auto completion #1420

### LINKED ISSUE #1420: class magic method missing auto completion
### Describe the Bug

Image

```py
class A:
    def __eq
```

### Sandbox Link

_No response_

v0.39.2

asukaminato0721:
fix #1420

Magic-method completions now surface when naming class methods, so typing def __ suggests the usual dunder overrides.

Added a curated magic-method catalog to feed completions when the identifier context is a method definition

Hooked the helper into the generic identifier branch so only method names trigger the new suggestions

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -143,6 +143,121 @@ pub enum DisplayTypeErrors {
 
 const RESOLVE_EXPORT_INITIAL_GAS: Gas = Gas::new(100);
 const MIN_CHARACTERS_TYPED_AUTOIMPORT: usize = 3;
+// Magic method names offered during method definition completions.
+const MAGIC_METHOD_NAMES: &[&str] = &[
+    "__abs__",
+    "__add__",
+    "__aenter__",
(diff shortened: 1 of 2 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -143,6 +143,121 @@ pub enum DisplayTypeErrors {
 
 const RESOLVE_EXPORT_INITIAL_GAS: Gas = Gas::new(100);
 const MIN_CHARACTERS_TYPED_AUTOIMPORT: usize = 3;
+// Magic method names offered during method definition completions.
+const MAGIC_METHOD_NAMES: &[&str] = &[
```
stroxler, line 147: This might better go in `pyrefly_python/src/dunder.rs`, we're trying to put generally useful information about Python there so that it's easier to access as a library without necessarily depending on the more complex parts of Pyrefly

kinto0: fixed this on merge

PR #1502, opened 2025-11-05
fix Simple alias to typing.Annotated is not recognized #1488

### LINKED ISSUE #1488: Simple alias to `typing.Annotated` is not recognized
### Describe the Bug

Minimal repro:
```python
from typing import Annotated

A = Annotated

def test(x: A[int, 0]) -> None:
    pass
```
Expected: No type error
Actual: 
- Expected a type argument for `Annotated`
- `type[Annotated]` is not subscriptable

In practice people probably won't do this kind of trivial aliasing within the same file. What would happen instead is that they tend to do it in version-compatibility shims, e.g.
```python
# my_library.py
import sys
import typing

if sys.version_info < (3, 9):
  Annotated = SOME_CUSTOM_IMPLEMENTATION_OF_ANNOTATED
else:
  Annotated = typing.Annotated

# my_app.py
from my_library import Annotated
def test(x: Annotated[int, 0]) -> None:
    pass
```
The pattern is currently (as of 11/14/2025) recognized by both Pyright and Ty, but not Mypy.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BBdOlwNUDGJgA66GQLoBeQcNHjJMmZhhhG8BgAp8iQQG1WDADR0ADAF0AlHQC0APjoA5XOhiIZdf3TEqHAIFiAArgzQcCTkiFS0jMysHFw8fMqkGuhguLiKmSBhZJTaUKSEDLRQFADEdAAKpKVg5XRoWHj4dADGXpBs4ZTiEF6EMvUAyjAwdAAWDAzEcIgA9KslZRW8bKsw6KuYuD1wq305EIPDUV6rdLmUdKgAbqjQqNiwvf2XQyNedFwxBu6BiMjIDDmXiczxglDgo3QBSkIAAzIQAIwAJhRMhAAF8wqgelFYQAxaAwCgdHAEWIEoA

_No response_

asukaminato0721:
fix #1488

Exposed binding_idx_for_name so bindings logic can inspect the current flow’s key and style when resolving aliases

Updated special-export detection to follow NameAssign/phi chains with cycle guards, allowing assignment aliases to inherit their source special form

Stopped treating bare Annotated assignments as implicit type aliases so aliasing the special form no longer triggers invalid-annotation errors

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -8,7 +8,6 @@
 use std::fmt;
 use std::fmt::Debug;
 use std::fmt::Display;
-use std::iter;
 use std::sync::Arc;
 
 use dupe::Dupe;
@@ -700,7 +699,11 @@ impl<'a> BindingsBuilder<'a> {
             | FlowStyle::Uninitialized => {
                 self.special_export_from_binding_idx(idx, visited_names, visited_keys)
             }
-            _ => None,
+            FlowStyle::MergeableImport(_)
+            | FlowStyle::Import(..)
+            | FlowStyle::ImportAs(_)
+            | FlowStyle::FunctionDef(..)
+            | FlowStyle::LoopRecursion => None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


pyrefly/lib/binding/bindings.rs
```diff
@@ -668,6 +686,86 @@ impl<'a> BindingsBuilder<'a> {
...
+    ) -> Option<SpecialExport> {
+        let (idx, style) = self.scopes.binding_idx_for_name(name)?;
+        match style {
+            FlowStyle::Other
+            | FlowStyle::ClassField { .. }
+            | FlowStyle::PossiblyUninitialized
+            | FlowStyle::Uninitialized => {
+                self.special_export_from_binding_idx(idx, visited_names, visited_keys)
+            }
+            _ => None,
```
stroxler: Could we make the negative cases explicit here?

Using `_` is only generally advisable when we know that a specific case or two are all we want to special-case, and that's likely to remain true in refactors.

Here, though, we're splitting `FlowStyle` close to half-and-half, which means any future change to `FlowStyle` is *likely* to silently create a bug by being handled as `_`; instead, we want the compiler to force anyone making such a change to actually decide

Issue #45, opened 2025-03-27
Disallow assigning to Final/@final members & variables


yangdanny97:

yangdanny97:
Final attrs which are initialized on the class may not be assigned to. For ones that aren't initialized on the class, naively extending the existing check would cause errors to be thrown when they are assigned in a constructor.

One way to solve this is to allow them to be assigned in constructors, but not outside of constructors. This isn't 100% correct (allowing assigning smth more than once in the constructor + flagging attrs that aren't initialized in the constructor but are initialized by some method), but it might be a reasonable tradeoff since there isn't really a way to model whether something is initialized or not in the type system.

PR #1509, opened 2025-11-05
Emit error when writing to a final field in a dataclass

fangyi-zhou:
When a final field of a dataclass is not initialised, the field is not marked as `Readonly` but as `Readwrite`. In these cases, we lookup if the field being written to is a dataclass field, and emit an error if the field is final.

We also handle the case we modify a final field in dataclass of the class instead of an instance.

To do so, we need to pass on the class base in a helper function, alongside the instance class. Maybe we should use an Either type since the two should not be passed at the same time.


yangdanny97:
I think this should apply to regular classes as well. Iirc the fact that our initial implementation of `Final` only errored if the field was initialized on the class is a known gap.

fangyi-zhou:
> I think this should apply to regular classes as well. Iirc the fact that our initial implementation of `Final` only errored if the field was initialized on the class is a known gap.

For my understanding, is this related to the comment here https://github.com/facebook/pyrefly/issues/45#issuecomment-3051031380 wrt to Final fields in general cases?

fangyi-zhou:
If the initialisation of `Final` remains an issue, let's land this diff as is and expand the feature in a follow up?

yangdanny97:
When I wrote the earlier comment I was a bit unclear over why dataclasses were different from regular classes, but I think I understand now - since the dataclass constructor is generated, there's no issue where uninitialized Final attrs can't be assigned to in the constructor.

PR #1512, opened 2025-11-06
fix [feature] treat deprecated functions differently in the IDE #716

### LINKED ISSUE #716: [feature] treat deprecated functions differently in the IDE
### Describe the Bug

We could either not surface them at all in autocomplete, or move them to the bottom & cross out the text to discourage use.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEaJFAO4CGIKqaAzptrvgCYCmCIHAxlTA5sSAAU7c+AoQAoARDAAWSJsqjie-QW1kBKIp2CwODGAEZpOqAFoAfFAByYFBwBcRKB6gIqDBiQNGJgBMFtZ2js5unl4+fkSCQA

_No response_

asukaminato0721:
fix #716

Introduced a completion post-processing tweak so any item carrying CompletionItemTag::DEPRECATED now receives a heavier sort_text, which pushes deprecated suggestions beneath non-deprecated ones while still keeping existing prioritization rules for auto-imports and private symbols intact.

### DIFF
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -149,9 +149,9 @@ foo.
 11 | foo.
          ^
 Completion Results:
+- (Field) x: int
 - (Field) [DEPRECATED] also_not_ok: int
 - (Method) [DEPRECATED] not_ok: def not_ok(self: Foo) -> None: ...
-- (Field) x: int
 "#
         .trim(),
         report.trim(),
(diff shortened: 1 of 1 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -2397,6 +2397,13 @@ impl<'a> Transaction<'a> {
                 "0"
             }
             .to_owned();
+            if item
```
kinto0, line 2400: can you inline this into the branches of sort_text instead of mutating it?

kinto0: it's probably late for you actually. ill just import + do this rn

PR #1514, opened 2025-11-06
fix Feat: Expand **kwargs with field names on hover when **kwargs is typed with Unpack (i.e. **kwargs: Unpack[T]) #758

### LINKED ISSUE #758: Feat: Expand **kwargs with field names on hover when `**kwargs` is typed with `Unpack` (i.e. `**kwargs: Unpack[T]`)
Say a function has a parameter `**kwargs` and its typed with `Unpack`. On hover it would be nice if **kwargs was replaced with the actual argument names and their types.

For instance, in the screenshot below, hovering on `func` reveals two arguments, `arg1` and `**kwargs`. It would be nice if instead hovering on func showing `arg1`, `extra_arg1`, and `extra_arg2`. 

Image

Thanks.

asukaminato0721:
fix #758

Hover output now expands **kwargs typed with typing.Unpack[TypedDict] into their constituent keyword-only parameters so people hovering a call see the real argument names.

Implemented the expansion by letting HoverValue accept a pre-computed display string, formatting callables via transaction.ad_hoc_solve, and inserting typed-dict fields as synthetic Param::KwOnly entries before the original **kwargs.

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -208,19 +208,6 @@ fn collect_typed_dict_fields_for_hover<'a>(
                     Some(fields)
                 }
             }
-            Type::PartialTypedDict(typed_dict) => {
-                let fields = solver.type_order().typed_dict_kw_param_info(typed_dict);
-                if fields.is_empty() {
-                    None
-                } else {
-                    Some(
-                        fields
-                            .into_iter()
-                            .map(|(name, ty, _)| (name, ty, Required::Optional(None)))
-                            .collect(),
-                    )
-                }
-            }
             _ => None,
         },
         _ => None,
@@ -251,6 +238,8 @@ fn expand_callable_kwargs_for_hover<'a>(
     }
 }
 
+/// Walks the type and rewrites any callable signatures so they display expanded TypedDict-based
+/// `**kwargs` entries, ensuring hover text shows the actual keyword names users can pass.
 fn expand_type_for_hover<'a>(solver: &AnswersSolver<TransactionHandle<'a>>, ty: &mut Type) {
     ty.visit_mut(&mut |t| match t {
         Type::Callable(callable) => expand_callable_kwargs_for_hover(solver, callable.as_mut()),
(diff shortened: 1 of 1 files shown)


pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -177,6 +194,126 @@ impl HoverValue {
...
+) {
+    if let Params::List(param_list) = &mut callable.params {
+        let mut expanded = Vec::with_capacity(param_list.len());
+        let mut changed = false;
+        for param in param_list.items() {
+            if let Param::Kwargs(_, ty) = param {
+                if let Some(fields) = collect_typed_dict_fields_for_hover(solver, ty) {
+                    changed = true;
+                    for (field_name, field_type, required) in fields {
+                        expanded.push(Param::KwOnly(field_name, field_type, required));
```
yangdanny97, line 229: Since required parameters can't come after non-required ones (there isn't a type checking issue here, just looks weird when we display it), we'd probably want to put the required TypedDict fields ahead of any non-required parameters. After the list is expanded, we can partition expanded by required-ness to make the new list.

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -177,6 +194,126 @@ impl HoverValue {
...
+                        expand_callable_kwargs_for_hover(solver, &mut forall.body.signature)
+                    }
+                }
+            }
+        }
+        _ => {}
+    });
+}
+
+fn format_type_with_expanded_kwargs(
```
yangdanny97: we can probably inline this since it's only called once

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -177,6 +194,126 @@ impl HoverValue {
...
+            }
+            expanded.push(param.clone());
+        }
+        if changed {
+            *param_list = ParamList::new(expanded);
+        }
+    }
+}
+
+fn expand_type_for_hover<'a>(solver: &AnswersSolver<TransactionHandle<'a>>, ty: &mut Type) {
```
yangdanny97, line 243: this could benefit from a docstring explaining what we're doing

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -177,6 +194,126 @@ impl HoverValue {
...
+        Type::Unpack(inner) => match inner.as_ref() {
+            Type::TypedDict(typed_dict) => {
+                let fields = solver.type_order().typed_dict_kw_param_info(typed_dict);
+                if fields.is_empty() {
+                    None
+                } else {
+                    Some(fields)
+                }
+            }
+            Type::PartialTypedDict(typed_dict) => {
```
yangdanny97: I don't think partial can appear in a kwargs annotation, it's only produced by local inference

here's its docstring

```
     /// Represents a "partial" version of a TypedDict that can be merged into the TypedDict
    /// (e.g., via its `update` method).
    /// For a TypedDict type `C`, `Partial[C]` represents an object with any subset of read-write
    /// keys from `C`, where each present key has the same value type as in `C`.
```

PR #1515, opened 2025-11-06
fix auto complete dict key when possible #1189

### LINKED ISSUE #1189: auto complete dict key when possible
### Describe the Bug

two cases:

```py
d = {
    "1": 1,
    "2": 3,
}
d[""] <- here can give Literal["1"] or "2" as a candidate
```

```py
d = {}
d["1"] = 1
d["2"] = 5
d[""] <- same here
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1189

Introduced dict key completions: added facet chains and literal facet propagation.

### DIFF
--- a/pyrefly/lib/test/typed_dict.rs
+++ b/pyrefly/lib/test/typed_dict.rs
@@ -1174,7 +1174,7 @@ def valid_kwargs(**kwargs: Unpack[Coord]):
 testcase!(
     test_items_as_key,
     r#"
-from typing import assert_type, TypedDict, Literal
+from typing import assert_type, TypedDict
 class TD(TypedDict):
     items: int
 td: TD = {'items': 1}


pyrefly/lib/test/lsp/completion.rs
```diff
@@ -12,9 +12,29 @@ use pyrefly_build::handle::Handle;
...
 
 use crate::state::lsp::ImportFormat;
+use crate::state::require::Require;
 use crate::state::state::State;
+use crate::test::util::extract_cursors_for_test;
 use crate::test::util::get_batched_lsp_operations_report;
 use crate::test::util::get_batched_lsp_operations_report_allow_error;
+use crate::test::util::mk_multi_file_state;
+
+fn strip_ansi(input: &str) -> String {
```
kinto0, line 25: what happens if we don't do this? can you add a comment?

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -126,6 +146,81 @@ Completion Results:
...
+    );
+}
+
+#[test]
+fn dict_key_completion_from_nested_literal() {
+    let code = r#"
+def nested():
+    config = {"user": {"name": "Alice"}}
+    config["user"][""]
+#               ^
```
kinto0: are you sure this is the correct column? could you attach a video of this in the IDE?

asukaminato0721: https://github.com/user-attachments/assets/c60cfff7-12cf-456f-a6bb-f2012d1a8541

asukaminato0721: ref https://stackoverflow.com/a/59774240/13040423

kinto0: is there a way to get these completions without triggering it manually? pyright [has trigger characters for `"`](https://github.com/microsoft/pyright/blob/6f3b72c148dc9db557e133b84e0f9a962e31a9b1/packages/pyright-internal/src/languageServerBase.ts#L655), what if we did that?

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -149,6 +169,81 @@ Completion Results:
...
+"#;
+    let (handles, state) = mk_multi_file_state(&[("main", code)], Require::indexing(), true);
+    let handle = handles.get("main").unwrap();
+    let position = extract_cursors_for_test(code)[0];
+    let completions = state
+        .transaction()
+        .completion(handle, position, ImportFormat::Absolute);
+    let labels: Vec<_> = completions.iter().map(|item| item.label.as_str()).collect();
+    assert!(labels.contains(&"name"));
```
kinto0: does it seem weird that we're testing that the completions on `user` result in the `name` completion? is it that we get completions for every nested dict value on all of them and the language client is supposed to filter it

kinto0: should we add a test that verifies the following behavior?
```
    config = {"user": {"name": "Alice"}}
    config["user"][""]
#                     ^ is only user and is not name
    config["user"][""]
#                             ^ is only name and is not user
```

pyrefly/lib/test/typed_dict.rs
```diff
@@ -1149,11 +1149,11 @@ def valid_kwargs(**kwargs: Unpack[Coord]):
 testcase!(
     test_items_as_key,
     r#"
-from typing import assert_type, TypedDict
+from typing import assert_type, TypedDict, Literal
 class TD(TypedDict):
     items: int
 td: TD = {'items': 1}
-assert_type(td['items'], int)
+assert_type(td['items'], Literal[1])
```
kinto0: could this mean it's harder to add items to a typeddict (i.e. all additions must be literal[1] and not int?

kinto0 (approved):

pyrefly/lib/test/typed_dict.rs
asukaminato0721: https://github.com/facebook/pyrefly/pull/1515/commits/47485e9dc86d5c48aec2d17452f56685d8583c81

so `td['items'] = 2` will fail but `td['items'] = a int value` will pass now. A bit too strict.

stroxler: @asukaminato0721 could we factor just the facet narrows out into a separate PR?

There's a lot of lsp logic in this change, but the typing implications are - I think - entirely due to the new facet narrows and we could focus on those separately

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -149,6 +180,95 @@ Completion Results:
...
+    config = {"user": {"name": "Alice"}}
+    config["user"][""]
+#                  ^
+"#;
+    let (handles, state) = mk_multi_file_state(&[("main", code)], Require::indexing(), true);
+    let handle = handles.get("main").unwrap();
+    let position = extract_cursors_for_test(code)[0];
+    let txn = state.transaction();
+    let labels = dict_field_labels(&txn, handle, position);
+    assert_eq!(labels, vec!["name".to_string()]);
```
kinto0: nit
```suggestion
    assert_eq!(labels, vec!["name".to_owned()]);
```

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -149,6 +180,95 @@ Completion Results:
...
+    config = {"user": {"name": "Alice"}}
+    config["user"][""]
+#           ^
+"#;
+    let (handles, state) = mk_multi_file_state(&[("main", code)], Require::indexing(), true);
+    let handle = handles.get("main").unwrap();
+    let position = extract_cursors_for_test(code)[0];
+    let txn = state.transaction();
+    let labels = dict_field_labels(&txn, handle, position);
+    assert_eq!(labels, vec!["user".to_string()]);
```
kinto0: nit
```suggestion
    assert_eq!(labels, vec!["user".to_owned()]);
```

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -12,9 +12,32 @@ use pyrefly_build::handle::Handle;
...
+use crate::test::util::mk_multi_file_state;
+
+// Some assertions compare literal completion dumps. pretty_assertions decorates
+// diffs with ANSI escapes, so we strip them to keep the expected strings stable.
+fn strip_ansi(input: &str) -> String {
+    let mut result = String::with_capacity(input.len());
+    let mut chars = input.chars();
+    while let Some(ch) = chars.next() {
+        if ch == '\u{1b}' {
+            while let Some(next) = chars.next() {
```
kinto0: nit
```suggestion
            for next in chars.by_ref() {
```

pyrefly/lib/state/lsp.rs
```diff
@@ -2270,6 +2276,193 @@ impl<'a> Transaction<'a> {
...
+        if let Some(base_type) = self.get_type_trace(handle, base_expr.range()) {
+            if let Some(typed_keys) = self.collect_typed_dict_keys(handle, base_type) {
+                for (key, ty) in typed_keys {
+                    let entry = suggestions.entry(key).or_insert(None);
+                    if entry.is_none() {
+                        *entry = Some(ty);
+                    }
+                }
+            }
+        }
```
kinto0: nit: combine branches

```suggestion
        if let Some(base_type) = self.get_type_trace(handle, base_expr.range())
          && let Some(typed_keys) = self.collect_typed_dict_keys(handle, base_type) {
              for (key, ty) in typed_keys {
                  let entry = suggestions.entry(key).or_insert(None);
                  if entry.is_none() {
                      *entry = Some(ty);
                  }
              }
        }
```

kinto0 (changes requested):
sorry, let's split this PR into two:
- lsp changes
- typing changes

asukaminato0721:
image

pyrefly/lib/test/typed_dict.rs
```diff
@@ -1174,7 +1174,7 @@ def valid_kwargs(**kwargs: Unpack[Coord]):
 testcase!(
     test_items_as_key,
     r#"
-from typing import assert_type, TypedDict
+from typing import assert_type, TypedDict, Literal
```
asukaminato0721:
```suggestion
from typing import assert_type, TypedDict
```

PR #1520, opened 2025-11-07
fix [Hover][Docstring] Docstrings from .py on attributes when base type is from .pyi #1090

### LINKED ISSUE #1090: [Hover][Docstring] Docstrings from .py on attributes when base type is from .pyi
### Describe the Bug
pyrefly in the IDE generally prefers the .py for docstrings because typeshed and type stubs generally don't include docstrings. we have the ability to prefer .py files from direct imports from [this issue](https://github.com/facebook/pyrefly/issues/406), but attributes are harder since we have to "solve" the type with the checker in order to get attributes, and the solver currently prefers .pyi.

example:
```
import numpy as np
#          ^ hover here shows the docstring from .py


def foo(x: np.darray):
#               ^ hover here should work, showing docstring from .py
    method= returns_np_darray().all
   #   ^                         ^ hover here will not show docstring, only type
``` 

test here:
https://github.com/facebook/pyrefly/blob/d0bdcaf7603c9a20b562b47f19561a1b20ed3dab/pyrefly/lib/test/lsp/lsp_interaction/definition.rs#L155

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1090

FindDefinitionItemWithDocstring now tracks an optional docstring_module, attribute fallback searches the executable .py counterpart (loading the module if needed), and records docstrings even when the definition came from a .pyi. The class/assignment context detection was tightened so nested members are resolved reliably.

Hover responses now read docstrings from the recorded module, so the rendered markdown includes the implementation docstring while keeping go‑to-definition on the stub.


kinto0:
thanks for implementing this! I like the approach a lot. I'll be landing it soon with some modifications, namely moving it up into definition so it works on go-to definition + hover

kinto0:
landed as 44515077623e502fed2efa91490bc354f9739255

PR #1522, opened 2025-11-07
fix Can't call function on object that is also a method #821

### LINKED ISSUE #821: Can't call function on object that is also a method
### Describe the Bug

```python
from typing import *

P = ParamSpec("P")
RV = TypeVar("RV", covariant=True)

class CachedMethod(Protocol, Generic[P, RV]):
    @staticmethod
    def clear_cache(cache: Any) -> None: ...

    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> RV: ...

def cache_on_self(fn: Callable[Concatenate[Any, P], RV]) -> CachedMethod[P, RV]:
    ...

class Foo:
    @cache_on_self
    def pointwise_read_writes(self): ...

    def f(self):
        self.pointwise_read_writes.clear_cache(self) # E: ERROR 20:9-47: Object of class `MethodType` has no attribute `clear_cache`
```

This should work as `pointwise_read_writes` is actually a `CachedMethod`.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQApQC8UpAhiDRAMoICmAxgBQBEpXAlEQBKANUpQAKohbC63EVwA0UNmABudJDRQwK4kAFcWAomwA2NAM4WoAYRpsAFiwAmAWRYwHYZx1LgYYCqmSgDiLCgsIEhsANqkSiIAunwAXERQGVAAAhYwNDDREB5ezumZzizAyqYsdAD6bPZOHI2OLClQAIIocHxQALQAfFAAcmARHQB00ySZUBVVdQ00pqZLHBYspsBKBHRoFh2kk-sWuwQA1gDup0eT16d9Q1AiUzNEC8pNLHXjdZvbDjAFAdOyrGgAIxqMRs40aMHC+RYMW6cCUpESCWEyQGwzsbTcxW8cSxiTSc2mkxIZks1gAYmAwOTMllWk5fih-ltgGUMp8cKgYFckJs6iBas46lcogiLBtuakoJTZuVKlBgPLtqleXMoADgJMBdphaLxTRJdKkLLJmZaiBlm1NcABEA

_No response_

asukaminato0721:
fix #821

Bound-method attribute resolution now also consults the wrapped function type before giving up, so descriptors that attach helpers like clear_cache remain visible after binding; this is handled by expanding the Type::BoundMethod branch to recurse into func.as_type() alongside the existing types.MethodType lookup.

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -1333,17 +1333,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     self.lookup_attr_from_attribute_base1((**protocol_base).clone(), attr_name, acc)
                 }
             }
-            AttributeBase1::Quantified(quantified, class) => {
-                if let Some(attr) = self.get_bounded_quantified_class_attribute(
-                    quantified.clone(),
-                    class,
-                    attr_name,
-                ) {
-                    acc.found_class_attribute(attr, base);
-                } else {
-                    acc.not_found(NotFoundOn::ClassObject(class.class_object().dupe(), base));
-                }
-            }
             AttributeBase1::BoundMethod(bound_func) => {
                 let method_type_base =
                     AttributeBase1::ClassInstance(self.stdlib.method_type().clone());
@@ -2383,9 +2372,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 self.completions_class_type(q.class_type(self.stdlib), expected_attribute_name, res)
             }
             AttributeBase1::ClassObject(class) => {
-                self.completions_class(class.class_object(), expected_attribute_name, res)
-            }
-            AttributeBase1::Quantified(_, class) => {
                 self.completions_class(class.class_object(), expected_attribute_name, res)
             }
             AttributeBase1::BoundMethod(bound_func) => {


pyrefly/lib/alt/attr.rs
```diff
@@ -1234,6 +1239,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     acc.not_found(NotFoundOn::ClassObject(class.class_object().dupe(), base));
                 }
             }
+            AttributeBase1::BoundMethodFunction(bound_func) => {
+                let mut func_bases = Vec::new();
+                self.as_attribute_base1(bound_func.clone().as_type(), &mut func_bases);
+                for base1 in func_bases {
+                    let not_found_len = acc.not_found.len();
+                    self.lookup_attr_from_attribute_base1(base1, attr_name, acc);
+                    acc.not_found.truncate(not_found_len);
```
stroxler, line 1344: This is the smell I was referring to - the attribute bases are supposed to be a union, which is why we propagate all 'not found' instances, and to make this work the current PR has to patch around that behavior here.

What we really want here is more like "intersection" behavior, but we don't currently have intersections so we probably want to make a single `AttributeBase1` that encompases everything we need

PR #1523, opened 2025-11-07
fix Go-to-def fails on unreachable branches #1254

### LINKED ISSUE #1254: Go-to-def fails on unreachable branches
### Describe the Bug

If we incorrectly infer the type of something, our control flow modeling will incorrectly treat branches as unreachable, and go-to-def does not work inside those branches.

The root cause of the unreachability in the sandbox example is https://github.com/facebook/pyrefly/issues/1159, but I think we would want go-to-def to work even if the branch is not reachable.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRyAxlFRw4dACrwGiOXT11MMMHXHjWEBqYAUcGFDABKHen2vOxvLihWADE91ugbb2hKh0ALx0AHK46LIugXR2ts6JbsFgoRF0AIx6AMR0bLgAtAylhsYA7rwA1moAFjCCAYEQxhlZEGrouPwxcalp%2BmGRnWGFxWUVRnRgqNCNzTAgADQgZIJgUKSE5TRQFIUACqRbO3RoWHj4dAqxkGwArpSoDBCxhHKFAMowMHQGgwGMQ4IgAPTgzZGHaEXhscEwdDgzC4BRwcH3dCPF5vD7Iua8OioIQLZTYWB3B4QZ6vd6xOi4Yj09BwL7oMgMBqxEoiShwfHZGQgADMhByACZhXIQABfdaoBTvEQAMWgK0QICuOAIJHIsqAA

_No response_

asukaminato0721:
fix #1254

Added a suppression stack so binding code can execute unreachable branches without emitting type errors and exposed a helper to gate this behavior to filesystem/memory modules only, skipping `builtins/__builtins__`.

Updated the if binding logic to analyze statically false branches only when the new gate allows it, then discard their flow so control-flow facts stay unchanged while IDE features (like go-to-def) gain the needed Keys.

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -73,6 +73,7 @@ use crate::export::definitions::Definition;
 use crate::export::definitions::DefinitionStyle;
 use crate::export::definitions::Definitions;
 use crate::export::definitions::MutableCaptureKind;
+use crate::export::definitions::Reachability;
 use crate::export::exports::LookupExport;
 use crate::export::special::SpecialExport;
 use crate::module::module_info::ModuleInfo;
@@ -157,6 +158,8 @@ pub struct NameWriteInfo {
     /// If this name only has one assignment, we will skip the `Anywhere` as
     /// an optimization, and this field will be `None`.
     pub anywhere_range: Option<TextRange>,
+    /// Whether the definition is reachable for the current sys_info configuration.
+    pub reachability: Reachability,
 }
 
 #[derive(Clone, Debug)]
@@ -229,6 +232,7 @@ struct StaticInfo {
     /// The range of the textually last assignment to this name. Used to check
     /// whether a captured variable is reassigned after a nested function definition.
     last_range: TextRange,
+    reachability: Reachability,
 }
 
 #[derive(Clone, Debug)]
@@ -309,7 +313,7 @@ impl StaticStyle {
 
     fn of_definition(
         name: Hashed<&Name>,
-        definition: Definition,
+        definition: &Definition,
         scopes: Option<&Scopes>,
         lookup: &dyn LookupExport,
         current_module: ModuleName,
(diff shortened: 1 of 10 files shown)


pyrefly/lib/binding/scope.rs
```diff
@@ -1418,6 +1418,22 @@ impl Scopes {
         self.current_mut().stat.expr_lvalue(x);
     }
 
+    /// Synthesize a static definition entry for `name` in the current scope if it
+    /// is missing. This is used when we deliberately analyze unreachable code for
+    /// IDE metadata; those code paths may not have been included in the up-front
+    /// static scan, so we add a lightweight placeholder on demand.
+    pub fn add_synthetic_definition(&mut self, name: &Name, range: TextRange) {
```
grievejia, line 2884: This feels a bit complicated to me. I wonder if a better alternative would be to just keep all defs in the very beginning when they are calculated, recording their unreachable status so downstream type checking logic would avoid emitting type errors for those instead.

rchen152:
I took a more careful look at this PR, and unfortunately, I don't think it's mergeable for several reasons.
* Architecturally, it adds parallel reachability plumbing with bindings-only error suppression, so errors from the answers phase are still emitted.
* It appears to enable diagnostics in unreachable code, not just go-to-def.
* It also binds and solves inactive branches, which breaks things like `sys.version_info` checks.
* As a result, mypy primer shows net +1,624 errors total, +818 errors in the default preset.
* It also seems to be incomplete; hover/go-to-def still don't work in examples like this:
```
if True:
    pass
else:
    x = 1
    print(x)  # Hover/go-to-definition still returns nothing
```

Issue #1498, opened 2025-11-05
Import root is wrongly inferred, no way to override?


nascheme:
### Describe the Bug

My project is old and follows a non-standard directory layout.  Pyrefly can't seem to handle this and ignores all of the files under the source folder.  At least, that is how it seems.   Running `pyrefly dump-config`, it says:
```
Import root (inferred from project layout): "/home/nas/src/foo"
```
Where "foo" is the top-level project folder.  The actual source files are stored under "\~/src/foo/lib".  Under that folder, there are Python packages, e.g. "\~/src/foo/lib/pkg1", "\~/src/foo/lib/pkg2".   

In modern Python layouts, the folder container the source code for the packages would be called "src" rather than "lib".  Indeed, moving "lib" to "src" and then created a symlink "lib -> src" seems to fix pyrefly.

Looking at the pyrefly source, I think it would be sufficient to make the "src" string inside `get_import_root()` be configurable.  I tried reproducing this bug with a smaller, example repo.  However, when I do that, the bug does not appear.

Another possible clue, the Hatch configuration has:
```
[tool.hatch.build.targets.wheel]
packages = [
    "lib/pkg1",
    "lib/pkg2",
]
```

This results in a pth file `./.venv/lib/python3.14/site-packages/_foo.pth` containing:
```
/home/nas/src/foo/lib
/home/nas/src/foo/lib
```
Removing the pkg2 package entirely doesn't fix things so it's not the duplicate path causing the issue.  Commenting out both lines in the `.pth` file does fix the issue.  My guess is that pyrefly is treating the files under lib as part of the Python install, since they appear in the .pth file.  However, if the files are stored under "src" rather than "lib", that problem is avoided due to the "import root" inference.

I'm normally using neovim (v0.12.0-dev-264+gedc8e9f40a) but the same issue appears with VSCode when using pyrefly as the LSP.  So, I don't think it's an IDE issue.

nascheme:
Building the 0.40.0 source with this change makes it work for my project layout:
```
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -111,7 +111,7 @@ impl ProjectLayout {
             );
             Self::default()
         };
-        let src_subdir = project_root.join("src");
+        let src_subdir = project_root.join("lib");
         match src_subdir.try_exists() {
             Ok(true) => return Self::Src,
             Ok(false) => (),
[...]

PR #1533, opened 2025-11-10
Filter project excludes for site package path entries if they would exclude search path

connernilsen:
Summary:
Pyrefly has some logic to add all `site-package-path` entries to `project-excludes`, but sometimes the Python interpreter includes the root of the project. Historically, we've filtered this out by using the project root, but if the actual project root differs from what Pyrefly thinks it is, we might end up excluding the project by default.

This happens in legacy projects, where instead of using `src` as the project directory, `lib` is used.

This diff instead uses all `search-path` entries, not just Pyrefly's heuristic project root, to determine if a site package path should be added. This way, anything that might be considered "project code" because it's in `search-path` will always be able to be included in `project-includes`.

Fixes https://github.com/facebook/pyrefly/issues/1498

Reviewed By: kinto0

Differential Revision: D86682740

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -570,7 +570,7 @@ impl ConfigFile {
         excludes.append(
             &self
                 .site_package_path()
-                .filter(|p| self.import_root.as_ref().is_none_or(|r| !r.starts_with(p)))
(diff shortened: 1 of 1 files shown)


kinto0:
looks like D86682740 was merged

PR #1542, opened 2025-11-11
fix documentation

kinto0:
Summary: it's just python.analysis, not pyright

Differential Revision: D86734468


ndmitchell (approved):

PR #1546, opened 2025-11-11
fix Enum should be subscriptable #916

### LINKED ISSUE #916: Enum should be subscriptable
### Describe the Bug

```python
from typing import *
from enum import *

def enum_object_hook(x: type[Enum]):
    x["test"] # E: `type[Enum]` is not subscriptable
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUKJFAKYoCu0WOehRRAJucBdRAPpgBGAVuQDGMLgAswYANYAKAB4AuWInIBtAKKcAugEoFRKIahzVAIhjkAzjFNaiQA&version=3.12

_No response_

asukaminato0721:
fix #916

class objects typed as type[Enum] (or any enum subclass) support string subscripting just like concrete Enum

Added an is_enum_class_type guard that returns the underlying enum instance type when indexing a Type[...] value with a str, and emits the existing “Enum type … can only be indexed by strings” error for non-string keys to keep diagnostics consistent.

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -2031,6 +2031,26 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         )
                     }
                 }
+                Type::Type(inner) if self.is_enum_class_type(inner.as_ref()) => {
+                    let base_display_ty = Type::Type(inner.clone());
+                    let enum_value_ty = *inner;
+                    if self.is_subset_eq(
+                        &self.expr(slice, None, errors),
+                        &self.stdlib.str().clone().to_type(),
+                    ) {
+                        enum_value_ty
+                    } else {
+                        self.error(
+                            errors,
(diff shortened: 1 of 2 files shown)


kinto0:
can you comment on the original issue so we can assign you?

yangdanny97:
@rchen152 this PR seems to fix the issue but the fix appears different from what you described here? https://github.com/facebook/pyrefly/issues/916#issuecomment-3185979287

rchen152:
> @rchen152 this PR seems to fix the issue but the fix appears different from what you described here? [#916 (comment)](https://github.com/facebook/pyrefly/issues/916#issuecomment-3185979287)

I haven't looked super carefully at this PR, but it looks like it's probably intercepting and special-casing enums rather than trying to go through the corresponding magic method call. We may still have to do the latter at some point, but special-casing enums seems fine to me; this is a pretty visible issue.

PR #1550, opened 2025-11-12
Identify methods in hover trace

### LINKED ISSUE #1544: Identify methods in hover trace
### Describe the Bug

Continuing from #1456, overloaded methods are currently shown as attributes on hover. This is because we only save `Callable` when recording an overload trace, which doesn't provide type information about the method itself. When aggregating the traces, we wrap each callable with `Type::Callable`, which I think is too generic.

We can and should provide richer context when saving an overload trace. One potential issue is the memory consumption used by the `Type` struct, which we can mitigate with `Arc`. Furthermore,  @kinto0 stated that there wasn't a noticeable performance hit in the large IG codebase with the initial changes made in #1456.

### Sandbox Link

_No response_

_No response_

aaron-ang:
Close #1544

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -296,8 +296,12 @@ pub fn get_hover(
         .into_iter()
         .next()
     {
+        let mut kind = metadata.symbol_kind();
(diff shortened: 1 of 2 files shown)


ndmitchell (approved):

PR #1554, opened 2025-11-12
fix False positive "Instance-only attribute is not visible on the class" #1103

### LINKED ISSUE #522: [unclear message] ClassVar initialized on __new__ method
### Describe the Bug

When a ClassVar is initialized on the \_\_new__ method, mypy indicate a `missing-attribute` (Instance-only attribute `k` of class `Foo` is not visible on the class)
I think an error like "The class attribute is initialized in the __new__ method and potentially doesn't exist" can be more explicit

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMogCmAboQIYA2A+vAoQDRQDCFZAzmwGpkgBQvAY1YcoAMTBgAXLyizMk5sK48A2qhgBdKAF4oABhlyAVgvU79h2ZagABIezYRCMABZgAJlADEUAHJh8IgBHAFckIndrd0JgKCoqFEIAd3iACiE2RgAqHjRMqCysgGsk3LYASmk5aswUNhgyFAFCczYQuhBU8oA6eMSUqnSKCutqjO6kKABqXQBGUblxo2m5hdlxovMAbwAiMp2FMuzi0pA8gF81gmcQkBRa+sbm-ljdcTBUgFZGACNtT-K-CIpEoNEQhFS7wm5W8igc3BAUGAYBCKEiCBA6khEmhvGB5GotAhUKMMJ8LHhPCRKLRvAxWJJgPxoKJ2LA3SKZKgAEk6g0moQALRgFAUOBQMgwGCYn4hGCEOmYlAwNkcrkAITltWulCQ8EYAkaUB+LTIAmaHEIniSepcsBcLXsHCAA

_No response_

### LINKED ISSUE #1103: False positive "Instance-only attribute is not visible on the class"
### Describe the Bug

I would expect the following to type check just fine in Python 3.12:

```python
from unittest import TestCase

class BaseTestCase(TestCase):
    _my_attr: int

class TestA(BaseTestCase):
    @classmethod
    def setUpClass(cls) -> None:
        cls._my_attr = 10
```

However, I get a `no-access` error: "Instance-only attribute `_my_attr` of class `TestA` is not visible on the class"

The same code type checks just fine with mypy in strict mode.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEAruhAC4vwt0Q3G6WcAVDgGFUcGAB10UgMZQxcOgCExMIXBajxACnWbVASkRS6pugH0apc6jaVEXdCymz5cRXoCC2leL1aYIxMzAAE5BRoYFgALXExg00wYMDpxFgBVYmE3OG05OAM6AFoAPjoAOVx0GGN0M3q6fMJLa1sWSjoAXjoARgAGKRAAGhAGFmg4EnJEEABiOnTxqFZSOjAmGXGquBcsZLW%2BGltzdAYabBhKbXwHCCdC0tT22vrKKIZKOrAJEHKzi-sdGA%2BAAvj9BiCRqhNhAAG4wABi0BgFDQWDwRDIIBBQA&version=3.12

_No response_

asukaminato0721:
fixes #1103
fixes #522

SpecialExport and decorator handling now recognize both `@classmethod` and @abc.abstractclassmethod, producing a new Decorators::is_classmethod flag so that we can differentiate class receivers when binding methods

Added MethodSelfKind plus an explicit MethodAttributeTarget on every recorded method assignment, so bindings remember whether an attribute came from self or cls

Carried that target information through ClassFieldDefinition and the raw/class-field initialization pipeline, treating class-target writes as class-body initializers (while still emitting existing diagnostics) so they no longer trigger NoAccess when accessed on the class

### DIFF
--- a/pyrefly/lib/binding/function.rs
+++ b/pyrefly/lib/binding/function.rs
@@ -538,7 +538,9 @@ impl<'a> BindingsBuilder<'a> {
             && !decorators.is_abstract_method
             && !is_ellipse(&body);
         let method_self_kind = if class_key.is_some()
-            && (decorators.is_classmethod || func_name.id == dunder::INIT_SUBCLASS)
+            && (decorators.is_classmethod
+                || func_name.id == dunder::INIT_SUBCLASS
+                || func_name.id == dunder::NEW)
         {
             MethodSelfKind::Class
         } else {
--- a/pyrefly/lib/test/attributes.rs
+++ b/pyrefly/lib/test/attributes.rs
@@ -187,17 +187,15 @@ class B(A):
 // Ref https://github.com/facebook/pyrefly/issues/370
 // Ref https://github.com/facebook/pyrefly/issues/522
 testcase!(
-    bug =
-        "Attributes initialized in `__new__` and `__init_subclass__` should not be instance-only.",
     test_cls_attribute_in_constructor,
     r#"
 from typing import ClassVar
 class A:
     def __new__(cls, x: int):
-        cls.x = x # E: Instance-only attribute `x` of class `A` is not visible on the class
+        cls.x = x
 class B:
     def __init_subclass__(cls, x: int):
-        cls.x = x # E: Instance-only attribute `x` of class `B` is not visible on the class
+        cls.x = x
 class C:
     x: ClassVar[int]
     def __new__(cls, x: int):


asukaminato0721:
seems also fix #522

yangdanny97 (approved):
Thanks, this generally looks good

My only comment would be that MethodSelfKind does the same thing as MethodAttributeTarget, so we could probably get rid of one of those two. I can patch that internally

PR #1562, opened 2025-11-12
don't look up all enum members when accessing `.value` on an enum literal

### LINKED ISSUE #1484: Pyrefly is marking `Literal[...]` with callables as non-callable at callsites
Clone [MkDocs](https://github.com/mkdocs/mkdocs)

Example 1: 
* in `mkdocs/structure/pages.py` around line 475, the use of `is_excluded()` in `path_to_url()`
  * also in `mkdocs/structure/files.py` around line 115, the use of `InclusionLevel.is_excluded` in `copy_static_files()`
* See that the `is_excluded()` function is marked as a `Literal[InclusionLevel.is_excluded]`, which is a function

I've also seen it marked as `Literal[Callable[[InclusionLevel], bool]]` as well

yangdanny97:
Summary:
fixes https://github.com/facebook/pyrefly/issues/1484

the original issue is caused by a cycle

Differential Revision: D86792749

### DIFF
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -1211,7 +1211,7 @@ impl Type {
         }
     }
 
-    fn is_toplevel_callable(&self) -> bool {
+    pub fn is_toplevel_callable(&self) -> bool {
         let mut is_callable = false;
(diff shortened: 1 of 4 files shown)


grievejia (approved):

PR #1564, opened 2025-11-13
Fix(solver): Allow Generator[Any] to match Iterable[T]

### LINKED ISSUE #1563: Generator failing to match Iterable (maybe due to unsolved Var?)
### Describe the Bug

Example distilled from castfit (https://github.com/facebook/pyrefly/issues/225):
```
from typing import Any

type TypeForm[T] = type[T] | Any

def _to_list[T](
    value: Any,
    kind: type[list[T]] = list,
) -> list[T]:
    return kind(to_type(val, Any) for val in value) # E: no matching overload

def to_type[T](value: Any, kind: TypeForm[T]) -> T: ...
```
Both mypy and pyright are okay with this, but pyrefly complains that `Generator[Any, None, None]` doesn't match any overloads of `list.__init__`, even though it should match the overload that takes an `Iterable`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BBdKQA66MU2Iw6AFWYwAYrxoBtGQF06AXkby1mgD6DhYsZhhg6AfQa4rUCHAb6AFGLoe6AN1RQArjCIxqQANO6eANasmEGSMCoOTvqaOokMYegAlHQAtAB8dGn6iOEelDAMfpTodFFYLrY28i4%2BUCHB2WC83r6cNa0BmaZYFox2ca4DgcHtdTGy8kqUqhrZ%2BbJBhFsgISB%2BDNBwJOSIIADEdACqBw5MdGB%2B6ADGB7jocMPmll3LqAxW6D8NGwMEoLnwQVYDDWBSclBKNU85Uq1XuIhAADkgSD4XRgPgAL7osQ7EBkcpgKCkQi2GhQCgXAAKpApVLoaCweHwdCeb0gbCqfwgb0IYguAGUYNIABYMBjEOCIAD0SvJFiphF4bCVMHQSswuCecCVvPQ-MFrz1926qB80FQ2FgPL5EAFlCFbzouGIlqOYjIDGlbxyXlBcGFNR06IAzIQAIwAJmJ6BABN2qBeEFDCmgMAoHJwBGOqaAA

_No response_

Slaytt:
Closes #1563

## What this PR does

This PR fixes a bug where the type checker would incorrectly report an error when a `Generator` (specifically one containing `Any`) was used where an `Iterable[T]` was expected. This would cause false positives on common patterns like `list(gen_expr)`.

## How it was fixed

The root cause was in `solver/subset.rs`, in the `is_subset_eq_impl` function. The logic for checking if a type `t1` is a subset of a `Type::Quantified(q)` (a TypeVar) was incomplete:

1.  It didn't correctly handle the case where `t1` is `Any`.
2.  It was missing match arms for `Restriction::Unrestricted` and `Restriction::Bound`, causing it to default to an error.

This patch adds the necessary logic to:
* Immediately return `Ok` if `t1` is `Any`.
* Correctly handle `Unrestricted` and `Bound` TypeVars.

### DIFF
--- a/pyrefly/lib/solver/subset.rs
+++ b/pyrefly/lib/solver/subset.rs
@@ -932,12 +932,21 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                     &want,
                 )
             }
-            (t1, Type::Quantified(q)) => match q.restriction() {
-                // This only works for constraints and not bounds, because a TypeVar must resolve to exactly one of its constraints.
-                Restriction::Constraints(constraints) => all(constraints.iter(), |constraint| {
-                    self.is_subset_eq(t1, constraint)
-                }),
-                _ => Err(SubsetError::Other),
+            (t1, Type::Quantified(q)) => {
+                // Allow 'Any' to match any TypeVar
+                if let Type::ClassType(t1_class) = t1 {
+                    if self.type_order.extends_any(t1_class.class_object()) {
+                        return Ok(());
+                    }
+                }
+                match q.restriction() {
+                    Restriction::Constraints(constraints) => all(constraints.iter(), |constraint| {
+                        self.is_subset_eq(t1, constraint)
+                    }),
+                    // Handle 'Unrestricted' and 'Bound' TypeVars
+                    Restriction::Unrestricted => Ok(()),
+                    Restriction::Bound(bound) => self.is_subset_eq(t1, bound),
+                }
             },
             (Type::Union(ls), u) => all(ls.iter(), |l| self.is_subset_eq(l, u)),
             (l, Type::Intersect(us)) => all(us.iter(), |u| self.is_subset_eq(l, u)),


pyrefly/lib/solver/subset.rs
```diff
@@ -932,12 +932,21 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
...
+                if let Type::ClassType(t1_class) = t1 {
+                    if self.type_order.extends_any(t1_class.class_object()) {
+                        return Ok(());
+                    }
+                }
+                match q.restriction() {
+                    Restriction::Constraints(constraints) => all(constraints.iter(), |constraint| {
+                        self.is_subset_eq(t1, constraint)
+                    }),
+                    // Handle 'Unrestricted' and 'Bound' TypeVars
```
yangdanny97, line 946: This is wrong, the comment that was deleted explains why

The added logic would allow programs like this to type check (it currently gives an [error](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGLrgNoAqAugBT6I3MCUNAWgB8PRAB10NGvhoBeGgEYpAYhoAXABYQ4NAO64ArlEw1sMGqihRcumCeyl1G8wCV4agE4QAxmoi50REQAVXQPdy9fO1MPDG8NCSlwtQMPSXxEmglSbjhPOToGdjEQEv5VGHxiGF8nbXVcU3M8jwAaUwM1Ggg1AHIdVF8DSyhHCHQ1EFaQMnCwUcI1XABbKApVAAVSOdGaNCw8GW8AyABzVNQ-AMIJVQBlGHMNNTViOEQAeg-Z6gXcD1OHxg6A%2BmFw3jgH2O6DOFyuIMKHgsADdUNBUNhYDRobDYvCaLhiPC4Dd0GRNAEBMiYB44P5JPISgBmQgKABMJQkIAAvtNBn5qQAxaAwCj7HAEEjkblAA))

```
def foo[T](x: T) -> T:
  x = 1  # this would be allowed by the Restriction::Unrestricted branch
  return x
  
y: str = foo("") # expect this to be str, but it's actually int
```

The reason it's safe to do for constrained type vars is because they must resolve to _exactly_ one of the constraints

so if you have something constrained by types A and B, it can be matched by a value that is both type A _and_ type B (for example, if A and B are classes we could have a class C that extends both A and B).

whereas when you have a bounded or unrestricted type var, it could be any type, so you can't be sure that a concrete type matches whatever it can resolve to.

yangdanny97 (changes requested):
Hi, thanks for the PR.

What makes you think that the root cause is that `Any` isn't handled? If the LHS is `Any` I would expect it to hit the branch on line 884.

Also, this PR is missing tests that demonstrate the fix works. Please refer to the examples in `pyrefly/lib/tests`

Slaytt:
Hi, I had the exact same thought so I checked the print logs from check_targs.

The logs show that the got_arg (the LHS) isn't actually the Type::Any variant. It's Type::ClassType(Class { ... name: "Any" ... }).

Because it's a Type::ClassType, it correctly bypasses the (Type::Any(_), _) arm on line 884.
It then falls all the way down to the (t1, Type::Quantified(q)) arm (line 935), where t1 is Type::ClassType(Any) and q is the TypeVar 'T'. This is where it was failing i think. 

But maybe we should add the specific check for this Type::ClassType(Any) inside the (t1, Type::Quantified(q)) ?

yangdanny97:
Interesting, I'll try to see what's going on there. There shouldn't be a class representation of `Any` since it's a special form.

yangdanny97:
TIL Any is a class in typeshed.

`canonicalize_all_class_types` is supposed to get rid of that when `Any` is used in a type position, but I think the issue is that it's being used in a non-type position in that example.

Making is_subset_eq handle this won't be enough, since other places in the codebase rely on Type::Any being the only representation of `Any`. We may need to add some code in `expr_infer` (and related functions) to intercept that Type::ClassType(Any) & replace it.

samwgoldman:
I looked into this a bit, and I think mypy and pyright pass this example for a different reason. If I remove the `| Any` from the definition of `TypeForm` in the original example, both mypy and pyright reject the code. I suspect that neither type checkers are actually checking against `T` here.

yangdanny97:
@samwgoldman so would you say that we're behaving correctly here then?

rchen152:
I'm going to close this, since https://github.com/facebook/pyrefly/issues/1563 has been fixed.

PR #1566, opened 2025-11-13
Disable project excludes heuristics when explicit config is given

connernilsen:
Summary: We should be able to disable project excludes when passing in `-c` with files

Differential Revision: D86925739


kinto0:
looks like it was merged

PR #1567, opened 2025-11-13
fix Auto-import completion missing label_detail #1553

### LINKED ISSUE #1553: Auto-import completion missing label_detail
### Describe the Bug

When getting auto-import suggestions in the completion list, the "auto-import" suggestion does not come with a label indicating that it will add an import unless I actually select the entry.

I have no knowledge about LSP protocols but it seems there's a `label_detail` attribute that is used to provide this information? Mentioning this in case it helps

With pyright (or Ty now - just tested it), it would include a label which I can see directly in the completion list (Ty screenshot attached).

Image

### Sandbox Link

_No response_

Neovim configured with lspconfig (defaults)

asukaminato0721:
fix #1553

Wired in CompletionItemLabelDetails with a reusable AUTO_IMPORT_LABEL_DETAIL constant so the LSP can consistently describe auto-import entries.

Auto-import symbol suggestions now publish label_details that show “auto-import” plus the module they’ll import from, so editors can display that context before the user accepts the completion.

image

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/completion.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/completion.rs
@@ -686,6 +686,10 @@ fn test_stdlib_class_completion() {
             "detail": "from email.errors import FirstHeaderLineIsContinuationDefect\n",
             "kind": 7,
             "sortText": "4",
+            "labelDetails": {
+                "detail": "auto-import",
+                "description": "email.errors",
+            },
             "additionalTextEdits": [{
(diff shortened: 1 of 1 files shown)


yangdanny97:
LGTM, i patched a few things after importing
1. adding a space before auto-import to make the formatting a bit nicer
2. adding auto-import text suffix to the label itself, if the client capabilities do not support label details

kinto0 (changes requested):

PR #1569, opened 2025-11-13
fix Go-to definition only finds definition from one element in a union #1470

### LINKED ISSUE #1470: Go-to definition only finds one class in a union
### Describe the Bug

given:
```
class State:
    def get_location(self) -> str: ...


class NewYork:
    def get_location(self) -> str:
        return "10016"

class Massachusetts:
    def get_location(self) -> str:
        return "10016"


def find_location(x: Massachusetts | NewYork):
    x.get_location()

def find_location2(x: NewYork | Massachusetts):
    x.get_location()
```
go-to definition on both get_location() calls only return massachusett's implementation

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAOugMZSpxwAEAygC6psyLVN9OYYYJgHMYbAPpRcNThFzoAFHBhQwASiYBaAHxM4bAE6ImhM9Qu16jJgDkYAdwCauQwGse6fgKGjxUmTkFZVUNbT0DY15vfkNxAFdDL0oQAEYABnTUgDYUyzoGZgBZQtQaAAt4lTY2OE9vQWExSWlZNnklFTVNXX0jepi%2BOLZE5LTMnLz0S0amSCwAto7FfBMSxjLK6tqmAB87Rxd3dQG%2BImbFoKV1Gd95zEv2hQAmFZN7Z1c3PaZ1tAqquJaidomdCBdWldFOoQAAaEDxdpQOAkciIEAAYiYAFUkRA2KQ5vFaE90HBbsIwK4ALacCToeLU7AwQxvJgQdBsHoRfqgpjDUZzFK2RnM4xMYD4AC%2BUzhIDIcTAUFIhDYuGpUAoWIACqRFcr9BgcAQmDQFJARIkroRqFiWDAYExyjViHUAPRuhVCZWEVwiN0wdBuzAyOBus3oC1W0luuauJioABuqGgqGwsFN5oglsMVyYuGIpJR1DIbHKCi0iZZcA6TAAvEwUgBmQipZ6yqXwsrtKsAMWgMAoaCweCIZBAUqAA

_No response_

asukaminato0721:
fix #1470

Playground has this issue in the sandbox because its go-to-definition call only returned the first Range, even though the core analyzer already reports every branch of a union.

Now collects every TextRangeWithModule from the transaction.goto_definition instead of truncating to the first match, serializes all of those ranges so the wasm bridge preserves the multiplicity.

### DIFF
--- a/pyrefly/lib/playground.rs
+++ b/pyrefly/lib/playground.rs
@@ -468,16 +468,21 @@ impl Playground {
         SemanticTokensLegends::lsp_semantic_token_legends()
     }
 
-    pub fn goto_definition(&mut self, pos: Position) -> Option<Range> {
-        let handle = self.handles.get(&self.active_filename)?;
+    pub fn goto_definition(&mut self, pos: Position) -> Vec<Range> {
+        let handle = match self.handles.get(&self.active_filename) {
+            Some(handle) => handle,
+            None => return Vec::new(),
+        };
         let transaction = self.state.transaction();
-        let position = self.to_text_size(&transaction, pos)?;
-        // TODO: Support goto multiple definitions
+        let position = match self.to_text_size(&transaction, pos) {
+            Some(position) => position,
+            None => return Vec::new(),
+        };
         transaction
             .goto_definition(handle, position)
             .into_iter()
-            .next()
             .map(|r| Range::new(r.module.display_range(r.range)))
+            .collect()
     }
 
     pub fn autocomplete(&self, pos: Position) -> Vec<AutoCompletionItem> {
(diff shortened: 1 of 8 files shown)


yangdanny97:
The description on #1470 seems to indicate that this issue affects non-playground surfaces as well, and that the issue is that the core analyzer does not return all the expected results. I'm surprised that a playground-specific fix would address it, unless the original issue was mislabeled.

~~Can you please write a test case that encodes the example from the original issue in https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/lsp/definition.rs ? That way we know the LSP works correctly and it cannot regress~~

Edit: Oh wait I see the test, thanks. IDK how i missed it earllier

PR #1571, opened 2025-11-13
fix Feature: inheritance checks for typed dictionaries #52

### LINKED ISSUE #52: Feature: override checks for typed dictionaries & named tuples

asukaminato0721:
fix #52

Tightened TypedDict override validation so mutable fields must keep exactly the parent type, while read-only fields may only narrow to types assignable to the parent.

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -10904,10 +10904,10 @@
     {
       "code": -2,
       "column": 4,
-      "concise_description": "Class member `Y1.x` overrides parent class `X1` in an inconsistent manner",
-      "description": "Class member `Y1.x` overrides parent class `X1` in an inconsistent manner\n  `Y1.x` has type `int`, which is not consistent with `str` in `X1.x` (the type of read-write attributes cannot be changed)",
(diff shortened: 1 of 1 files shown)


yangdanny97:
It seems like we already emit an error for this from the regular multiple inheritance checks. This change appears to modify the error message in some cases & emit a second/duplicate error in other places.

I don't see a correctness improvement here, so is the intent mainly to improve the error message clarity?

asukaminato0721:
> mainly to improve the error message clarity

Yes, also, is there any place left that #52 has not fixed?

yangdanny97:
Given that #52 is complete (based on conformance test results), I don't think we should add a duplicate code path to handle TypedDicts.

I'm also not convinced the error message is clearer. For example, looking at the duplicated error we have:

old error - "Field `x` has inconsistent types inherited from multiple base classes\n  Inherited types include:\n    `int` from `X2`\n    `str` from `Y2`"

new error - "TypedDict field `x` in `XYZ2` has incompatible type `int`; expected `str` from parent TypedDict `Y2`"

IMO the old error is clearer, since it's a multiple inheritance conflict and `XYZ2` does not define `x` directly.

I'm happy to take suggestions for improvements for error messages here, I just don't think we need to add any additional logic.

PR #1573, opened 2025-11-13
fix [feature] emit an error when trying to call an abstract method #592

### LINKED ISSUE #592: [feature] emit an error when trying to call an abstract method

asukaminato0721:
fix #592

Added an early diagnostic plus helper methods to emit abstract-method-call whenever an abstract function, staticmethod, or classmethod is invoked directly (or via a class object), preventing previously silent failures while still allowing instance dispatch.

Introduced the new AbstractMethodCall error kind

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -685,15 +685,51 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let metadata = call_target.function_metadata();
         if let Some(meta) = metadata
             && meta.flags.is_abstract_method
-            && self.should_error_on_abstract_call(&call_target)
         {
-            let method_name = meta.kind.format(self.module().name());
-            self.error(
-                errors,
-                range,
-                ErrorInfo::new(ErrorKind::AbstractMethodCall, context),
-                format!("Cannot call abstract method `{method_name}`"),
-            );
+            // We only raise here for calls where we know the callee is the
+            // abstract definition itself. That includes:
+            //   * direct calls on the defining class object (e.g. Base.build())
+            //   * super() lookups that surface the abstract method
+            // We skip calls via variables of type[Base] for non-final classes,
+            // because those values might point at a concrete subclass.
+            let should_error = match &call_target {
+                CallTarget::Function(_) | CallTarget::FunctionOverload(..) => true,
+                CallTarget::BoundMethod(obj, _) | CallTarget::BoundMethodOverload(obj, ..) => {
+                    // Definite class object?
+                    let mut is_definite_abstract_owner = matches!(obj, Type::ClassDef(_));
+
+                    // super() always refers to a specific class in the MRO
+                    if matches!(obj, Type::SuperInstance(_)) {
+                        is_definite_abstract_owner = true;
+                    }
+
+                    // Final classes can't be refined by subclasses, so treat
+                    // any reference to them as definite.
+                    let is_final = match obj {
+                        Type::ClassDef(cls) => self.get_metadata_for_class(cls).is_final(),
+                        Type::Type(inner) => match &**inner {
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/call.rs
```diff
@@ -856,6 +868,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                self.is_class_object_type(obj)
+            }
+            _ => false,
+        }
+    }
+
+    fn is_class_object_type(&self, ty: &Type) -> bool {
+        match ty {
+            Type::ClassDef(_) => true,
+            Type::Type(inner) => self.is_class_object_type(inner),
```
yangdanny97: I don't think we need this case

pyrefly/lib/test/abstract_methods.rs
```diff
@@ -190,6 +190,51 @@ c = Concrete()
 "#,
 );
 
+testcase!(
```
yangdanny97, line 193: I'd like to see a couple of edge cases from the original issue reflected in this test suite:
- If `X.foo()` is illegal, we should also test that calling `foo` on some arbitrary variable of `type[X]` should NOT error unless `X` is final, since it's possible that the variable refers to a subclass of `X` that implements the method.
- We also need a test for calling abstract methods on `super()`

pyrefly/lib/alt/call.rs
```diff
@@ -856,6 +868,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    fn should_error_on_abstract_call(&self, target: &CallTarget) -> bool {
+        match target {
+            CallTarget::Function(_) | CallTarget::FunctionOverload(..) => true,
```
yangdanny97: Is this code path exercised in a test? Can we have an abstract function/functionOverload be the call target when only methods can be marked abstract?

My hunch is that it's not necessary

pyrefly/lib/alt/call.rs
```diff
@@ -856,6 +868,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            CallTarget::BoundMethod(obj, _) | CallTarget::BoundMethodOverload(obj, ..) => {
+                self.is_class_object_type(obj)
```
yangdanny97: given there's only one call site let's inline the implementation here

pyrefly/lib/alt/call.rs
```diff
@@ -856,6 +868,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    fn should_error_on_abstract_call(&self, target: &CallTarget) -> bool {
```
yangdanny97: Can we make a comment here that we're checking against calls to abstract static and class methods from the class, only when the class name is explicitly provided?

pyrefly/lib/test/abstract_methods.rs
```diff
@@ -190,6 +190,84 @@ c = Concrete()
...
+    test_call_abstract_staticmethod_errors,
+    r#"
+from abc import ABC, abstractmethod
+
+class Base(ABC):
+    @staticmethod
+    @abstractmethod
+    def helper() -> None: ...
+
+Base.helper()  # E: Cannot call abstract method `Base.helper`
```
yangdanny97, line 217: We'll still get an error if we have something like

```
x: type[Base] = ...
x.helper()
```

which could potentially lead to false-positives.

This might not be easy to solve, since we don't have access to whether or not the class name is directly referenced when checking static methods (since it's just CallTarget::Function)

The rest of the changes in the PR look good, so I'll merge everything except for the static method checks, and leave that as a TODO for another PR.

Issue #317, opened 2025-05-20
Inlay hint insertion should consider imports/use qualified paths to auto import correctly 


ckuethe:
### Describe the Bug

When letting pyrefly suggest type annotations on code that imports `re` and calls functions like `re.compile`, the suggested annotation do not include the full module name, at least not in VSCode.

I would have expected that if I `import re`, and use `re.compile()`, the inferred type would be `re.Pattern`, instead of `Pattern`. What if I'm also using another package that has a different something also called `Pattern`? I see that same pattern with `argparse`. If I `import argparse` and say `ap = argparse.ArgumentParser()`, the inferred type is `ArgumentParser`, not `argparse.ArgumentParser`, and the type of `ap.parse_args()` is inferred as `NameSpace`, not `argparse.NameSpace`.

![Image](https://github.com/user-attachments/assets/3db1f55d-e389-4380-98c1-434bed89734c)

Also, what do you think about automatically importing `Any` with `from typing import Any` if I haven't already done so and pyrefly is suggesting `list[Any]`? If I accept that suggestion without having imported `Any`, VSCode tells me it "Could not find name `Any`Pyrefly(unknown-name)"

### Sandbox Link

[...]

PR #1576, opened 2025-11-13
fix Inlay hint insertion should consider imports/use qualified paths to auto import correctly #317

asukaminato0721:
fix part of #317

Type pretty-printing now tracks every module name it emits, so we can tell the client which imports a hint depends on; TypeDisplayContext stores the module set in a RefCell, exposes it via referenced_modules, and records each module.name combination as it formats the text

The LSP transaction layer wraps each hint in a new InlayHintWithEdits struct, renders annotations with fully qualified names, applies any local aliases (e.g., import typing as t), and bundles the necessary import edits by consulting the ImportTracker. This ensures that accepting a type hint also inserts import typing when the module hasn’t been brought into scope.

Downstream consumers now understand the richer hint payload: the server translates InlayHintWithEdits into multiple textEdits, the playground/WASM bindings expose the new label format, and the inlay-hint regression tests cover both standard and notebook scenarios to verify the extra import edits.


crates/pyrefly_types/src/display.rs
```diff
@@ -87,6 +89,7 @@ pub struct TypeDisplayContext<'a> {
     /// Should we display for IDE Hover? This makes type names more readable but less precise.
     hover: bool,
     always_display_module_name: bool,
+    display_modules: RefCell<SmallSet<ModuleName>>,
```
kinto0:
```suggestion
    modules: RefCell<SmallSet<ModuleName>>,
```
maybe we don't need the display_ prefix here since it's already part of the displaycontext struct. a docstring with more information would be nice

pyrefly/lib/state/lsp.rs
```diff
@@ -483,6 +493,145 @@ impl IdentifierWithContext {
     }
 }
 
+#[derive(Default)]
+struct ImportTracker {
```
kinto0: let's put this + the related impl below + the struct + the render/format_type_for_annotation function in it's own file with tests

pyrefly/lib/state/lsp.rs
```diff
@@ -3137,6 +3286,9 @@ impl<'a> Transaction<'a> {
                 }
         };
         let bindings = self.get_bindings(handle)?;
+        let ast_arc = self.get_ast(handle);
+        let ast_ref = ast_arc.as_deref();
+        let import_tracker = ast_ref.map(ImportTracker::from_ast);
```
kinto0: nit: less verbose
```suggestion
        let import_tracker = (*ast).map(ImportTracker::from_ast);
```

pyrefly/lib/state/lsp.rs
```diff
@@ -483,6 +493,145 @@ impl IdentifierWithContext {
...
 }
 
+#[derive(Default)]
+struct ImportTracker {
+    canonical_modules: SmallSet<ModuleName>,
+    alias_modules: Vec<(ModuleName, String)>,
+}
+
+impl ImportTracker {
+    fn from_ast(ast: &ModModule) -> Self {
```
kinto0: can all of these functions have docstrings for what they do? and can ImportTracker have a top level docstring?

pyrefly/lib/test/lsp/lsp_interaction/inlay_hint.rs
```diff
@@ -30,27 +30,39 @@ fn test_inlay_hint_default_config() {
         id: interaction.server.current_request_id(),
         result: Some(serde_json::json!([
             {
-                "label":" -> tuple[Literal[1], Literal[2]]",
+                "label":" -> tuple[typing.Literal[1], typing.Literal[2]]",
```
kinto0: cc @jvansch1 as you are also currently making changes to this

kinto0: can we keep the `Literal` text (without the module) but still have the import edits add the correct import? we have received feedback that the inlay hints are too long and I worry that modules on more types makes this even worse

kinto0 (changes requested):
awesome work! a few suggestions + it looks like there are some merge conflicts. it might also be worth coordinating with @jvansch1 over discord to make sure you're not stepping on each other's toes since he's editing the same file

pyrefly/lib/test/lsp/lsp_interaction/inlay_hint.rs
```diff
@@ -36,6 +36,10 @@ fn test_inlay_hint_default_config() {
                 "textEdits":[{
                     "newText":" -> tuple[Literal[1], Literal[2]]",
```
kinto0:
```suggestion
                    "newText":" -> tuple[typing.Literal[1], typing.Literal[2]]",
```
if we `import typing`, we need to qualify these types with it

kinto0 (changes requested):
looks like the insertion text no longer matches the type

pyrefly/lib/test/lsp/lsp_interaction/inlay_hint.rs
kinto0: I think this is still an issue

kinto0 (changes requested):

asukaminato0721:
Inlay hint labels now render with module qualifiers (e.g. typing.Literal) while skipping builtins and the current module, and respecting import X as alias.

When the module isn’t imported, the inlay hint now includes extra text edits to insert import \<module\> alongside the hint insertion.

LSP tests updated to expect the qualified label parts.

kinto0:
looks like there are still some test failures

kinto0:
looks like some tests are failing

crates/pyrefly_types/src/display.rs
```diff
@@ -153,6 +154,7 @@ pub struct TypeDisplayContext<'a> {
     /// quantified by an enclosing Forall), but shown for free variables from outer scopes
     /// (e.g. `F1@bar.f1` inside a nested function `f2[F2]` — F1 is free, F2 is bound).
     forall_tparam_uniques: RefCell<Vec<QuantifiedIdentity>>,
+    display_modules: RefCell<SmallSet<ModuleName>>,
```
rchen152: Rather than storing display_modules in TypeDisplayContext, could we add an annotation-specific TypeOutput implementation? That way we can avoid having mutable state on the general display context.

pyrefly/lib/lsp/wasm/inlay_hints.rs
```diff
@@ -92,6 +105,45 @@ pub fn normalize_singleton_function_type_into_params(type_: Type) -> Option<Vec<
...
+    }
+
+    fn render_type_hint(
+        &self,
+        ty: &Type,
+        handle: &Handle,
+        tracker: Option<&ImportTracker>,
+        ast: Option<&ModModule>,
+    ) -> RenderedTypeHint {
+        let (mut text, modules) = Self::format_type_for_annotation(ty);
```
rchen152: This transmits and rewrites arbitrary text, which seems a bit fragile. IMO it would be better for `format_type_annotation` to return structured parts. E.g., something like this:
```
enum AnnotationPart {
    Text(String),
    Reference {
        module: ModuleName,
        name: String,
    },
}
```
ImportTracker could resolve each Reference, then concatenate the parts.

pyrefly/lib/state/import_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Returns the locally imported name for a `from module import name` statement.
+    pub fn imported_name_alias(&self, module: ModuleName, name: &str) -> Option<&str> {
+        self.imported_names
+            .get(&module)
+            .and_then(|names| names.get(name))
+            .map(|s| s.as_str())
+    }
+
+    /// Replace imported module prefixes with aliases where possible.
+    pub fn apply_aliases(&self, text: &str) -> String {
```
rchen152: This processes one byte at a time, which will corrupt unicode annotations like `Literal["é"]`.

crates/pyrefly_types/src/display.rs
rchen152, line 240: I believe the new `referenced_modules` will accidentally grab this fake module.

pyrefly/lib/state/import_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+                self.alias_modules
+                    .push((module_name, asname.id.to_string()));
+            } else {
+                self.canonical_modules.insert(module_name);
+            }
+        }
+    }
+
+    /// Record a `from ... import ...` statement into the tracker.
+    pub fn record_import_from(&mut self, stmt_import_from: &StmtImportFrom) {
```
rchen152: This function doesn't seem to be used anywhere? It's a little difficult to tell without tests, but it seems like we might always use `foo.Bar` even when `from foo import Bar` is available?

pyrefly/lib/lsp/wasm/inlay_hints.rs
```diff
@@ -92,6 +105,45 @@ pub fn normalize_singleton_function_type_into_params(type_: Type) -> Option<Vec<
...
+        let mut import_edits = Vec::new();
+        if let (Some(tracker), Some(ast)) = (tracker, ast) {
+            text = tracker.apply_aliases(&text);
+            for module in tracker
+                .missing_modules(&modules, handle.module())
+                .into_iter()
+            {
+                if let Some(handle_to_import) = self.import_handle(handle, module, None).finding() {
+                    let (position, insert_text, _) =
```
rchen152, line 132: So this computes each needed import edit independently, which seems like it would run into trouble if an inlay hint needs multiple imports. We probably need some coordination to ensure we return a non-overlapping and deduplicated set of edits.

rchen152 (changes requested):
Thanks for the PR! I'm not super familiar with some of this code (@kinto0 would probably have the most context), so please feel free to push back if any of my suggestions don't make sense. But my overall feel is that directionally this PR is good, but the mechanism is a bit fragile (passing around and editing raw strings), which leads to issues like incorrect unicode handling. It would be better if we can create structured data representations to pass around. This PR is also missing tests.

pyrefly/lib/state/import_tracker.rs
```diff
@@ -0,0 +1,212 @@
...
+        let target = module.as_str();
+        for (alias_module, alias_name) in &self.alias_modules {
+            let alias_module_str = alias_module.as_str();
+            if alias_module_str.is_empty() {
+                continue;
+            }
+            if target == alias_module_str {
+                return Some(alias_name.clone());
+            }
+            if target.len() > alias_module_str.len()
```
rchen152: This assumes that importing a parent module also loads every descendant module, which may not be true. For example, `import foo as f` does not guarantee that `f.bar` exists, so resolving a reference to `foo.bar.C` as `f.bar.C` may produce an invalid annotation. I think we need to require an exact module match. The same issue applies to has_canonical below.

pyrefly/lib/state/import_tracker.rs
```diff
@@ -0,0 +1,212 @@
...
+            if let Stmt::Import(stmt_import) = stmt {
+                for alias in &stmt_import.names {
+                    let module = ModuleName::from_str(alias.name.as_str());
+                    if let Some(asname) = &alias.asname {
+                        tracker.alias_modules.push((module, asname.id.to_string()));
+                    } else {
+                        tracker.canonical_modules.insert(module);
+                    }
+                }
+            } else if let Stmt::ImportFrom(stmt_import_from) = stmt {
```
rchen152, line 41: This ignores ImportFrom.level, so relative imports are mistakenly treated like absolute imports.

rchen152 (changes requested):
Thanks! This is looking much better. Back to you with a couple more comments on the specifics of the import tracking.

grievejia (approved):

PR #1578, opened 2025-11-14
fix tuple[()] | tuple[int, *tuple[int, ...]] should be equivalent to tuple[int, ...] #273

### LINKED ISSUE #273: `tuple[()] | tuple[int, *tuple[int, ...]]` should be equivalent to `tuple[int, ...]`
### Describe the Bug

Interesting example posted by @jorenham in https://discuss.python.org/t/unbounded-tuple-unions/92472, thanks!

Related to contextual subtyping + type simplification

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #273

Now invokes a new tuple-specific simplifier so the general union collapse path can discard tuple[()] whenever another member already guarantees that the empty tuple case is covered.

Introduces collapse_tuple_unions_with_empty, which rewrites unions.

### DIFF
--- a/crates/pyrefly_types/src/simplify.rs
+++ b/crates/pyrefly_types/src/simplify.rs
@@ -263,17 +263,15 @@ fn collapse_tuple_unions_with_empty(types: &mut Vec<Type>) {
             }
             Type::Tuple(Tuple::Unpacked(unpacked)) => {
                 let (prefix, middle, suffix) = &**unpacked;
-                if prefix.len() + suffix.len() == 1 {
-                    if let Type::Tuple(Tuple::Unbounded(elem)) = middle {
-                        if prefix
-                            .iter()
-                            .chain(suffix.iter())
-                            .all(|fixed| fixed == elem.as_ref())
-                        {
-                            *ty = Type::Tuple(Tuple::Unbounded(Box::new(elem.as_ref().clone())));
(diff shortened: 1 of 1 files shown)


yangdanny97:
This specific change looks fine to me, though weirdly enough it seems like ty no longer does this type of equivalence: https://play.ty.dev/f956d93a-bb1b-4cc9-9e6e-a01bf28c8c16

Jelle's implementation of this might be a good reference for a more complex equivalence we could support, though I don't think we need to prioritize adding it unless people ask for it: https://github.com/JelleZijlstra/pycroscope/pull/28

pyrefly/lib/test/simple.rs
```diff
@@ -1870,6 +1870,17 @@ testcase!(
     "#,
 );
 
+// Ref https://github.com/facebook/pyrefly/issues/273
+testcase!(
+    test_union_empty_tuple_and_variadic_tuple,
+    r#"
+from typing import assert_type
+
+def f(x: tuple[()] | tuple[int, *tuple[int, ...]]):
```
yangdanny97, line 1879: ~~for completeness, could you please add the other cases from the example in https://discuss.python.org/t/unbounded-tuple-unions/92472 ?~~

actually NVM, I'll add it when I import the PR

PR #1579, opened 2025-11-14
fix infer type of parameter from super type #1139

### LINKED ISSUE #1139: infer type of parameter from super type
### Describe the Bug

```py
class A:
    def f(self, i: int):
        ...

class B(A):
    def f(self, i):
        reveal_type(i)  # Unknown, expect int
```

here the unannotated `i` can be inferred from the super type as `int`

### Sandbox Link

_No response_

_No response_

asukaminato0721:
fix #1139

Added inherited_method_signature that walks the class MRO, instantiates parent method fields with the subclass Self, and returns the usable ParamList plus decorator flags so override inference can start from a concrete base signature.

Introduced ParentParamHints and threaded it through undecorated_function and get_params_and_paramspec, allowing each unannotated parameter (except the true self/cls) to pick up the appropriate hinted type before falling back to Any/defaults; get_param_type_and_requiredness now consumes those hints in priority orde.r

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -15,7 +15,6 @@ use pyrefly_derive::TypeEq;
 use pyrefly_derive::VisitMut;
 use pyrefly_python::ast::Ast;
 use pyrefly_python::dunder;
-use pyrefly_python::module_path::ModulePathDetails;
 use pyrefly_types::callable::FuncFlags;
 use pyrefly_types::callable::FuncId;
 use pyrefly_types::callable::FunctionKind;
@@ -2655,13 +2654,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let derived_instance = self.instantiate(cls);
         for ancestor in self.get_mro_for_class(cls).ancestors(self.stdlib) {
             let parent_cls = ancestor.class_object();
-            match parent_cls.module_path().details() {
-                ModulePathDetails::BundledTypeshed(_)
-                | ModulePathDetails::BundledTypeshedThirdParty(_) => {
-                    continue;
-                }
-                _ => {}
-            }
             let Some(member) = self.get_class_member(parent_cls, name) else {
                 continue;
             };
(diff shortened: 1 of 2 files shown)


rchen152:
I like the idea of inheriting method type signatures - having to type out the type annotations again when inheriting from a class gets tedious very quickly. But I share @stroxler's concerns about the non-local inference being potentially confusing and how well it would work on poorly typed code bases.

Random idea: what if we restricted this to methods decorated with `@override` (https://typing.python.org/en/latest/spec/class-compat.html#override)? Then users who want to opt into this behavior could do so by adding `@override` without adding type annotations, untyped code would be unaffected, and anyone reading the code has an explicit indicator that the method exists in a parent class, which would give them a good idea of where to look for type annotations.

pyrefly/lib/test/class_overrides.rs
```diff
@@ -150,6 +150,25 @@ class Derived2(Base):
     "#,
 );
 
+testcase!(
```
stroxler, line 153: Could we add a second method to the example without the explicit `@override`?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -2572,6 +2574,34 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+    pub(crate) fn inherited_method_signature(
+        &self,
+        cls: &Class,
+        name: &Name,
+    ) -> Option<(ParamList, FuncFlags)> {
+        let derived_instance = self.instantiate(cls);
+        for ancestor in self.get_mro_for_class(cls).ancestors(self.stdlib) {
+            let parent_cls = ancestor.class_object();
+            match parent_cls.module_path().details() {
+                ModulePathDetails::BundledTypeshed(_)
```
rchen152: Why do we want to skip classes in bundled typeshed?

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -3284,3 +3314,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .and_then(|attr| attr.as_instance_method())
     }
 }
+
+impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
+    fn callable_params_and_flags(ty: Type) -> Option<(ParamList, FuncFlags)> {
```
rchen152: Could you rewrite this using the helpers in https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_types/src/types.rs? The logic for matching over functions/callables is sort of fiddly, and I'd prefer to keep it all in one place.

I think what you can do is use `transform_toplevel_func_metadata` to grab FuncFlags (I know it has "transform" in its name, but the transformation function can do anything you like, including just copying out the flags): https://github.com/facebook/pyrefly/blob/a4d0bfa8abb78ef1ecffbf6c5eae765be1998c5b/crates/pyrefly_types/src/types.rs#L1144. And use `callable_signatures` to get the function signature that you can then extract the ParamList from: https://github.com/facebook/pyrefly/blob/a4d0bfa8abb78ef1ecffbf6c5eae765be1998c5b/crates/pyrefly_types/src/types.rs#L1317C9-L1317C31.

pyrefly/lib/test/class_overrides.rs
```diff
@@ -150,6 +150,28 @@ class Derived2(Base):
...
```
rchen152, line 153: The test coverage here seems rather sparse for such a complex feature. Could you also add tests for these?
- a method in a subclass that does override a method in a parent but doesn't use the `@override` decorator
- a method that uses both decorator hints and parent hints
- a static method and a class method
- a method that uses more parameter types than just Pos (PosOnly, Kwarg, etc.)

rchen152 (dismissed):
Thanks! Back to you with some comments.

rchen152:
Thanks! I've put this up for internal review.

PR #1580, opened 2025-11-14
fix Pyrefly inserted typing import in sub-optimal location #1490

### LINKED ISSUE #1490: Pyrefly inserted `typing` import in sub-optimal location
### Describe the Bug

When I started typing `ClassVar` in a file in which I hadn't imported `typing.ClassVar`, Pyrefly brought up `ClassVar` as a completion suggestion (very helpful!), but it inserted `from typing import ClassVar` at the very top of my imports list. Ideally, it would have merged `ClassVar` into my existing `from typing import ...` line, or at least inserted the new `typing` import line right above/below the existing one.

Image

### Sandbox Link

_No response_

_No response_

### LINKED ISSUE #3382: VSCode Auto-import inserts imports above `from __future__ import annotations`
### Describe the Bug

VS Code auto-import Quick Fix inserts imports above `from __future__ import annotations`, causing invalid Python syntax. `__future__` imports must remain at the top of the file, but Pyrefly’s code action places new imports before them.

Image

Image

### Sandbox Link

_No response_

VSCode: Version: 1.119.1
meta.pyrefly@1.0.0

asukaminato0721:
fixes #1490 fixes https://github.com/facebook/pyrefly/issues/3382

Adjusted so insert_import_edit* returns an ImportEdit with both the actual edit and a stable display string, and added logic to append new symbols to an existing matching from ... import ... statement instead of duplicating the import line. The helper now shares this richer info with callers so UI text stays human-readable.

Updated LSP code to consume the new struct, ensuring code-action titles and auto-import completion details still show from module import name while the actual edit mutates the existing line in place.

Taught the infer command to handle the new edit type and to skip empty/duplicate insertions when mutating files.

Revised the LSP code-action test to assert the merged-import behavior.


pyrefly/lib/test/lsp/completion.rs
```diff
@@ -100,7 +100,7 @@ fn get_test_report(
                     report.push_str(" with text edit: ");
                     report.push_str(&format!("{:?}", &text_edit));
                 }
-                if let Some(documentation) = documentation {
```
jvansch1, line 150: Why were these changes needed?

pyrefly/lib/state/ide.rs
```diff
@@ -256,12 +263,38 @@ pub fn insert_import_edit_with_forced_import_format(
...
+            ast,
+            module_name_to_import.as_str(),
+            export_name,
+            display_text.clone(),
+            module_name_to_import.as_str(),
+        )
+    {
+        return edit;
+    }
+    let position = if let Some(first_stmt) = ast.body.iter().find(|stmt| !is_docstring_stmt(stmt)) {
```
jvansch1: I wonder if there is some way we can avoid traversing the entire AST again? It might lead to slow performance for large modules.

pyrefly/lib/test/lsp/completion.rs
asukaminato0721: We were trying to match the historical golden strings for completion reports.

Previously, every entry, even plain keywords or literal values, had a blank line appended because the reporter always added one after the optional docstring block. That produced extra empty lines compared to the expected fixtures, which is why dozens of completion tests started failing.

The adjustment keeps the docstring formatting logic intact, but stops unconditionally inserting that trailing blank line unless there was actually extra content to separate. This wa,y the rendered report matches the snapshots again without altering runtime behavior.

pyrefly/lib/state/ide.rs
asukaminato0721: One way to avoid the second walk would be to fold both concerns into a single pass (track the insertion position while scanning for compatible imports)

Or to compute the insertion point once before calling try_extend_existing_from_import and pass it in.

jvansch1:
@asukaminato0721 Are you still planning to make changes to this PR? I see there are some failing checks.

yangdanny97:
@jvansch1 let's get this one merged?

yangdanny97:
Sorry this one's taking a bit longer to review because it's doing a lot of things at once. No action needed from you ATM but in the future a smaller more focused changeset would help us get things reviewed faster.

PR #1608, opened 2025-11-18
Allow recursive definition in functional typed dict

fangyi-zhou:
Consider the following snippet

```
from typing import NamedTuple, Optional, TypedDict

X = NamedTuple("X", [("x", Optional["X"])])
Y = TypedDict("Y", {"y": Optional["Y"]})
```

Pyrefly errors on the line for `TypedDict` but not for `NamedTuple`, which is reflected in conformance testsuite `typeddicts_required.py` (L71).

I've examined the code between processing a functional named tuple declaration and a functional typed dict. The only difference seems to be an extra `.clone()` for typed dict. Somehow that was the actual culprit --- removing the extraneous clone causes the error to go away!

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -11442,17 +11442,6 @@
       "severity": "error",
       "stop_column": 34,
       "stop_line": 60
-    },
-    {
-      "code": -2,
-      "column": 75,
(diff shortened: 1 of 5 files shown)


yangdanny97:
nice, thanks!

stroxler (approved):

PR #1614, opened 2025-11-18
Simplify expect_publish_diagnostics_error_count

samwgoldman:
Summary:
A few changes here:
1. Use the simpler expect_message util, since we always skip over unexpected messages
2. Match on the _parsed_ params type instead of directly on JSON
3. Skip apparently unnecessary path canonicalization (and remove redundant comments)
4. Only wait for the first publishDiagnostics request, then assert about it, instead of waiting for a publishDiagnostics request w/ the expected attributes

I validated (3) just by running tests locally. It's possible this will break when run on Mac/Win, but I'll check CI. The blame diff didn't explain this choice.

For (4), I think the proposed behavior is better, and tests continue to pass. In the scenario where we do expect multiple publishDiagnostics notifications, I think it's better to expect them instead of relying on skipping them. (I expected to need to fix up tests because of this change, but happily everything just worked out.)

Differential Revision: D87360101

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/object_model.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/object_model.rs
@@ -24,6 +24,7 @@ use lsp_server::Request;
 use lsp_server::RequestId;
 use lsp_server::Response;
 use lsp_server::ResponseError;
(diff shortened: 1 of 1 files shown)


kinto0 (approved):

Issue #1348, opened 2025-10-21
on hover lacking parameter highlight for some attributes


asukaminato0721:
### Describe the Bug

similar to #979

some works, some don't.

<img width="750" height="244" alt="Image" src="https://github.com/user-attachments/assets/ca355810-c4ef-40db-971f-9e92b6bd36b6" />

<img width="750" height="170" alt="Image" src="https://github.com/user-attachments/assets/d333138b-e3dd-47c8-a5a4-17be6ea3f555" />

```py
import matplotlib.pyplot as plt

# Set up the figure and axis
fig, ax = plt.subplots(figsize=(10, 6))

ax.legend(loc="upper right", bbox_to_anchor=(0.85, 0.98), fontsize=24)

# Add grid for better readability
ax.grid(True, alpha=0.3, axis="y")

```

### (Only applicable for extension issues) IDE Information

vscode v0.38.1

samwgoldman:
Thanks for the report! The root cause seems to be the same as #979. The hover card for `legend` does not include `def legend`.

asukaminato0721:
<img width="1340" height="328" alt="Image" src="https://github.com/user-attachments/assets/cc153229-7983-4217-b3b9-c911e27cbb86" />

This is another highlight problem, 

```py
from sqlalchemy.orm import DeclarativeBase

class Base(DeclarativeBase):
    metadata = metadata

class AppModelConfig(Base):
    id = mapped_column(StringUUID, server_default=sa.text("uuid_generate_v4()"))
```

I'm not sure if it is another issue, so put it here.

PR #1618, opened 2025-11-19
fix numpy function hover missing name #1418

### LINKED ISSUE #1418: numpy function hover missing name
### Describe the Bug

similar to #971 

Image

```py
import numpy as np

a = np.array([1])
```

### Sandbox Link

_No response_

v0.39.2

asukaminato0721:
fix #1418

image

Added metadata-backed hover names by threading a display_name through every FindDefinitionItemWithDocstring so hover headers can still show identifiers even when we land in stub files or empty ranges.

Updated the hover formatter to try definition snippets, the metadata name, a type-derived fallback, and finally a lexical identifier; this restored headers for third-party symbols and ensured both WASM and native LSP paths behave identically

Introduced rich function rendering: callable tooltips now show (function) name plus a syntactically valid def name(...): ..., improving syntax highlighting in VS Code for cases like np.array.

### DIFF
--- a/crates/pyrefly_types/src/display.rs
+++ b/crates/pyrefly_types/src/display.rs
@@ -776,9 +776,22 @@ impl Display for Type {
 
 impl Type {
     pub fn as_hover_string(&self) -> String {
+        self.as_hover_string_with_fallback_name(None)
+    }
+
+    pub fn as_hover_string_with_fallback_name(&self, fallback_name: Option<&str>) -> String {
         let mut c = TypeDisplayContext::new(&[self]);
         c.set_display_mode_to_hover();
-        c.display(self).to_string()
+        let rendered = c.display(self).to_string();
+        if let Some(name) = fallback_name
+            && self.is_function_type()
+        {
+            let trimmed = rendered.trim_start();
+            if trimmed.starts_with('(') {
+                return format!("def {}{}: ...", name, trimmed);
+            }
+        }
+        rendered
     }
 
     pub fn get_types_with_locations(&self) -> Vec<(String, Option<TextRangeWithModule>)> {
@@ -802,54 +815,6 @@ impl<'a> ClassDisplayContext<'a> {
 
     pub fn display(&'a self, cls: &'a Class) -> impl Display + 'a {
         Fmt(|f| self.0.fmt_qname(cls.qname(), f))
-    }
-}
-
-/// Small helper describing the formatted code that should appear inside a hover block.
-///
-/// `body` is the snippet that will live inside the fenced code block, while
-/// `default_kind_label` lets callers know which `(kind)` prefix to use when the
-/// hover metadata can't provide one (e.g. when resolving a stub-only function).
-#[derive(Debug, Clone)]
-pub struct HoverCodeSnippet {
-    pub body: String,
-    pub default_kind_label: Option<&'static str>,
-}
-
-/// Format the string returned by `Type::as_hover_string()` so that callable types
-/// always resemble real Python function definitions. When `name` is provided we
-/// will synthesize a `def name(...): ...` signature if the rendered type only
-/// contains the parenthesized parameter list. This keeps IDE syntax highlighting
-/// working even when we fall back to hover strings built from metadata alone.
-///
-/// `display` is typically `Type::as_hover_string()`, but callers may pass their
-/// own rendering (for example, after expanding TypedDict kwargs).
-pub fn format_hover_code_snippet(
-    type_: &Type,
-    name: Option<&str>,
-    display: String,
-) -> HoverCodeSnippet {
-    if type_.is_function_type() {
-        let body = match name {
-            Some(name) => {
-                let trimmed = display.trim_start();
-                if trimmed.starts_with('(') {
-                    format!("def {}{}: ...", name, trimmed)
-                } else {
-                    display
-                }
-            }
-            None => display,
-        };
-        HoverCodeSnippet {
-            body,
-            default_kind_label: Some("(function) "),
-        }
-    } else {
-        HoverCodeSnippet {
-            body: display,
-            default_kind_label: None,
-        }
     }
 }
 
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -26,7 +26,6 @@ use pyrefly_types::callable::Param;
 use pyrefly_types::callable::ParamList;
 use pyrefly_types::callable::Params;
 use pyrefly_types::callable::Required;
-use pyrefly_types::display::format_hover_code_snippet;
 use pyrefly_types::types::BoundMethodType;
 use pyrefly_types::types::Forallable;
 use pyrefly_types::types::Type;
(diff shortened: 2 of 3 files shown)


asukaminato0721:
improve #1416

image

asukaminato0721:
fix #1348

image

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -176,16 +179,34 @@ impl HoverValue {
             .clone()
             .unwrap_or_else(|| self.type_.as_hover_string());
 
+        let is_function = self.type_.is_function_type();
+        if is_function && kind_formatted.trim().is_empty() {
```
kinto0: can we put this in the map_or above to avoid the mut?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -176,16 +179,34 @@ impl HoverValue {
...
+            kind_formatted = "(function) ".to_owned();
+        }
+        let trimmed_type_display = type_display.trim_start();
+        let needs_def_wrapper = trimmed_type_display.starts_with('(');
```
kinto0: can we move this logic into the creation of type_display? I don't like the logic for what a type display is being fragmented like this

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -176,16 +179,34 @@ impl HoverValue {
...
+        let (heading, body) = if is_function {
+            let function_name = self.name.clone().unwrap_or_else(|| "function".to_owned());
```
kinto0: when do we use this `function` fall back?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -176,16 +179,34 @@ impl HoverValue {
...
+            let heading = format!("{}{}\n", kind_formatted, function_name);
+            let body = if needs_def_wrapper {
```
kinto0: the other logic for adding `def ` is in pyrefly/crates/pyrefly_types/src/display.rs. can we move this logic there instead?

kinto0 (changes requested):
Thanks for looking into this! A few questions

it's unclear why these names are missing from your summary. are all of these bugs the same situation?

do we need all of these fallbacks? which ones are necessary in which situations?

I would like to see all this logic localized to one section. can this entire PR be a change to `pyrefly/crates/pyrefly_types/src/display.rs`?

asukaminato0721:
Resolver didn’t keep a textual name when the definition expression was empty (e.g., stub attributes, keyword arguments, or type-only nodes). Fix: FindDefinitionItemWithDocstring now always carries a display_name, so when we do have definition metadata, we reuse that string.

Type-only lookup path – when we fail to find any definition (third-party binary stubs), we still need something readable. That’s why we fall back to metadata from Type::Function/Callable/BoundMethod/Forall.

Literal identifier under the cursor – in the degenerate case where neither of the above provides anything (e.g., a failing resolver on a local scratch buffer), we lex the identifier so we can at least echo the token.

crates/pyrefly_types/src/display.rs
```diff
@@ -604,6 +604,47 @@ impl<'a> ClassDisplayContext<'a> {
...
 }
 
+#[derive(Debug, Clone)]
+pub struct HoverCodeSnippet {
+    pub heading: Option<String>,
+    pub body: String,
+    pub default_kind_label: Option<&'static str>,
+}
+
+pub fn format_hover_code_snippet(
```
kinto0: this function doesn't make sense upon reading it with no context here.
- why is it needed when we already have a display for types?
- when should we provide name?
- what is display?
a docstring would help

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -175,17 +172,32 @@ impl HoverValue {
             .display
             .clone()
             .unwrap_or_else(|| self.type_.as_hover_string());
+        let snippet = format_hover_code_snippet(&self.type_, self.name.as_deref(), type_display);
```
kinto0: there should be a comment: why do we have to format_hover_code_snippet when we already have a type_display?

begs the question: can we instead make an `as_hover_string` that takes a fallback name? that would make this API one less step

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -236,6 +248,60 @@ fn expand_callable_kwargs_for_hover<'a>(
         }
     }
 }
+
+fn fallback_hover_name_from_type(type_: &Type, current_module: ModuleName) -> Option<String> {
```
kinto0: can you add a docstring for what this is doing?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -305,16 +374,25 @@ pub fn get_hover(
...
-        )
+        let name = {
+            let snippet = module.code_at(definition_location);
+            if snippet.chars().any(|c| !c.is_whitespace()) {
+                Some(snippet.to_owned())
+            } else if let Some(name) = display_name.clone() {
+                Some(name)
+            } else {
+                fallback_name_from_type.clone()
+            }
```
kinto0, line 473: we can make name not `mut` by putting the logic to fallback to identifier_text_at here

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -236,6 +248,60 @@ fn expand_callable_kwargs_for_hover<'a>(
...
+        _ => None,
+    }
+}
+
+fn identifier_text_at(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    position: TextSize,
+) -> Option<String> {
+    let module = transaction.get_module_info(handle)?;
```
kinto0, line 378: we have a `Module::code_at` function already. can we reuse that and avoid `identifier_text_at` entirely?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -333,3 +411,55 @@ pub fn get_hover(
...
+            signature: Callable::ellipsis(Type::None),
+            metadata,
+        }))
+    }
+
+    #[test]
+    fn fallback_uses_function_metadata() {
+        let ty = make_function_type("numpy", "arange");
+        let fallback = fallback_hover_name_from_type(&ty, ModuleName::from_str("user_code"));
+        assert_eq!(fallback.as_deref(), Some("numpy.arange"));
```
kinto0: can we instead just use the function name and exclude the module? we don't put the module name here in other contexts

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -175,17 +172,32 @@ impl HoverValue {
...
                 kind: MarkupKind::Markdown,
                 value: format!(
-                    "```python\n{}{}{}\n```{}{}",
-                    kind_formatted,
-                    name_formatted,
-                    type_display,
-                    docstring_formatted,
-                    symbol_def_formatted
+                    "```python\n{}{}\n```{}{}",
+                    heading, snippet.body, docstring_formatted, symbol_def_formatted
```
kinto0: I don't love the heading since it repeats what exists on the next line. can we implement this in a way that keeps the appearance the same?

kinto0 (changes requested):

pyrefly/lib/lsp/wasm/hover.rs
asukaminato0721: Module::code_at takes a TextRange; hovering supplies only a cursor offset. We still need to expand that offset into the full identifier span before we can call code_at. That expansion (walking left/right while characters are [_0-9A-Za-z]) is exactly what identifier_text_at does.

PR #1623, opened 2025-11-19
fix [deprecated] error should show the deprecation message #1582

### LINKED ISSUE #1582: deprecated method missing the info
### Describe the Bug

Image

and if I check the source

```
    @typing_extensions.deprecated('The `schema` method is deprecated; use `model_json_schema` instead.', category=None)
```

it has message.

### Sandbox Link

_No response_

v0.41.3

asukaminato0721:
fix #1582

Added a shared helper that parses `@deprecated` decorators (including attribute forms), captures the optional message, and exposes formatting utilities. Wired this through exports, so Export now carries structured.

DeprecatedDecoration data, and all binding/LSP/state code that synthesizes Export instances were updated accordingly.

Function/class metadata now stores deprecation messages, so solver diagnostics, overload checks, attribute completions, etc., emit “foo is deprecated: <reason>” when provided.

Binding structures and type flags were extended so decorated definitions retain the message through compilation phases; BindingClassMetadata also tracks class-level deprecations.

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -51,6 +51,7 @@ use crate::binding::binding::Key;
 use crate::binding::pydantic::PydanticConfigDict;
 use crate::binding::pydantic::VALIDATION_ALIAS;
 use crate::config::error_kind::ErrorKind;
+use crate::deprecation::DeprecatedDecoration;
 use crate::error::collector::ErrorCollector;
 use crate::error::context::ErrorInfo;
 use crate::error::style::ErrorStyle;
@@ -114,6 +115,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         is_new_type: bool,
         pydantic_config_dict: &PydanticConfigDict,
         django_primary_key_field: Option<&Name>,
+        deprecated: Option<&DeprecatedDecoration>,
         errors: &ErrorCollector,
     ) -> ClassMetadata {
         // Get class decorators.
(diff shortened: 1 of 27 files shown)


pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -258,9 +260,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let is_final = decorators.iter().any(|(decorator, _)| {
             decorator.ty().callee_kind() == Some(CalleeKind::Function(FunctionKind::Final))
         });
-        let is_deprecated = decorators
+        let mut is_deprecated = decorators
```
fangyi-zhou, line 263: Do we still need this? Could we do `let is_deprecated = deprecated.is_some()`?

pyrefly/lib/export/exports.rs
```diff
@@ -41,6 +42,7 @@ pub struct Export {
     pub symbol_kind: Option<SymbolKind>,
     pub docstring_range: Option<TextRange>,
     pub is_deprecated: bool,
```
fangyi-zhou, line 44: Do we still need this field? We could use `deprecation.is_some()` instead

fangyi-zhou (commented):
LGTM, left some nits!

pyrefly/lib/alt/class/class_metadata.rs
asukaminato0721: We still need is_deprecated, independent of the optional DeprecatedDecoration. Deprecated currently only reflects syntactic markers we carry forward (e.g., decorator messages captured on binding). decorators.iter().any(|…| decorator.ty().is_deprecation_marker()) continues to cover semantic cases discovered later (like inferred deprecations on inherited decorators or when we only see types, not original AST).

Keeping the existing scan ensures class metadata mirrors the solver’s previous behavior even when the binding step didn’t attach a DeprecatedDecoration. If you want to drop the scan, you’d also have to guarantee that deprecated is set for every path—including synthesized classes, protocol fallbacks, etc. —which isn’t true yet.

pyrefly/lib/export/exports.rs
asukaminato0721: Callers (e.g., LSP hover/completions) currently rely on a cheap boolean check to tag items without needing the richer decoration (and some exports come from stubs or synthetic sources where we don’t populate deprecation). 

Switching to deprecation.is_some() would be equivalent only if every deprecated export had a DeprecatedDecoration. That isn’t the case: we keep boolean-only flags for compatibility when we detect deprecation, but don’t have a message.

PR #1638, opened 2025-11-20
Fix: Correctly detect await in nested generator expressions

### LINKED ISSUE #1611: Comprehensions are inferred as Generator not AsyncGenerator when await keyword is nested in the condition
### Describe the Bug

```python
import asyncio
import typing

async def some_async_func(x: int) -> bool:
    return x % 2 == 0

async def main() -> None:
    generator = (x for x in [1, 2, 3] if await some_async_func(x) == True)
    typing.reveal_type(generator)
    assert await anext(generator) == 2

# asyncio nonsense to make this work in the sandbox
asyncio.ensure_future(main(), loop=asyncio.get_event_loop())
```
In the code snippet above, Pyrefly infers the type of generator to be `Generator[int, None, None]` rather than `AsyncGenerator[int, None]`.

I previously raised #1299 (very similar, but note the change to the condition expression in the generator) for the case where the await expression comprised the full condition of the generator, and that has since been fixed. However, I have now noticed that it doesn't work in the case where the `await` is nested inside the condition expression. I skimmed the code to try to understand what was going on, and I suspect it is because the visitor that visits that expression:
https://github.com/facebook/pyrefly/blob/cd06d6700296b54f7d00f15acdfa82f77c587fcd/crates/pyrefly_python/src/ast.rs#L331-L335
doesn't recurse like the example does in the documentation of `visit`:
https://github.com/facebook/pyrefly/blob/d6f28bef41aeccca62669df9dc6a82afc20767b2/crates/pyrefly_util/src/visit.rs#L47-L52

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIEAtsbgE4AuABKnKegMYS4A661tRkwaliEdAHM%2BfVuw5NMMMEzi4qMAPqzOGsAFdOACnyIm4hgEomAWgB8TPLiiI%2BTN0zowGeuuib4mAFImACYmAF5wpgAGaXRteUVlKlRxQys7JgA5XHQYFz93CRg8ulQGegimYyYwSoDxJgBtAEYAGlCOgGYAXTNlVAB3VOZVdS02HX0jfCtIpgAVOj0YC1d3ETFJQk8ANxhUKA1NmENi0vL6NcK3VjgYISGRljz8BjOSh8u6OaiQuIAxCxJlxcEx0Ll7uh7sIwSkANYwYQACwgcCYg3o8LMfgYyKRaCweHwMhB3EIJTgPk0%2Bm8nkMKTSFg6UFwuGI4QS5OKDA0MH26F5rPZ6QsIDaIDInjAUFIhAqVCgFCBAAVSNLZSoMDgCEwOLlIBIfOVuOhCHwgQBlGBI5EMBjEOCIAD0zqlSllhHoEmdJWdmFwHDgzv16ENxoYpudtUqqF2qSgqGwsD1BogRrKkdyTHZWeh5vQZDxuWs%2BzocFNVR4IC6hBa-xAfBAAF8JagOJH9gAxaAwCiEnVEMgtoA

_No response_

AryanBagade:
Fix #1611: Correctly detect await in nested generator expressions

Description:
The contains_await() function was not recursing into child expressions, causing generator comprehensions with await nested inside comparisons (e.g., `if await func() == True`) to be incorrectly typed as Generator instead of AsyncGenerator.
Fixed by rewriting contains_await() to follow the visitor pattern correctly, ensuring it recursively checks all child expressions for await keywords.

Added test case that reproduces the exact bug from issue #1611

### DIFF
--- a/crates/pyrefly_python/src/ast.rs
+++ b/crates/pyrefly_python/src/ast.rs
@@ -331,11 +331,14 @@ impl Ast {
 
     pub fn contains_await(expr: &Expr) -> bool {
         let mut found = false;
-        expr.visit(&mut |node: &Expr| {
-            if matches!(node, Expr::Await(_)) {
-                found = true;
+        // Recursive function that checks this node and recurses to children
+        fn check(expr: &Expr, found: &mut bool) {
+            if matches!(expr, Expr::Await(_)) {
+                *found = true;
             }
-        });
+            expr.recurse(&mut |child: &Expr| check(child, found));
+        }
+        expr.visit(&mut |node: &Expr| check(node, &mut found));
         found
     }
 }
--- a/pyrefly/lib/test/yields.rs
+++ b/pyrefly/lib/test/yields.rs
@@ -446,6 +446,21 @@ async def test_implicit_generators() -> None:
 );
 
 testcase!(
+    test_async_generator_comprehension_with_nested_await,
+    r#"
+from typing import AsyncGenerator, assert_type
+
+async def some_async_func(x: int) -> bool:
+    return x % 2 == 0
+
+async def main() -> None:
+    # await is nested inside a comparison expression in the condition
+    generator = (x for x in [1, 2, 3] if await some_async_func(x) == True)
+    assert_type(generator, AsyncGenerator[int, None])
+"#,
+);
+
+testcase!(
     bug = "We don't understand yield in lambda, and misattribute the yield to the surrounding function",
     test_lambda_yield,
     r#"


AryanBagade:
We can make the approach iterative as well and stops when found, no need to check entire tree, let me know if you want to go through this approach!

stroxler (approved):
Looks good to me, I'll try to get this merged

PR #1641, opened 2025-11-20
populate display name when untyping a type alias

yangdanny97:
Summary:
- populate display name when untyping a union
- keep name when simplifying
- display the expanded type when the top level type being printed is the alias (only use the display name for nested types)

Differential Revision: D87558543


rchen152 (approved):

Issue #881, opened 2025-08-10
check_type is no longer narrowing types (regression from 0.27.0 -> 0.27.1)


kilroy42:
### Describe the Bug

Test code:

```python
from typeguard import check_type
from typing import reveal_type

x: dict[str, int] | float = {}
reveal_type(x)
x = check_type(x, dict[str, int])
reveal_type(x)
```

In pyrefly 0.27.0 (correct output):

```
 INFO revealed type: dict[str, int] | float [reveal-type]
  |
7 | reveal_type(x)
  |            ---
  |
 INFO revealed type: dict[str, int] [reveal-type]
  |
9 | reveal_type(x)
  |            ---
  |
 INFO 0 errors
```

In pyrefly 0.27.1 (wrong output):

```
 INFO revealed type: dict[str, int] | float [reveal-type]
  |
7 | reveal_type(x)
  |            ---
  |
 INFO revealed type: dict[str, int] | float [reveal-type]
  |
9 | reveal_type(x)
  |            ---
  |
 INFO 0 errors
```

It should have narrowed the type down to the dict with `x = check_type(x, dict[str, int])`

BTW: I would have appended a sandbox link but the sandbox fails with "Pyrefly encountered an internal error: {}." - even with the demo code.

vedant1022:
kindly assign me this work , i'll do needfull

samwgoldman:
I caused this regression in a206d0075d1be9977e883d0158065bb0c8186b4c. When we see `x = check_type(...)` we use the expected type of `x`, which is `dict[str, int] | float` to instantiate the type parameter `T`. This is what we want for pushing down hints, but we want to actually instantiate `T` using the argument.

@vedant1022 appreciate your willingness to help, but I think this is not a great issue. I'll take this one on.

stroxler:
#1635 (closed in favor of this issue) another example of the same problem

rchen152:
I believe this is fixed - the original example no longer errors, and all of the examples from the deduped issues now work correctly as well, except https://github.com/facebook/pyrefly/issues/1143#issuecomment-3873306428. It's not obvious to me that that one's even the same issue, so I'll open a new bug for it.

PR #1647, opened 2025-11-21
fix: prevent hints from poisoning generic function inference in OR expressions

### LINKED ISSUE #1635: Bug Report: wrong type-hint
### Describe the Bug

```python

import os
from pathlib import Path 

def load_config(config_path: str | None = None):
    config_path = None

    config_path = config_path or os.getenv("CONFIG_PATH", "./configs/deploy.yaml")

    config_file = Path(config_path)

    return config_file

print(load_config())
```

Python: 3.12

`config_path` should be `str` instead of `str | None` in Path function. 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAOuhALbG4BOALgAS5zViO62vGpmACygRsrOgxasACoKGtq1TDDCsouVJgD6AY1zpIAcwAU%2BwxCPaBwxKzjNGrAD6sAcgZisAvO88BKRGpWENZzY2t5Hz90GCV0ULCDCJsFX3DLSOF2Jw5CIxhmGHQANxNKEABhAHk3ADEASQBxbRkAQQAVAAkKgBpFEEIAegyjOCGVYg1SQlJUWigK-3jE0e1IWGi5YTNkzNTl9GDQxkKAV0YEtY24o-RiRgh0ZhMNLT09039-EF6QM%2BY0DgJHIiBAAGJWABVQGiZikVhgM7oXSAgycO4qNRgJi0QTadBnWjYGCMEz4OxPZj%2BVgAWgAfPZHEEEidzpdERU3ESSYw7MB8ABfCrUX4gMinMBQGbMXhQCiQmSkSXS%2BwYHAEJIWIwXQQQAyEaiQgDKMC8QmYzGIcEQQyGEtU0sITCMQ2KE1wunGo11aPQQ0RTFYqBKqGgqGwmx9jD1BnYxD9wOoZGEBlpJVJcH1CV8FQAzIQAIwAJhF6BAgr%2BqFREAzdWgMAoaCweCIZArQA

_No response_

Imran-S-heikh:
## Summary
Fixes #1635.

This PR fixes a type inference bug where generic functions (like `os.getenv`) returned incorrectly widened types when used on the right-hand side of an `or` expression.

**The Issue:**
When reassigning a variable like `config: str | None`, the type checker passed the variable's current type (`str | None`) as a "context hint" to the right-hand side of the `or`. Generic functions like `os.getenv(key, default)` accepted this hint, widening their return type to match the hint (inferring `T` as `None`) rather than inferring the specific type from their arguments (`default="string"` -> `T=str`).

This resulted in false positive type errors (e.g., `Argument 'str | None' is not assignable...`) even when a valid default value was provided.

## The Fix
I updated `boolop` in `expr.rs` to adjust how hints are propagated in `OR` expressions:

- **For Function Calls (`Expr::Call`):** The context hint is now dropped. This forces the function to infer its return type strictly from its arguments (Inside-Out inference), preventing "poisoning" from the surrounding context.
- **For Literals/Others:** The context hint is preserved. This ensures that expressions like `x: List[int] = None or []` still correctly infer `List[int]` instead of `List[Any]`.

## Test Plan
Added a regression test `test_or_generic_function_hint_poisoning_fix` in `operators.rs`.

The test simulates the `os.getenv` behavior using a generic `identity` function and verifies that:
1. A variable defined as `str | None` set to `None`.
2. Reassigned via `variable or identity("default")`.
3. Correctly narrows to `str` (proving the hint `str | None` was ignored by the function inference).

**Verification:**
Ran tests locally:
`cargo test -p pyrefly --lib test_or_generic_function_hint_poisoning_fix` (Passed)

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1073,16 +1073,23 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let mut t_acc = Type::never();
         let last_index = values.len() - 1;
         for (i, value) in values.iter().enumerate() {
-            // If there isn't a hint for the overall expression, use the preceding branches as a "soft" hint
-            // for the next one. Most useful for expressions like `optional_list or []`.
-            let hint = hint.or_else(|| {
-                if t_acc.is_never() {
-                    None
-                } else {
-                    Some(HintRef::soft(&t_acc))
-                }
-            });
-            let mut t = self.expr_infer_with_hint(value, hint, errors);
+            let operand_hint = if matches!(op, BoolOp::Or) && matches!(value, Expr::Call(_)) {
+                // Drop the hint for function calls in 'OR' expressions.
+                // This prevents the expected type (e.g. Optional[str]) from improperly influencing
+                // a generic function's inference (forcing it to return Optional[str]
+                // instead of str), while still preserving hints for literals like [].
+                None
+            } else {
+                // Use external hint, or fall back to previous branches
+                hint.or_else(|| {
+                    if t_acc.is_never() {
+                        None
+                    } else {
+                        Some(HintRef::soft(&t_acc))
+                    }
+                })
+            };
+            let mut t = self.expr_infer_with_hint(value, operand_hint, errors);
             self.expand_vars_mut(&mut t);
             // If this is not the last entry, we have to make a type-dependent decision and also narrow the
             // result; both operations require us to force `Var` first or they become unpredictable.
--- a/pyrefly/lib/test/operators.rs
+++ b/pyrefly/lib/test/operators.rs
@@ -681,3 +681,33 @@ def f[S, T](x: type[S], y: type[T]):
     return x == y
     "#,
 );
+
+testcase!(
+    test_or_generic_function_hint_poisoning_fix,
+    r#"
+from typing import TypeVar, assert_type
+
+T = TypeVar("T")
+
+def get_value(default: T) -> T:
+    return default
+
+def test(config: str | None):
+    # 1. Setup: config is explicitly set to None.
+    # The variable 'config' still carries the type hint 'str | None'.
+    config = None
+    
+    # 2. The Test:
+    # We reassign to 'config' using an OR expression with a generic function call.
+    #
+    # Without the fix, the 'str | None' hint from the variable definition flows 
+    # into 'get_value', causing it to infer T='str | None' instead of T='str'.
+    #
+    # With the fix, the hint is dropped for the function call, so it infers 
+    # T='str' purely from the argument "default".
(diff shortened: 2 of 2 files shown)


pyrefly/lib/alt/expr.rs
```diff
@@ -1073,16 +1073,23 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-            // for the next one. Most useful for expressions like `optional_list or []`.
-            let hint = hint.or_else(|| {
-                if t_acc.is_never() {
-                    None
-                } else {
-                    Some(HintRef::soft(&t_acc))
-                }
-            });
-            let mut t = self.expr_infer_with_hint(value, hint, errors);
+            let operand_hint = if matches!(op, BoolOp::Or) && matches!(value, Expr::Call(_)) {
```
stroxler, line 1076: I'm not convinced this is specific to `or`; for example if we had
```
x: None | bool | str
```
and then
```
x = (not x) and get_value("default")
```
then we probably should be able to narrow `x` to `bool | str`, but the existing logic would pass down the hint and we'd get `None | bool | str`.

------

More generally, I think what might really be going on is that return types shouldn't be used contextually in the way they are now, because (using your test case as an example)
```
config = get_value("default")
```
really ought to result in `config` being narrowed to `str`, and it's not even in a bool op - I think the real issue is that we should only be using context when it's necessary to make the assignment legal; in cases where we can get a narrower type we want the narrower type

cc @samwgoldman for thoughts, it's unclear to me how hard this would be to do. We might be able to just analyze the function twice when necessary, similar to how we handle overloads

samwgoldman: Is this just #881? Agree that the way we contextually type calls to generic functions (and ctors) is wrong. My plan was to treat these hints differently, but I only have a sketch of an idea. 

If this PR is working around a specific instance of 881 I think we should probably work on the underlying issue instead.

stroxler: Ah, yeah I forgot about #881, I think this probably is just another case of it.

After giving it more thought, it seems like if the constraint solver understood that a return type hint is only an upper bound (and should not affect the result unless necessary) then we'd get the right answer.

I'm guessing that's at least roughly what your idea is?

I do think it's likely possible to use two attempts at solving the call to get this behavior, similar to overloads. But assuming the solver can do it natively in one pass that seems better

stroxler (changes requested):
Marking as changes requested since I think we likely want a different approach.

For what it's worth I think the diagnosis is right and the fix works, but we probably want a broader solution than this.

yangdanny97:
closing the PR because the original issue has been closed

PR #1650, opened 2025-11-21
fix: preserve generics when applying function decorators

### LINKED ISSUE #1646: Contextmanager lossing Generics
### Describe the Bug

when using generics inside a contextmanager the generic is lost 

```python
from contextlib import contextmanager
from typing import Iterator


@contextmanager
def add_source_channel[T](source: list[T]) -> Iterator[list[T]]: ...

t: list[int] = [1]
m = add_source_channel(t) # not working  ->  _GeneratorContextManager[list[Unknown], None, None] 
# should be _GeneratorContextManager[list[int], None, None] 


def add_source_channel2[T](source: list[T]) -> Iterator[list[T]]: ...

m = add_source_channel2(t) # working just fine

```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEAxrugC4z4tQTZ0Q3G5KLRszYcaGVAHMYlADrow1ei1LEI6Kb36DhASTaVULQQrPoAAk1bsWE9NNkLMMMHVSZMAfTi4ArpQMMF4MABYY6DBQANoAKgC6ABS%2BAUGIdFxwLHHxAJR0ALQAfHQGssaC0ZnZCfHphA3mLOnV0Ros8XQAvHTRAIzxCvQ9Ht4pgcFhEVGJLLnmzq7unj7%2BEyHh6JFQAEw5yWtpGRBZOfnFpYYVlFUnNfF1dA2E5sPLY4eTm9s7s-PoIAANCA-CxoHASORECAAMR0ACqYK4qjoYD86AYYOYcHMLjcYEEEhYXnQfho2FkiXw6Xa5xKWUoiAUdBZdEoMBYAXQqLkIAAcmSKYy6MB8ABfXkKIEgMjssBQUiEEw0KAUOEABVIcoVdDQWDw%2BBEiggUgCxggzBe6DhAGUYDA6KEWCxiHBEAB6d2y1wKwiCKTumDod2YXAMODu6yQU1GLHB1GCdwAN1Q0FQ2FgRujZrjdFwxDjEIUZBYoWYBSTsjgFu5PV5AGZCH0dpKAWLgahMRBKwAxaAwCh6nAESEgMVAA

_No response_

Imran-S-heikh:
## Summary
Fixes #1646.

Previously, using decorators like `@contextmanager` on generic functions caused the generic type parameters to be lost (resolving to `Unknown` or unbound variables). The `apply_function_decorator` logic did not propagate the `Forall` binders through the decorator's type inference process.

## The Fix
I updated `apply_function_decorator` in `function.rs` to:
1. **Unwrap** the `Forall` body of the generic function before passing it to `call_infer`. This ensures the type inference sees the raw function signature with its type variables intact.
2. **Re-wrap** the result in `Forall` using the original type parameters, provided the decorator didn't return a new generic structure itself.

This ensures that `Iterator[T] -> ContextManager[T]` transformations correctly bind `T`.

## Test Plan
Added `test_contextmanager_preserves_generics` in `decorators.rs`.
- Verifies that `@contextmanager` on a function `[T](x: T) -> Iterator[T]` produces a `ContextManager[T]`.
- Verifies that calling this manager with `int` results in `ContextManager[int]`, not `ContextManager[Unknown]`.

Ran tests locally: `cargo test -p pyrefly --lib test_contextmanager_preserves_generics` (Passed)

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -28,6 +28,7 @@ use ruff_python_ast::StmtFunctionDef;
 use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
+use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
 use vec1::Vec1;
 
@@ -885,13 +886,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         errors: &ErrorCollector,
     ) -> Type {
         // Unwrap Forall if present.
+        // We pass the BODY of the generic function to the decorator.
+        // This prevents `call_infer` from replacing the type variables with fresh inference variables (placeholders).
         let (tparams_opt, decoratee_arg) = match &decoratee {
             Type::Forall(forall) => (Some(forall.tparams.clone()), forall.body.clone().as_type()),
             _ => (None, decoratee.clone()),
         };
 
+        // Preserve function metadata, so things like method binding still work.
         let call_target =
             self.as_call_target_or_error(decorator, CallStyle::FreeForm, range, errors, None);
+
         let arg = CallArg::ty(&decoratee_arg, range);
 
         let inferred_ty =
@@ -940,31 +945,49 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 returned_ty => returned_ty,
             };
 
-        // Re-wrap with *relevant* original Type Parameters.
+        // Re-wrap with relevant type parameters
         if let Some(tparams) = tparams_opt {
-            if !matches!(inferred_ty, Type::Forall(_)) {
-                // FIX: Wrap this analysis in a block { ... }
-                // This ensures `used_quantifieds` (which borrows `inferred_ty`) is dropped
-                // BEFORE we try to move `inferred_ty` in the match statement below.
-                let relevant_tparams: Vec<TParam> = {
-                    let mut used_quantifieds = SmallSet::new();
-                    inferred_ty.collect_quantifieds(&mut used_quantifieds);
+            // Identify which original tparams are actually used in the result
+            // We scope this in a block to drop the borrow on inferred_ty immediately after scanning
+            let relevant_tparams_vec: Vec<TParam> = {
+                let mut used_quantifieds = SmallSet::new();
+                inferred_ty.collect_quantifieds(&mut used_quantifieds);
+                tparams
+                    .iter()
+                    .filter(|p| used_quantifieds.contains(&p.quantified))
+                    .cloned()
+                    .collect()
+            };
 
-                    tparams
-                        .iter()
-                        .filter(|p| used_quantifieds.contains(&p.quantified))
-                        .cloned()
-                        .collect()
+            if !relevant_tparams_vec.is_empty() {
+                let new_tparams = Arc::new(TParams::new(relevant_tparams_vec));
+
+                return match inferred_ty {
+                    // Case A: Result is already Forall (e.g. decorator adds params)
+                    // We must merge our preserved params with the new ones.
+                    Type::Forall(box forall) => {
+                        let mut merged_tparams = (*new_tparams).clone();
+                        merged_tparams.extend(&forall.tparams);
+                        // We re-wrap the inner body with the merged list
+                        forall.body.forall(Arc::new(merged_tparams))
+                    }
+
+                    // Case B: Result is Callable/Function (Standard case)
+                    // We wrap it in Forall.
+                    Type::Function(f) => Forallable::Function(*f).forall(new_tparams),
+                    Type::Callable(c) => Forallable::Callable(*c).forall(new_tparams),
+
+                    // Case C: Result is NOT callable (e.g. list[T])
+                    // We cannot wrap it in Forall. To prevent leaking unbound variables,
+                    // we must replace them with Any.
+                    ty => {
+                        let any = Type::any_implicit();
+                        let substitution_map: SmallMap<_, _> =
+                            new_tparams.iter().map(|p| (&p.quantified, &any)).collect();
+
+                        ty.subst(&substitution_map)
+                    }
                 };
-
-                if !relevant_tparams.is_empty() {
-                    let new_tparams = Arc::new(TParams::new(relevant_tparams));
-                    return match inferred_ty {
-                        Type::Function(f) => Forallable::Function(*f).forall(new_tparams),
-                        Type::Callable(c) => Forallable::Callable(*c).forall(new_tparams),
-                        _ => inferred_ty,
-                    };
-                }
             }
         }
 
--- a/pyrefly/lib/test/decorators.rs
+++ b/pyrefly/lib/test/decorators.rs
@@ -587,3 +587,52 @@ def test(x: int, items: List[int]):
     assert_type(m, contextlib._GeneratorContextManager[List[int], None, None])
 "#,
 );
+
+// Decorator returns non-callable (The "Leak" Case)
+testcase!(
+    test_decorator_returns_non_callable_leaks_sanitized,
+    r#"
+from typing import TypeVar, Callable, List, assert_type, Any
+
+T = TypeVar("T")
+
+# A decorator that returns a list
+def list_decorator(f: Callable[[T], T]) -> List[T]: ...
+
+@list_decorator
+def my_func(x: T) -> T:
+    return x
+
+def test():
+    # The result is list[T], but since it's not a function, we can't wrap it in Forall[T].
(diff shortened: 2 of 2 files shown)


stroxler (approved):
Thanks for the contribution!

This looks like it might work to me, but it's tricky so I'll aim to get a second review from someone more experienced in our call resolution code

stroxler:
Just thinking about the problem at a higher level - this fix seems reasonable, but I think an open question might be whether we should actually be doing this not just for decorators but for all calls

rchen152 (changes requested):

rchen152:
I'm skeptical that this is the right solution. contextmanager preserves the generic-ness of its input because of how it's typed: https://github.com/python/typeshed/blob/cc792321dd482a5336225229bf584ad51a3e85cd/stdlib/contextlib.pyi#L89. I don't think it's true in general that a decorator should do this.

pyrefly/lib/alt/function.rs
```diff
@@ -884,54 +884,81 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        // This handles decorators like `@contextmanager` that transform a generic function
+        // into a generic callable (e.g. `Iterator[T]` -> `ContextManager[T]`).
+        if let Some(tparams) = tparams_opt {
+            if !matches!(inferred_ty, Type::Forall(_)) {
+                return match inferred_ty {
+                    Type::Function(f) => Forallable::Function(*f).forall(tparams),
+                    Type::Callable(c) => Forallable::Callable(*c).forall(tparams),
+                    // If the result isn't a callable (e.g. an instance), we cannot wrap it in Forall.
+                    _ => inferred_ty,
+                };
```
stroxler: Assuming that other concerns get addressed I think this logic may not be sufficient

First off, if `inferred_ty` *is* a `Forall`, it might still close over some of the `tparams`; for example I think it might be possible to run into this with
```
def decorator(f: Callable[[T], T]) -> Callable[[T, S], tuple[T, S]):
    return lambda x, y: (f(x), y)
```
We would probably get the wrong final type and also leak unbound `Quantified` in this case

Related, I think it's also possible for us to return something that *doesn't* meet these restrictions, and we need to catch this and eliminate the tparams to avoid leaking unbound `Quantified`
```
def decorator(f: Callable[[T], T]) -> list[T]:
    # admittedly it's impossible to actually implement this, but someone could certainly write down the type
    
# I think we would need the final type here to be `list[Unknown]`
@decorator
def f(x): return x
```

stroxler:
> I'm skeptical that this is the right solution. contextmanager preserves the generic-ness of its input because of how it's typed: https://github.com/python/typeshed/blob/cc792321dd482a5336225229bf584ad51a3e85cd/stdlib/contextlib.pyi#L89. I don't think it's true in general that a decorator should do this.

I do think there are some gaps in the logic - in particular it seems like the return type could leak unbound Quantified.

If those gaps were fixed, I think this would still be a risky change in that it's hard to reason about it (part of the problem is the typing library - a lack of closed-form representation for `ForAll` means we're stuck with implicit behavior) but I'm not convinced it doesn't work. Do you have examples in mind.

In particular when you say "because of how it's typed", what exactly about the signature do you think is sufficient to propagate the ForAll-ness?

pyrefly/lib/alt/function.rs
stroxler: Another related example from @rchen152 is where some part of `tparams` *doesn't* appear in the returned type, then we get a weird dangling forall where the quantified aren't used; this probably doesn't actually matter for semantics but it would be confusing:
```
from typing import Callable, reveal_type
def decorate(_f) -> Callable[[int], int]:
    def g(x):
        assert type(x) is int
        return x
    return g
    
@decorate
def f[T](x: T) -> T:
    return x

reveal_type(f)
```

stroxler:
@Imran-S-heikh FWIW the tests are mostly failing because of lack of autoformatting, I think you can run `cargo fmt` to take care of that

Imran-S-heikh:
@stroxler Thank you for the detailed feedback\! You are right about the risk of leaking unbound `Quantified` variables if the decorator transforms the type (e.g. `Decorator -> int`).

 I have updated the logic to address this:

 1.  **Filtering:** I now scan the `inferred_ty` (the result of the decorator) to collect used quantified variables.
 2.  **Selective Wrapping:** I only re-apply the `tparams` that are *actually present* in the result.

 This ensures that:

   * `Iterator[T] -> ContextManager[T]` preserves `T`.
   * `Iterator[T] -> int` does **not** get wrapped as `Forall[T]. int`.

regarding @rchen152's point: I agree that ideally the typeshed signature would handle this. However, currently `call_infer` seems to instantiate the input function (stripping the `Forall`) before matching it against the signature. Since no arguments are present during the decorator call to constrain `T`, it resolves to `Unknown`. This PR prevents that premature erasure by unwrapping/rewrapping manually.

I've also ran `cargo fmt`. Ready for re-review!

pyrefly/lib/alt/function.rs
```diff
@@ -884,54 +884,91 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                        .cloned()
+                        .collect()
+                };
+
+                if !relevant_tparams.is_empty() {
+                    let new_tparams = Arc::new(TParams::new(relevant_tparams));
+                    return match inferred_ty {
+                        Type::Function(f) => Forallable::Function(*f).forall(new_tparams),
+                        Type::Callable(c) => Forallable::Callable(*c).forall(new_tparams),
+                        _ => inferred_ty,
```
stroxler: I think we're still missing logic here - we actually know for sure that `relevant_tparams` is nonempty, which means we'll definitely leak a `Quantified` if we don't promote it here.

The example I have in mind is still
```
def decorator(f: Callable[[T], T]) -> list[T]): ...
@decorator
def f(x): ...
```
which I think would result in `list[@T]` at this point, we do need to sweep up the unusable quantifieds and turn it into `list[Any]`

pyrefly/lib/alt/function.rs
```diff
@@ -884,54 +884,91 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-                } else {
-                    cls.to_type()
+                }
+                Type::ClassType(cls) if cls.has_qname("functools", "_Wrapped") => decoratee.clone(),
+                returned_ty => returned_ty,
+            };
+
+        // Re-wrap with *relevant* original Type Parameters.
+        if let Some(tparams) = tparams_opt {
+            if !matches!(inferred_ty, Type::Forall(_)) {
```
stroxler: I still think we're missing some handling for the `else` case here: if the returned type adds an *additional* tparam, it will probably actually be a `Type::Forall` - the example
```
def decorator(f: Callable[[T], T]) -> Callable[[T, S], tuple[T, S]):
    return lambda x, y: (f(x), y)
```

This is a problem because if the returned type is `forall S . Callable[[@T, S], tuple[@T, S]]` (where I'm using `@T` to indicate `T` as a Quantified type rather than a type parameter) then we'll leak `@T`

What we want to do is probably rewrite the type into `forall T, S . Callable[[T, S], tuple[T, S]]`; at the very least we should convert to `Callable[[Any, S], tuple[Any, S]]` but I think we can do better by aggregating the `forall`

stroxler (commented):
The  fix to use only relevant tparams looks great to me.

I still think we're lacking two more hooks though, both of which are related to potentially leaking quantified:
- it seems like if the returned type is *not* callable, we're still leaking `Type::Quantified`
- it also seems like we're mishandling the case where `inferred_ty` already is a `Forall`; in that case, the `Forall` needs to be expanded

See inline comments for details and examples (which we should probably include as unit tests)

Note that since both of these scenarios require `relevant_tparams`, we likely want to move the `relevant_tparams` collection to be outside of the
```
matches!(inferred_ty, Type::Forall(_))
```
because I think we're going to need them whether or not the inferred type is a forall

Imran-S-heikh:
@stroxler Thanks for the detailed guidance! I've pushed a new commit that addresses the edge cases you identified:

1.  **Refactoring:** I moved the `relevant_tparams` calculation outside the match block so it runs for all cases.
2.  **Merging:** If the decorator returns a `Type::Forall` (e.g. adding a new generic `S`), I now **merge** the original `T` with the new `S` instead of nesting them.
3.  **Sanitizing:** If the decorator returns a non-callable (e.g. `list[T]`), I now substitute the leaked type variables with `Any` to prevent unbound variables.
4.  **Tests:** I added regression tests in `decorators.rs` covering both the "Merge" and "Leak" scenarios.
5.  **Formatting:** Ran `cargo fmt`.

Ready for re-review!

rchen152:
@Imran-S-heikh thanks for adding the additional checks! That addressed my concerns about this unwrapping/re-wrapping happening when it shouldn't.

Issue #1632, opened 2025-11-20
Pyrefly VS Code extension ignores the interpreter selected in “Python: Select Interpreter”


c0rychu:
### Describe the Bug

Even if I selected the Python interpreter in VS Code via `> Python: Select Interpreter`, the Pyrefly extension does not respect that. For instance, in an open `.py` file, it complains about `Could not find import ...` like
```
Could not find import of `rich.console`
  Looked in these locations (from config in `/Users/username/XXXXX/pyproject.toml`):
  Search path (from config file): ["/Users/username/XXXXX/src"]
  Import root (inferred from project layout): "/Users/username/XXXXX/src"
  Site package path queried from interpreter: ["/Users/username/.pyenv/versions/3.12.12/lib/python3.12", "/Users/username/.pyenv/versions/3.12.12/lib/python3.12/lib-dynload", "/Users/username/.pyenv/versions/3.12.12/lib/python3.12/site-packages"]
Pyrefly(missing-import)
```

Only if I set
```
[tool.pyrefly]
python-interpreter-path = ".venv/bin/python"
```
in my `pyproject.toml`, the Pyrefly extension can then find the correct interpreter.

### (Only applicable for extension issues) IDE Information

VS Code: 
```
Version: 1.106.2 (Universal)
Commit: 1e3c50d64110be466c0b4a45222e81d2c9352888
Date: 2025-11-19T16:56:50.023Z
Electron: 37.7.0
ElectronBuildId: 12781156
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.1.0
```

Pyrefly Extension:
```
Identifier
meta.pyrefly
Version
0.42.2
```

kinto0:
Sorry about the issue! 

us finding the `/Users/username/.pyenv/versions/3.12.12/lib/python3.12` interpreter is surprising. can you confirm a few things?
- are you using the new [Microsoft python-environments extension](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-python-envs) or the normal Python (`ms-python.python`) extension? (we don't yet support the new one as it's in early adopters)
- how many workspace folders do you have?
- where is the workspace folder root compared to your pyrefly.toml/project root? (we prioritize scope of the workspace folder over the pyrefly.toml)
- which workspace folder setting did you set the interpreter to?
- do you have a "workspace level" interpreter set in addition to the workspace-folder level interpreter setting?

[...]

PR #1654, opened 2025-11-21
clear interpreter cache on changes

kinto0:
Summary:
[bug report](https://github.com/facebook/pyrefly/issues/1632#issuecomment-3560579415)

  Root Cause:
  1. PythonEnvironment::get_interpreter_env() caches interpreter environments in a global static registry (INTERPRETER_ENV_REGISTRY at environment.rs:30-32)
  2. When invalidate_config() is called (state.rs:1317), it clears the ConfigFinder cache
  3. ConfigFinder::clear() (finder.rs:195-199) calls clear_extra_caches()
  4. But clear_extra_caches (config_finder.rs:145-154) does NOT clear INTERPRETER_ENV_REGISTRY
  5. Result: When you switch interpreters, the cached PythonEnvironment from the old interpreter persists

  This explains the LSP trace perfectly:
  - First switch .venv → miniforge3: Works (no cache yet)
  - Second switch miniforge3 → .venv: Broken - diagnostics show miniforge3's site-packages even though .venv interpreter is selected

  The Fix

  Add a public function to clear the interpreter cache and call it from clear_extra_caches:

  When the interpreter changes:
  1. Extension sends DidChangeConfigurationNotification
  2. Server calls request_settings_for_all_workspaces() → gets new pythonPath
  3. workspace_configuration_response() sets modified = true
  4. invalidate_config_and_validate_in_memory() is called
  5. config_finder.clear() now clears the interpreter cache
  6. Next config load re-queries the interpreter for site-packages
  7. Diagnostics show correct site-packages for the new interpreter

  This ensures that switching interpreters immediately invalidates all cached environment data, forcing a fresh query of the new interpreter's site-packages.

Differential Revision: D87643729

### DIFF
--- a/crates/pyrefly_config/src/environment/environment.rs
+++ b/crates/pyrefly_config/src/environment/environment.rs
@@ -220,6 +220,13 @@ print(json.dumps({'python_platform': platform, 'python_version': version, 'site_
         &INTERPRETER_STDLIB_PATH_REGISTRY
     }
 
+    /// Clear the cached interpreter environments.
+    /// This should be called when the workspace configuration changes
+    /// to ensure we re-query interpreters for their current environment.
+    pub fn clear_interpreter_cache() {
+        INTERPRETER_ENV_REGISTRY.lock().clear();
+    }
+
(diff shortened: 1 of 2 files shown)


kinto0:
does not fix

PR #1658, opened 2025-11-21
Disable completions in comments

### LINKED ISSUE #1371: completion popup menu appears in a top level comment
### Describe the Bug

I don't know 100% if the issue here is with pyrefly, vim or the ALE plugin.
If this isn't pyrefly, please close :)

I use vim in a lightweight environment where I also have pyrefly available so I wanted to wire it up with ALE so that vim can have syntax checking, type checking, auto completions, etc.

I got it set up, and it is mostly working great but there is a quirk I notice.

If I type a dot (period) inside the scope of a comment that's in a function or class scope, then as I would expect, the autocomplete menu does _not_ pop up.

But if I type a period inside a comment that's in global/module scope, it pops out the autocomplete menu, which I would not expect.

Python code which demonstrates the issue:
```
# bad comment
def a_func() -> None:
    # good comment
# also good comment
    pass
```

Pyrefly version:
```
# pyrefly --version
pyrefly 0.38.2
```

Python version:
```
# python3 --version
Python 3.14.0
```

.vimrc file:
```
# cat ~/.vimrc
" --- basics ---
set nocompatible
set signcolumn=yes
set encoding=utf-8
syntax on
filetype plugin indent on

" Make CursorHold fire faster (default is 4000 ms)
set updatetime=300

" --- ALE: pyrefly linting and lsp, no fixers ---
let g:ale_linters = { 'python': ['pyrefly'] }
let g:ale_lint_delay = 250    " ms (default is 200)
let g:ale_fix_on_save = 0

" Re-lint on normal-mode edits (undo/redo, dd, p, etc.)
let g:ale_lint_on_text_changed = 'normal'

" Still lint after leaving insert (so you get feedback after typing)
let g:ale_lint_on_insert_leave = 1

" Also lint on saves
let g:ale_lint_on_save = 1

" ASCII signs to avoid E239
let g:ale_sign_error = 'E'
let g:ale_sign_warning = 'W'

" --- statusline that won't throw E119 ---
function! LintStatus() abort
  let l:c = ale#statusline#Count(bufnr(''))
  return printf('E:%d W:%d', l:c.error + l:c.style_error, l:c.warning + l:c.style_warning)
endfunction
set laststatus=2
set statusline=%<%f\ %h%m%r%=%-14.(%l,%c%V%)\ %P\ %{LintStatus()}

" Turn on ALE-backed completion and route omnifunc to ALE
let g:ale_completion_enabled = 1
autocmd FileType python setlocal omnifunc=ale#completion#OmniFunc
set completeopt=menuone,noinsert,noselect

" If a buffer ever ends up with filetype=pyi, tell ALE to treat it like python
let g:ale_linter_aliases = get(g:, 'ale_linter_aliases', {})
let g:ale_linter_aliases.pyi = 'python'

" --- binds: Jump between diagnostics
nmap <silent> ]d <Plug>(ale_next_wrap)
nmap <silent> [d <Plug>(ale_previous_wrap)
```

ALE linter setup:
```
# cat ~/.vim/ale_linters/python/pyrefly.vim
if exists('g:loaded_ale_linters_python_pyrefly') | finish | endif
let g:loaded_ale_linters_python_pyrefly = 1

function! ale_linters#python#pyrefly#GetExecutable(buffer) abort
  return 'pyrefly'
endfunction

function! ale_linters#python#pyrefly#GetCommand(buffer) abort
  " Start Pyrefly's LSP server
  return 'pyrefly lsp'
endfunction

call ale#linter#Define('python', {
\   'name': 'pyrefly',
\   'lsp': 'stdio',
\   'executable': function('ale_linters#python#pyrefly#GetExecutable'),
\   'command': function('ale_linters#python#pyrefly#GetCommand'),
\   'project_root': function('ale#python#FindProjectRoot'),
\})
```

vim version:
```
# vim --version
VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Sep 15 2025 16:38:04)
Included patches: 1-948, 950-967
Modified by team+vim@tracker.debian.org
Compiled by team+vim@tracker.debian.org
Huge version without GUI.  Features included (+) or not (-):
+acl               +file_in_path      +mouse_urxvt       -tag_any_white
+arabic            +find_in_path      +mouse_xterm       -tcl
+autocmd           +float             +multi_byte        +termguicolors
+autochdir         +folding           +multi_lang        +terminal
-autoservername    -footer            -mzscheme          +terminfo
-balloon_eval      +fork()            +netbeans_intg     +termresponse
+balloon_eval_term +gettext           +num64             +textobjects
-browse            -hangul_input      +packages          +textprop
++builtin_terms    +iconv             +path_extra        +timers
+byte_offset       +insert_expand     -perl              +title
+channel           +ipv6              +persistent_undo   -toolbar
+cindent           +job               +popupwin          +user_commands
-clientserver      +jumplist          +postscript        +vartabs
-clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +vim9script
+cmdline_hist      +langmap           -python            +viminfo
+cmdline_info      +libcall           +python3           +virtualedit
+comments          +linebreak         +quickfix          +visual
+conceal           +lispindent        +reltime           +visualextra
+cryptv            +listcmds          +rightleft         +vreplace
+cscope            +localmap          -ruby              +wildignore
+cursorbind        -lua               +scrollbind        +wildmenu
+cursorshape       +menu              +signs             +windows
+dialog_con        +mksession         +smartindent       +writebackup
+diff              +modify_fname      +sodium            -X11
+digraphs          +mouse             -sound             +xattr
-dnd               -mouseshape        +spell             -xfontset
-ebcdic            +mouse_dec         +startuptime       -xim
+emacs_tags        +mouse_gpm         +statusline        -xpm
+eval              -mouse_jsbterm     -sun_workshop      -xsmp
+ex_extra          +mouse_netterm     +syntax            -xterm_clipboard
+extra_search      +mouse_sgr         +tag_binary        -xterm_save
-farsi             -mouse_sysmouse    -tag_old_static
   system vimrc file: "/etc/vim/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
 3rd user vimrc file: "~/.config/vim/vimrc"
      user exrc file: "$HOME/.exrc"
       defaults file: "$VIMRUNTIME/defaults.vim"
  fall-back for $VIM: "/usr/share/vim"
Compilation: gcc -c -I. -Iproto -DHAVE_CONFIG_H -Wdate-time -g -O2 -Werror=implicit-function-declaration -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/build/vim-RmNAuM/vim-9.1.0967=. -flto=auto -ffat-lto-objects -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -mbranch-protection=standard -fdebug-prefix-map=/build/vim-RmNAuM/vim-9.1.0967=/usr/src/vim-2:9.1.0967-1ubuntu6 -DSYS_VIMRC_FILE=\"/etc/vim/vimrc\" -DSYS_GVIMRC_FILE=\"/etc/vim/gvimrc\" -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1
Linking: gcc -Wl,-Bsymbolic-functions -flto=auto -ffat-lto-objects -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -o vim -lm -ltinfo -lselinux -lsodium -lacl -lattr -lgpm -L/usr/lib/python3.13/config-3.13-aarch64-linux-gnu -lpython3.13 -ldl -lm
```

### Sandbox Link

_No response_

_No response_

kshitijgetsac:
fixes #1371
The disabled_ranges_for_module function now detects three types of ranges where completions should be disabled:

Unreachable code (existing) - code after return, break, etc.
Comments (new) - lines starting with #
String literals (new) - text inside quotes

Implementation

Extended disabled_ranges_for_module() to accept a Module parameter for accessing source text
Added collect_comment_ranges() - scans source line-by-line using existing find_comment_start_in_line() utility
Added collect_string_literal_ranges() - walks AST to find StringLiteral nodes
Updated all call sites to pass the module_info parameter
@yangdanny97

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2635,6 +2635,7 @@ impl<'a> Transaction<'a> {
         results.dedup_by(|item1, item2| item1.label == item2.label && item1.detail == item2.detail);
         (results, is_incomplete)
     }
+
     fn completion_disabled_ranges_for_module(module: &ModuleInfo) -> Vec<TextRange> {
         let mut ranges = Vec::new();
         let source = module.lined_buffer().contents();
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -2420,3 +2420,33 @@ Completion Results:
         report.trim(),
     );
 }
+
+#[test]
+fn test_no_completion_in_comments() {
+    let code = r#"
+# This is a comment
+#      ^
+import sys
+x = sys.version
+#       ^
+"#;
+    let (handles, state) = mk_multi_file_state(&[("main", code)], Require::indexing(), true);
+    let handle = handles.get("main").unwrap();
+    let positions = extract_cursors_for_test(code);
+    let txn = state.transaction();
+
+    // Position 0: inside comment - should return empty
+    let comment_completions = txn.completion(&handle, positions[0], ImportFormat::Absolute, true);
+    assert!(
+        comment_completions.is_empty(),
+        "Expected no completions in comment, but got {} completions",
+        comment_completions.len()
+    );
+
+    // Position 1: normal code - should return completions
+    let normal_completions = txn.completion(&handle, positions[1], ImportFormat::Absolute, true);
+    assert!(
+        !normal_completions.is_empty(),
+        "Expected completions in normal code but got none"
+    );
+}


pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -433,8 +433,67 @@ fn collect_disabled_ranges_from_stmt(
...
+}
+
+fn collect_string_ranges_from_expr(expr: &Expr, ranges: &mut Vec<TextRange>) {
+    match expr {
+        Expr::StringLiteral(string_lit) => {
+            ranges.push(string_lit.range());
+        }
+        _ => {}
+    }
+}
```
kinto0: we are actually planning [to suggest completions on string ilterals ](https://github.com/facebook/pyrefly/pull/1515)so I don't think we want to disable them

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -433,8 +433,67 @@ fn collect_disabled_ranges_from_stmt(
     }
 }
 
-pub(crate) fn disabled_ranges_for_module(ast: &ModModule, sys_info: &SysInfo) -> Vec<TextRange> {
```
kinto0, line 436: i'm not sure this semantic tokens disabled range is what we want to change. this stuff (which should have a comment), turns off semantic tokens for the spots instead of greying out text. it's great to reuse code, but I don't think the scope of these is the same

kshitijgetsac:
@kinto0 
Hi! Thanks for the feedback on the scope concern.
I've removed the string literal detection since you mentioned planning to support completions in strings. The fix now only adds comment detection to disabled_ranges_for_module.
However, I'm concerned about your comment that disabled_ranges_for_module is for semantic tokens (greying out text) rather than disabling completions.
My questions:

Should completion blocking use the same disabled_ranges as semantic tokens, or should they be separate?
I see disabled_ranges is checked in completion_sorted_opt_with_incomplete - is that the right place, or should comment detection be elsewhere?
Would you prefer a separate function like completion_disabled_ranges_for_module that only handles comments (not unreachable code)?

The current implementation works (completions are blocked in comments), but I want to make sure it's architecturally correct before finalizing.

pyrefly/lib/state/semantic_tokens.rs
yangdanny97: Maybe we can disable it for triple-quoted string literals only?

yangdanny97:
> Should completion blocking use the same disabled_ranges as semantic tokens, or should they be separate?

Separate

> I see disabled_ranges is checked in completion_sorted_opt_with_incomplete - is that the right place, or should comment detection be elsewhere?

I can't find this check - could you please give a code pointer?

> Would you prefer a separate function like completion_disabled_ranges_for_module that only handles comments (not unreachable code)?

Yes. Additionally, the string detection AST traversal utils should go in a different place, not in semantic_tokens.rs

kshitijgetsac:
@yangdanny97 Thanks for asking for the code pointer. After thorough searching, I can confirm there is no check for disabled ranges in the completion code currently.
The completion flow is:

server.rs:~2315 - LSP handler calls transaction.completion_with_incomplete()
lsp.rs:2642 - completion_with_incomplete() wrapper for sorting
lsp.rs:2666 - completion_sorted_opt_with_incomplete() generates completions

None of these check if the position is in a comment. That's why autocomplete appears in comments.
Proposed Fix:
Add the check in completion_with_incomplete() (lsp.rs:2642), right after line 2648:
Questions:
Is completion_with_incomplete() the right place for this check?
Where should the new completion_disabled_ranges_for_module() function live?
Should I create a new file (e.g., completion_disabled_ranges.rs) or add it to an existing file?

yangdanny97:
> Is completion_with_incomplete() the right place for this check?

Seems reasonable to me

> Where should the new completion_disabled_ranges_for_module() function live? Should I create a new file (e.g., completion_disabled_ranges.rs) or add it to an existing file?

Probably right under `completion_with_incomplete` for now. We might move all the completion stuff to separate file later but that's out of scope for this PR.

kshitijgetsac:
@yangdanny97 could you review these changes and let me know if they are alright

yangdanny97:
Two things, could you run cargo fmt, and could you write a test case in https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/lsp/completion.rs ?

pyrefly/lib/state/lsp.rs
```diff
@@ -2626,6 +2635,22 @@ impl<'a> Transaction<'a> {
         results.dedup_by(|item1, item2| item1.label == item2.label && item1.detail == item2.detail);
         (results, is_incomplete)
     }
+    fn completion_disabled_ranges_for_module(module: &ModuleInfo) -> Vec<TextRange> {
```
yangdanny97, line 2639: I think there needs to be an empty line before this fn

Issue #43, opened 2025-03-27
Generic ParamSpec


yangdanny97:
paramspecs should be able to resolve to parameters from generic functions or constructors of generic classes, but currently it does not work

(edit by Steven) Sandbox link: [[Sandbox](https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDCAhgDYlEBGJApgDRREDOj1eA%2BvAtQFDcAm1YFGABtACoBdABQAPAFxQxASigBaAHyK53KLqghqMAK4gUUGbwFC0IgFS2ACvQBK04AuJlKNEU6iuVDUJScipqXxcJbT0oK0wUFFYpWyIQNEYFBwA6VPT6ewBrAHdcjKhs4tKlaJi9A2NTYWTS-NtKtMYlHTrDEzNURJBeAAsoAF4oNClgLqYWdk5qKWGpAEYlelQYWeZWGA5EJZWAImONqEYYECUgA&version=3.12)]

The sandbox link below by @lolpack now works, but the functionality here is still not yet implemented

lolpack:
Sandbox example: https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAKAhiERAMoICmAxgDRQAqiVAaiQwMJEA2PRAIx5UGAcSooqIJDQBQoSFGABXFDRhgwPAM6ZsufAHdSCbbMbFSEXQF5CJMpVoAKAEQWH11wEpzAJSoYZRAUKDtmanYQN0YAoJCfczCmFiiYxJp+bV1GLCpKIhQALlkoMqgAEypgKAB9WppeHnrnbSoeYAYVNSKobj5BYQBtDyttBljA4JQAXW8oAFoAPj6mwaoRyzJxpjjpmZLyo6gAAWMiU2duml9j8qqa84RqaIAqEjRtXtHtgDoPnavV4Aa0MAO+W2sv1BAPmy12UxChzuxxAiNC12c7xAnwYQJhOO0txRaPioSeL1kmSI2SgACEkGhOFltM5xJJpDQRnNkWUTrkIPkEIVnPMoABiJgACyQum0UrAyh4FSgKDAhighlwwNK92qSlUNFa7U6qrIVG+cJWjF5qPRZsFQA

[...]

PR #1660, opened 2025-11-22
fix: Generic ParamSpec for constructors

jackulau:
## Summary

For #43

This fixes a type checking bug where generic class constructors would lose their type parameters when passed through ParamSpec functions. The type checker now correctly preserves type variables (`@_`) instead of finalizing them to `Unknown`, allowing proper type inference.

## The Issue

When passing a generic class constructor through a function using ParamSpec, the type parameters were incorrectly finalized to `Unknown`. For example:

```rust
def identity[**P, R](x: Callable[P, R]) -> Callable[P, R]:
    return x

class C[T]:
    def __init__(self, x: T) -> None:
        self.x = x

c2 = identity(C)
reveal_type(c2)  # ERROR: (x: Unknown) -> C[Unknown]
```

At runtime, generic class constructors should preserve their type parameters for inference. However, Pyrefly's subset checking in `subset.rs` wasn't instantiating fresh type variables for generic classes, causing them to be finalized to `Unknown` during ParamSpec inference.

## The Fix

I made several changes to handle generic class constructors properly:

1. Modified `CallTarget::Class` in pyrefly/lib/alt/call.rs:16 to use `TargetWithTParams`:
```rust
Class(TargetWithTParams<ClassTarget>),
```

2. Added fresh variable instantiation for generic classes in pyrefly/lib/solver/subset.rs:34:
```rust
let fresh_class = self.type_order.instantiate_fresh_class(&class, self.type_order.solver);
```

3. Added helper methods in pyrefly/lib/solver/type_order.rs:17 for accessing class type parameters and instantiating fresh class instances.

This ensures generic type parameters remain as solver variables (`@_`) that can be unified during inference, rather than being prematurely finalized to `Unknown`.

## Test Plan

Updated the existing test case `test_param_spec_generic_constructor` in pyrefly/lib/test/paramspec.rs:

- Removed the `bug = "Generic class constructors don't work with ParamSpec"` marker
- Changed expected output from `# E: revealed type: (x: Unknown) -> C[Unknown]` to `# E: revealed type: (x: @_) -> C[@_]`
- Added verification line `x: C[int] = c2(1)` to ensure type inference works

Verification:
- Ran test locally:
```bash
cargo test test_param_spec_generic_constructor
```
Result: passes

- Ran all ParamSpec tests to ensure no regressions:
```bash
cargo test paramspec
```
Result: passes

### DIFF
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -43,7 +43,6 @@ use crate::error::context::TypeCheckKind;
 use crate::solver::type_order::TypeOrder;
 use crate::types::callable::Callable;
 use crate::types::callable::Function;
-use crate::types::callable::Param;
 use crate::types::callable::Params;
 use crate::types::module::ModuleType;
 use crate::types::simplify::simplify_tuples;
@@ -546,14 +545,7 @@ impl Solver {
                         let mut representative_quantifieds: Vec<Quantified> = Vec::new();
 
                         for param in params.items() {
-                            let ty = match param {
-                                Param::PosOnly(_, ty, _) => ty,
-                                Param::Pos(_, ty, _) => ty,
-                                Param::VarArg(_, ty) => ty,
-                                Param::KwOnly(_, ty, _) => ty,
-                                Param::Kwargs(_, ty) => ty,
-                            };
-                            ty.universe(&mut |t| {
+                            param.as_type().universe(&mut |t| {
                                 if let Type::Var(v) = t {
                                     if let Some(q) = self.get_quantified_for_var(*v) {
                                         var_to_representative.insert(*v, q.clone());
@@ -593,11 +585,16 @@ impl Solver {
                         });
 
                         // Build new tparams from representative Quantifieds
+                        //
+                        // TODO: We use PUndefined (Bivariant) because Quantified doesn't store
+                        // variance and Variable::Quantified only stores Quantified, not TParam.
+                        // The proper fix would be to track variance through fresh_quantified.
+                        // PUndefined is safe (won't cause false errors) but may miss some errors.
                         let new_tparams: Vec<TParam> = representative_quantifieds
                             .into_iter()
                             .map(|q| TParam {
                                 quantified: q,
-                                variance: PreInferenceVariance::PInvariant,
+                                variance: PreInferenceVariance::PUndefined,
                             })
                             .collect();
 
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -12,9 +12,11 @@ use std::cell::RefMut;
 use std::fmt;
 use std::fmt::Display;
 use std::mem;
+use std::sync::Arc;
 
 use pyrefly_types::quantified::Quantified;
 use pyrefly_types::simplify::intersect;
+use pyrefly_types::type_var::PreInferenceVariance;
 use pyrefly_types::types::TArgs;
 use pyrefly_util::gas::Gas;
 use pyrefly_util::lock::Mutex;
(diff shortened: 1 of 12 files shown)


yangdanny97:
Thanks! This change looks good, though I don't think this completely handles #43, per sam's comment here: https://github.com/facebook/pyrefly/issues/43#issuecomment-3226102136

yangdanny97:
Hmm, one issue I found is that the type variable is not being solved by the call arguments.

In the test case you added, I see that `x: C[int] = c2(1)` works. We should also be able to infer `C[int]` from the argument to `c2`, wihtout the hint. However,  doing a reveal_type on y after `y = c2(1)` gives `C[Unknown]`

yangdanny97:
I think the solution for #43 probably requires making a new Forallable type for ParamSpecValue and possibly also Concatenate

That would allow passing the type params when we solve a ParamSpec, which would enable us to properly synthesize a generic function type later.

rchen152:
> I think the solution for #43 probably requires making a new Forallable type for ParamSpecValue and possibly also Concatenate
> 
> That would allow passing the type params when we solve a ParamSpec, which would enable us to properly synthesize a generic function type later.

I wonder if we might be able to do something similar to https://github.com/facebook/pyrefly/pull/1650; the fix there is specific to decorators, but we might be able to extend it.

pyrefly/lib/solver/subset.rs
```diff
@@ -567,6 +571,66 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
         }
     }
 
+    /// Create a ParamSpecValue type, wrapping it in Forall if it contains quantified variables
+    /// (either directly as Type::Quantified or as Type::Var backed by Variable::Quantified)
+    ///
+    /// IMPORTANT: We do NOT substitute Vars with Quantifieds here. The Vars must be preserved
+    /// so that after unification (e.g., T unified with R), we can correctly resolve all
+    /// references to the same representative Quantified in simplify_mut.
+    fn create_paramspec_value(&self, params: ParamList) -> Type {
```
yangdanny97, line 580: cc @samwgoldman for review of this var-related logic

pyrefly/lib/solver/subset.rs
```diff
@@ -567,6 +571,66 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
...
+            Type::ParamSpecValue(params)
+        } else {
+            // Create TParams from the quantified variables
+            // Note: We keep the original Vars in the params - substitution will happen
+            // later in simplify_mut after all unifications are complete.
+            let tparams: Vec<TParam> = quantifieds
+                .into_iter()
+                .map(|q| TParam {
+                    quantified: q,
+                    variance: PreInferenceVariance::PInvariant,
```
yangdanny97: Is setting all the type vars to invariant safe to do? What happens if the type var is explicitly set to be covariant?

pyrefly/lib/solver/solver.rs
```diff
@@ -505,6 +523,94 @@ impl Solver {
...
+                        // 3. Substitute Vars with representative Quantifieds in both params and ret
+                        // 4. Build new tparams from the representative Quantifieds
+
+                        let mut params = paramlist.prepend_types(ts).into_owned();
+
+                        // Collect Vars from params and their representative Quantifieds
+                        let mut var_to_representative: SmallMap<Var, Quantified> = SmallMap::new();
+                        let mut representative_quantifieds: Vec<Quantified> = Vec::new();
+
+                        for param in params.items() {
```
yangdanny97, line 547: this code looks pretty similar to the code in subset.rs - is there a way to reuse it?

NathanTempest:
Closing this PR as this has not seen activity in quite a while. Will re-open if the user responds with a follow-up.

PR #1664, opened 2025-11-22
Fix: Allow list[str] for pandas DataFrame columns and index parameters

### LINKED ISSUE #1657: Flags using list[str] for `columns` and `index` on pd.DataFrame
### Describe the Bug

Extension flags errors for using list[str] for `columns` and `index` on `pandas.DataFrame` , even though both accept array-like

`
Argument `list[str]` is not assignable to parameter `columns` with type `ExtensionArray | Index | SequenceNotStr[Unknown] | Series | ndarray[tuple[Any, ...], dtype[Any]] | range | None` in function `pandas.core.frame.DataFrame.__init__`

Argument `list[str]` is not assignable to parameter `index` with type `ExtensionArray | Index | SequenceNotStr[Unknown] | Series | ndarray[tuple[Any, ...], dtype[Any]] | range | None` in function `pandas.core.frame.DataFrame.__init__`
`

code
```
import pandas as pd 
df = pd.DataFrame(columns=['a', 'b', 'c'], index=['a','b','c'])
```

### Sandbox Link

_No response_

_No response_

jackulau:
# Summary

Fixes #1657.

This fixes a type checking bug where Pyrefly incorrectly flagged errors when using `list[str]` for `columns` and `index` parameters in the `pd.DataFrame` constructor:

```python
import pandas as pd
df = pd.DataFrame(columns=['a', 'b', 'c'], index=['a','b','c'])
# Error: Argument `list[str]` is not assignable to parameter `columns`
# Error: Argument `list[str]` is not assignable to parameter `index`
```

The root cause is pandas 2.x's incorrect `SequenceNotStr` protocol definition in typeshed. The protocol's `index()` method signature is:

```python
def index(self, value: Any, /, start: int = 0, stop: int = ...) -> int
```

This makes only `value` position-only, but `list.index()` requires ALL parameters to be position-only. Since `list.index(value, start, stop, /)` doesn't structurally match the protocol (parameter kinds differ), `list[str]` fails to satisfy the protocol constraint.

## The Fix

Added corrected pandas stubs to pyrefly's bundled typeshed matching the fix from pandas main branch
(targeted for pandas 2.3/3.0). The corrected `SequenceNotStr` protocol in `pandas/_typing.pyi` now defines:

```python
def index(self, value: Any, start: int = ..., stop: int = ..., /) -> int
```

All parameters are position-only, matching `list.index()` signature. This allows `list[str]` to properly satisfy the `SequenceNotStr[str]` protocol used by DataFrame's `columns` and `index` parameters.

The stubs include:
- `pandas/_typing.pyi` - Corrected `SequenceNotStr` protocol and type aliases
- `pandas/__init__.pyi` - DataFrame, Series exports and common functions
- `pandas/core/frame.pyi` - DataFrame class definition
- `pandas/core/series.pyi` - Series class definition
- `METADATA.toml` - Package metadata
  - `pandas/core/__init__.pyi` - Core module stub
  Total: 6 stub files (82 lines) + test (140 lines)

## Test Plan

1. Added regression tests in `pyrefly/lib/test/pandas/dataframe.rs`:
```bash
cargo test test_dataframe
```
Result: passes

references :
  - https://github.com/pandas-dev/pandas/issues/56995
  - https://github.com/pandas-dev/pandas/blob/main/pandas/_typing.py

### DIFF
--- a/pyrefly/lib/solver/subset.rs
+++ b/pyrefly/lib/solver/subset.rs
@@ -120,20 +120,6 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                     l_arg = l_args.next();
                     u_arg = u_args.next();
                 }
-                // EDGE CASE: Allow PosOnly parameters to match Pos parameters in protocols
-                // This handles cases like list.index() (which has position-only params) matching
-                // SequenceNotStr.index() from pandas 2.x typeshed stubs (which incorrectly
-                // lacks position-only markers, fixed in pandas 3.0).
-                // From a typing perspective, this is sound: if a protocol allows a parameter
-                // to be passed by position or keyword (Pos), an implementation that only allows
-                // positional (PosOnly) is more restrictive, which is acceptable.
-                (Some(Param::PosOnly(_, l, l_req)), Some(Param::Pos(_, u, u_req)))
-                    if (*u_req == Required::Required || matches!(l_req, Required::Optional(_))) =>
-                {
-                    self.is_subset_eq(u, l)?;
-                    l_arg = l_args.next();
-                    u_arg = u_args.next();
-                }
                 (Some(Param::Pos(l_name, l, l_req)), Some(Param::Pos(u_name, u, u_req)))
                     if *u_req == Required::Required || matches!(l_req, Required::Optional(_)) =>
                 {
@@ -424,6 +410,17 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
         if !self.recursive_assumptions.insert(recursive_check) {
             // Assume recursive checks are true
             return Ok(());
+        }
+        // HACK: pandas 2.x typeshed stubs have a bug where SequenceNotStr.index() has incorrect
+        // parameter kinds (not all position-only), causing list[str] to fail the protocol check.
+        // This is fixed in pandas 3.0 stubs. Until then, we hard-code that list/tuple satisfy
+        // SequenceNotStr. See https://github.com/pandas-dev/pandas/issues/56995
+        if protocol.has_qname("pandas._typing", "SequenceNotStr") {
+            if let Type::ClassType(got_cls) = &got {
+                if got_cls.is_builtin("list") || got_cls.is_builtin("tuple") {
+                    return Ok(());
+                }
+            }
         }
         let protocol_members = self
             .type_order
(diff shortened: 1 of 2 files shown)


stroxler:
Thanks for the PR, and for looking into Pandas to diagnose our issues!

Requesting review from @rchen152 because I think she has the best context on stub bundling - as a rule of thumb I don't think we would want to override released stubs, but if the typeshed stubs are badly broken for a library this important we might need to consider it.

We also might be able to hack Pyrefly internals a bit if necessary as an alternative - in theory I think an edge case in `is_subset_eq` might actually be easier to maintain than stub overrides.

jackulau:
Hi @stroxler,

Thanks for the suggestion! I added an edge case in `is_subset_eq` that allows position-only parameters (`PosOnly`) to match regular positional parameters (`Pos`) in protocol checking.

Added `test_dataframe_with_broken_stubs` which uses the actual broken pandas 2.x stubs and verifies `list[str]` correctly matches `SequenceNotStr[Any]`.

## Testing

```bash
cargo test test_dataframe
```
Result: passes

pyrefly/lib/solver/subset.rs
```diff
@@ -120,6 +120,20 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                     l_arg = l_args.next();
                     u_arg = u_args.next();
                 }
+                // EDGE CASE: Allow PosOnly parameters to match Pos parameters in protocols
+                // This handles cases like list.index() (which has position-only params) matching
+                // SequenceNotStr.index() from pandas 2.x typeshed stubs (which incorrectly
+                // lacks position-only markers, fixed in pandas 3.0).
+                // From a typing perspective, this is sound: if a protocol allows a parameter
```
rchen152: This seems backwards to me. The implementation should be *less* restrictive, so that it can be used everywhere the protocol can be used.

stroxler: @Jack-GitHub12 what I had in mind if we wanted a hard-coding based fix is that we could just say `SequenceNotStr <: list[str]`, so that we never even get to the level of protocol callable checks where this branch gets hit

If we do it that way, the hack is specific to Pandas `SequenceNotStr` instead of affecting arbitrary protocol checks

grievejia (approved):

PR #1666, opened 2025-11-22
fix: quote marks during autocomplete now take care of the existing quotes 

### LINKED ISSUE #1086: Literal autocomplete should take care of existing quote marks

Karman-singh15:
# Summary

Fix code completion to avoid inserting extra quotes when cursor is already inside a string literal.

When typing `foo('` and triggering completion for a `Literal["apple", "pear"]` parameter, the completion previously inserted `'apple'` (with quotes), resulting in invalid syntax `foo(''apple'')`. Now it correctly inserts just `apple`, producing valid code `foo('apple')`.

Fixes completion behavior for `typing.Literal` string values when the user has already typed an opening quote.

Fixes #1086 

# Test Plan

## Automated Testing
- Added new test case [completion_literal_quote_test] in [pyrefly/lib/test/lsp/completion_quotes.rs]
- Test verifies that completions insert unquoted text when cursor is inside a string
- Run with: `cargo test -p pyrefly --lib test::lsp::completion_quotes::completion_literal_quote_test`

## Manual Testing
1. Create a Python file with:
   ```python
   from typing import Literal
   def foo(fruit: Literal["apple", "pear"]) -> None: ...
   foo('

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2653,19 +2653,19 @@ impl<'a> Transaction<'a> {
                 let label = lit.to_string_escaped(true);
                 let insert_text = if in_string_literal {
                     if let Lit::Str(s) = lit {
-                        Some(s.to_string())
+                        s.to_string()
                     } else {
-                        None
+                        label.clone()
                     }
                 } else {
-                    None
+                    label.clone()
                 };
                 completions.push(CompletionItem {
                     // TODO: Pass the flag correctly for whether literal string is single quoted or double quoted
                     label,
                     kind: Some(CompletionItemKind::VALUE),
                     detail: Some(format!("{param_type}")),
-                    insert_text,
+                    insert_text: Some(insert_text),
                     ..Default::default()
                 });
             }
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -1322,8 +1322,8 @@ foo(
 5 | foo(
         ^
 Completion Results:
-- (Value) 'bar': Literal['bar']
-- (Value) 'foo': Literal['foo']
+- (Value) 'bar': Literal['bar'] inserting `'bar'`
+- (Value) 'foo': Literal['foo'] inserting `'foo'`
 - (Variable) x=: Literal['bar', 'foo']
 "#
         .trim(),
@@ -1381,7 +1381,6 @@ Completion Results:
     );
 }
 
-// todo(kylei): provide editttext to remove the quotes
 #[test]
 fn completion_literal_do_not_duplicate_quotes() {
     let code = r#"


pyrefly/lib/error/expectation.rs
```diff
@@ -57,3 +62,9 @@ impl Expectation {
         }
     }
 }
+
+fn normalize_message(s: &str) -> String {
+    s.replace("\\'", "'") // unescape single quotes
+        .replace("\"", "\"") // keep double-quote escapes as-is
```
kinto0: is it worth including this line? let's keep the comment but remove the no-op

pyrefly/lib/test/lsp/completion_quotes.rs
```diff
@@ -0,0 +1,104 @@
...
+                    }
+                }
+            }
+        }
+        report
+    }
+}
+
+#[test]
+fn completion_literal_quote_test() {
```
kinto0: instead of making this new test file, is there a way to include this test in the normal lsp/completion.rs test?

for example, `completion_literal_do_not_duplicate_quotes` already exists. I notice that test result didn't change. is there a way to update that test instead?

Karman-singh15:
@kinto0 , i will work on your suggestions but can you please tell what are the merge conflicts and how do i go about fixing that. it would be great help if you could guide me through that

pyrefly/lib/test/lsp/completion_quotes.rs
Karman-singh15: should i write a new test in the completion.rs or i can just rewrite the completion_literal_do_not_duplicate_quotes test?

kinto0: if rewriting completion_literal_do_not_duplicate_quotes is easier, go for it!

kinto0:
> * pyrefly/lib/state/lsp.rs

not sure exactly what the conflicts are but it says there are conflicts in pyrefly/lib/state/lsp.rs. if you pull + merge, it should tell you. let me know if you need more help!

pyrefly/lib/error/expectation.rs
Karman-singh15: sure i will remove the line 68 but leave the comment there?
is there a problem with the code tho?

kinto0: no issue with the code. but replacing something with itself (a no-op) isn't worth doing here imo

Karman-singh15: yeah that makes sense ofc. i will look into this

pyrefly/lib/test/lsp/completion_quotes.rs
Karman-singh15: okay great i will work on this rn

Karman-singh15:
hi i think i made some mistake while trying to resolve the conflict, but it seems fine to me now. sorry for the mess of commits i made, still new to this.
can you please check once
sorry for the inconvenience

kinto0:
looks great now! thank you for working on this

pyrefly/lib/state/lsp.rs
```diff
@@ -2370,20 +2375,34 @@ impl<'a> Transaction<'a> {
         }
```
kinto0, line 2641: hmm actually it looks like this doesn't build. I think you missed a call here

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -1263,30 +1264,22 @@ Completion Results:
...
-- (Value) 'foo': Literal['foo']
-- (Value) 1: Literal[1]
-"#
-        .trim(),
-        report.trim(),
-    );
+    
+    // We expect the completion to NOT insert extra quotes if we are already in a quote.
+    // Currently it likely inserts quotes.
+    println!("{}", report);
```
kinto0: nit: is there a reason we can't keep the test how it used to be?

pyrefly/lib/test/lsp/mod.rs
```diff
@@ -9,6 +9,7 @@
 
 mod code_actions;
 mod completion;
+
```
kinto0: nit: remove

pyrefly/lib/test/lsp/completion.rs
Karman-singh15: no actually this test was written for the specific issue occurring in the sandbox link i think we can use the previous one. should i change it?

pyrefly/lib/state/lsp.rs
Karman-singh15: yeah i had missed a call i added that

Karman-singh15:
only changing the test is left you can tell me if you want the previous test, i will add that tonight rest changes i have made you can take a look

pyrefly/lib/test/lsp/completion.rs
kinto0: if the previous one describes the same problem (I think it does) we should keep it that way to make this PR smaller / easier to review. 

but if you want to use the exact example from the sandbox (perfectly good too), I would suggest writing it similar to the existing assert_eq(the entire result.trim(), report.trim()). this entire result makes it much easier to read at a glance and it's already formatted to exclude useless information

Karman-singh15:
hi, sorry for the delay i have reverted the test to the previous test. you can look and tell me if there is any issues

Karman-singh15:
@kinto0 is there any update?

Karman-singh15:
@kinto0 is there any update

kinto0:
> @kinto0 is there any update

sorry about the delay. we've been taking some time off with the holidays and I haven't got a chance to get through my backlog. if you fix the merge conflicts, shoot me a message on discord and i'll make sure to review it again in time

Karman-singh15:
@kinto0 i am sorry i accidentally closed this pr, i am still figuring out git github. i request you to reopen this 
sorry for the inconvenience

yangdanny97:
We can't reopen the PR for you, I think only you can

Karman-singh15:
@kinto0 i have redid the issue and test file please check and sorry for the problems before still new to this. thanks for your cooperation.

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -1398,7 +1399,7 @@ foo(''
 5 | foo(''
          ^
 Completion Results:
-- (Value) 'foo': Literal['foo']
+- (Value) 'foo': Literal['foo'] inserting `foo`
```
kinto0, line 1401: please remove the todo on line 1384 of this test

pyrefly/lib/state/lsp.rs
```diff
@@ -2482,24 +2484,39 @@ impl<'a> Transaction<'a> {
...
+                    if let Lit::Str(s) = lit {
+                        Some(s.to_string())
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
                 completions.push(CompletionItem {
                     // TODO: Pass the flag correctly for whether literal string is single quoted or double quoted
```
kinto0, line 2664: do we still need this?

pyrefly/lib/state/lsp.rs
```diff
@@ -2482,24 +2484,39 @@ impl<'a> Transaction<'a> {
...
+        in_string_literal: bool,
+    ) {
         match param_type {
             Type::Literal(lit) => {
+                let label = lit.to_string_escaped(true);
+                let insert_text = if in_string_literal {
+                    if let Lit::Str(s) = lit {
+                        Some(s.to_string())
+                    } else {
+                        None
```
kinto0: is the intuition here that insert_text of `None` means we will use the label as the insert text? I think it might be easier just to always use insert_text for clarity. i.e. never make insert text `None`

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -1295,8 +1296,8 @@ foo('
 4 | foo('
          ^
 Completion Results:
-- (Value) 'bar': Literal['bar']
-- (Value) 'foo': Literal['foo']
+- (Value) 'bar': Literal['bar'] inserting `bar`
```
kinto0, line 1299: it seems like this case is identifier as in a string literal?

is it possible to add a test for `foo(` (no quotes at all) where we see `inserting `"bar"`` so it's clear that the lack of quotes mean it's not inserting quotes?

Karman-singh15: yes so this test exists under completion_literal_union_alias() i added the inserting "bar"

pyrefly/lib/state/lsp.rs
Karman-singh15: i made it explicit

pyrefly/lib/test/lsp/completion.rs
Karman-singh15: done

pyrefly/lib/state/lsp.rs
Karman-singh15: sorry, i didnt get you

Karman-singh15:
@kinto0 i made the necessary changes please review them. sorry for the delay

pyrefly/lib/state/lsp.rs
kinto0: i meant the comment. I originally thought your PR accomplished this so it could be deleted. but it still seems like a bug. I will move the comment up to the relevant place (where the `to_string_escaped` call is)

kinto0 (approved):
awesome!! thanks for doing this

Karman-singh15:
hi, @kinto0 this shows that the pr was merged with no commits. why does this happen

yangdanny97:
@Karman-singh15

pyrefly's source of truth is actually a mercurial repo, when we merge a PR from github it involves importing it to mercurial, merging, and syncing the change back to github.

I guess somewhere in the process the commit hash gets overwritten because the syncing system needs to edit the commit message, which makes Github's UI show the PR as "closed" instead of "merged". But if you look in the commit history for Pyrefly it should show you as the author of those commits, just under a different hash.

PR #1671, opened 2025-11-23
feat: allow redefinition of variables

### LINKED ISSUE #816: Allow redefinition of variables
### Describe the Bug

```python
def f(x: int):
   x: str = str(x) # E: ERROR 2:4-5: `x` cannot be annotated with `str`, it is already defined with type `int`
```

Would be nice if we can redefine variables afresh. Or alternatively, give it a separate error could (redefinition) and then people could just ignore that one error to make it work.

### Sandbox Link

https://pyrefly.org/sandbox/?code=CYUwZgBGAUAeBcECWA7ALgSngKAnhEAzmgE4QC8RpcG2QA

_No response_

gvozdvmozgu:
Fixes #816

### DIFF
--- a/crates/pyrefly_config/src/args.rs
+++ b/crates/pyrefly_config/src/args.rs
@@ -141,14 +141,6 @@ pub struct ConfigOverrideArgs {
         num_args = 0..=1
     )]
     infer_with_first_use: Option<bool>,
-    /// Allow re-annotating a variable with a new, conflicting type without emitting a redefinition error.
-    #[arg(
-        long,
-        default_missing_value = "true",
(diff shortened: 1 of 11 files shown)


stroxler (changes requested):
Thanks for the PR @gvozdvmozgu!

I think this can actually be simplified quite a lot though. Our usual mechanism for optional checks is to make a new error kind - which this diff does - and then encourage people who want less-strict checks to just opt out of that error code.

As a result, the `allow-redefinitions` flag is probably unnecessary, we can just merge the new error code and get most of the benefits

Issue #1589, opened 2025-11-15
parity with pylance/basedpyright


asukaminato0721:
### Describe the Bug

Related links.

[https://github.com/DetachHead/basedpyright/issues?q=is:issue+is:open+pylance+label:"pylance+parity"
](https://github.com/DetachHead/basedpyright/issues?q=is%3Aissue%20is%3Aopen%20label%3A%22pylance%20parity%22)

https://docs.basedpyright.com/dev/benefits-over-pyright/pylance-features/

asukaminato0721:
missing:

Convert import to relative path / absolute path (and Convert all... variants)

Add pytest fixture type annotation (and Add all... variants)

asukaminato0721:
#claim

Issue #1577, opened 2025-11-13
Go to type definition by cmd + clicking on a type hint.


jvansch1:
### Describe the Bug

Currently you cannot command + click on a type hint in Pyrefly and have it go to the type definition. We should enable the ability to cmd + click on a type hint and to be taken to the type definition.

yangdanny97:
Wait why? I thought cmd+click is gotodef, and would allow the user to jump to the definition of the inferred type, without needing to write it into the file?

if I remember correctly, that's what Rust analyzer does

jvansch1:
@yangdanny97 I misunderstood the initial issue. Updated the issue to reflect what we are actually trying to do.

jvansch1:
Completed with commit e0bf26e4be6f3cc13c9382d7d3dfaec214d68b4f

PR #1677, opened 2025-11-24
impl "go to definition" on inlay hints / location

asukaminato0721:
# Summary

impl https://github.com/DetachHead/basedpyright/issues/651

part of #1589 

Introduced InlayHintLabelSegments/InlayHintLabelSegment to capture each rendered fragment plus its optional TextRangeWithModule, and taught Transaction::inlay_hints to emit these richer labels instead of raw strings so we can preserve click targets coming from the type formatter. The non-type hints still use the helper from_text, so downstream code only needs to call .text() when a plain string is required.

Updated the LSP server to convert those segments into InlayHintLabelParts (falling back to strings when no locations exist) so
 VS Code and others now support clickable inlay hints; text edits still use concatenated text, so formatting behavior remains unchanged.

# Test Plan

Added an explicit regression that asserts label parts carry the document URI/range for MyClass, proving that go-to-definition on inlay hints now works as requested in basedpyright#651.

Existing notebook and CLI tests read the assembled text, so no user-facing breakage.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -2640,7 +2640,7 @@ impl Server {
         )?;
         let res = t
             .into_iter()
-            .filter_map(|(text_size, label_parts)| {
+            .filter_map(|(text_size, label)| {
                 // If the url is a notebook cell, filter out inlay hints for other cells
                 if info.to_cell_for_lsp(text_size) != maybe_cell_idx {
                     return None;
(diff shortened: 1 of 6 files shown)


jvansch1:
Hey @asukaminato0721 we appreciate you looking into this! This was being tracked with https://github.com/facebook/pyrefly/issues/1577 which was actually just closed. 

However, there is some follow up work that needs to be done on top of what was implemented if you are interested in looking at either of these issues:
- https://github.com/facebook/pyrefly/issues/1680
- https://github.com/facebook/pyrefly/issues/1681

PR #1683, opened 2025-11-25
fix Tuple Types in type hints are currently not clickable #1680

### LINKED ISSUE #1680: Tuple Types in type hints are currently not clickable
### Describe the Bug

Clickable type hints were introduced with https://github.com/facebook/pyrefly/issues/1577. 

Currently if you have a type like `tuple[int, str]`, the `int` and `str` portion will be clickable, but the `tuple` will not be. The tuple arm at https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_types/src/display.rs#L548 probably needs some special handling to make this work properly. 

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1680

TypeDisplayContext now tracks an optional tuple QName, exposes new_with_tuple, and Type gained get_types_with_locations_with_stdlib, so callers with a stdlib can request clickable tuple bases while legacy callers stay untouched.

The formatting stack gained TypeOutput::write_tuple_keyword, Tuple::fmt_with_type now routes through it, and OutputWithLocations injects the tuple location when available, giving us localized spans without changing the rendered text.

The inlay-hint provider fetches the stdlib once per request and uses the new helper so tuple keywords inside hints are now cmd+clickable.

# Test Plan

cargo test -p pyrefly_types type_output::tests::test_output_with_locations_tuple_base_clickable_when_qname_available 

cargo test -p pyrefly_types display::tests::test_get_types_with_location_tuple.

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/inlay_hint.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/inlay_hint.rs
@@ -6,8 +6,8 @@
  */
 
 use lsp_types::request::InlayHintRequest;
-use serde_json::json;
 use serde_json::Value;
+use serde_json::json;
 
 use crate::test::lsp::lsp_interaction::object_model::ClientRequestHandle;
 use crate::test::lsp::lsp_interaction::object_model::InitializeSettings;
@@ -92,8 +92,7 @@ fn test_inlay_hint_default_config() {
                 }]
             }
         ]),
-    )
-        .unwrap();
+    );
 
     interaction.shutdown().unwrap();
 }
@@ -223,8 +222,7 @@ fn test_inlay_hint_disable_variables() {
                 "range":{"end":{"character":15,"line":14},"start":{"character":15,"line":14}}
             }]
         }]),
-    )
-        .unwrap();
+    );
 
     interaction.shutdown().unwrap();
 }
@@ -275,8 +273,7 @@ fn test_inlay_hint_disable_returns() {
                 "range":{"end":{"character":6,"line":11},"start":{"character":6,"line":11}}
             }]
         }]),
-    )
-        .unwrap();
+    );
 
     interaction.shutdown().unwrap();
 }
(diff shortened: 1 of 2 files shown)


crates/pyrefly_types/src/display.rs
```diff
@@ -91,11 +92,19 @@ pub struct TypeDisplayContext<'a> {
     /// Should we display for IDE Hover? This makes type names more readable but less precise.
     hover: bool,
     always_display_module_name: bool,
+    tuple_qname: Option<&'a QName>,
```
jvansch1: Could we just add the tuple qname to the existing qnames map instead? I think if we did that we would not have to handle the tuple qname in a special way and can rely on existing logic.

crates/pyrefly_types/src/display.rs
```diff
@@ -742,7 +755,21 @@ impl Type {
...
+    }
+
+    pub fn get_types_with_locations_with_stdlib(
+        &self,
+        stdlib: &Stdlib,
+    ) -> Vec<(String, Option<TextRangeWithModule>)> {
+        self.get_types_with_locations_with_tuple_qname(Some(stdlib.tuple_object().qname()))
+    }
+
+    fn get_types_with_locations_with_tuple_qname(
```
jvansch1: Do we need to have both this function as well as get_type_locations_with_stdlib? Currently there is only one place where we call get_types_with_locations so it should be fairly easy to update that to make sure we pass in stdlib when we call it.

My concern with adding all of this now is that we are adding complexity that is not really warranted yet. Ideally we always want to have stdlib when we are using OutputWithLocations, so it doesn't really make sense to have one function which does not have stdlib(get_types_with_locations) and one that does(get_types_with_locations_with_stdlib). Until we see the need for both of those I think we should try to keep the code simpler.

crates/pyrefly_types/src/tuple.rs
```diff
@@ -64,7 +64,7 @@ impl Tuple {
         output: &mut O,
         write_type: &impl Fn(&Type, &mut O) -> fmt::Result,
     ) -> fmt::Result {
-        output.write_str("tuple")?;
+        output.write_tuple_keyword()?;
```
jvansch1: It would be great it we could avoid having these special helper function for each type. If we could find a way to instead just have this be write_qname instead of write_tuple_keyword and make the logic more generic I think that would be great.

asukaminato0721:
https://github.com/facebook/pyrefly/commit/39fba165db74de948c07d2282b8168a96e6e6f07

PR #1684, opened 2025-11-25
fix Hovering on function docstrings shows unnecessary Literal["..."] #1665

### LINKED ISSUE #1665: Hovering on function docstrings shows unnecessary `Literal["..."]`
### Describe the Bug

When hovering on a docstring in a function or a class, the dialog for type info and ... (i don't know what it's called) opens and show the `Literal["..."]` and for long docstrings it is very annoying.

I don't experience this behavior in other lsp's that i used before (basedpyright).

pyrefly rpc messages:  
```json
// Send:
{"jsonrpc":"2.0","id":222,"method":"textDocument/hover","params":{"textDocument":{"uri":"file:///home/sina/test.py"},"position":{"line":44,"character":26}}}

// Receive:
{"jsonrpc":"2.0","id":222,"result":{"contents":{"kind":"markdown","value":"```python\nLiteral['No need to show this text']\n```"}}}
```

basedpyright on the same thing:  
```json
// Send:
{"jsonrpc":"2.0","id":207,"method":"textDocument/hover","params":{"textDocument":{"uri":"file:///home/sina/test.py"},"position":{"line":44,"character":26}}}

// Receive:
{"jsonrpc":"2.0","id":207,"result":null}
```

I don't know if this is intended behavior or a bug.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn0%2BjA0pYvVGKkxsMCACpFcrqAFoMpujC4blYSwCmt0BYWNg0gqsKJOPAONsahEOLtar5qL7dCB-f6AHLcdAwCYHaiuGXlZ34QLaVChOx8XyUtJqnnRdiYdFwDgOMKBgMBn2%2BkTA3xgc2W622tJu%2BnoJLLVYbLZ2aBQDTSzAWRjsdSbPrUNLSiXRMFOKBQZTmKzZm7UQQiMQSNxwJIY4gBCPVid8NktIq%2BbbJ6KqGwLFwFput9rtvrbJz57YAFgATLsmyA5CADbOJDkIgVBzF0UCOBIYAcqcZLoJuzaOmU%2B4WhwVo2naQhuogHr7HUAB8UYFt6vrlseQhVsGGFpDhwD4AAvoG6C-iAZCruIhBcCwUAUDQK6iCKlSyPgmToJAWhFAUcHxAYUjhtQPAcBwxBwIgAD0alsQJpCELCWhqTA6BqbmjBwGpzBiRAElCFJsRqWaEqoEsSZQKg2A5BZ4mSV0PK4MQPnwTJZDcugdRUnAcHUAAvNQ-oAMyEAAjB%2BTEgPRf5TF0KwAGLQDAFBCdUQFpUAA

I'm using zed 0.213.3 with pyrefly 0.42.1

asukaminato0721:
# Summary

Fixes #1665

Added position_in_docstring plus an early-return guard so the hover pipeline now recognizes module/class/function docstrings and skips building Literal[...] tooltips when the cursor is inside those ranges, resolving the behavior reported in issue #1665

# Test Plan

Introduced a regression test that hovers over a docstring literal and asserts that the server responds with None, ensuring the scenario stays fixed.

cargo test -p pyrefly hover_returns_none_for_docstring_literals -- --nocapture

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -111,9 +111,8 @@ fn position_in_docstring(
     handle: &Handle,
     position: TextSize,
 ) -> bool {
-    let ast = match transaction.get_ast(handle) {
-        Some(ast) => ast,
-        None => return false,
+    let Some(ast) = transaction.get_ast(handle) else {
+        return false;
     };
     fn body_contains_docstring(body: &[Stmt], position: TextSize) -> bool {
         if let Some(range) = Docstring::range_from_stmts(body)


pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -105,6 +106,42 @@ fn format_suppressed_errors_hover(errors: Vec<Error>) -> Hover {
     }
 }
 
+fn position_in_docstring(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    position: TextSize,
+) -> bool {
+    let ast = match transaction.get_ast(handle) {
```
yangdanny97: nit:
```
let Some(ast) = transaction.get_ast(handle) else {
  return false
}
```

yangdanny97 (approved):
LGTM

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -105,6 +106,41 @@ fn format_suppressed_errors_hover(errors: Vec<Error>) -> Hover {
     }
 }
 
+fn position_in_docstring(
```
jvansch1, line 109: To me the name `position_in_docstring` sounds like it should be returning the position of something within the docstring. Instead this returns a boolean which indicates to me that it is actually determining something if part of the docstring and maybe should be called `positioned_in_docstring` or `is_positioned_in_docstring`. 

I think either way it would be good to add a docstring for this function to help it easier to understand what is going on here.

yangdanny97: I'll patch this after import

stroxler (approved):

Issue #1685, opened 2025-11-25
Auto import suggestions for `collections.abc` should take precedence over `typing`


DanielNoord:
### Describe the Bug

When running `pyrefly` as LSP I'd expect imports that are available in `collections.abc` to take precedence of those coming from `typing`.
Considering they are deprecated, see for example https://docs.python.org/3/library/typing.html#typing.Iterable

For example in `def func(x: Iterable[str]) -> None: ...` the import `from collections.abc import Iterable` should come before `from typing import Iterable`.
However, `pyrefly` doesn't find it, or suggests to import it from `_collections_abc`.

I believe the logic for sorting these actions is in:
https://github.com/facebook/pyrefly/blob/66dfdbfd6a75a0d70ab51b658f220e7ee5456c60/pyrefly/lib/state/lsp.rs#L1676

That should probably learn about a certain precedence of certain modules over others.
However, at the moment `pyrefly` seems unable to even recognize `collections.abc` as a shim for `_collections_abc`.

I found the following code which could explain this:
https://github.com/facebook/pyrefly/blob/66dfdbfd6a75a0d70ab51b658f220e7ee5456c60/pyrefly/lib/state/lsp.rs#L2822-L2828

_I'd love to contribute to this project as I'd like to learn a bit more Rust and consider myself to have more than adequate knowledge of Python, but this seems like it might be too involved for a first contribution. If somebody is willing to coach me, I'd be up for trying that though!_

### Sandbox Link

I don't think the Sandbox supports generating samples of LSP output

stroxler:
Thanks for the bug report! I think you're right, we shold prefer the `collections` variants.

We ideally would also downrank anything marked as `@deprecated`, but that's a separate implementation I think, and probably not as important as the core collection types

DanielNoord:
@stroxler Do you think this could serve as a good first issue? Or is it too complex?

yangdanny97:
As steven mentioned, this could be split on 2 parts:
1. downranking deprecated imports
2. fixing `_collections_abc` to be `collections.abc`

I'm not sure how to do the second one (cc @kinto0 on the implementation details) but the first item should be easy enough to be a good first issue, if you'd like to tackle it.

DanielNoord:
I have tried to implement the check for `deprecated` imports in #1694.

I went with removing `deprecated` exports from the selection. There are some questions about this behaviour and the code in question that we can discuss there.

[...]

PR #1694, opened 2025-11-26
flag `deprecated` exports in autoimport and move to bottom

DanielNoord:
# Summary

References https://github.com/facebook/pyrefly/issues/1685

This implements one of the improvements discussed in the issue: taking into account deprecations when checking exports for autoimport.

I have added two tests. One tests whether we suggest imports from the stdlib. It doesn't seem we test this yet and I was figuring out how the tests worked and came up with this. I can remove it, but since I wrote it I thought I might as well keep it? It also shows an improvement we should make (`typing` over `ast`).

The second test covers the change I make in the PR. It is pretty simple really.

# Test Plan

See above. Added two tests.

Also tried to run the CI on my local fork in https://github.com/DanielNoord/pyrefly/pull/1 but it seems the setup is quite different from the checks I see running on other PRs in this repository.

I did. There are quite a lot of warnings about unused lint ignores and unused exports, but my changes seem okay? Let me know if I missed anything.


DanielNoord:
For the reviewer, I also found:
https://github.com/facebook/pyrefly/blob/c0ebd8ce89543649ebe757181b22902e45dd0f86/pyrefly/lib/state/lsp.rs#L2226-L2230

There is a difference between completions and code actions, I guess? I'm not quite sure what the difference between them is, but I think the code on `main` already shows there is a potential for drift. In completions deprecation is taken into account while for code actions it isn't.
I have now excluded deprecated exports from code actions (causing another type of drift with completions), mainly because `search_exports_exact` doesn't really provide a nice API to "downrank" handles and just either returns them or not.

Do we need to address this drift? And if so, how?
And does this PR actually address the error I'm trying to fix? Or does the code action not really matter in practice?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -236,3 +236,66 @@ my_export
         report.trim()
     );
 }
+
+#[test]
+fn test_import_from_stdlib() {
```
kinto0, line 324: this test is great, i'd like to keep it 

the one concern would be that different python / typeshed versions change where `TypeVar` comes from, but since we bundled typeshed it should match on all systems. an integration test of this behavior (where we actually use system python) would be less ideal than what you did.

pyrefly/lib/state/lsp.rs
```diff
@@ -2887,9 +2887,15 @@ impl<'a> Transaction<'a> {
...
-            if let Some(export) = exports.get(&Name::new(name)) {
-                match export {
-                    ExportLocation::ThisModule(_) => vec![handle.dupe()],
+            if let Some(export_location) = exports.get(&Name::new(name)) {
+                match export_location {
+                    ExportLocation::ThisModule(export) => {
+                        // Ignore deprecated exports in autoimport suggestions
+                        if export.deprecation.is_some() {
+                            return Vec::new();
+                        }
```
kinto0: I wonder if we can add metadata making it clear that these are deprecated instead of ignoring these (lots of people might still want to use them). 

- completion items are the most likely to be accidentally imported, but they [have the tag called `deprecated`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem) that formats it as strikethrough. maybe the next step in improving this is deranking them with sortText?
- [quick fix code actions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction) are more of an opt-in action, so I'm not too concerned about people accidentally importing deprecated functionality. it might make sense to instead format it with a strike-through or write `(deprecated)` next to it. what do you think?

kinto0 (changes requested):
thanks for the change! I like the small change with lots of tests. 

> Do we need to address this drift? And if so, how?

I'm not especially concerned about the drift right now. from what I see, completion autoimports use a fuzzy search and quick fixes don't. that makes sense to me: I don't want a quick fix changing the identifier - I only want it adding imports. 

if I were to work on this, I'd likely change search_exports_exact's API to return some notion of `deprecated` as well and use that metadata to format the quick fixes with some type of warning similar to imports (along with downranking them if you want)

but then the bigger question of "how do we sort these correctly" is still unanswered.... I'll comment on the original task for that

pyrefly/lib/state/lsp.rs
DanielNoord: Do we know if IDEs take the `deprecated` into consideration when sorting completions? To me it feels more of a concern of the IDE (or whatever is taking in the LSP output) than of the LSP itself. We already provide all "necessary" data for the sorting algorithm to decide (by indicating it as being deprecated).

For quick fix I think adding `(deprecated)` makes the most sense?

However, perhaps that should be tackled in a follow up (related to the linked issue). It seems we want more things to be taken into consideration when sorting, which might influence the type of "metadata"/context we show for the action?

DanielNoord:
> but then the bigger question of "how do we sort these correctly" is still unanswered.... I'll comment on the original task for that

This is also what I struggle with. I'm scared of overscoping this PR but am not sure what you're preferred course of action is. If this already provides value in your opinion I would merge it, but can also understand if you first want to answer the bigger question :)

pyrefly/lib/state/lsp.rs
kinto0: > Do we know if IDEs take the deprecated into consideration when sorting completions? To me it feels more of a concern of the IDE (or whatever is taking in the LSP output) than of the LSP itself. We already provide all "necessary" data for the sorting algorithm to decide (by indicating it as being deprecated).

Since we provide `sortText`, I doubt they'd push these results to the bottom out of `sortText` order. It seems like language servers should decide this. 

> For quick fix I think adding (deprecated) makes the most sense?

agreed

kinto0:
> > but then the bigger question of "how do we sort these correctly" is still unanswered.... I'll comment on the original task for that
> 
> This is also what I struggle with. I'm scared of overscoping this PR but am not sure what you're preferred course of action is. If this already provides value in your opinion I would merge it, but can also understand if you first want to answer the bigger question :)

I think this provides value if it adds the (deprecated) field. But I think removing quick fixes might not be what some users want. Let me look into how pyright sorts / filters these completion items....

pyrefly/lib/state/lsp.rs
DanielNoord: @kinto0 Could you provide some pointers on the datastructure to return and where to define it?

I'm not too experienced with larger `rust` codebases like these. I thought of maybe combining `Handle` and `ExportLocation` in the a `struct? But perhaps that is not efficient enough.
Very open to any kind of feedback/tips :)

kinto0: If I were you I'd scope this as follows:
- this PR: add `deprecated` warning on autoimport
- another PR: sort `deprecated` to the bottom

in order to do these, I might start by changing `search_exports_exact` to return a Vec<(Handle, bool (representing is_deprecated)>. then you can reuse is_deprecated upstream in sorting / text.

pyrefly/lib/state/lsp.rs
```diff
@@ -1686,7 +1689,12 @@ impl<'a> Transaction<'a> {
                                 import_format,
                             );
                             let range = TextRange::at(position, TextSize::new(0));
-                            let title = format!("Insert import: `{}`", insert_text.trim());
+                            let mut title = format!("Insert import: `{}`", insert_text.trim());
+
+                            if export.deprecation.is_some() {
```
DanielNoord: This can probably be done more efficiently 😅

pyrefly/lib/state/lsp.rs
```diff
@@ -1709,7 +1717,16 @@ impl<'a> Transaction<'a> {
...
-        code_actions.sort_by(|(title1, _, _, _), (title2, _, _, _)| title1.cmp(title2));
+
+        // Sort code actions: non-deprecated first, then alphabetically
+        code_actions.sort_by(|(title1, _, _, _), (title2, _, _, _)| {
+            match (title1.contains("deprecated"), title2.contains("deprecated")) {
+                (true, false) => Ordering::Greater,
+                (false, true) => Ordering::Less,
+                _ => title1.cmp(title2),
+            }
+        });
```
DanielNoord, line 1747: This also feels very non-idiomatic...

DanielNoord:
@kinto0 I have changed the test and code. It now shows we prefer the import from `b.py` over `a.py` due to deprecation and that deprecation is being shown in the action as well. Let me know what you think!

pyrefly/lib/state/lsp.rs
kinto0: it might easiest just to throw it all in the original format with a `{:?}` and `export.deprecation.map(|_| " (deprecated")`. that should only insert it if is_some and avoid any `mut`

pyrefly/lib/state/lsp.rs
kinto0: if we sort within this function, we should rename it to `local_quickfix_code_actions_sorted` so callers know not to re-sort (I think most other `state` functions return unsorted items). keeping it in here is probably the simplest thing to do. 

maybe instead of this sort, we add a sort_key to code_actions. when we notice deprecated and add it, we put a sort-key that will put it at the bottom. then in this function, we can just sort by key once at the end. this will let us add other things to sort by in the future. we don't need to return the sort_key, so we can map it out after/during sorting.

pyrefly/lib/state/lsp.rs
DanielNoord: TIL, thanks!

That did produce an unwanted `None` in cases where `deprecation` was `None`. I fixed that by using `export.deprecation.map_or("", |_| " (deprecated)")` and `{}` instead. Hope that also works for you!

pyrefly/lib/state/lsp.rs
DanielNoord: I did the rename!

I'm not quite sure how to implement the `sort_key`. I grepped the code base but could only find examples in Python code. Do you have an example of an idiomatic `rust` implementation of such a property/attribute/whatever?

kinto0: in the code_actions.sort_by call, we would do something like this:

```
code_actions.sort_by_key(|(_, _, _, _, sort_key)| sort_key)

// remove sort key
code_actions.map(|(a, b, c, d, _)| (a, b, c, d));
```

this means all the logic of sorting can be confined to when we add the code action:
```
    if (export.deprecation.is_some()) {
        let title = format!(
            "Insert import: `{}`",
            insert_text.trim(),
        );
        code_actions.push((title, module_info.dupe(), range, insert_text, ""));
    } else {
        let title = format!(
            "Insert import: `{}` (deprecated)",
            insert_text.trim(),
        );
        code_actions.push((title, module_info.dupe(), range, insert_text, "zzzz"));
    }
```
(sorry this removes the other suggestion of insert import deduplication unless you can figure out a clean way to keep it)

DanielNoord: Hmm, I'm not sure how to do this cleanly. Because you'll also still want to sort by `title` itself so you'd get `"zzzz - {}", title` as the `sort_key`. It also doesn't feel very extensible to other sorting considerations?

If you think it is better to use the `sort_key` I can change it of course, but I'd probably keep as is for now so and come up with a better datastructure for all of our sorting needs in the rest of the linked issue.

DanielNoord:
@kinto0 Is there anything left to do here? If possible I'd like to avoid future merge conflicts :smile:

kinto0:
> @kinto0 Is there anything left to do here? If possible I'd like to avoid future merge conflicts 😄

sorry!!! not sure how I missed this. I swear I imported it when I renamed it. oh well, thanks for fixing the conflicts.

PR #1695, opened 2025-11-26
[easy] Fix runtime checkable protocol unsafe overlap with inheritance test case

fangyi-zhou:
# Summary
This diff does two minor fixes.
(1)
The test `test_runtime_checkable_unsafe_overlap_with_inheritance` was not a bug. `@runtime_checkable` requires a direct inheritance of `Protocol`, so the error message is valid.
Otherwise, a runtime error would be thrown.
This diff adds the direct inheritance and the expected error for unsafe overlap is emitted.

(2)
When there are multiple unsafe overlaps exists for a runtime checkable protocol, multiple errors were emitted.
This treatment was suboptimal, so this diff groups the errors into a single error, and explains all the overlapping fields in a multi-line error message.

# Test Plan
`test.py`


stroxler (commented):
LGTM, I'll try to get a second review

PR #1712, opened 2025-11-29
Fix: LSP import completions with trailing spaces and builtin filtering (#1704)

### LINKED ISSUE #1704: Wrong import suggestions
### Describe the Bug

When importing from another module, many objects that are not actually present in that module are suggested, making it quite hard to find what is really relevant and actually exists there.

This can be replicated in the initial Sandbox by simply trying to get intellisense to suggest an import, where the only suggestion should be `format_number`.

Image

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn0%2BjA0pYvVGKkxsMCACpFcrqAFoMpujC4dQFVhRJx4BxtjUIhxdrVfNQXSJgb4wLCWAU1ugLCxsGlbfT0EllqsNls7NAoBppZgLIx2OpNn1qGlpRLomCnFAoMpzFZouxUwJhKIRW44EkMcQAhNcrDPmyWkVfNtQuVrKgFi4OEIQ%2BH2pG%2BtsnHAbQAWABMuxDIDkIAN%2BZI5EQVDmXSgjgkYA5pzJ6GroYtZU9Qm9HF9-sDQltiHt%2BzqAD5qBOhE6Xa7W0IPboQAAcjeaQPsA%2BAAL7-voC4gGQIhiKQhBcCwUAUDQgjwSKlSyPgmToJAWhFAUh7xAYUgwOwPAcBwxBwIgAD09FwRWiGwlo9EwOg9GYOicD0cw%2BEQIRQjEbE9GNhKqBLKg0CoNgOQCQRRFdDyuDECpR6kWQ3LoHUVJwIe1AALxmiAADMhAAIzTlB6AgOBi5TF0KwAGLQDAFDYdUq72UAA

_No response_

vinismarques:
# Summary

This PR fixes #1704 by filtering out implicitly re-exported builtins from import suggestions and improves the robustness of completions when trailing whitespace is present.

**Specific changes:**
-   **Filter Builtin Exports**: Filters out exports that originate from `builtins` or `extra_builtins` (e.g., `str`, `Exception`) when providing completion candidates for a module. This prevents polluting the completion list with global builtins that aren't explicitly exported by the target module. Standard module attributes (like `__annotations__`, `__doc__`) remain available.
-   **Trailing Whitespace Support**: Adds logic to handle `import` and `from ... import` statements ending in trailing spaces. Previously, the parser yielded no nodes for these cursor positions, preventing completions; now, we detect the context and provide appropriate suggestions (e.g., showing the module list after `import  <cursor>`).
-   **Refactoring**: Extracted completion logic into `add_module_exports_completions` and `add_all_modules_completions` to reduce code duplication and simplify the main `completion` function.
-   **Docs**: Fixed a minor typo in `CONTRIBUTING.md`.

Fixes #1704

# Test Plan

Ran `python3 test.py` to verify the following scenarios:

-   **Builtin Filtering**: Verified that importing from a module does not suggest builtin types like `str` or `int`, but still suggests valid exports and module attributes (verified via `test_completion_from_import_builtins`).
-   **Trailing Spaces**: Verified that completions appear correctly when the cursor is separated from the `import` keyword by spaces (verified via `test_completion_import_trailing_space` and `test_completion_from_import_trailing_space`).
-   **Conformance**: All existing conformance tests passed.


vinismarques:
Hey @stroxler, I tried tackling the issue opened last week. If you can, this is ready for review.

connernilsen:
Thanks for doing this @vinismarques. We just got back from holiday, so I'll add Steven to review the diff

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -691,6 +704,21 @@ from foo import func
 2 | from foo import func
                ^
 Completion Results:
+- (Variable) deprecated
+- (Variable) func_ok
+- (Variable) __annotations__
```
yangdanny97: is this change intended? I don't think it was part of the original ask

vinismarques: Good catch, I reverted these tests. I had misunderstood them, but upon reviewing, it became clear that these completions should not be returned.

pyrefly/lib/state/lsp.rs
```diff
@@ -2861,24 +2896,100 @@ impl<'a> Transaction<'a> {
...
                 if let Some(mod_module) = self.get_ast(handle) {
                     let nodes = Ast::locate_node(&mod_module, position);
+
+                    // If no node covers the current position, try to find a preceding import statement
+                    let mut processed_as_import_statement = false;
                     if nodes.is_empty() {
-                        self.add_keyword_completions(handle, &mut result);
-                        self.add_local_variable_completions(handle, None, position, &mut result);
-                        self.add_builtins_autoimport_completions(handle, None, &mut result);
+                        for stmt in mod_module.body.iter().rev() {
```
yangdanny97, line 2903: The logic in lsp.rs is very difficult for me to follow. I would prefer if we could avoid doing string slicing to figure out whether we're in an import or not.

Based on poking around with the ruff AST, https://play.ruff.rs/c109ef63-0112-4318-b7d0-031fce65d738 it seems like trailing space imports generate a StmtImport or StmtImportFrom with `names=[]`. Would it be possible to check that instead?

yangdanny97 (changes requested):
Hi, thanks for making this PR!

The test cases look good, but I think there is an opportunity to simplify the added logic in lsp.rs. Let me know what you think

pyrefly/lib/state/lsp.rs
yangdanny97: Also, it's a bit odd we iterate thru the statements in reverse when it's more likely to find import statements at the top

pyrefly/lib/state/lsp.rs
```diff
@@ -2652,6 +2652,60 @@ impl<'a> Transaction<'a> {
...
+                    } else {
+                        None
+                    },
+                    ..Default::default()
+                })
+            }
+        }
+    }
+
+    fn add_all_modules_completions(&self, handle: &Handle, completions: &mut Vec<CompletionItem>) {
```
yangdanny97, line 2690: I don't understand why it was added, the logic didn't exist before, could you please explain?

pyrefly/lib/state/lsp.rs
yangdanny97: I would prefer the AST logic to be extracted into a helper like `locate_immediately_preceding_import` in the same module as `Ast::locate_node`

PR #1726, opened 2025-12-01
Fix #1565: Handle LoopRecursive in is_subset_eq

### LINKED ISSUE #1565: `int` is not assignable to `int` (caused by inconsistent types when breaking cycles)
Type error assigning `int` to `int` in while loop. [repro](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABFQMa5UAUciNcALgE4CUiAHXQ0RNCDQC8NAAxDRNAO4ALaDDE0APDVjpW-OfNEArSWIOHFK2DWbmLouAG0jAXRoBCKQJDe78-cL28iYA1FIAjH6i4lKhNJHoIAA0IACunNBwJOSIIADENACqGVAQnKQ0YKno9Bm46HBCQlS0YLjcALaonAD66Kkd2DDczPjsEOicvDQAtAB8HDyCgSLcMJyp3MJg3gByA0Pc7MD4AL6%2BiSlka2BQpIScuB1QFAUACqQ3dxwYOAQ0jHQkAA5ptuhB6oQhAUAMowNRKTicYhsAD0qOu1DuhHawNRMHQqMwuHocFRgJBYLqhMq7RoqAAbqhoKhsNYKRBQdxwfUaLhiNSskIyJwlPUZgzhnAIcIvCAAMyEcIAJguIFOKVQtQgkoAYqoKGgsHgiGR1UA)

Repro contents:
```
def decode(s: str):
    i = 0
    while i < len(s):
        j = i
        while (
            s[j] != ""
        ):
            j += 1
        i = j + 1
```

AryanBagade:
# Summary

This PR fixes a false positive type error in nested loops with cyclic variable dependencies.

Previously, when a `LoopRecursive` variable appeared as the `want` parameter in `is_subset_eq`, the type checker would throw an internal error, causing false positives like "`int` is not assignable to `int`" in nested while loops.

The fix handles the `Variable::LoopRecursive` case by extracting the prior type and continuing the subset check with the underlying type, instead of erroring.

Fixes #1565

# Test Plan

- ✅ Test with exact code from issue #1565 now passes

### DIFF
--- a/pyrefly/lib/test/flow_looping.rs
+++ b/pyrefly/lib/test/flow_looping.rs
@@ -685,3 +685,20 @@ for i in range(10):
     a = d.get('x', a)
     "#,
 );
+
+// Test for https://github.com/facebook/pyrefly/issues/1565
+testcase!(
+    bug = "False positive: int is not assignable to int with nested while loops",
+    test_nested_while_int_assignment,
+    r#"
+def decode(s: str):
+    i = 0
+    while i < len(s):
+        j = i
+        while (
+            s[j] != ""
+        ):
+            j += 1
+        i = j + 1
+    "#,
+);
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -1324,9 +1324,17 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                 let variable1 = variables.get(*v1);
                 let variable2 = variables.get(*v2);
                 match (&*variable1, &*variable2) {
-                    (_, Variable::LoopRecursive(..)) => Err(SubsetError::internal_error(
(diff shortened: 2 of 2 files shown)


pyrefly/lib/test/flow_looping.rs
```diff
@@ -685,3 +685,20 @@ for i in range(10):
     a = d.get('x', a)
     "#,
 );
+
+// Test for https://github.com/facebook/pyrefly/issues/1565
+testcase!(
+    bug = "False positive: int is not assignable to int with nested while loops",
```
stroxler, line 691: We generally only use `bug = "..."` to track bugs that aren't fixed; in this case it looks like we fixed the issue. That's just for future reference - I'll patch it up since I have to import the PR anyway in order to merge

yangdanny97 (approved):

Issue #1655, opened 2025-11-21
pyrefly lsp cannot find imports from virtual environment


bmarroquin:
### Describe the Bug

When i run `pyrefly check` there are no errors.
```
$ poetry run pyrefly check
 INFO Checking project configured at `C:\Users\...\pyproject.toml`
 INFO 0 errors
```
LSP seems not be checking the Poetry create venv.

```
Could not find import of `advanced_alchemy.extensions.litestar`
Looked in these locations (from config in `C:\Users\...\pyproject.toml`):
Import root (inferred from project layout): "C:\\Users\\..\\src"
Site package path queried from interpreter: ["C:\\Python313\\DLLs", "C:\\Python313\\Lib", "C:\\Python313", "C:\\Users\\<user>\\AppData\\Roaming\\Python\\Python313\\site-packages", "C:\\Python313\\Lib\\site-packages"] Pyrefly (missing-import)
```

### (Only applicable for extension issues) IDE Information

Running on pycharm and the issue appears in the following configurations.
* Latest stable release using LSP4IJ plugin
* Running Pycharm EAP with the integrated settings.

In both situations i am sure that pyrefly is being called from the venv, it is the only placed installed on the system.

stroxler:
What does your project layout look like? I believe we have heuristics on what a virtual environment folder is likely to be named, we may not be finding the virtualenv if it's laid out differently.

In most cases you can work around gaps in our heuristics by specifying an interpreter path in pyrefly.toml

bmarroquin:
also impacting vs code plugin. 

quick recreate:
1. prep project
```
# set up pyrefly
poetry new test-pyrefly-lsp
cd test-pyrefly-lsp
poetry add --dev pyrefly
poetry run pyrefly init

# add to site-pacakges
poetry add sqlalchemy

```
2. open code in project.
3. open any file
4. import sqlalchemy
5. error!
```
Could not find import of `sqlalchemy`
  Looked in these locations (from config in `c:\test-pyrefly-lsp\pyproject.toml`):
  Import root (inferred from project layout): "c:\\test-pyrefly-lsp\\src"
[...]

PR #1731, opened 2025-12-02
track modules that can't be resolved

kinto0:
Summary:
- on invalidate find, we only set a file to dirty if our dependencies changed
- we currently only trace `deps`, the dependencies that can be resolved

we also need to track deps that can't be resolved. this bug manifests as type errors that remain longer than they should

fixes https://github.com/facebook/pyrefly/issues/1655
fixes https://github.com/facebook/pyrefly/issues/1632

Differential Revision: D87952758


samwgoldman (approved):

PR #1735, opened 2025-12-02
Use LazyBlocking indexing mode to fix test failure

rchen152:
Summary: Maybe?

Differential Revision: D88134869


grievejia (approved):

PR #1743, opened 2025-12-03
Handle await inside async comprehensions

### LINKED ISSUE #1732: False positive Syntax Error: `await` can only be used inside an async function
### Describe the Bug

This error is basically wrong: 

> `await` can only be _**used**_ inside an async function

The reality is that await can't be **_executed_** outside async functions, which means that inside a definition of an async generator (that is later executed inside a async function) is perfectly ok:

For instance this code works perfectly with no syntax error:

```python
import asyncio

async def test2():
    return "test"

async def a_gen():
    for i in range(3):
        yield test2()

a = (await cor async for cor in a_gen()) # [Invalid-syntax] `await` can only be used inside an async function

async def main():
    async for i in a:
        print(i)

if __name__ == "__main__":
    asyncio.run(main())
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIEAtsbgE4AuABKnKegMYS4A66frdhyaYYYJg3gMATAAoAlIj5MVTOjAYBXOuiY8QkuA318BbTiLEsA%2BgHMY6BUt2qw9JhA%2B66Ge7IDMisqqIaQQMFCYElJy8qboqEwAvEyyqADuqBDMHO6CFm50TLlFELqodg4Kcfy1%2BcKi4lRZjkEuKvVMhR5eLM4hIcR0ZQyyEDV8EOLW1glUMDPJKfozzWUz%2Bv2q9dyEdJqOa63yIAA0IJoM0HAk5IggAMRMAKpXUNmkXQccV7jocPFGl16M0GLNNFRsDA6LJ8IgvAx5EwALQAPiYRjoWxU6i0Oi6%2BgAchCoVimMB8ABfEzoM4gMjqMBQUiEBi4KhQChPAAKpEZzIxGBwBGKf0gtm0qF%2B6EIfCeAGUYDAmAALBgMYhwRAAem1DLEzMI9Fs2oc2swuA4cG1uXQ4sl0u1wKKqAAblkoKhsLBRXaIBKfNKmLhiNKbnwyAwVX9ka7oXBuLpliB-IQAIzSGkgSnnVA-CBxgBi0BgFDQWDwRDI2aAA

_No response_

KaranPradhan266:
# Summary

Previously, Pyrefly treated async comprehensions as non-async contexts, so await inside an async comprehension at module scope was wrongly flagged (“await can only be used inside an async function”). This PR marks comprehension scopes as async when appropriate and updates the await check to honor them, fixing the false error.

Fixes #1732

# Test Plan

Added a regression test covering the a = (await cor async for cor in a_gen()) case, with exact code from issue #1732

### DIFF
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -766,9 +766,7 @@ impl<'a> BindingsBuilder<'a> {
                 self.ensure_expr(&mut x.value, usage);
                 let in_async_def = self.scopes.is_in_async_def();
                 let in_generator_element = self.in_generator_await_context();
-                if !in_async_def
-                    && !in_generator_element
-                    && !self.module_info.path().is_notebook()
+                if !in_async_def && !in_generator_element && !self.module_info.path().is_notebook()
                 {
                     self.error(
                         x.range(),


yangdanny97:
I don't think this is quite right, we should be handling generators differently from other comprehensions, whether the comprehension is async or not shouldn't matter.

Can you please add some additional test cases to verify the behavior is correct? I borrowed these from https://github.com/python/mypy/pull/18152/files#r1841480778

```
from typing import Any
def test(xs: Any):
    (await x for x in xs)  # OK
    [await x for x in xs]  # E:
    {await x for x in xs}  # E:
    {0: await x for x in xs}  # E:
    {await x: 0 for x in xs}  # E:

    (x async for x in xs)  # OK
    [x async for x in xs]  # E:
    {x async for x in xs}  # E:
    {x: 0 async for x in xs}  # E:

    (x for x in await xs)  # E:
    [x for x in await xs]  # E:
    {x for x in await xs}  # E:
    {x: 0 for x in await xs}  # E:
 ```

**Comment by KaranPradhan266** [comment]
Hi @yangdanny97  Thanks for pointing that out! My previous change didn’t correctly handle those cases. I’ve updated the logic so that only generator expressions open an async-capable scope. 

Async list/set/dict comprehensions do not create that scope and therefore must appear inside an enclosing async def. If they occur in a synchronous context, I’ll emit an “asynchronous comprehension outside of an async function” error. 

This keeps module/sync contexts correctly rejecting async list/set/dict comprehensions, while still allowing async generator expressions anywhere and letting await work inside them as intended.

Let me know if this approach looks good — I’ll also cover additional edge cases beyond the ones you mentioned.
image
(After implementing above change)

**Comment by KaranPradhan266** [comment]
In a case like:

def testCase6(async_iterable):
    gen = (
        [x * 2 async for x in inner_iterable]
        for inner_iterable in async_iterable
    )
    return gen


What behavior should we expect here?

**Comment by yangdanny97** [comment]
@KaranPradhan266 

Something like "`async` not allowed outside of async function", like Pyright does

https://pyright-play.net/?code=CYUwZgBALiDOUGECGsQDYAUKCeA7AxgPoCWMATkgEYA2IAlAFwBQErEA5iLhALwQYs2AbQAeEAFQQATBBwEIYAPZkIY4t3W4QZEuSq0AuoNZKVm7bu36QEdbNh4ipKzRDGIdd2RBQArmW5OXCA

File: pyrefly/lib/binding/expr.rs [hunk 1]
```diff
@@ -374,14 +374,24 @@ impl<'a> BindingsBuilder<'a> {
...
             self.ensure_expr(&mut comp.iter, usage);
             if i == 0 {
-                self.scopes.push(Scope::comprehension(range));
+                // Async list/set/dict comprehensions must be inside an async def. Async generator
+                // expressions are allowed to stand alone because they produce an async generator.
+                if comp.is_async && !is_generator && !self.scopes.is_in_async_def() {
+                    self.error(
+                        range,
+                        ErrorInfo::Kind(ErrorKind::InvalidSyntax),
+                        "asynchronous comprehension outside of an async function".to_owned(),
```
yangdanny97: nit: "`async` outside of an async function" would be more concise

KaranPradhan266:
@yangdanny97 I reworked how await legality is tracked so that await is permitted in generator-expression elements but still rejected in iterables, filters, and other non-async positions. Regression tests are included for the updated generator-await semantics and expected diagnostics.

I also removed the old generator-aware comprehension-scope flag and the longer “asynchronous comprehension outside of an async function” error. Instead, await validity is now determined entirely through explicit await-context tracking: generator elements get an async-capable context, while iterables, filters, and list/set/dict elements do not. The new tests exercise these updated semantics and verify the revised error wording.
image
image

PR #1744, opened 2025-12-03
Fix #1699: Install python-dateutil in sandbox using micropip

### LINKED ISSUE #1699: Sandbox has source not found for dateutil
### Describe the Bug

If you put this code in the sandbox:

```python
from dateutil.relativedelta import relativedelta
print(relativedelta(days=1))
```

Then you get the error:

```
ERROR sandbox.py:1:1-49: Stubs for `dateutil.relativedelta` are bundled with Pyrefly but the source files for the package are not found. [[missing-source-for-stubs](https://pyrefly.org/en/docs/error-kinds/#missing-source-for-stubs)]
```

Does this make sense in the sandbox? If you then run it you get a failure:

```
The module 'python-dateutil' is included in the Pyodide distribution, but it is not installed.
```

Should we install it into Pyodide?

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEmqALjAK5PSGUxTMQBuMTDyao6EGsVyUmdbrw6DhUUQB1064pQjomACnl8lI1Hsak4AXgCMASlsgANCHbQ4JcohABiOgFUOKAgmUjowVnQAYw5cdDh1dWEwMOkaZgB9dFYabBhKPXxEcV1bOgBaAD46OCZKRHU6RrkYJlZKdDDVEAA5bNy6umB8AF8u9ScQMm4wKFJCJlooCl8ABVJp2eqMHAI6SNjIAHM2vljCdV8AZRgYOgALJiZiOEQAelepmBm56UPXmHQr0wuEicFe%2B3QRxOMUBKUodFQ-FQ0FQ2FgewOEGOlFOHVwxBh7k0pCYd1iZUElDgEFidEsdC6AGZCNYAExjdAgYbOVDRAQwABi0BgFDQWDwRDIXKAA

_No response_

AryanBagade:
# Summary
Load python-dateutil when initializing Pyodide using micropip.install()This PR enables users to import and use `python dateutil` in the Pyrefly sandbox by automatically installing the package during Pyodide initialization.

**Changes:**
- Modified `pythonWorker.ts` to load micropip and install python-dateutil during Pyodide init
- Added error handling to prevent initialization failures if package loading fails

Fixes #1699 

# Test Plan

**Local Testing:**
- ⚠️ Unable to fully verify locally due to SRI (Subresource Integrity) check failures on Pyodide packages in dev environment

**Verification needed:**
- Maintainers should verify this works in production/CI where SRI checks pass correctly

### DIFF
--- a/website/src/sandbox/pythonWorker.ts
+++ b/website/src/sandbox/pythonWorker.ts
@@ -82,6 +82,18 @@ const initPyodide = async () => {
         },
     });
 
+    // Load python-dateutil package for sandbox use
+    // See: https://github.com/facebook/pyrefly/issues/1699
+    try {
+        await py.loadPackage('micropip');
+        await py.runPythonAsync(`
+import micropip
+await micropip.install('python-dateutil')
+        `);
+    } catch (error) {
+        console.error('[Pyodide] Failed to install python-dateutil:', error);
+    }
+
     return py;
 };
 


yangdanny97:
Questions here:
1. does the snippet that does the instlalation run at page load, or only when you hit "Run"?
2. if the latter, is there a way to detect that the code uses that package and only install it when necessary? One package is probably fine but if we end up having a whole list of packages installed each time the page loads it could get fairly slow and expensive.

yangdanny97:
re: testing, @migeed-z might know

AryanBagade:
> 1. does the snippet that does the instlalation run at page load, or only when you hit "Run"?

The installation runs on **first "Run" click**, not at page load. Pyodide initializes lazily when the worker receives user's first code execution, so the package is installed then and cached for all subsequent runs.

yangdanny97:
Upon further consideration I don't think we should add various 3p dependencies to the sandbox, that way we don't slow down execution.

PR #1747, opened 2025-12-03
fix panic: range start index 1 out of range for slice of length 0 #1745

### LINKED ISSUE #1745: panic: range start index 1 out of range for slice of length 0
### Describe the Bug

```python
def f[*Ts](*x: *Ts):
    x[1:0]
```

```
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/alt/subscript.rs:176:31:
range start index 1 out of range for slice of length 0
```

### Version info
pyrefly 0.44.0

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1745

Added a guard for positive-bound tuple slices so that when start >= stop we immediately return tuple[()] instead of slicing the prefix

# Test Plan

Introduced test_type_var_tuple_slice_empty to lock in the expectation that slicing a variadic tuple with reversed bounds always produces an empty tuple.

### DIFF
--- a/pyrefly/lib/alt/subscript.rs
+++ b/pyrefly/lib/alt/subscript.rs
@@ -115,14 +115,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .parse_slice_literal(upper_expr)
             .ok()?
             .unwrap_or(elts.len() as i64);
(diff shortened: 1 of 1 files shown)


yangdanny97:
Thanks!

PR #1749, opened 2025-12-03
fix Support property.deleter - Known attribute labeled as unknown #1711

### LINKED ISSUE #1711: Known attribute labeled as unknown
### Describe the Bug

The hover text on the `full_url` attribute is `(attribute) full_url: Unknown`
```python
import urllib.request
urllib.request.Request('http://example.com', method='GET').full_url
``` 

Image

This seem to happen on property attribute with setter and deleter.

Minimal reproducible example:
```python
class A:
    @property
    def prop(self) -> int:
        return 42
    @prop.setter
    def prop(self, value: int) -> None:
        pass
    @prop.deleter
    def prop(self) -> None:
        pass

A().prop
```

Image

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn0%2BjA0pYvVGKkxsMCACp9N0YXDqEUoFAINhCCIAI4WeAcfSoVAAXiNJrNlutcA4hAASjArTbtgByHgcDjERAAenD91QmNghAxAYUbG5mAdAcafwAKgHdoQwBZjWsjfplqsNlttvbdiA5CAAtA4CRyIgqHMuibNrkOacyeg4AqsKJcrCWAU1ugLCxsGltjUIhx9nUAHzUD1CWq%2BahbkTA3xgXQgAByk%2Bn6%2BowHwAF8D-oayAyCIxKRCFwWFAKDRBI%2BRZVZPhMugkBaEUBS9vEBhSDA7DBqGcARuGD6iOIhCwloUboOGmDonA4bMIBEDAUIoGxOGw4SqgSyoNAqDYDkeFASBXQ8rgxBMX24FkNy6B1FScC9tQDrUAeADMhAAIwAEw3ugICXrWUxdCsABi0AwBQv7VE2slAA

_No response_

asukaminato0721:
# Summary

Fixes #1711

Updated property attribute lookup so that both getter and setter functions are treated as property bases, and .deleter access now records a PropertyPayload that carries the getter plus any known setter.

Added PropertyPayload plumbing through FuncFlags and SpecialDecorator::PropertyDeleter, allowing solve_function_binding to substitute the previously inferred setter (or getter) type whenever a property deleter is defined. This keeps the class attribute type stable even after `@foo.deleter` runs.

# Test Plan

Reproduced the original failure with a new regression test that exercises getter + setter + deleter, ensuring reads, writes, deletes, and reveal_type all behave as expected

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -352,9 +352,21 @@ impl Deprecation {
 #[derive(
     Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Visit, VisitMut, TypeEq
 )]
-pub struct PropertyPayload {
+pub enum PropertyRole {
+    Getter,
+    Setter,
+    SetterDecorator,
+    DeleterDecorator,
+}
+
+#[derive(
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Visit, VisitMut, TypeEq
+)]
+pub struct PropertyMetadata {
+    pub role: PropertyRole,
     pub getter: Type,
     pub setter: Option<Type>,
+    pub has_deleter: bool,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
@@ -365,19 +377,10 @@ pub struct FuncFlags {
     pub is_classmethod: bool,
     /// A function decorated with `@deprecated`
     pub deprecation: Option<Deprecation>,
-    /// A function decorated with `@property`
-    pub is_property_getter: bool,
+    /// Metadata for `@property`, `@foo.setter`, and `@foo.deleter`.
+    pub property_metadata: Option<PropertyMetadata>,
     /// A function decorated with `functools.cached_property` or equivalent.
     pub is_cached_property: bool,
-    /// A `foo.setter` function, where `foo` is some `@property`-decorated function.
-    /// When used to decorate a function, turns the decorated function into a property setter.
-    pub is_property_setter_decorator: bool,
-    /// If None, this is a function decorated with `@foo.setter`, where `foo` is
-    /// a property (i.e. a function decoratoed with `@property`)
-    ///
-    /// The stored type is `foo` (the getter).
-    pub is_property_setter_with_getter: Option<Type>,
-    pub property_deleter_payload: Option<PropertyPayload>,
     pub has_enum_member_decoration: bool,
     pub is_override: bool,
     pub has_final_decoration: bool,
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -47,7 +47,8 @@ use crate::solver::solver::SubsetError;
 use crate::types::callable::FuncMetadata;
 use crate::types::callable::Function;
 use crate::types::callable::FunctionKind;
-use crate::types::callable::PropertyPayload;
+use crate::types::callable::PropertyMetadata;
+use crate::types::callable::PropertyRole;
 use crate::types::class::Class;
 use crate::types::class::ClassType;
 use crate::types::literal::Lit;
@@ -414,6 +415,7 @@ struct AttributeBase(Vec1<AttributeBase1>);
 struct PropertyAttr {
     getter: Type,
     setter: Option<Type>,
+    deleter: bool,
 }
 
 /// A single, "atomic" attribute base, not coming from a union or an intersection.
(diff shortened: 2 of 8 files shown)


pyrefly/lib/alt/attr.rs
```diff
@@ -1323,14 +1330,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
                         getter, base,
                     )
+                } else if attr_name == "deleter" {
+                    let mut getter = property.getter.clone();
+                    let payload = PropertyPayload {
+                        getter: property.getter.clone(),
+                        setter: property.setter.clone(),
+                    };
+                    getter.transform_toplevel_func_metadata(|meta: &mut FuncMetadata| {
+                        meta.flags.property_deleter_payload = Some(payload.clone());
```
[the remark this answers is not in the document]
asukaminato0721: you made a mistake.

pyrefly/lib/alt/attr.rs
```diff
@@ -1323,14 +1330,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
                         getter, base,
                     )
+                } else if attr_name == "deleter" {
+                    let mut getter = property.getter.clone();
+                    let payload = PropertyPayload {
+                        getter: property.getter.clone(),
+                        setter: property.setter.clone(),
+                    };
+                    getter.transform_toplevel_func_metadata(|meta: &mut FuncMetadata| {
+                        meta.flags.property_deleter_payload = Some(payload);
```
asukaminato0721:
```suggestion
                        meta.flags.property_deleter_payload = Some(payload.clone());
```

crates/pyrefly_types/src/callable.rs
```diff
@@ -368,6 +376,7 @@ pub struct FuncFlags {
     ///
     /// The stored type is `foo` (the getter).
     pub is_property_setter_with_getter: Option<Type>,
+    pub property_deleter_payload: Option<PropertyPayload>,
```
yangdanny97: Do both fields of the payload ever get used? I would think just `Option<Type>` is enough, if we're only using one of the setter or getter types.

pyrefly/lib/alt/attr.rs
```diff
@@ -432,8 +439,8 @@ enum AttributeBase1 {
     /// before falling back to `Never`.
     TypeNever,
     /// Properties are handled via a special case so that we can understand
-    /// setter decorators.
-    Property(Type),
+    /// setter/deleter decorators.
+    Property(PropertyAttr),
```
yangdanny97, line 448: If the PropertyPayload is unnecessary then this one is too

crates/pyrefly_types/src/callable.rs
asukaminato0721: `property_deleter_payload` gets populated when the solver handles `@foo.deleter`. We wrap the getter type, along with any previously discovered setter, into a PropertyPayload and stash it in the function metadata. A simple boolean would lose the getter/setter types we need later.

stroxler: The logic in this change seems mostly correct, but the way that flags and enum values are distributed through the types seems hard to follow.

Do you think it's possible to use a single `property_metadata: Option<PropertyMetadata>` field here, and use an elaborated version of your current `PropertyPayload` for the metadata - something like
```
pub struct PropertyPayload {
    pub getter: Type,
    pub setter: Option<Type>,
    pub deleter: bool
}
```

That way, all of the property information is collected in one place, it's a lot easier to see how the types, mutations, and downstream behaviors work. It's also easier to see that the getter is always present but the setter and deleter may not be.

This also makes it more obvious that the getter is always present if we have a property, but both the setter and deleter are optional

Issue #1709, opened 2025-11-28
`type[NewType]` in a `dict` leads to strange error


DanielNoord:
### Describe the Bug

```python
from typing import NewType

Thing = NewType("Thing", int)
ThingType = type[Thing]

MAPPING: dict[int, ThingType] = {1: Thing}
```

```
ERROR sandbox.py:6:33-43: `dict[int, type[Thing]]` is not assignable to `dict[int, type[Thing]]` [[bad-assignment](https://pyrefly.org/en/docs/error-kinds/#bad-assignment)]
```

Python: 3.12

I don't think `type[NewType]` is spec'd very well as I know that `mypy` also has issues with it.
However, the error in this case is very confusing and doesn't point to what the user can do to fix this.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQCCiAOuvR-ZjGPWAK7pKACjgwoYADT18iehHQAXAJT0AtAD56AOVzoYcwkbbVaDAELDGy1u07deAoaPFT6AfTkKV6rbv2GxuggkiD8itBwJOSIIADE9ACqEVAQiqR8gpQRenBsbA58uABOALaoiu7o-KXYMMXCsvJKqpr0cIrFtpz0xTCK-MXsYCwg2jV1XfTA%2BAC%2Bo2whIGR9YFCkhIq4pVAUCQAKpKvr7Rg4BPSUepAA5oMVEHqEbAkAyjAw9AAWiorEcIgAPSAlY8daEEo3QEwdCAzC4ShwQFXdC3e45WFFYr0VAAN1Q0FQ2Fgl2uEDuxQeenouGIGKibDIii%2BejUuPqcEe7AAvPRRgBmQgARgATAtgrNQqhshB2QAxaAwChoLB4IhkECzIA

DanielNoord:
```python
from typing import NewType

Thing = NewType("Thing", int)
ThingType = type[Thing]

def func(x: ThingType) -> None: ...

func(Thing)
```

Gives:
```
ERROR sandbox.py:10:6-11: Argument `type[Thing]` is not assignable to parameter `x` with type `type[Thing]` in function `func` [[bad-argument-type](https://pyrefly.org/en/docs/error-kinds/#bad-argument-type)]
```

That is also not too helpful.

rchen152:
[...]

PR #1752, opened 2025-12-04
Reject `type[...]` with `NewType` aliases at definition time

terror:
Resolves https://github.com/facebook/pyrefly/issues/1709

This diff validates `type[...]`/`Type[...]` arguments during annotation so `typing.NewType` aliases produce an immediate invalid-annotation error instead of flowing through subtyping.

### DIFF
--- a/pyrefly/lib/test/new_type.rs
+++ b/pyrefly/lib/test/new_type.rs
@@ -145,10 +145,10 @@ Thing = NewType("Thing", int)
 ThingType = type[Thing]  # E: NewType `Thing` is not a class and cannot be used with `type` or `Type`
 OtherThingType = Type[Thing]  # E: NewType `Thing` is not a class and cannot be used with `type` or `Type`
 
-mapping: dict[int, ThingType] = {1: Thing}
+mapping: dict[int, ThingType] = {1: Thing}  # E: `dict[int, type[Thing]]` is not assignable to `dict[int, type[Unknown]]`
 
 def func(x: ThingType) -> None: ...
-func(Thing)
+func(Thing)  # E: Argument `type[Thing]` is not assignable to parameter `x` with type `type[Unknown]` in function `func`
     "#,
 );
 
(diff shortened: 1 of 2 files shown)


pyrefly/lib/test/new_type.rs
```diff
@@ -136,6 +136,21 @@ y: type[Any] = Foo  # E: `type[Foo]` is not assignable to `type[Any]`
     "#,
 );
 
+testcase!(
+    test_new_type_type_argument,
+    r#"
+from typing import NewType
+
+Thing = NewType("Thing", int)
+ThingType = type[Thing]
```
terror: I'm wondering if we should also spend effort flagging invalid `type[NewType]` usages at annotation/alias definition time? Would be willing to follow up in a separate PR.

yangdanny97:
@migeed-z can you take this one?

pyrefly/lib/test/new_type.rs
yangdanny97: Yes, we should add that check here: https://github.com/facebook/pyrefly/blob/f245fb35b4ce6013977f11b48166103c5bab5205/pyrefly/lib/alt/expr.rs#L1939

yangdanny97: I think if we have a `type[NewType]` we may want to just emit the error at the definition site, and infer the type as `Any`.

Then we wouldn't need a custom error everywhere it's being used, so the error at the definition site would supersede this PR.

terror: Yeah this makes more sense to me, I've re-purposed this PR to go down this route instead.

pyrefly/lib/test/new_type.rs
```diff
@@ -136,6 +136,22 @@ y: type[Any] = Foo  # E: `type[Foo]` is not assignable to `type[Any]`
...
+testcase!(
+    test_new_type_type_argument,
+    r#"
+from typing import NewType, Type
+
+Thing = NewType("Thing", int)
+ThingType = type[Thing]  # E: NewType `Thing` is not a class and cannot be used with `type` or `Type`
+OtherThingType = Type[Thing]  # E: NewType `Thing` is not a class and cannot be used with `type` or `Type`
+
+mapping: dict[int, ThingType] = {1: Thing}  # E: `dict[int, type[Thing]]` is not assignable to `dict[int, type[Unknown]]`
```
terror, line 148: note: I explicitly chose to keep cascading errors here. There may be a case for hiding them, but I think it makes sense to surface them 🤔

yangdanny97: Sure, if we later find that this causes a lot of noise in production we can disable it, but in absence of data it should be OK

Issue #1482, opened 2025-11-04
Ability to disable and suppress all errors but those explicitly enabled


connernilsen:
Would be awesome to have something where I could tell Pyrefly to disable all errors except what I pass in. Same with doing suppressions. This would make it easier to migrate some projects if I'm focusing on one error at a time, which works well with the way that we can enable/disable by config error.

This could be a big motivator for error profiles (all off, strict, basic/default, ...)

samwgoldman:
Can you clarify a bit more? What would you "pass in"?

connernilsen:
I was thinking that we have `--include`, `--exclude`, and `--warn`, which are explicit overrides to error configuration. So in this case, there could be a flag `--disable-all`, which when combined with `--include` and `--warn`, would allow us to turn on specific errors to look at or `--suppress`.

Even better if `--disable-all` were something like `--error-profile <all|off|strict|basic|...>`.

yangdanny97:
`--error-profile <default|all|explicit>` for now, perhaps (we'd add strict later).

`explicit` = "only emit errors explicitly set via --warn  or --error"

in terms of implementation, I imagine these would affect the "default severity" for errors, which would apply before any severity overrides from config or CLI.

Once we deicde on the naming for the flags, we should mark this as a "good first issue"

yangdanny97:
Actually I think we could reuse the `--only` flag, and make it work like error/warn/info where you can pass multiple error codes

https://github.com/facebook/pyrefly/blob/6055447c2f47fdd0e5809c8007af8d752afcf06c/pyrefly/lib/commands/check.rs#L239

rchen152:
https://github.com/facebook/pyrefly/pull/1753 does the CLI piece of this (but doesn't add an equivalent config file option).

PR #1753, opened 2025-12-04
Make `--only` filter error list as well as error summary

rchen152:
Summary:
For https://github.com/facebook/pyrefly/issues/1482.

* Extends `--only` filtering to individual shown errors as well as the error summary.
* Makes `--only` take a list (passing a single error kind still works).

This doesn't fully fix the linked issue because this is CLI-only, and we should also make it possible to set this in a config file.

Differential Revision: D88340480


yangdanny97 (approved):

PR #1769, opened 2025-12-05
Fix --remove-unused-ignores leaving blank lines

### LINKED ISSUE #1768: `--remove-unused-ignores` shouldn't leave behind extra blank lines
### Describe the Bug

If you use `--suppress-errors` to auto-add pyrefly ignores and then `--remove-unused-ignores` to remove the ignores as pyrefly improves or you fix type errors, you're left with unsightly blank lines in your code. Example:

Starting file, containing a type error:
```
def f(x: str, y, z):
    pass
f(
    0,
    1,
    2,
)
```

After running `pyrefly check --suppress-errors`:
```
def f(x: str, y, z):
    pass
f(
    # pyrefly: ignore [bad-argument-type]
    0,
    1,
    2,
)
```

After changing `x: str` to `x: int` and running `pyrefly check --remove-unused-ignores`:
```
def f(x: int, y, z):
    pass
f(

    0,
    1,
    2,
)
```

Notice the extra blank line between `f(` and `0,` that wasn't there before.

### Sandbox Link

_No response_

_No response_

AryanBagade:
# Summary
When removing unused ignore comments, standalone comment lines were leaving blank lines behind. This fix ensures that when a line becomes empty after removing the ignore comment, the entire line (including the newline) is removed.

Fixes #1768

## Changes
- Updated the logic in `remove_unused_ignores` to conditionally write newlines

# Test Plan
  - ✅ All 20 tests in `error::suppress::tests` pass
  - ✅ Manually tested the exact workflow from the issue:

### DIFF
--- a/pyrefly/lib/error/suppress.rs
+++ b/pyrefly/lib/error/suppress.rs
@@ -372,8 +372,9 @@ pub fn remove_unused_ignores(loads: &Errors, all: bool) -> usize {
                         let new_string = regex.replace_all(line, "");
                         if !new_string.trim().is_empty() {
                             buf.push_str(new_string.trim_end());
+                            buf.push('\n');
                         }
-                        buf.push('\n');
+                        // Skip writing newline if the line becomes empty after removing the ignore
                         continue;
                     }
                     buf.push_str(line);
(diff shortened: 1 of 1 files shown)


rchen152 (approved):
Thank you!

PR #1771, opened 2025-12-05
fix [Bug] No error on missing __all__ implementation #204

### LINKED ISSUE #204: [Bug] No error on missing `__all__` implementation

asukaminato0721:
# Summary

Fixes #204

Added a new method invalid_dunder_all_entries() that validates explicitly user-defined `__all__` entries. This method iterates through each DunderAllEntry::Name and checks whether the name exists in the module's own definitions or is available through wildcard imports (from x import *). It returns a vector of invalid entries along with their source locations.

Added a validation call in Bindings::new() after processing all statements. For each invalid entry returned by the new method, it emits a MissingModuleAttribute error

# Test Plan

updated the existing tests

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -414,7 +414,7 @@ impl Bindings {
         assert_eq!(builder.scopes.loop_depth(), 0);
 
         // Validate that all entries in __all__ are defined in the module
-        for (range, name) in exports.invalid_dunder_all_entries(lookup) {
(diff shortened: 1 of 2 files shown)


stroxler (approved):
Thanks! LGTM, I'll try to get it merged

PR #1772, opened 2025-12-05
fix Instance attribute should take precedence over non-data descriptors in attribute writes #253

### LINKED ISSUE #253: Instance attribute should take precedence over non-data descriptors in attribute writes
### Describe the Bug

Minimal repro:
```python
class NonDataDescriptor:
    def __get__(self, instance: object, owner: type | None = None) -> str:
        return "derp"

class C:
    x = NonDataDescriptor()

    def f(self):
        self.x = 1  # This is okay
```
Expected: No error
Actual: Attribute `x` of class `C` is a read-only descriptor with no `__set__` and cannot be set

### Sandbox Link

https://pyrefly.org/sandbox/?code=MYGwhgzhAEByD2A7AImALmZBTCwBOAlgA5rx4BcAUNDdACZYBm0A+iwOZZpsAUEWIRgBpoBRBAyJgWctHgAjAFZZgaEfADuiLBWhoAnkSzQAPnCTGAvOe0BKaAFoAfNAkVqtT3i4BXPImgAIgY8IkDKSlBIGABhKk8AD2hrBBR0TBx8YlI8HlsIzwZmRj4BRlt4z1p+QQA6JOsARhoAYmgAFQALAhgeuQBrMH0gA

_No response_

asukaminato0721:
# Summary

Fixes #253

now recognizes descriptor lookups without __set__, treating the assignment as a regular attribute write typed as Any so instance writes succeed without invoking the descriptor protocol while still type-checking the RHS expression.

exposes a Descriptor::has_setter() helper, letting other modules check setter availability without touching private fields.

# Test Plan

drop the expected read-only errors for assigning to non-data descriptors, aligning the fixtures with Python’s runtime behavior.

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -30,6 +30,7 @@ use crate::alt::answers::LookupAnswer;
 use crate::alt::answers_solver::AnswersSolver;
 use crate::alt::callable::CallArg;
 use crate::alt::class::class_field::ClassAttribute;
+use crate::alt::class::class_field::DescriptorBase;
 use crate::alt::expr::TypeOrExpr;
 use crate::binding::binding::ExprOrBinding;
 use crate::binding::binding::KeyExport;
@@ -906,6 +907,32 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     );
                 }
                 Attribute::ClassAttribute(class_attr) => {
+                    if let (
+                        ClassAttribute::Descriptor(
+                            descriptor,
+                            DescriptorBase::Instance(_class_type),
+                        ),
+                        AttributeBase1::ClassInstance(_),
+                    ) = (&class_attr, &found_on)
+                        && !descriptor.has_setter()
+                    {
+                        // Non-data descriptors (no __set__) are overridden by instance
+                        // assignments. Treat the assignment as writing to a plain attribute
+                        // so it type-checks but does not attempt to call a descriptor setter.
+                        self.check_set_read_write_and_infer_narrow(
+                            Type::any_explicit(),
+                            attr_name,
+                            got,
+                            range,
+                            errors,
+                            context,
+                            false,
+                            &mut narrowed_types,
+                        );
+                        should_narrow = false;
+                        continue;
+                    }
+
                     // If we are writing to an instance, we may need access to
                     // the class to special-case dataclass converters.
                     let instance_class = match &found_on {
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -189,6 +189,12 @@ pub struct Descriptor {
     setter: bool,
 }
 
+impl Descriptor {
+    pub fn has_setter(&self) -> bool {
+        self.setter
+    }
+}
+
 #[derive(Clone, Debug)]
 pub enum DescriptorBase {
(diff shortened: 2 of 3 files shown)


stroxler (changes requested):
Thanks for the PR!

I think that issue #253 deserves a bit more discussion.

First, the issue is actually slightly incorrect as written, cc @grievejia; the example in the unit tests is arguably ok but the example given in the issue is this:
```python
class NonDataDescriptor:
    def __get__(self, instance: object, owner: type | None = None) -> str:
        return "derp"

class C:
    x = NonDataDescriptor()
    def f(self):
        self.x = 1  # This is okay
```
but the assignment to `x` is *not* okay, it changes the read type from `str` to `int` which is clearly unsafe for statically typed code.

That said, I think we *could* support this pattern - for it to be type-safe, the logic when there's no setter would have to check the *get* type, and allow only assignments of valid subtypes.

That might be a good idea, and we could tweak this PR to do so

---------

That said, before we modify the logic I'd like to clarify something with @grievejia: I'm a little surprised at issue #253 to start with. It's true that the runtime allows this, but other type checkers do not (both basedpyright and mypy complain).

The issue doesn't come with much explanation, it just says "we should do this thing that diverges from other type checkers". Was there a specific codebase you were looking at when this was filed?

I don't necessarily mind allowing a type-safe operation that other type checkers don't allow, but it feels like it deserves an explanation of the motivation, and it might be that we should use a dedicated (maybe default-off) error for users who don't want this, because my understanding is that given other type checker behaviors, a missing `__set__` might be sometimes used to declare the intent that a descriptor is read-only.

grievejia:
> but the assignment to x is not okay, it changes the read type from str to int which is clearly unsafe for statically typed code.

I think one could argue that it depends on how the type of `x` is inferred -- the class definition by itself wouldn't be "unsafe" if we infer the type of `x` to be `int | str`. That said, I think it's also reasonable to infer `x` to be `int` and error on the descriptor definition. Inferring `x` to be `str` and erroring on the assignment can still be acceptable but I think is not as ideal as the other two options, given how the precedence at runtime works. 

> The issue doesn't come with much explanation, it just says "we should do this thing that diverges from other type checkers". Was there a specific codebase you were looking at when this was filed?

IIRC that series of issues were filed when past Jia was looking at [`ty`'s mdtests](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/descriptor_protocol.md#data-and-non-data-descriptors). Now that I look at it again, they put an `Any` there in the inferred type which might explain why they don't error. That part I no longer think is ideal for us. Current Jia wouldn't oppose closing the original issue 😛 .

stroxler:
Hmm. I strongly disagree that it would be okay to let `x` be `int` and not `str` - the class body definition takes precedence when present, I think all type checkers agree on this and it's pretty baked into our architecture.

As a result, if the class body says it's a descriptor then it's a descriptor, and any methods using instance shadowing need to respect the descriptor semantics. I very much do not think we should do it the other way around - in other words, in the example I definitely think we should be erroring on the assignment to `x` inside of `f`, not the assignment in the class body.

I'd be fine with closing the issue, especially if it was possibly filed mistakenly. I'm also okay with supporting this, as long as we add type safety checks to make sure the read type from the instance shadowing matches the descriptor read type.

@asukaminato0721 what do you think?

asukaminato0721:
I am also fine about :)

PR #1785, opened 2025-12-07
fix Special output format for Github actions #1599

### LINKED ISSUE #1599: Special output format for Github actions
### Describe the Bug

Detect when Pyrefly is running as a Github action, and emit errors using the Github workflow output format so that errors show up inline in PRs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands

See basedpyright for reference: 
https://docs.basedpyright.com/latest/benefits-over-pyright/improved-ci-integration/
when it find a type error, it will show comment on the pr. Not just ci error.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1599

Added emit_github_actions_annotations plus helpers to build safe ::error/::warning/::notice workflow commands (severity mapping, relative paths, escaping, Windows handling) so every CLI error can surface as a GitHub Actions annotation when GITHUB_ACTIONS=true

Hooked the annotation emission into the main reporting path so it runs after standard console/file output without affecting existing formats.

# Test Plan

Introduced unit tests covering the formatting logic, severity mapping, and escape helpers to guard against regressions.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -24,6 +24,9 @@ use anyhow::Context as _;
 use clap::Parser;
 use clap::ValueEnum;
 use dupe::Dupe as _;
+use percent_encoding::AsciiSet;
+use percent_encoding::CONTROLS;
+use percent_encoding::utf8_percent_encode;
 use pyrefly_build::handle::Handle;
 use pyrefly_config::args::ConfigOverrideArgs;
 use pyrefly_config::config::ConfigFile;
@@ -55,6 +58,7 @@ use crate::config::finder::ConfigFinder;
 use crate::error::error::Error;
 use crate::error::error::print_error_counts;
 use crate::error::legacy::LegacyErrors;
+use crate::error::legacy::severity_to_str;
 use crate::error::summarize::print_error_summary;
 use crate::error::suppress;
 use crate::module::typeshed::stdlib_search_path;
@@ -430,20 +434,26 @@ impl OutputFormat {
     }
 }
 
-fn should_emit_github_annotations() -> bool {
+fn should_emit_github_errors() -> bool {
     matches!(
         std::env::var("GITHUB_ACTIONS"),
         Ok(value) if value.eq_ignore_ascii_case("true")
     )
 }
 
+fn severity_to_github_command(severity: Severity) -> Option<&'static str> {
+    let normalized = severity_to_str(severity);
+    match normalized.as_str() {
+        "ignore" => None,
+        "warn" => Some("warning"),
+        "info" => Some("notice"),
+        "error" => Some("error"),
+        _ => None,
+    }
+}
+
 fn github_actions_command(error: &Error, relative_to: &Path) -> Option<String> {
-    let command = match error.severity() {
-        Severity::Error => "error",
-        Severity::Warn => "warning",
-        Severity::Info => "notice",
-        Severity::Ignore => return None,
-    };
+    let command = severity_to_github_command(error.severity())?;
     let range = error.display_range();
     let file = github_actions_path(error.path().as_path(), relative_to);
     let params = format!(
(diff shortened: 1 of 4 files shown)


.github/workflows/pyrefly.yml
```diff
@@ -50,5 +50,5 @@ jobs:
         JQ: jq
         TEST_PY: ${{ github.workspace }}/test.py
         PYREFLY_PY: ${{ github.workspace }}/pyrefly/python
-      run: scrut test test
```
kinto0, line 53: what's the reason for deleting this?

asukaminato0721: not deleting, add `env -u GITHUB_ACTIONS` or the output will fail because of new output.

pyrefly/lib/commands/check.rs
```diff
@@ -384,9 +387,193 @@ impl OutputFormat {
...
+    }
+
+    fn write_error_github_to_console(relative_to: &Path, errors: &[Error]) -> anyhow::Result<()> {
+        Self::buffered_write_error_github(stdout(), relative_to, errors)
+    }
+}
+
+fn should_emit_github_annotations() -> bool {
+    matches!(
+        std::env::var("GITHUB_ACTIONS"),
```
maggiemoss, line 439: Is this a requirement? Or would passing in `--output=github` be sufficient? I would prefer the CLI command, but want to understand if there's a reason for this.

asukaminato0721: Because GitHub ci will set this env var automatically. So by this, it can detect itself is in ci, and print diagnosis.

maggiemoss: Thanks for the explanation! :) Let's rename to `should_emit_github_errors`

pyrefly/lib/commands/check.rs
```diff
@@ -384,9 +387,193 @@ impl OutputFormat {
...
+        Ok(value) if value.eq_ignore_ascii_case("true")
+    )
+}
+
+fn github_actions_command(error: &Error, relative_to: &Path) -> Option<String> {
+    let command = match error.severity() {
+        Severity::Error => "error",
+        Severity::Warn => "warning",
+        Severity::Info => "notice",
+        Severity::Ignore => return None,
```
maggiemoss: We have a method 

```
fn severity_to_str(severity: Severity) -> String {
    match severity {
        Severity::Ignore => "ignore".to_owned(),
        Severity::Info => "info".to_owned(),
        Severity::Warn => "warn".to_owned(),
        Severity::Error => "error".to_owned(),
    }
}
```
in `errors/legacy.rs`, using that here could simplify this logic

pyrefly/lib/commands/check.rs
```diff
@@ -384,9 +387,193 @@ impl OutputFormat {
...
+    value
+        .replace('%', "%25")
+        .replace('\r', "%0D")
+        .replace('\n', "%0A")
+}
+
+fn escape_workflow_property(value: &str) -> String {
+    escape_workflow_data(value)
+        .replace(':', "%3A")
+        .replace(',', "%2C")
```
maggiemoss: we do have `percent-encoding` as a dependency, are there helper methods there that we could use for this?

pyrefly/lib/error/legacy.rs
```diff
@@ -14,7 +14,7 @@ use serde::Serialize;
 
 use crate::error::error::Error;
 
-fn severity_to_str(severity: Severity) -> String {
+pub(crate) fn severity_to_str(severity: Severity) -> String {
```
maggiemoss, line 17: is this a copy paste issue?

asukaminato0721: Not a copy/paste slip.

we made severity_to_str pub(crate) on purpose so the GitHub formatter in check.rs could reuse the existing severity normalization.

Issue #746, opened 2025-07-24
Extraction-related and Copy/Move refactor support


SamChou19815:
#364 lists a bunch of refactor support we want to add. This issue covers the implementation strategy for extraction related ones, which might include:

- [ ] Extract to function
- [ ] Extract to method
- [x] Extract to variable
- [x] Extract to field

All of the features are relatively easy to implement once we have a set of common infra, which isn't trivial. I will describe them below, using extract to function as an example:

- [ ] AST based diff printing: these refactors are significantly more complex than quick fixes. An AST based diff printing should generate a smallest diff between the AST before and after the refactor.
- [ ] Selection range -> statement/expression node. All these extraction refactors start from a user selection range. We need to map the range into a list of statement/expression AST nodes. Not all selection ranges make sense. e.g. If the user select only part of a statement, then maybe we shouldn't create an extraction refactor for that statement.
- [ ] Find insertion points of extracted function, and their scope. There might be multiple places to insert the extracted function.
- [ ] Generate parameters for extracted function.
  - [ ] Find undefined variables after extraction (this depends on the insertion point). The selected statements might refer to variables that are locally defined within a function. If you naively refactor the code by copy-pasta, then these variables will be undefined. These variables will be the parameters for the extracted function.
  - [ ] We also need to know the types of these variables, and we should convert the internal representation of the type into AST annotation nodes.
- [ ] Generate returns for extracted function.
  - [ ] Find escaping definition/assignments. These are the local assignments of variables within the selected range, that are also used later down the line outside of the selection range.
  - [ ] These variables need to be returned as a tuple

I implemented the [support for these features in Flow](https://github.com/facebook/flow/blob/52bebe6c8cb4f88852d225966a58f73d01302015/src/services/code_action/refactor_extract.ml), which might be helpful for Pyrefly implementation.

lolpack:
Extending this ticket to also cover the Copy/Move functionality to a new file as described in PyCharm docs: https://www.jetbrains.com/help/pycharm/move-refactorings.html

Requested here: https://forum.cursor.com/t/poor-performance-for-python-code-navigation-after-upgrade/133041/24?u=lolpack

asukaminato0721:
> AST based diff printing: these refactors are significantly more complex than quick fixes. An AST based diff printing should generate a smallest diff between the AST before and after the refactor.

This tool may help https://github.com/Wilfred/difftastic.

PR #1786, opened 2025-12-07
impl extract function, part of #746

asukaminato0721:
# Summary

part of #746

Introduced extract_function_code_actions plus LocalRefactorCodeAction plumbing so the state layer can produce workspace edits for a selected block; it now dedents the selection, synthesizes helper definitions/calls, infers parameters/returns (including augmented assignments), and rejects unsupported selections (returns/breaks/etc.).

Added a visitor-based identifier collector that walks statements/expressions to classify loads, stores, post-selection reads, and synthetic aug-assign loads, which drives the parameter/return heuristics.

Updated the non-wasm LSP server to advertise REFACTOR_EXTRACT, convert LocalRefactorCodeAction edits into URIs, and merge them alongside existing quick fixes.

# Test Plan

Added an integration test that selects a block via markers, requests refactor actions, applies the returned edits, and asserts the helper/function call match the expected structure.

### DIFF
--- a/crates/pyrefly_bundled/src/lib.rs
+++ b/crates/pyrefly_bundled/src/lib.rs
@@ -83,10 +83,10 @@ fn extract_pyi_files_from_archive(filter: PathFilter) -> anyhow::Result<SmallMap
         let mut relative_path_components = relative_path_context.components();
 
         let first_component = relative_path_components.next();
-        if let Some(expected) = filter.expected_first_component() {
-            if first_component.is_none_or(|component| component.as_os_str() != expected) {
-                continue;
-            }
+        if let Some(expected) = filter.expected_first_component()
+            && first_component.is_none_or(|component| component.as_os_str() != expected)
+        {
+            continue;
         }
         // For ThirdPartyStubs, we need to put first_component back into the path
 
--- a/pyrefly/lib/state/lsp/quick_fixes/extract_function.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/extract_function.rs
@@ -37,12 +37,8 @@ pub(crate) fn extract_function_code_actions(
     if selection.is_empty() {
         return None;
     }
-    let Some(module_info) = transaction.get_module_info(handle) else {
-        return None;
-    };
-    let Some(ast) = transaction.get_ast(handle) else {
-        return None;
-    };
+    let module_info = transaction.get_module_info(handle)?;
+    let ast = transaction.get_ast(handle)?;
     let selection_text = module_info.code_at(selection);
     if selection_text.trim().is_empty() {
         return None;
@@ -59,9 +55,7 @@ pub(crate) fn extract_function_code_actions(
     }
     let post_loads = collect_post_selection_loads(ast.as_ref(), module_stmt_range, selection.end());
     let block_indent = detect_block_indent(selection_text);
-    let Some(mut dedented_body) = dedent_block_preserving_layout(selection_text) else {
-        return None;
-    };
+    let mut dedented_body = dedent_block_preserving_layout(selection_text)?;
     if dedented_body.ends_with('\n') {
         dedented_body.pop();
         if dedented_body.ends_with('\r') {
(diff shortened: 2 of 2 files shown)


asukaminato0721:
this aim to be a basic impl, as copilot said, too many edge cases...

pyrefly/lib/state/lsp.rs
```diff
@@ -1713,6 +1725,129 @@ impl<'a> Transaction<'a> {
...
+            if seen_returns.contains(&ident.name) || !post_loads.contains(&ident.name) {
+                continue;
+            }
+            seen_returns.insert(ident.name.clone());
+            returns.push(ident.name.clone());
+        }
+        let helper_params = params.join(", ");
+        let indented_body = if dedented_body.trim().is_empty() {
+            "    pass\n".to_owned()
+        } else {
```
kinto0: what's an example where we'd end up with nothing in the body? can you add a test?

pyrefly/lib/state/lsp.rs
```diff
@@ -2974,6 +3109,203 @@ impl<'a> Transaction<'a> {
...
+        }
+        return line
+            .chars()
+            .take_while(|c| c.is_whitespace())
+            .collect::<String>();
+    }
+    String::new()
+}
+
+fn dedent_selection(selection_text: &str) -> Option<String> {
```
kinto0: any way to reuse what we do for docstrings?

pyrefly/lib/state/lsp.rs
```diff
@@ -2974,6 +3109,203 @@ impl<'a> Transaction<'a> {
...
+        dedented.push_str(&dedented_line);
+        dedented.push('\n');
+    }
+    if !selection_text.ends_with('\n') {
+        dedented.push('\n');
+    }
+    Some(dedented)
+}
+
+fn indent_block(block: &str, indent: &str) -> String {
```
kinto0: should this instead be called "prefix_lines_with"?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -236,3 +282,59 @@ my_export
...
+        mk_multi_file_state_assert_no_errors(&[("main", code)], Require::Everything);
+    let handle = handles.get("main").unwrap();
+    let transaction = state.transaction();
+    let module_info = transaction.get_module_info(handle).unwrap();
+    let selection = find_marked_range(module_info.contents());
+    let actions = transaction
+        .extract_function_code_actions(handle, selection)
+        .unwrap_or_default();
+    assert!(!actions.is_empty(), "expected extract refactor action");
+    let updated = apply_refactor_edits_for_module(&module_info, &actions[0].edits);
```
kinto0: can this stuff be part of the test boilerplate so we don't have to repeat it every test?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -236,3 +282,59 @@ my_export
...
+    return total_sum
+
+
+if __name__ == "__main__":
+    data = [1, 5, 12, 8, 15]
+    result = process_data(data)
+    print(f"The final sum is: {result}")
+"#;
+    assert_eq!(expected.trim(), updated.trim());
+}
```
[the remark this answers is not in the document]
kinto0: agreed. please add more tests, even if they are disabled

pyrefly/lib/state/lsp.rs
```diff
@@ -2974,6 +3109,203 @@ impl<'a> Transaction<'a> {
...
+    for stmt in &ast.body {
+        visit_stmt(stmt, selection, &mut found);
+        if found {
+            break;
+        }
+    }
+    found
+}
+
+fn find_enclosing_module_statement_range(
```
kinto0: @jvansch1 i think you needed something similar for call hierarchy. any way to reuse it?

pyrefly/lib/state/lsp.rs
```diff
@@ -2974,6 +3109,203 @@ impl<'a> Transaction<'a> {
     }
 }
 
+#[derive(Clone, Debug)]
+struct IdentifierRef {
+    name: String,
+    position: TextSize,
+    synthetic_load: bool,
```
kinto0: please put a docstring here for what this means

pyrefly/lib/state/lsp.rs
```diff
@@ -2974,6 +3109,203 @@ impl<'a> Transaction<'a> {
...
+}
+
+fn contains_range(outer: TextRange, inner: TextRange) -> bool {
+    outer.start() <= inner.start() && outer.end() >= inner.end()
+}
+
+fn ranges_overlap(a: TextRange, b: TextRange) -> bool {
+    a.start() < b.end() && b.start() < a.end()
+}
```
kinto0: you should be able to use TextRange::contains_inclusive and TextRange::overlap for these

pyrefly/lib/test/lsp/code_actions.rs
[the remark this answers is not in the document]
kinto0: some other ones I can think of:
- function is within a class (but no classes in range), does the indent work?

PR #1793, opened 2025-12-08
feat: support checking dataclasses.replace

### LINKED ISSUE #1789: Support checking dataclasses.replace
### Describe the Bug

Currently Pyrefly reports no errors for the following snippet

```python
from typing import *

from dataclasses import dataclass, replace

@dataclass
class Foo:
    x: int
    y: str

f = Foo(1, "a")

replace(f, x="wrong")  # expect type error
replace(f, z=3)        # expect type error
```

Mypy gives an error: https://mypy-play.net/?mypy=latest&python=3.12&gist=d0d5de9a216936fd220dd091ee7a7e3e

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRYavUyoGqAMZRUcOPE7de-KTPmK4AGjqUYxBbJgiRAAU1yFSkdqV0AYrlyIRdT3XyJO6Bh5epL5wDJTm6GB0ALzOrgAUAIz6QiCoqQCUEYbGcjDxYPr40akA7tTsmZ4AxHQw%2BMQwsvxMjXWU1OHoOSb5hXQAXtEAzBle43S19Y3NjMww7Z0guiAArgzQcCTkiCC1AKobUBBMdGCr6M0QuOhwEZgwUWC8NNIA%2BuirNNgwlPE%2BfgYYwAtAA%2BOihSjudDjQwMVaUGFgVIAOS%2BPyhdGA%2BAAvqkRMsQGRDGAoKRCAxaFAKLUAAqkElkiEYHAEOiyG6QNgI6TXdCEES1ADKMAWAAsGAxiHBEAB6WXEx5kwi8NiymDoWWYXCyOCyjmRCDcyi8m6ys68OioABuqGgqGwsHZnKNPI2NzouGI7tuAvQZAYYpuwOtvzgfJidFSw0IiQATPj0CAcSs5BtQ05oDAKGgsHgiGRk0A

_No response_

gvozdvmozgu:
Fixes #1789

### DIFF
--- a/pyrefly/lib/alt/class/dataclass.rs
+++ b/pyrefly/lib/alt/class/dataclass.rs
@@ -198,63 +198,95 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let call_keywords = keywords.map(CallKeyword::new);
         let obj_name = Name::new_static("obj");
 
-        let obj_ty = match args.first() {
-            Some(first) => self.expr_infer(first, errors),
-            None => Type::Any(AnyStyle::Implicit),
+        if args.is_empty() {
+            return self.callable_infer(
+                Callable::list(
+                    ParamList::new(vec![
+                        Param::PosOnly(
+                            Some(obj_name.clone()),
+                            Type::Any(AnyStyle::Implicit),
+                            Required::Required,
+                        ),
+                        Param::Kwargs(None, Type::Any(AnyStyle::Implicit)),
+                    ]),
+                    Type::Any(AnyStyle::Implicit),
+                ),
+                Some(&FunctionKind::DataclassReplace),
+                None,
+                None,
+                &call_args,
+                &call_keywords,
+                range,
+                errors,
+                errors,
+                None,
+                None,
+                None,
+            );
+        }
+
+        let obj_expr = &args[0];
+        let obj_ty = self.expr_infer(obj_expr, errors);
+
+        let is_dataclass = |cls: &ClassType| {
+            let cls_metadata = self.get_metadata_for_class(cls.class_object());
+            cls_metadata.dataclass_metadata().is_some_and(|m| {
+                m.field_specifiers.iter().any(|k| {
+                    matches!(
+                        k,
+                        CalleeKind::Function(FunctionKind::DataclassField)
+                            | CalleeKind::Class(ClassKind::DataclassField)
+                    )
+                })
+            })
         };
-        let mut dataclass_types = SmallSet::new();
+
         let mut saw_any = false;
+        let mut saw_dataclass = false;
         let mut saw_non_dataclass = false;
-        self.distribute_over_union(&obj_ty, |ty| {
-            match ty {
-                Type::Any(_) => saw_any = true,
-                Type::ClassType(cls) => {
-                    let cls_metadata = self.get_metadata_for_class(cls.class_object());
-                    let allow_dataclass = cls_metadata.dataclass_metadata().is_some_and(|m| {
-                        m.field_specifiers.iter().any(|k| {
-                            matches!(
-                                k,
-                                CalleeKind::Function(FunctionKind::DataclassField)
-                                    | CalleeKind::Class(ClassKind::DataclassField)
-                            )
-                        })
-                    });
-                    if allow_dataclass {
-                        dataclass_types.insert(cls.clone());
-                    } else {
-                        saw_non_dataclass = true;
-                    }
-                }
-                _ => saw_non_dataclass = true,
-            }
-            ty.clone()
+        self.map_over_union(&obj_ty, |ty| match ty {
+            Type::Any(_) => saw_any = true,
+            Type::ClassType(cls) if is_dataclass(cls) => saw_dataclass = true,
+            _ => saw_non_dataclass = true,
         });
 
-        let metadata = FuncMetadata {
-            kind: FunctionKind::DataclassReplace,
-            flags: Default::default(),
-        };
-
-        if dataclass_types.is_empty() {
-            if !saw_any {
-                self.error(
-                    errors,
-                    args.first().map(|x| x.range()).unwrap_or(range),
-                    ErrorInfo::Kind(ErrorKind::InvalidArgument),
-                    "dataclasses.replace() should be called on dataclass instances".to_owned(),
-                );
-            }
+        if saw_any {
             return self.callable_infer(
                 Callable::list(
                     ParamList::new(vec![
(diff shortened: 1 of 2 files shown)


rchen152:
Thanks! I'll review this as soon as I can.

pyrefly/lib/test/dataclasses.rs
```diff
@@ -67,6 +67,234 @@ Data(0, 1)  # E: Argument `Literal[1]` is not assignable to parameter `y` with t
...
+    item: T
+
+b = Box(item=1)
+replace(b, item=2)
+replace(b, item="wrong")  # E: Argument `Literal['wrong']` is not assignable to parameter `item` with type `int`
+    "#,
+);
+
+testcase!(
+    test_replace_union_of_two_dataclasses_rejects_bad_kw,
```
rchen152: This test seems to be a duplicate of the one at line 187.

pyrefly/lib/test/dataclasses.rs
```diff
@@ -67,6 +67,234 @@ Data(0, 1)  # E: Argument `Literal[1]` is not assignable to parameter `y` with t
...
+class Bar:
+    x: int
+
+def f(obj: Union[Foo, Bar]):
+    replace(obj, x=0)  # E: dataclasses.replace() expects a dataclass instance; got Bar | Foo
+    "#,
+);
+
+testcase!(
+    test_replace_union_two_dataclasses_rejects_bad_kw,
```
rchen152, line 187: Could you also add a test for an accepted `replace` call on a union of two dataclasses?

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -180,6 +187,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        let obj_ty = match args.first() {
+            Some(first) => self.expr_infer(first, errors),
+            None => Type::Any(AnyStyle::Implicit),
+        };
+        let mut dataclass_types = SmallSet::new();
+        let mut saw_any = false;
+        let mut saw_non_dataclass = false;
+        self.distribute_over_union(&obj_ty, |ty| {
+            match ty {
+                Type::Any(_) => saw_any = true,
```
rchen152: Could you add a test for the `saw_any = true` case?

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -180,6 +187,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                Type::ClassType(cls) => {
+                    let cls_metadata = self.get_metadata_for_class(cls.class_object());
+                    let allow_dataclass = cls_metadata.dataclass_metadata().is_some_and(|m| {
```
rchen152: nit: `allow_dataclass` is a slightly confusing name for this variable; how about `is_dataclass`?

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -180,6 +187,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        range: TextRange,
+        errors: &ErrorCollector,
+    ) -> Type {
+        let call_args = args.map(CallArg::expr_maybe_starred);
+        let call_keywords = keywords.map(CallKeyword::new);
+        let obj_name = Name::new_static("obj");
+
+        let obj_ty = match args.first() {
+            Some(first) => self.expr_infer(first, errors),
+            None => Type::Any(AnyStyle::Implicit),
```
rchen152: If `args` is empty, can we just fall back to a regular function call?

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -180,6 +187,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        if dataclass_types.is_empty() {
+            if !saw_any {
+                self.error(
+                    errors,
+                    args.first().map(|x| x.range()).unwrap_or(range),
+                    ErrorInfo::Kind(ErrorKind::InvalidArgument),
+                    "dataclasses.replace() should be called on dataclass instances".to_owned(),
+                );
+            }
+            return self.callable_infer(
```
rchen152: Similar to the comment about the empty `args` case above - I'd prefer falling back to the regular function call logic  in edge cases rather than adding more special behavior.

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -180,6 +187,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        let obj_name = Name::new_static("obj");
+
+        let obj_ty = match args.first() {
+            Some(first) => self.expr_infer(first, errors),
+            None => Type::Any(AnyStyle::Implicit),
+        };
+        let mut dataclass_types = SmallSet::new();
+        let mut saw_any = false;
+        let mut saw_non_dataclass = false;
+        self.distribute_over_union(&obj_ty, |ty| {
```
rchen152: I think you can simplify this function a lot by moving the core logic to be inside `distribute_over_union`, rather than first collecting the dataclass types and then trying to handle the union. I.e., do something like this:
```
self.distribute_over_union(&obj_ty, |ty| {
  // 1. Check if `ty` is a dataclass.
  // 2. If it is, construct and call a `replace` function like you do below.
}
```
Then I believe you can delete all the code that deals with checking the number of overloads, finding common keywords, etc.

rchen152 (changes requested):
Thanks for the PR, and sorry for the slow review! Most of the pieces look good; I had some suggestions for simplifying `call_dataclasses_replace`, which I think is more complex than it needs to be.

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -180,6 +187,232 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                errors,
+                errors,
+                None,
+                None,
+                None,
+            );
+        }
+
+        if !saw_dataclass {
+            self.error(
```
rchen152, line 277: FYI (no action needed on your part) I ended up patching the non-dataclass handling after importing this PR. The main issue was that I had to remove this check because it was generating too many false positives in internal tests.

rchen152 (approved):
Thanks! Getting an internal review now.

PR #1797, opened 2025-12-08
Consider an uninitialised ClassVar as abstract

fangyi-zhou:
# Summary

As per title, this diff addresses the last remaining issue in the
conformance test `protocols_explicit.py` by requiring declared classvar
members to be initialised for protocols.

The exception is that uninitialised class variables in interface files are not considered abstract. This exception is necessary to get around issues in typeshed.

# Test Plan

`python test.py`

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -1209,17 +1209,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     .cloned(),
             );
         }
-        // skip __hash__, but we might need to find a better way to handle it in the future.
-        // Context:
-        // https://github.com/facebook/pyrefly/pull/1797#issuecomment-3629910569
-        // and
-        // https://github.com/python/typeshed/issues/2148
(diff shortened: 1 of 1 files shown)


yangdanny97:
We may need to special-case `__hash__`, or update typeshed to add a `pyrefly: ignore` (pref the former, Mypy only added the type ignore because the referenced change was in typeshed & mypy's behavior was pre-existing)

https://github.com/python/typeshed/issues/2148

fangyi-zhou:
Got another issue with typeshed:

https://github.com/facebook/pyrefly/blob/d8e792838d99d16b14ec291e5cfdd11e266b1dc5/crates/pyrefly_bundled/third_party/typeshed/stdlib/pathlib/__init__.pyi#L58

I start to think we should not consider `ClassVar` declarations in `.pyi` files as abstract, does that sound reasonable?

yangdanny97:
> I start to think we should not consider ClassVar declarations in .pyi files as abstract

How widespread is it?

I think it sounds reasonable to exclude ClassVars in stubs for now, especially if it's common.

stroxler (approved):

PR #1798, opened 2025-12-09
#1795 Implement PEP 765: disallow return/break/continue in finally blocks

### LINKED ISSUE #1795: PEP 765: Disallow return/break/continue that exit a finally block
### Describe the Bug

https://peps.python.org/pep-0765/

I guess we'd want to emit a syntax error if the Python version is 3.14 or higher. If there is an enclosing loop in the finally block, then break/continue is OK inside that loop. If there is a nested function in the finally block, then return is OK inside that function.

The PEP mentions that runtime emits a SyntaxWarning not a SyntaxError, which may influence the severity of the error we emit.

```python
def test():
    try:
        pass
    finally:
        return # expect error for 3.14+

    for _ in "abc":
        try:
            pass
        finally:
            break # expect error for 3.14+

    for _ in "abc":
        try:
            pass
        finally:
            continue # expect error for 3.14+
```

More examples from the pep below

Should error

```
 def f():
     try:
         ...
     finally:
         return 42

 for x in o:
     try:
         ...
     finally:
         break  # (or continue)
```

Should NOT error

```
try:
     ...
 finally:
     def f():
         return 42

 try:
     ...
 finally:
     for x in o:
         break  # (or continue)
```

Python: 3.14

### Sandbox Link

https://pyrefly.org/sandbox/

_No response_

jackulau:
## Summary

  For #1795

  This implements [PEP 765](https://peps.python.org/pep-0765/) which disallows `return`, `break`, and `continue` statements that directly exit a `finally` block. Starting in Python
  3.14, these statements emit a `SyntaxWarning` because they can silently mask exceptions.

  ## The Issue

  When a `return`, `break`, or `continue` statement exits a `finally` block, any exception being propagated is silently discarded:

  ```python
  def f():
      try:
          raise ValueError("important error")
      finally:
          return 42  # Silences the ValueError!

  for x in range(10):
      try:
          raise ValueError()
      finally:
          break  # Silences the ValueError!
```
  However, these statements are valid when inside a nested function or nested loop within the finally block:
```python
  def f():
      try:
          pass
      finally:
          def inner():
              return 42  # OK: exits inner function, not finally

          for x in range(5):
              break  # OK: exits inner loop, not finally
```
  The Fix

  I made several changes to detect these violations during the binding phase:

    - Added InvalidReturnInFinally error kind in crates/pyrefly_config/src/error_kind.rs with Warn severity (matching Python's SyntaxWarning)
    - Added finally_depth tracking in pyrefly/lib/binding/scope.rs:
    - Added finally_depth: usize field to Scope struct
    - This resets to 0 when entering a new function scope, so nested functions are correctly allowed
    - Added enter_finally(), exit_finally(), and in_finally() methods to Scopes
    - Added in_finally flag on loops in pyrefly/lib/binding/scope.rs:
    - Added in_finally: bool field to Loop struct
    - Records whether a loop was started inside a finally block
    - Added loop_protects_from_finally_exit() method to check if break/continue targets a nested loop
    - Added checks for return/break/continue in pyrefly/lib/binding/stmt.rs:
    - Check in record_return() for return statements
    - Check in Stmt::Break and Stmt::Continue handlers
    - All checks are gated to Python 3.14+ via self.sys_info.version().at_least(3, 14)
    - Added documentation in website/docs/error-kinds.mdx for invalid-return-in-finally

  Test Plan

  | Test                                     | Description                    |
  |------------------------------------------|--------------------------------|
  | test_pep765_return_in_finally            | return in finally → warning    |
  | test_pep765_break_in_finally             | break in finally → warning     |
  | test_pep765_continue_in_finally          | continue in finally → warning  |
  | test_pep765_return_in_nested_function_ok | return in nested function → OK |
  | test_pep765_break_in_nested_loop_ok      | break in nested loop → OK      |
  | test_pep765_continue_in_nested_loop_ok   | continue in nested loop → OK   |
  | test_pep765_return_python313_no_error    | no warning on Python 3.13      |

  Verification:

  Run PEP 765 tests
  cargo test test_pep765
  Result: 7 passed

  Run all flow_branching tests
  cargo test flow_branching
  Result: 86 passed

  Run all tests
  cargo test
  Result: 3057 passed

  Fixes #1795

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -2580,7 +2580,10 @@ impl<'a> BindingsBuilder<'a> {
         // speculatively insert phi keys upfront.
         self.scopes.current_mut().flow =
             self.insert_phi_keys(base.clone(), range, loop_header_targets);
-        self.scopes.current_mut().loops.push(Loop::new(base, in_finally));
+        self.scopes
+            .current_mut()
+            .loops
+            .push(Loop::new(base, in_finally));
     }
 
     pub fn teardown_loop(


yangdanny97:
thanks, I'll get this reviewed

yangdanny97:
Found an edge case, if you have another finally inside a loop inside a finally, the error does not get emitted:

```
try:
    pass
finally:
    for _ in []:
        try:
            pass
        finally:
            break # expected error
```

I'll patch the scoping logic internally, no action needed from you

PR #1810, opened 2025-12-10
format params same as signature

kinto0:
Summary:
per LSP spec:
> a label of type string should be a substring of its containing signature label.

we must format these the same or editors might not display them

Differential Revision: D88863608


stroxler (approved):

Issue #1812, opened 2025-12-10
Flag definitely unreachable return/yield


yangdanny97:
### Describe the Bug

For cases where we know a return or yield is definitely unreachable, we should emit an error. This includes: anything after a `return` ,`raise`, `break`, `continue`, etc. Yields can follow other yields.

We should be conservative here and not use any narrowed type information, since that can be wrong and we can have false positives. Since we don't need type information, we can probably do this in the bindings phase.

```python
def test():
    return 1
    return 2 # expected error
    yield 1 # expected error
```

This really should be more of a lint rule, but it doesn't seem like Ruff flags this so we might as well.

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/

PR #1815, opened 2025-12-11
Flag unreachable `return` and `yield` statements

terror:
Resolves https://github.com/facebook/pyrefly/issues/1812

This diff adds detection for return and yield statements that can never be reached because they follow a statement that always terminates flow, such as return, raise, break, or continue. We now have a new `unreachable-statement` error kind.

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -249,6 +249,10 @@ pub enum ErrorKind {
     /// Identity comparison (`is` or `is not`) between types that are provably disjoint
     /// or between literals whose comparison result is statically known.
     UnnecessaryComparison,
+    /// A return or yield statement that can never be reached.
+    /// This occurs when a return/yield follows a statement that always exits,
+    /// such as return, raise, break, or continue.
+    UnreachableStatement,
     /// Attempting to use a feature that is not yet supported.
     Unsupported,
     /// Attempting to `del` something that cannot be deleted
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -482,6 +482,14 @@ impl<'a> BindingsBuilder<'a> {
     }
 
     fn record_yield(&mut self, mut x: ExprYield) {
+        // Check if this yield is unreachable (comes after a terminating statement)
+        if self.scopes.has_flow_terminated() {
+            self.error(
+                x.range,
+                ErrorInfo::Kind(ErrorKind::UnreachableStatement),
+                "This `yield` expression is unreachable".to_owned(),
+            );
+        }
         let mut yield_link = self.declare_current_idx(Key::YieldLink(x.range));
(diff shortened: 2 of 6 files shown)


yangdanny97:
I found 2 cases where this would give false positives:

error swallowing context managers

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEAxrugC4z4tQTZ0Q3G5KLOnACuxYpXhwAOujmYYYXnAD6ALxjUAFPkS9WASkRy6ZugHcILABYjxk6doBaW3ABEIANwhwIzAFFKakpjU3MIgEY6AHo6fHCIsykWUUp0OgAxVCg4GES6FLSMgBVKURgQABoQMikwKFJCFlooCgBiOgAFUnrGkQwcAkZmSABzNNQWf3RCOU6AZRgYOhsWFmI4RBiYuqVGwkExmJh0GMxcBjgYpnRxyenmOLBBOlQvVGhUbFgRu4gJpQpjM6LhiI90HA5ugyLZmABaLxaPzMOgAXjoMhAAGZCJEAExYuQgAC%2BNVQDGmSMy0EqiBAaCweCIZFJQA

Since the body of the context manager can throw, the return statement inside may not be execute so the bottom is still reachable

sys.version_info checks

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIEAtsbgE4AuABHKXADrqeYxhMPwGACgCUiTk0lMIfVnEIA3GHTgRc6APoR0YXEwB8AXiZCAzABomARitiJUh3RgMArnXRMAKnRcx7kp1d3JgAxVCg4GBBzEDInMChSQgZcKigKAGImAAVSeMSWDBwCJgBjdUgAczdUBjV0Qk4sgGUYGCYACwYGYjhEAHp%2BuN5EwnpK-ph0fsxcUrh%2B8p0Iarpa%2Bv6mXTomVAVUaFRsWDKKlZq69SZcYkv0eU4yBg71AFolFXqmY3YQU0IrAAmX6cEAAXxiqFKdSUIWgUUQIDQWDwRDI4KAA

Even though we can statically determine which path the code takes by evaluating sys.version_info, libraries still need to support multiple versions.

I'm not sure if it would over-complicate this PR to support those cases. Will think about this some more.

yangdanny97:
For context managers, we can't determine whether it's error-swallowing or not at the bindings phase, so the current analysis assumes it is non-swallowing.

This is a tricky tradeoff, since it might produce better default behavior in other cases, but for this specific analysis we really want a "are we 100% sure this statement is unreachable", and the way we treat context managers makes `has_terminated` not have that property.

Perhaps we can just add a different flag called `is_unreachable` that does encode the property that we want, where we treat context managers and statically evaluated conditions differently than we currently do.

I'll play around with that idea next week, no action needed from you for now, if it's easy enough to do I'll patch it onto this PR before merging.

PR #1822, opened 2025-12-11
Resolve django foreign key string and self references

ska-kialo:
# Summary

Resolves ForeignKey models defined as `"self"` or a string forward ref.
This latter requires that the model is imported in the module (e.g. under `if TYPE_CHECKING`).

I'm unsure about how robust/clean this code is as my knowledge of this project (and Rust in general) is pretty limited at the moment.

# Test Plan

Removed expected failures for cases that work now and added a test with an imported model.


rchen152 (approved):
This looks great to me, thank you so much! I'll work on getting this merged.

Issue #1811, opened 2025-12-10
Semantic Token Inconsistencies between Editors


miversen33:
### Describe the Bug

I am noticing some bugs around semantic tokens in neovim. Below is a screenshot of some code I am working on (code attached as well for reference)

<img width="3438" height="1438" alt="Image" src="https://github.com/user-attachments/assets/8d9459f6-3786-4383-ac3d-1a89ed22ce83" />

On the left is vscode, running pyrefly 0.45.1 and on the right is neovim running pyrefly 0.45.1.

You can see that the semantic tokens in vscode all seem to be applying correctly but in neovim they aren't, The one that caught my eye is the coloring for "MetadataManager", a custom class I am writing

<img width="1078" height="295" alt="Image" src="https://github.com/user-attachments/assets/f5f8dcba-688b-43a1-af81-2f5cf9fe45df" />

<img width="1197" height="380" alt="Image" src="https://github.com/user-attachments/assets/8fff5cb2-ba12-4db6-8a09-dce775a79f4c" />

Interestingly, pyrefly _is_ applying semantic tokens for _some_ stuff in Neovim, exceptions for example are being correctly highlighted

<img width="1048" height="477" alt="Image" src="https://github.com/user-attachments/assets/93b4d72d-94c4-428c-9aaa-d389a97f485b" />

I have noticed lots of weirdness around semantic tokens lately but I finally decided to submit something to see if I am doing something wrong here.
I am not sure how to "inspect" the highlighting of a token in vscode (or if that is even possible).

In neovim, I am using neovim's native `lsp` and my command is simply `pyrefly lsp`. Currently I am passing no options along to the lsp server.
I also attached the log from neovim's lsp client, maybe that is useful? Please let me know if there is anything else I can provide! 

[test_manager.py](https://github.com/user-attachments/files/24088374/test_manager.py)

[pyrefly.log](https://github.com/user-attachments/files/24088376/pyrefly.log)

yangdanny97:
> I am not sure how to "inspect" the highlighting of a token in vscode (or if that is even possible).

This is tricky, because the format that the tokens are represented isn't human-readable. cc @kinto0 

Thanks for the logs, we'll look into this.

kinto0:
In VSCode you can inspect tokens like this:

<img width="604" height="93" alt="Image" src="https://github.com/user-attachments/assets/f83425ed-0105-4fc1-9e42-75a9ecad3db5" />

I wonder if there's a similar way in neovim?

[...]

PR #1828, opened 2025-12-12
fix semantic tokens not having full bindings

kinto0:
Summary:
https://github.com/facebook/pyrefly/issues/1811 reported imports not working for semantic tokens in nvim, but not vscode.

nvim requests semantic tokens full, whereas vscode requests only the range of the window. I imagine a full semantic tokens request might not have everything ready, whereas ranged might. this could explain nvim showing semantic tokens everywhere except imports (we didn't have those bindings).

Differential Revision: D89005703

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -1050,7 +1050,7 @@ impl Server {
                         self.send_response(new_response(
                             x.id,
                             Ok(self
-                                .semantic_tokens_full(&transaction, params)
+                                .semantic_tokens_full(&mut transaction, params)
                                 .unwrap_or(default_response)),
                         ));
                     }
@@ -1067,7 +1067,7 @@ impl Server {
                         self.send_response(new_response(
                             x.id,
                             Ok(self
-                                .semantic_tokens_ranged(&transaction, params)
+                                .semantic_tokens_ranged(&mut transaction, params)
                                 .unwrap_or(default_response)),
                         ));
                     }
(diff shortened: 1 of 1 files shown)


miversen33:
I am certainly no rust expert but I am unable to built this. When I run `cargo build --release` from inside the repo, it seems to pull the wrong version of rustc?
image

This causes the cargo build to fail because it has the wrong version of rustc. Not really sure what I am missing here :/
image


Edit: I am running an x64 arch installation for testing purposes, though I would like to be able to build it myself as opposed to having someone spoon feed me :)

kinto0:
> I am certainly no rust expert but I am unable to built this. When I run `cargo build --release` from inside the repo, it seems to pull the wrong version of rustc? image
> 
> This causes the cargo build to fail because it has the wrong version of rustc. Not really sure what I am missing here :/ image
> 
> Edit: I am running an x64 arch installation for testing purposes, though I would like to be able to build it myself as opposed to having someone spoon feed me :)

hmm... I'm on `rustc 1.91.0-nightly` and `cargo 1.91.0-nightly`. the `rust-toolchain.toml` file should figure out which version to compile with, but maybe you need to start on a higher version so it can support those features?

kinto0:
here are some pip wheels for you to try if you don't want to do that:
[wheels-x86_64-unknown-linux-gnu.zip](https://github.com/user-attachments/files/24131554/wheels-x86_64-unknown-linux-gnu.zip)
[wheels-aarch64-unknown-linux-gnu.zip](https://github.com/user-attachments/files/24131555/wheels-aarch64-unknown-linux-gnu.zip)

kinto0:
if a binary is easier than a wheel, here's a linux x64 one:
https://drive.google.com/file/d/1YLtsF0Kgcxoz3ux27SP7IkW1wjYdTHwq/view?usp=sharing

(unfortunately it's 30 MB and github only allows uploading up to 25MB. maybe we could compress the stubs we bundle more)

miversen33:
I just removed the rust-toolchain file and switched to nightly and it built :)

Same issue though.
image

This time its pretty consistently not working, I am unsure if that is progress or not though
[lsp.log](https://github.com/user-attachments/files/24132663/lsp.log)
Here is my lsp log from the screenshot above

PR #1829, opened 2025-12-12
Do not enforce ordering for attrs classes

### LINKED ISSUE #1825: Feature:  support `attrs.field`
### Describe the Bug

Inside an `attrs` class, with the line `device: torch.device = field(default=Factory(lambda: torch.device("cpu")))`, I got an error `Dataclass field `device` without a default may not follow dataclass field with a default`.  Neither this is a dataclass, nor this is a wrong usage per `attrs` documentation.

This issue occurs with both the CLI and the extension, which are running the same version.

### Sandbox Link

_No response_

### IDE Information

IDE: `Windsurf 1.12.160+next.67a0e47281`, `VSCode OSS Version: 1.106.0-next`
Extension: `meta.pyrefly` with version `0.45.2`

validatedev:
# Summary
This pull request introduces a new mechanism for enforcing field ordering rules in dataclass-like class transformations, specifically to ensure that fields without defaults do not follow fields with defaults when required. The enforcement can be toggled based on the context, and a new test case is added to verify behavior when enforcement is not required.

Field ordering enforcement:

* Added a new `enforce_field_ordering` boolean field to the `DataclassMetadata` struct to control whether the rule "fields without defaults cannot follow fields with defaults" should be enforced.
* Updated the logic in `AnswersSolver` to set `enforce_field_ordering` to `true` for certain dataclass-like transformations (e.g., when using `@dataclass` and similar) and to `false` for others, allowing for flexible enforcement based on the decorator used. [[1]](diffhunk://#diff-4f2b77bb8cfa712de3bb3e0c6b55b101b63330176b6287cce88714adee81adb0R787) [[2]](diffhunk://#diff-4f2b77bb8cfa712de3bb3e0c6b55b101b63330176b6287cce88714adee81adb0R808) [[3]](diffhunk://#diff-4f2b77bb8cfa712de3bb3e0c6b55b101b63330176b6287cce88714adee81adb0R822)
* Modified the field validation logic in `dataclass.rs` to check `enforce_field_ordering` before applying the rule, so that enforcement is context-sensitive.

Testing:

* Added a new test case `test_dataclass_transform_allows_required_after_default` to verify that when `enforce_field_ordering` is `false`, required fields can follow fields with defaults without error.

Fixes #1825

Currently in draft.

# Test Plan

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -784,6 +784,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         alias_keyword: alias_keyword.clone(),
                         init_defaults: init_defaults.clone(),
                         default_can_be_positional,
+                        enforce_field_ordering: true
                     });
                 }
                 // `@dataclass(...)`
@@ -804,6 +805,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         alias_keyword: alias_keyword.clone(),
                         init_defaults: init_defaults.clone(),
                         default_can_be_positional,
+                        enforce_field_ordering: true
                     });
                 }
                 _ => {}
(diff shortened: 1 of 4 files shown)


validatedev:
This pull request does not address the actual issue. Closing.

PR #1830, opened 2025-12-12
fix panic: capacity overflow with large tuple expansion #1826

### LINKED ISSUE #1826: panic: `capacity overflow` with large tuple expansion
### Describe the Bug

```python
a: int | list[int] | tuple[int, ...] | bool
a + (a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a, a)
```

```
ERROR Thread panicked, shutting down: panicked at /rustc/02c7b1a7ac1d739663878030510508372e46f254/library/core/src/iter/traits/iterator.rs:2027:9:
capacity overflow
```

### Version info
0.45.2

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1826

Prevent tuple-argument overload expansion from overflowing by counting combinations with usize::saturating_mul and bailing out as soon as the limit is exceeded, keeping the old GAS threshold behavior intact while avoiding panics

# Test Plan

Added a regression test that exercises a long tuple literal built from a tuple-or-bool union to ensure we no longer crash while still reportingthe expected unsupported-operation diagnostic

### DIFF
--- a/pyrefly/lib/test/tuple.rs
+++ b/pyrefly/lib/test/tuple.rs
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+use crate::test::util::TestEnv;
 use crate::testcase;
 
 testcase!(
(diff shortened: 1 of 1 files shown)


kinto0:
awesome!!

- I am verifying this test failed before your changes
- would you mind commenting on the original issue so we can assign you? (I wonder if we can just give you assign permissions...)

pyrefly/lib/test/tuple.rs
```diff
@@ -397,6 +397,14 @@ def test(x: tuple[int] | tuple[str]) -> None:
 "#,
 );
 
+testcase!(
+    test_tuple_concat_large_union,
+    r#"
+def test(a: tuple[int, ...] | bool) -> None:
```
kinto0: hmm it looks like even in dev mode this test passes before your changes. I wonder if you had the same experience? I'd like to make sure this regression test ensures we don't accidentally create this issue again.

asukaminato0721: aa, I checked again, this test worked before, but origin issue's will trigger very long error.

I set a function becase use raw a will raise `a` is uninitialized.

will update this test example.

kinto0: thanks!! looks great now

PR #1831, opened 2025-12-12
impl method extraction, part of #746

asukaminato0721:
# Summary

part of #746

Added method-aware extraction plumbing without disrupting the original helper flow: helper actions are still emitted first while method-specific actions get appended, receiver detection skips `@staticmethod/@classmethod`, and helper construction utilities now share indentation logic

# Test Plan

add new tests


stroxler (approved):

PR #1832, opened 2025-12-12
Add constant highlighting for ALL_CAPS variables

### LINKED ISSUE #1780: Semantic highlighting: Highlight some top level variables as constants
### Describe the Bug

Right now they are highlighted the same as other variables, but perhaps we should highlight the variables different if they only contain capital letters, numbers, and underscore.

If we want to get REALLY fancy, maybe we could also check that 
1. they defined in an assignment where the RHS is a literal (or the variable has a literal type)
2. they are never re-assigned

I think maybe just implementing the basic condition is good enough.

### Sandbox Link

_No response_

_No response_

AryanBagade:
Fixes #1780

# Summary
Configure VSCode extension and web sandbox to highlight constants (ALL_CAPS variables) in a distinct teal color. The backend already detects ALL_CAPS variables as constants and sends the `readonly` modifier - this change ensures they're visually distinguished in both the VSCode extension and the pyrefly.org sandbox.

# Test Plan

sandbox ui 
CleanShot 2025-12-11 at 22 49 06@2x

vscode
CleanShot 2025-12-11 at 22 48 11@2x


kinto0:
cool! thanks for the research and exhaustive test plan

PR #1833, opened 2025-12-12
fix Pyrefly Fails to Report Non-Exhaustive match Statements and Incorrect Enum Case Patterns #400

### LINKED ISSUE #400: Pyrefly Fails to Report Non-Exhaustive match Statements and Incorrect Enum Case Patterns
### Describe the Bug

**Observed Behavior (with Pyrefly):**

When analyzing Python `match` statements, Pyrefly currently reports no errors under the following conditions where issues are present:

* **Non-exhaustive matching:** When `case` statements within a `match` block do not collectively handle all possible members of an `Enum` or all types within a `Union` that the matched variable could be.
* **Incorrect syntax for Enum member case patterns:** When `Enum` members used in `case` patterns are incorrectly written with parentheses as if they were being called (e.g., `MyEnum.MEMBER()` instead of the correct syntax `MyEnum.MEMBER`).

This behavior implies Pyrefly currently considers these potentially problematic patterns as valid.

**Expected Behavior:**

Pyrefly should identify and report errors in the scenarios described above to help developers catch potential bugs and ensure code correctness:

* **For non-exhaustive matching:** Pyrefly should issue an error or warning, ideally specifying which `Enum` members or `Union` types are not handled by any `case` statement.
* **For incorrect Enum syntax:** Pyrefly should report a syntax or type error when `Enum` members are incorrectly used with parentheses in `case` patterns (e.g., similar to Pyright's `"Literal[...] is not a class"` report).

This would align Pyrefly's strictness with other type checkers and provide more robust static analysis for `match` statements.

Pyright's Output (for reference, from the original user report):
```
/Users/davi/gits/chief-app/test_types.py
  /Users/davi/gits/chief-app/test_types.py:21:11 - error: Cases within match statement do not exhaustively handle all values
    Unhandled type: "OrderStatus"
    If exhaustive handling is not intended, add "case _: pass" (reportMatchNotExhaustive)
  /Users/davi/gits/chief-app/test_types.py:22:14 - error: "Literal[OrderStatus.Ready]" is not a class (reportGeneralTypeIssues)
  /Users/davi/gits/chief-app/test_types.py:24:14 - error: "Literal[OrderStatus.Shipped]" is not a class (reportGeneralTypeIssues)
  /Users/davi/gits/chief-app/test_types.py:30:11 - error: Cases within match statement do not exhaustively handle all values
    Unhandled type: "Literal[OrderStatus.Scheduled]"
    If exhaustive handling is not intended, add "case _: pass" (reportMatchNotExhaustive)
  /Users/davi/gits/chief-app/test_types.py:52:11 - error: Cases within match statement do not exhaustively handle all values
    Unhandled type: "int"
    If exhaustive handling is not intended, add "case _: pass" (reportMatchNotExhaustive)
5 errors, 0 warnings, 0 information 
```
**Steps to Reproduce:**

1.  **Prepare the code:** Save the Python code snippet provided in the "Sandbox Link" section of this GitHub issue (which contains functions `handle_order_1`, `handle_order_2`, and `handle`) into a local file (e.g., `test_match_issues.py`). This code demonstrates the non-exhaustive matches and incorrect Enum syntax.
2.  **Analyze with Pyrefly:** Run `pyrefly` on the created file:
    ```bash
    pyrefly check test_match_issues.py
    ```
3.  **Observe Pyrefly's output:** Notice that Pyrefly does not report any errors for the problematic `match` statements within the `handle_order_1`, `handle_order_2`, or `handle` functions.
4.  **(Optional Comparison) Analyze with Pyright:** Run `pyright` on the same file:
    ```bash
    pyright test_match_issues.py
    ```
5.  **Observe Pyright's output:** Notice that Pyright correctly identifies and reports multiple errors related to non-exhaustiveness and incorrect enum case patterns for the same functions, as detailed in the "Pyright's Output" section of this issue. This highlights the behavior expected from Pyrefly.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBApgdgV2gSwgBzCALlAooiAKEIGIoB5EZAc2TgEMAbKAZTARAGMYAuKACyxY0AZx4B6cf3oBrZACN4yLPQB0nSOLQBPLPzBwAtBG07DMAB7SEIrMgBuMQ534xOcuNVWkoAFX7IIlAaACYwsBZojJgwQZbWtg7wsUEubh7UUHRQEPRYLlC2eTAQ8Fgi3mQAYphQAAq6+nBQAMyqAIztANQANFBYpnRe9CIiMNgA+nAwjiBZQfT29MiM9PKMvAJCohLiA2hDqiAw9CF6MCFgnBXIYOLw4qtYsVjiIuxcMOIIcMf0LmsNqpBBBGD5-OFLPR0BsgopogB3KD0Y5QGwXfpgfovCIJOyOaajYKudxDHwI5T8eqNAytDoABgA5EFcvkqUVnqU4FhVNSqNRBMyoIppPZbnMABQaEDHThYRjaLJhbnIYDaIawGWYEQASnmUDgYGeIUqfgCQWOGGwAho-EYtvK-WkOB0x2ACqgEoRAQKlswxo9dBCyE4xSCACIwDJw1BaobNeAQEFgLVzmNgiNYjqfCJ9AhGCElSdGB6mO8iyq1U7YuEMM8VUx5iIELEoCm5uB5DYcPtwmlSZ5vIROKsiZQwiAWCosDYJfgkDqeIQoCuoAAlE4hRUAXig4b+W-Dy9XLDSIXzGN34ZEZ4vISPq9YATQaEve9zyBfFyPPgabo9ubsAWRZMB6frYEuZDtLyAByBjmFY9DdkkOR5AUErjuMU55DYqinq454bIWECBCIQw6qaABMvIAJJwNKsounkzwgM0AAGmGTtOuEbqc2gSjqbGeoB+aFooUAcSAE7YTOFS8VubEUT4bRQHRDFuExQjjOxnEybhLDPq+IQCUJEoicB4mSdJ3EVAZn5GYp3hhMAAj0HAIQbBMmAThM7RmTZfC6TZeqGAAfFAcHTEuj6suhHI2IuUBkH+DpQPa0xQJRAAcWTNJgNB0I2UIwjAx6PiuobpkFOFyZu-GJcl2j8oIaV0OElEAJxleVj5oFQ3IStez6xlJ4zhtmPUVZmFCjVxNV4YZFwCXwjXNTg6XhC0lHdZNUB9XQWCDQCnjhN5Y3Zr+TUwO6irmYWyBhKBirgVgkFQNBEXwfESGJI4qFsp61WyXht5ETkpHkU512ue5nlnSAEyUf5NWBbNem6lAYWfVF3WxVSyOyQ1fKpRtUAACz0rlI0FQwzDFVEpW7ZV4RAzxdXRbtq77QNQ2fiNE7jTtq7MzN1nzXZX4hBznN7f1h3hsd1CnbNguPiAyzpgAakwLa4FqIASsA4bo1AADe8UiAAvvqho4Es9r3hdZD+IE+rAG1fSutdAF5sBL0Gli4yJhUhDOTDHkwF5s0TC0BM2KjYuE9L-1xcFyePiLrO1Xx6eTdz8sfmg-PnULU1VWjNkgwRd65z1+eDYXxcgKrk2ZxX4uLVLpflfXCvSCdTeC4QdFYJQU5zLuB1QAAPoUWAgMQq3e7dvv3Y9JbPTAVqvT4H2RQheIoXjnpsQdQm9uDoyQ6H0P9xHEoWFAfAj2P88heFkW8LjaFUhYK3EwKdabUoAAHZKJU3yrQWmERoQM27iLWwBtFzd16nLQaAAJNyEdHZAA

_No response_

asukaminato0721:
# Summary

Fixes #400

now records per-case metadata (literal/class patterns, guard info, coverage range) and emits a Binding::MatchStmt, enabling later passes to reason about matches.

defines the new MatchStmtInfo/MatchCaseInfo structures, adds the MatchStmt binding variant, and wires it into symbol_kind/display logic so the solver can consume the recorded metadata.

adds check_match_stmt plus helpers that a) warn when a match over a single Enum lacks coverage for some members (NonExhaustiveMatch) and b) flag class patterns that are actually Enum members (InvalidEnumPattern). Enum domain analysis handles unions of literals, skips flags, and only triggers when the user is enumerating specific members.

introduces invalid-enum-pattern (error) and non-exhaustive-match (warn) error kinds with default severities.

documents both new diagnostics with examples so the public docs stay in sync.

# Test Plan

adds integration tests covering both new diagnostics (Color.RED() misuse and missing Enum cases) to keep behaviour locked down.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -86,7 +86,6 @@ use crate::binding::binding::KeyLegacyTypeParam;
 use crate::binding::binding::KeyUndecoratedFunction;
 use crate::binding::binding::LastStmt;
 use crate::binding::binding::LinkedKey;
-use crate::binding::binding::MatchStmtInfo;
 use crate::binding::binding::NoneIfRecursive;
 use crate::binding::binding::RaisedException;
 use crate::binding::binding::ReturnTypeKind;
@@ -117,7 +116,6 @@ use crate::types::class::Class;
 use crate::types::class::ClassType;
 use crate::types::display::TypeDisplayContext;
 use crate::types::literal::Lit;
-use crate::types::literal::LitEnum;
 use crate::types::module::ModuleType;
 use crate::types::param_spec::ParamSpec;
 use crate::types::quantified::Quantified;
@@ -177,25 +175,6 @@ pub enum TypeFormContext {
     /// Variable annotation outside of a class definition
     /// Is the variable assigned a value here?
     VarAnnotation(AnnAssignHasValue),
-}
-
-#[derive(Clone)]
-enum EnumMatchDomain {
-    Full {
-        class: ClassType,
-    },
-    Restricted {
-        class: ClassType,
-        allowed: SmallMap<Name, LitEnum>,
-    },
-}
-
-impl EnumMatchDomain {
-    fn class(&self) -> &ClassType {
-        match self {
-            EnumMatchDomain::Full { class } | EnumMatchDomain::Restricted { class, .. } => class,
-        }
-    }
 }
 
 impl TypeFormContext {
@@ -1629,6 +1608,58 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         self.solver().expand_vars_mut(ty);
     }
 
+    fn is_pure_enum_class(&self, cls: &ClassType) -> bool {
+        self.get_metadata_for_class(cls.class_object())
+            .enum_metadata()
+            .is_some_and(|meta| !meta.is_flag)
+    }
+
+    fn is_pure_enum_type(&self, ty: &Type) -> bool {
+        match ty {
+            Type::ClassType(cls) | Type::SelfType(cls) => self.is_pure_enum_class(cls),
+            Type::Literal(Lit::Enum(lit_enum)) => {
+                let lit_enum = lit_enum.as_ref();
+                self.is_pure_enum_class(&lit_enum.class)
+            }
+            Type::Union(union) => {
+                let union = union.as_ref();
+                !union.members.is_empty()
+                    && union
+                        .members
+                        .iter()
+                        .all(|member| self.is_pure_enum_type(member))
+            }
+            _ => false,
+        }
+    }
+
+    fn format_enum_literal_cases(&self, ty: &Type) -> Option<String> {
+        fn collect_cases(ty: &Type, acc: &mut Vec<String>) -> bool {
+            match ty {
+                Type::Literal(Lit::Enum(lit_enum)) => {
+                    let lit_enum = lit_enum.as_ref();
+                    acc.push(format!("{}.{}", lit_enum.class.name(), lit_enum.member));
+                    true
+                }
+                Type::Union(union) => {
+                    let union = union.as_ref();
+                    union
+                        .members
+                        .iter()
+                        .all(|member| collect_cases(member, acc))
+                }
+                _ => false,
+            }
+        }
+
+        let mut cases = Vec::new();
+        if collect_cases(ty, &mut cases) {
+            Some(cases.join(", "))
+        } else {
+            None
+        }
+    }
+
     fn check_del_typed_dict_field(
         &self,
         typed_dict: &Name,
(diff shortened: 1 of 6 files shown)


yangdanny97:
Thanks for the PR, I'll give this a more detailed review when I have time. Just a quick comment on architecture for context:

Match exhaustiveness affects control flow modeling in the bindings layer, but it can't be modeled precisely because it depends on resolved types, which are not available until the solving layer in our exports -> binding -> solving architecture.

So things like knowing when a match w/o `_` cases is exhaustive & using it to determine reachability of subsequent statements, for example, would be very hard to do under the current architecture.

Overall, I think this PR is a good start in terms of reporting potential bugs/errors to users, but there will be some remaining gaps around control flow that we can address later, which may require architecture changes.

grievejia:
I get the impression that the PR only deals with non-exhaustive enum matching, not anything else. And it doesn't quite deal with any type-induced consequences with control flow.

yangdanny97:
Additionally:

> flag class patterns that are actually Enum members (InvalidEnumPattern)

I think this can be done in a separate PR. My preferred implementation would be a new binding that gets generated for each class-pattern case which does the checking + solving for the class type, and updating the `PatternMatchClassPositional`  + `PatternMatchClassKeyword` bindings to use the result of that pattern instead of storing the expression and solving it ad-hoc each time.

> Enum domain analysis handles unions of literals, skips flags, and only triggers when the user is enumerating specific members.

These all good properties, I think. The only thing is that I'm not sure the "only triggers when the user is enumerating specific members" is necessary - I think our type narrowing behavior should work pretty well for enums and literal unions, so we may not need to be so strict here.

pyrefly/lib/alt/solve.rs
```diff
@@ -1608,6 +1608,58 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                        .all(|member| self.is_pure_enum_type(member))
+            }
+            _ => false,
+        }
+    }
+
+    fn format_enum_literal_cases(&self, ty: &Type) -> Option<String> {
+        fn collect_cases(ty: &Type, acc: &mut Vec<String>) -> bool {
+            match ty {
+                Type::Literal(Lit::Enum(lit_enum)) => {
```
yangdanny97, line 1639: Lit implements display, you can do `lit_enum @ Lit::Enum(_)` and then `format!("{}", lit_enum)` and it will do the same thing

pyrefly/lib/binding/pattern.rs
```diff
@@ -388,13 +397,31 @@ impl<'a> BindingsBuilder<'a> {
...
             negated_prev_ops.and_all(new_narrow_ops.negate());
-            self.stmts(case.body, parent);
+            self.stmts(body, parent);
             self.finish_branch();
         }
         if exhaustive {
             self.finish_exhaustive_fork();
         } else {
             self.finish_non_exhaustive_fork(&negated_prev_ops);
+            if let Some(subject_name) = match match_narrowing_subject.as_ref() {
```
yangdanny97, line 407: There's a bug somewhere here. if the subject is `NarrowingSubject::Facets` I don't think it works, since we don't preserve the facet information into the solving step.

For example, this:

```
from enum import Enum

class Color(Enum):
    RED = "red"
    BLUE = "blue"

class X:
    color: Color

def describe(x: X) -> str:
    match x.color:
        case Color.RED:
            return "danger"
    return "cool"
```

If we add a case for Color.BLUE the error still gets shown

yangdanny97:
I'll see if it's easy to patch internally, no action needed from your for now

yangdanny97:
filed https://github.com/facebook/pyrefly/issues/2153 as followup

PR #1834, opened 2025-12-12
fix Narrowed type is not reflected in nested scopes #40

### LINKED ISSUE #40: Narrowed type is not reflected in nested scopes
```python
from typing import Callable

class Foo:
    _window_function: Callable[[str], int] | None

    def foo(self) -> None:
        if (window_function := self._window_function):
            def bar() -> None:
                print(window_function('foo'))
            bar()

```
```
$ pyrefly check g.py 
ERROR g.py:9:23-38: Expected a callable, got None [not-callable]
```

Spotted in [Narwhals](https://github.com/narwhals-dev/narwhals)

asukaminato0721:
# Summary

Fixes #40

Added Scope::nested_function so nested defs use
FlowBarrier::AllowFlowUnchecked, letting them inherit narrowings and first-use info from the enclosing flow while still protecting top-level functions and methods with the stricter barrier.

Updated push_function_scope to automatically select the nested flow barrier whenever we enter a function inside another function, so closures and local defs now see the narrowed types available at their creation site.

# Test Plan

Reproduced the GitHub report with a new regression test that exercises a nested def capturing a walrus-assigned callable, ensuring the narrowed callable type remains visible inside the nested scope

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -835,8 +835,8 @@ impl ConfigFile {
         self.handle_from_module_path_with_fallback(module_path, &FallbackSearchPath::Empty)
     }
 
-    pub fn handle_from_module_path_with_fallback<'a>(
-        &'a self,
+    pub fn handle_from_module_path_with_fallback(
+        &self,
         module_path: ModulePath,
         fallback_search_path: &FallbackSearchPath,
     ) -> Handle {
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -1047,6 +1047,9 @@ impl<'a> BindingsBuilder<'a> {
         if matches!(style, FlowStyle::Other) {
             self.scopes.register_variable(name);
         }
+        if !matches!(style, FlowStyle::Uninitialized) {
+            self.scopes.record_assignment(name);
+        }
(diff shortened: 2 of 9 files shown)


yangdanny97:
Sorry I didn't see this was updated since the status still said "changes requested".

I'll need more time to give a detailed review. We need to be super careful about capturing names from enclosing scopes - I did something similar for preserving whether or not variables from enclosing scopes are initialized, and it really tanked the performance (tripling type checking times because there were just so many names). This change is conceptually similar to that one.

The primer error delta looks decent, which is a good sign though I think it's worth investigating some of the new errors.

yangdanny97:
I ended up doing this in a different way, by front-loading some calculations at the definitions stage instead of bindings to get better performance.

The high level overview is that in definitions we calculate which variables are read in a given scope, and the last assignment to a each variable in that scope.

The variables that are read in a function's body, which are not defined within the body or parameters are the set of variables eligible to be captured. We always treat captured variables are initialized, and additionally we persist their narrowed type from outside if the last assignment to that variable in the enclosing scope is before the closure.

So we can handle patterns like this

```
x: int | None = ...
assert x is not None
def foo():
   # x should not get narrowed here
x = None
foo()
```

Thanks for your work on this PR, it was very useful prototyping this path even though we didn't end up using it.

PR #1844, opened 2025-12-13
Revert back to `debug = 1` for now

stroxler:
Summary:
We aren't sure why but the build seems to have broken, we're getting
a failure with `rustc` on exactly one of the build jobs (build binaries)
with an error
```
rustc-LLVM ERROR: out of memory
Allocation failed
```

Let's roll back the change to see whether debug symbols are indeed
causing an OOM; sadly this means we still can't debug some issues,
but we can't do it with a broken build in any case.

Differential Revision: D89101563


rchen152 (approved):

PR #1846, opened 2025-12-13
fix Disallow assigning to Final/@final members & variables #45

### LINKED ISSUE #45: Disallow assigning to Final/@final members & variables

asukaminato0721:
# Summary

Fixes #45

Attribute assignments now carry an allow_read_only flag that is set only when we’re inside a recognized instance initializer (e.g., `__init__`); this relies on the new Scopes::method_that_sets_attr helper so we can distinguish constructor writes from all other contexts.

The solver and attribute checker honor that flag and now treat every Final class or instance attribute as read-only by default, only permitting writes when allow_read_only is true (constructor initialization). This required threading the flag through the assignment pipeline and updating read-only handling so only ReadOnlyReason::Final can be bypassed.

# Test Plan

Added a regression test that demonstrates self.x = ... inside another method and a.x = ... at call sites now properly error for Final instance attributes, while initialization in `__init__` remains allowed

### DIFF
--- a/pyrefly/lib/test/attributes.rs
+++ b/pyrefly/lib/test/attributes.rs
@@ -420,7 +420,7 @@ class A:
     def __init__(self):
         self.x: Final = 0
 def f(a: A):
-    assert_type(a.x, Literal[0])  # E: assert_type(int, Literal[0])
+    assert_type(a.x, Literal[0])
     "#,
 );
 


pyrefly/lib/alt/class/class_field.rs
```diff
@@ -1853,22 +1853,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         field_definition: &ClassFieldDefinition,
     ) -> Option<ReadOnlyReason> {
         if let Some(ann) = annotation {
-            // TODO: enable this for Final attrs that aren't initialized on the class;
```
yangdanny97, line 1856: hooray

pyrefly/lib/alt/attr.rs
```diff
@@ -727,6 +727,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         got: &ExprOrBinding,
         range: TextRange,
         errors: &ErrorCollector,
+        allow_read_only: bool,
```
yangdanny97, line 730: nit: range/errors/ctx is typically at the end of the parameter list

pyrefly/lib/alt/attr.rs
```diff
@@ -850,6 +852,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         errors: &ErrorCollector,
         context: Option<&dyn Fn() -> ErrorContext>,
         todo_ctx: &str,
+        allow_read_only: bool,
```
yangdanny97, line 855: nit: same as above, I think this should go right after `got`

pyrefly/lib/binding/binding.rs
```diff
@@ -1384,7 +1384,13 @@ pub enum Binding {
     SuperInstance(SuperStyle, TextRange),
     /// The result of assigning to an attribute. This operation cannot change the *type* of the
     /// name to which we are assigning, but it *can* change the live attribute narrows.
-    AssignToAttribute(ExprAttribute, Box<ExprOrBinding>),
+    AssignToAttribute {
+        attr: ExprAttribute,
+        value: Box<ExprOrBinding>,
+        /// Some contexts (like initializing a `Final` field inside `__init__`) are allowed to
+        /// write to otherwise read-only attributes.
+        allow_read_only: bool,
```
yangdanny97, line 1392: nit: since it's only used for final, we should name it as such, something like `allow_assign_to_final`

pyrefly/lib/binding/scope.rs
```diff
@@ -1449,6 +1449,39 @@ impl Scopes {
...
+                            ),
+                            instance_or_class: receiver_kind,
+                        });
+                    }
+                }
+                _ => {}
+            }
+        }
+        None
+    }
```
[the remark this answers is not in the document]
yangdanny97: Based on this example, Mypy allows assignments inside nested functions inside `__init__`, while Pyright does not.

```
from typing import Final
class C:
    x: Final[int]
    def __init__(self):
        def inner():
            self.x = 1
```

so I think we could go either way here. Perhaps we can match mypy since that's more permissive, tho either way will have tradeoffs.

We should add a comment here to be explicit about the decision, and add an explicit integration test case

pyrefly/lib/binding/target.rs
```diff
@@ -159,14 +160,25 @@ impl<'a> BindingsBuilder<'a> {
         } else {
             self.declare_current_idx(Key::Anon(attr.range))
         };
+        let allow_read_only = self
```
yangdanny97, line 163: nit: rename variable

PR #1847, opened 2025-12-13
fix TypedDict NotRequired key narrowing + access error #84

### LINKED ISSUE #84: TypedDict `NotRequired` key narrowing + access error
I believe this is optional in the typing spec

1. support narrowing a `NotRequired` key to be `Required` using `in` and `.get()`
2. warn on accessing `NotRequired` keys that have not been narrowed

asukaminato0721:
# Summary

Fixes #84

Added AtomicNarrowOp::HasKey/NotHasKey plus a membership matcher so comparisons like 'foo' in mapping (and their not in negations) now narrow the key facet ahead of time, including .get subjects.

Extended the solver to consume the new ops, storing key presence/absence and reusing it through control flow; accessing a string key now first records the guard-aware facet and keeps the base type in sync.

subscript_infer now emits a not-required-key-access warning whenever a NotRequired/total=False key is read without a preceding guard, and helper logic walks union/var cases before reporting.

Introduced the new NotRequiredKeyAccess error kind + documentation entry so the warning can be configured/suppressed like other diagnostics.

# Test Plan

Added regression tests covering both the guarded ("in") and unguarded access paths to make sure .get + membership checks narrow correctly and bare subscripts warn.

### DIFF
--- a/pyrefly/lib/binding/narrow.rs
+++ b/pyrefly/lib/binding/narrow.rs
@@ -388,14 +388,8 @@ impl NarrowOps {
     }
 
     fn or(&mut self, name: Name, op: NarrowOp, range: TextRange) {
-        match self.0.entry(name) {
-            Entry::Occupied(mut entry) => {
-                entry.get_mut().0.or(op);
-            }
-            Entry::Vacant(entry) => {
-                entry.insert((op, range));
-            }
-        }
+        let existing_op = self.get_or_placeholder(name, range);
+        existing_op.or(op)
     }
 
     pub fn or_all(&mut self, other: Self) {


pyrefly/lib/binding/narrow.rs
```diff
@@ -492,10 +527,18 @@ impl NarrowOps {
...
                     }
                 }
+                let mut left_for_membership = left;
+                let mut membership_narrows = Vec::new();
                 let mut ops = cmp_ops
                     .iter()
                     .zip(comparators)
                     .filter_map(|(cmp_op, right)| {
+                        if let Some(extra) =
+                            Self::from_membership_comparison(left_for_membership, *cmp_op, right)
```
yangdanny97: are we able to fold this into the pattern match below instead of having a special logic to the side?

pyrefly/lib/binding/narrow.rs
```diff
@@ -74,6 +74,8 @@ pub enum AtomicNarrowOp {
     TypeNotEq(Expr),
     In(Expr),
     NotIn(Expr),
+    HasKey(Name),
```
yangdanny97, line 79: may want a docstring explaining how this differs from In/NotIn (in/notin is `<subject> in <container>` while this is `<key> in <subject>`

pyrefly/lib/alt/expr.rs
```diff
@@ -1207,6 +1207,60 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         false
     }
 
+    fn type_has_optional_typed_dict_key(&self, ty: &Type, key: &Name) -> bool {
```
yangdanny97: nit: if we do end up with helpers here, I'd rather name these notrequired to align with the actual type name (and also because the current can be conflated w/ a key that has type `typing.Optional[T]`

pyrefly/lib/test/subscript_narrow.rs
```diff
@@ -241,6 +241,35 @@ def use2(mapping: TD) -> None:
...
+from typing import TypedDict, NotRequired, assert_type
+
+class TD(TypedDict):
+    foo: NotRequired[int]
+
+def use(mapping: TD) -> None:
+    if "foo" in mapping:
+        assert_type(mapping["foo"], int)
+    else:
+        mapping["foo"]  # E: TypedDict key `foo` may be missing; guard this access with `'foo' in obj` or `obj.get('foo')`
```
yangdanny97, line 256: can you add a test case for `.get` as well, please? I don't remember if its behavior is special-cased.

pyrefly/lib/test/subscript_narrow.rs
```diff
@@ -241,6 +241,35 @@ def use2(mapping: TD) -> None:
...
+    r#"
+from typing import TypedDict, NotRequired
+
+class TD(TypedDict):
+    foo: NotRequired[int]
+
+def bad(mapping: TD) -> int:
+    return mapping["foo"]  # E: TypedDict key `foo` may be missing; guard this access with `'foo' in obj` or `obj.get('foo')`
+"#,
+);
```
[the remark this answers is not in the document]
yangdanny97: Agree w/ copilot here

pyrefly/lib/test/subscript_narrow.rs
yangdanny97: also, a test case for `if "foo" not in mapping` + accessing the key in else branch

pyrefly/lib/binding/narrow.rs
```diff
@@ -436,6 +442,35 @@ impl NarrowOps {
         narrow_ops
     }
 
+    fn from_membership_comparison(left: &Expr, cmp_op: CmpOp, right: &Expr) -> Option<Self> {
+        match cmp_op {
```
yangdanny97: nit: we can lift the cmp_op check outside the function call, and pass a boolean flag for in/not_in

pyrefly/lib/alt/expr.rs
```diff
@@ -1255,7 +1309,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 })
             }
             Expr::StringLiteral(ExprStringLiteral { value: key, .. }) => {
-                TypeInfo::at_facet(base, &FacetKind::Key(key.to_string()), || {
+                let key_str = key.to_string();
+                self.warn_if_optional_typed_dict_key_access(base, &key_str, slice.range(), errors);
```
yangdanny97: subscript_infer_for_type already has a case for handling TypedDicts, I'd prefer to avoid adding places where we match on the Type struct, so I think we should just add that warning there.

Right now we don't check whether the field is NotRequired, it would be fairly simple to add a condition and emit the warning there. If the narrowed facet exists, that code path will not be hit at all.

https://github.com/facebook/pyrefly/blob/db0d49be8e8424f1ba8e6893436a961dcf8f8a67/pyrefly/lib/alt/expr.rs#L2196

Then the additional logic in type_has_optional_typed_dict_key + warn_if_optional_typed_dict_key_access are no longer required.

yangdanny97:
Sorry this is taking a while to get merged, there are some changes I need to make on top of the PR and merge together.

The error is fairly noisy on codebases that make use of total=False dicts unsafely, so I'm considering whether we should split the error code or make it off-by-default. I'm not decided on the approach yet.

Also, it seems like there's a weird interaction between this and `.get()` narrowing. The merge of the truthy/falsy branches will yield a facet, which will then obscure the not-requiredness of the key. Even outside the guard, future accesses will not raise a warning.

No action needed from you, just keeping you updated on the status

yangdanny97:
I'll make this error off-by-default for now. The reasoning is that in test files, users may want to perform unsafe accesses as an assertion ("I expect the key to be present, otherwise crash and fail the test").

PR #1848, opened 2025-12-13
fix Using a Literal key to update values in a TypeDict #329

### LINKED ISSUE #329: Using a Literal key to update values in a TypeDict
### Describe the Bug

I'm not sure how exactly to describe this, but it isn't working as I expect.

Even though I have guarded my key variable with a Literal, pyrefly does not seem to respect the None check.

This works as expected when my key is the actual string.

```
from typing import Any, TypedDict, Literal

class MyDict(TypedDict):
    my_key: list[str] | None


def fun1():
    data: MyDict = {'my_key': None}
    if data['my_key'] is None:
        data['my_key'] = []
    else:
        data['my_key'].append('data')


def fun2():
    data: MyDict = {'my_key': None}
    key: Literal['my_key'] = 'my_key'

    if data[key] is None:
        data[key] = []
    else:
        data[key].append('data')  # this fails
        data['my_key'].append('data')  # also fails
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFACqICmAJgCJIDGMZAMkjFSAIYA2AUP0a9uAZ1FQAsnAbMAFJQS1ZMAJQAuflG1QIcAPoBrKnHVReSUTADaVkAF0oAHygA5MCiqD+NKsCjAAK4oAIxyGlo6NNww3GbSKlAAvFAA3gDkekYm6WbungC+kdpI-tGx1pkGxnDpjpZuHlSaOq1Q5dyVWTV1yVDW9sVQVLyizUOtHV3VOfYAdNwISig0cukd6arevv5BKABM4S1RMXFSMkz4KRndOXlNRa01ZuycPLzT2bWOKVVf6YJWqV2qdrDV6hJ8uM2icKuC+gMhiMxscYVNwQsllQVmsNqptABiWAACwawG4SFGE1hnT+PXmi2Wq3Wp02hKgfFEYACFKpQA

_No response_

asukaminato0721:
# Summary

Fixes #329

now threads a BindingsBuilder through NarrowOps, so expr_to_subjects can resolve data[key] by inspecting the binding that defines key.  We added helper logic that looks up the original Binding (either a NameAssign or FunctionParameter) to pull its Literal annotation via the new BindingsBuilder::get_annotation_by_idx helper. The existing literal-subscript walkers were generalized to accept resolver callbacks, and the resolver backed by annotations is used everywhere in the binder.

gained literal_string_key_from_expr, letting expression inference recover literal keys from either syntax or inferred type.  subscript_infer now calls this helper so TypeInfo::at_facet is entered even when the index expression is a Literal-typed variable.

switches Binding::AssignToSubscript over to the same resolver (via the newly exported _with_resolver helpers) so assignments such as data[key] = [] update the correct TypedDict facet even when key is a literal-typed variable.

other assertion helpers were updated to pass the builder into from_single_narrow_op, ensuring unit-test helpers keep benefiting from the new literal detection.

# Test Plan

add tests

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -874,22 +874,26 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     ) -> TypeInfo {
         match op {
             NarrowOp::Atomic(subject, AtomicNarrowOp::HasAttr(attr)) => {
-                let base_ty = match subject {
-                    Some(facet_subject) => {
-                        self.get_facet_chain_type(type_info, &facet_subject.chain, range)
-                    }
+                let resolved_chain = match subject {
+                    Some(facet_subject) => match self.resolve_facet_subject_chain(facet_subject) {
+                        Some(chain) => Some(chain),
+                        None => return type_info.clone(),
+                    },
+                    None => None,
+                };
+                let base_ty = match resolved_chain.as_ref() {
+                    Some(chain) => self.get_facet_chain_type(type_info, chain, range),
                     None => self.force_for_narrowing(type_info.ty()),
                 };
                 // We only narrow the attribute to `Any` if the attribute does not exist
                 if !self.has_attr(&base_ty, attr) {
-                    let attr_facet = FacetKind::Attribute(attr.clone());
-                    let facets = match subject {
-                        Some(facet_subject) => {
-                            let mut new_facets = facet_subject.chain.facets().clone();
-                            new_facets.push(attr_facet);
+                    let facets = match resolved_chain.as_ref() {
+                        Some(chain) => {
+                            let mut new_facets = chain.facets().clone();
+                            new_facets.push(FacetKind::Attribute(attr.clone()));
                             new_facets
                         }
-                        None => Vec1::new(attr_facet),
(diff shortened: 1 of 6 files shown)


yangdanny97 (changes requested):
I'm going to request changes on this because the way we're trying to resolve which values are literals at the bindings level is a little too hacky.

One of the challenges in our current exports -> bindings -> solving architecture is that the bindings stage can't access resolved types, which are only available during the solving stage. Getting around that is often difficult and hacky, and I'd like to work within the framework unless absolutely necessary.

Inspecting initial values and annotations at the bindings level can only get so far, and there are a lot of practical cases that it wouldn't be able to handle.

I would prefer an approach that instead allows us to create narrows for arbitrary expressions in subscripts (or even just literal expressions & names), and only creates the facet narrowing if the expression resolves to a literal str/int.

yangdanny97:
Thanks for making the changes, I'll review in detail once I have time. It will be after https://github.com/facebook/pyrefly/pull/1847, since there will be merge conflicts.

yangdanny97:
Sorry this took so long, I was investigating a possible regression but it turns out it was pre-existing

https://github.com/facebook/pyrefly/issues/2027

PR #1851, opened 2025-12-15
Fix: unbound-name now considers range() and non-empty literals #1706

### LINKED ISSUE #1706: `unbound-name` does not consider `range()` with an integer correctly
### Describe the Bug

```python
for _ in range(3):
    work = 1

print(work)
```

```
ERROR sandbox.py:4:7-11: `work` may be uninitialized [[unbound-name](https://pyrefly.org/en/docs/error-kinds/#unbound-name)]
```

Python: 3.12

If the argument passed to `range` can be determined to be above `0` `work` should not be considered `unbound`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4uATgAQD6NE6NVGA5jABQDMAlIgB1mNEQHdqAaxoBeGgEYhQ4lSYAXTuKoTeIADQgArquhwS5RCADENAKrGoEVaRpgD6AMbHc6OIqwwwF2oAW1RVOnQDYOwYKk58REZ0VV4aAFoAPho4VSpBYREqGFUDKmYwARAAOSiYvJpgfABfSqE9EDIisChSQlVcYKgKawAFUi6e7IwcAhp3b0g2UrCIb0IhawBlGBgaAAtVVWI4RAB6U86AnsJqNlOYdFPMXHc4U-n0ReWvR6DaVAAbqhoKhsLA5gsIEtWD8aLhiD9TEpSKo9t40gDYnBVsxZJVuIQ5AAmVroEBNfSoTwQTEAMWgMAoaCweCIZHJQA

_No response_

AryanBagade:
# Summary
For-loops iterating over definitely non-empty iterables now correctly recognize that variables assigned in the loop body will always be defined.
This adds special case handling for:
  - `range(N)` where N is a positive integer literal (e.g., `range(3)`)
  - Non-empty list literals (e.g., `[1, 2, 3]`)
  - Non-empty tuple literals (e.g., `(1, 2, 3)`)
  - Non-empty set literals (e.g., `{1, 2, 3}`)
  
  Before this change:
  ```python
  for _ in range(3):
      work = 1
  print(work)  # ERROR: `work` may be uninitialized
  ```

  After this change:
  ```python
  for _ in range(3):
      work = 1
  print(work)  # No error - loop definitely runs at least once
  ```

Fixes #1706

# Test Plan
- Added 9 new test cases in flow_branching.rs
- All Test Cases Passed
- Manually tested in the playground

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -638,9 +638,10 @@ impl FlowStyle {
                     // we have to be lax about whether boolean ops define new names
                     match merge_style {
                         MergeStyle::BoolOp => FlowStyle::Other,
-                        MergeStyle::Loop | MergeStyle::Exclusive | MergeStyle::Inclusive => {
-                            FlowStyle::PossiblyUninitialized
-                        }
+                        MergeStyle::Loop
+                        | MergeStyle::LoopDefinitelyRuns
+                        | MergeStyle::Exclusive
+                        | MergeStyle::Inclusive => FlowStyle::PossiblyUninitialized,
                     }
                 }
             }
@@ -2334,6 +2335,11 @@ enum MergeStyle {
     /// This is a loopback merge for the top of a loop; the base flow is part of the
     /// merge.
     Loop,
+    /// This is a loopback merge for a loop that definitely runs at least once
+    /// (e.g., `for _ in range(3)` or `for _ in [1, 2, 3]`). The base flow is NOT
+    /// counted as a branch for the purpose of determining if names are always defined,
+    /// since the loop body will always execute at least once.
+    LoopDefinitelyRuns,
     /// This is a fork in which the current flow should be discarded - for example
     /// the end of an `if` statement with an `else` branch.
     ///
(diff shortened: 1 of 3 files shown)


pyrefly/lib/binding/scope.rs
```diff
@@ -2443,10 +2449,19 @@ impl<'a> BindingsBuilder<'a> {
...
+            .as_ref()
+            .map_or(false, |b| b.value.is_some());
         // If this is a loop, we want to use the current default in any phis we produce,
-        // and the base flow is part of the merge.
-        let loop_prior = if matches!(merge_style, MergeStyle::Loop)
-            && let Some(base) = merge_item.base
+        // and the base flow is part of the merge for type inference purposes.
+        let loop_prior = if matches!(
+            merge_style,
+            MergeStyle::Loop | MergeStyle::LoopDefinitelyRuns
```
stroxler, line 2463: For future reference - this match happens a number of times, it can be useful to extract into a `MergeStyle.is_loop` helper. Don't worry about changing it, this is fine and I have a fix-forward internally, just figured it would be worth pointing out.

pyrefly/lib/test/flow_branching.rs
```diff
@@ -1402,3 +1402,105 @@ for _ in []:
         continue
     "#,
 );
+
```
stroxler, line 1405: For future reference, there'a  `flow_looping.rs` test suite that's a little better as a permanent home for this - Don't worry about changing this either, it's fine as-is and I have a fix-forward internally, just figured it would be worth pointing out.

grievejia (approved):

PR #1852, opened 2025-12-15
fix: highlighting for shadow builtins

### LINKED ISSUE #1849: When declaring a variable that shadows a built-in, it should not be highlighted as a built-in.
### Describe the Bug

Image

In the first assignment to `input` the variable is yellow

### Sandbox Link

_No response_

_No response_

shruti2522:
# Summary

Fixes #1849 

fixes the bug where variables shadowing built-ins were highlighted as built-ins instead of variables.

before:
pyrefly

after fix:
Screenshot 2025-12-15 174837

**Changes:**
- Ensure locally defined names are not treated as library symbols when applying the `defaultLibrary` modifier
- Always highlight assignment targets as variables in `process_stmt`
- Process the AST before built-in keys so locally defined variables take precedence

# Test Plan

added test `shadowed_builtin_not_highlighted_as_builtin` verifying shadowed built-ins are variables and actual built-ins keep the modifier

### DIFF
--- a/pyrefly/lib/state/semantic_tokens.rs
+++ b/pyrefly/lib/state/semantic_tokens.rs
@@ -228,15 +228,40 @@ impl SemanticTokenBuilder {
         key: &Key,
         definition_module: ModuleName,
         symbol_kind: SymbolKind,
+        current_module: ModuleName,
     ) {
         let reference_range = key.range();
         let (token_type, mut token_modifiers) =
             symbol_kind.to_lsp_semantic_token_type_with_modifiers();
+
+        let is_import = matches!(key, Key::Import(..));
+        // check if this is a local definition key (like Key::Definition).
+        let is_local_definition_key = matches!(
+            key,
+            Key::Definition(_)
+                | Key::MutableCapture(_)
+                | Key::CompletedPartialType(_)
+                | Key::PartialTypeWithUpstreamsCompleted(_)
+                | Key::FacetAssign(_)
+                | Key::Delete(_)
+        );
+
+        let is_truly_local_definition =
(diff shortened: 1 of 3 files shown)


pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -228,15 +228,40 @@ impl SemanticTokenBuilder {
...
+            key,
+            Key::Definition(_)
+                | Key::MutableCapture(_)
+                | Key::CompletedPartialType(_)
+                | Key::PartialTypeWithUpstreamsCompleted(_)
+                | Key::FacetAssign(_)
+                | Key::Delete(_)
+        );
+
+        let is_truly_local_definition =
```
yangdanny97, line 249: can you please explain why we need is_truly_local_definition + is_local_definition, and what the difference is between them?

I feel like the conditionals here are a little too complex. Ideally, this would be a very small change that doesn't add the DEFAULT_LIBRARY modifier if the name is on the LHS of some assignment

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -303,7 +328,7 @@ impl SemanticTokenBuilder {
                 x.recurse(&mut |x| self.process_stmt(x));
             }
             Stmt::Assign(assign) => {
-                if self.is_disabled(assign.range()) {
+                if !self.is_disabled(assign.range()) {
```
yangdanny97, line 331: could you please explain this change? was it a bug before?

yangdanny97:
@grievejia rewrote semantic highlighting to use the AST, so the issue has been fixed

PR #1857, opened 2025-12-15
Update uv tab label in installation guide

zanieb:
# Summary

👋 Please use our recommended style for uv's name to avoid spreading confusion. Thanks!

# Test Plan

I didn't test this since it seems trivial.


zanieb:
Thank you!

PR #1864, opened 2025-12-16
fix Generate intersections for isinstance/issubclass #235

### LINKED ISSUE #235: Generate intersections for isinstance/issubclass
### Describe the Bug

See sandbox link, calling `issubclass()` on a `TypeVar` bound to a `type[object]` does not correctly specialize to that subclass inside an `if` statement

### Sandbox Link

https://pyrefly.org/sandbox/?code=JYWwDg9gTgLgBDAnmYA7A5gKEwYwDYCGAzkXAGIQQBcmcdcAJgKYBmcOAFkzgNYAURJnhYBKOAFoAfHAByEVExr1lcMMSLYAKnAC8CZGnQA6TciYA1AlD4AiTTYA0cAEYQArqgY6kYJgG0IZwArbhgAXRFsZjZObn4YKjhNMSlZeUVaemA2YBI3Z3x1PhgnCggRJRU6H0MjKCYANyYCPAB9HyZiyKrqo1jePhEgA

_No response_

asukaminato0721:
# Summary

Fixes #235

Extended narrow_issubclass so we can still refine class objects when the controlling expression is a type variable: we now try to unwrap the left-hand side, compute a safe intersection with the right-hand class info (via the new helper and allowed_type_for_typevar), and fall back to directly substituting the class constraint when untype_opt can’t peel the variable.

# Test Plan

Added a regression test that mirrors issue #235

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -158,13 +158,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     fn intersect_subclass_type(&self, left: &Type, right: &Type) -> Type {
         if let Some(allowed) = self.allowed_type_for_typevar(left) {
             let right_type_form = Type::type_form(right.clone());
-            if self.is_subset_eq(&right_type_form, &allowed) {
-                return right.clone();
-            } else {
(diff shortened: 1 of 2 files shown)


yangdanny97 (approved):

PR #1884, opened 2025-12-18
fix panic: entered unreachable code: Expected a QuantifiedValue, got @_ #1659

### LINKED ISSUE #1659: panic: entered unreachable code: Expected a QuantifiedValue, got @_
### Update

The original issue (preserved below) is fixed, but pyrefly still crashes on this:
```
class C[T: C](set[object]):
    pass
```
---

### Describe the Bug

Recently tried running on the repo below which caused the crash logs below.

Steps to reproduce:
```shell
git clone git@github.com:Orcaset/orcaset.git && cd orcaset
uv venv && uv pip install -r pyproject.toml 
source .venv/bin/activate
uvx pyrefly check
```

Output:
```shell
% uvx pyrefly check
 INFO Found `~/orcaset/pyproject.toml` marking project root, checking root directory with default configuration
[00:00:00] ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░       0/58     ERROR Thread panicked, shutting down: panicked at pyrefly/lib/alt/solve.rs:1359:30:
internal error: entered unreachable code
Backtrace:
   0: __mh_execute_header
   1: __mh_execute_header
   2: __mh_execute_header
   3: __mh_execute_header
   4: __mh_execute_header
   5: __mh_execute_header
   6: __mh_execute_header
   7: __mh_execute_header
   8: __mh_execute_header
   9: __mh_execute_header
  10: __mh_execute_header
  11: __mh_execute_header
  12: __mh_execute_header
  13: __mh_execute_header
  14: __mh_execute_header
  15: __mh_execute_header
  16: __mh_execute_header
  17: __mh_execute_header
  18: __mh_execute_header
  19: __mh_execute_header
  20: __mh_execute_header
  21: __mh_execute_header
  22: __mh_execute_header
  23: __mh_execute_header
  24: __mh_execute_header
  25: __mh_execute_header
  26: __mh_execute_header
  27: __mh_execute_header
  28: __mh_execute_header
  29: __mh_execute_header
  30: __mh_execute_header
  31: __mh_execute_header
  32: __mh_execute_header
  33: __pthread_cond_wait

PANIC Sorry, Pyrefly crashed, this is always a bug in Pyrefly itself.
PANIC Please report the bug at https://github.com/facebook/pyrefly/issues/new
```

Python 3.13.9
pyrefly 0.42.3

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1659

Classes with header type parameters now consistently emit a KeyTParams binding and defer their Class tparams to that binding. BindingTParams holds the owning ClassDefIndex so the solver can identify which class is mid-computation.

While class tparams are being computed we register a temporary placeholder copy so that any recursive reference to the class can safely call get_class_tparams without re-entering the exact computation. The guard automatically removes the placeholder once the real TParams have been validated.

scoped_type_params no longer asks the solver to resolve each TypeParameter (which caused the recursive Type::Var). It now reads the TypeParameter binding directly, builds the Quantified via the shared helper quantified_from_type_parameter, and takes the active ErrorCollector so diagnostics still flow to the correct sink. Call sites in class tparams, function tparams, and scoped type aliases were updated to pass errors.

solve_tparams and class_definition now pass the owning ClassDefIndex through to the calculator so we can key the placeholder map correctly.

# Test Plan

Added a regression test that models class C[T: C] to lock in the crash reported in #1659.

### DIFF
--- a/pyrefly/lib/binding/class.rs
+++ b/pyrefly/lib/binding/class.rs
@@ -129,7 +129,6 @@ impl<'a> BindingsBuilder<'a> {
         self.scopes.push(Scope::annotation(x.range));
 
         let has_scoped_type_params = x.type_params.is_some();
-        let scoped_type_params_clone = x.type_params.clone();
         let scoped_type_param_names = x
             .type_params
             .as_mut()
@@ -309,10 +308,9 @@ impl<'a> BindingsBuilder<'a> {
         // Insert a `KeyTParams` / `BindingTParams` pair if there are scoped type parameters or legacy
         // tparams that require binding.
         let legacy_tparams = legacy_tparam_collector.lookup_keys();
-        let needs_tparams_binding = has_scoped_type_params || !legacy_tparams.is_empty();
-        let tparams_require_binding = needs_tparams_binding;
-        if needs_tparams_binding {
-            let scoped_type_params = scoped_type_params_clone.clone();
+        let tparams_require_binding = has_scoped_type_params || !legacy_tparams.is_empty();
+        if tparams_require_binding {
+            let scoped_type_params = x.type_params.clone();
             self.insert_binding(
                 KeyTParams(class_indices.def_index),
                 BindingTParams {


rchen152:
Thanks, I've put this up for internal review.

PR #1885, opened 2025-12-18
impl extract field, final part of #746

### LINKED ISSUE #746: Extraction-related and Copy/Move refactor support

asukaminato0721:
# Summary

Fixes #746

Introduced extract_field_code_actions that validates single-expression selections, finds the surrounding instance/class method, rejects static methods and locals (via find_definition), then inserts a uniquely named class field plus a receiver.field replacement.

Shared helpers (split_selection, is_exact_expression, line_indent_and_start) are now pub(super) so both variable/field extractors reuse the same trimming and indentation logic; extract_function now imports that helper instead of duplicating it.

Wired the new quick fix through the LSP transaction API and server dispatcher so field actions appear alongside the existing variable/function refactors.

# Test Plan

Added LSP harness helpers and three coverage tests describing the happy path, rejection of method-local dependencies, and classmethod behavior.

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/extract_field.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/extract_field.rs
@@ -11,10 +11,8 @@ use pyrefly_build::handle::Handle;
 use ruff_python_ast::Expr;
 use ruff_python_ast::ExprContext;
 use ruff_python_ast::ModModule;
-use ruff_python_ast::Parameters;
 use ruff_python_ast::Stmt;
 use ruff_python_ast::StmtClassDef;
-use ruff_python_ast::StmtFunctionDef;
 use ruff_python_ast::helpers::is_docstring_stmt;
 use ruff_python_ast::visitor::Visitor;
 use ruff_text_size::Ranged;
@@ -22,9 +20,11 @@ use ruff_text_size::TextRange;
 use ruff_text_size::TextSize;
 
 use super::extract_function::LocalRefactorCodeAction;
-use super::extract_variable::is_exact_expression;
-use super::extract_variable::line_indent_and_start;
-use super::extract_variable::split_selection;
+use super::extract_shared::first_parameter_name;
+use super::extract_shared::function_has_decorator;
+use super::extract_shared::is_exact_expression;
+use super::extract_shared::line_indent_and_start;
+use super::extract_shared::split_selection;
 use crate::state::lsp::FindPreference;
 use crate::state::lsp::Transaction;
 
@@ -125,7 +125,7 @@ fn find_field_context_in_class(
     for stmt in &class_def.body {
         match stmt {
             Stmt::FunctionDef(function_def) if function_def.range().contains_range(selection) => {
-                if is_static_method(function_def) {
+                if function_has_decorator(function_def, "staticmethod") {
                     return None;
                 }
                 let receiver_name = first_parameter_name(&function_def.parameters)?;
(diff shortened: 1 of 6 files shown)


yangdanny97:
You're on a roll this week! :D

pyrefly/lib/state/lsp/quick_fixes/extract_field.rs
```diff
@@ -0,0 +1,269 @@
...
+        .first()
+        .map(|param| param.name().id.to_string())
+}
+
+fn is_static_method(function_def: &StmtFunctionDef) -> bool {
+    function_def
+        .decorator_list
+        .iter()
+        .any(|decorator| decorator_matches_name(&decorator.expression, "staticmethod"))
+}
```
kinto0: this is very similar to is_static_or_class_method from extract_function.rs. could we reuse these?

pyrefly/lib/state/lsp/quick_fixes/extract_field.rs
```diff
@@ -0,0 +1,269 @@
...
+
+fn decorator_matches_name(decorator: &Expr, expected: &str) -> bool {
```
kinto0: duplicate of decorator_matches_name from extract_function.rs. can we reuse them?

pyrefly/lib/state/lsp/quick_fixes/extract_variable.rs
```diff
@@ -64,7 +64,7 @@ pub(crate) fn extract_variable_code_actions(
     Some(vec![action])
 }
 
-fn split_selection<'a>(
+pub(super) fn split_selection<'a>(
```
kinto0: instead of extract field depending on extract variable, can these methods be moved into a shared extraction module? (along with the decorator_matches_name duplication above)

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -170,6 +170,43 @@ fn assert_no_extract_action(code: &str) {
     );
 }
 
+fn compute_extract_field_actions(
```
kinto0: nit: can this be in it's own test module?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -814,6 +851,83 @@ def sink(values):
...
+    def make(cls):
+        return (
+            # EXTRACT-START
+            cls.extracted_field
+            # EXTRACT-END
+        )
+"#;
+    assert_eq!(expected.trim(), updated.trim());
+}
+
```
kinto0, line 976: could you add a test for nested classes?

pyrefly/lib/state/lsp/quick_fixes/extract_field.rs
```diff
@@ -0,0 +1,269 @@
...
+    match decorator {
+        Expr::Name(identifier) => identifier.id.as_str() == expected,
+        Expr::Attribute(attribute) => attribute.attr.as_str() == expected,
+        Expr::Call(call) => decorator_matches_name(call.func.as_ref(), expected),
+        _ => false,
+    }
+}
+
+struct IdentifierRef {
```
kinto0, line 169: nit: should we reuse identifierref from the other refactors?

asukaminato0721: it has an extra field `synthetic_load: bool`, so I choose to "dup" here.

yangdanny97 (approved):

PR #1902, opened 2025-12-19
fix Support types like pydantic.types.PositiveInt #1451 able to check call site

asukaminato0721:
# Summary

part of #1451

Added call-time constraint checking for BaseModel fields: after building the dataclass metadata we now collect each field’s gt/ge/lt/le bounds, map positional slots plus aliases back to field names, infer the argument literals, and report BadArgumentType when a literal violates its declared range, so cases like Model(x=0) or Model(x=15) are caught at the callsite.

Hooked the new checker into AnswersSolver::construct_class, so every BaseModel instantiation is validated immediately after `__init__` analysis; since we now reuse the class metadata, this avoids re-fetching it and ensures the diagnostics fire for both positional and keyword arguments.

Exposed iter_fields from the dataclass helper so the Pydantic logic can share the same ordering rules, keeping kw-only handling consistent with the existing initializer synthesis.

fix https://github.com/facebook/pyrefly/pull/1458/changes#r2487290587

# Test Plan

Updated the regression test so the `Field(gt=0, lt=10)` example now expects errors for the invalid calls, reflecting the fix for issue.

### DIFF
--- a/pyrefly/lib/alt/class/pydantic.rs
+++ b/pyrefly/lib/alt/class/pydantic.rs
@@ -58,6 +58,7 @@ use crate::types::class::Class;
 use crate::types::types::Type;
 
 fn int_literal_from_type(ty: &Type) -> Option<&LitInt> {
+    // We only currently enforce range constraints for literal ints.
     match ty {
         Type::Literal(Lit::Int(lit)) => Some(lit),
         _ => None,
@@ -94,6 +95,12 @@ impl PydanticRangeConstraints {
 struct PydanticParamConstraint {
     field_name: Name,
     constraints: PydanticRangeConstraints,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+enum PydanticParamKey {
+    Position(usize),
+    Name(Name),
 }
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/class/pydantic.rs
```diff
@@ -53,6 +57,45 @@ use crate::error::context::ErrorInfo;
 use crate::types::class::Class;
 use crate::types::types::Type;
 
+fn int_literal_from_type(ty: &Type) -> Option<&LitInt> {
```
rchen152, line 60: Can you merge this with https://github.com/facebook/pyrefly/blob/2276094063b6126db59e20608aa380aa451b9836/pyrefly/lib/alt/class/pydantic.rs#L417-L424?

pyrefly/lib/test/pydantic/field.rs
```diff
@@ -21,8 +21,8 @@ class Model(BaseModel):
    x: int = Field(gt=0, lt=10)
 
 Model(x=5) 
-Model(x=0)  
-Model(x=15)
+Model(x=0)  # E: Argument value `Literal[0]` violates Pydantic `gt` constraint `Literal[0]` for field `x`
```
rchen152, line 23: The `bug = ...` at line 16 can be deleted now.

rchen152, line 23: This needs much more test coverage. Off the top of my head, there should be tests for the remaining comparison operators, positional arguments, kw-only parameters, init_by_name, and init_by_alias.

pyrefly/lib/alt/class/pydantic.rs
```diff
@@ -499,4 +542,158 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                self.emit_pydantic_argument_constraint(&value_ty, info, kw.range, errors);
+            }
+        }
+    }
+
+    fn collect_pydantic_constraint_params(
+        &self,
+        cls: &Class,
+        dataclass: &DataclassMetadata,
+    ) -> Option<(Vec<Option<Name>>, SmallMap<Name, PydanticParamConstraint>)> {
```
rchen152: I think you can simplify this a bit by returning a single map from parameter indices and names (you'd probably want a two-variant enum to represent these) to PydanticParamConstraint. Then `check_pydantic_argument_range_constraints` can use `.get` to fetch the corresponding constraint for both args and keywords, rather than having separate positional slots.

rchen152 (approved):
Looks good to me, thank you!

PR #1908, opened 2025-12-20
fix Infer SQLAlchemy mapped_column type #1600

### LINKED ISSUE #1600: Infer SQLAlchemy mapped_column type
### Describe the Bug

In SQLalchemy, the mapped_column will return Mapped[Any], even when I set the type like sa.String . So I think if this can be inferred automatically will save a lot of effort.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1600

Added a dedicated FunctionKind::SqlAlchemyMappedColumn plus a SpecialExport::SqlAlchemyMappedColumn / ClassKind match so that references to sqlalchemy.orm.mapped_column (and the re-exported MappedColumn class) are treated as dataclass-style field specifiers instead of opaque callables.

Introduced a special-call path for `mapped_column(...)` that inspects the supplied column type (positional or type_= keyword), walks the `TypeEngine[T]` hierarchy to recover the Python type argument, and rewrites the returned descriptor so its generic argument is that inferred type.

# Test Plan

Added a regression test with a minimal SQLAlchemy stub package; the test asserts that reveal_type now shows MappedColumn[str] / MappedColumn[int] when using mapped_column(String()) and mapped_column(Integer).


rchen152:
Hey @asukaminato0721 - just want to give you a heads up that a lot of our team is out for the rest of the year, so it may take us a little longer than usual to get to your PRs. Thanks for all the excellent work, and happy holidays!

rchen152 (changes requested):

crates/pyrefly_types/src/callable.rs
```diff
@@ -860,6 +862,10 @@ impl FunctionKind {
             ("typing" | "typing_extensions", None, "disjoint_base") => Self::DisjointBase,
             ("numba.core.decorators", None, "jit") => Self::NumbaJit,
             ("numba.core.decorators", None, "njit") => Self::NumbaNjit,
+            ("sqlalchemy.orm", None, "mapped_column")
+            | ("sqlalchemy.orm._orm_constructors", None, "mapped_column") => {
+                Self::SqlAlchemyMappedColumn
+            }
```
grievejia: Would you mind elaborating why 2 entires are added here? I'd usually expect `sqlalchemy.orm.mapped_column` to be the canonical name and the other one just an implementation detail that's ok to drop, unless things get moved around between Python versions or something like that.

crates/pyrefly_types/src/class.rs
```diff
@@ -170,7 +170,9 @@ impl ClassKind {
             ("enum", "property") => Self::Property(name.clone()),
             ("enum", "member") => Self::EnumMember,
             ("enum", "nonmember") => Self::EnumNonmember,
-            ("dataclasses", "Field") => Self::DataclassField,
+            ("dataclasses", "Field")
+            | ("sqlalchemy.orm", "MappedColumn")
+            | ("sqlalchemy.orm.properties", "MappedColumn") => Self::DataclassField,
```
grievejia: Two things:
- Could you check if this change is actually necessary? SQLAlchemy uses `dataclass_transform`, so I would've expected `MappedColumn` to already be treated as a field specifier even without this special-casing. If tests failed without this, there may be a gap in Pyrefly's `dataclass_transform` implementation that should be investigated separately.
- Similar to above, why two module names?

pyrefly/lib/export/special.rs
```diff
@@ -71,6 +71,7 @@ pub enum SpecialExport {
     BuiltinsFrozenset,
     BuiltinsFloat,
     Deprecated,
+    SqlAlchemyMappedColumn,
```
grievejia, line 78: Can you point me to where this is used? I was not able to find any.

pyrefly/lib/alt/special_calls.rs
```diff
@@ -277,6 +279,152 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+    fn sqlalchemy_mapped_column_python_type(&self, call: &ExprCall) -> Option<Type> {
+        // Check up to the first two positional arguments for a SQLAlchemy TypeEngine-derived column type.
+        for expr in call.arguments.args.iter().take(2) {
+            if let Some(ty) = self.python_type_from_type_engine_expr(expr) {
+                return Some(ty);
+            }
+        }
+        for keyword in &call.arguments.keywords {
+            if let Some(arg) = &keyword.arg
+                && matches!(arg.as_str(), "__type_pos" | "type_")
```
grievejia: __type_pos seems to be [posonly](https://github.com/sqlalchemy/sqlalchemy/blob/a96c607cc53f16b364d5025c5a5de1470661d3c9/lib/sqlalchemy/orm/_orm_constructors.py#L103)

pyrefly/lib/alt/special_calls.rs
```diff
@@ -277,6 +279,152 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                }
+                "primary_key" => {
+                    if let Some(value) = Self::expr_bool_literal(&keyword.value) {
+                        primary_key = value;
+                    }
+                }
+                _ => {}
+            }
+        }
+        if primary_key { false } else { nullable }
```
grievejia:
```suggestion
        !primary_key && nullable
```

pyrefly/lib/alt/special_calls.rs
```diff
@@ -277,6 +279,152 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                    if let Some(member_ty) = self.python_type_from_type_engine_type(member) {
+                        match &inferred {
+                            Some(existing) if existing != &member_ty => return None,
+                            None => inferred = Some(member_ty),
+                            _ => {}
+                        }
+                    }
+                }
+                inferred
+            }
```
grievejia, line 598: Union together the inferred types might be more complete?

pyrefly/lib/alt/special_calls.rs
```diff
@@ -277,6 +279,152 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            _ => None,
+        }
+    }
+
+    fn python_type_from_type_engine_class(&self, cls: &ClassType) -> Option<Type> {
```
grievejia, line 603: Using `get_mro_for_class` + `ancestors_no_object` might be cleaner and more consistent with pre-existing patterns in the codebase.

pyrefly/lib/alt/special_calls.rs
```diff
@@ -277,6 +279,152 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            if let Some(ty) = self.python_type_from_type_engine_class(base) {
+                return Some(ty);
+            }
+        }
+        None
+    }
+
+    fn is_sqlalchemy_type_engine_class(class: &Class) -> bool {
+        class.has_toplevel_qname("sqlalchemy.sql.type_api", "TypeEngine")
+    }
```
grievejia: Simple checks like this can probably be inlined

pyrefly/lib/alt/special_calls.rs
```diff
@@ -277,6 +279,152 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        });
+        ty
+    }
+
+    fn is_sqlalchemy_mapped_class(class: &Class) -> bool {
+        class.has_toplevel_qname("sqlalchemy.orm.base", "Mapped")
+            || class.has_toplevel_qname("sqlalchemy.orm.base", "_MappedAnnotationBase")
+            || class.has_toplevel_qname("sqlalchemy.orm.base", "_DeclarativeMapped")
+            || class.has_toplevel_qname("sqlalchemy.orm.properties", "MappedColumn")
+    }
```
grievejia: `mapped_column` seems to [always return a `MappedColumn`](https://github.com/sqlalchemy/sqlalchemy/blob/a96c607cc53f16b364d5025c5a5de1470661d3c9/lib/sqlalchemy/orm/_orm_constructors.py#L142). Checks on `Mapped`, `_MappedAnnotationBase`, `_DeclarativeMapped` etc. are parent classes in the hierarchy that are already implied if the `MappedColumn` check succeeds.

pyrefly/lib/test/sqlalchemy.rs
```diff
@@ -0,0 +1,120 @@
...
+class Model:
+    name = mapped_column(String())
+    quantity = mapped_column(Integer)
+    sku = mapped_column(String(), nullable=False)
+    pk = mapped_column(Integer, primary_key=True)
+
+reveal_type(Model.name)  # E: revealed type: MappedColumn[str | None]
+reveal_type(Model.quantity)  # E: revealed type: MappedColumn[int | None]
+reveal_type(Model.sku)  # E: revealed type: MappedColumn[str]
+reveal_type(Model.pk)  # E: revealed type: MappedColumn[int]
```
grievejia: Use of `assert_type` would be preferable over `reveal_type`.

grievejia (changes requested):
Sorry for the late response! @rchen152 was kind enough to perform an in-depth review of the PR. I'm just forwarding her comments here.

asukaminato0721:
seems there are some problems with MappedColumn and Mapped.

Issue #1835, opened 2025-12-12
Support non-ipynb notebook formats


yangdanny97:
### Describe the Bug

For example, Marimo's .py notebooks

Need to figure out what happens if a file is simultaneously open as a notebook & as a regular file. Right now we don't support that because open files are keyed on the path.

michaelfortunato:
Cool ! So would this in theory apply to quarto mark down notebooks? Its a decent format. I use otter.nvim to get pyrefly to understand the .qmd files currently.

yangdanny97:
@michaelfortunato ooh yeah, markdown notebooks would be another good one to support

looks like the format is here: https://pandoc.org/MANUAL.html#jupyter-notebooks

For neovim otter sounds like a good workaround for now. FWIW, that's what Pyrefly does internally to jupyter notebooks (squashing all the cells into a single virtual file & re-mapping the position), we just don't currently sync files with `.md` extension.

yangdanny97:
Marimo support is documented here: https://docs.marimo.io/guides/editor_features/language_server/#pyrefly

kinto0:
closing because https://github.com/jackulau/pyrefly/commit/bdab7a022ab0e98ee76ec1f962b3e7681af03bdc claims it is complete

PR #1909, opened 2025-12-21
feat: Support non-ipynb notebook formats (#1835) 

jackulau:
**Summary**                                                                                                                             
                                                                                                                                                
  - Add support for non-Jupyter notebook formats (e.g., Marimo .py notebooks, Quarto .qmd)                                                      
  - Protect notebooks from text document operations when the same file could be opened as both notebook and text file                           
  - Skip textDocument/didOpen, textDocument/didChange, and textDocument/didClose when the file is already open as a notebook                    
                                                                                                                                                
**Test plan**                                                                                                                              
                                                                                                                                                
  - Added test_notebook_non_ipynb_extension - verifies .py files work correctly as notebooks                                                    
  - Added test_notebook_text_open_conflict - verifies text document operations are skipped when file is open as notebook                        
  - All 3223 existing tests pass                                                                                                                
  - All 20 notebook tests pass                                                                                                                  
                                                                                                                                                
  adds on to https://github.com/facebook/pyrefly/issues/1835

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -684,6 +684,16 @@ impl Server {
         None
     }
 
+    /// Check if a path is currently open as a notebook.
+    /// This is used to handle the case where a file could be opened as both
+    /// a notebook (e.g., via Marimo or Jupyter) and as a regular text file.
+    fn is_open_as_notebook(&self, path: &PathBuf) -> bool {
+        self.open_files
+            .read()
+            .get(path)
+            .is_some_and(|file| file.is_notebook())
+    }
+
     fn extract_request_params_or_send_err_response<T>(
         &self,
         params: Result<T::Params, serde_json::Error>,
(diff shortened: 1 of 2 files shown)


yangdanny97:
Q: what happens if a file is open as a text document first, then as a notebook?

jackulau:
When a file opens as text first, then as notebook, the notebook opening will "upgrade" the file state. The notebook representation overwrites the text file entry in open_files and any subsequent text document operations (didChange, didClose) are ignored since is_open_as_notebook() now returns true

yangdanny97:
Thanks for clarifying. I have a different change I'm playing around with that lets someone open a file as both a source and a notebook. Given that it will conflict with this change, once I'm done prototyping that one I'll compare and see which one works better

kinto0:
looks like a commit did something similar here?
https://github.com/jackulau/pyrefly/commit/bdab7a022ab0e98ee76ec1f962b3e7681af03bdc

if it's not, feel free to reopen!

Issue #1286, opened 2025-10-14
Exhaustiveness check in match statement


sander76:
### Describe the Bug

See  [sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2o4cGHwD66GADcRAHXRyAxlEFw6AQUSEtcxcqF0AQpu3ysMMHTa4AFDUxREagD4GAlIjl0vdGqgYKACx97D3RvcLoFQRh1a3dPCMTiSlYGawByNXTXBMSo4UM40MSklPQ09INs3Ij8mNFikvCVEQZxKRFbe1cQABoQMkpzKFJCBlooCgBiOgAFUiGwEbo0LDx8SNx0SDYAV0o-CC3CORmAZRgYgIYGYjhEAHoHweHR3jYHmHQHzFwFOAeCi2O32hy2DzoYF4AkkqGgqGwsE22wgewODCOYVwxAxWzgJ3QZAYAS2AFppJQ4Ji6ABeOgyEAAZkIAEYAEwMuQgAC%2B-VQCgx0gAYtAYBRVjgCCRyNygA)

```python
from typing import assert_never

class A:...

class B:...

def go(mdl:A|B):
    match mdl:
        case A():
            print('A')
        case B():
            print('B')
        case _:
            assert_never(mdl)
```

assert_never raises an issue which it shouldn't be raising?:

`Argument `A | B` is not assignable to parameter `arg` with type `Never` in function `typing.assert_never`

yangdanny97:
Thanks for reporting! This looks similar to the issues described here: https://github.com/facebook/pyrefly/issues/400

There are some gaps with how we model match statements which means that exhaustiveness is not detected correctly.

stroxler:
Another example. from duplicate issue #1511
```
from typing import *

class Something:
    pass


def typecheck_well(items: list[Something | Exception]) -> Something:
    for item in items:
        match item:
            case Something():
                return item
            case Exception():
                raise item

def fail_to_typecheck(things: list[Something | Exception], next_thing_index: int) -> Something:
    next_thing = things[next_thing_index]
    match next_thing:
        case Something():
            return next_thing
        case Exception():
            raise next_thing
```
[...]

PR #1913, opened 2025-12-21
Detect when previous patterns were exhaustive in a trailing `_` case

ArchieBinnie:
This PR fixes an issue where Pyrefly incorrectly reported exhaustiveness errors when using `assert_never` in the default case of a `match` statement, specifically when matching against classes (e.g. `case C():`).

for https://github.com/facebook/pyrefly/issues/1286

```python

def f(x: int | str):
    match x:
        case int():
            pass
        case str():
            pass
        case _:
            assert_never(x)  # Pyrefly incorrectly reported error here
```

### Root Cause
When processing Class Patterns (e.g., `case int():`), the solver injected a conservative `Placeholder` operation:

```rust
// lib/binding/pattern.rs
// We're not sure whether the pattern matches all possible instances of a class, and
// the placeholder prevents negative narrowing from removing the class in later branches.
let placeholder = NarrowOps::from_single_narrow_op_for_subject(
    subject.clone(),
    AtomicNarrowOp::Placeholder,
    x.cls.range(),
);
narrow_for_subject.and_all(placeholder);
```

This prevented the solver from narrowing the type in the `else` branch, leaving the subject as `int | str` instead of `Never` even after all cases were covered.

### Fix
This change optimizes `MatchClass` pattern binding in `lib/binding/pattern.rs`. We now explicitly check if the class pattern has **no arguments** (positional or keyword, e.g., `case int():`). If so, we skip adding the `Placeholder`.

```rust
// lib/binding/pattern.rs
// However, if there are no arguments, it's just an isinstance check, so we don't need
// the placeholder.
if !x.arguments.patterns.is_empty() || !x.arguments.keywords.is_empty() {
    let placeholder = NarrowOps::from_single_narrow_op_for_subject(
        subject.clone(),
        AtomicNarrowOp::Placeholder,
        x.cls.range(),
    );
    narrow_for_subject.and_all(placeholder);
}
narrow_for_subject
```

I believe this is safe because, according to [Python's match statement specification](https://docs.python.org/3/reference/compound_stmts.html#class-patterns), a class pattern without arguments is semantically equivalent to an `isinstance()` check. The docs state: *"If the subject value is not an instance of name_or_attr (tested via isinstance()), the class pattern fails"* and *"If no pattern arguments are present, the pattern succeeds."* This allows the solver to correctly deduce that the type has been fully excluded.

#### Comparison with other tools
Neither `mypy` nor `Ty` report this exhaustiveness error for empty class patterns.

### Related Work

**Note**: This is a partial fix focusing on the `MatchClass` empty-argument case to resolve the immediate false positive. 

While researching this issue, I noticed [PR #1833](https://github.com/facebook/pyrefly/pull/1833) which adds an exhaustiveness reporting framework. This fix seems to be fully compatible with that PR - it improves the solver's accuracy to make exhaustiveness reporting correct for class patterns.

## Test Plan
- `lib/test/pattern_match.rs`:
    - Modified `test_negated_exhaustive_class_match` (the reproduction case) to expect success.
    - Added safety tests:
        - `test_class_match_with_args_not_exhaustive` (guarantees we don't regress on guarded/argument patterns)
        - `test_class_match_with_guard_not_exhaustive`
        - `test_class_match_with_positional_args_not_exhaustive`
- Verified all tests pass with `cargo test` (3232 passed).
- Verified formatting and linting with `cargo fmt --check` and `cargo clippy` (no new warnings).

### DIFF
--- a/pyrefly/lib/binding/pattern.rs
+++ b/pyrefly/lib/binding/pattern.rs
@@ -233,12 +233,16 @@ impl<'a> BindingsBuilder<'a> {
                     );
                     // We're not sure whether the pattern matches all possible instances of a class, and
                     // the placeholder prevents negative narrowing from removing the class in later branches.
-                    let placeholder = NarrowOps::from_single_narrow_op_for_subject(
-                        subject.clone(),
-                        AtomicNarrowOp::Placeholder,
-                        x.cls.range(),
-                    );
-                    narrow_for_subject.and_all(placeholder);
+                    // However, if there are no arguments, it's just an isinstance check, so we don't need
+                    // the placeholder.
+                    if !x.arguments.patterns.is_empty() || !x.arguments.keywords.is_empty() {
+                        let placeholder = NarrowOps::from_single_narrow_op_for_subject(
+                            subject.clone(),
+                            AtomicNarrowOp::Placeholder,
+                            x.cls.range(),
+                        );
+                        narrow_for_subject.and_all(placeholder);
+                    }
                     narrow_for_subject
                 } else {
                     NarrowOps::new()
--- a/pyrefly/lib/test/pattern_match.rs
+++ b/pyrefly/lib/test/pattern_match.rs
@@ -86,7 +86,6 @@ def foo(x: Literal['A'] | Literal['B']):
 );
 
 testcase!(
-    bug = "We currently never negate class matches; ideally we would be smarter about when the match is exhaustive",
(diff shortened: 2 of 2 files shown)


yangdanny97:
Thanks, I'll have a look at the code tomorrow. While this PR doesn't fully solve the linked issue, from reading the summary I believe this is a good change.

PR #1920, opened 2025-12-23
Forbid re-assignment of `Final` fields in constructor

fangyi-zhou:
# Summary

Previously, the check for `Final` fields in classes did not check whether the final fields are provided with an initial value. This diff adds check for whether the final fields have been initialised in their declaration, and raises an error if they are re-assigned in constructors or other methods that are permitted to modify attributes.

Fixes one failing test in the conformance testsuite.

# Test Plan

`test.py`


rchen152 (approved):
Thanks! This looks pretty good to me.

PR #1922, opened 2025-12-23
fix: Resolve TypeVars in generic TypedDict inheritance with Unpack

kitagry:
# Summary

Pyrefly incorrectly reports type errors when using `Unpack` with a `TypedDict` that has a `TypeVar` in its generic parameter. Pyright handles this correctly.

## Problem

When using `Unpack[Child[int]]` where `Child` inherits from `Base[T]`, Pyrefly failed to properly resolve type parameters through the inheritance chain. Fields inherited from the parent TypedDict retained unresolved TypeVars instead of being substituted with concrete types.

**Example:**

```python
from typing import TypedDict, Type, Generic
from typing_extensions import TypeVar, Unpack

_T = TypeVar("_T", default=str)

class Base(TypedDict, Generic[_T], total=False):
    default: _T | None

T = TypeVar('T')

class Child(Base[T], total=False):
    other: Type[T]

def test(**kwargs: Unpack[Child[int]]) -> None:
    pass

test(other=int, default=5)  # Error: Argument `Literal[5]` not assignable to `_T | None`
```

## Solution

Modified class_field_to_typed_dict_field() in typed_dict.rs to:

1. Track which class in the MRO defines each field
2. Retrieve that class's type arguments from the MRO (expressed in terms of the child's type parameter
s)
3. Apply the child's substitution to get the parent's concrete type arguments
4. Compose substitutions properly through the inheritance chain

This ensures that when Child[int] inherits from Base[T], the type parameter _T is correctly resolved to int.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -3115,7 +3115,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         })
     }
 
-    fn get_class_member_with_defining_class(
+    pub(in crate::alt::class) fn get_class_member_with_defining_class(
         &self,
         cls: &Class,
         name: &Name,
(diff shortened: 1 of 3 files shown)


yangdanny97:
Thanks for your patience here. The direction is correct, but I've modified the implementation to reuse the existing class field instantiation logic (which does something very similar to what you're doing now), as opposed to manually looking up the parent class and performing a substitution.

No action needed on your end, I'll get this reviewed and hopefully merged soon

PR #1923, opened 2025-12-23
[easy] Improve error message for assignments to final variables

fangyi-zhou:
# Summary

- Unify two different error messages depending on call sites
- Add backticks around variable names

# Test Plan
`python test.py`


yangdanny97 (approved):

PR #1924, opened 2025-12-23
fix: Pyrefly should narrow to intersection, not Never, for a TypeIs call

### LINKED ISSUE #1597: Pyrefly should narrow to intersection, not Never, for a TypeIs call
### Describe the Bug

See:
```
from typing import TypeIs, reveal_type

class A: ...

def f(x) -> TypeIs[A]:
    return True

class B: ...

def g(b: B):
    if f(b):
        reveal_type(b) # pyrefly incorrectly reveals Never
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjACScADR1KMAG4xUUAPpNiMADroNAYyio4cOgEFEdQmY0bMMMHTAAKfAEo6AWgB8g4WIDahgLqIGnTBUjAMAK6U6IKU4eqa6Dp6BgBCJmaEFljWdGx22CYpjoHRIRA29tjFQSG10nIKysL5jiDiIOEM0HAk5IggAMR0AKpdUBBMtuGJXbjocFlWFbw0qAyK6OE02DCUDiasDM7udHAMlCV1YZHRYGogAHJbOxd0wPgAvvcabSBk0mAoKRCAxaFAKEMAAqkAFA04YHAEOhaOaQNiRNYQOaZdBDADKMBgdAAFgwGMQ4IgAPRU-7WIGEXhsKkwdBUzC4LRwKko9BojGzNm2Xh0VAyVDQVDYWDI1EQdGUTFzOi4YiCnoaMgMYlzFxyShwLHRAC8dHuAGZCABGc3fdAgD7tVBaLpyABi0BgFDQWDwRDIDqAA

_No response_

tannguyencse19:
# Summary

- `AtomicNarrowOp::TypeIs` now use `distribute_over_union` return types and uses `intersect_with_fallback` per member

Fixes #1597 

# Test Plan

Using the [issue sandbox link](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjACScADR1KMAG4xUUAPpNiMADroNAYyio4cOgEFEdQmY0bMMMHTAAKfAEo6AWgB8g4WIDahgLqIGnTBUjAMAK6U6IKU4eqa6Dp6BgBCJmaEFljWdGx22CYpjoHRIRA29tjFQSG10nIKysL5jiDiIOEM0HAk5IggAMR0AKpdUBBMtuGJXbjocFlWFbw0qAyK6OE02DCUDiasDM7udHAMlCV1YZHRYGogAHJbOxd0wPgAvvcabSBk0mAoKRCAxaFAKEMAAqkAFA04YHAEOhaOaQNiRNYQOaZdBDADKMBgdAAFgwGMQ4IgAPRU-7WIGEXhsKkwdBUzC4LRwKko9BojGzNm2Xh0VAyVDQVDYWDI1EQdGUTFzOi4YiCnoaMgMYlzFxyShwLHRAC8dHuAGZCABGc3fdAgD7tVBaLpyABi0BgFDQWDwRDIDqAA) as the testcase

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -283,6 +283,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     CallTargetLookup::Ok(Box::new(CallTarget::Union(targets)))
                 }
             }
+            Type::Intersect(intersect) => {
+                let (_, fallback) = *intersect;
+                self.as_call_target_impl(fallback, quantified, dunder_call)
+            }
             Type::Any(style) => CallTargetLookup::Ok(Box::new(CallTarget::Any(style))),
             Type::TypeAlias(ta) => {
                 self.as_call_target_impl(ta.as_value(self.stdlib), quantified, dunder_call)
--- a/pyrefly/lib/test/narrow.rs
+++ b/pyrefly/lib/test/narrow.rs
@@ -918,7 +918,7 @@ def f(x: Cat | Dog):
 testcase!(
     test_typeis_union,
     r#"
-from typing import TypeIs, assert_type
+from typing import TypeIs, assert_type, reveal_type
 class A: ...
 class B: ...
 class C: ...
@@ -930,7 +930,7 @@ def f(x:  A | B | C, y: A | C):
     else:
         assert_type(x, C)
     if is_a_or_b(y):
-        assert_type(y, A)
+        reveal_type(y)  # E: (B & C) | A
     else:
         assert_type(y, C)
     "#,
@@ -1898,7 +1898,7 @@ testcase!(
     test_callable,
     r#"
 from typing import Callable
-def f(x: int | Callable[[], int]) -> int:
+def f(x: int | Callable[[], int]) -> object:
     if callable(x):
         return x()
     else:


tannguyencse19:
> Thanks, @tannguyencse19! This PR looks about right to me, but it looks like there are a couple test failures that need fixing:
> 
> * The `test::narrow::test_typeis_union` failure seems to be due to the asserted type in the test being wrong. Could you change [this line](https://github.com/facebook/pyrefly/blob/df8a540d8e328500af649e91473cc251231ba80f/pyrefly/lib/test/narrow.rs#L933) from `assert_type(y, A)` to `reveal_type(y, A)  # E: (B & C) | A`? (You'll also need to import `reveal_type`).
> 
> * The `test::narrow::test_callable` failure is due to a missing case for `Type::Intersect` in [`as_call_target_impl`](https://github.com/facebook/pyrefly/blob/df8a540d8e328500af649e91473cc251231ba80f/pyrefly/lib/alt/call.rs#L177). Implementing the Intersect case fully might be a bit tricky, but for now, I think you can get away with just using the [fallback type](https://github.com/facebook/pyrefly/blob/df8a540d8e328500af649e91473cc251231ba80f/crates/pyrefly_types/src/types.rs#L669-L671).

Okay, I will do it

tannguyencse19:
@rchen152 I fixed. Can you take a look

rchen152:
> @rchen152 I fixed. Can you take a look

Thanks! I'll take a look when I'm back in the office on Monday.

pyrefly/lib/alt/narrow.rs
```diff
@@ -672,7 +672,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     let ret =
                         self.call_infer(*call_target, &args, &kws, range, errors, None, None, None);
                     if let Type::TypeIs(t) = ret {
-                        return self.intersect(ty, &t);
+                        return self.distribute_over_union(&t, |right| {
+                            self.intersect_with_fallback(ty, right, &|| right.clone())
```
rchen152, line 676: This change turned out to be a little too noisy when I ran internal tests, so I tweaked this to fall back to Never when the lhs is a union. (No action needed on your part, just FYI.)

rchen152 (approved):
Thanks! I'll work on getting this merged.

pyrefly/lib/alt/narrow.rs
tannguyencse19: Okay

yangdanny97 (approved):

PR #1925, opened 2025-12-23
Check re-assignment of `Final` variables in less-common binding constructs

fangyi-zhou:
# Summary

This diff extends existing re-assignment checks for variables declared as `Final` to a few other binding constructs that are less common (namely unpacking assignments, context manager assignments, and loop variable assignments). Overall, we have 3 more passing conformance test cases.

# Test Plan

`python test.py`


fangyi-zhou:
Rebased

PR #1930, opened 2025-12-25
fix: use function instead of module docstring

### LINKED ISSUE #1929: Wrong documentation in signature help
### Describe the Bug

Pyright
Image

Pyrefly
Image

### Sandbox Link

_No response_

_No response_

aaron-ang:
Close #1929.

## Changes
- Use `callee_range.end()` to get docstring for the callee, instead of `start()` which gets the docstring for the module or object before the callee. For example, `np.linspace(` would previously generate signature information with documentation for `numpy` instead of `linspace`.
- Moved repeated range and module search logic from `parameter_documentation_for_callee` and `function_docstring_for_callee` into helper method.


yangdanny97:
Sorry, ignore the primer comments. Seems like there's a bug somewhere.

kinto0 (approved):
thanks for working on this!

PR #1934, opened 2025-12-25
fix: Recognize string type annotation in explicit type arguments

### LINKED ISSUE #1474: Recognize string type annotation in explicit type arguments
### Describe the Bug

```

if TYPE_CHECKING:
    class A:
        pass

class C[T]:
    pass

x = C["A"]() # unexpected not-a-type error
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVAJoAFAKIB9AMIAJMVIDSASQByAcQA66LRDCDRk2fOXrEWuhboBjKKjhw6AQTPpLbusTtwtWm17pSANoCALoubp72Puj4dAC8AYEaII7JIQAUAJQgADQgAK4M0HAk5IggAMR0AKpFUBBMdGD56FZFuOje2lgwemC8NKgMEuj5NNgwlOn4iJzoDJl0ALQAfHRwDJThlpQwDPmUrmDJKmMTW3TA%2BAC%2ByVq5IGS7YFCkhAy0UBRVIqTPr%2BsMDgCNYOpA2AchhAOoQtFUAMowGB0AAWDAYxDgiAA9Ninr1XoReGxsTB0NjMLgrHBsVYwRAIZQoR1sU1eHRUAA3VDQVDYWCg9DgyHtVy4YiikpaMgMFEdJacyZwaGuBLJADMhAAjAAmO7oEDXPKoNoQRUAMWgMAoaCweCIZENQA

_No response_

tannguyencse19:
# Summary

Fix by treating explicit generic type-argument slices as type expressions, including forward-reference strings like C["A"]

Fixes #1474

# Test Plan

```bash
cargo test -p pyrefly test_type_alias_with_string
```

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -4191,6 +4191,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         type_form_context: TypeFormContext,
         errors: &ErrorCollector,
     ) -> Type {
+        if let Expr::StringLiteral(string_literal) = x
+            && let Some(string_literal) = string_literal.as_single_part_string()
+            && let Ok(string_type_expression) = Ast::parse_type_literal(string_literal)
+        {
+            return self.expr_untype(&string_type_expression, type_form_context, errors);
+        }
         let result = match x {
             Expr::List(x)
                 if matches!(
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -50,6 +50,7 @@ use crate::binding::bindings::LegacyTParamId;
 use crate::binding::bindings::NameLookupResult;
 use crate::binding::narrow::AtomicNarrowOp;
 use crate::binding::narrow::NarrowOps;
+use crate::binding::scope::FlowStyle;
 use crate::binding::scope::Scope;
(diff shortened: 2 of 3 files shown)


grievejia:
Thanks for taking a stab at this! I’ve reviewed the code, but I don't think this approach will work given the current constraints of the binding and solving phases.

There are two main blockers:

**1. Binding Phase Assumptions**
The current implementation treats all subscripts with an identifier base as type expressions. However:
* **Complex Bases:** Non-identifier bases can still be class types requiring type expressions.
* **Value Subscripts:** Identifier bases often have non-type subscripts (e.g., `MyEnum["X"]`).

**2. Solving Phase Safety**
Fabricating expressions during the solving phase that were never examined by the binding phase is unsafe. Since these expressions lack entries in the binding table, performing a use-def lookup on them causes the type checker to panic immediately.

**The Path Forward**
A correct solution is tricky because it requires satisfying two difficult constraints:
* Only applying subscript type treatments to bases that are generic class types.
* Ensuring forward references are handled without looking up nonexistent binding entries.

Because the solution will likely be much more involved architecturally, I’m going to close this PR for now. Let's discuss the design before opening a new attempt.

PR #1936, opened 2025-12-26
fix: find submodule LSP definition

### LINKED ISSUE #1935: Missing submodule definition
### Describe the Bug

Pyrefly:
Image

Pyright:
Image

Context: LSP definition enables following token references into the file where it was defined, usually by Control-click.

I am not able to get LSP definition for `nn` submodule in `torch.nn.Linear` in Pyrefly, but am able to do so in both LSPs in `import torch.nn`.

### Sandbox Link

_No response_

_No response_

aaron-ang:
# Summary

Fixes #1935.

- In `completions_module`, create `AttrInfo` when the expected module attribute isn't exported but exists.
- In `resolve_attribute_definition`, fallback using `find_definition_for_imported_module` to get submodule text range.
- Add submodule type information for LSP highlight.

# Test Plan
New test in `pyrefly/lib/test/lsp/lsp_interaction/definition.rs` and `completion.rs`.

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -409,13 +409,16 @@ impl<'a> DefinitionsBuilder<'a> {
                 }
             }
             Stmt::ImportFrom(x) => {
-                let name = self.module_name.new_maybe_relative(
+                let maybe_imported_module = self.module_name.new_maybe_relative(
                     self.is_init,
                     x.level,
-                    x.module.as_ref().map(|x| &x.id),
+                    x.module.as_ref().map(|m| &m.id),
                 );
+                let is_relative_self_import = self.is_init
+                    && x.module.is_none()
+                    && maybe_imported_module == Some(self.module_name);
                 if self.is_init
-                    && let Some(imported_module) = name
+                    && let Some(imported_module) = maybe_imported_module
                     && let Some(submodule) =
                         implicitly_imported_submodule(self.module_name, imported_module)
                 {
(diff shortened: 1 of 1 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -1362,6 +1362,58 @@ impl<'a> Transaction<'a> {
...
+                        return self.find_attribute_definition_for_base_type(
+                            handle,
+                            preference,
+                            base_type,
+                            identifier.id(),
+                        );
+                    };
+                }
+            }
+        };
```
grievejia: Would you mind explaining / do you have a test on when this branch would trigger?

pyrefly/lib/state/lsp.rs
```diff
@@ -1362,6 +1362,58 @@ impl<'a> Transaction<'a> {
...
+    fn find_attribute_definition_fallback(
+        &self,
+        handle: &Handle,
+        covering_nodes: &[AnyNodeRef],
+        identifier: &Identifier,
+        preference: FindPreference,
+    ) -> Vec<FindDefinitionItemWithDocstring> {
+        // Extract the base name from the ExprAttribute node.
+        // Index 1 is the ExprAttribute node itself (index 0 is the identifier).
+        if let Some(AnyNodeRef::ExprAttribute(attr)) = covering_nodes.get(1)
```
grievejia: I suspect we could just make it such that `IdentifierContext::Attribute` contains the base expr directly so there's no need to reach in for `covering_nodes` and break the abstraction.

grievejia (commented):
May I ask your thoughts behind adding a new `find_attribute_definition_fallback()` function, as opposed to fixing the original `find_definition_for_attribute()` function? This is not me suggesting doing things the other way -- just trying to align my understanding of the codebase.

aaron-ang:
> May I ask your thoughts behind adding a new `find_attribute_definition_fallback()` function, as opposed to fixing the original `find_definition_for_attribute()` function? This is not me suggesting doing things the other way -- just trying to align my understanding of the codebase.

the additional logic is long (especially for a matched arm) so I moved it into a separate helper function.

pyrefly/lib/state/lsp.rs
aaron-ang: ok, I will look into this.

aaron-ang:
hi @grievejia, I did some refactoring on `IdentifierContext::Attribute` so that the `find_definition_for_attribute` function uses `base_expr` directly. This requires some modifications to `CalleeKind::Method` since it also calls that function. I'm not sure whether that will have a noticeable effect on performance.

I also modified the submodule resolution to support nested submodules. To handle import aliasing, we walk `ExprAttribute` to get the real module name.

aaron-ang:
I'm thinking of approaching it from a different angle. Currently, `nn` in `torch.nn` completions is resolved as an attribute, not a module. My guess is if I can fix the module resolution in completions, I don't have to perform module resolution in `find_definition_for_attribute`.

aaron-ang:
hi @grievejia could you take a look at this?

grievejia:
Hi @aaron-ang, apologies for the long turnaround time on this!

I really like the direction you took here. However, upon closer look, I think some of the logic is a bit too permissive, allowing for gotodef/autocomplete in cases where we'd ideally want to skip.

This actually isn't on you: after some digging, I realized our pre-existing tests (`submodule_access_test` and `deep_submodule_chain_test`) were misleading and weren't actually testing the intended behavior. I’ve spent some time cleaning those up this week and have put forward some simpler fixes that incorporate your ideas.

I’ll link this PR to those fixes so you still get credit, and I'll close this out once they land. Thank you so much for the contribution and for digging into this!

PR #1937, opened 2025-12-26
fix:(hover,definition,semantic_token): multipart module import

aaron-ang:
# Summary

- Fixed hover and go-to-definition for components of multi-part imports (e.g., `torch` in `import torch.nn` or `nn` in `from torch.nn.parallel import DataParallel`). Previously, these resolved to the full module name instead of the component under the cursor.
- Updated `find_definition` in `lsp.rs` to compute the cursor offset within the identifier and resolve to the component at that position.
- Fixed semantic token coloring so module attributes use `NAMESPACE` instead of `PROPERTY`.

# Test Plan

New tests in `pyrefly/lib/test/lsp/definition.rs`, `pyrefly/lib/test/lsp/hover.rs`, and `pyrefly/lib/test/lsp/semantic_tokens.rs`.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -1520,7 +1520,6 @@ impl<'a> Transaction<'a> {
             }) => {
                 // Build the module name for lookup based on identifier position.
                 let components = module_name.components();
-
                 let target_module_name =
                     if let Some(idx) = components.iter().position(|c| c == &identifier.id) {
                         // Identifier matches a module component.
--- a/pyrefly/lib/state/semantic_tokens.rs
+++ b/pyrefly/lib/state/semantic_tokens.rs
@@ -270,6 +270,7 @@ impl SemanticTokenBuilder {
                         SemanticTokenType::FUNCTION
                     }
                     Some(Type::ClassDef(_)) => SemanticTokenType::CLASS,
+                    Some(Type::Module(_)) => SemanticTokenType::NAMESPACE,
                     _ => SemanticTokenType::PROPERTY,
                 };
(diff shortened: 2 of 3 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -1498,14 +1498,42 @@ impl<'a> Transaction<'a> {
...
-                .map_or(vec![], |item| vec![item]),
+            }) => {
+                // Build the module name for lookup based on identifier position.
+                let components = module_name.components();
+
+                let target_module_name =
+                    if let Some(idx) = components.iter().position(|c| c == &identifier.id) {
+                        // Identifier matches a module component.
+                        ModuleName::from_parts(&components[..=idx])
+                    } else if identifier.id == module_name.as_name() {
```
grievejia: nit: Any chance we could just extract `&str` from both sides and compare them? This would avoid the need of adding a `MoudleName::as_name()` function.

aaron-ang: yes, we could use `as_str` for both structs.

kinto0 (commented):
this looks great! thank you for working on this! I wonder what happens when a module does not exist at that location:

example: clicking `foo` in `import foo.bar` where we only have foo/bar.py. 

I'll add a test for this situation when I import it.

kinto0:
failing some snapshot tests related to indexing results of go-to definition on imports, waiting on someone to return from PTO tomorrow to confirm

aaron-ang:
rebased

kinto0:
sorry again for the issue - thanks for your patience. we have fixed their tests and once I get a stamp we can merge this

PR #1938, opened 2025-12-26
fix: imports with aliases

aaron-ang:
# Summary

Previously, pyrefly wasn’t resolving or highlighting imported aliased names. To fix, we push the tokens that are aliased (contains `asname`) and resolve the imported tokens with `name_after_import` in `IdentifierContext::ImportedName`.

# Test Plan

New tests cases in:
- `pyrefly/lib/test/lsp/definition.rs`
- `pyrefly/lib/test/lsp/hover.rs`
- `pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs`
- `pyrefly/lib/test/lsp/semantic_tokens.rs`

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -780,10 +780,17 @@ impl<'a> Transaction<'a> {
             }
             Some(IdentifierWithContext {
                 identifier: _,
-                context: IdentifierContext::ImportedName { .. },
+                context:
+                    IdentifierContext::ImportedName {
+                        name_after_import, ..
+                    },
             }) => {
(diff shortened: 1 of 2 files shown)


aaron-ang:
hi @yangdanny97, is this good to merge? I think the test in `lib/test/lsp/lsp_interaction/` may be unnecessary since there is an equivalent test in`lib/test/lsp/semantic_tokens.rs`. lmk if I should remove it.

yangdanny97:
We can patch the PR before merging. Things are probably just a bit delayed because of holidays, we'll try to get it merged soon

It looks fine to me but I'll defer to @grievejia on when to merge

PR #1940, opened 2025-12-26
fix: unbound-name after assignment to singleton attribute ; unexpected unbound-name error on second walrus conditional assignment for same variable

### LINKED ISSUE #1378: unbound-name after assignment to singleton attribute
### Describe the Bug

We have a pattern where a singleton class is created to represent some data that needs to be maintained the full lifecycle of a process.

If I assign a value to an attribute of this singleton as part of a constructor for another class, future references to the singleton flag "unbound-name" error, which I don't believe makes sense.

If I remove the attribute assignment, or if I move the code into base module scope, the error is no longer flagged.

Python version: `3.14.0`
Pyrefly version: `0.38.2`

Pyrefly error observed:
```
# pyrefly check singleton.py caller.py
 WARN PYTHONPATH environment variable is set to `/myapp`. Checks in other environments may not include these paths.
ERROR `globals_inv` may be uninitialized [unbound-name]
 --> caller.py:9:9
  |
9 |         globals_inv.initialization_status = True
  |         ^^^^^^^^^^^
  |
 INFO 1 error
```

caller.py:
```
from singleton import globals_inv

class Foo: # unbound-name error is gone if __init__ contents moved out of class, into module scope!
    def __init__(self) -> None:
        globals_inv.config_file_inventory = {"foo": 1}  # unbound-name error also gone if this is commented out!
        for _ in globals_inv.config_file_inventory.values():
            if globals_inv.initialization_status:
                break
        globals_inv.initialization_status = True
```

singleton.py:
```
class GlobalInventory:
    _instance: GlobalInventory | None = None
    _initialized: bool = False

    initialization_status: bool
    config_file_inventory: dict[str, int]

    def __new__(cls) -> GlobalInventory:
        if cls._instance is None:
            cls._instance = super().__new__(cls)
        assert cls._instance is not None, "GlobalInventory instance creation failed"
        return cls._instance

    def __init__(self) -> None:
        # Avoid re-initializing the instance if already done.
        if not self._initialized:
            self.initialization_status = False
            self.config_file_inventory = {}

            GlobalInventory._initialized = True

globals_inv: GlobalInventory = GlobalInventory()
```

Is there a valid reason for the unbound-name error or is this a bug?

### Sandbox Link

_No response_

_No response_

### LINKED ISSUE #1397: unexpected unbound-name error on second walrus conditional assignment for same variable
### Describe the Bug

I have a code sequence that's doing walrus assignments of regex matches within if statements, using the same variable name for the walrus assignment, and then using the walrus-assigned variable within the body of the if statement.

Due to it being part of the if condition itself, once inside the if statement body it is my understanding, there is no way that the walrus-assigned variable could ever be unbound.

Yet, this is an outcome I see with a specific code sequence.

Interestingly I'm doing two walrus assignments, if I remove the first once, the error goes away.
And if I keep the first one but slightly tweak the if condition of the second one, it also goes away.

Full details are below.

Python code in question:
```
from re import compile

interface_re = compile(r"^foo")
ipv4_re = compile(r"bar$")
line = str()

if match := interface_re.match(line):  # unbound-name error gone if this is removed
    pass

# if match := ipv4_re.search(line):
if line and (match := ipv4_re.search(line)):  # unbound-name error also gone if above alternate if is used instead
    print(match)
```

Relevant software versions:
```
# pyrefly --version
pyrefly 0.39.1

# python3 --version
Python 3.14.0
```

Pyrefly error:
```
# pyrefly check walrus.py
 WARN PYTHONPATH environment variable is set to `/myapp`. Checks in other environments may not include these paths.
ERROR `match` may be uninitialized [unbound-name]
  --> walrus.py:12:11
   |
12 |     print(match)
   |           ^^^^^
   |
 INFO 1 error
```

### Sandbox Link

_No response_

_No response_

tannguyencse19:
# Summary

Updated the flow-merge logic so names that are present in all flows (even if only via attribute “narrows”) are treated as defined after a loop

fix #1378 
fix #1397 

# Test Plan

```bash
cargo test -p pyrefly test_attribute_assignment_to_imported_singleton_in_method
cargo test -p pyrefly test_walrus_reuse_name_in_if_condition
```

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -2573,12 +2573,6 @@ impl<'a> BindingsBuilder<'a> {
             MergeStyle::Loop => n_branch_flow_infos == n_branches.saturating_sub(1),
             _ => n_branch_flow_infos == n_branches,
         };
-        let has_uninitialized_style = styles.iter().any(|style| {
-            matches!(
-                style,
-                FlowStyle::Uninitialized | FlowStyle::PossiblyUninitialized
-            )
-        });
         // For LoopDefinitelyRuns, a name is always defined if:
         // - It was defined before the loop (base_has_binding), OR
         // - It's defined in all loop body branches (since the loop definitely runs at least once).
@@ -2588,7 +2582,6 @@ impl<'a> BindingsBuilder<'a> {
         let this_name_always_defined = match merge_style {
             MergeStyle::Loop => base_has_binding && is_name_exists_in_all_branch_flow,
             MergeStyle::LoopDefinitelyRuns => base_has_binding || is_name_exists_in_all_branch_flow,
-            MergeStyle::BoolOp if has_uninitialized_style => false,
             MergeStyle::BoolOp => false,
             _ => is_name_exists_in_all_branch_flow,
         };


pyrefly/lib/binding/scope.rs
```diff
@@ -2572,9 +2571,14 @@ impl<'a> BindingsBuilder<'a> {
         // - It was defined before the loop (base_has_value), OR
         // - It's defined in all loop body branches (since the loop definitely runs at least once)
         // For regular loops and other merges, a name is always defined if it's in all branches.
+        let is_name_exists_in_all_branch_flow = match merge_style {
```
yangdanny97, line 2572: This block of code was inserted between the comment above & the piece of code it describes below. I think it should be moved above the comment


The comment might also need to be updated since you made changes below

yangdanny97:
This says #1397 is fixed, but I don't see the test cases for that. Do you mind adding the test case from the original issue as well as the one from the comment below?

pyrefly/lib/binding/scope.rs
```diff
@@ -2512,6 +2512,7 @@ impl<'a> BindingsBuilder<'a> {
         let n_branch_flow_infos = flow_infos.len();
         // Track if base has a value for this name (for LoopDefinitelyRuns init check)
         let base_has_value = merge_item.base.as_ref().is_some_and(|b| b.value.is_some());
+        let base_has_name = merge_item.base.is_some();
```
yangdanny97: Are we able to do this by modifying base_has_value? Perhaps instead of `b.value.is_some()` we should check taht `b.value.is_some() || b.narrow.is_some()` and if one of them is Some, its `style` shouldn't be `FlowStyle::Uninitialized`

yangdanny97 (changes requested):
Thanks for the PR, back to you with some comments

I'm not 100% sure on the logic either, but see if you can get it to work without adding `base_has_name`. If we absolutely have to we can, but it might be simpler if we don't

tannguyencse19:
@yangdanny97 Okay. I will try

pyrefly/lib/binding/scope.rs
tannguyencse19: I pushed new commit: https://github.com/facebook/pyrefly/pull/1940/commits/cdbed60c3002d62500797411c04c4f300c3f3fba

---

I moved the new logic so it sits directly under the comment it describes, and updated the comment to match the final behavior

tannguyencse19:
> This says #1397 is fixed, but I don't see the test cases for that. Do you mind adding the test case from the original issue as well as the one from the comment below?

I pushed new commit: https://github.com/facebook/pyrefly/pull/1940/commits/cdbed60c3002d62500797411c04c4f300c3f3fba

---

I added the testcase

pyrefly/lib/binding/scope.rs
tannguyencse19: I pushed new commit: https://github.com/facebook/pyrefly/pull/1940/commits/cdbed60c3002d62500797411c04c4f300c3f3fba

---

Yes — I removed base_has_name and switched to a single `base_has_binding = base.value.is_some() || base.narrow.is_some()` so loop merges treat names that only have narrows as still present/defined.

yangdanny97 (changes requested):
From the mypy primer output, it sounds like we're getting rid of a bunch of `may be uninitialized` errors (good) but adding a lot of other errors (bad/not expected).

Could you please look into what's happening here? I don't see why we would be getting these new errors

yangdanny97:
This issue is likely related: https://github.com/facebook/pyrefly/issues/2027

pyrefly/lib/binding/scope.rs
```diff
@@ -2568,13 +2569,28 @@ impl<'a> BindingsBuilder<'a> {
...
+        // - It's defined in all loop body branches (since the loop definitely runs at least once).
+        // For Loop, the name must exist in the base flow and in all loop body branches.
+        // For BoolOp, avoid propagating uninitialized styles across the merge.
+        // For other merges, a name is always defined if it appears in every branch flow.
         let this_name_always_defined = match merge_style {
-            MergeStyle::LoopDefinitelyRuns => base_has_value || n_branch_flow_infos == n_branches,
-            _ => n_values == n_branches,
+            MergeStyle::Loop => base_has_binding && is_name_exists_in_all_branch_flow,
+            MergeStyle::LoopDefinitelyRuns => base_has_binding || is_name_exists_in_all_branch_flow,
+            MergeStyle::BoolOp if has_uninitialized_style => false,
```
yangdanny97: this case is redundant with the one below

tannguyencse19: @yangdanny97 I fixed in my latest commit: https://github.com/facebook/pyrefly/pull/1940/commits/46590336dc4cc1fc4d85396e1c58fa43e861964b

yangdanny97:
What are your thoughts on the new mypy primer errors? Also, i see there are merge conflicts

tannguyencse19:
@yangdanny97 Sorry. But right now I need to focus on something else so I can't contribute more. Thank you for spending your time on review my pull request.

yangdanny97:
No problem, thanks for your contributions so far!

PR #1941, opened 2025-12-26
fix: Narrow typevar bounds in isinstance check

### LINKED ISSUE #1274: Narrow typevar bounds in isinstance check
### Describe the Bug

Actual:
```python


def test[T: int | str](value: T) -> T:
    if isinstance(value, int):
        reveal_type(value) # revealed type: int
        return value # This errors
    else:
        reveal_type(value)
        return value # This is ok

```

Python: 3.13

Type narrowing for typeVars seems to be wrong.
After narrowing a variable that has a typeVar as its type, Pyrefly seems to forget that it is a typeVar so, it is not compatible with the original type anymore, even though the value didn't change.

In the example above I would expect Pyrefly to narrow the type of `value` to `T & int` or to also narrow `T` to `int`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRy5mGGEbwGAbQAqiTun4AfOnAaUAugAohYgK4wdmgJR0AtAD462uXW%2BcVEOKzGGADGMJY2MAA0ugwOiF4%2BiYIiYpLMYVZQtg4Jid6CDNaU6HSZtrl0MFBwdhVJwqISUhkROSV5%2BTCFxaURciCRIGSCYFCkhAy0UBQAxHQACqQjY0YYOAR0wbjokGxFqAwQ24RycwDKMDB0ABYMDMRwiAD0T8PKY4S8bE8w6E%2BYuGCcCeWx2ED2lAORz%2BdDAvDoqCs0FQ2Fgm22u32h22dFwxGx6DgJ3QZAY122zhElACOIAvHQZCAAMyEACMTMZ-QAvoNUMFDiIAGLQGAUNBYPBEMggLlAA

tannguyencse19:
# Summary

When narrowing encounters a solver `Var` that represents a quantified type parameter, preserve it as `Type::Quantified` instead of forcing it to its bound.
This allows `isinstance` narrowing to compute an anonymous intersection like `T & int`, which remains assignable to the original `T`.

Fixes #1274

# Test Plan

```bash
cargo test -p pyrefly test_isinstance_narrow_bounded_type_param --lib
```

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -1586,7 +1586,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     ) -> Type {
         match ty {
             Type::Var(v) => {
-                if let Some(_guard) = self.recurse(*v) {
+                if let Some(quantified) = self.solver().quantified_type_for_solver_variable(*v) {
+                    quantified
+                } else if let Some(_guard) = self.recurse(*v) {
                     let forced = self.solver().force_var(*v);
                     self.force_for_narrowing(&forced, range, errors)
                 } else {
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -434,6 +434,17 @@ impl Solver {
         }
     }
 
+    pub fn quantified_type_for_solver_variable(&self, solver_variable: Var) -> Option<Type> {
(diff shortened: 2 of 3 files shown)


yangdanny97:
It seems like the linked issue is also fixed by #1864

PR #1942, opened 2025-12-27
fix: Incorrect return path identification with try/finally

### LINKED ISSUE #1091: Incorrect return path identification with try/finally

tannguyencse19:
# Summary

Update to: Only let the finally block override the try path analysis when the finally block itself always terminates, for example it always returns or always raises.

If the finally block can fall through, analyze the try body, else, and handlers as usual.

Fixes #1091

# Test Plan

```bash
cargo test test_return_try_finally_explicit_return
```

### DIFF
--- a/pyrefly/lib/binding/function.rs
+++ b/pyrefly/lib/binding/function.rs
@@ -809,27 +809,27 @@ fn function_last_expressions<'a>(
                 }
             }
             Stmt::Try(x) => {
-                // If final body is not empty, _and_ contains a return statement,
-                // process it.
-                if !x.finalbody.is_empty()
-                    && x.finalbody
-                        .iter()
-                        .any(|stmt| matches!(stmt, Stmt::Return(_)))
-                {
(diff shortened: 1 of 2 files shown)


yangdanny97:
Hmm, the linked issue was fixed by https://github.com/facebook/pyrefly/pull/1118, but I actually don't think that PR is correct. Let me discuss with the team

yangdanny97:
After discussions w/ the team, I realize I misunderstood the previous PR. That is an adequate fix for the original issue, so this PR should not be necessary - the added test case passes without it

PR #1946, opened 2025-12-28
fix: assert callable narrows too much

### LINKED ISSUE #911: `assert callable` narrows too much
### Describe the Bug

```python
from typing import *

def f(x):
    assert callable(x)
    reveal_type(x()) # Should be Any, but returns object
```

The resulting type is object, and that seems like what the type signature suggests is correct. But Pyright gets Any, which is much more likely to be what is intended.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQCYCmwUwAFAB4CUAXEVG1AIYDOX5eUAYw4AbYRwBGw8vQat2IcgDdyIgPrwE0ujQYMgA&version=3.12

_No response_

tannguyencse19:
# Summary

Fix assert callable(x) narrowing because callability does not imply an object return type

Fixes #911

# Test Plan

```bash
cargo test test_assert_callable_narrow_any
cargo test test_builtins_callable_narrow
```

### DIFF
--- a/crates/pyrefly_bundled/third_party/typeshed/stdlib/builtins.pyi
+++ b/crates/pyrefly_bundled/third_party/typeshed/stdlib/builtins.pyi
@@ -4099,7 +4099,7 @@ def breakpoint(*args: Any, **kws: Any) -> None:
     By default, this drops you into the pdb debugger.
     """
     ...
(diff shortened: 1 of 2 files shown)


tannguyencse19:
@yangdanny97 I haven't think that way. Let me take a look

PR #1949, opened 2025-12-29
fix Auto import suggestions for collections.abc should take precedence over typing #1685

### LINKED ISSUE #1685: Auto import suggestions for `collections.abc` should take precedence over `typing`

asukaminato0721:
# Summary

Fixes #1685

Addressed issue 1685 by making autoimport prefer public shim modules (like collections.abc) over their private implementations and ensuring private entries are deprioritized when a public alternative exists.

Updated to recognize dotted-name private shims (e.g., _collections_abc -> collections.abc) and to sort autoimport candidates so private imports fall behind public ones when both exist.

# Test Plan

test autoimport_prefers_public_reexport_for_dotted_private_module

### DIFF
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -1794,13 +1794,13 @@ T = Literal
 2 | T = Literal
             ^
 Completion Results:
-- (Variable) AnyOrLiteralStr: from _typeshed import AnyOrLiteralStr
-
 - (Variable) Literal: from typing import Literal
 
 - (Variable) Literal: from typing_extensions import Literal
 
 - (Variable) LiteralString: from typing import LiteralString
+
+- (Variable) AnyOrLiteralStr: from _typeshed import AnyOrLiteralStr
 
 - (Variable) StrOrLiteralStr: from _typeshed import StrOrLiteralStr
 "#


yangdanny97:
This might conflict with some in-flight changes from @javabster, so heads up that this might need to be rebased at some point in the next week or so

kinto0:
should this be reviewed (it's currently draft so we've been ignoring it)

asukaminato0721:
removed depth-based sorting in autoimport completions in lsp.rs, replacing it with a stable sort using label plus private rank to prefer public over private while keeping order consistent.

pyrefly/lib/state/lsp.rs
```diff
@@ -2569,8 +2568,14 @@ impl<'a> Transaction<'a> {
                 };
                 let auto_import_label_detail = format!(" (import {imported_module})");
 
+                let is_private_import = handle_to_import_from
+                    .module()
+                    .components()
+                    .last()
```
kinto0: i wonder if non-last components should also be considered?

pyrefly/lib/state/lsp.rs
```diff
@@ -2589,7 +2594,7 @@ impl<'a> Transaction<'a> {
                     } else {
                         None
                     },
-                    sort_text: Some(format!("4{}", depth)),
+                    sort_text: Some(format!("{name}:{private_rank}")),
```
kinto0: this is pretty confusing and isn't obvious based on lsp trace why we are sorting them. could it make sense to include metadata in the sort text?

e.g.: "private: {private_rank}"

pyrefly/lib/state/lsp.rs
```diff
@@ -2589,7 +2594,7 @@ impl<'a> Transaction<'a> {
                     } else {
                         None
                     },
-                    sort_text: Some(format!("4{}", depth)),
```
kinto0: why remove the 4?

pyrefly/lib/state/lsp.rs
kinto0: I think that's what the referenced (#1685) issue is (_collections_abc.<>) so I'm not sure this PR fixes it without this change

pyrefly/lib/state/lsp.rs
```diff
@@ -2550,14 +2550,13 @@ impl<'a> Transaction<'a> {
                 {
                     continue;
                 }
-                let depth = handle_to_import_from.module().components().len();
```
kinto0: we should also still consider depth. looks like CI didn't fail on this - would you be able to add a regression test that ensures shorter modules are preferred?

kinto0 (changes requested):
thanks for working on this! this `sort_text` stuff is getting pretty confusing. I wonder if we can make a single function that sorts based on all the inputs (private module, length, etc) for autoimports, and a single function for completions (which calls the autoimports one for the import completions)? the logic being scattered across the file makes it difficult to understand

kinto0:
looks like there are some test failures

kinto0:
still a draft. please publish when ready, thanks!

PR #1957, opened 2025-12-30
Buck2 docs

cbarrete:
# Summary

This PR adds some documentation about integrating Pyrefly with Buck2.
It also cleans up some formatting based on the existing Prettier configuration.

# Test Plan

This is just changing documentation, there is no functional impact.

I tried to run `test.py`, but

```
env: ‘fbpython’: No such file or directory
```

I didn't bother to try more since this PR makes no code changes.

### DIFF
--- a/website/docs/configuration.mdx
+++ b/website/docs/configuration.mdx
@@ -598,6 +598,142 @@ allowlist or denylist). Regular
 - Default: `true`
 - Flag equivalent: `--use-ignore-files`
 
+### `build-system`
+
+Pyrefly supports integrating into build systems to discover targets to type
+check and their dependencies. It currently natively supports
+[Buck2](https://buck2.build), as well as arbitrary build systems via custom
+queries.
+
+**Note that support for build systems is currently unstable, and breakage may
+occur without notice. Support will likely be lower priority than other issues
+for a while.**
+
+#### Buck2
+
+To configure Pyrefly to use Buck2 as a build system, add the following to your
+`pyrefly.toml`:
+
+```toml
+[build-system]
+type = "buck"
+# Optional: The isolation dir for Buck2 to use.
+isolation-dir = "pyrefly"
+# Optional: Extra flags passed to Buck2.
+extras = ["--verbose", "4"]
+```
+
+Behind the scenes, Pyrefly will run the
+`prelude//python/sourcedb/pyrefly.bxl:main` [BXL
+script](https://buck2.build/docs/bxl/) to get the necessary information about
+the targets to type check.
+
+Here is a description of the supported optional options:
+
+**`isolation-dir`**
+
+Name of the isolation dir to use when running the BXL query (see `buck2 --help`
+or [Buck's documentation](https://buck2.build/docs/concepts/isolation_dir/) for
+more information about isolation dirs).
+
+Type: string
+Default: none (Buck's own default is `v2`)
+
+**`extras`**
+
+Extra command line arguments passed to `buck2` when running the query.
+
+- Type: list of strings
+- Default: `[]`
+
+#### Custom queries
+
+Arbitrary build systems can be integrated using the `custom` type:
+
+```toml
+[build-system]
+type = "custom"
+command = ["some", "command", "..."]
+# Optional: The root of the repository. Repo roots here will be shared between configs.
+repo_root = "src/"
+```
+
+Here is a description of the supported options:.
+
+**`command`**
+
+The command executed to query the build system about available targets.
+
+Pyrefly will call this in the form `<command> @<argfile>`, where `<argfile>`
+has the format:
+
+```text
+--
+<arg-flag>
+<arg>
+...
+```
+
+`<arg-flag>` is either `--file` or `--target`, depending on the type of
+`<arg>`, and `<arg>` is an absolute path to a file or a build system's target.
+
+The command should output a JSON file with the following structure:
+
+```json
+{
+    "root": "/path/to/this/repository",
+    "db": {
+        "//colorama:py-stubs": {
+            "srcs": {
+                "colorama": [
+                    "colorama/__init__.pyi"
+                ]
+            },
+            "deps": [],
+            "buildfile_path": "colorama/BUCK",
+            "python_version": "3.12",
+            "python_platform": "linux"
+        },
+        "//colorama:py": {
+            "srcs": {
+                "colorama": [
+                    "colorama/__init__.py"
+                ]
+            },
+            "deps": ["//colorama:py-stubs"],
+            "buildfile_path": "colorama/BUCK",
+            "python_version": "3.12",
+            "python_platform": "linux"
+        },
+        "//colorama:colorama": {
+            "alias": "//colorama:py"
+        }
+}
+```
+
+Where:
+
+- `root` is the absolute path the root of the repository.
+- `db` is a map from target names to either:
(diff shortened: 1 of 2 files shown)


website/docs/configuration.mdx
```diff
@@ -599,6 +598,62 @@ allowlist or denylist). Regular
 - Default: `true`
 - Flag equivalent: `--use-ignore-files`
 
+### `build-system`
+
+Pyrefly supports integrating into build systems to discover targets to lint and
```
connernilsen: I think we want to stay away from the term "lint" for a while, since there are some strong opinions about that.
```suggestion
Pyrefly supports integrating into build systems to discover targets to type check and
```

website/docs/configuration.mdx
```diff
@@ -599,6 +598,62 @@ allowlist or denylist). Regular
...
+Behind the scenes, Pyrefly will run the
+`prelude//python/sourcedb/pyrefly.bxl:main` [BXL
+script](https://buck2.build/docs/bxl/) to get the necessary information about
+the targets to typecheck.
+
+The following optional configuration options are also supported:
+
+- `isolation_dir` (`str`): name of the isolation dir to use when running the
+  BXL query (see `buck2 --help` or
+  <https://buck2.build/docs/concepts/isolation_dir/> for more information about
```
connernilsen:
```suggestion
  [Buck's documentation](https://buck2.build/docs/concepts/isolation_dir/) for more information about
```

website/docs/configuration.mdx
```diff
@@ -599,6 +598,62 @@ allowlist or denylist). Regular
...
+[build-system]
+type = "buck"
+```
+
+Behind the scenes, Pyrefly will run the
+`prelude//python/sourcedb/pyrefly.bxl:main` [BXL
+script](https://buck2.build/docs/bxl/) to get the necessary information about
+the targets to typecheck.
+
+The following optional configuration options are also supported:
```
connernilsen: Can you update the options below to follow the format used for other config options?

Something like the following would be preferred

```
#### Buck2

<description>

**`<option>`**

<description>

- Type: ...
- Default: ...
```

website/docs/configuration.mdx
```diff
@@ -599,6 +598,62 @@ allowlist or denylist). Regular
...
+
+- `isolation_dir` (`str`): name of the isolation dir to use when running the
+  BXL query (see `buck2 --help` or
+  <https://buck2.build/docs/concepts/isolation_dir/> for more information about
+  isolation dirs).
+- `extras` (`list[str]`): extra command line arguments passed to `buck2` when
+  running the query.
+
```
connernilsen, line 648: Might be worth saying the command that will be built here, which is
```bash
buck2 [--isolation-dir <isolation_dir>] bxl --reuse-current-config [<extras>...] prelude//python/sourcedb/pyrefly.bxl:main @<tempfile>
```

website/docs/configuration.mdx
```diff
@@ -599,6 +598,62 @@ allowlist or denylist). Regular
...
+Arbitrary build systems can be integrated using the `custom` type:
+
+```toml
+[build-system]
+type = "custom"
+command = ["some", "command", "..."]
+# Optional: The root of the repository. Repo roots here will be shared between configs.
+repo_root = "src/"
+```
```
connernilsen, line 659: The style of this differs from the style of the Buck section. Can you either make the Buck section demo all of the values, with similar comments as here, or move the example up to the main `### build-system` section looking something like this
```toml
[build-system]
type = "<type>"
# type-specific arguments
```

website/docs/configuration.mdx
```diff
@@ -599,6 +598,62 @@ allowlist or denylist). Regular
...
+```text
+--
+<arg-flag>
+<arg>
+...
+```
+
+`<arg-flag>` is either `--file` or `--target`, depending on the type of
+`<arg>`, and `<arg>` is an absolute path to a file or a build system's target.
+
```
connernilsen, line 679: It may also be worth describing the JSON we consume. If you don't want to do it yourself, I can fill that in for you. If you do though, [the structure is here](https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_build/src/query/mod.rs#L253-L257).

connernilsen (changes requested):
Hey @cbarrete, thanks for putting this together!

Can you pull out the formatting changes into a separate PR? There are two things going on here, so I'd like to focus on the changes in each of them separately. The format PR should be quick to review and pull in :).

As for the build system documentation, I'm not sure if we're in a state where we're ready to publicly support things yet. We're still at a point where the versions of Pyrefly and Buck's prelude are still tightly coupled, and breaking/backward incompatible changes are happening relatively often. I also don't think we have the bandwidth to offer much support around this yet. 

If we're adding this documentation, I'd like to make it clear that it's pretty unstable and will continue to be for a while, and that while feature requests and bug reports are welcome, support will likely be lower priority than other issues for a while.

Let me know your thoughts around this, and I'm very open to discussion if you disagree with my requests or have other questions/ideas.

website/docs/configuration.mdx
connernilsen: I see below that you have this at the start of the custom section, so please put this wherever you end up keeping the custom command we invoke

website/docs/configuration.mdx
cbarrete: Fair enough, my bad!

website/docs/configuration.mdx
cbarrete: I'd be happy to, but this is a more complex option than all others, because it has multiple variants, and its value is a product type.

I've done my best, let me know if you'd like me to tweak it further.

website/docs/configuration.mdx
cbarrete: Done, I've added optional options to the Buck2 section demo.

website/docs/configuration.mdx
cbarrete: I've done what I could based on the sample value in a unit test I found.
Let me know if this is good enough.

cbarrete:
Thanks for the feedback! I've tried to address all your comments, let me know how it looks to you.

Also, apologies for the awkward stacking, I'm not very used to branch-based code review tools.

I've opened https://github.com/facebook/pyrefly/pull/1969 under this review for the formatting changes alone.

I've also added a big disclaimer that this is not well supported yet.
Frankly, I've opened this PR to help a friend that is starting to use Buck2 for Python. While I think that "unstable" docs are better than nothing, I would understand if you didn't want to merge this commit at this time. All I'd like in return is for it to be documented when it _does_ become stable, if that's not too much to ask!

connernilsen:
Awesome, this is looking great! I'll pull it in and start the review process internally. Thanks for working on this.

> Also, apologies for the awkward stacking, I'm not very used to branch-based code review tools.

No worries, I also haven't worked with open source git stuff in a while, so I may ask for some things that might be unreasonable, and it's totally fine to push back if my requests are weird.

> I've also added a big disclaimer that this is not well supported yet.

> Frankly, I've opened this PR to help a friend that is starting to use Buck2 for Python. While I think that "unstable" docs are better than nothing, I would understand if you didn't want to merge this commit at this time. All I'd like in return is for it to be documented when it does become stable, if that's not too much to ask!

That was the big thing I wanted to see addressed before the documentation was added. I think this is looking great so far, so I'll see if anyone else on the team thinks there's anything that needs to change with this, and I'll try to get it merged in soon.

kinto0 (approved):

PR #1963, opened 2025-12-30
fix Support constrained TypeVars #1422

### LINKED ISSUE #1422: Support exhaustiveness checks for constrained quantified type vars
### Describe the Bug

It seems like pyrefly doesn't support TypeVar constraints. From the docs (https://docs.python.org/3/library/typing.html#typing.TypeVar):

> Using a constrained type variable, however, means that the TypeVar can only ever be solved as being exactly one of the constraints given

which means that this should work:

```python
from typing import TypeVar

T = TypeVar("T", int, str)

def f(x: T) -> T:
    if isinstance(x, int):
        return x + 1
    elif isinstance(x, str):
        return x + "1"
```

Instead, pyrefly complains:

```
ERROR sandbox.py:5:16-17: Function declared to return `T`, but one or more paths are missing an explicit `return` [[bad-return](https://pyrefly.org/en/docs/error-kinds/#bad-return)]
ERROR sandbox.py:7:16-21: Returned type `int` is not assignable to declared return type `T` [[bad-return](https://pyrefly.org/en/docs/error-kinds/#bad-return)]
ERROR sandbox.py:9:16-23: Returned type `str` is not assignable to declared return type `T` [[bad-return](https://pyrefly.org/en/docs/error-kinds/#bad-return)]
```

[Pyright](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgGoCGIAUFQVALyEkUgAUARAewDSYoy8AzjBABKGgBNiwKMFYAPAFyFRUALQA%2BQoqpQ9mGUkGph5FAGNiC3qhiid%2Bx1BDEYAVxAoo8qAGooAIy6%2BsQANkiGxiimFlbyQiL2wU7Orh5ePv7sAexAA) and mypy both seem allow this function to pass without errors.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjABqqSgBo6lGADcYqKAH0mxGAB10mgXQC8g4WMoAKdSAFmprBlLgNKASk2bMMMHTDH8iQQ7oBaAD5BRE06cM53CDhWOwwAYxgvK3QGB1D0CKzpGAYAV0pM-DoAajoARjCImCgIKJj0OPRE5Lo7RwzsiJl8wrpisrNKkBAJEDyGaDgSckQQAGI6AFVJ2qYPPObJ3EbnLDcPXhpUBiV0PJpsGBNvTlS-ILb7Tu7cgsywMwA5C6vKH2A%2BAAvmZNKMQGQZGAoKRCAxaFAKIsAAqkKEwtoYHAEOjxHaQNgFE4QHaETSLADKMBgdAAFgwGMQ4IgAPQsyFuGGEXhsFkwdAszC4eJwFl49AEonbAWHSh0VCyVDQVDYWC4-EQQmUYk7Oi4YjS6aaMgMWk7fzySgxXX6MwAZkI5QATKD0CAgWNUPFJvIAGLQGAUNBYPBEMjuoA

_No response_

asukaminato0721:
# Summary

Fixes #1422

Adjusted constrained TypeVar handling so return checks accept any constraint and missing-return is suppressed when an isinstance chain covers all constraints, then added a regression test and updated expectations.

Updated constrained TypeVar subtyping to accept any matching constraint.

Added a conservative exhaustive-isinstance check for constrained TypeVar parameters to avoid false missing-return errors.

# Test Plan

Added a new return regression test and removed the now-obsolete return-type error expectation.


stroxler (commented):
Thanks for the PR. I think it works, but I actually think the implementation isn't the right direction, we fixed the reported bug but not the root cause, which is missing exhaustiveness checks.

In particular, this snippet has exactly the same bug:
```
def g(x: int | str) -> int | str:
    if isinstance(x, int):
        return x
    elif isinstance(x, str):
        return x
```


I have some related work in draft match exhaustiveness changes I have (which are blocked from landing right away because I have major features that depend on one another + on some internal constraints)

I think I may need to come back to this PR once that work is unblocked to better understand how it all fits together, but my instinct is that we need a much more general mechanism for handling whether a final statement with branching control flow "falls off the end" or not in function return logic.

pyrefly/lib/solver/subset.rs
```diff
@@ -991,7 +991,7 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
             }
             (t1, Type::Quantified(q)) => match q.restriction() {
                 // This only works for constraints and not bounds, because a TypeVar must resolve to exactly one of its constraints.
-                Restriction::Constraints(constraints) => all(constraints.iter(), |constraint| {
+                Restriction::Constraints(constraints) => any(constraints.iter(), |constraint| {
```
stroxler, line 1631: I think this change is probably not closely related, and it looks like it's *probably* right to me.

Would you want to submit a PR with just this one change?

rchen152:
Closing because the linked issue is fixed

PR #1964, opened 2025-12-30
fix LSP: Provide key autocompletion on annotated TypedDict #1886

### LINKED ISSUE #1886: LSP: Provide key autocompletion on annotated `TypedDict`
### Describe the Bug

I'd like to have key autocompletion suggestions when constructing an annotated `TypedDict`.

Image

Example:

```python
from typing import TypedDict

class Dimension(TypedDict):
    x: int
    y: int
    z: int

def main():
    dimensions: Dimension = {"x": 5}

if __name__ == "__main__":
    main()
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjEwARCAGMGAHXSzZEqKjhw64mjHRwIudAAohxEeKkBKRLLpW6%2BRJ3Qz01uqTutHzgF5uH8uVhgwOhpUVj1zS2tMLk1tXTg7dVidJwBeOmBpEHwsuwBWAF8-WQgggH0y9FQNCrpU9KyKkNYK3MirZv1TWRAAGhAAVwZoOBJyRBAAYjoAVWGoCCY6MAH0KRS4P0xA5d4QhkqBmmwYSj1bewZTOgBaAD46OAZKCydrShgGAconMCyAOSOJxeGXwRRAPX6ZA%2BYCgpEIDFoUAo0wACqQYXDHhgcAQ6BJdJA2N9UMNdIRZNMAMowGB0AAWDAYxASAHpWdDAnDCLw2KzNKzMLgJHBWQT0ESSWT0KzdpQ6KgAG6hJTYWD4wkQYmUUkpOi4YjS0ayMgMem6G6K05xNJ0LIAZkIAEYAExZHoFfqodZWgBi0BgFDQWDwRDIIAKQA

_No response_

asukaminato0721:
# Summary

Fixes #1886

Added DictKeyLiteralContext plus helpers so completion logic can recognize both subscript expressions and literal dictionary keys, enabling us to reuse range/facet information for either source.

Reworked add_dict_key_completions to use the new context.

# Test Plan

Covered the regression with a new LSP test that exercises completions while editing a typed dict literal and asserts the expected field suggestions appear.

### DIFF
--- a/pyrefly/lib/state/lsp/dict_completions.rs
+++ b/pyrefly/lib/state/lsp/dict_completions.rs
@@ -192,30 +192,43 @@ impl<'a> super::Transaction<'a> {
         let nodes = Ast::locate_node(module, position);
         let mut best: Option<(u8, TextSize, ExprDict, ExprStringLiteral)> = None;
         for node in nodes {
-            if let AnyNodeRef::ExprDict(dict) = node {
-                for item in &dict.items {
-                    let Some(key_expr) = item.key.as_ref() else {
-                        continue;
-                    };
-                    if let Expr::StringLiteral(literal) = key_expr {
-                        let (priority, dist) =
-                            Self::string_literal_priority(position, literal.range());
-                        let should_update = match &best {
-                            Some((best_prio, best_dist, _, _)) => {
-                                priority < *best_prio
-                                    || (priority == *best_prio && dist < *best_dist)
-                            }
-                            None => true,
-                        };
-                        if should_update {
-                            best = Some((priority, dist, dict.clone(), literal.clone()));
-                            if priority == 0 && dist == TextSize::from(0) {
-                                break;
-                            }
-                        }
+            let AnyNodeRef::ExprDict(dict) = node else {
+                continue;
+            };
+            let mut best_in_dict: Option<(u8, TextSize, ExprStringLiteral)> = None;
+            for item in &dict.items {
+                let Some(key_expr) = item.key.as_ref() else {
+                    continue;
+                };
+                let Expr::StringLiteral(literal) = key_expr else {
+                    continue;
+                };
+                let (priority, dist) = Self::string_literal_priority(position, literal.range());
+                let should_update = match &best_in_dict {
+                    Some((best_prio, best_dist, _)) => {
+                        priority < *best_prio || (priority == *best_prio && dist < *best_dist)
+                    }
+                    None => true,
+                };
+                if should_update {
+                    best_in_dict = Some((priority, dist, literal.clone()));
+                    if priority == 0 && dist == TextSize::from(0) {
+                        break;
                     }
                 }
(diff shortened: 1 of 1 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -2789,6 +2824,60 @@ impl<'a> Transaction<'a> {
...
+                            Self::string_literal_priority(position, literal.range());
+                        let should_update = match &best {
+                            Some((best_prio, best_dist, _, _)) => {
+                                priority < *best_prio
+                                    || (priority == *best_prio && dist < *best_dist)
+                            }
+                            None => true,
+                        };
+                        if should_update {
+                            best = Some((priority, dist, dict.clone(), literal.clone()));
```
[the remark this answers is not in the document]
kinto0: please do this

pyrefly/lib/state/lsp.rs
```diff
@@ -2913,7 +3001,7 @@ impl<'a> Transaction<'a> {
             }
         }
 
-        if let Some(base_type) = self.get_type_trace(handle, base_expr.range())
+        if let Some(base_type) = self.get_type_trace(handle, context.base_range())
```
kinto0: could you add comments explaining why each case does what it does?

as I understand it base_range is the entire dict range. does get_type_trace on the entire dict literal correctly infer the type? which tests cover this behavior?

pyrefly/lib/state/lsp.rs
```diff
@@ -92,6 +93,40 @@ mod quick_fixes;
 
 pub(crate) use self::quick_fixes::extract_function::LocalRefactorCodeAction;
 
+#[derive(Clone)]
```
kinto0: this stuff is starting to get very specific. could it go in it's own file instead of lsp.rs?

pyrefly/lib/state/lsp.rs
```diff
@@ -92,6 +93,40 @@ mod quick_fixes;
...
+enum DictKeyLiteralContext {
+    KeyAccess {
```
kinto0: could you add examples for all of these so it is easy to skim and see which is which?

pyrefly/lib/state/lsp.rs
```diff
@@ -2789,6 +2824,60 @@ impl<'a> Transaction<'a> {
         }
     }
 
+    fn dict_key_literal_context(
```
kinto0: this stuff also could all go in it's own file about dictionary completions

kinto0 (changes requested):
thanks for working on this! just a few suggestions about code cleanup. as this file gets bigger it's even more important to move related functionality into new files so it isn't as confusing

PR #1965, opened 2025-12-30
Fix: Skip bad-return validation for Protocol methods (#1916)

### LINKED ISSUE #1916: Protocol definitions complain that "Function declared to return `X` but is missing an explicit `return`"
### Describe the Bug

I have a Protocol defined like this:

```python
class View(Protocol):
    """blah blah"""

    @property
    def sort_state(self) -> SortState:
        """Return the current sorting/grouping settings."""
```

and this causes pyrefly 0.46.0 to emit

```
ERROR Function declared to return `SortState` but is missing an explicit `return` [bad-return]
   --> src/.../views/base.py:206:29
    |
206 |     def sort_state(self) -> SortState:
    |                             ^^^^^^^^^
```

I expected no error (and the code is accepted fine by mypy and most other Python type checkers I've tested).

### Sandbox Link

[Here you are](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AFag1wBjXFAA66adNFRUcOHQDKvBioaoGMRNLoG6xRXFkz08k3QBqEGAHcAFENwjxUAJR70hupJD%2B-tgKABZ0waghgQEB5r4AAsTUxDB8pPqGmDBgdHDqAPpwWjqOcDBQYB50ALQAfKrqmtq6Gb4G0f4ASjAMAK6UPgwhMHSi-ZQw6Px5fKxsAPRs1L0s7Lk9DHNwhB0gIAA0IL2bUNtkFADEdACqJxBMdGC9Fpu46KbmWTlgvDTa%2BehejRsKlHPhEJwplU6rkGJRvL4Jn0Bo9-AA5IEg%2BF0YD4AC%2BgXQBxAZAmYCgpEIIhoUEuglIZIpuQwOAIozekDY-W0EDeO3QVxUMBGIQYDGIcEQ83mpOyFMIvAWk3mmDEcHm4nQnO5r3Q80evDoqAAbqhoKhgiNNdrKDy3nRcMRddtpGQhm9qsbUnBeT4ALx%2BEAAZkIAEYAEyEkB4w6oUSbL0AMWgMAoaCweCI5zxQA)

_No response_

AryanBagade:
# Summary
Protocol methods with only a docstring (no `...` or implementation body) were incorrectly emitting "Function declared to return X but is missing an explicit return" errors. This is valid Python for Protocol definitions and is accepted by mypy and other type checkers.

**Changes:**
  - Added `class_key` field to `ReturnTypeKind::ShouldValidateAnnotation` to track the enclosing class
  - In `solve.rs`, check if the class is a Protocol and skip return validation for Protocol methods
  - Added regression test covering the reported case

Fixes #1916

# Test Plan
- Added `test_protocol_method_with_docstring` test case covering:
    - `@property` with docstring only
    - Regular method with docstring only
    - Method with `pass`
    - Method with ellipsis (already worked)
- Verified regular (non-Protocol) classes still error correctly
- Ran `python3 test.py` - all tests pass
- Ran `cargo test --lib protocol::` - all 37 protocol tests pass

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -3155,7 +3155,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         annotation,
                         stub_or_impl,
                         decorators,
-                        class_key,
                         implicit_return,
                         is_generator,
                         has_explicit_return,
@@ -3164,18 +3163,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         // It will result in an implicit Any type, which is reasonable, but we should
                         // at least error here.
                         let ty = self.get_idx(*annotation).annotation.get_type().clone();
-                        // Check if this method is in a protocol class
-                        let is_in_protocol = class_key.is_some_and(|key| {
-                            self.get_idx(key)
-                                .0
-                                .as_ref()
-                                .is_some_and(|cls| self.get_metadata_for_class(cls).is_protocol())
-                        });
-                        // If the function body is stubbed out, if the function is decorated with
-                        // `@abstractmethod`, or if the function is a method in a protocol class,
-                        // we blindly accept the return type annotation.
+                        // If the function body is stubbed out or if the function is decorated with
+                        // `@abstractmethod`, we blindly accept the return type annotation.
                         if *stub_or_impl != FunctionStubOrImpl::Stub
-                            && !is_in_protocol
                             && !decorators.iter().any(|k| {
                                 let decorator = self.get_idx(*k);
                                 match decorator.ty.callee_kind() {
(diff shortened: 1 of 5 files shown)


samwgoldman (changes requested):
Thank you!

I think we can do this a bit more simply, by instead adjusting the logic in `BindingsBuilder::function_body`, where we determine the `stub_or_impl` value. The class key is already available in that code, so we can also avoid threading it through the binding.

For what it's worth, it looks like we always treat `...` as a stub. We did this to match pyright & because the conformance tests are written this way. The conformance tests were recently updated to avoid the need to accept `...` always (https://github.com/python/typing/pull/2134). For now, I think we should leave this behavior alone, but might be an interesting follow-up.

PR #1973, opened 2025-12-31
Fix implicit-any error for bare tuple and Callable (#1970)

### LINKED ISSUE #1970: `implicit-any` doesn't warn on bare `type` or `tuple`
### Describe the Bug

When I enable the `implicit-any` error, pyrefly warns on a bare `list` annotation but not a bare `type` or `tuple` annotation:
```
def f(
    x: list,  # pyrefly errors here (good)
    y: type,  # pyrefly doesn't error here, but I want it to
    z: tuple,  # pyrefly doesn't error here, but I want it to
):
    pass
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABQA66NrN%2BiNUEcALgDTM2NUp16liMQSzYAvMQFdisaQEpEQtsVRwE-EGQBO1KKUK9cAWygUAxDQAKpY2FM00WPPhoBjXOkgAcwVDVF4If0JmewBlGBgaAAteXmI4RAB6DKMTM1xDQIyYdAzMXB84DL8AiGDQ8P8MunyaVAA3VGhUbFhffyCQsIiWXGIG9Dgo9BhDQ3zJiEtlCB8IXgBaDFIaAF4aXkMFGBAAX31UH3C2mAAxaGPEEA8cAhJyE6A

_No response_

jackulau:
# Summary

Fixes #1970

When the `implicit-any` error is enabled, bare `tuple` and `Callable` annotations now generate errors, matching the existing behavior for `list`.

## Before
```python
def f(
    x: list,      # ERROR: implicit-any
    y: tuple,     # no error
    z: Callable,  # no error 
):
    pass
```

## After
```python
def f(
    x: list,      # ERROR: implicit-any
    y: tuple,     # ERROR: implicit-any
    z: Callable,  # ERROR: implicit-any
):
    pass
```

## Changes

- Add `ImplicitAny` error for `SpecialForm::Tuple`, `SpecialForm::Callable`, and `SpecialForm::Type` in `canonicalize_all_class_types`
- Add `ImplicitAny` error for `ClassDef(tuple)` to handle the builtin `tuple`
- Add test case covering the fix

## Notes

- Bare builtin `type` still doesn't error because it's not defined as generic in typeshed (the `type` class has no type parameters). However, `typing.Type` does error correctly.

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1278,48 +1278,20 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     ) -> Type {
         ty.transform(&mut |ty| match ty {
             Type::SpecialForm(SpecialForm::Tuple) => {
-                errors.add(
-                    range,
-                    ErrorInfo::Kind(ErrorKind::ImplicitAny),
-                    vec1![
-                        "Cannot determine the type parameter for generic class `tuple`".to_owned(),
-                        "Either specify the type argument explicitly, or specify a default for the type variable.".to_owned(),
-                    ],
-                );
+                Self::add_implicit_any_error(errors, range, "tuple", None);
                 *ty = Type::unbounded_tuple(Type::Any(AnyStyle::Implicit));
             }
             Type::SpecialForm(SpecialForm::Callable) => {
-                errors.add(
-                    range,
-                    ErrorInfo::Kind(ErrorKind::ImplicitAny),
-                    vec1![
-                        "Cannot determine the type parameter for generic class `Callable`".to_owned(),
-                        "Either specify the type argument explicitly, or specify a default for the type variable.".to_owned(),
-                    ],
-                );
+                Self::add_implicit_any_error(errors, range, "Callable", None);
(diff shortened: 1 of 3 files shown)


pyrefly/lib/alt/expr.rs
```diff
@@ -1278,16 +1278,48 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     ) -> Type {
         ty.transform(&mut |ty| match ty {
             Type::SpecialForm(SpecialForm::Tuple) => {
+                errors.add(
+                    range,
+                    ErrorInfo::Kind(ErrorKind::ImplicitAny),
+                    vec1![
```
rchen152: Would it be possible to pull out the error message creation from https://github.com/facebook/pyrefly/blob/713e5b7d90f3515b343e1bc731e6001c61d5d109/pyrefly/lib/alt/class/targs.rs#L176-L183 into a reusable helper method rather than copy-pasting?

rchen152 (changes requested):
Thanks! The logic looks good, but it would be nice if we could reduce code duplication a bit.

rchen152 (approved):

PR #1974, opened 2025-12-31
Fix bad-specialization false positive for bounded TypeVar in callable (#1960)

### LINKED ISSUE #1960: False positive bad-specialization
mypy and pyright accept this code, pyrefly 0.46.1 does not.

```python


class Box[U]:
    pass


class Result[T: int]:
    def __init__(self, x: T | Box[T]) -> None:
        pass


def f[T: int](factory: Callable[[T], Result[T]]) -> None:
    pass


f(factory=Result)
```

```
ERROR `Box[Unknown] | Unknown` is not assignable to upper bound `int` of type variable `T` [bad-specialization]
  --> example.py:27:2
   |
27 | f(factory=Result)
   |  ^^^^^^^^^^^^^^^^
```

jackulau:
# Summary

Fixes #1960

When passing a generic class as a factory to a function expecting `Callable[[T], Result[T]]`, pyrefly incorrectly reported a `bad-specialization` error

## The Problem
```python

class Box[U]:
    pass

class Result[T: int]:
    def __init__(self, x: T | Box[T]) -> None:
        pass

def f[T: int](factory: Callable[[T], Result[T]]) -> None:
    pass

f(factory=Result)  # ERROR: `Box[Unknown] | Unknown` is not assignable to upper bound `int`
```

## Root Cause

When checking `$T <: SomeType` where `$T` has a bound (e.g., `T: int`), the solver was:

1. Setting `$T = SomeType`
2. Then checking if `SomeType` satisfies the bound

For this case, the constructor callable for `Result[Any]` has parameter type `Any | Box[Any]`. The solver set `$T = Any | Box[Any]` and then checked `Any | Box[Any] <: int`, which failed.

## The Fix

Before setting the type variable, check if the bound already satisfies the constraint. If `bound <: target`, then any value within the bound automatically satisfies the constraint by transitivity, so we can return early without incorrectly setting the variable.

For this case: `int <: Any | Box[Any]` is true (since `int <: Any`), so the constraint is satisfied without needing to set `$T = Any | Box[Any]`.

## Test Plan

- Added regression test `test_callable_constructor_bounded_typevar_with_union_param`

### DIFF
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -1440,7 +1440,10 @@ impl<'a, Ans: LookupAnswer> Subset<'a, Ans> {
                         // is also a subtype of `t2`, so the constraint is satisfied.
                         // This handles cases like `$T: int` vs `Any | SomeType` where
                         // the bound `int` is a subtype of the target type.
+                        // We set the variable to the bound to ensure it gets resolved.
                         if self.is_subset_eq(&bound, t2).is_ok() {
+                            let variables = self.solver.variables.lock();
+                            variables.update(*v1, Variable::Answer(bound));
                             return Ok(());
                         }
                         let variables = self.solver.variables.lock();


samwgoldman:
Thanks for investigating! This is a clever idea, but I think solving the variable to its bound is a bit problematic. It looks like the new errors in mypy primer arise from this as well.

I know our current behavior of solving a type variable to the "first" concrete type it sees in constraints leads to issues like this. I took a moment to analyze this example against our currently planned design for improving type variable solving. First, let's simplify the example a little bit:

```

def f[T: int](g: Callable[[T], T]):
    pass

def g(x: int|str) -> int:
    return 0

f(g)
```

Here, we are passing a callable type `int|str -> int` into a position expecting `T -> T`, and we want to solve for `T`.

```
* create fresh X <: int, substitute to get `f' : X -> None`
* int|str -> None <: X -> None  [call f'(g)]
   * X <: int | str             [func arg contravariant]
      * X = int | str           [eager variable solution *]
   * int <: X                   [func return covariant]
      * int <: int|str          [X=int|str]
* int|str <: int                [check solution against bound **]
```

Things go "wrong" when we eagerly solve for X (*) and we realize later when we go check against the bound (**).

Notice that we see both `X <: int|str` and `int <: X` when checking the call. If we considered both of these constraints together, we would see that the better, more precise solution for `T` is `int`. In general, solving a type variable to an upper bound when a lower bound exists will be imprecise -- this is also why solving to the typevar bound, an upper bound, leads to errors.

Delaying the solution, solving #105, seems like the best solution to this problem as well.

rchen152:
I'm going to close this PR, since it sounds like we should work on a more general solution for #105 instead.

PR #1977, opened 2026-01-01
fix Reading `__v` from the outside of the class, pyrefly doesn't give any error while Python interpreter gives error #1971

### LINKED ISSUE #1971: Reading `__v` from the outside of the class, pyrefly doesn't give any error while Python interpreter gives error
### Describe the Bug

*Memo:
- pyrefly check
- pyrefly 0.46.1
- python test.py
- Python 3.14.0
- Windows 11

Reading `__v` from the outside of the class, pyrefly doesn't give any error as shown below:

*Memo:
- A double leading underscore(__abc) can make attributes private in a class.

```python
class Cls:
    __v = 100

cls: Cls = Cls()

print(cls.__v) # No error
```

Actually, Python interpreter gives the error properly as shown below:

> AttributeError: 'Cls' object has no attribute '__v'

So, pyrefly should give an error like Python interpreter gives as shown below:

> ERROR 'Cls' object has no attribute '__v'

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1971

now records both the class key and syntactic class name for every double-underscore attribute access using the nearest enclosing class scope, so we know exactly which class body the access belongs to even when it appears inside methods or static/class methods. The public getter returns the tuple so downstream consumers can see both pieces.

# Test Plan

Added regression test

### DIFF
--- a/pyrefly/lib/alt/special_calls.rs
+++ b/pyrefly/lib/alt/special_calls.rs
@@ -54,8 +54,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let ret = if args.len() == 2 {
             let expr_a = &args[0];
             let expr_b = &args[1];
-            let a = self.expr_infer_with_hint(expr_a, hint, errors);
-            let b = self.expr_untype(expr_b, TypeFormContext::FunctionArgument, errors);
+            let a_raw = self.expr_infer_with_hint(expr_a, hint, errors);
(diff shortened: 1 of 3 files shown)


asukaminato0721:
aa, here comes the problem

```py
from enum import Enum
class MyEnum(Enum):
    X = 1
    Y = 2
    __PRIVATE = 3
    def foo(self):
        self.__PRIVATE # works
MyEnum.__PRIVATE # error
```

asukaminato0721:
Leave it here, a bit too tricky :(

PR #1981, opened 2026-01-02
fix(lsp): stop at import for module bindings in go-to declaration

### LINKED ISSUE #1954: [go-to declaration] support module bindings
### Describe the Bug

example from [here](https://github.com/facebook/pyrefly/issues/1291#issuecomment-3673794744)

```
import torch
import numpy as np

def main():
    torch
 #   ^ cursor here, goto-decl called, hoping to get cursor to import torch line
    np.abs
 #   ^ cursor here, goto-decl called, hopoing to get cursor to import numpy as np line
 #.     ^ cursor here, goto-decl called, falls back to goto def in this case.  

```

In order to implement it, first add a test for the behavior in `pyrefly/pyrefly/lib/test/lsp/declaration.rs`, then look for the place to update. go-to declaration is implemented as `find_definition` but with ImportBehavior::StopAtEverything. maybe ImportBehavior::StopAtEverything is not stopping for this case?  

### Sandbox Link

_No response_

_No response_

bv-saketha-rama:
Support module bindings in go-to declaration by respecting ImportBehavior::StopAtEverything for IntermediateDefinition::Module.

Previously, go-to declaration would follow through to the module definition instead of stopping at the import statement for module imports like 'import torch'. Now it correctly navigates to the import line.

Changes:
- Add TextRange to IntermediateDefinition::Module to track import location
- Check ImportBehavior before following through to module definition
- Add test for module binding go-to declaration

Fixes #1954

# Summary

This PR fixes the "go-to declaration" feature to correctly handle module bindings (e.g., `import torch`). Previously, the logic would always jump through the import to the module definition itself, even when configured to `StopAtEverything`.

Changes:

- ide.rs: Updated the `IntermediateDefinition::Module`variant to include a `TextRange`, allowing it to store the location of the import statement.

- lsp.rs: Modified 
`resolve_intermediate_definition`
 to `checkpreference.import_behavior`. If set to `StopAtEverything`, it now returns the import's location instead of resolving the full module handle.

- answers.rs: Updated pattern matching for `IntermediateDefinition::Module` to accommodate the new field.

# Test Plan

A new test case 
goto_declaration_module_binding
 was added to 
pyrefly/lib/test/lsp/declaration.rs
 which verifies that go-to declaration on a module-level binding correctly identifies the 
import
 statement as the declaration site.

Ran the following tests to ensure correctness and no regressions: 

`cargo test goto_declaration -p pyrefly --lib `

# Results 

```
test test::lsp::declaration::goto_declaration_stops_at_import ... ok
test test::lsp::declaration::goto_declaration_aliased_import ... ok
test test::lsp::declaration::goto_declaration_module_binding ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 3287 filtered out; finished in 6.24s
```

### DIFF
--- a/pyrefly/lib/state/ide.rs
+++ b/pyrefly/lib/state/ide.rs
@@ -139,7 +139,10 @@ fn create_intermediate_definition_from(
                     // actual module that corresponds to the key must be `x.y`.
                     name.dupe()
                 };
-                return Some(IntermediateDefinition::Module(def_key.range(), imported_module_name));
+                return Some(IntermediateDefinition::Module(
+                    def_key.range(),
+                    imported_module_name,
+                ));
             }
             Binding::Function(idx, ..) => {
                 let func = bindings.get(*idx);


bv-saketha-rama:
Facebook Internal - Builds & Tests are failing, how can I find out the exact issue? 
I can re-commit after running some linting checks locally, is that enough?

kinto0:
> Facebook Internal - Builds & Tests are failing, how can I find out the exact issue? I can re-commit after running some linting checks locally, is that enough?

it's just linting. I'll fix it on import, you don't need to worry about it.

PR #1983, opened 2026-01-02
fix: Pyrefly gets confused if a class has the same name as an overloaded method

### LINKED ISSUE #1697: Pyrefly gets confused if a class has the same name as an overloaded method
### Describe the Bug

Example:
```
from typing import assert_type, overload

class A:
    @overload
    def B(self, x: int) -> B: ...
    @overload
    def B(self, x: str) -> B: ... # Expected a type form, got instance of `(self: Self@A, x: int) -> B`
    def B(self, x):
        return B()

class B:
    x: int

assert_type(A().B(0).x, int)
assert_type(A().B("1").x, int) # assert_type(Any, int) failed
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2o4cGHwD6TYjAA0dXADcRUXKkwAddOoDGUQXDoBBROrom6AAXmLla9KbqYYYOgCEAFMKhgZ%2BRJ3QMASjoAWgA%2BF19CKONTCwVKJRUYkwcnNw8vOh86OAZKILCIuijCZPtHF3cYT28Ao1s7U0oYBgBXSls3APUtHSEIsuzWBh70XREGcWYYV31XAMI3AAYF-Blh7rGhCanJWfnF11UQAEZj1fX-AJApEFaGaDgSckQQAGI6AFUHqAgmOjArXQmgeuHQcFGqQBvBoqEm6FaNGwIlcQyuIXCuUo9TszTaHQBxwAcojkdi6MB8ABfY7qG4gMjNMBQUiEBi0KAUD4ABVITJZOQwOAIdE0YMgbHacIgYNK6A%2BAGUYDA6AALBgMYhwRAAeh1jMcLMIvDYOpg6B1mFwmjgOrF6AlUtBFuhlAEclQ0FQ2FgovFEEllGlYNkxGdT3UZAYqrBwXicBltgAvHRjgBmQgnABMtPQICpt1QIIgCgAYtAYBQ0Fg8EQyPmgA

_No response_

tannguyencse19:
# Summary

Add a static type name lookup variant that skips class-scope flow entries whose value style is `FlowStyle::FunctionDef`.
Route lookups performed under `Usage::StaticTypeInformation` to this static type lookup path, leaving all other usages unchanged.

Fixes #1697

# Test Plan

```
cargo test -p pyrefly test_overload_method_name_matches_class
```

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -603,8 +603,13 @@ pub enum FlowStyle {
     /// would get `foo.bar` here.
     ImportAs(ModuleName),
     /// Am I a function definition? Used to chain overload definitions.
-    /// If so, does my return type have an explicit annotation?
-    FunctionDef(Idx<KeyDecoratedFunction>, bool),
+    /// Track whether the return type has an explicit annotation and whether this
+    /// definition is marked as an overload.
+    FunctionDef {
+        function_idx: Idx<KeyDecoratedFunction>,
+        has_return_annotation: bool,
+        is_overload: bool,
+    },
     /// Am I a class definition?
     ClassDef,
     /// The name is possibly uninitialized (perhaps due to merging branches)
@@ -1263,7 +1268,10 @@ impl Scopes {
         name: &Name,
     ) -> Option<(Idx<Key>, Idx<KeyDecoratedFunction>)> {
         if let Some(value) = self.current().flow.get_value(name)
-            && let FlowStyle::FunctionDef(fidx, _) = value.style
+            && let FlowStyle::FunctionDef {
+                function_idx: fidx,
+                ..
+            } = value.style
         {
             return Some((value.idx, fidx));
         }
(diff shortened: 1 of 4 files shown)


pyrefly/lib/binding/scope.rs
```diff
@@ -2251,12 +2251,36 @@ impl Scopes {
...
             let is_class = matches!(scope.kind, ScopeKind::Class(_));
 
             if let Some(flow_info) = scope.flow.get_info_hashed(name)
                 && flow_barrier < FlowBarrier::BlockFlow
             {
+                if skip_class_flow_function_definitions
+                    && is_class
+                    && flow_info
+                        .value()
+                        .is_some_and(|value| matches!(value.style, FlowStyle::FunctionDef(..)))
```
rchen152: This isn't quite right. If a name is shadowed by a function definition, a type checker should error if it is used as a type annotation: https://github.com/python/typing/blob/211b0cb5470e856f6093f7c3d688295f72f78bb6/conformance/tests/annotations_forward_refs.py#L89. We should only skip `@overload`-decorated signatures.

tannguyencse19: @rchen152 I push a new commit to address this: https://github.com/facebook/pyrefly/pull/1983/commits/55189db67bfe9bb2fdcb6c1646baf8dc698c21d9 . Please review. Thanks

PR #1984, opened 2026-01-02
fix: Narrowing for isinstance and type[T] is unsound

### LINKED ISSUE #713: Narrowing for isinstance and type[T] is unsound
### Describe the Bug

I came across something during my own work on pytype2 and thought I'd check whether pyrefly has the same problem. Turns out it does:

```py
def f(cls: type[int], x: int | str) -> None:
    if isinstance(x, cls):
        reveal_type(x)  # reveals int - ok
    else:
        reveal_type(x)  # reveals str - too narrow
```

The first revealed type is fine but the second revealed type is too narrow. Someone could call `f(bool, 42)` and would end up in the `else` branch thinking that `42` is a `str`.

I think you simply cannot narrow the negative case for `isinstance` if its second argument is of type `type[T]` but not the literal name of a class.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQCYCmwUwAFAMYA2AzgFyyLkDaqMAugDRQAHmx5QAPlCYwQASigBaAHxQAcmBTkWRKLsxUkTVNICGKOuRpDBjJrO17HUEOQBu5EwwD68BJaGyOnrkzFpBTs5uHt6+-oFAA

_No response_

tannguyencse19:
# Summary

Fix unsound isinstance negative narrowing by skipping else-branch refinement when the classinfo is a runtime type[T] value rather than a literal class/ tuple/union of classes.

Fixes #713

# Test Plan

```
cargo test -p pyrefly -- test_isinstance_type_classinfo_no_negative_narrow
```

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -17,6 +17,7 @@ use ruff_python_ast::Expr;
 use ruff_python_ast::ExprNumberLiteral;
 use ruff_python_ast::Int;
 use ruff_python_ast::Number;
+use ruff_python_ast::Operator;
 use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/narrow.rs
```diff
@@ -219,6 +220,35 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         self.unions(res)
     }
 
+    fn allow_negative_isinstance_classinfo(
```
rchen152, line 223: Is it possible to specifically *disallow* `type[T]` rather than trying to enumerate everything that is allowed? This implementation makes me a bit nervous, since it seems easy to miss cases that we should allow.

rchen152 (changes requested):

pyrefly/lib/alt/narrow.rs
tannguyencse19: Let me check that

PR #1986, opened 2026-01-02
[RFC] Canonicalize `type` as `type[Any]`

fangyi-zhou:
# Summary
This diff canonicalised the builtin type (which is a class) as a `Type::Type` with an implicit `Any`.

The typing spec
(https://typing.python.org/en/latest/spec/special-types.html#type) requires that
"Plain `type` without brackets, the root of Python’s metaclass hierarchy, is equivalent to `type[Any]`."

Whereas assignments of `type` and `type[Any]` are allowed by Pyrefly before this diff, they are not considered as identical types by `assert_type` call. After the canonicalisation introduced by this diff, they are considered as identical. As a result, we are able to pass 3 more conformance test cases.

Since `type` itself is a class, we need to add some special treatment in the type system on inheritance, so that it can be inherited and used as a metaclass, while being a `Type::Type`. Some of the treatments are similar to how `tuple` is handled.

# Test Plan
`python test.py`


fangyi-zhou:
cc @grievejia who authored https://github.com/facebook/pyrefly/commit/6767f684d0d0286c5aad037e413e88639694adeb

grievejia:
Thanks for the patch! Will take a look later this week.

PR #1992, opened 2026-01-03
fix: Type narrowing not working with dict with literal keys

### LINKED ISSUE #1616: Type narrowing not working with dict with literal keys
### Describe the Bug

In the following example, I expected `reveal_type(x)` to output `int`, but the actual output was `dict[Literal['value'], int] | int`.
As a result, the type checker gives a `[bad-return]` error.

```py
from typing import Literal, reveal_type

def get_value(x: dict[Literal['value'], int] | int) -> int | None:
    if isinstance(x, dict):
        return x.get('value')
    reveal_type(x) # Revealed type is `dict[Literal['value'], int] | int`, but it should be `int`
    return x
```

(This example works as expected in Pyright.)

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn0%2BjA0pYvVGKkxsMCABlHGl2gplqsNlsFQZqAAVHiqQywoZwfREso6DhrJbtCwwNY2oQjMLbGr8iCnADacA41IiHAAutQAD6R-Z1AB8kbj1AAcrEYLVfNQ8xAykojhwMIwdvgFALTrsc3m63WRMDfERndsAORuqAetv03N5w3tY19f37GjJ1OoZT3E5OTD6euNorN000ZIymBUhTRVSYXDwajoXCvW0O0S4lqu92ew8u72%2BrT%2BxCBkM6pxCdrBjtXttRhShaNpv%2BiYpv%2BaaZug2bznWBYqHAxaluWlZBhwNZQfW-YtEu1Ati07adt2vYLhuRqbMO%2BCjtQABKxHtBMNywdQAAGVYcMGr56lAn74TAP5-i8Mbxv%2BjEKNgAQqIEriWFA1zYOwjFCWhbKYUIy7oKaIByCAATQHAJDkIgVBzF0UCOBIYAcqcZLoPaalYGeYCwiwBRrOgFgsLJQiPgm1DjmGQi1g2ym%2BGAuggOmbkeU%2BwD4AAvqF%2BgaSAZAiGIpCEFwLBQBQNCCClIqVLI%2BCZOgkBaEUBRWfEZqSDA7A8BwHDEHAiAAPQtclojiIQsJaC1MDoC1u6MHALXMCVEBle%2BXSxC1uSwmobrQKg2A5GNpXldNvi4MQm26foZDcugdRUnBPIALzUKFADMhAAIwAEzxegIAxZpUxdCsABi0AwBQBXVHpL1AA

_No response_

tannguyencse19:
# Summary

Update negative `isinstance` narrowing to drop a union member when it is a runtime instance of the checked class:
- For `Type::ClassType`, use class hierarchy checks (`has_superclass`) on the underlying class objects.
- For `Type::TypedDict` and `Type::PartialTypedDict`, treat them as `dict` at runtime.
- Fall back to the old `is_subset_eq` check as an additional removal condition.

Fixes #1616

# Test Plan

```
cargo test -p pyrefly -- test_dict_literal_key_isinstance_narrowing
```

### DIFF
--- a/pyrefly/lib/test/narrow.rs
+++ b/pyrefly/lib/test/narrow.rs
@@ -1457,6 +1457,18 @@ def f(a: A | list[int]):
 );
 
 testcase!(
+    test_dict_literal_key_isinstance_narrowing,
+    r#"
+from typing import Literal, reveal_type
+def get_value(x: dict[Literal["value"], int] | int) -> int | None:
+    if isinstance(x, dict):
+        return x.get("value")
+    reveal_type(x) # E: revealed type: int
+    return x
+    "#,
+);
+
+testcase!(
     test_nested_or_with_multiple_vars,
     r#"
 from typing import assert_type


tannguyencse19:
> Looks reasonable to me, thanks for the PR! I'll try to get a second review and get it merged next week

Thank you

stroxler:
This one may not wind up making it in. I dug around some more and realized that the actual problem is much deeper in Pyrefly, it has to do with the use of `Var` and the way that we pin `Var`s causing `is_subset_eq` to misbehave on invariant types with literal type arguments.

The fix works, but is probably not solving the root cause in the way we need for long-term project health; I think what we actually need to do is change some patterns codebase-wide, including in narrowing

stroxler:
We have a different fix that better addresses the root cause, which involved literal pinning of `Var`. But thanks for the PR, it's good work!

PR #1993, opened 2026-01-03
Support module-level __getattr__ in stubs for from-imports fixes

### LINKED ISSUE #1988: Feature request: support module-level `__getattr__` in stubs
### Describe the Bug

A top-level `__getattr__` definition in a stub file indicates that the file is incomplete and any name can be accessed (https://typing.python.org/en/latest/guides/writing_stubs.html#incomplete-stubs). Pyrefly doesn't currently respect this.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEYuudENxulALnQBogA0IMpRhgopQp1pQKAYjoAFUiLGk6aLHnx0AxrnSQA5gFdKqThH2EAOunkBlGDDoALTp2JxEAem-DR4oQcht4w6N6YuDpw3noGECZmFvreDBx0qABuqNCo2LC6%2Bkam5pbodLjEyehwNuhknC76ALSZMJRwZXQAvHTWIADMhACMAEz9tgLgTCSkEBRg1PScpMQQ6IYsbBzcAILopNaUtpiidAD654YwnOaclJcAFOioNDAAlHTNAHx0%2B6SIOiEYEgAC%2BglQOgsbQAYtAYBQNDgCLMwUA

_No response_

AryanBagade:
## Summary
When a stub file has a top-level `__getattr__` function, it indicates the stub is incomplete and any name can be imported from that module. This is documented in the [typing spec for incomplete stubs](https://typing.python.org/en/latest/guides/writing_stubs.html#incomplete-stubs)

Previously, `from module import name` would fail with "Could not import `name` from `module`" even when `__getattr__` was defined. Now, the imported name correctly gets the return type of `__getattr__`.
**Example:**
  ```python
  # foo.pyi
  def __getattr__(name: str) -> int: ...

  # main.py
  from foo import x, y  # Previously failed, now works
  assert_type(x, int)   # x is correctly typed as int
```

Fixes #1988

# Test Plan
Test Plan
- Added test_module_getattr_from_import - tests basic from foo import x with __getattr__
- Added test_module_getattr_stub_incomplete - tests .pyi stub with Incomplete return type
- Verified existing test_module_getattr still passes (tests import foo; foo.x)
- Ran conformance tests - all pass
- Manually tested with CLI: pyrefly check on test files shows 0 errors

CleanShot 2026-01-02 at 19 25 24@2x

### DIFF
--- a/pyrefly/lib/binding/stmt.rs
+++ b/pyrefly/lib/binding/stmt.rs
@@ -1191,30 +1191,31 @@ impl<'a> BindingsBuilder<'a> {
                         self.error_multiline(x.range, ErrorInfo::Kind(ErrorKind::Deprecated), msg);
                     }
                     Binding::Import(m, x.name.id.clone(), original_name_range)
-                } else if exported.contains_key(&dunder::GETATTR) {
-                    // Module has __getattr__, which means any attribute can be accessed.
-                    // See: https://typing.python.org/en/latest/guides/writing_stubs.html#incomplete-stubs
-                    Binding::ImportViaGetattr(m, x.name.id.clone())
                 } else {
+                    // Try submodule lookup first, then fall back to __getattr__
                     let x_as_module_name = m.append(&x.name.id);
                     let (finding, error) = match self.lookup.get(x_as_module_name) {
                         FindingOrError::Finding(finding) => (true, finding.error),
                         FindingOrError::Error(error) => (false, Some(error)),
                     };
-                    let error = error.is_some_and(|e| matches!(e, FindError::NotFound(..)));
-                    if error {
-                        self.error(
-                            x.range,
-                            ErrorInfo::Kind(ErrorKind::MissingModuleAttribute),
-                            format!("Could not import `{}` from `{m}`", x.name.id),
-                        );
-                    }
+                    let is_not_found = error.is_some_and(|e| matches!(e, FindError::NotFound(..)));
                     if finding {
                         Binding::Module(x_as_module_name, x_as_module_name.components(), None)
-                    } else if error {
-                        Binding::Type(Type::any_error())
+                    } else if exported.contains_key(&dunder::GETATTR) {
+                        // Module has __getattr__, which means any attribute can be accessed.
+                        // See: https://typing.python.org/en/latest/guides/writing_stubs.html#incomplete-stubs
+                        Binding::ImportViaGetattr(m, x.name.id.clone())
                     } else {
-                        Binding::Type(Type::any_explicit())
+                        if is_not_found {
+                            self.error(
+                                x.range,
+                                ErrorInfo::Kind(ErrorKind::MissingModuleAttribute),
+                                format!("Could not import `{}` from `{m}`", x.name.id),
+                            );
+                            Binding::Type(Type::any_error())
+                        } else {
+                            Binding::Type(Type::any_explicit())
+                        }
                     }
                 };
                 // __future__ imports have side effects even if not explicitly used,
--- a/pyrefly/lib/test/attributes.rs
+++ b/pyrefly/lib/test/attributes.rs
@@ -971,6 +971,26 @@ assert_type(y, Any)
 );
 
 testcase!(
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by AryanBagade** [comment]
i see errors in AutoSplit, black, pydantic these occur because their dependency stubs have `__getattr__` returning `object`/`list` instead of `Any`. When we return that type, subsequent attribute access fails.

This is technically correct behavior —> we return whatever `__getattr__` declares. The intended use case (stubs with `-> Incomplete` or `-> Any`) works as expected.

Let me know if you'd prefer we only apply this fallback when `__getattr__` returns `Any`/`Incomplete`.

File: pyrefly/lib/test/attributes.rs [hunk 1]
```diff
@@ -940,6 +940,36 @@ del foo.y  # E: No attribute `y` in module `foo`
...
+    ),
+    r#"
+from typing import assert_type, Any
+from foo import x, y
+# Incomplete is essentially Any, so x and y should be Any
+assert_type(x, Any)
+assert_type(y, Any)
+    "#,
+);
+
```
rchen152, line 972: Could you add a test case for a `foo.pyi` that has both a `__getattr__` and a module-level variable with a different type, to check that we can look up the variable's type and don't incorrectly use the `__getattr__`?

pyrefly/lib/binding/stmt.rs
```diff
@@ -1190,6 +1191,10 @@ impl<'a> BindingsBuilder<'a> {
                         self.error_multiline(x.range, ErrorInfo::Kind(ErrorKind::Deprecated), msg);
                     }
                     Binding::Import(m, x.name.id.clone(), original_name_range)
+                } else if exported.contains_key(&dunder::GETATTR) {
```
rchen152: This obscures imports of submodules ([example](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRhq9MLlyduvftlSU5gkWMnMYACi2UAlHToBiOgEEwDGJToAFAEo2AjAE4fAZgAaRgALCDhGIzoAYxCMNnhGVQADVgZkkECQMkEwKFJCBlooCltXUlz8ujQsPHwY3HkINgBXSlQGCEbCOVsAZRgYOhCGBmI4RAB6SZyYPILeNkmYdEnMXGi4SejGyFb2zsbJumUXVCFUaFRsWAam-Y6u9DpcYkP0OB70MgYQxoBaESUOBPOgAXjoMhA-kIXgATFC5JlwCpJuJxKwIAx0SRSBAKJg5nR0QkGB0GJR0cZ0KgaDBLP8AHycdAMRB0Qic5HKXCTMy4-GIEAgAC%2BWVQ0U6IgAYtAYBQajgCLjRUA)). I tested mypy and pyright, and in that example, both still look up module `foo.bar` rather than using the `__getattr__` in `foo/__init__.pyi`. Could you change this PR to let submodule lookup take precedence?

rchen152:
> i see errors in AutoSplit, black, pydantic these occur because their dependency stubs have `__getattr__` returning `object`/`list` instead of `Any`. When we return that type, subsequent attribute access fails.
> 
> This is technically correct behavior —> we return whatever `__getattr__` declares. The intended use case (stubs with `-> Incomplete` or `-> Any`) works as expected.
> 
> Let me know if you'd prefer we only apply this fallback when `__getattr__` returns `Any`/`Incomplete`.

Always applying the fallback seems like correct behavior to me, no need to change it!

pyrefly/lib/test/attributes.rs
AryanBagade: sure

pyrefly/lib/binding/stmt.rs
AryanBagade: sure

grievejia (approved):

PR #2015, opened 2026-01-06
Do not allow `Callable` to be passed as an argument of `type`

fangyi-zhou:
# Summary

According to the spec:
https://typing.python.org/en/latest/spec/special-types.html#type
> Any other special constructs like `tuple` or `Callable` are not allowed as an argument to type.

Therefore, this diff disallows `Callable` from being passed to `type` in the subtyping implementation, whereas permitting its use in calls to `isinstance`.

P.S. The conformance testsuite and the spec seems to disagree on whether `tuple` should be allowed, which I filed an issue in the typing repository.

We should consider restricting other special forms to be passed as arguments to `type` as well, but the rules regarding special forms are a bit not clear to me.

# Test Plan

`python test.py`

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -9853,6 +9853,17 @@
     },
     {
       "code": -2,
+      "column": 7,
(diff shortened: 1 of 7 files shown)


rchen152 (approved):

PR #2017, opened 2026-01-07
Fix TypeGuard/TypeIs validation for functions without explicit return annotation

### LINKED ISSUE #1998: Functions returning a TypeGuard value are incorrectly propagating the TypeGuard as their return type
### Describe the Bug

```py
from typing import TypeGuard


def is_int(x: int | str) -> TypeGuard[int]:
    return isinstance(x, int)


class X:
    def __init__(self, param: int | str):
        self.param = param

    def has_int(self):
        return is_int(self.param)
```

This reports `X::has_int` as an invalid TypeGuard function, because it is detected to return a TypeGuard, and has no parameters. Seems to work fine if the function does not always return the typeguard.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjADiAV1SVMAHXSzZmGGE5wA%2BqwYAKfIk7p%2BAHzpwGlAJR0AtAD5Bw8ZMwBtDQF1Esul7qUYDMZToKqwmGADGMNoANHoMZvJy6GFQqHBwdAAaHkHeisqq6ugQDAWacDBQYDHEkqg0uhp0Ribm2d7txhVgJLX0ALx0NZR1Ce15dAAWqYVa5ZVmbR1evv6BKjNlXT3DNPHoIFEgYgzQcCTkiCAAxHQAqidQxaR0YGJJJ7jocAnjYLw0qBK6DENGwMEo2ga%2BgsNmMpkWyz8ASCYGkIAAciCwZRdMB8ABfNGyA4gMi%2BMBQUiEBi0KAUG4ABVI5MpxgwOAIdDCn0gbACgIgn0IshuAGUYDBJgwGMQ4IgAPTyslKSmEXhseUwdDyzC4MJweXc9C8-kfbUvXh0VAAN1Q0FQ2FgXJ5ED5wzNdFwxDNZ1kZAYE0%2Blmt4LggqCAzRAGZCABGABMRP2%2BMOqDCJxDADFoDAKGgsHgiGQQPigA

_No response_

AryanBagade:
# Summary
Functions returning a TypeGuard/TypeIs value were incorrectly being validated as type guard functions even when they had no explicit return annotation.

For example:
```python
  from typing import TypeGuard

  def is_int(x: int | str) -> TypeGuard[int]:
      return isinstance(x, int)

  class X:
      def __init__(self, param: int | str) -> None:
          self.param = param

      def has_int(self):
          return is_int(self.param)
```
has_int was being reported as an invalid TypeGuard function ("Type guard functions must accept at least one positional argument") because Pyrefly inferred its return type as TypeGuard[int] and then validated it as a type guard function.

The fix ensures TypeGuard/TypeIs validation only runs when the function has an explicit return annotation.

Fixes #1998

# Test Plan
- Added test_typeguard_return_without_annotation and test_typeis_return_without_annotation test cases
- Verified existing TypeGuard/TypeIs validation tests still pass (test_typeguard_argument_number, test_typeis_argument_number, test_typeis_subtyping)

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -531,25 +531,30 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 );
             }
         }
-        if matches!(&ret, Type::TypeGuard(_) | Type::TypeIs(_)) {
-            self.validate_type_guard_positional_argument_count(
-                &def.params,
-                def.id_range(),
-                &def.defining_cls,
-                def.metadata.flags.is_staticmethod,
-                errors,
-            );
-        };
+        // Only validate TypeGuard/TypeIs functions when they have an explicit return annotation.
+        // Functions that return a TypeGuard value without an explicit annotation should not be
+        // treated as TypeGuard functions.
+        if has_return_annotation {
+            if matches!(&ret, Type::TypeGuard(_) | Type::TypeIs(_)) {
+                self.validate_type_guard_positional_argument_count(
+                    &def.params,
+                    def.id_range(),
+                    &def.defining_cls,
+                    def.metadata.flags.is_staticmethod,
+                    errors,
+                );
+            }
 
-        if let Type::TypeIs(ty_narrow) = &ret {
-            self.validate_type_is_type_narrowing(
-                &def.params,
-                stmt,
-                &def.defining_cls,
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

PR #2021, opened 2026-01-07
fix panic: Name not found in static scope of module __unknown__ #1987

### LINKED ISSUE #1987: panic: Name not found in static scope of module `__unknown__`
### Describe the Bug

```python
async = 1
```

```
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/binding/bindings.rs:1094:17:
Name `` not found in static scope of module `__unknown__`.
```

### Version info
0.46.3

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1987
Fixes https://github.com/facebook/pyrefly/issues/2047

Implemented safe handling for parser-synthesized targets so keyword assignments no longer panic.

treat all empty identifiers as synthesized so later passes skip binding logic.

route synthesized lvalues and empty Identifiers through anonymous bindings, and short-circuit bind_single_name_assign when parser recovery produces an empty target, ensuring we still analyze RHS expressions without touching static scopes.

# Test Plan

added regression test covering async = 1 and documented Ruff’s diagnostic wording.

### DIFF
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -407,13 +407,10 @@ impl<'a> BindingsBuilder<'a> {
                 }
                 self.scopes.push(Scope::comprehension(range, is_generator));
             }
-            let key = Key::Anon(comp.iter.range());
-            let iterable_value_idx = self.existing_binding_idx(&key).unwrap_or_else(|| {
-                self.insert_binding(
-                    key,
-                    Binding::IterableValue(None, comp.iter.clone(), IsAsync::new(comp.is_async)),
-                )
-            });
+            let iterable_value_idx = self
+                .insert_binding_if_missing(Key::Anon(comp.iter.range()), || {
+                    Binding::IterableValue(None, comp.iter.clone(), IsAsync::new(comp.is_async))
+                });
             self.scopes.add_lvalue_to_current_static(&comp.target);
             // A comprehension target cannot be annotated, so it is safe to ignore the
             // annotation (which is None) and just use a `Forward` here.
--- a/pyrefly/lib/binding/target.rs
+++ b/pyrefly/lib/binding/target.rs
@@ -335,10 +335,9 @@ impl<'a> BindingsBuilder<'a> {
                     self.ensure_expr(assigned, &mut usage);
                 }
                 let key = Key::Anon(illegal_target.range());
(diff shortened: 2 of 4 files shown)


pyrefly/lib/test/semantic_syntax_errors.rs
```diff
@@ -119,6 +119,13 @@ match p:
 "#,
 );
 
+testcase!(
+    test_keyword_assignment_reports_parse_error,
+    r#"
+async = 1  # E: Parse error: Expected `def`, `with` or `for` to follow `async`, found `=`
+"#,
```
migeed-z, line 128: Can we add the testcase in https://github.com/facebook/pyrefly/issues/2047 as well to make sure this PR fixes it?

pyrefly/lib/binding/target.rs
```diff
@@ -455,6 +460,13 @@ impl<'a> BindingsBuilder<'a> {
         mut value: Box<Expr>,
         direct_ann: Option<(&Expr, Idx<KeyAnnotation>)>,
     ) -> Option<Idx<KeyAnnotation>> {
+        if name.id.as_str().is_empty() {
```
migeed-z: The suggestion by copilot here looks legit. Could we commit it?

pyrefly/lib/binding/expr.rs
```diff
@@ -404,12 +404,13 @@ impl<'a> BindingsBuilder<'a> {
...
                 self.scopes.push(Scope::comprehension(range, is_generator));
             }
-            // Incomplete nested comprehensions can have identical iterators
-            // for inner and outer loops. It is safe to overwrite it because it literally the same.
-            let iterable_value_idx = self.insert_binding_overwrite(
-                Key::Anon(comp.iter.range()),
-                Binding::IterableValue(None, comp.iter.clone(), IsAsync::new(comp.is_async)),
-            );
+            let key = Key::Anon(comp.iter.range());
+            let iterable_value_idx = self.existing_binding_idx(&key).unwrap_or_else(|| {
```
migeed-z: is the `existing_binding_idx` helper function, doing something similar to [`insert_if_missing`](https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_graph/src/index_map.rs#L76)? we may be able to use that instead.

stroxler (approved):

Issue #630, opened 2025-07-03
Feature Request: support `__slots__`


rchen152:
### Describe the Bug

When a class defines `__slots__`, pyrefly shouldn't allow writing to attributes that aren't specified in `__slots__`. This is tested in the dataclasses_slots conformance test, which we currently fail:
https://github.com/facebook/pyrefly/blob/47f5efe3397a6f9881e5b3398fcd86dc96ed9158/conformance/third_party/dataclasses_slots.py#L24-L25

### Sandbox Link

https://pyrefly.org/sandbox/?code=MYGwhgzhAECCBcAoaLoH00RAewC4Q2gF5oAKAcgA9yAaASmVQBMBTAM3TQEsA7L3DKQgsQbOklSTow0QDpKxaABYATIynSRbWQE9FqlAGJoAFQAWXGBDPYAriCY9yuaACMW0MCBwB3FkyA

yangdanny97:
I've been thinking about this and `__match_args__` which has similar properties.

I was thinking we should try and compute these while solving class metadata and store it as something better than `Type::Tuple([Type::Literal(...), Type::Literal(...)]`.

Adist319:
I can take a look at this!

rchen152:
Assigned, thank you!

Adist319:
PR created here: #2029

migeed-z:
> PR created here: [#2029](https://github.com/facebook/pyrefly/pull/2029)

Thanks a lot! As we discussed, this PR still has some open questions regarding whether the errors it introduced are false positives or not and had not been active for sometime, unless @Adist319 would still like to look into the false positives, I can take it and/or open to the community

PR #2029, opened 2026-01-07
Add __slots__ validation for attribute assignments

Adist319:
# Summary

Validate that attribute assignments like `self.y = 3` respect `__slots__` restrictions. When a class defines `__slots__`, Pyrefly now errors if you try to assign to an attribute not in the slots.

  Key changes:
  - Add slots check in `ClassAttribute` handling path (not just `check_setattr` which only runs for unfound attributes)
  - Rename `slots_metadata()` to `compute_slots_metadata()` and make it public, called on-demand to ensure fields are fully solved
  - Handle classes without `__slots__` correctly (they get implicit `__dict__` and allow any attributes)

  Addresses #630.

  # Tests

  Added 6 test cases covering:
  - `@dataclass(slots=True)` preventing unknown attributes
  - Manual `__slots__ = ("x",)` preventing unknown attributes
  - `__dict__` in slots allowing dynamic attributes
  - Slot inheritance from parent classes
  - Child class without `__slots__` getting implicit `__dict__`

Open to feedback!


migeed-z:
Thanks for the PR! I will take a look!

pyrefly/lib/alt/attr.rs
```diff
@@ -801,6 +816,85 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        // - If any class in MRO has `__dict__` in slots, allow everything
+        // - If any class in MRO is missing `__slots__`, it has implicit `__dict__`, allow everything
+        // - Otherwise, the attribute must be in the union of all slots in the MRO
+
+        let mut all_slots: SmallSet<Name> = SmallSet::new();
+        let mut has_any_slots = false;
+
+        // Check current class - compute slots metadata on-demand to ensure fields are solved
+        let metadata = self.get_metadata_for_class(class);
+        let slots = self.compute_slots_metadata(class, metadata.dataclass_metadata());
```
migeed-z, line 835: If we stored the slots metadata in the class, why are we computing it here as well?  More generally, why is recomputing needed/can we find a way to compute this only once for each class?

pyrefly/lib/alt/class/class_metadata.rs
```diff
@@ -1254,4 +1261,83 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        let slots_ty = slots_field.ty();
+
+        // Extract string literals from the slots type.
+        let mut slots = SmallSet::new();
+        let mut has_dict_slot = false;
+
+        match &slots_ty {
+            Type::Tuple(Tuple::Concrete(elts)) => {
+                for elt in elts.iter() {
+                    if let Type::Literal(Lit::Str(s)) = elt {
```
migeed-z, line 1308: I am seeing various new errors coming from mypy primer. Curious if there is any sense on whether those are legitimate? 

Also, do we know that at this point, pyrefly is inferring the literal types we expect for slots?

pyrefly/lib/test/dataclasses.rs
```diff
@@ -1569,3 +1569,77 @@ class C:
...
+    __slots__ = ("x",)
+
+class Child(Parent):
+    pass  # No __slots__, gets implicit __dict__
+
+    def set_attrs(self):
+        self.x = 1  # OK
+        self.z = 3  # OK - child has implicit __dict__
+    "#,
+);
```
migeed-z, line 1657: can we add a testcase invoking the init method? 

```
class Base:
    __slots__ = ("x",)
    
    def __init__(self, x: int, flag: bool = False):
        self.x = x

class Child(Base):
    __slots__ = ("y",)
    
    def __init__(self, x: int, y: int):
        super().__init__(x)  
        self.y = y
```
which should succeed?

pyrefly/lib/alt/attr.rs
Adist319: Good catch, I actually tried switching to use the stored `metadata.slots_metadata()` but it broke the tests. I believe the issue is timing: `slots_metadata` gets computed during `class_metadata_of()` which happens early in the solving process, before the `__slots__` field type is fully resolved. At that point the type might still be `Any` or not have the literal strings extracted yet. If we want to avoid recomputation in the future, we'd probably need some kind of lazy/deferred computation pattern (compute on first access rather than during class metadata construction). Could potentially be a followup discussion?

pyrefly/lib/test/dataclasses.rs
Adist319: Yup, added this test case, thanks

pyrefly/lib/alt/class/class_metadata.rs
Adist319: Looked into these, most are false positives. The slots check was running for all attribute types including properties/descriptors, but Python's `__slots__` only restricts instance attributes. For example, dulwich's `Blob.data` and many discord.py attributes are properties, not instance slots.

Fixed by only checking slots for `ReadWrite`/`ReadOnly` class attributes, skipping `Property` and `Descriptor` variants.

For the literal type inference, the code extracts literal strings from the tuple (the `Type::Tuple(Tuple::Concrete(elts))` match).

migeed-z: Thank you! 
Are the current mypy primer results meant to be the latest results? I am still seeing a lot of false positives.

Adist319: Yes, those are the latest results. I'm not completely sure myself which ones are still false positives (after adding the descriptor/property fix), but I believe they appear because slots checking didn't exist before this PR, so they're "new" in the sense that we're now checking something we weren't before.

Open to more feedback for the primer results.

rchen152:
Closing this PR since it's stale. Please feel free to open another one if you're still interested in working on this.

Adist319:
No problem, have an updated PR at #2695

PR #2031, opened 2026-01-08
Allow `typing.Mapping` to form type aliases

fangyi-zhou:
# Summary

- Add `typing.Mapping` as a special export
- Recongnise that subscript in `typing.Mapping` to be a type alias

This fix is targeted to the failing conformance test suite (although I guess recursive aliases are still not working since recursive types are not supported afaik). Nonetheless, I think we need a longer term solution since there are other related "special exports" (e.g. `collections.abc.Mapping`, `typing.MutableMapping`, `typing.Sequence`).

# Test Plan

`python test.py`

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -600,19 +600,7 @@
       "stop_line": 65
     }
   ],
-  "aliases_recursive.py": [
-    {
-      "code": -2,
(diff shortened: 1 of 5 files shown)


fangyi-zhou:
Regenerated the conformance test results, should be good now @stroxler

PR #2035, opened 2026-01-08
impl use function

asukaminato0721:
# Summary

Fixes part of #2207 

Pattern matcher and refactor edits targets single‑return, non‑decorated, non‑async top‑level functions with only positional params and return expressions that reference only those params (no and/or/ternary or chained comparisons to avoid evaluation‑order changes).

Hooked into LSP code actions and exposed as refactor.rewrite .

https://rope.readthedocs.io/en/latest/overview.html#use-function-refactoring

# Test Plan

Added a multi‑module test and updated LSP capability expectations


pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,895 @@
...
+    bindings: HashMap<String, ArgBinding>,
+}
+
+#[derive(Clone, Debug)]
+struct CallTarget {
+    callee: String,
+    import_edit: Option<(TextRange, String)>,
+}
+
+fn find_enclosing_top_level_function(
```
jvansch1, line 238: There is a function in `fbcode/pyrefly/pyrefly/lib/lsp/non_wasm/call_hierarchy.rs` called find_containing_function_for_call which seems like it does something very similar. Can we try to re-use that?

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,895 @@
+/*
```
jvansch1, line 1: I think most of the functions in this file need a docstring. A lot of these functions are quite large and it is not immediately clear what their purpose is.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,895 @@
...
+
+fn module_has_top_level_binding(ast: &ModModule, name: &str) -> bool {
+    for stmt in &ast.body {
+        match stmt {
+            Stmt::FunctionDef(def) => {
+                if def.name.id.as_str() == name {
+                    return true;
+                }
+            }
+            Stmt::ClassDef(def) => {
```
jvansch1: This is the same as the arm above. Can we combine them into one?

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,895 @@
...
+                {
+                    return true;
+                }
+            }
+            Stmt::AnnAssign(ann) => {
+                if matches!(&*ann.target, Expr::Name(ExprName { id, .. }) if id.as_str() == name) {
+                    return true;
+                }
+            }
+            Stmt::AugAssign(aug) => {
```
jvansch1: Same here. This is the same as the arm above so it would be good if we can combine them.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,895 @@
...
+            matches.push(ExprMatch {
+                range: expr_range,
+                bindings,
+            });
+        }
+    });
+    matches
+}
+
+fn match_expr(
```
jvansch1, line 607: Can we break this function up into smaller, easier to read functions. This function is over 300 lines and I think it will be easier to maintain if we can break this into smaller functions which are easier to reason about.

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -1778,6 +1787,312 @@ def caller():
...
+my_var = 3 ** 2
+"#;
+    let mod2 = r#"
+another_var = 4 ** 2
+"#;
+    let (handles, state) = mk_multi_file_state_assert_no_errors(
+        &[("mod1", mod1), ("mod2", mod2)],
+        Require::Everything,
+    );
+    let handle = handles.get("mod1").unwrap();
```
jvansch1: There is a lot of repeated work between these tests. Can we break this out into a helper function that we can reuse?

jvansch1 (changes requested):
Good start! I have a couple suggestions for some things that we can change.

Also, for these code action PRs it would be helpful if you could either include some instructions for how we can test these in our own IDE(we mostly use VScode) or a video showing how they work.

asukaminato0721:
[Screencast_20260204_042421_c.webm](https://github.com/user-attachments/assets/10d48246-1849-4aba-b138-d9da0970c7db)

jvansch1:
@asukaminato0721 Looks like there is a merge conflict here. Once this gets updated I will do another review of this.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -856,6 +856,7 @@ pub fn capabilities(
                 CodeActionKind::REFACTOR_REWRITE,
                 CodeActionKind::new("refactor.move"),
                 CodeActionKind::REFACTOR_INLINE,
+                CodeActionKind::REFACTOR_REWRITE,
```
jvansch1: This was already added up on line 856 so we can remove this.

pyrefly/lib/state/lsp/ast_helpers.rs
```diff
@@ -0,0 +1,42 @@
...
+/// A function definition plus its enclosing class (if any).
+pub(crate) struct ContainingFunction<'a> {
+    pub function_def: &'a StmtFunctionDef,
+    pub class_def: Option<&'a StmtClassDef>,
+}
+
+/// Finds the innermost function definition covering `position`.
+///
+/// Returns the function and its enclosing class when the function is a method.
+pub(crate) fn find_containing_function_def(
```
jvansch1, line 24: Thanks for refactoring this!

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+    let selection_start = selection.start();
+    let function_def = find_enclosing_top_level_function(ast.as_ref(), selection_start)?;
+    if function_def.is_async || !function_def.decorator_list.is_empty() {
+        return None;
+    }
+    let params = collect_supported_params(&function_def)?;
+    let return_expr = extract_return_expr(&function_def)?;
+    let mut param_set = HashSet::new();
+    for param in &params {
+        param_set.insert(param.clone());
```
jvansch1: Instead of looping manually here could we just do something like params.iter().cloned().collect::<HashSet<_>>()?

I also wonder why we need this represented as both a vector and a hashset. Could we get away with just having one of them and saving some memory?

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+        let target_info = match transaction.get_module_info(&target_handle) {
+            Some(info) => info,
+            None => continue,
+        };
+        if transaction.is_third_party_module(&target_info, &target_handle)
+            && !transaction.is_source_file(&target_info, &target_handle)
+        {
+            continue;
+        }
+        let target_ast = match transaction.get_ast(&target_handle) {
```
jvansch1: Can we just use let Some() syntax here or even `?` syntax?

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+            Some(ast) => ast,
+            None => continue,
+        };
+        let call_target =
+            match call_target_for_module(&target_handle, target_ast.as_ref(), &pattern, handle) {
```
jvansch1: Same here. I don't think we need the full match statement written out here.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+                Some(target) => target,
+                None => continue,
+            };
+        let same_module = target_handle.module() == pattern.module_name;
+        let min_start = if same_module {
```
jvansch1, line 160: Can we use `let Some()`?

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+use crate::state::lsp::ast_helpers::find_containing_function_def;
+
+/// Builds use-function refactor actions for the supplied selection.
+pub(crate) fn use_function_code_actions(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    selection: TextRange,
+) -> Option<Vec<LocalRefactorCodeAction>> {
+    let module_info = transaction.get_module_info(handle)?;
+    if transaction.is_third_party_module(&module_info, handle)
```
jvansch1, line 65: We don't have to do it in this PR, but it seems like this a check we are doing often. We check `transaction.is_third_party_module(&target_info, &target_handle) && !transaction.is_source_file(&target_info, &target_handle)`  twice in this function and we also do it in lib/state/lsp.rs. At this point we can probably extract this out to a helper function.

Again not something that I think needs to be done in the PR but something that would improve code quality a bit.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+        kind: CodeActionKind::REFACTOR_REWRITE,
+    }])
+}
+
+#[derive(Clone, Debug)]
+struct FunctionPattern {
+    name: String,
+    params: Vec<String>,
+    param_set: HashSet<String>,
+    param_counts: HashMap<String, usize>,
```
jvansch1, line 210: Do we need all these representations of params? I can see how having the counts is useful, but wonder if we really need to have both params as a vector and as a set.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+    None
+}
+
+/// Finds the local alias for a module import (`import module as alias`).
+fn find_imported_module_alias(ast: &ModModule, module_name: &str) -> Option<String> {
+    for stmt in &ast.body {
+        let Stmt::Import(StmtImport { names, .. }) = stmt else {
+            continue;
+        };
+        for alias in names {
```
jvansch1, line 463: This is very similar to the block on line 405. Maybe a small helper function should be introduced to reduce repeated code? I know we are comparing a module in one case and a function in another case, so if it becomes too complicated to make it one function then I think it is fine to leave as is.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,998 @@
...
+    lhs.len() == rhs.len()
+        && lhs
+            .iter()
+            .zip(rhs.iter())
+            .all(|(lhs, rhs)| match_expr(lhs, rhs, param_set, param_counts, bindings, module_info))
+}
+
+/// Matches dict expressions by comparing key/value pairs.
+fn match_dict_expr(
+    lhs: &ExprDict,
```
jvansch1, line 871: It looks like we have a lot of functions that do very similar things here with the main difference being the type of `lhs` and `rhs`. Maybe we could make this a generic function that has a match statement that could properly handle the logic for different types of `lhs` and `rhs`?

jvansch1 (changes requested):

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
asukaminato0721: We need a stable argument order to build call text, so the Vec is necessary.

The HashSet is used for fast membership checks in match_expr/validate_return_expr.

I kept both for clarity and performance. If we really wanted to reduce fields, we could drop the set and use params.contains(name) everywhere, but that becomes O(n) per check and makes the matching logic noisier.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
asukaminato0721: A single generic would likely recreate a big match internally

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,993 @@
...
+        return None;
+    }
+    match body[0] {
+        Stmt::Return(ret) => ret.value.as_deref(),
+        _ => None,
+    }
+}
+
+/// Validates that `expr` is a matchable return expression and counts param uses.
+fn validate_return_expr(
```
jvansch1, line 295: I noticed that this function does not have a match arm for "BooOp". I think what would happen is that if we have a BoolOp in this case, we will always fall back to {} => false. Then in match_expr_nodes the "BoolOP" arm is unreachable since a BoolOp will never be validated in this function.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
jvansch1: I think a match might be preferable In this case. I guess we either need to have

1. A lot of smaller functions like this
2. One huge function containing all the match arms

I don't think either great but if we had to choose one I think what we have here is better since at least we have named functions which describe what is happening. Overall there is a done of duplicate logic so it would be great to see if we could get rid of that if possible.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
jvansch1: @asukaminato0721 Still wondering about this.

jvansch1 (changes requested):
Can we also add some tests for the following cases:
1. A function with zero arguments
2. A function that has an argument with a default value.

jvansch1:
@asukaminato0721 I did see an issue with import when using this functionality. It seems like we are using the full import name rather than just importing the function we need. It would be nice if we could fix this before we merge this in.

https://github.com/user-attachments/assets/abba8c4e-2c02-4b76-b481-84a0068e5a7e

kinto0 (changes requested):

jvansch1:
@asukaminato0721 Once all the CI checks are passing I will continue to review this PR.

pyrefly/lib/state/lsp/quick_fixes/use_function.rs
```diff
@@ -0,0 +1,977 @@
...
+}
+
+/// Finds the top-level function that covers `position`, if any.
+///
+/// This reuses the same AST traversal as call hierarchy but filters out
+/// methods and nested functions.
+fn find_enclosing_top_level_function(
+    ast: &ModModule,
+    position: TextSize,
+) -> Option<StmtFunctionDef> {
```
jvansch1: I think this will clone the AST node which would be good to avoid. Could we instead do something like 

```
fn find_enclosing_top_level_function<'a>(
    ast: &'a ModModule,
    position: TextSize,
) -> Option<&'a StmtFunctionDef> {
```

asukaminato0721:
```
thread 'tests::test_get_source_db_always_configures_paths' (6788) panicked at crates\pyrefly_build\src\lib.rs:209:9:
assertion `left == right` failed
  left: ["D:\\root\\path\\to\\project", "D:\\absolute\\path\\to\\project"]
 right: ["/root\\path/to/project", "/absolute/path/to/project"]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

the ci err is strange

jvansch1 (changes requested):

jvansch1:
@asukaminato0721 I am noticing that this functionality is not working when I would expect it to. See the example video below.

I am also wondering how this would work if we have multiple projects in a workspace. If we call `use_function` in one project, will this also try to replace similar instances for another project in the workspace?


https://github.com/user-attachments/assets/3bf88f8a-291e-4f0e-8e96-2ed7885ea2aa

asukaminato0721:
The “use function” code action now rewrites expression statements inside function bodies (like `x ** 2` in other) by traversing expressions via statements instead of only visiting top-level expressions.

Added a regression test.

jvansch1:
@asukaminato0721 I ran into a few other issues when testing

I created very simple function:

```
def one():
    return 1
```

1. We try to use this in several places where it does not seem appropriate including in typeshed stub files
Screenshot 2026-02-18 at 9 37 59 AM

We also use it in places where it does not seem totally appropriate. In this case there was a hardcoded value of `1` and we replaced it with a function. I have questions about how useful that actually is.
Screenshot 2026-02-18 at 9 41 37 AM

2. This check happens across all projects in a workspace which I do not think is correct.

Say for example I have two projects in the same VSCode workspace. One project is called `main` and one is called `other`

In `main` I defined this `one()` function. When I call use_function on this it also replaces code inside the `other` project. This is incorrect since these are entirely different projects that just happen to be in the same workspace instance. 

These issues are both pretty critical and would need to be fixed before moving forward with this change.

asukaminato0721:
> In this case there was a hardcoded value of 1 and we replaced it with a function. I have questions about how useful that actually is.

I think skip trivial constant-return functions could solve this?

jvansch1:
@asukaminato0721 I am going to try and finalize this review as well. Could you resolve the merge conflict when you have a chance?

asukaminato0721:
done

asukaminato0721:
done

Issue #364, opened 2025-05-26
Complicated refactoring support


neel04:
### Describe the Bug

Are there any plans for adding extensive refactoring support, similar to the ROPE plugin for PyLSP?

lolpack:
Thanks for reaching out! I've copied the features here for tracking the individual ones:

**Rename:**

- [x] implemented: variables, classes, functions (disabled by default)
- [ ] coming soon: modules, packages (disabled by default)

**Code Action:**

- [x] extract method
- [x] extract variable
- [ ] inline method/variable/parameter
- [ ] use function
- [ ] method to method object
- [ ] convert local variable to field
- [ ] organize imports
- [ ] introduce parameter
- [ ] generate variable/function/class from undefined variable

kinto0:
hi @neel04, we added basic rename support [in 0.22.0](https://github.com/facebook/pyrefly/releases/tag/0.22.0), please try it out and let us know your thoughts!

i'll rename this issue to "complicated refactoring"

lolpack:
PyCharm refactoring features: https://www.jetbrains.com/help/pycharm/refactoring-source-code.html

- [ ] Change signature
- [ ]  Convert to Package and Module
- [ ]  Extract/Introduce refactorings
- [ ]  Inline
- [ ]  Invert boolean
- [ ]  Copy and move refactorings
- [ ]  Pull members up, push members down
- [ ]  Rename refactorings
- [ ]  Safe delete

asukaminato0721:
ref1 https://www.jetbrains.com/help/pycharm/refactoring-source-code.html 
ref2 https://rope.readthedocs.io/en/latest/overview.html#refactorings

asukaminato0721:
PyCharm refactoring features: https://www.jetbrains.com/help/pycharm/refactoring-source-code.html

- [ ] Change signature
- [x] Convert to Package and Module
- [x] Introduce refactorings #2037
- [x] Extract #1885
- [x] Inline #2033
- [ ] Invert boolean #2060
- [x] Copy and move refactorings
- [x] Pull members up, push members down
- [ ] Rename refactorings
- [ ] Safe delete

yangdanny97:
It would be great if we documented the full set of refactoring features we support on the website, and added to it as we go.

I wonder if it should be put on a separate page from https://pyrefly.org/en/docs/IDE-features/ since that's getting kind of long

cc @kinto0 for opinions

kinto0:
[...]

PR #2038, opened 2026-01-08
impl push up/push down

asukaminato0721:
# Summary

Fixes part of #364

https://www.jetbrains.com/help/pycharm/pull-members-up.html#push_down

Implemented pull‑up and push‑down member refactor actions (including a “push to all subclasses” option) in the LSP quick‑fix pipeline, using new move-member helpers that reindent and edit class bodies safely. Added code‑action wiring and capability exposure for refactor.move.

# Test Plan

introduced unit tests covering basic pull/push and the all‑subclasses variant.

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/move_members.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/move_members.rs
@@ -29,36 +29,36 @@ pub(crate) fn pull_members_up_code_actions(
     handle: &Handle,
     selection: TextRange,
 ) -> Option<Vec<LocalRefactorCodeAction>> {
-    let module_info = transaction.get_module_info(handle)?;
-    let ast = transaction.get_ast(handle)?;
-    let source = module_info.contents();
-    let selection_point = selection_anchor(source, selection);
-    let (class_def, member_stmt) = find_member_context(ast.as_ref(), selection_point)?;
-    let member_name = member_name_from_stmt(member_stmt)?;
-    let member_indent = line_indent_and_start(source, member_stmt.range().start())?.0;
-    let base_classes = base_classes_in_module(ast.as_ref(), class_def);
-    let mut actions = Vec::new();
-    for base_class in base_classes {
-        if class_has_member_named(base_class, &member_name) {
-            continue;
-        }
-        if let Some(action) = build_move_action(
-            &module_info,
-            member_stmt,
-            &member_indent,
-            class_def,
-            base_class,
-            format!("Pull `{member_name}` up to `{}`", base_class.name.id),
-        ) {
-            actions.push(action);
-        }
-    }
-    actions.sort_by(|a, b| a.title.cmp(&b.title));
-    if actions.is_empty() {
-        None
-    } else {
-        Some(actions)
-    }
+    with_move_context(
+        transaction,
+        handle,
+        selection,
+        |module_info, ast, class_def, member_stmt, member_name, member_indent| {
+            let base_classes = base_classes_in_module(ast, class_def);
+            let mut actions = Vec::new();
+            for base_class in base_classes {
+                if class_has_member_named(base_class, member_name) {
+                    continue;
+                }
+                if let Some(action) = build_move_action(
+                    module_info,
+                    member_stmt,
+                    member_indent,
+                    class_def,
+                    base_class,
+                    format!("Pull `{member_name}` up to `{}`", base_class.name.id),
+                ) {
+                    actions.push(action);
+                }
+            }
+            actions.sort_by(|a, b| a.title.cmp(&b.title));
+            if actions.is_empty() {
+                None
+            } else {
+                Some(actions)
+            }
+        },
+    )
 }
 
 /// Builds push-members-down refactor actions for the supplied selection.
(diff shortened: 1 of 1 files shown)


kinto0:
thank you for working on all of these!! super excited about these features. I don't think I have time to take a look until at least tomorrow (but someone else might be able to earlier). 

In the meantime, is it possible to add these to the [IDE supported features](https://pyrefly.org/en/docs/IDE-features/) page with descriptions and videos?

The way I have been adding those is by compressing the screen recording like the test plan [here](https://github.com/facebook/pyrefly/commit/730d0138804e39cf098c2ef24114b07d02e0b77b)

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -632,6 +632,7 @@ pub fn capabilities(
             code_action_kinds: Some(vec![
                 CodeActionKind::QUICKFIX,
                 CodeActionKind::REFACTOR_EXTRACT,
+                CodeActionKind::new("refactor.move"),
```
jvansch1, line 635: Could we add a REFACTOR_MOVE variant to the CodeActionKind enum rather than creating it inline here?

pyrefly/lib/state/lsp/quick_fixes/move_members.rs
```diff
@@ -0,0 +1,470 @@
...
+    let has_docstring = class_def.body.first().is_some_and(is_docstring_stmt);
+    let mut non_docstring = class_def
+        .body
+        .iter()
+        .filter(|stmt| !is_docstring_stmt(stmt));
+    let only_stmt = non_docstring.next();
+    let is_only_non_docstring = non_docstring.next().is_none()
+        && only_stmt.is_some_and(|stmt| stmt.range() == member_stmt.range());
+
+    if has_docstring {
```
jvansch1: I am a little confused why we have this if statement since if we pass this we are returning the same value anyway.

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -170,6 +172,50 @@ fn assert_no_extract_action(code: &str) {
...
 }
 
+fn compute_pull_up_actions(
+    code: &str,
+) -> (
+    ModuleInfo,
+    Vec<Vec<(Module, TextRange, String)>>,
+    Vec<String>,
+) {
+    let (handles, state) =
```
jvansch1, line 189: I wonder if we can pull some of this out into a shared function that can be re-used? It looks like the first couple lines of compute_pull_up_actions and compute_pull_down_actions are identical. From what I can tell really the only thing that changes between the 2 functions is what function on the transaction is being called.

pyrefly/lib/lsp/non_wasm/server.rs
asukaminato0721: CodeActionKind comes from here: https://docs.rs/crate/lsp-types/0.94.0/source/src/code_action.rs#195

crates/pyrefly_config/src/config.rs
```diff
@@ -835,8 +835,8 @@ impl ConfigFile {
         self.handle_from_module_path_with_fallback(module_path, &FallbackSearchPath::Empty)
     }
 
-    pub fn handle_from_module_path_with_fallback<'a>(
-        &'a self,
+    pub fn handle_from_module_path_with_fallback(
```
jvansch1, line 838: Was this change intentional? It doesn't seem like we actually use this function anywhere else in the PR. If the lifetime is wrong here then it might be good to open another issue and PR to track it rather than doing it here.

pyrefly/lib/state/lsp/quick_fixes/move_members.rs
```diff
@@ -0,0 +1,458 @@
...
+
+const DEFAULT_INDENT: &str = "    ";
+
+/// Builds pull-members-up refactor actions for the supplied selection.
+pub(crate) fn pull_members_up_code_actions(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    selection: TextRange,
+) -> Option<Vec<LocalRefactorCodeAction>> {
+    let module_info = transaction.get_module_info(handle)?;
```
jvansch1: Similar comment here. There is a lot of shared logic between this function and push_members_down_code_actions. I may be good to try and extract what we can into something shared.

pyrefly/lib/state/lsp/quick_fixes/move_members.rs
```diff
@@ -0,0 +1,458 @@
+/*
```
jvansch1, line 1: In general there a lot added in this file. I think we at least need some docstrings here to give a description of what some of these functions are for. For example, functions like replaceable_pass_stmt are a little unclear about what they do and why we need them.

crates/pyrefly_config/src/config.rs
asukaminato0721: this is by `cargo clippy --fix --allow-dirty`.

pyrefly/lib/state/lsp/quick_fixes/move_members.rs
asukaminato0721: done

kinto0 (approved):

PR #2048, opened 2026-01-08
Disallow invalid string literals around `|` in type annotation

fangyi-zhou:
# Summary
This diff prevents invalid union types from being constructed using string literals (as a forward reference) and a type (e.g. `"str" | int`). Pyrefly currently reports no error for these runtime errors.

Note that union types can still be constructed if either side of the union operator is a type variable.

Implementation-wise, when we parse a string literal in a type annotation into an expression, we use the `node_index` field to convey that the parsed name originated from a string literal. This is somewhat hacky but is probably the easiest way to achieve this goal without making major changes to the existing code.

With this change, we are passing 2 more conformance test cases.

# Test Plan

`python test.py`

### DIFF
--- a/conformance/third_party/results.json
+++ b/conformance/third_party/results.json
@@ -3,7 +3,7 @@
   "pass": 109,
   "fail": 29,
   "pass_rate": 0.79,
-  "differences": 126,
+  "differences": 113,
   "passing": [
     "aliases_explicit.py",
     "aliases_newtype.py",
@@ -117,9 +117,9 @@
   ],
   "failing": {
     "aliases_implicit.py": 5,
-    "aliases_recursive.py": 11,
-    "aliases_type_statement.py": 4,
-    "aliases_typealiastype.py": 4,
+    "aliases_recursive.py": 2,
+    "aliases_type_statement.py": 1,
+    "aliases_typealiastype.py": 3,
     "aliases_variance.py": 4,
     "annotations_forward_refs.py": 2,
     "callables_annotation.py": 2,


fangyi-zhou:
Converting to draft because I didn't realise there are new features in Python3.14 (https://docs.python.org/3/whatsnew/3.14.html#pep-649-pep-749-deferred-evaluation-of-annotations) that affects annotation behaviour.

By sampling some of the errors from open source code, I realised that deferring annotation evaluation means that the runtime error might not arise unless the annotation is evaluated, e.g. `x: "int" | str` will not throw immediately in Python3.14

```
$ python3.14
Python 3.14.2 (main, Dec  5 2025, 16:49:16) [Clang 17.0.0 (clang-1700.4.4.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x: "int" | str
>>> __annotate__(1)
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    __annotate__(1)
    ~~~~~~~~~~~~^^^
  File "<python-input-0>", line 1, in __annotate__
    x: "int" | str
       ~~~~~~^~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'type'
```

However, I've also neglected handling of generic alias (?) which should work in Python3.13, which was revealed in the open source codebases.

```
$ python3.13
Python 3.13.9 (main, Oct 14 2025, 13:52:31) [Clang 17.0.0 (clang-1700.4.4.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import IO
>>> x: IO | "str"
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    x: IO | "str"
       ~~~^~~~~~~
TypeError: unsupported operand type(s) for |: 'type' and 'str'
>>> x: IO[bytes] | "str"
```

fangyi-zhou:
cc @rchen152 who authored https://github.com/facebook/pyrefly/commit/05a2ab9bba3b8c42c720ea46e023045b14dd659d

rchen152:
Thanks for taking this on! To minimize false positives, it looks like we need to avoid erroring in these cases:
* Target Python version is 3.14+
* `from __future__ import annotations` is present
* A specialized generic is one of the other  options in the union

fangyi-zhou:
I still intend to work on it, just taking slightly longer

fangyi-zhou:
Hi @rchen152 , I think this PR is ready for review now. I've restricted the error to one side being string, and the other side being a type to reduce the number of false positives.

rchen152:
Thanks! Sorry for the delay in review; I'll get to this as soon as I can.

pyrefly/lib/alt/operators.rs
```diff
@@ -253,6 +255,42 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             && let Some(l) = self.untype_opt(lhs.clone(), x.left.range(), errors)
             && let Some(r) = self.untype_opt(rhs.clone(), x.right.range(), errors)
         {
+            // Use `node_index` to detect if either side is a forward reference
```
rchen152: Rather than using `node_index` to detect forward references, this seems like it would be a perfect place to use `KeyExpect` and `BindingExpect` to communicate information between the bindings and answers phases. What that would look like is (1) add a new variant to both KeyExpect (https://github.com/facebook/pyrefly/blob/6bfc770934f66c71e41b83a7821ae2c8a426109e/pyrefly/lib/binding/binding.rs#L706) and BindingExpect (https://github.com/facebook/pyrefly/blob/6bfc770934f66c71e41b83a7821ae2c8a426109e/pyrefly/lib/binding/binding.rs#L798), (2) insert the new key/binding pair in `ensure_type_impl` when we see an `|` involving a string literal (https://github.com/facebook/pyrefly/blob/6bfc770934f66c71e41b83a7821ae2c8a426109e/pyrefly/lib/binding/expr.rs#L862), and (3) put the actual check logic into solve_expectation (https://github.com/facebook/pyrefly/blob/6bfc770934f66c71e41b83a7821ae2c8a426109e/pyrefly/lib/alt/solve.rs#L1740).

That would be cleaner than messing around with `node_index`, and I think you could also avoid copying around the python version and has_future_annotations around by doing that part of the check in `ensure_type_impl`.

pyrefly/lib/binding/bindings.rs
```diff
@@ -236,6 +239,8 @@ pub struct BindingsBuilder<'a> {
     semantic_syntax_errors: RefCell<Vec<SemanticSyntaxError>>,
     /// BoundName lookups deferred until after AST traversal
     deferred_bound_names: Vec<DeferredBoundName>,
+    /// Whether `from __future__ import annotations` is present
+    pub has_future_annotations: bool,
```
rchen152: `Scope` would be a more natural place to put this flag, since that's where we have other import tracking: https://github.com/facebook/pyrefly/blob/6bfc770934f66c71e41b83a7821ae2c8a426109e/pyrefly/lib/binding/scope.rs#L1019

pyrefly/lib/test/annotation.rs
```diff
@@ -21,12 +24,53 @@ assert_type(D.x, int)  # E: assert_type(Any, int) failed
 );
 
 testcase!(
-    bug = "conformance: Union syntax cannot be used with string operand; use quotes around entire expression: PR #2048 contains a draft implementation",
-    test_quoted_type_union_operator_runtime_error,
+    test_union_operator_with_bare_string_literal,
```
rchen152, line 27: Could we add an explicit python version to this test so its behavior doesn't depend on pyrefly's default target verion?

rchen152, line 27: A few more things I'd like to see tested:
* int | "list[str]" (a more complicated type in quotes)
* "int" | "list[str]" (both sides in quotes)
* SomeTypeAlias | "str" (I can't remember at what point we "untype" the type alias, so it may not be possible to detect the type alias here. If that's the case, please just add a test with a `bug` marker to document that we don't handle this well.)

rchen152 (changes requested):
Thanks for the PR, and sorry again for the slow review! Suggestions in-line - the most significant one is that I think you can simplify this by reusing pyrefly's `KeyExpect` / `BindingExpect` pattern.

pyrefly/lib/alt/operators.rs
fangyi-zhou: TIL! Thanks for the detailed explanation!

fangyi-zhou:
Rebased and (Claude) addressed comments! Thanks for the detailed review

stroxler (approved):

PR #2049, opened 2026-01-09
Fix sequence pattern matching type narrowing

### LINKED ISSUE #1806: [match-case]: sequence pattern not properly supported
### Describe the Bug

The code sample below ([pyright playground](https://pyright-play.net/?code=GYJw9gtgBAxmA28CmMAuBLMA7AzgOgEMAjGKdCABzBFSgGUkBHAVySxiQChRIpUBPCuiwBzMpWq0COHEhoB9LEgBucgDRRpshQIpdOAEyTA%2BSHKnmzG8igVSo5WABQAPAFz0mrdkgDaw1ABdAEooAFoAPigAOWwkN04oJKgIOxgACyh3ROTcmGkkKF8AKmUCeFYcQITc2uStOQtdJCcyirMNeHRzfywg4Jy6-NkoeRq62oaFJVUQVwHDY1NzSyYbO3kAd3RUdPlmLExndzI%2BqAAfTxY2Dl7%2B8KjYpXGk1NQMrJfa4cKA1vLWMEvhNNDJGvJmv92hoAgMQT8iqUAWZqoMQaDtE1BC02pVOt1UHcQmi8gVRsCJlMLDM5PMgA)) should pass without issues, but `pyrefly` reports a bunch of false positives

```python
from typing import assert_never, assert_type

def test_seq_pattern(x: Sequence[int]) -> None:
    match x:
        case [*values]:
            assert_type(values, list[int])
        case _:
            assert_never(x)

def test_seq_pat_with_union(x: int | Sequence[int]) -> None:
    match x:
        case int(value):
            assert_type(value, int)
        case [*values]:
            assert_type(values, list[int])
        case _:
            assert_never(x)
```

```
ERROR sandbox.py:9:26-27: Argument `Sequence[int]` is not assignable to parameter `arg` with type `Never` in function `typing.assert_never` [bad-argument-type]
ERROR sandbox.py:15:14-23: Type `int` is not iterable [not-iterable]
ERROR sandbox.py:15:15-22: Type `int` is not iterable [not-iterable]
ERROR sandbox.py:16:24-43: assert_type(list[int] | list[Any], list[int]) failed [assert-type]
ERROR sandbox.py:18:26-27: Argument `Sequence[int] | int` is not assignable to parameter `arg` with type `Never` in function `typing.assert_never` [bad-argument-type]
```

Not that for [sequence patterns](https://peps.python.org/pep-0634/#sequence-patterns), without additional subpatterns, we should have that:

- `[]` -> equivalent to `TypeIs[Sequence[Never]]`
- `[*args]` -> equivalent to `TypeIs[Sequence[Any]]`
- `[a, b, c]` -> equivalent to `TypeIs[Sequence[Any]]` and `len==3`
- `[a, b, c, *args]` -> equivalent to `TypeIs[Sequence[Any]]` and `len>=3`

There are more issues if we consider subpatterns, but probably worth it to make a separate issue for that.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEAxrlLAwC4S7pyGrYN0INYrkps6AZRgBHAK4x0DGAB10YavTaliEdAHNBw0eNRw4MMQH10MAG4WANHVPmrW4ivSrMMMHTbwbJbm0pbEqGwBlOgAFPiIkjLyijAA2rpsALoAlHQAtAB8dAByXDCIqnRVdDQRDAAWdPGV1a0MpjB0qQBUtqhQ8nCZFeitY9UuFkHuMDF9A-BOUBBwbOnoWdkt4%2B3mdJYj4%2BOTVjb2lHFbXli%2B-oHBMmERlgDuEGz1lrLonLHxghs6AAfRJyBRKdabfJFUo2Q7VWpsBpNeE7DoAthzfrybKoo5VE7TbSzebyJwZK743adHqk%2BDDbb45xmKaWGZYhZwJYrNYZHKMtrog4Co6E6x2CyXEAOECyDhQHhkCgAYjoAFV5e9SHQwN92L84KpvLcwKJEdZZDRsJL-hToXRVpQ8ZQYGxZNEdcoQMVLdanXRgPgAL5e1TSkBkF1gKCkQhsWhQFV0AAKpCjMYdGBwBEYXEgendEV%2BhFUqqknXqkWIcEQAHpa5HfDHCKI9LWFLXMLgGHBa0w1BAC5Qi1xazrRM4%2BtA%2BLBcwOhyPRrhiBwuDxVGQPlw8uc4L86ABeOhegDMhAAjAAmUPoEBBmWofX2ABi0BgFDQWDwRCVQaAA

_No response_

Adist319:
# Summary

Adds proper type narrowing for sequence patterns in match/case statements. Previously, sequence patterns like [*values] did not narrow union types correctly, and captured variables were not properly typed.

This PR:
- Adds Sequence type to the stdlib for type checking
- Introduces IsSequence/IsNotSequence narrowing operations
- Emits IsSequence narrowing for sequence patterns to properly filter union types
- Correctly types captured variables (e.g., [*values] gives values: list[T] when matching Sequence[T])
- Excludes str, bytes, and bytearray from sequence pattern matching per PEP 634

Fixes #1806.

# Test Plan

Added 9 new test cases covering:
    - Basic star capture patterns ([*values])
    - Union type narrowing (int | Sequence[int])
    - Fixed-length patterns ([a, b])
    - Mixed patterns ([first, *middle, last])
    - String exclusion per PEP 634
    - List and tuple patterns

### DIFF
--- a/crates/pyrefly_types/src/stdlib.rs
+++ b/crates/pyrefly_types/src/stdlib.rs
@@ -62,6 +62,7 @@ pub struct Stdlib {
     async_iterable: StdlibResult<(Class, Arc<TParams>)>,
     async_iterator: StdlibResult<(Class, Arc<TParams>)>,
     mutable_sequence: StdlibResult<(Class, Arc<TParams>)>,
+    sequence: StdlibResult<(Class, Arc<TParams>)>,
     generator: StdlibResult<(Class, Arc<TParams>)>,
     async_generator: StdlibResult<(Class, Arc<TParams>)>,
     awaitable: StdlibResult<(Class, Arc<TParams>)>,
@@ -202,6 +203,7 @@ impl Stdlib {
             async_iterable: lookup_generic(typing, "AsyncIterable", 1),
             async_iterator: lookup_generic(typing, "AsyncIterator", 1),
             mutable_sequence: lookup_generic(typing, "MutableSequence", 1),
+            sequence: lookup_generic(typing, "Sequence", 1),
             generator: lookup_generic(typing, "Generator", 3),
             async_generator: lookup_generic(typing, "AsyncGenerator", 2),
             awaitable: lookup_generic(typing, "Awaitable", 1),
@@ -458,6 +460,10 @@ impl Stdlib {
 
     pub fn mutable_sequence(&self, x: Type) -> ClassType {
         Self::apply(&self.mutable_sequence, vec![x])
+    }
+
+    pub fn sequence(&self, x: Type) -> ClassType {
+        Self::apply(&self.sequence, vec![x])
     }
 
     pub fn generator(&self, yield_ty: Type, send_ty: Type, return_ty: Type) -> ClassType {
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -88,6 +88,34 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         )
     }
 
+    /// Check if a type is a sequence type (subtype of Sequence[T] for some T).
+    /// This is used for sequence pattern matching in match/case statements.
+    /// Note: str, bytes, and bytearray are excluded as per PEP 634.
+    fn is_sequence_type(&self, ty: &Type) -> bool {
+        // Handle special cases first
+        match ty {
+            // Tuples are sequences
+            Type::Tuple(_) => return true,
+            // str, bytes, and bytearray are NOT matched by sequence patterns per PEP 634
+            Type::ClassType(cls)
+                if cls.is_builtin("str")
+                    || cls.is_builtin("bytes")
+                    || cls.is_builtin("bytearray") =>
+            {
+                return false;
+            }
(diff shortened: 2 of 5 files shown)


yangdanny97:
Overall approach looks good, I had a few pieces of feedback and noted some edge cases that need to be fixed, but it might be easier for Steven to just patch them before merging

I noticed that some of the test cases had assert_never imported but unused, was it because Pyrefly couldn't check it properly?

Adist319:
Yeah exactly, pyrefly's exhaustiveness checking only kicks in for enums and literal unions right now. so even though the IsNotSequence narrowing logic would correctly narrow Sequence[int] to Never after case [*values]:, the exhaustiveness checker bails out early for non-enum types (theres an early return in check_match_exhaustiveness in narrow.rs).

So if we actually used assert_never(x) in the wildcard case it would error saying the type isnt Never. Imported it mostly to show the intent that the branch should be unreachable, but had to use pass instead.

PR #2059, opened 2026-01-10
Fix match case dict pattern error after matching None (#2058)

### LINKED ISSUE #2058: negative narrowing not applied for `None` in match case
This is possibly not implemented yet or a bug. `match..case` can handle the second case without eliminating the None first. But pyrefly shows an error for dict matching.

```python
dict_or_none: dict[str, Any] | None = {"a": "b"}

match dict_or_none:
    case None:
        pass
    case {"a": "b"}: # Object of class `NoneType` has no attribute `__getitem__`
        pass
    case _:
        pass
```

_Originally posted by @satya-waylit in https://github.com/facebook/pyrefly/discussions/2050_

AryanBagade:
The fix creates a narrowed subject_idx for each case by applying the accumulated negated_prev_ops before passing it to bind_pattern - following the same pattern already used by MatchSequence and MatchClass.

Fixes #2058

Test Plan
  - Added regression test test_match_mapping_after_none
  - Verified the exact reproducer from the issue now returns 0 errors

### DIFF
--- a/pyrefly/lib/binding/pattern.rs
+++ b/pyrefly/lib/binding/pattern.rs
@@ -206,7 +206,6 @@ impl<'a> BindingsBuilder<'a> {
             Pattern::MatchMapping(x) => {
                 let mut narrow_ops = NarrowOps::new();
                 let mut subject_idx = subject_idx;
-                // Narrow the match subject to mapping types (filters out None, etc.)
                 if let Some(subject) = &match_subject {
                     let narrow_op = AtomicNarrowOp::IsMapping;
                     subject_idx = self.insert_binding(


satya-waylit:
Thanks for the quick PR. Would it also allow the following variant? I've used this construct in my code. No errors reported by pyright and mypy. And Python handles it fine too.

```python
match dict_or_none:
    case {"a": "b"}: # Object of class `NoneType` has no attribute `__getitem__`
        pass
    case None:
        pass
    case _:
        pass
```

AryanBagade:
> Thanks for the quick PR. Would it also allow the following variant? I've used this construct in my code. No errors reported by pyright and mypy. And Python handles it fine too.
> 
> ```python
> match dict_or_none:
>     case {"a": "b"}: # Object of class `NoneType` has no attribute `__getitem__`
>         pass
>     case None:
>         pass
>     case _:
>         pass
> ```

Good catch! Yeah this fix only handles the case where None comes first. The reverse order would need a different approach to fix, would look into it as well

AryanBagade:
as per my assumptions, MatchMapping itself needs to filter out non-mapping types (like None) BEFORE calling __getitem__ similar to how MatchSequence uses LenEq to filter out non-sequences.
That's a bigger change, needs a new narrow op or different approach in the solver. 
@stroxler your views??

yangdanny97:
Maybe we could add a IsMapping/IsNotMapping narrowing op like https://github.com/facebook/pyrefly/pull/2049 does for sequences

yangdanny97:
From the PEP:

For a mapping pattern to succeed the subject must be a mapping, where being a mapping is defined as its class being one of the following:
- a class that inherits from collections.abc.Mapping
- a Python class that has been registered as a collections.abc.Mapping
- a builtin class that has its Py_TPFLAGS_MAPPING bit set
- a class that inherits from any of the above (including classes defined before a parent’s Mapping registration)

The standard library classes dict and mappingproxy will have their Py_TPFLAGS_MAPPING bit set.

My thoughts:

I think checking if it extends `typing.Mapping` is pretty good step for now. I don't understand the "python class has been registered" part. dict and types.MappingProxyType already extend Mapping directly.

AryanBagade:
> Maybe we could add a IsMapping/IsNotMapping narrowing op like #2049 does for sequences

Done! Added IsMapping/IsNotMapping narrowing op that checks against typing.Mapping, Also added a test for the reverse case (dict pattern before None).

AryanBagade:
@yangdanny97 no hurry, just review once you get time!

pyrefly/lib/alt/narrow.rs
```diff
@@ -908,6 +908,20 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     ty.clone()
                 }
             }
+            AtomicNarrowOp::IsMapping => {
+                let mapping_type = self
+                    .stdlib
+                    .mapping(Type::any_implicit(), Type::any_implicit());
+                // Wrap in Type::Type to make it look like a class reference for isinstance
+                self.narrow_isinstance(ty, &Type::Type(Box::new(mapping_type.to_type())))
```
yangdanny97: You can add a helper to the stdlib that has just returns the Type::Class for mapping, similar to stdlib.set_object() v.s. stdlib.set(...)

https://github.com/facebook/pyrefly/blob/8052ae3f798f63ac5c8cd834622ab326027c73ba/crates/pyrefly_types/src/stdlib.rs#L475

pyrefly/lib/binding/pattern.rs
```diff
@@ -205,6 +205,24 @@ impl<'a> BindingsBuilder<'a> {
             }
             Pattern::MatchMapping(x) => {
                 let mut narrow_ops = NarrowOps::new();
+                let mut subject_idx = subject_idx;
+                // Narrow the match subject to mapping types (filters out None, etc.)
```
yangdanny97: i think this comment is redundant

yangdanny97:
The new mypy primer errors seem like we're behaving as expected, if we narrow `Any` to `Mapping` there's no way to know it's mutable.

https://github.com/common-workflow-language/cwltool/blob/main/cwltool/cwlprov/provenance_profile.py#L477

I wonder if we could special-case it so that `IsMapping` narrows to `MutableMapping` instead of `Mapping` for `Any`. I have no idea if that's better than what we have currently tho, it could cause more issues.

yangdanny97 (changes requested):
Thanks for the PR, one small change and it should be good.

It might be interesting to experiment with the MutableMapping stuff from the other comment, but out of scope for this PR.

pyrefly/lib/alt/narrow.rs
AryanBagade: got it!!

PR #2060, opened 2026-01-11
impl invert boolean

asukaminato0721:
# Summary

Fixes part of #364

Implemented an “Invert boolean” refactor code action that flips a local variable’s stored boolean value and updates all load sites (adds/removes not) across the file, with conservative guards against multi-target assignments and deletes. Wired it into the LSP code action pipeline as a refactor.rewrite.

# Test Plan

added focused LSP code action tests.

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
@@ -25,12 +25,14 @@ use crate::state::lsp::IdentifierContext;
 use crate::state::lsp::LocalRefactorCodeAction;
 use crate::state::lsp::Transaction;
 
+/// A text replacement for a single load of the target name.
 #[derive(Clone, Debug)]
 struct LoadEdit {
     range: TextRange,
     replacement: String,
 }
 
+/// Summary of a single assignment to the target name, used to build replacements.
 #[derive(Clone, Debug)]
 struct AssignmentPlan {
     target_range: TextRange,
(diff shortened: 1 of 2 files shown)


jvansch1:
@asukaminato0721 I am going to review this but I also wanted to flag that we just did a large refactor of code action code in Pyrefly. It may be worth it to rebase this on main and see if there is any logic that we added that you can re-use and make sure there is not anything conflicting.

asukaminato0721:
cc @jvansch1 I think this can be reviewed.

pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
```diff
@@ -0,0 +1,333 @@
...
+        has_del: false,
+    };
+    collector.visit_body(&ast.body);
+    StoreInfo {
+        stores: collector.stores,
+        has_del: collector.has_del,
+    }
+}
+
+fn collect_assignment_plans(
```
jvansch1: It seems like the logic in this function and in `collect_load_edits` are very similar and we are doing one AST traversal in each one. I wonder if we could instead do one AST traversal and collect all the information we need. What do you think?

pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
```diff
@@ -0,0 +1,333 @@
...
+use crate::state::lsp::Transaction;
+
+#[derive(Clone, Debug)]
+struct LoadEdit {
+    range: TextRange,
+    replacement: String,
+}
+
+#[derive(Clone, Debug)]
+struct AssignmentPlan {
```
jvansch1, line 37: Can we get a docstring on this? Just from looking at it I think it is a but unclear what this struct us supposed to represent

pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
```diff
@@ -0,0 +1,333 @@
...
+    let mut collector = AssignmentCollector {
+        target_name,
+        references,
+        plans: Vec::new(),
+    };
+    collector.visit_body(&ast.body);
+    collector.plans
+}
+
+fn assignment_value_kind(value: &Expr) -> AssignmentValueKind {
```
jvansch1: Can we instead call this something like `assignment_value_kind` from expression. I think that makes it more clear what this function is doing.

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -101,6 +101,28 @@ fn find_marked_range_with(source: &str, start_marker: &str, end_marker: &str) ->
     )
 }
 
+fn find_nth_range(source: &str, needle: &str, occurrence: usize) -> TextRange {
```
jvansch1, line 108: Can we get a docsring here explaining why we need this function and what it does?

pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
```diff
@@ -0,0 +1,333 @@
+/*
```
jvansch1, line 1: We should try to add at least some docstrings explaining the purpose of functions in this file.

pyrefly/lib/state/lsp/quick_fixes/invert_boolean.rs
```diff
@@ -0,0 +1,333 @@
...
+    }
+    Some(vec![LocalRefactorCodeAction {
+        title: format!("Invert boolean `{target_name}`"),
+        edits,
+        kind: CodeActionKind::REFACTOR_REWRITE,
+    }])
+}
+
+struct StoreInfo {
+    stores: Vec<TextRange>,
```
jvansch1: What exactly does "store" mean in this context. Is there a more descriptive name we can use?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -101,6 +101,32 @@ fn find_marked_range_with(source: &str, start_marker: &str, end_marker: &str) ->
     )
 }
 
+/// Finds the text range for the Nth occurrence of `needle` in `source`.
+///
+/// This is used by tests that need to select a specific repeated token without
+/// adding extra inline markers.
+fn find_nth_range(source: &str, needle: &str, occurrence: usize) -> TextRange {
```
jvansch1, line 108: If this is used only by tests can we put it behind a #[cfg(test)] directive?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -101,6 +101,28 @@ fn find_marked_range_with(source: &str, start_marker: &str, end_marker: &str) ->
     )
 }
 
+fn find_nth_range(source: &str, needle: &str, occurrence: usize) -> TextRange {
+    assert!(occurrence > 0, "occurrence is 1-based");
+    let mut start = 0;
+    let mut seen = 0;
+    while let Some(found) = source[start..].find(needle) {
+        let abs = start + found;
```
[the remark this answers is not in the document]
jvansch1: This seems concerning. We should make sure we handle the case where "needle" could be empty.

stroxler (approved):

PR #2070, opened 2026-01-12
Fix false negatives in uninitialized local detection (#2055)

### LINKED ISSUE #2055: False negatives in uninitialized local detection for branches that end in function calls (due to unknown Never/NoReturn status)
### Describe the Bug

Consider this example:
```
from typing import *


def foo() -> None: pass
def condition() -> bool: ...

if condition():
    x = 5
else:
    pass
    # if we comment this out, the uninitialized local error is detected!
    foo()

print(x)
```

Pyrefly fails to detect the uninitialized local, because the implementation of `Never` and `NoReturn` support to fix #361 made it clear that branch termination is actually a type-level concern, but uninitialized local checks happen at binding time. As a result, the best we can do is consider branches that end in a function call "potentially terminating" and treat them as such in initialization checks.

This means we get false negatives for any code where the last statement in a branch is a call to a function that *doesn't* have a Never / NoReturn type.

I'm not sure it's possible to fix the bug without reworking initialization from scratch. The possible approaches I see are:
- type-level (i.e. answers-solver-time) "coloring" of names with initialization, so that the checks move to the solver. Most likely we would add an extra initialization flag to `TypeInfo` if we did this, since initialization is about top-level names rather than arbitrary types.
- remove uninitialized local check entirely, and instead use it as a test bed for our first type-aware *lint* rule: this is actually appealing, because given our existing scope handling in bindings, initialization would be trivial to get right if we already had a typed AST - the reason it's hard now is because we're trying to do it *before* we have types

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRIzDDB0wuXAAoAlHQC0APjoA5XOhiI6xVHDjjJdAMbbMEBhG2KV6vLih7CbseghSLWa7fSKiCJ0IXT4dAC8dACsIjBQcLrBoQZGySEAxJxSAO4w5rQ0MOj8DAAWEHB0uACuDAA0jGX5NZ6eNqhQEABeMJh0ULhmnXQwlNSUnFUSDDBms5gAhOnSsooexJSsDHL4CiD1IHXQcCTkiCBZAKo2XUzSrfP%2BxqJYpjKUNKgMAProNTRsGNdnptko1HQ4AxKEF0KE6JQYAwapQ4WAhCANACgTC6MB8ABfDEiA4gMiIsBQUiEBi0KAULIABVIFKpkIwOAIBXQkDYKO%2B-kIIiyAGUYPkygwGMQ4IgAPRy8mSKmEXhsOXFOWYIZwOU%2BXn8mzaOWrSaoABuqGgqGwsG5BsoAu01WIRvQpxEZHK2mU5rGcH8kToGIAzIQAIwAJmJ6BABMOqCefoAYtAYBQ0Fg8EQyHGgA

_No response_

jackulau:
## Summary

Fixes #2055 - False negatives in uninitialized local detection when branches end with function calls.

**The Bug:** When a branch ended with a function call like `foo()`, pyrefly incorrectly assumed the branch might terminate with `NoReturn`, suppressing uninitialized variable warnings even for regular functions.

```python
def foo() -> None: pass

if condition():
    x = 5
else:
    foo()  # Bug: this suppressed the warning

print(x)  # x may be uninitialized - but no error was reported!
```

**The Fix:** Move initialization checking to solve time where we have type information to determine if a function actually returns `Never`.

## Approach

The root cause is that "branch termination is actually a type-level concern, but uninitialized local checks happen at binding time."

This PR defers the initialization check to solve time by:

1. **Tracking termination keys** - When merging flows, track the last expression statement for branches that don't initialize a variable
2. **New `Binding::InitCheck`** - Instead of emitting errors at binding time, create an `InitCheck` binding that defers the check
3. **Solve-time verification** - At solve time, check if ALL termination keys resolve to `Never` type. Only suppress the error if all non-initializing branches truly return `NoReturn`

## Test Cases Added

- `test_if_defines_variable_in_one_side_with_call`
- `test_if_defines_variable_with_noreturn_branch`
- `test_nested_if_with_inner_regular_call`
- `test_multiple_branches_all_noreturn`

## Test Plan
- All 90 flow_branching tests pass
- All existing NoReturn-related tests pass

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -2738,11 +2738,34 @@ impl<'a> BindingsBuilder<'a> {
         // - It was defined before the loop (base_has_value), OR
         // - It's defined in all loop body branches (since the loop definitely runs at least once)
         //
+        // For regular Loop where the variable is defined in all loop body branches but not
+        // in the base: we need to decide whether to flag "may be uninitialized".
+        // - If the loop body has a termination key (function call that might be NoReturn),
+        //   defer to solve time to check if it's actually Never. This handles patterns like
+        //   else-branch + while where the while condition is implied by the else condition.
+        // - Otherwise, flag it as uninitialized (the loop might not run).
+        //
         // For other merges, a name is always defined if it's defined in all branches.
         // If some branches are missing but have termination keys, we defer to solve time
         // to check if those termination keys are Never.
+        let n_loop_body_branches = if matches!(merge_style, MergeStyle::Loop) {
+            n_branches - 1 // n_branches includes +1 for base
+        } else {
+            n_branches
+        };
+        let defined_in_all_loop_body_branches = n_values == n_loop_body_branches;
+        // Check if any loop body branch has a termination key
+        let loop_has_any_term_key = all_termination_keys.iter().any(|k| k.is_some());
         let this_name_always_defined = match merge_style {
             MergeStyle::LoopDefinitelyRuns => base_has_value || n_values == n_branches,
+            MergeStyle::Loop if defined_in_all_loop_body_branches && loop_has_any_term_key => {
+                // Variable is in all loop body branches, and at least one branch has a
+                // termination key. Be lenient: treat as "always defined" for now, and
+                // rely on solve-time check of the termination key to catch real issues.
+                // This handles patterns like else-branch + while where the condition
+                // implies the loop runs.
+                true
+            }
             _ => n_values == n_branches,
         };
         match value_idxs.len() {


stroxler:
Thanks for the PR! I'm not entirely sure when I'll get to this - I likely have a competing fix as part of some active work on match exhaustiveness, which is why I had assigned the issue to myself.

I'll try to take a look later this week, but I may wind up closing this PR because my existing work (and the fact that it's intertwined with match logic) might take precedence.

From a skim, the fix looks pretty good to me

stroxler:
(Just as a general note - if an issue is assigned I'd recommend pinging the assignee before you take it on - this often means someone is already working on a fix, or they believe the issue needs to be part of a bigger chunk of changes rather than tackled as a one-off)

jackulau:
Thanks for the heads up, I'll keep that in mind for future contributions!

yangdanny97:
Closing this because the original issue has been fixed

PR #2074, opened 2026-01-12
impl Add support for Django reverse relationships #2071

### LINKED ISSUE #2071: Add support for Django reverse relationships
### Describe the Bug

I'm aware that the [docs](https://pyrefly.org/en/docs/django/#reverse-relationships) specify that this is not supported, but it would be _awesome_ to have!

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2071

Implemented Django reverse relationship synthesis by scanning model field bindings in the module, inferring reverse accessors for ForeignKey/ManyToManyField (and OneToOneField), parsing related_name (including %(class)s/%(app_label)s), and adding a RelatedManager helper for reverse FK accessors.

  Notes/limits:

Reverse accessors are synthesized only from relation fields declared in the same module and with literal related_name/None (dynamic names are skipped).

~~Cross-module reverse inference isn’t covered yet.~~ supported.

# Test Plan

Added assertions for reverse accessors


migeed-z:
Thank you so much! I will import and take a closer look.

rchen152 (changes requested):

migeed-z:
Hello! we got to review the diff in more details. Here are some open questions:
- The implementation requires iterating over every single `ClassField` binding. Can we do better? this would be expensive for large django models. Can we store do one pass in the bindings phase and collect `ForeignKey`, `ManyToManyField`, and `OneToOneField` assignments and store the names of those fields in class metadata and go from there?
- Cross module support seems important to me on realistic codebases and ultimately the goal. Is this approach going in the right direction towards that?

asukaminato0721:
Moved Django reverse-relationship synthesis to a module-level binding/index so we compute reverse fields once per module (instead of scanning every KeyClassField for each model).

The binder now records candidate relation fields during the class-body pass, and the solver builds a reverse-relation index keyed by target class that get_django_model_synthesized_fields can query.

asukaminato0721:
Started cross‑module support by building a global reverse‑relation index once per solve and caching it in the thread state. The index merges each module’s KeyDjangoRelations result, so reverse accessors now come from all modules loaded in the current transaction (not just the current module).

This required exporting KeyDjangoRelations and adding LookupAnswer::modules() so the solver can enumerate available modules and combine their per‑module indexes.

pyrefly/lib/test/django/many_to_many.rs
```diff
@@ -30,6 +30,9 @@ assert_type(book.authors.all(), QuerySet[Author, Author])
 assert_type(book.authors.filter(name="Bob"), QuerySet[Author, Author])
 assert_type(book.authors.create(name="Alice"), Author) 
 
+author = Author()
```
migeed-z: could we please revert all the test files back to the original except the reverse_relations.rs one? I added that so all reverse relation tests could live there and added the various corner cases and the macro that tests cross module support.

migeed-z (changes requested):

migeed-z:
Thank you so much for the updates. We are currently making some changes to incremental checks and since this change will affect incremental, we need to hold off on merging till the changes are made. But I reviewed the diff in detail and the high level approach looks pretty good, so once those changes are made, we should be good to go :)

yangdanny97:
cc @kinto0 this involves exports changes

at a high level its how i would expect it to be solved, but will likely need modifications to work properly w/ incremental checking

we need to make sure that the reverse dependency edge actually gets added, and make sure that creating a bunch of small cycles doesn't kill perf

@asukaminato0721 we're in the middle of a fairly large change to the exports logic, so we'll give this a closer look after we finish that

kinto0:
> cc @kinto0 this involves exports changes
> 
> at a high level its how i would expect it to be solved, but will likely need modifications to work properly w/ incremental checking
> 
> we need to make sure that the reverse dependency edge actually gets added, and make sure that creating a bunch of small cycles doesn't kill perf
> 
> @asukaminato0721 we're in the middle of a fairly large change to the exports logic, so we'll give this a closer look after we finish that

thanks for the tag. once my code lands, we will need to make sure that `KeyDjangoRelations` updates state correctly. 

if you want to get a headstart (though no guarantee this will stay the same so it's probably worth waiting), you will need to do something similar to #2202 for this new key. and you'll need tests similar to #2203

rchen152:
@asukaminato0721 Thanks for updating the PR!

@kinto0 @yangdanny97 Is incremental checking in a state where we can move forward with this? Do we want to wait until we have more perf testing?

yangdanny97:
Ideally we would have better CI benchmarking to be in place first, but if whoever imports this also runs the benchmarks manually it should be fine?

In addition to the usual suite we'd also want to check before/after for some larger project that uses django.

rchen152:
Thanks, Danny! I likely won't have time this coming week, since I'm taking some time off starting Thursday, but I can import this and run the benchmarks when I get back if no one gets to it in the meantime.

SHxKM:
This is great. Are there any plans to merge this?

rchen152:
> This is great. Are there any plans to merge this?

Sorry for the lack of movement on this one. Yes, I do still intend to review and merge this; it's just a very large PR and we're rather backed up on reviews.

asukaminato0721:
interesting...

```
 INFO Checking project configured at `/home/runner/work/pyrefly/pyrefly/primer_base/projects/zulip/pyrefly.toml`
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/state/state.rs:3006:14:
called `Option::unwrap()` on a `None` value
```

rchen152:
Thanks for your patience here. I did some perf testing, and unfortunately cross-module django reverse relationships support is a noticeable perf hit - about 5x (600ms to 3s) on the mypy_primer zulip project. I'm going to start by merging the pieces of this PR that don't involve cross-module support and then see what we can do to optimize it. The good news is that the slowdown only appears to occur on code that uses django, so a bit of performance is probably an acceptable tradeoff for this feature, but 5x is definitely too much.

rchen152:
I've merged the parts of this PR that add same-module support for Django reverse relationships (with the appropriate author attribution ;) ). Based on some quick sampling of popular GitHub projects, I believe this will allow pyrefly to resolve ~40% of reverse relationships, which is already a great improvement - thanks @asukaminato0721!

For the cross-module portion, unfortunately I don't think the approach in this PR is workable. It looks at all currently loaded modules, which (1) is very slow, and (2) isn't viable in the IDE, because we need to load modules lazily, and modules can also appear or disappear as the code is modified. My feeling is that we'll need to ask the user to somehow provide us some info about where models are located, for us to discover reverse relationships reliably and performantly. I'm exploring some options and will update the issue with some ideas.

PR #2075, opened 2026-01-12
impl LSP: Navigation support for pytest fixtures #1901

### LINKED ISSUE #1901: LSP: Navigation support for pytest fixtures
### Describe the Bug

Navigation features related to pytest [fixtures](https://docs.pytest.org/en/stable/explanation/fixtures.html) like the "Go to Definition" and "Go to References" are not supported. I would expect to be able to jump to the definition from the fixture in the test parameters, as well as getting all usages listed when doing "Go to references" on the fixture implementation.

 These are special cases since the fixtures are not called directly but resolved via name-based dependency injection by pytest during collection. Currently they are treated as common functions and parameters.

Tested on:
**Vscode**: 1.106.3
**Pyrefly Extension**: 0.46.0

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1901

Implemented pytest import/alias detection plus fixture discovery and parameter scanning in pyrefly/lib/state/lsp.rs, and hooked it into go‑to‑definition and local references.

todo: should these pytest functions in their own mod?

# Test Plan

cargo test pytest_fixture_parameter_goes_to_fixture_definition

cargo test pytest_fixture_references_include_injected_parameters


kinto0:
> todo: should these pytest functions in their own mod?

for sure. maybe a pyrefly/lib/third_party or something similar?

pyrefly/lib/state/lsp.rs
```diff
@@ -1478,6 +1484,61 @@ impl<'a> Transaction<'a> {
             .collect()
     }
 
+    fn pytest_fixture_definitions_for_parameter(
+        &self,
+        handle: &Handle,
```
migeed-z, line 2116: I think code involving pytest fixtures should be in its own modules. Can we edit the PR to isolate this code into its own module as much as possible?

migeed-z (changes requested):

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,207 @@
...
+                        }
+                    }
+                    for param in function_def.parameters.kwonlyargs.iter() {
+                        if param.name() != "self"
+                            && param.name() != "cls"
+                            && param.name().id() == fixture_name
+                        {
+                            references.push(param.name().range());
+                        }
+                    }
```
migeed-z: this part of the code seems repeated. Could we add a helper?

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,207 @@
...
+
+// Tracks local names that refer to pytest so we can resolve @pytest.fixture and fixture aliases.
+#[derive(Default)]
+pub(crate) struct PytestAliases {
+    pytest_module_aliases: Vec<String>,
+    fixture_aliases: Vec<String>,
+}
+
+impl PytestAliases {
+    pub(crate) fn from_module(module: &ModModule) -> Self {
```
migeed-z: This function iterates over every statement in the module per invocation.  Is there a better solution?

pyrefly/lib/test/lsp/definition.rs
```diff
@@ -156,6 +156,34 @@ Definition Result:
     );
 }
```
migeed-z, line 161: A more high level comment/question is, do we want to add type support? or is this exclusively an IDE feature? For example, it would be great if the following worked:

```
import pytest

@pytest.fixture
def my_fixture() -> int:
    return 42

def test_foo(my_fixture):
    reveal_type(my_fixture) # int
```


That would require extending the binding layer. 

Additionally, how would conftest.py fit into this support?  Overall, I think there are some open questions when it comes to this feature. 

On a different note, I would also say that attr support could be a nice problem to look at (if you are specifically interested in 3p support!)

migeed-z (changes requested):

pyrefly/lib/test/lsp/definition.rs
asukaminato0721: > do we want to add type support

https://github.com/facebook/pyrefly/pull/2201 I think it's here?

migeed-z: I was thinking more like what we're doing for pydantic and django in the binding layer. Was curious why we didn't follow that direction.

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+}
+
+/// Returns fixture definitions for a parameter when the cursor is in a pytest test/fixture.
+pub(crate) fn pytest_fixture_definitions_for_parameter(
+    module: &ModModule,
+    identifier: &Identifier,
+    covering_nodes: &[AnyNodeRef],
+) -> Option<Vec<PytestFixtureDefinition>> {
+    let module_info = PytestModuleInfo::from_module(module)?;
```
migeed-z: my understanding is that this will be recomputed on every call? if so, I would love to learn more on why we are not taking a binding approach to store some of the information we need?

migeed-z (changes requested):

pyrefly/lib/test/lsp/definition.rs
asukaminato0721: Pytest fixtures here are only used to improve navigation (definition/refs) in the LSP, not typing.

The LSP path already has AST access and is allowed to do local scans. Avoiding extra binding state keeps the feature isolated

For now, it’s a tradeoff compared to adding new binding data structures and incremental update logic.

asukaminato0721: And add typing will be too big for a single pr?

migeed-z: thanks for the improvements! the diff is looking much better. Let me discuss with the team those tradeoffs.

migeed-z: I discussed with the team. I think this tradeoff between storing data in the binding layer vs the lsp layer is still open. We need to better understand the perf implications of this.

asukaminato0721: now it bind to types

asukaminato0721:
https://stackoverflow.com/questions/77318902/why-using-yield-instead-of-return-in-pytest-fixtures aa, not only return but yield

asukaminato0721:
currently I fixed these cases in pytest.fixture

return
yield
async/await

asukaminato0721:
about scrapy (https://github.com/scrapy/scrapy) , mainly because the usage of Mock

kinto0:
I wonder if it's worth looking into some of these errors and seeing if we find issues. the scale of these issues concerns me since I don't think it's a great UX to get errors in tests when there weren't any before

asukaminato0721:
let llm to do some analysis on prefect (https://github.com/PrefectHQ/prefect)

```
  2) prefect-sqlalchemy async/sync union + narrowings that pyrefly can’t prove
  File: src/integrations/prefect-sqlalchemy/tests/test_database.py

  The fixture managed_connector_with_data returns either a SqlAlchemyConnector (sync) or an AsyncSqlAlchemyConnector (async), and you branch on connector_with_data._is_async. Pyrefly
  does not use _is_async as a type guard, so it treats the object as a union of both types everywhere.

  This causes:

  - “Type tuple[Any] is not awaitable” / “list[tuple[Any]] is not awaitable”
    In async branches like:

    results = await managed_connector_with_data.fetch_one(...)
    pyrefly sees that fetch_one might be the sync version, which returns a tuple, not awaitable. Same for fetch_many, fetch_all, etc.
    Lines around 291, 344, 409, etc.
    Lines around 291, 344, 409, etc.
  - “Type None is not awaitable”
    Calls like await managed_connector_with_data.close() or await managed_connector_with_data.reset_connections() are legal for the async connector, but the sync connector versions
    return None, so the union is Awaitable[None] | None.
    Lines around 236, 309, 425, etc.
  - “Cannot use Connection/AsyncConnection as a context manager”
    In prefect_sqlalchemy/database.py, get_connection() is annotated to return Connection, but when begin=True it actually returns engine.begin(...), which is a context manager (or
    async context manager).
    In tests, you do:

    async with connection as conn: ...
    with connection as conn: ...

    Pyrefly believes connection is a plain Connection / AsyncConnection, which it doesn’t treat as a context manager, so it flags it.
    Lines ~255, 258, 278, 281.
  - _exit_stack.close() + NoneType
    _exit_stack is typed as Optional[ExitStack] for the sync connector and Optional[AsyncExitStack] for async. In the cleanup branch you do:

    connector_with_data._exit_stack.close()

    Pyrefly sees Optional[AsyncExitStack] or None, and AsyncExitStack doesn’t have .close() (it has .aclose()).
    Line ~239.
  - len(managed_connector_with_data._unique_results)
    _unique_results is annotated as Optional[Dict[...]], so len(...) on it is flagged unless you check it’s not None first (even though it’s initialized at runtime).
    Lines around 306, 310, 330, etc.
```

this seems to be mix sync with async and "None" problem

asukaminato0721:
cloud-init

```
  1. Dynamic namedtuple fixtures are inferred as plain tuple
     In tests/unittests/analyze/test_show.py, mock_io returns a runtime namedtuple, so pyrefly infers tuple[Any, ...]. That’s why .infile is flagged and why passing it to analyze_show
  (expects argparse.Namespace) is flagged.
     Same pattern in tests/unittests/config/test_cc_install_hotplug.py: the mocks fixture yields a namedtuple, so the checker sees a tuple and doesn’t know
  about .m_which, .m_update_enabled, etc.
  2. Optional returns used without a None check
     cloudinit.sources.pkl_load is typed as Optional[DataSource]. In tests/unittests/test_upgrade.py the fixture returns it directly and the tests access attributes without checking
  for None, so you get NoneType attribute errors (dsname, distro, paths, etc.).
     paths.get_ipath(...) is typed as Optional[str], but util.load_text_file expects a str; this is why you see the tests/unittests/test_data.py errors.
  3. Mocks created dynamically but still typed as the original function
     Pyrefly doesn’t know that mocker.patch.object or mock.patch has turned a method into a Mock, so attributes like .call_count or .return_value are flagged as missing on
  FunctionType. This is what’s happening in tests/unittests/sources/test_oracle.py and similar spots.
  4. Attributes inferred as str because of sentinel UNSET
     In cloudinit/sources/DataSourceAzure.py, _metadata_imds is initialized with sources.UNSET (a str), so the checker infers the attribute as str. Tests then assign dicts to it in
  tests/unittests/sources/test_azure.py, yielding bad-assignment.
     Same pattern in OpenStack and other data source tests: attributes are typed as str or str | None, but tests assign dicts/lists (tests/unittests/sources/test_openstack.py, tests/
  unittests/sources/test_oracle.py, tests/unittests/sources/test_smartos.py, etc.).
  5. Base-class signature vs subclass extra kwargs
     cloudinit.distros.Distro.apply_locale doesn’t accept keyname, but Debian’s override does. The test uses keyname on a value typed as base Distro, so pyrefly rejects the keyword
  (tests/unittests/distros/test_debian.py).
  6. TypedDict / strict type expectations
     cloud_init_modules is typed as list[str], but tests/unittests/runs/test_simple_run.py sets it to None to represent “empty”. That’s a type error by definition.
     Other TypedDict-ish errors (e.g., test_upcloud) come from treating str | None as sized or dict-like without narrowing.
```

asukaminato0721:
setuptools (https://github.com/pypa/setuptools)

```
  1) Dynamic command options on MyCmd
  File: setuptools/_distutils/tests/test_cmd.py
  MyCmd subclasses distutils.cmd.Command but its initialize_options is empty:

  class MyCmd(Command):
      def initialize_options(self):
          pass

  The tests then set arbitrary attributes (option1, option2, not_string_list, etc.) on the instance. Pyrefly doesn’t infer “this class allows dynamic attributes,” so it reports
  missing attributes. This is the source of all the MyCmd has no attribute … errors.

  2) Intentionally “wrong” argument types for tests
  Same file, test_make_file:

  cmd.make_file(infiles=True, outfile='', func='func', args=())

  The test is intentionally passing invalid types to confirm runtime errors. Pyrefly sees func='func' but make_file expects a Callable, so it reports bad-argument-type. This is a
  false positive in the static checker because the test is purposely invalid.

  3) Assigning test doubles to methods with strict signatures
  cmd.execute = _execute and cmd.announce = _announce are test stubs.
  Pyrefly treats execute and announce as methods with a precise signature. Assigning a function with a looser signature is flagged as bad-assignment. It’s another case of dynamic
  testing patterns vs. static typing.

  4) Environment.paths on a str subclass
  File: setuptools/tests/test_egg_info.py
  Environment is just class Environment(str): pass, and the fixture sets env.paths = {...} dynamically. Pyrefly doesn’t permit adding new attributes to a str subclass without an
  annotation, so every access to env.paths is flagged.
```

so many dynamic.

asukaminato0721:
scrapy (https://github.com/scrapy/scrapy)

```
  1) tests/test_engine.py missing-attribute on _slot
  ExecutionEngine._slot is annotated as optional in scrapy/core/engine.py:

  - self._slot: _Slot | None = None

  In the tests, you set engine._slot = Mock() and then access engine._slot.add_request / remove_request. A static type checker treats _slot as Optional[_Slot] in the class definition
  and does not narrow the type just because you assign a Mock in a test. So any access to engine._slot.* without a guard is treated as “might be None”, hence:

  - “Object of class NoneType has no attribute add_request / remove_request”

  That’s a static typing rule issue, not a runtime issue. It’s likely “new” because the checker got stricter about attribute optionality or started analyzing these tests more deeply.

  2) tests/test_utils_misc/test_return_with_argument_inside_generator.py bad-argument-type
  warn_on_generator_with_return_value is typed to accept a real Spider:

  def warn_on_generator_with_return_value(spider: Spider, callable: Callable[..., Any]) -> None:

  Your tests pass MockSpider, which is not a subclass of Spider (it only has .settings). A nominal type checker won’t accept that unless Spider is a Protocol (it isn’t). So you get:

  - “Argument MockSpider is not assignable to parameter spider with type Spider”

```

asukaminato0721:
Conclusion: this maybe default close since too much noise for now.

asukaminato0721:
https://github.com/astral-sh/ty/issues/2864 this may also good feature

jaens:
_In case_ disabling this also disables the navigation (don't really understand if it does :sweat_smile: ):

Perhaps the "go to definition" functionality could be separated from fixture argument type inference (like Pylance does it) so there would be no need to deal with the errors (make the type inference optional)?


In Pylance (as far as I have experienced), the fixture arguments internally have type `Any`, but parameter inlay hints and go-to-definition behave as if the location/type of the fixture was inferred.

asukaminato0721:
> _In case_ disabling this also disables the navigation (don't really understand if it does 😅 ):
> 
> Perhaps the "go to definition" functionality could be separated from fixture argument type inference (like Pylance does it) so there would be no need to deal with the errors (make the type inference optional)?
> 
> In Pylance (as far as I have experienced), the fixture arguments internally have type `Any`, but parameter inlay hints and go-to-definition behave as if the location/type of the fixture was inferred.

At first it is separated, https://github.com/facebook/pyrefly/pull/2075/changes/5d0c4307c5659223f8e9d658d4253fd6dad8a6ef

But then since we need to track the definition position to provide navigation, why not infer the type as well, so I implemented the type binding as well.

asukaminato0721:
> _In case_ disabling this also disables the navigation (don't really understand if it does 😅 ):
> 
> Perhaps the "go to definition" functionality could be separated from fixture argument type inference (like Pylance does it) so there would be no need to deal with the errors (make the type inference optional)?
> 
> In Pylance (as far as I have experienced), the fixture arguments internally have type `Any`, but parameter inlay hints and go-to-definition behave as if the location/type of the fixture was inferred.

Your idea make sense, I change to default Any.

kinto0:
seems like a frequently requested feature: https://www.reddit.com/r/Python/comments/1tbxzza/comment/olmfhg5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

crates/pyrefly_config/src/base.rs
```diff
@@ -285,6 +285,11 @@ pub struct ConfigBase {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub spec_compliant_overloads: Option<bool>,
 
+    /// Whether to infer injected pytest fixture parameter types from fixture definitions.
+    /// When false (the default), injected fixture parameters fall back to `Any` unless explicitly annotated.
```
kinto0: this piece is where I think bugs might arise

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,273 @@
...
+                    out,
+                    nested_class_key,
+                );
+            }
+            _ => {}
+        }
+    }
+}
+
+fn has_fixture_definition_at_range(
```
kinto0: this function seems redundant with collect_pytest_fixture_definitions_for_name. can we instead map `collect_pytest_fixture_definitions_for_name` into an `any` checking range?

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,273 @@
...
+
+fn is_pytest_fixture_function(
+    function_def: &StmtFunctionDef,
+    class_key: Option<Idx<KeyClass>>,
+    pytest_info: &PytestBindingInfo,
+) -> bool {
+    pytest_info.is_fixture_definition(&function_def.name, class_key.as_ref())
+}
+
+fn maybe_add_fixture_param_reference(
```
kinto0: only used in place. maybe we should inline?

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,273 @@
...
+        .posonlyargs
+        .iter()
+        .chain(parameters.args.iter())
+        .chain(parameters.kwonlyargs.iter())
+    {
+        maybe_add_fixture_param_reference(param, fixture_name, references);
+    }
+}
+
+fn collect_pytest_fixture_parameter_ranges(
```
kinto0: can the logic in this be deduped with collect_pytest_fixture_definitions_for_name so we can map across it?

crates/pyrefly_config/src/base.rs
```diff
@@ -285,6 +285,11 @@ pub struct ConfigBase {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub infer_pytest_fixture_types: Option<bool>,
```
kinto0: we definitely need this documented in the configuration docs. it's unclear whether navigation will work with/without this setting (it looks like it does but that's not easy to understand from a glance).

pyrefly/lib/binding/bindings.rs
```diff
@@ -570,6 +589,64 @@ impl Bindings {
...
+        let info = self.0.pytest_info.as_ref()?;
+        if matches!(param.id.as_str(), "self" | "cls") {
+            return None;
+        }
+        if !self.is_pytest_test_or_fixture(def, class_key, info) {
+            return None;
+        }
+        let defs = info.fixture_definitions(&param.id)?;
+        if !self.0.infer_pytest_fixture_types {
+            return Some(PytestFixtureParamHint::Any);
```
kinto0: what's the difference here between returning `None` and returning `PytestFixtureParamHint::Any`?

it looks like `PytestFixtureParamHint::Any` will result in a "cleansing" of the type by turning it into `Type::any_explicit()`. but is that desirable over simply not inferring it (returning `None` here)? why/why not?

with untyped code, we would previously get missing annotation errors. does this behavior still exist after the "cleansing"?

could we add tests for this (pytest fixture with no annotation, does it error?)? and what was the behavior before this?

pyrefly/lib/binding/bindings.rs
```diff
@@ -2357,3 +2438,23 @@ impl<'a> SemanticSyntaxContext for BindingsBuilder<'a> {
         self.scopes.is_bound_parameter(name)
     }
 }
+
+impl BindingsBuilder<'_> {
+    pub(crate) fn record_pytest_fixture_definition(
```
kinto0: let's put a docstring here. why do we record, what does it do, etc

kinto0: nit: this doesn't always record. maybe a better name is `maybe_record_pytest_fixture_definition`?

pyrefly/lib/state/lsp/pytest.rs
```diff
@@ -0,0 +1,66 @@
...
+use ruff_text_size::TextRange;
+
+use super::DefinitionMetadata;
+use super::FindDefinitionItemWithDocstring;
+use crate::state::pytest::pytest_fixture_definitions_for_parameter as pytest_fixture_definitions_for_parameter_in_module;
+use crate::state::pytest::pytest_fixture_parameter_references as pytest_fixture_parameter_references_in_module;
+use crate::state::state::Transaction;
+
+impl<'a> Transaction<'a> {
+    pub(super) fn pytest_fixture_definitions_for_parameter(
```
kinto0, line 27: let's put a docstring on this. let's include performance characteristics: why is it ok to call this on every definition? because it is very cheap when pytest bindings don't exist?

pyrefly/lib/state/lsp/pytest.rs
```diff
@@ -0,0 +1,66 @@
...
+use pyrefly_build::handle::Handle;
+use pyrefly_python::symbol_kind::SymbolKind;
+use ruff_python_ast::AnyNodeRef;
+use ruff_python_ast::Identifier;
+use ruff_python_ast::name::Name;
+use ruff_text_size::TextRange;
+
+use super::DefinitionMetadata;
+use super::FindDefinitionItemWithDocstring;
+use crate::state::pytest::pytest_fixture_definitions_for_parameter as pytest_fixture_definitions_for_parameter_in_module;
```
kinto0: why the rename? can we have them called two separate things?

pyrefly/lib/test/pytest.rs
```diff
@@ -0,0 +1,89 @@
...
+def test_data(sample_data):
+    reveal_type(sample_data)  # E: revealed type: Literal[42]
+"#,
+);
+
+testcase!(
+    test_pytest_fixture_async_parameter_type,
+    env_pytest_fixture(),
+    r#"
+import pytest
```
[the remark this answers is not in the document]
kinto0: let's add this

kinto0 (changes requested):
I like the approach a lot and I can tell a lot of time went into this. Thank you, and sorry about the delay in reviewing. Let's get this over the line. 

For next time, a summary of what / how you did a huge change like this would make it easier to review:

"we inject these in bindings", "we use the types in ...", "we add support in ...", "the configuration controls <>", etc

mostly small nits. 

question: do you see people ever setting `infer_pytest_fixture_types` to True? what's the use case? should we even have it?

website/docs/configuration.mdx
```diff
@@ -752,6 +752,22 @@ When set to `true`, parameter list compatibility is checked strictly even when
...
 - Flag equivalent: `--strict-callable-subtyping`
 
+### `infer-pytest-fixture-types`
+
+Controls whether unannotated pytest fixture parameters get their type from the
+fixture definition with the same name.
+
+When set to `false` (the default), Pyrefly recognizes pytest-injected fixture
+parameters and treats them as `Any`. This avoids noisy missing-annotation errors
+in test functions while keeping fixture navigation support enabled.
```
kinto0: cleaned up a bit
```suggestion
When set to `false` (the default), Pyrefly treats pytest-injected fixture
parameters as `Any`. This avoids noisy missing-annotation errors
in test functions while keeping fixture navigation support enabled.
```

pyrefly/lib/state/pytest.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+/// Returns fixture definitions for a parameter when the cursor is in a pytest test/fixture.
+pub(crate) fn find_pytest_fixture_definitions_for_parameter(
+    module: &ModModule,
+    bindings: &Bindings,
+    identifier: &Identifier,
+    covering_nodes: &[AnyNodeRef],
+) -> Option<Vec<PytestFixtureDefinition>> {
```
kinto0: why not just return a Vec1 here instead of asserting it is one after?

pyrefly/lib/state/lsp.rs
```diff
@@ -2278,18 +2279,28 @@ impl<'a> Transaction<'a> {
...
-                Ok(vec1![FindDefinitionItemWithDocstring {
-                    metadata: item.metadata,
-                    definition_range: item.definition_range,
-                    module: item.module,
-                    docstring_range: None,
-                    display_name: Some(identifier.id.to_string()),
-                }])
+                    &covering_nodes,
+                ) {
+                    Ok(Vec1::try_from_vec(pytest_definitions).expect(
```
kinto0: let's avoid this by returning a Vec1

kinto0:
a few small nits. I can clean them up on import

pyrefly/lib/state/pytest.rs
kinto0: even better, it can be a non-optional vec where empty means there were none

yangdanny97 (changes requested):

pyrefly/lib/test/lsp/definition.rs
```diff
@@ -159,6 +159,34 @@ Definition Result:
...
 
+#[test]
```
kinto0, line 162: from @yangdanny97 :
do we want tests to check what happens if the parameter name doesn't exist as a function somewhere, or if the function exists but isn't a fixture, or if the function we're trying GTD isn't a top-level test function?

website/docs/configuration.mdx
```diff
@@ -752,6 +752,22 @@ When set to `true`, parameter list compatibility is checked strictly even when
 - Default: `false`
 - Flag equivalent: `--strict-callable-subtyping`
 
+### `infer-pytest-fixture-types`
```
kinto0: from @yangdanny97:

we should split this into it's own PR with it's own rationale. it's not related to navigation support

pyrefly/lib/binding/bindings.rs
```diff
@@ -570,6 +589,67 @@ impl Bindings {
...
+            return None;
+        }
+        let defs = info.fixture_definitions(&param.id)?;
+        if !self.0.infer_pytest_fixture_types {
+            // `None` would mean "not a fixture parameter" and leave the parameter on the
+            // ordinary unannotated fallback path. Returning explicit Any records that pytest
+            // supplies this argument at runtime, which avoids noisy missing-annotation errors.
+            return Some(PytestFixtureParamHint::Any);
+        }
+        let selected = match class_key {
```
kinto0: @yangdanny97  suggests checking pytest's scoping rules to make sure the logic is consistent
https://docs.pytest.org/en/stable/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

pyrefly/lib/binding/pytest.rs
```diff
@@ -0,0 +1,160 @@
...
+        self.fixture_aliases
+            .iter()
+            .any(|alias| alias == name.as_str())
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct PytestFixtureDefinition {
+    pub name: Name,
+    pub return_type_key: ShortIdentifier,
```
kinto0: @yangdanny97:
I don't fully understand this: why do we store a function name, then later iterate every definition and check it against the name? is that the best way to do it?

pyrefly/lib/binding/pytest.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Clone, Debug)]
+pub struct PytestBindingInfo {
+    aliases: PytestAliases,
+    fixtures: SmallMap<Name, Vec<PytestFixtureDefinition>>,
+}
+
+impl PytestBindingInfo {
+    pub fn from_module(module: &ModModule) -> Option<Self> {
+        let aliases = PytestAliases::from_module(module);
```
kinto0, line 97: @yangdanny97:
is there a way to skip this entirely if the module does not import pytest

pyrefly/lib/binding/pytest.rs
```diff
@@ -0,0 +1,160 @@
...
+            aliases,
+            fixtures: SmallMap::new(),
+        })
+    }
+
+    pub fn aliases(&self) -> &PytestAliases {
+        &self.aliases
+    }
+
+    pub fn add_fixture_definition(&mut self, definition: PytestFixtureDefinition) {
```
kinto0: @yangdanny97 
> We store these fixtures in a flat list, which means scoping information is lost - not sure what the risk is of having duplicate fixtures in nested scopes

I don't think nested scopes work

asukaminato0721: fixed, removed the flat list and now store fixture definitions by pytest visibility scope.

pyrefly/lib/binding/pytest.rs
asukaminato0721: now we only create PytestBindingInfo if the module imports pytest/fixture; otherwise it is None.

pyrefly/lib/binding/pytest.rs
asukaminato0721: The map key is now the fixture name, so individual fixture records no longer duplicate the name.

yangdanny97 (approved):

PR #2081, opened 2026-01-12
Resolved incorrect parenthesized tuple type argument

### LINKED ISSUE #2063: A generic type argument accepts a tuple
### Describe the Bug

*Memo:
- pyrefly check
- pyrefly 0.46.1
- Python 3.14.0

A generic type argument accepts a tuple as shown below:

```python
       # ↓↓↓↓↓↓
v1: list[(int,)]
v1 = [0, 1, 2]       # No error
v1 = ['A', 'B', 'C'] # Error

       # ↓↓↓↓↓↓
v2: list[(str,)]
v2 = [0, 1, 2]       # Error
v2 = ['A', 'B', 'C'] # No error
```

```python
        # ↓↓↓↓↓↓↓↓↓↓
v1: tuple[(int, ...)]
v1 = (0, 1, 2)       # No error
v1 = ('A', 'B', 'C') # Error

        # ↓↓↓↓↓↓↓↓↓↓
v2: tuple[(str, ...)]
v2 = (0, 1, 2)       # Error
v2 = ('A', 'B', 'C') # No error
```

### Sandbox Link

_No response_

_No response_

Adist319:
# Summary

Pyrefly incorrectly flattened parenthesized tuple type arguments:
- `list[(int,)]` was interpreted as `list[int]` instead of `list[tuple[int]]`
- `dict[(int, str)]` was treated as two type arguments instead of one

**Root cause:** `Ast::unpack_slice` unconditionally unpacked all tuples, ignoring Python's AST `parenthesized` flag.

**Fix:** Check the `parenthesized` flag to distinguish:
- `dict[int, str]` (non-parenthesized) → unpack as 2 type args (unchanged)
- `list[(int,)]` (parenthesized) → single arg, interpret as `tuple[int]`

| Syntax | Before | After |
|--------|--------|-------|
| `list[(int,)]` | `list[int]` | `list[tuple[int]]` |
| `dict[(int, str)]` | `dict[int, str]` | Error: expected 2 args |
| `list[(int, str)]` | Error: expected 1 arg | `list[tuple[int, str]]` |
| `dict[int, str]` | `dict[int, str]` | `dict[int, str]` (unchanged) |

Fixes #2063

# Test Plan

- Added 6 test cases in `generic_basic.rs`
- Manual verification:
```python
v1: list[(int,)]
v1 = [0, 1, 2]          # Now errors (list[int] != list[tuple[int]])
v1 = [(0,), (1,), (2,)] # OK
```

Open to feedback or different approaches for this specific issue.

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -6757,8 +6757,8 @@
     {
       "code": -2,
       "column": 16,
-      "concise_description": "assert_type(tuple[int, *tuple[Any, ...]], tuple[int]) failed",
-      "description": "assert_type(tuple[int, *tuple[Any, ...]], tuple[int]) failed",
+      "concise_description": "assert_type(tuple[int, tuple[()]], tuple[int]) failed",
+      "description": "assert_type(tuple[int, tuple[()]], tuple[int]) failed",
       "line": 51,
       "name": "assert-type",
       "severity": "error",
--- a/conformance/third_party/conformance.result
+++ b/conformance/third_party/conformance.result
@@ -185,7 +185,7 @@
   "generics_typevartuple_concat.py": [],
   "generics_typevartuple_overloads.py": [],
   "generics_typevartuple_specialization.py": [
-    "Line 51: Unexpected errors ['assert_type(tuple[int, *tuple[Any, ...]], tuple[int]) failed']",
+    "Line 51: Unexpected errors ['assert_type(tuple[int, tuple[()]], tuple[int]) failed']",
     "Line 94: Unexpected errors ['assert_type(tuple[float, *tuple[Any, ...]], tuple[float]) failed']",
     "Line 153: Unexpected errors ['Unpacked argument cannot be used for type parameter T1']",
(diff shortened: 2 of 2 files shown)


Adist319:
The two failing tests on Windows CI seem **unrelated to this PR**:
  - `test_inlay_hint_paramspec_has_location`
  - `test_inlay_hint_typevartuple_has_location`

Happy to help investigate this separately if useful, but wanted to clarify it doesn't look like a regression from my changes.

yangdanny97:
Thanks for the PR, but I think this makes us behave differently from all other type checkers so I'm not in favor of making the change. We can discuss further on the issue and if a strong motivating reason comes up we can reconsider

Additionally, we shouldn't interpret the tuple of types as a tuple type. `(int,)` is not a valid way to write `tuple[int]`

For more context why, see https://jellezijlstra.github.io/why-cant-we.html#tuples-as-tuple-types

Adist319:
Thanks for the link, that makes sense. I agree it’s better to stay aligned with other type checkers here.

Issue #1662, opened 2025-11-22
[Feature] Add ability to set command to discover the interpreter


bmarroquin:
This would help in the situation where the IDE does not start the LSP with the Virtual Environment activated and where the virtual environment path is not fixed. See #1655 

```
[tool.pyrefly]
# poetry interpreter path is different for all users, use info command instead
# python-interpreter-path=""C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\test-pyrefly-lsp-Pe66xNAj-py3.13\Scripts\python.exe""
python-interpreter-find-cmd = ["poetry", "env", "info", "-e"]
```

Alternatively, just specifying the tool that supports a `run` sub command could make this generic
```
[tool.pyrefly]
venv_manager  = "poetry | hatch | uv | pipenv"
```
`interpreter = run([venv_manager, "run", "python", "-c", "import sys;print(sys.executable)"])`

stroxler:
Thanks for the suggestion! This seems like a great idea for the long tail of package managers that we probably can't fully support with baked-in heuristics and might not always use relative paths for their python interpreters (e.g. all the ones you listed, nix and guix, and I'm sure people are coming up with new ones)

I think it's at least possible you'd need the ability to run two commands in one shell - one to load up an environment and another to find the interpreter - for some use cases. In particular I think supporting a direnv-like setup where a shell hook sets a bunch of environment variables might be important - nix relies on PATH manipulation to load dependencies and I think that might not be terribly unusual

cc @connernilsen

connernilsen:
Yeah, I think this could be a useful thing to add! I'll open it up to OSS contribution

bmarroquin:
This would be solved by [pep-0832](https://peps.python.org/pep-0832/)

PR #2083, opened 2026-01-13
Auto-detect .venv interpreter when configured path is invalid

HamzaYslmn:
## Summary
This PR improves the interpreter discovery logic to automatically fall back to a `.venv` directory in the project root if the configured `python-interpreter-path` is invalid or refers to a non-existent file.

## Motivation
Modern Python workflows (using tools like `uv`, `poetry`, or standard `pip`) often create a local `.venv` directory in the project root by default. Currently, users are often required to explicitly configure the full path to this interpreter in `pyproject.toml` or [pyrefly.toml]

This change aims to reduce configuration boilerplate by attempting to auto-discover the `.venv` directory if the initially configured path fails validation. This makes the "out-of-the-box" experience much smoother for standard project layouts.

## Changes
- Modified `ConfigFile::configure` in [crates/pyrefly_config/src/config.rs]
- Added a check: if `interpreter.exists()` returns `false`, the config now attempts to find a valid environment using `crate::environment::venv::find(root)`.
- If a valid `.venv` is found, [pyrefly] automatically switches to using it instead of throwing an error or falling back to a system interpreter.

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -60,11 +60,13 @@ use crate::base::ConfigBase;
 use crate::base::UntypedDefBehavior;
 use crate::environment::environment::PythonEnvironment;
 use crate::environment::interpreters::Interpreters;
+use crate::environment::venv;
 use crate::error::ErrorConfig;
 use crate::error::ErrorDisplayConfig;
 use crate::finder::ConfigError;
 use crate::module_wildcard::Match;
 use crate::pyproject::PyProject;
+use crate::util::ConfigOrigin;
 
 pub static GENERATED_FILE_CONFIG_OVERRIDE: LazyLock<
     RwLock<SmallMap<ModulePathBuf, ArcId<ConfigFile>>>,
@@ -1006,11 +1008,10 @@ impl ConfigFile {
             }
             match self.interpreters.find_interpreter(self.source.root()) {
                 Ok(mut interpreter) => {
-                    if !interpreter.exists() {
-                         let root = self.source.root().unwrap_or(Path::new("."));
+                    if !interpreter.exists() && let Some(root) = self.source.root() {
                          // Attempt to auto-discover .venv if the configured path is invalid
-                         if let Some(found) = crate::environment::venv::find(root) {
-                             interpreter = crate::util::ConfigOrigin::auto(found);
+                         if let Some(found) = venv::find(root) {
+                             interpreter = ConfigOrigin::fallback(found);
                          }
                     }
 
--- a/crates/pyrefly_config/src/util.rs
+++ b/crates/pyrefly_config/src/util.rs
@@ -45,6 +45,10 @@ pub(crate) enum ConfigOrigin<T> {
     #[serde(skip)]
     Auto(T),
 
+    /// This value was auto-discovered as a fallback when configured path was invalid.
+    #[serde(skip)]
+    Fallback(T),
+
     /// This value was explicitly provided by the IDE using "workspace.configuration" using LSP protocol.
     #[serde(skip)]
     Lsp(T),
@@ -88,6 +92,7 @@ impl<T> Deref for ConfigOrigin<T> {
             Self::CommandLine(value)
             | Self::ConfigFile(value)
             | Self::Auto(value)
+            | Self::Fallback(value)
             | Self::Lsp(value) => value,
         }
     }
@@ -99,6 +104,7 @@ impl<T> DerefMut for ConfigOrigin<T> {
             Self::CommandLine(value)
             | Self::ConfigFile(value)
             | Self::Auto(value)
+            | Self::Fallback(value)
             | Self::Lsp(value) => value,
         }
     }
@@ -126,6 +132,11 @@ impl<T> ConfigOrigin<T> {
         Self::Auto(value)
     }
 
+    /// Construct a new [`ConfigOrigin::Fallback`] with the given value.
+    pub(crate) fn fallback(value: T) -> Self {
+        Self::Fallback(value)
+    }
+
     /// Construct a new [`ConfigOrigin::Lsp`] with the given value.
     pub(crate) fn lsp(value: T) -> Self {
         Self::Lsp(value)
@@ -142,6 +153,7 @@ impl<T> ConfigOrigin<T> {
             ConfigOrigin::ConfigFile(value) => ConfigOrigin::ConfigFile(f(value)),
             ConfigOrigin::CommandLine(value) => ConfigOrigin::CommandLine(f(value)),
             ConfigOrigin::Auto(value) => ConfigOrigin::Auto(f(value)),
+            ConfigOrigin::Fallback(value) => ConfigOrigin::Fallback(f(value)),
             ConfigOrigin::Lsp(value) => ConfigOrigin::Lsp(f(value)),
         }
     }
@@ -151,6 +163,7 @@ impl<T> ConfigOrigin<T> {
             ConfigOrigin::ConfigFile(ref value) => ConfigOrigin::ConfigFile(value),
             ConfigOrigin::CommandLine(ref value) => ConfigOrigin::CommandLine(value),
             ConfigOrigin::Auto(ref value) => ConfigOrigin::Auto(value),
+            ConfigOrigin::Fallback(ref value) => ConfigOrigin::Fallback(value),
             ConfigOrigin::Lsp(ref value) => ConfigOrigin::Lsp(value),
         }
     }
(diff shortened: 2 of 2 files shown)


connernilsen:
Hey @HamzaYslmn, thanks for making this!

I'm not sure I fully understand the use case you're trying to address though. Can you walk me through an example setup where you would expect to use this? Have you run into this problem before?

Right now, I'm kind of leaning toward the idea that explicitly configured interpreters are a source of truth, and we generally shouldn't be attempting to fall back to something else automatically, so I think trying to understand your reasoning behind this will help me to decide if this is the right approach, or if something else might work (maybe something like #1662).

> Currently, users are often required to explicitly configure the full path to this interpreter in `pyproject.toml` or `pyreflyl.toml`

I just want to clarify, there are a lot of cases where this isn't required and should just work 'by default'. Are you seeing issues with this, or do you have a setup incompatible with auto-configuration?

HamzaYslmn:
@connernilsen

My use case is: when I create projects with `uv` / `pip`, the layout is usually (global usage):

```
pyproject.toml
.venv/ (next to the project)
```

But Pyrefly can’t automatically detect the interpreter inside that `.venv`. Because of that, I end up having to add this to every project:

```toml
[tool.pyrefly]
python-interpreter-path = ".venv/Scripts/python.exe"
```

What my PR does is **not** to override anything. If an interpreter path is explicitly configured, it stays the source of truth. The change only applies when there is **no configured path** and auto-discovery fails , then it falls back to checking for a local `.venv` next to `pyproject.toml`.

So it’s a safe fallback, not a replacement.


Example:
image



Pyrefly cannot handle such situations. Have to manually configure .venv for every project, which is really tedious. Tools like uv automatically detect pyproject.toml and use the adjacent .venv, but with Pyrefly I still need to add an extra config each time.

WhoIsMrSentry:
I want to echo that I’m hitting the same problems outlined in this PR. The current behavior has caused confusion and extra overhead in my own workflow, particularly in cases similar to the ones referenced here.

I’ve spent a fair amount of time trying to understand whether the issues were coming from my setup or my code, only to realize they align very closely with what’s being discussed in this change.

From my perspective, this PR targets a real and practical limitation rather than a theoretical edge case. Having this fixed would remove a lot of unnecessary friction when using Pyrefly on larger or more complex projects. Really appreciate the work going into this.

crates/pyrefly_config/src/config.rs
```diff
@@ -975,7 +975,15 @@ impl ConfigFile {
                 ));
             }
             match self.interpreters.find_interpreter(self.source.root()) {
-                Ok(interpreter) => {
+                Ok(mut interpreter) => {
+                    if !interpreter.exists() {
+                         let root = self.source.root().unwrap_or(Path::new("."));
```
connernilsen: Can we get rid of the unwrap? `.` is kind of a complicated thing to use for tools, especially if they can be used as language servers. When run from an IDE, `.` is often where the IDE was started from, which doesn't make sense in most contexts.
```suggestion
                    if !interpreter.exists() && let Some(root) = self.source.root() {
```

crates/pyrefly_config/src/config.rs
```diff
@@ -975,7 +975,15 @@ impl ConfigFile {
...
+                         // Attempt to auto-discover .venv if the configured path is invalid
+                         if let Some(found) = crate::environment::venv::find(root) {
+                             interpreter = crate::util::ConfigOrigin::auto(found);
```
connernilsen: Can you `use` these crates so that we just call `venv::find()` and `ConfigOrigin` directly here?

Also, can you make another `ConfigOrigin` variant for `Fallback` or something similar?

HamzaYslmn:
I added them to utils so they would be easier to track, and I made the changes you requested.

crates/pyrefly_config/src/config.rs
HamzaYslmn: > Can we get rid of the unwrap? `.` is kind of a complicated thing to use for tools, especially if they can be used as language servers. When run from an IDE, `.` is often where the IDE was started from, which doesn't make sense in most contexts.

You're right

crates/pyrefly_config/src/config.rs
HamzaYslmn: done

connernilsen:
Awesome, thanks for the changes! I'll pull this in and merge it, so you should hopefully see it in our release next week.

grievejia (changes requested):

connernilsen:
Hey @HamzaYslmn, we've been discussing these changes internally, and decided this might not fully be the approach we want to take with solving this problem. There are a few changes in a different direction.

First though, I think I actually misunderstood the use case you were describing. The problem you're running into actually seems like more of a bug than unimplemented functionality.

>What my PR does is not to override anything. If an interpreter path is explicitly configured, it stays the source of truth. The change only applies when there is no configured path and auto-discovery fails , then it falls back to checking for a local .venv next to pyproject.toml.

(Just a quick note that this comment contradicts what's said in the PR summary in a few ways, but is the more recent description of the diff from our conversation, so I'm going to assume this is what you intended to do. Let me know if you actually meant what's in the PR description, and we can have a discussion around that instead.)

Technically, what your diff does is search for a venv next to `config.source.root()` if the found interpreter path does not exist. This should really only be possible if an explicitly configured interpreter (from the command line, config, IDE, or activated conda/venv) is provided to Pyrefly but doesn't exist. Falling back to finding a venv next to `config.source.root()` should already be happening after the above options are checked, but in your case, that's likely not happening. So maybe there is a value set somewhere? I would assume an activated conda or venv environment that's invalid.

I think there are a few things we can do here to continue forward:
1. Produce an error from `find_interpreter` if the returned path does not exist for any explicitly configured fields. This we'll pass back to `Config::configure()` so that it can be added to the list of configure errors that are collected.
2. Figure out if there is a bug in how your interpreter is being selected (@WhoIsMrSentry this is probably the same for you too).
3. Determine if continuing down the list of automatically configured interpreters is the right way to continue (I think it would be a reasonable thing to do!). This would change the signature of the `find_interpreters()` function and move the `if !interpreter.exists()` logic into `find_interpreter()` (sorry for the churn here).

To start looking into whether there's a bug on interpreter selection, can both of you send your `pyrefly.toml` or `[tool.pyrefly]` sections of your `pyproject.toml` here? Can you also send me the output from `pyrefly dump-config`.

HamzaYslmn:
> `pyproject.toml` here?
```
[project]
name = "korustan-metal"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
...
]

[tool.pyrefly]
project-includes = ["src"]
search-path = ["src"]
ignore-errors-in-generated-code = true
python-interpreter-path = ".venv/Scripts/python.exe"

[tool.pyrefly.errors]
bad-argument-type = false
implicit-import = false
bad-assignment = false
bad-return = false
missing-attribute = false
bad-function-definition = false
no-matching-overload = false
unsupported-operation = false
```


image

I have to add this line to every project:
```
python-interpreter-path = “.venv/Scripts/python.exe”
```

Normally, it should find the .venv or venv alongside pyproject.toml. But it lacks this capability; there's no provision for it in the code. That's why I added it.

Pyrefly requires the “.venv” folder to be located in the workspace root.
image


That's my intention.

image

HamzaYslmn:
I also believe that manually entered routes are sacred, but if there is no route, find one automatically.

connernilsen:
@HamzaYslmn 

> I also believe that manually entered routes are sacred, but if there is no route, find one automatically.

I totally agree, and it is what we *should* already be attempting. I didn't realize it when I was originally looking at your diff, but the changes you make attempt to do the same exact venv lookup we're already doing.

With the way your project is set up, we should be able to automatically find your config. I'm going to add extra logging around how we detect the interpreter to make things more debuggable for you in the future.

My best guess as to what's happening for you is that by explicitly setting the interpreter, we bypass the IDE interpreter check we do. I think what may be happening is that your IDE is providing Pyrefly with a default (or selected from another session), and that takes precedence over everything except for interpreters in configs.

Which IDE do you use? I can try to figure out how it's done for your IDE if you're not sure, but if you can, see if there's a way to select a Python interpreter. Don't forget to remove the explicit interpreter from your config.

HamzaYslmn:
The code I implemented expects the pyproject path, whereas your latest version points to the project root. It’s all good, though; you'll figure it out once you get a chance to look at the logs. I’ve moved on to using uv + ty in the meantime.

connernilsen:
@HamzaYslmn Project root and Pyproject path should be the same thing. What your PR passes into the venv finder is the same path we pass to `interpreters::find_interpreter`, which eventually calls the same function. I can look more into this if you can throw together a repro, but I think there's some misalignment between what I think you're describing and what you're trying to describe. 

@WhoIsMrSentry are you still running into issues here? Maybe we can open a separate issue for this if there are still problems to discuss.

PR #2084, opened 2026-01-13
fix(definition): import submodule as alias

aaron-ang:
# Summary

Resolve goto-definitions for submodule imports as aliases. Previously, only hover information was available.
For example,
```python
from torch.nn import (
    attention as attention,
    functional as functional,
    init as init,
    modules as modules,
    parallel as parallel,
    parameter as parameter,
    utils as utils,
)
```

# Test Plan


yangdanny97 (approved):

PR #2088, opened 2026-01-13
Test disabled reverse relations

migeed-z:
Summary: We can use the "+" symbol to avoid creating a reverse relation. This is important to test so we can make sure we are avoiding creating that field if it is not needed

Reviewed By: aleivag, yangdanny97

Differential Revision: D90517596


yangdanny97 (approved):

PR #2100, opened 2026-01-13
Add JSON schemas for pyrefly.toml and pyproject.toml validation

### LINKED ISSUE #536: Add Pyrefly config to schemastore
[Schemastore](https://www.schemastore.org/) is a nice config file validation framework supported by most major editors. It would be nice if we could add one here for Pyrefly's `pyproject.toml` and `pyrefly.toml` config options.

Here's an example of [Pyright's](https://raw.githubusercontent.com/microsoft/pyright/main/packages/vscode-pyright/schemas/pyrightconfig.schema.json)

Prathamesh-tech-eng:
**Description:**
This PR adds JSON schemas (`pyrefly.json` and `pyproject-tool-pyrefly.json`) to enable editor support (auto-completion, validation, and documentation) for Pyrefly configuration files in `pyrefly.toml` and `pyproject.toml`.

Fixes #536

**Notes:**
- I've constructed this schema based on `crates/pyrefly_config/src/config.rs` and and `crates/pyrefly_config/src/base.rs`
- I used kebab-case for keys (e.g., `project-includes`) as defined by the Serde structs.
- I placed the schema in the  `schemas/` directory

### DIFF
--- a/schemas/pyproject-tool-pyrefly.json
+++ b/schemas/pyproject-tool-pyrefly.json
@@ -9,7 +9,7 @@
       "type": "object",
       "properties": {
         "pyrefly": {
-          "$ref": "https://pyrefly.org/schemas/pyrefly.json"
+          "$ref": "pyrefly.json"
         }
       }
     }
--- a/schemas/pyrefly.json
+++ b/schemas/pyrefly.json
@@ -9,21 +9,24 @@
       "type": "object",
       "properties": {
         "errors": {
-          "description": "Configure the severity for each kind of error that Pyrefly emits. Set error code to true to enable, false to disable. Any valid Pyrefly error code can be used, not just the ones listed here.",
+          "description": "Configure the severity for each kind of error that Pyrefly emits. Set error code to true to enable (default severity), false to disable, or use a severity string ('ignore', 'info', 'warn', 'error'). Any valid Pyrefly error code can be used, not just the ones listed here.",
           "type": "object",
           "additionalProperties": {
-            "type": "boolean"
+            "oneOf": [
+              { "type": "boolean" },
+              { "type": "string", "enum": ["ignore", "info", "warn", "error"] }
+            ]
           },
           "properties": {
-            "bad-assignment": {"type": "boolean"},
-            "bad-return": {"type": "boolean"},
-            "missing-import": {"type": "boolean"},
-            "invalid-argument": {"type": "boolean"},
-            "assert-type": {"type": "boolean"},
-            "bad-override": {"type": "boolean"},
-            "incompatible-type": {"type": "boolean"},
-            "unused-import": {"type": "boolean"},
-            "undefined-variable": {"type": "boolean"}
+            "bad-assignment": {},
+            "bad-return": {},
+            "missing-import": {},
+            "invalid-argument": {},
+            "assert-type": {},
+            "bad-override": {},
+            "incompatible-type": {},
+            "unused-import": {},
+            "undefined-variable": {}
           },
           "default": {}
         },
@@ -74,9 +77,24 @@
           "type": "array",
           "items": {
             "type": "string",
-            "enum": ["type", "pyrefly", "mypy", "pyright", "pyre"]
+            "enum": ["type", "pyrefly", "mypy", "pyright", "pyre", "ty"]
           },
           "default": ["type", "pyrefly"]
+        },
+        "tensor-shapes": {
+          "description": "Whether to enable tensor shape checking.",
+          "type": "boolean"
+        },
+        "recursion-depth-limit": {
+          "description": "Maximum recursion depth for type evaluation. Set to 0 to disable the limit.",
+          "type": "integer",
+          "minimum": 0
+        },
+        "recursion-overflow-handler": {
+          "description": "How should Pyrefly handle recursion overflow during type evaluation?",
+          "type": "string",
+          "enum": ["break-with-placeholder", "panic-with-debug-info"],
+          "default": "break-with-placeholder"
         }
       }
     }
@@ -94,7 +112,7 @@
           "items": {
             "type": "string"
           },
-          "default": ["**/*.py*"]
+          "default": ["**/*.py*", "**/*.ipynb"]
         },
         "project-excludes": {
           "description": "The glob patterns used to describe which files to avoid type checking as a way to filter files that match project-includes.",
@@ -173,6 +191,10 @@
           "type": "boolean",
           "default": false
         },
+        "baseline": {
+          "description": "Path to a baseline file for suppressing existing errors.",
+          "type": "string"
+        },
         "build-system": {
           "description": "Pyrefly supports integrating into build systems to discover targets to type check and their dependencies. Currently supports Buck2 and custom queries.",
           "type": "object",
@@ -182,9 +204,18 @@
               "type": "string",
               "enum": ["buck", "custom"]
             },
-            "repo_root": {
-              "description": "Path to the root of the repository.",
-              "type": "string"
+            "ignore-if-build-system-missing": {
+              "description": "Whether to silently ignore the build system configuration if the build system is not found.",
+              "type": "boolean",
+              "default": false
+            },
+            "search-path-prefix": {
+              "description": "Prefix paths to prepend to search paths discovered by the build system.",
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "default": []
             },
             "isolation-dir": {
               "description": "The isolation dir for Buck2 to use.",
--- a/schemas/test-pyproject.toml
+++ b/schemas/test-pyproject.toml
@@ -25,6 +25,9 @@ untyped-def-behavior = "check-and-infer-return-type"
 infer-with-first-use = true
 ignore-errors-in-generated-code = false
 disable-type-errors-in-ide = false
+tensor-shapes = false
+recursion-depth-limit = 100
+recursion-overflow-handler = "break-with-placeholder"
 
 # Import handling
 replace-imports-with-any = ["sympy.*", "*.series"]
@@ -35,11 +38,15 @@ use-ignore-files = true
 permissive-ignores = false
 enabled-ignores = ["type", "pyrefly"]
 
+# Baseline
+# baseline = "baseline.json"
+
 # Error configuration
 [tool.pyrefly.errors]
 bad-assignment = false
 bad-return = true
-missing-import = true
+missing-import = "warn"
+invalid-argument = "error"
 
 # SubConfig example 1: Test files
 [[tool.pyrefly.sub-config]]
--- a/schemas/test-pyrefly.toml
+++ b/schemas/test-pyrefly.toml
@@ -24,6 +24,9 @@ untyped-def-behavior = "check-and-infer-return-type"
 infer-with-first-use = true
 ignore-errors-in-generated-code = false
 disable-type-errors-in-ide = false
+tensor-shapes = false
+recursion-depth-limit = 100
+recursion-overflow-handler = "break-with-placeholder"
 
 # Import handling
 replace-imports-with-any = ["sympy.*", "*.series"]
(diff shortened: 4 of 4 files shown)


schemas/pyrefly.json
```diff
@@ -0,0 +1,266 @@
...
+          "untyped-def-behavior": {
+            "description": "Override untyped function behavior for matched files.",
+            "type": "string",
+            "enum": ["check-and-infer-return-type", "check-and-infer-return-any", "skip-and-infer-return-any"]
+          },
+          "ignore-errors-in-generated-code": {
+            "description": "Override generated code error handling for matched files.",
+            "type": "boolean"
+          }
+        }
```
connernilsen: I'm not sure if this is possible, so if you struggle to find a way to do this, don't worry about it.

Could we pull this out into a subschema and do two things with it?
1. flatten the subschema into the top-level object
2. reuse that sub-schema here

This would help us avoid redefining definitions in multiple places, which means we won't have to remember to update multiple things if there are changes to the config base object.

Again, not sure if flattening is possible, but it would be awesome if it works!

schemas/pyproject-tool-pyrefly.json
```diff
@@ -0,0 +1,277 @@
+{
```
connernilsen, line 1: If you look at how the [`pyproject.toml` in schemastore works](https://github.com/SchemaStore/schemastore/blob/42c1f83e9c33ca07ba2f74e5ce6870d2b3836820/src/schemas/json/pyproject.json#L980-L989), we might be able to get away with just the `pyrefly.json` config below and reference it in a pyproject definition.

When we merge into schemastore, I'd like to follow those examples, so this file will be redundant.

schemas/pyrefly.json
```diff
@@ -0,0 +1,266 @@
...
+      "description": "The value used with conditions based on type checking against sys.platform values.",
+      "type": "string",
+      "default": "linux",
+      "enum": ["linux", "darwin", "win32", "cygwin", "freebsd", "openbsd", "netbsd", "aix"]
+    },
+    "python-version": {
+      "description": "The Python version to use for type checking. Format: <major>[.<minor>[.<micro>]]",
+      "type": "string",
+      "default": "3.13.0",
+      "pattern": "^\\d+(\\.\\d+)?(\\.\\d+)?$"
```
connernilsen: Awesome validation!

schemas/pyrefly.json
```diff
@@ -0,0 +1,266 @@
...
+            }
+          },
+          "then": {
+            "required": ["command"],
+            "properties": {
+              "command": true,
+              "repo_root": true
+            }
+          }
+        }
```
connernilsen: Is it possible to set `not` conditions here for the custom-only fields when using the `buck` type, and vice versa for `custom` with buck-only fields?

It might be possible to do a subschema for the `buck` and `custom` fields, specify both subschemas in the field definitions, and just list the subschemas here to simplify the conditions.

schemas/pyrefly.json
```diff
@@ -0,0 +1,266 @@
...
+          "if": {
+            "properties": {
+              "type": {"const": "buck"}
+            }
+          },
+          "then": {
+            "properties": {
+              "isolation-dir": true,
+              "extras": true
+            }
```
connernilsen: Just to double check, these aren't required fields, right? Just fields that are allowed with the `buck` type?

schemas/pyrefly.json
```diff
@@ -0,0 +1,266 @@
...
+          "if": {
+            "properties": {
+              "type": {"const": "custom"}
+            }
+          },
+          "then": {
+            "required": ["command"],
+            "properties": {
+              "command": true,
+              "repo_root": true
```
connernilsen: Repo root should be for all types, not just `custom`

Prathamesh-tech-eng:
@connernilsen I've addressed all the feedback! Here is a summary of the changes:

1. Extracted configBase to avoid repetition 
I created a $defs/configBase section containing all the shared properties (like errors, untyped-def-behavior, replace-imports-with-any, etc.).

2. Flattened ConfigBase into the root 
All configBase properties are now duplicated in the main properties section so they're available at the root level. This ensures the schema works correctly with additionalProperties: false.

3. Applied ConfigBase to sub-config 
The sub-config items now use allOf with { "$ref": "#/$defs/configBase" }. This ensures that any future changes to the base config automatically propagate to sub-configs without manual duplication.

4. Fixed Build System Logic 
I cleaned up the validation logic:

Moved repo_root: It is now a top-level property inside build-system, accessible to both buck and custom types (addressing your comment about it being for all types).

Added strict not conditions:

If type: "buck", the schema explicitly forbids command.

If type: "custom", the schema requires command but forbids isolation-dir and extras.

Clarified optional fields: isolation-dir and extras are defined as properties but not required (answering your question about them).

5. Simplified pyproject-tool-pyrefly.json 
I updated this to just reference pyrefly.json via $ref rather than duplicating 270+ lines. This aligns with the SchemaStore pattern you mentioned.

Once this is approved, I'll proceed with opening the PR to SchemaStore.

Prathamesh-tech-eng:
Hi @connernilsen ,

Hope you're doing well. It's been a while since we last connected, so I just wanted to follow up and check if there have been any updates regarding the changes.

connernilsen:
Hey @Prathamesh-tech-eng, sorry, I've had it open in my browser but haven't been able to spare the time to look into it yet. I'll sit down sometime this week and give a review

.gitignore
```diff
@@ -1,3 +1,8 @@
 debug.json
 debug.js
 target/
+
+# Ignore all files in schemas/ except specific JSON schemas
+schemas/*
```
connernilsen, line 6: qq: are there other schemas that you're expecting to be added here?

schemas/pyrefly.json
```diff
@@ -0,0 +1,328 @@
...
+                "type": "string"
+              }
+            }
+          }
+        ]
+      },
+      "default": []
+    }
+  },
+  "additionalProperties": false
```
connernilsen: Can we allow additional properties until we're sure we have this working really well?

schemas/pyrefly.json
```diff
@@ -0,0 +1,328 @@
...
+    "configBase": {
+      "type": "object",
+      "properties": {
+        "errors": {
+          "description": "Configure the severity for each kind of error that Pyrefly emits. Set error code to true to enable, false to disable.",
+          "type": "object",
+          "additionalProperties": {
+            "type": "boolean"
+          },
+          "properties": {
```
connernilsen, line 20: We have a bunch of other error kinds, but this is a good start. You don't have to add them all in this diff, but just to double check, errors that aren't listed here can be used right?

schemas/pyrefly.json
```diff
@@ -0,0 +1,328 @@
...
+      "items": {
+        "type": "string"
+      },
+      "default": []
+    },
+    "python-platform": {
+      "description": "The value used with conditions based on type checking against sys.platform values.",
+      "type": "string",
+      "default": "linux",
+      "enum": ["linux", "darwin", "win32", "cygwin", "freebsd", "openbsd", "netbsd", "aix"]
```
connernilsen: This one should also be non-exhaustive. Is there a way to specify that?

schemas/pyrefly.json
```diff
@@ -0,0 +1,328 @@
...
+    "skip-interpreter-query": {
+      "description": "Skip querying any interpreters and do not do any environment autoconfiguration. This means that Pyrefly will take hard-coded defaults for python-version and python-platform, and will use an empty site-package-path.",
+      "type": "boolean",
+      "default": false
+    },
+    "typeshed-path": {
+      "description": "Override the version of typeshed that's being used for type checking. The provided path should point to the root of typeshed.",
+      "type": "string"
+    },
+    "errors": {
```
connernilsen: Oh wait, is there no way to fold the `configBase` def you made above into here?

connernilsen (changes requested):
Sorry just a few more requests.

Also test instructions if you have just so I can validate some things

.gitignore
Prathamesh-tech-eng: No, just pyrefly.json and pyproject-tool-pyrefly.json. I've updated the .gitignore to also track the supporting files (test configs, validation script, docs).

schemas/pyrefly.json
Prathamesh-tech-eng: Done, changed additionalProperties to true.

schemas/pyrefly.json
Prathamesh-tech-eng: Yes ,   additionalProperties: { type: boolean } allows any error code string. The listed ones just provide autocomplete hints. Updated the description to clarify this.

schemas/pyrefly.json
Prathamesh-tech-eng: Done,  replaced enum with examples. The Rust PythonPlatform::new() accepts any string, so the schema now does too.

schemas/pyrefly.json
Prathamesh-tech-eng: Done,  the top-level now uses allOf: [$ref configBase, {top-level-only properties}] instead of duplicating. Both top-level and sub-config share the same configBase definition.

Prathamesh-tech-eng:
HI @connernilsen, Addressed all the feedback ,  here's a summary of the changes:
- additionalProperties set to true
- python-platform changed from enum to examples (non-exhaustive)
- errors description clarified to note any valid error code works
- Top-level schema now uses allOf with $ref to configBase (no duplication)
- .gitignore updated to track test files and docs

To test : Run pip install jsonschema toml referencing then python schemas/validate_schemas.py. It validates both test TOML files against the schemas. You can also test in VS Code with the "Even Better TOML" extension.

I looked into schemars (https://github.com/GREsau/schemars), which can derive JSON Schema from Rust structs with #[derive(JsonSchema)]. That could auto-generate these schemas from ConfigBase/ConfigFile and we could add a CI check to detect drift between the generated schema and what's in schemastore. Happy to explore that as a follow-up PR if you think it's worth pursuing.

Prathamesh-tech-eng:
I have also Performed additional fixes found during cross-referencing the schema against the Rust config source of truth:

- **`errors` values**: Now accepts both booleans and severity strings (`"ignore"`, `"info"`, `"warn"`, `"error"`) via `oneOf`, matching the Rust `Severity` enum
- **Missing `configBase` fields**: Added `tensor-shapes`, `recursion-depth-limit`, `recursion-overflow-handler` (these were present in Rust's `ConfigBase` but missing from the schema; they also apply to `sub-config` items via `$ref`)
- **`enabled-ignores`**: Added missing `"ty"` variant (for `# ty: ignore` directives)
- **`project-includes` default**: Fixed from `["**/*.py*"]` to `["**/*.py*", "**/*.ipynb"]` to match the Rust default
- **`baseline`**: Added missing top-level string field for suppressing existing errors
- **`build-system`**: Removed non-existent `repo_root` field (computed at runtime, not a config option), added `ignore-if-build-system-missing` (boolean) and `search-path-prefix` (array) which were present in the Rust `BuildSystem` struct
- **`pyproject-tool-pyrefly.json`**: Changed `$ref` from `https://pyrefly.org/schemas/pyrefly.json` to relative `pyrefly.json` so it resolves locally without depending on the URL being live
- **Test files**: Updated to exercise severity string values (`missing-import = "warn"`, `invalid-argument = "error"`) and the newly added fields

connernilsen:
> I looked into schemars (https://github.com/GREsau/schemars), which can derive JSON Schema from Rust structs with #[derive(JsonSchema)]. That could auto-generate these schemas from ConfigBase/ConfigFile and we could add a CI check to detect drift between the generated schema and what's in schemastore. Happy to explore that as a follow-up PR if you think it's worth pursuing.

Yeah, that would be awesome if you're interested! If not, then we can file that as a task and come back to it later!

Also, thanks for doing all of this, it looks awesome! I'm going to pull this in and start testing it, and I'll merge it and let you know if there are any followups.

Prathamesh-tech-eng:
Thanks! Yeah, I'm definitely interested.  I'll start looking into schemars and see how it can be integrated to auto-generate the schemas from the Rust config structs. I'll open a follow-up PR once I have something working. Might ping you if I run into any blockers!

And glad the schemas look good, let me know if anything comes up during testing!

PR #2101, opened 2026-01-13
impl move refactor #364

asukaminato0721:
# Summary

Added new refactor.move actions for moving module members and for making local functions/methods top‑level, then wired them into the LSP code action pipeline. The new logic moves a top‑level definition into a sibling module and inserts an import back into the source module to keep references working.

For local functions/methods, it lifts the function to module scope and replaces methods with a class‑body shim (foo = foo / staticmethod / classmethod) so existing call sites still work. New tests cover module member move, local function lift, and method/staticmethod lift. 

Fixes part of #364

https://www.jetbrains.com/help/pycharm/move-refactorings.html

# Test Plan

Added tests in pyrefly/lib/test/lsp/code_actions.rs

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/move_module.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/move_module.rs
@@ -136,7 +136,9 @@ pub(crate) fn make_local_function_top_level_code_actions(
             return None;
         }
     }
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #2102, opened 2026-01-13
impl convert module

asukaminato0721:
# Summary

Fixes part of #364

Implemented LSP refactor actions to mirror PyCharm’s “convert module ↔ package” behavior: module→package renames foo.py to `foo/__init__.py`, and package→module is only offered when the package directory contains only `__init__.py` (empty package), keeping content in `__init__.py` as the module body.

Added workspace-edit refactor actions (rename + delete operations; gated on client support for documentChanges/resourceOperations).

# Test Plan

Added LSP interaction tests and fixtures.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/convert_module_package.rs
+++ b/pyrefly/lib/lsp/non_wasm/convert_module_package.rs
@@ -1,0 +1,191 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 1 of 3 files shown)


stroxler (commented):
Should the new helpers be in a dedicated module rather than `server.rs`? I think in general we're trying to avoid too much growth in the central LSP modules as we add refactoring capabilities which often require a lot of code

PR #2109, opened 2026-01-14
fix: 'Go to symbol' now includes methods inside a class

### LINKED ISSUE #2051: 'Go to symbol' excludes methods inside a class
### Describe the Bug

When using either of the commands in VS code:
1. Go to symbol in Editor
2. Go to symbol in workspace

This command doesn't reveal methods within a class. Only functions, or classes

Is this intentional?
Is this configurable?

### Sandbox Link

_No response_

Extension Version 0.46.3

Karman-singh15:
# Summary

Logic was added to post-process symbols inside a ClassDef.
it allows Stmt::FunctionDef to remain simple (always creating a FUNCTION), and shifts the responsibility of context-awareness (knowing it is a method) to the parent ClassDef.
Any child symbol with SymbolKind::FUNCTION is now converted to SymbolKind::METHOD

Fixes #2051 

# Test Plan

I updated the existing test 
test_document_symbols_normal_file in pyrefly/lib/test/lsp/lsp_interaction/document_symbols.rs
The test uses an existing file normal.py which contains NormalClass and normal_method.
I updated the test assertion to verify that normal_method inside NormalClass has SymbolKind::METHOD.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/document_symbols.rs
+++ b/pyrefly/lib/lsp/non_wasm/document_symbols.rs
@@ -61,6 +61,14 @@ fn recurse_stmt_adding_symbols<'a>(
         Stmt::ClassDef(stmt_class_def) => {
             let mut children = Vec::new();
             children.append(&mut recursed_symbols);
+
+            // Functions defined inside a class are methods.
+            for child in &mut children {
+                if child.kind == lsp_types::SymbolKind::FUNCTION {
+                    child.kind = lsp_types::SymbolKind::METHOD;
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


stroxler:
Thanks for the PR! Looks good to me, I'll try to get a second review and merge in the next day or two

Heads up that the red CI signal is just format, you can run `cargo fmt` to avoid it in the future but I can patch it up so no need to worry about it on this commit

stroxler (approved):

grievejia (approved):

kinto0:
This doesn't fix the issue. this change simply changed the symbol kind of document symbols if they had existed.

PR #2142, opened 2026-01-17
Fix list concatenation under contextual list hint

### LINKED ISSUE #2141: Casting list of union type to list[Base]
## Summary

Pyrefly reports a `bad-return` error when returning a list containing instances of subclasses, even when the return type is explicitly annotated as `list[Base]`. The inferred type `list[A | B]` should be assignable to `list[Base]` when both `A` and `B` inherit from `Base`.

```python

class Base(ABC):

    @abstractmethod
    def foo(self, x: int) -> None: ...


class A(Base):

    def foo(self, x: int) -> None:
        print(x)

class B(Base):

    def foo(self, x: int) -> None:
        pass

def return_object(name: str) -> list[Base]:
    return [A()] + [B()]


def return_object_work_for_non_list(name: str) -> Base:
    o = None
    if name == "a":
        o = A()
    else:
        o = B()
    reveal_type(o)
    return o

```

## Actual Behavior

```
ERROR sandbox.py:21:12-23: Returned type `list[A | B]` is not assignable to declared return type `list[Base]` [bad-return]
```

## Expected Behavior

No error. Since `A` and `B` are both subclasses of `Base`, the union `A | B` is a subtype of `Base`. While `list` is technically invariant, the list is created locally and immediately returned—there's no opportunity for type-unsafe mutation.

Mypy accepts this code without error.

Can we do something like **if T is assignable to V** then **list[T] is assignable to list[V]**?

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEq2AxnRDcbpQC50CCAQgGEANA2xwulVEy40YXABa5MAHXRqmUVHDh1%2B2mAAoBggJSI1autboABRhKky5i5VZuYYYOmFy5DcDBQYKL4iKzoXKZ0ALQAfHQAcrjoMOGEGZbq6JrauryG%2BoHmWTZ0nt6%2B-oHBoeEQkdHxSSlp7mXWxJQNXIb4plm5OnqFBiXZZRU%2BfgFBIXRhEVGxCcmpFugdNsR5WVOU8gCulOgA%2BqhQUKe42ABWMDJwhk0JUBASANpFMAC6G2UXK43e6POgAXjoHwKph%2BdAA1JD%2BM8fu06AcuMdNoDrncHlw4GoQMIQIcuNA4CRyIgqLQxCw2BxuHwhKIHJJpLJ5EpVOgwNR6FxSMQGgBzVjsTg8A4ANxgF1OguIMEGWmG32MQnGqPs4nZzi5bk2Hi802qczqSxeLXWdAyhCyGlV%2BVGxQ2qKmVVmtQW9UaK2tbSNWy6PT6A2yQ10SO%2BWqD5RNnpq80WPSta0DW06u2y%2ByOJxxIN66FQcnCjitb0%2B3z%2BqPRmMh0NhCK%2ByIdWBNdfzwLxpwA7pwANanXyUU7oFKnStFktpOjl-3ff42XDggOoiDeYtycEQlQgVB7pdbFcQ6GooKBI8dE8jcNlGVyq6Koy4O82TubXCE4lkA5gKCkIQXC0FAFAAMR0AACqQf4AXOGA4AQdBMCkkAiscqBkik9roBBADKMAwHQChcFwxBwIgAD0lG-l4AGEJwIqUTA6CUZguBMHAlEobyEDoVIWGsdMlAMNKqDQIwsDIahfEYYJdC4MQgkUmoZCuOgMSypQcAQCkq57gAzIQACMABMe6EgAvsSHIQLKABi0AwBQpLkpSICWUAA

_No response_

praskr-wisdom:
Fixes #2141.

### What
When a `list[T]` contextual hint is available (assignment/return annotation) and **both** operands of `+` are fresh list-producing expressions (list literal or list comprehension), propagate the hint into both operands.

This makes cases like:
- `l2: list[Base] = [A()] + [B()]`
- `def f() -> list[Base]: return [A()] + [B()]`
work without making `list` covariant globally.

### Why
List literals already use contextual typing via `decompose_list`, but `+` previously inferred operands without passing the hint (except `[X] * int`). That caused `[A()] + [B()]` to be inferred as `list[A | B]` and fail against `list[Base]` due to invariance.

### Tests
- Added assignment regression test covering `l2` and ensuring no-context `reveal_type([A()] + [B()])` remains `list[A | B]`.
- Added return regression test for #2141.

### Notes
The hint is only propagated when *both* operands are fresh to avoid masking invariance errors for expressions involving non-fresh operands.

### DIFF
--- a/pyrefly/lib/alt/operators.rs
+++ b/pyrefly/lib/alt/operators.rs
@@ -201,6 +201,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    /// Returns true if the expression is a "fresh" list-producing construct
+    /// (list literal or list comprehension) that can safely be inferred under
+    /// a contextual list type hint.
+    fn is_fresh_list_expr(expr: &Expr) -> bool {
+        matches!(expr, Expr::List(_) | Expr::ListComp(_))
+    }
+
     pub fn binop_infer(
         &self,
         x: &ExprBinOp,
@@ -235,6 +242,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             } else {
                 lhs = self.expr_infer(&x.left, errors);
             }
+        } else if x.op == Operator::Add
+            && hint.is_some()
+            && hint
+                .as_ref()
+                .map(|h| self.decompose_list(*h).is_some())
+                .unwrap_or(false)
+            && Self::is_fresh_list_expr(&x.left)
+            && Self::is_fresh_list_expr(&x.right)
+        {
+            // If we have a list[T] contextual hint and BOTH operands are fresh
+            // list-producing expressions (list literal or list comprehension), pass the
+            // hint to both operands so that e.g. `[A()] + [B()]` in a `list[Base]` context
+            // is inferred as `list[Base]` instead of `list[A | B]`.
+            //
+            // We require both operands to be fresh to avoid unsoundness: if one operand
+            // is a variable like `xs: list[A]`, we should not coerce the result to
+            // `list[Base]` because that would hide invariance violations.
+            lhs = self.expr_infer_with_hint(&x.left, hint, errors);
+            rhs = self.expr_infer_with_hint(&x.right, hint, errors);
         } else {
             lhs = self.expr_infer(&x.left, errors);
             rhs = self.expr_infer(&x.right, errors);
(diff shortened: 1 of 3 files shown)


praskr-wisdom:
hey, sorry I didn't get time to get back to this. Feel free to fix the conflicts and merge otherwise I can do it tomorrow

PR #2143, opened 2026-01-17
Show instance type in constructor signatures

### LINKED ISSUE #1805: [lsp] Seeing signatures that contain `-> None` when constructing a class is confusing
See screenshots. If I'm constructing a class or hovering over the name of a class with a custom `__init__` method, I see the `__init__` signature, which is really nice, but the `-> None` at the end confused me the first time I saw it, because I thought it was saying that the constructor call returned `None`. It would be nice if we could replace the `None` with the class name for display.

Image
Image

sgavriil01:
Fixes #1805

   Shows the class instance type instead of `-> None` for constructor calls in both signature help and hover.

   ## Changes
   - Constructor calls like `Person()` now show `-> Person` instead of `-> None`
   - Direct `__init__` calls still show `-> None` 
   - Generic constructors show the specialized type (e.g., `Box[str]`)

   ## Tests
   - Added 6 hover tests covering constructors, direct __init__ calls, methods, and arguments
   - Added 1 signature help test for method calls
   - All 24 signature help tests passing
   - All 73 hover tests passing

### DIFF
--- a/pyrefly/lib/lsp/wasm/signature_help.rs
+++ b/pyrefly/lib/lsp/wasm/signature_help.rs
@@ -37,6 +37,34 @@ use crate::types::callable::Param;
 use crate::types::callable::Params;
 use crate::types::types::Type;
 
+pub(crate) fn is_constructor_call(callee_type: Type) -> bool {
+    matches!(callee_type, Type::ClassDef(_))
+        || matches!(callee_type, Type::Type(inner) if matches!(inner.as_ref(), Type::ClassType(_) | Type::ClassDef(_)))
+}
+
+pub(crate) fn override_constructor_return_type(type_: Type) -> Option<Type> {
+    let mut callable = type_.clone().to_callable()?;
+    if !callable.ret.is_none() {
+        return None;
+    }
+
+    let mut should_override = false;
+    if let Params::List(ref params_list) = callable.params
+        && let Some(Param::Pos(name, self_type, _) | Param::PosOnly(Some(name), self_type, _)) =
+            params_list.items().first()
+        && (name.as_str() == "self" || name.as_str() == "cls")
+    {
+        callable.ret = self_type.clone();
+        should_override = true;
+    }
+
+    if should_override {
+        Some(Type::Callable(Box::new(callable)))
+    } else {
+        None
+    }
+}
+
 /// The currently active argument in a function call for signature help.
 #[derive(Debug)]
 pub(crate) enum ActiveArgument {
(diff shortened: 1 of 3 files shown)


pyrefly/lib/lsp/wasm/signature_help.rs
```diff
@@ -280,11 +280,46 @@ impl Transaction<'_> {
...
+            && let Some(mut callable) = type_.clone().to_callable()
+            && callable.ret.is_none()
+        {
+            // Check if first parameter is self or cls
+            let should_override = if let Params::List(ref params_list) = callable.params {
+                if let Some(
+                    Param::Pos(name, self_type, _) | Param::PosOnly(Some(name), self_type, _),
+                ) = params_list.items().first()
+                {
+                    if name.as_str() == "self" || name.as_str() == "cls" {
```
yangdanny97: I think the should_override should be inititalized to false and set to true if all the conditions succeed, would reduce the size of these conditionals

pyrefly/lib/lsp/wasm/signature_help.rs
```diff
@@ -335,6 +370,38 @@ impl Transaction<'_> {
...
+                let is_constructor_call = self
+                    .get_module_info(handle)
+                    .map(|module| {
+                        let contents = module.contents();
+                        let start = callee_range.start().to_usize();
+                        let end = callee_range.end().to_usize();
+                        contents
+                            .get(start..end)
+                            .map(|callee_text| {
+                                if callee_text.ends_with(".__init__") {
```
yangdanny97: this logic appears duplicated w/ the hover code - maybe we can add a util function to share it?

I also don't really like that we're doing string slicing here - is there a way to do it by detecting the CallTarget? If it's a call to a class name it would show up as CallTarget::Class, directly calling `__init__` would not

https://github.com/facebook/pyrefly/blob/8052ae3f798f63ac5c8cd834622ab326027c73ba/pyrefly/lib/alt/call.rs#L85

pyrefly/lib/test/lsp/hover.rs
```diff
@@ -904,6 +905,26 @@ from mymod.submod.deep import Bar
...
+    let code = r#"
+class Person:
+    def __init__(self, name: str, age: int) -> None: ...
+
+Person()
+#^
+"#;
+    let report = get_batched_lsp_operations_report_allow_error(&[("main", code)], get_test_report);
+    assert!(
+        report.contains("-> Person"),
```
yangdanny97: can we assert on the whole string in these tests

yangdanny97: I'm curious if we should do `def Person(...) -> Person` here, it might give it better highlighting properties

yangdanny97:
cc @rchen152 for second review since she wrote the original task

sgavriil01:
Thanks for the review! I've addressed all the feedback:

1. **Duplicated logic** - Extracted `is_constructor_call()` utility function shared between `signature_help.rs` and `hover.rs`
2. **String slicing** - Replaced with type-based detection using `Type::ClassDef(_)` and `Type::Type(ClassType)`  
3. **`should_override` initialization** - Now initialized to `false` and set to `true` when conditions match, simplifying the conditional
4. **Test assertions** - Updated to assert on complete signature strings

Regarding the `def Person(...) -> Person` suggestion: I tested it, but Pyrefly resolves it as `Unknown` since no class definition exists. The class syntax with `__init__` is the correct pattern for this feature. Supporting the `def Person(...)` syntax would require additional changes to Pyrefly's type resolution logic.

stroxler (approved):

yangdanny97:
thanks @sgavriil01

Issue #2136, opened 2026-01-16
Bad Type Variable resolution with outer context (dict.get)


randolf-scholz:
### Describe the Bug

In the example below, I create a `dict`-like class

```python
from typing import Any, reveal_type

class Map[K, V]:
    def set(self, key: K, value: V) -> None: ...
    def get[T=None](self, key: Any, default: T = None, /) -> V | T: ...

d_any: Map[str, Any] = Map()

# as expected, these are both `Any | None`
reveal_type(d_any.get("key"))  # Any | None  ✅️
reveal_type(d_any.get("key", None))  # Any | None  ✅️
# outer context should not overrule the argument type!
result: str = reveal_type(d_any.get("key", None))  # Any | str ❌️
```

## Expectation

`d_any.get("key", None)` produces `Any | None`, and we have `str` as outer context. This gives an unsolvable situation for the type variable `T` as we would need both `T <: str` and `T :> None`. It seems `pyrefly` gives priority to the outer constraint `T <: str` here, which seems misguided and leads to an illogical error message:

> Argument `None` is not assignable to parameter `default` with type `str` in function `Map.get`

[...]

PR #2146, opened 2026-01-18
Fix outer context hints overriding argument types in generic calls (#2136)

jackulau:
# Summary

Added selective hint fallback for generic function calls
Added `has_error_kind()` method to `ErrorCollector` for checking specific error types
Preserves type variable bound violations while fixing hint-induced argument errors
# Problem

When calling generic functions with an outer type hint from an assignment context, the hint was incorrectly constraining type variable resolution:

```python
d_any: Map[str, Any] = Map()
result: str = d_any.get("key", None)  # Was: error on argument type
                                      # Should: error on assignment
```

The type checker was using the `str` hint to solve `T`, then erroring that `None` doesn't match `str`, instead of correctly reporting that `Any | None` is not assignable to `str`.

# Solution

When a generic function call with a hint produces errors:

1. Check if errors are `BadSpecialization` (type variable bound violations)
2. 2. If NOT bound violations → retry without hint (hint-induced errors)
3. 3. If bound violations → keep the errors (legitimate type checking)
This ensures:
- Outer context hints don't incorrectly constrain argument types
- - Type variable bound checking still works correctly
# Test plan

- Added `test_call_hint_does_not_override_arg` for the reported issue
- - Verified `test_nested_call_preserves_bound` still passes
- - All 3542 tests pass

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -867,54 +867,253 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         metadata,
                     },
                 ),
-            ) => self.callable_infer(
-                signature,
-                Some(&metadata.kind),
-                tparams.as_deref(),
-                Some(obj),
-                args,
-                keywords,
-                range,
-                errors,
-                errors,
-                context,
-                hint,
-                ctor_targs,
-            ),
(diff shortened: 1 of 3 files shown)


rchen152:
@jackulau Apologies that this slipped through the cracks! I'll try to get to this as soon as I can, although my review queue is a little backed up right now.

yangdanny97:
I'm going to re-trigger mypy primer

PR #2157, opened 2026-01-19
fix Error when class patterns are used as enums in pattern match #2153

### LINKED ISSUE #2153: Error when class patterns are used as enums in pattern match
### Describe the Bug

For example, error when `case Color.RED():` is used, it should be `case Color.RED:`

This was originally implemented as part of https://github.com/facebook/pyrefly/pull/1833 but we didn't merge that change since it felt separate from the core change in the PR.

We would be open to accepting a PR that implements this change separately.

The error should be something like `invalid-pattern`

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2153

Added a pattern-specific class-pattern narrow op that validates the class target and emits the new invalid-pattern error when it isn’t a class object (e.g., Color.RED())

# Test Plan

a regression test and docs update.

### DIFF
--- a/pyrefly/lib/test/pattern_match.rs
+++ b/pyrefly/lib/test/pattern_match.rs
@@ -221,7 +221,7 @@ class Color(Enum):
     RED = "red"
 
 def describe(color: Color) -> None:
-    match color:
+    match color:  # E: Match on `Color` is not exhaustive
         case Color.RED():  # E: Expected class object, got `Literal[Color.RED]`
             pass
 "#,


pyrefly/lib/alt/narrow.rs
```diff
@@ -752,6 +753,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 let right = self.expr_infer(v, errors);
                 self.narrow_isinstance(ty, &right)
             }
+            AtomicNarrowOp::IsInstancePattern(v) => {
```
yangdanny97, line 756: does this need to be a separate narrow op?

PR #2158, opened 2026-01-19
fix Missing support for typing.Literal in pydantic #2156

### LINKED ISSUE #2156: Missing support for `typing.Literal` in pydantic
### Describe the Bug

```python
from typing import Literal

class MyPydantic(pydantic.BaseModel):
    value: Literal["MyLiteral"]


MyPydantic(value=2) # Does not error (incorrectly)


@dataclass
class MyPythonDataclass:
    value: Literal["MyLiteral"]


MyPythonDataclass(value=2) # Errors (correctly)
```

```bash
➜ uvx pyrefly --version
pyrefly 0.48.2

➜  uvx pyrefly check .
ERROR Argument `Literal[2]` is not assignable to parameter `value` with type `Literal['MyLiteral']` in function `MyPythonDataclass.__init__` [bad-argument-type]
  --> iterpy/test_arr.py:40:25
   |
40 | MyPythonDataclass(value=2)
   |                         ^
   |
 INFO 1 error (8 suppressed)
```

Does not error on the pydantic dataclass, which it should. 

Again, thanks for awesome software and a nice vision. Would love to see this resolved!

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2156

Adjusted Pydantic lax conversion to preserve literal types so Literal[...] fields stay strict even when lax mode is enabled. This prevents converting Literal into Any, which is why pydantic dataclasses were silently accepting bad values.

# Test Plan

update test, add a test

### DIFF
--- a/pyrefly/lib/alt/class/pydantic_lax.rs
+++ b/pyrefly/lib/alt/class/pydantic_lax.rs
@@ -167,7 +167,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
 
     fn expand_type_for_lax_mode(&self, ty: &Type) -> Type {
         match ty {
-            Type::None => ty.clone(),
+            Type::None | Type::Literal(_) | Type::LiteralString(_) => ty.clone(),
             Type::Type(box inner) => Type::Type(Box::new(self.expand_type_for_lax_mode(inner))),
             // Tuple types: convert to Iterable[T] where T is a union of expanded element types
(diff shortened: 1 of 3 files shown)


grievejia:
Sorry @migeed-z seems to have fixed it before this PR gets merged :( 

We might be able to avoid this kind of "race condition" if you comment on the issue you want to work on and let us assign them to you before submitting PRs.

PR #2191, opened 2026-01-22
fix provide auto completion when match Literal #2178

### LINKED ISSUE #2178: provide auto completion when match Literal
### Describe the Bug

```py
def f(x : Literal["xxxx", "yyyyy"]):
    match x:
        case "x"
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2178

Added match-pattern literal completions by detecting `PatternMatchValue`/`PatternMatchSingleton` at the cursor and using the match subject’s inferred type to surface Literal values.

# Test Plan

A regression test for string literal matching.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2591,7 +2591,7 @@ impl<'a> Transaction<'a> {
     /// infer literals from. Instead, we look for a match value/singleton
     /// pattern at the cursor and pull the `match` subject's type to surface
     /// its Literal members.
-    fn add_match_literal_completions(
+    pub(crate) fn add_match_literal_completions(
         &self,
         handle: &Handle,
         covering_nodes: &[AnyNodeRef],
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -599,6 +599,12 @@ impl Transaction<'_> {
                     let in_string_literal = nodes
                         .iter()
                         .any(|node| matches!(node, AnyNodeRef::ExprStringLiteral(_)));
+                    self.add_match_literal_completions(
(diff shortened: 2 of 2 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -2784,6 +2784,33 @@ impl<'a> Transaction<'a> {
         }
     }
 
+    fn add_match_literal_completions(
```
jvansch1: I think it would be valuable to add a docstring here. It would be especially useful to add an explanation as to why this functionality is needed for match literal and why existing functionality in Pyrefly does not cover this case.

pyrefly/lib/state/lsp.rs
```diff
@@ -2784,6 +2784,33 @@ impl<'a> Transaction<'a> {
...
+        let is_match_value_pattern = covering_nodes.iter().any(|node| {
+            matches!(
+                node,
+                AnyNodeRef::PatternMatchValue(_) | AnyNodeRef::PatternMatchSingleton(_)
+            )
+        });
+        if !is_match_value_pattern {
+            return;
+        }
+        let Some(subject) = covering_nodes.iter().find_map(|node| match node {
```
jvansch1: I wonder if we can combine this block and the block in line 2794 into one? We are doing 2 iterations through the covering node which is not a huge deal but always good to avoid unnecessary work if we can.

jvansch1:
@asukaminato0721 Looks like you have some conflicts here.

asukaminato0721:
@jvansch1 solved

jvansch1:
Hey @asukaminato0721 Looks like merge conflicts again. Once those are resolved I will get a final review of this PR completed.

kinto0 (approved):

PR #2193, opened 2026-01-22
impl Autocomplete Abstract, Protocol and method override #1589

asukaminato0721:
# Summary

Fixes part of #1589 #2207

https://github.com/microsoft/pylance-release/issues/146

Implemented an “Implement abstract members” quick fix that derives missing abstract members via the solver, copies signatures/decorators/docstrings, and inserts stubs into the selected class, then wired it into LSP code actions.

# Test Plan

added a test.

### DIFF
--- a/pyrefly/lib/test/lsp/code_actions.rs
+++ b/pyrefly/lib/test/lsp/code_actions.rs
@@ -614,6 +614,27 @@ fn compute_implement_abstract_actions(
     (module_info, edit_sets, titles)
 }
 
+fn compute_implement_abstract_actions_allow_errors(
+    code: &str,
+) -> (
+    ModuleInfo,
+    Vec<Vec<(Module, TextRange, String)>>,
+    Vec<String>,
+) {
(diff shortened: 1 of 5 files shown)


pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -895,7 +939,7 @@ class Base:
         pass
     # MOVE-END
 
-class Child(Base):
+class Child(Base, ABC):
```
asukaminato0721, line 3371: test used mk_multi_file_state_assert_no_errors, and a concrete Child(Base) that doesn’t implement abstract members would raise an ImplicitAbstractClass error. Making it inherit ABC marks it explicitly abstract, so the test setup stays error‑free.

mohaidoss:
Up we need this <3

asukaminato0721:
cc @jvansch1

asukaminato0721:
cc @jvansch1

PR #2195, opened 2026-01-22
impl Convert import to relative path / absolute path (and Convert all... variants) #1589

asukaminato0721:
# Summary

Fixes part of #1589

Added convert‑import code actions (single + all) for relative/absolute paths, wired them into LSP.

# Test Plan

Added tests for both from and import forms.

### DIFF
--- a/crates/pyrefly_python/src/module_name.rs
+++ b/crates/pyrefly_python/src/module_name.rs
@@ -399,7 +399,13 @@ impl ModuleName {
                 }
             };
         }
-        Self::from_relative_path_components(components).ok()
+        let module = Self::from_relative_path_components(components).ok()?;
+        // Joining empty components renders one fewer dot, so restore it when there is no suffix.
+        if module.as_str().chars().all(|c| c == '.') {
+            Some(Self::from_string(format!(".{}", module.as_str())))
+        } else {
+            Some(module)
+        }
     }
 
     pub fn append(self, name: &Name) -> Self {
@@ -674,6 +680,9 @@ mod tests {
         assert_module_name("bar.py", "foo/baz.py", ".foo.baz");
         assert_module_name("foo/bar.py", "baz.py", "..baz");
         assert_module_name("foo/bar/boz.py", "baz.py", "...baz");
+        assert_module_name("foo/bar.py", "foo/__init__.py", ".");
+        assert_module_name("foo/bar/baz.py", "foo/__init__.py", "..");
+        assert_module_name("foo/bar.py", "foo/sub/__init__.py", ".sub");
     }
 
     #[test]
--- a/pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
@@ -1,0 +1,364 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+use dupe::Dupe;
+use lsp_types::CodeActionKind;
+use pyrefly_build::handle::Handle;
+use pyrefly_python::ast::Ast;
+use pyrefly_python::module::Module;
+use pyrefly_python::module_name::ModuleName;
+use ruff_python_ast::Alias;
+use ruff_python_ast::AnyNodeRef;
+use ruff_python_ast::ModModule;
+use ruff_python_ast::Stmt;
+use ruff_python_ast::StmtImport;
+use ruff_python_ast::StmtImportFrom;
+use ruff_python_ast::visitor::Visitor;
+use ruff_text_size::Ranged;
+use ruff_text_size::TextRange;
+use ruff_text_size::TextSize;
+
+use super::extract_shared::line_indent_and_start;
+use super::extract_shared::selection_anchor;
+use crate::state::ide::handle_require_absolute_import;
+use crate::state::lsp::LocalRefactorCodeAction;
+use crate::state::lsp::Transaction;
+
+#[derive(Clone, Copy, Debug)]
+enum ConvertTarget {
+    Relative,
+    Absolute,
+}
+
+#[derive(Clone, Copy, Debug)]
+enum ImportStmtRef<'a> {
+    Import(&'a StmtImport),
+    ImportFrom(&'a StmtImportFrom),
+}
+
+pub(crate) fn convert_import_code_actions(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    selection: TextRange,
+) -> Option<Vec<LocalRefactorCodeAction>> {
+    let module_info = transaction.get_module_info(handle)?;
+    let ast = transaction.get_ast(handle)?;
+    let source = module_info.contents();
+    let selection_point = selection_anchor(source, selection);
+    let import_stmt = find_import_stmt(ast.as_ref(), selection_point)?;
+
+    let mut actions = Vec::new();
+    if let Some(action) = convert_single_import_action(
+        transaction,
+        handle,
+        &module_info,
+        source,
+        import_stmt,
+        ConvertTarget::Relative,
+    ) {
+        actions.push(action);
+    }
+    if let Some(action) = convert_single_import_action(
+        transaction,
+        handle,
+        &module_info,
+        source,
+        import_stmt,
+        ConvertTarget::Absolute,
+    ) {
+        actions.push(action);
+    }
+    if let Some(action) = convert_all_imports_action(
+        transaction,
+        handle,
+        &module_info,
+        source,
+        ast.as_ref(),
+        ConvertTarget::Relative,
+    ) {
+        actions.push(action);
+    }
+    if let Some(action) = convert_all_imports_action(
+        transaction,
+        handle,
+        &module_info,
+        source,
+        ast.as_ref(),
+        ConvertTarget::Absolute,
+    ) {
+        actions.push(action);
+    }
+
+    if actions.is_empty() {
+        None
+    } else {
+        actions.sort_by(|a, b| a.title.cmp(&b.title));
+        Some(actions)
+    }
+}
+
+fn convert_single_import_action(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    module_info: &Module,
+    source: &str,
+    stmt: ImportStmtRef<'_>,
+    target: ConvertTarget,
+) -> Option<LocalRefactorCodeAction> {
+    let edit = convert_import_stmt(transaction, handle, source, stmt, target)?;
+    let title = match target {
+        ConvertTarget::Relative => "Convert import to relative path",
+        ConvertTarget::Absolute => "Convert import to absolute path",
+    };
+    Some(LocalRefactorCodeAction {
+        title: title.to_owned(),
+        edits: vec![(module_info.dupe(), edit.0, edit.1)],
+        kind: CodeActionKind::REFACTOR_REWRITE,
+    })
+}
+
+fn convert_all_imports_action(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    module_info: &Module,
+    source: &str,
+    ast: &ModModule,
+    target: ConvertTarget,
+) -> Option<LocalRefactorCodeAction> {
+    let mut edits = Vec::new();
+    for stmt in collect_import_stmts(ast) {
+        if let Some(edit) = convert_import_stmt(transaction, handle, source, stmt, target) {
+            edits.push((module_info.dupe(), edit.0, edit.1));
+        }
+    }
+    if edits.is_empty() {
+        return None;
+    }
+    let title = match target {
+        ConvertTarget::Relative => "Convert all imports to relative path",
+        ConvertTarget::Absolute => "Convert all imports to absolute path",
+    };
+    Some(LocalRefactorCodeAction {
+        title: title.to_owned(),
+        edits,
+        kind: CodeActionKind::REFACTOR_REWRITE,
+    })
+}
+
+fn convert_import_stmt(
+    transaction: &Transaction<'_>,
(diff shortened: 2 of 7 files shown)


asukaminato0721:
The test is missing init parameter, which is in other prs.

jvansch1:
@asukaminato0721 Once merge conflicts are resolved on this PR I can start review.

asukaminato0721:
cc @jvansch1

asukaminato0721:
cc @jvansch1

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,364 @@
...
+    if let Some(action) = convert_all_imports_action(
+        transaction,
+        handle,
+        &module_info,
+        source,
+        ast.as_ref(),
+        ConvertTarget::Absolute,
+    ) {
+        actions.push(action);
+    }
```
kinto0: nit: can this be a `map` over a list of these possible converttarget so we don't duplicate the code so much?

kinto0 (changes requested):

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,364 @@
...
+    impl<'a> Visitor<'a> for ImportCollector<'a> {
+        fn visit_stmt(&mut self, stmt: &'a Stmt) {
+            match stmt {
+                Stmt::Import(import) => self.imports.push(ImportStmtRef::Import(import)),
+                Stmt::ImportFrom(import_from) => {
+                    self.imports.push(ImportStmtRef::ImportFrom(import_from));
+                }
+                _ => {}
+            }
+            ruff_python_ast::visitor::walk_stmt(self, stmt);
```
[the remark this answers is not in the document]
kinto0: what do you think?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+}
+
+fn split_relative_module(relative_module: &str, absolute_module: &str) -> Option<(String, String)> {
+    if relative_module == "." {
+        let leaf = absolute_module
+            .rsplit_once('.')
+            .map_or(absolute_module, |(_, leaf)| leaf);
+        return Some((".".to_owned(), leaf.to_owned()));
+    }
```
kinto0: AI comment: wrong dot count for multi-level relative import X. In split_relative_module, rsplit_once('.') miscounts leading dots when the relative module is N≥2 dots directly followed by a leaf:

  import baz            # in foo/bar.py, baz.py at repo root
    relative_module_name_between → "..baz"   (see module_name.rs test line 666)
    split_relative_module("..baz", "baz")
      "..baz".rsplit_once('.') → (".", "baz")   # base=".", non-empty, kept
      → "from . import baz"                       # ❌ should be "from .. import baz"
  from . import baz resolves to foo.baz, which doesn't exist — a semantically broken edit. ...leaf is likewise off by one dot. The .foo.baz case works only because the intermediate package absorbs the miscount. Suggested fix — count leading dots explicitly instead of rsplit_once:
  let dots = relative_module.len() - relative_module.trim_start_matches('.').len();
  let rest = &relative_module[dots..];
  let (pkg, leaf) = rest.rsplit_once('.').unwrap_or(("", rest));
  let base = format!("{}{}", ".".repeat(dots), pkg); // ".." + "" = "..", "." + "foo" = ".foo"

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,364 @@
...
+    if let Some((base, leaf)) = relative_module.rsplit_once('.') {
+        let base = if base.is_empty() { "." } else { base };
+        return Some((base.to_owned(), leaf.to_owned()));
+    }
+    Some((".".to_owned(), relative_module.to_owned()))
```
kinto0: is this unreachable? can we verify that with the type system or make unreachable!()

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,364 @@
...
+    } else {
+        default_name.to_owned()
+    }
+}
+
+fn relative_module_string(from: &Handle, to: &Handle) -> Option<String> {
+    let module =
+        ModuleName::relative_module_name_between(from.path().as_path(), to.path().as_path())?;
+    let raw = module.as_str();
+    if raw.is_empty() {
```
kinto0: is this reachable? can we avoid it?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -4987,3 +5010,105 @@ class Derived(Base):
...
+        .position(|title| title == "Convert import to relative path")
+        .unwrap();
+    let updated = apply_refactor_edits_for_module(&module_info, &actions[idx]);
+    let expected = r#"
+# IMPORT-START
+from . import b as mod_b
+# IMPORT-END
+"#;
+    assert_eq!(expected.trim(), updated.trim());
+}
```
kinto0, line 5954: Untested edge cases (all reachable through the code paths):
  - Multi-level relative (.., ...) — the buggy case above; not covered by any test.
  - from . import x (module=None, level>0) → absolute. convert_from_import_to_absolute handles it (new_maybe_relative(is_init, level, None) → from pkg import x) but nothing exercises it.
  - handle_require_absolute_import = true (stdlib / search-path / site-packages) should suppress the relative action — logic looks correct, untested.
  - Grouping: import a as x, b as y → merged from . import a as x, b as y. Untested.
  - Multi-line parenthesized from x import (A, B) collapses to a single line (and any comments inside the parens are dropped, since render_imported_names only joins names). Trailing end-of-line comments are safe (outside stmt.range()), but the in-parens case is lossy and untested.
  - is_init files: relative levels differ for __init__.py; no __init__ test.

kinto0 (changes requested):

crates/pyrefly_python/src/module_name.rs
```diff
@@ -389,7 +389,12 @@ impl ModuleName {
                 }
             };
         }
-        Self::from_relative_path_components(components).ok()
+        let module = Self::from_relative_path_components(components).ok()?;
+        if to.file_stem()?.to_str()? == dunder::INIT {
+            Some(Self::from_string(format!(".{}", module.as_str())))
+        } else {
+            Some(module)
```
kinto0, line 407: Blocker: the module_name.rs change is wrong for descendant packages

  ModuleName::relative_module_name_between (crates/pyrefly_python/src/module_name.rs:376) is shared — its only other caller is the auto-import path at pyrefly/lib/state/ide.rs:350. The PR appends a dot
  whenever to is an __init__.py. That over-applies.

  The real defect is narrower: from_parts joins with . between components, so 1 + levels_up empty components render as levels_up dots. That only loses a dot when the part below the dots is empty:

  ┌────────────────┬─────────────────────┬─────────┬───────────┬──────────┐
  │      from      │         to          │ correct │ before PR │ after PR │
  ├────────────────┼─────────────────────┼─────────┼───────────┼──────────┤
  │ foo/bar.py     │ foo/__init__.py     │ .       │ ""        │ . ✅     │
  ├────────────────┼─────────────────────┼─────────┼───────────┼──────────┤
  │ foo/bar/baz.py │ foo/__init__.py     │ ..      │ .         │ .. ✅    │
  ├────────────────┼─────────────────────┼─────────┼───────────┼──────────┤
  │ foo/bar.py     │ foo/sub/__init__.py │ .sub    │ .sub      │ ..sub ❌ │
  └────────────────┴─────────────────────┴─────────┴───────────┴──────────┘

  ..sub from foo/bar.py resolves to a sibling of foo. This hits both the new code action (import foo.sub as s → from .. import sub as s) and the pre-existing relative auto-import. Condition on the collapse,
  not on the filename:

  let module = Self::from_relative_path_components(components).ok()?;
  // `join` puts separators *between* components, so an all-dots result is one dot short.
  if module.as_str().chars().all(|c| c == '.') {
      Some(Self::from_string(format!(".{}", module.as_str())))
  } else {
      Some(module)
  }

  Add assert_module_name("foo/bar.py", "foo/sub/__init__.py", ".sub") to test_relative_module_name_between (module_name.rs:657) — no existing case covers a target __init__.py below the source.

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+enum ImportStmtRef<'a> {
+    Import(&'a StmtImport),
+    ImportFrom(&'a StmtImportFrom),
+}
+
+pub(crate) fn convert_import_code_actions(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    selection: TextRange,
+) -> Option<Vec<LocalRefactorCodeAction>> {
```
kinto0: nit: should this be `option<vec1` because an empty list doesn't make sense?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    if actions.is_empty() {
+        None
+    } else {
+        actions.sort_by(|a, b| a.title.cmp(&b.title));
+        Some(actions)
+    }
+}
+
+fn convert_single_import_action(
```
kinto0: nit: what's the contract here? when will this return empty?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+        ConvertTarget::Absolute => "Convert all imports to absolute path",
+    };
+    Some(LocalRefactorCodeAction {
+        title: title.to_owned(),
+        edits,
+        kind: CodeActionKind::REFACTOR_REWRITE,
+    })
+}
+
+fn convert_import_stmt(
```
kinto0: what's the contract here? when will this return empty?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+    handle: &Handle,
+    source: &str,
+    stmt: ImportStmtRef<'_>,
+    target: ConvertTarget,
+) -> Option<(TextRange, String)> {
+    let stmt_range = match stmt {
+        ImportStmtRef::Import(import) => import.range(),
+        ImportStmtRef::ImportFrom(import_from) => import_from.range(),
+    };
+    if source
```
kinto0: is this the best way to do this? do we do something similar in other code actions? why / why not?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+        _ => return None,
+    };
+    Some((stmt_range, new_text))
+}
+
+fn convert_import_to_relative(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    import: &StmtImport,
+) -> Option<Vec<String>> {
```
kinto0: signature is not clear from looking at this

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+    Some(
+        grouped
+            .into_iter()
+            .map(|(base, names)| format!("from {base} import {}", names.join(", ")))
+            .collect(),
+    )
+}
+
+fn convert_from_import_to_relative(
```
kinto0, line 200: can we share more logic with the function above?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
```diff
@@ -0,0 +1,344 @@
...
+    handle: &Handle,
+    import_from: &StmtImportFrom,
+) -> Option<String> {
+    let module = handle.module().new_maybe_relative(
+        handle.path().is_init(),
+        import_from.level,
+        import_from.module.as_ref().map(|module| &module.id),
+    )?;
+    let module_str = module.as_str();
+    if module_str.is_empty() {
```
kinto0, line 223: why this guard? when would this happen?

kinto0 (changes requested):
some concerns upon closer look. could you audit the high-level design and make sure API boundaries are clean?

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
asukaminato0721: fixed

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
asukaminato0721: done

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
asukaminato0721: It returns None when the selected import cannot be safely converted to the requested target.

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
asukaminato0721: renamed the function to convert_plain_import_to_relative

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
asukaminato0721: `from . import x`

pyrefly/lib/state/lsp/quick_fixes/convert_import.rs
asukaminato0721: when no import in the file has an applicable.

PR #2201, opened 2026-01-22
impl pytest add type

asukaminato0721:
# Summary

Fixes part of #1589

The code action now adds type annotations for fixture parameters only in test functions, resolving fixtures first in the current module then upward to conftest files (closest wins).

Haven't implemented full pytest fixture resolution outside conftest but the current approach covers typical use.

But problem: should LSP just infer the version as inlay hint, then click to insert?

relevant https://github.com/DetachHead/basedpyright/issues/311

# Test Plan

pytest_fixture_type_annotation_code_actions

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
@@ -9,7 +9,9 @@ use std::collections::HashMap;
 use std::collections::HashSet;
 
 use dupe::Dupe;
+use lsp_types::CodeActionKind;
 use pyrefly_build::handle::Handle;
+use pyrefly_python::module::Module;
 use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::module_name::ModuleNameWithKind;
 use pyrefly_python::module_path::ModulePath;
@@ -108,12 +110,19 @@ fn collect_fixture_functions<'a>(stmts: &'a [Stmt], out: &mut Vec<&'a StmtFuncti
     }
 }
 
-fn should_skip_annotation(rendered: &str, ty: &Type) -> bool {
-    ty.is_any()
+fn render_annotation(ty: &Type) -> Option<String> {
+    let rendered = ty.as_lsp_string(LspDisplayMode::SignatureHelp);
+    if ty.is_any()
+        || ty.is_never()
         || rendered.contains("Any")
         || rendered.contains("Unknown")
         || rendered.contains("Never")
         || rendered.contains('@')
+    {
+        None
+    } else {
+        Some(rendered)
+    }
 }
 
 fn is_test_name(name: &Name) -> bool {
@@ -124,7 +133,7 @@ fn is_test_function(func: &StmtFunctionDef, class_name: Option<&Name>) -> bool {
     if !is_test_name(&func.name.id) {
         return false;
     }
-    class_name.map_or(true, |name| name.as_str().starts_with("Test"))
+    class_name.is_none_or(|name| name.as_str().starts_with("Test"))
 }
 
 fn collect_test_functions<'a>(
@@ -148,18 +157,16 @@ fn collect_test_functions<'a>(
 }
 
 #[derive(Debug)]
-struct FixtureAnnotationEdit {
-    def_range: TextRange,
+struct AnnotationEdit {
     insert_range: TextRange,
     insert_text: String,
     import_edits: Vec<(TextSize, String)>,
 }
 
 #[derive(Debug)]
-struct FixtureParamAnnotationEdit {
-    insert_range: TextRange,
-    insert_text: String,
-    import_edits: Vec<(TextSize, String)>,
+struct ScopedAnnotationEdit {
+    scope_range: TextRange,
+    annotation: AnnotationEdit,
 }
 
 fn fixture_return_type(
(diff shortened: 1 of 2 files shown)


kinto0 (changes requested):
PR is quite old. sorry about the delayed review. happy to review it if you're still interested in merging - it just will need some merge conflicts resolved

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -135,8 +135,14 @@ fn compute_extract_actions(
     Vec<Vec<(Module, TextRange, String)>>,
     Vec<String>,
 ) {
-    let (handles, state) =
-        mk_multi_file_state_assert_no_errors(&[("main", code)], Require::Everything);
+    let pytest_stub = r#"
```
kinto0: nit: could we instead do this for only the pytest tests?

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
```diff
@@ -0,0 +1,542 @@
...
+            return;
+        }
+        if seen_imports.insert(insert_text.clone()) {
+            import_edits.push((position, insert_text));
+        }
+    });
+    import_edits
+}
+
+/// Builds code actions that add inferred return annotations to pytest fixtures.
```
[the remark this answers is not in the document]
kinto0: even better: can we split these responsibilities into two functions?

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
```diff
@@ -0,0 +1,546 @@
...
+            Stmt::FunctionDef(func) => out.push(func),
+            Stmt::ClassDef(class_def) => collect_fixture_functions(&class_def.body, out),
+            _ => {}
+        }
+    }
+}
+
+fn should_skip_annotation(rendered: &str, ty: &Type) -> bool {
+    ty.is_any()
+        || rendered.contains("Any")
```
kinto0, line 117: hmm I wonder if we have a helper function here that does something similar in a better way? I don't think we can land it like this

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
```diff
@@ -0,0 +1,546 @@
...
+                };
+                let rendered = ty.as_lsp_string(LspDisplayMode::SignatureHelp);
+                if should_skip_annotation(&rendered, &ty) {
+                    continue;
+                }
+                if !seen_params.insert(identifier.range()) {
+                    continue;
+                }
+                edits.push((
+                    module.dupe(),
```
kinto0: should we use FixtureParamAnnotationEdit here too?

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
```diff
@@ -0,0 +1,546 @@
...
+    let is_memory = matches!(module_path.details(), ModulePathDetails::Memory(_));
+    let mut conftest_paths = Vec::new();
+    loop {
+        let conftest_pyi = dir.join("conftest.pyi");
+        let conftest_py = dir.join("conftest.py");
+        if is_memory {
+            conftest_paths.push(ModulePath::memory(conftest_pyi.clone()));
+            conftest_paths.push(ModulePath::memory(conftest_py.clone()));
+        } else {
+            if conftest_pyi.exists() {
```
kinto0: can we use state to understand if this exists instead of a filesystem operation?

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
```diff
@@ -0,0 +1,546 @@
...
+            edits.push((
+                module.dupe(),
+                TextRange::at(*position, TextSize::new(0)),
+                text.clone(),
+            ));
+        }
+        actions.push(LocalRefactorCodeAction {
+            title: "Add pytest fixture type annotation".to_owned(),
+            edits,
+            kind: lsp_types::CodeActionKind::QUICKFIX,
```
kinto0: probably REFACTOR is a better kind if that exists?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -4094,3 +4100,77 @@ class Child(Base):
...
+
+@pytest.fixture
+def user():
+    return "alice"
+
+def test_one(answer: int, user: str):
+    print(answer, user)
+"#;
+    assert_eq!(expected.trim(), updated_all.trim());
+}
```
kinto0, line 4195: there's no test for the "Add pytest fixture type annotation" action (adding -> str to the user() fixture definition)

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -4094,3 +4100,77 @@ class Child(Base):
 "#;
     assert!(apply_first_safe_delete_action(code).is_none());
 }
+
+#[test]
+fn pytest_fixture_type_annotation_code_actions() {
+    let conftest = r#"
+import pytest  # type: ignore
```
kinto0, line 4101: does it work if they import pytest as something_else?

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
```diff
@@ -0,0 +1,546 @@
...
+}
+
+impl PytestAliases {
+    fn collect(ast: &ModModule) -> Self {
+        let mut aliases = Self::default();
+        for stmt in &ast.body {
+            match stmt {
+                Stmt::Import(import_stmt) => {
+                    for alias in &import_stmt.names {
+                        if alias.name.id.as_str() == "pytest" {
```
kinto0, line 52: i wonder if there's a better way we can do this with fully qualified paths?

kinto0 (changes requested):
doesn't look like you've replied to / resolved the comments? if there's a reason you can't do a suggestion, can you reply to the suggestion?

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
asukaminato0721: I didn’t switch this to fully-qualified resolution in this diff. The closest existing machinery is much heavier than what we need here, so I kept the alias-based approach and added alias coverage.

pyrefly/lib/test/lsp/code_actions.rs
asukaminato0721: added test for `import pytest as pt  # type: ignore`

pyrefly/lib/state/lsp/quick_fixes/pytest_fixture.rs
asukaminato0721: I collapsed the repeated filtering into `render_annotation`

PR #2203, opened 2026-01-22
add tests for incremental on field keys

kinto0:
Summary:
# This stack

This stack modifies the optimization for incremental updates found in the stack containing D90647539. 

Original optimization: Use import statements alone to understand which imports a module uses

Problem: `import foo` will always invalidate `foo`

This stack will modify this optimization to instead track these dependencies based on usages during solve. Every type depended on from a module will be referenced in solve and we only will need to invalidate an entire module when there is an import *.

# This diff
adds tests for field keys incrementality. similar to test_enum_member_change_propagates, when a field or part of a class gets added, removing a type error, we need to ensure dependents get updated. this is necessary because without D91173578, the semantic invalidation will not work

Reviewed By: yangdanny97

Differential Revision: D91172436

### DIFF
--- a/pyrefly/lib/test/incremental.rs
+++ b/pyrefly/lib/test/incremental.rs
@@ -1272,3 +1272,197 @@ fn test_mixed_import_failed_export_invalidated() {
         "Expected error after foo exports y"
     );
 }
+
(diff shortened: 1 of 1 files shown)


yangdanny97 (approved):

PR #2217, opened 2026-01-24
make --remove-unused-ignores a no-op for files with no unused ignores

### LINKED ISSUE #2185: Feature: Don't unnecessarily rewrite files with --remove-unused-ignores
### Describe the Bug

I'm looking to integrate pyrefly into [treefmt](https://treefmt.com/) (to support my pre-commit hook), but the `--remove-unused-ignores` flag rewrites all files even when there were no unused ignores to remove. This violates treefmt's [Formatter specification rule #2](https://github.com/numtide/treefmt/blob/main/docs/site/reference/formatter-spec.md#2-write-to-changed-files), which requires that correctly-formatted files (in this case, files with no unused ignores to remove) not be written to at all.

### Sandbox Link

_No response_

_No response_

self-sasi:
# Summary

Fixes #2185

This change makes `--remove-unused-ignores` avoid rewriting files when there are no unused ignores to remove.

Files that have no ignore comments, or only have used ignore comments, are now left untouched. This prevents unnecessary file writes and avoids introducing formatting-only changes such as added newlines.

# Test Plan

Added tests to verify no newline characters are appended to input when there are no unused ignores to remove.

### DIFF
--- a/pyrefly/lib/error/suppress.rs
+++ b/pyrefly/lib/error/suppress.rs
@@ -374,7 +374,9 @@ pub fn remove_unused_ignores(loads: &Errors, all: bool) -> usize {
     let mut all_ignores: SmallMap<&PathBuf, &Ignore> = SmallMap::new();
     for (module_path, ignore) in loads.collect_ignores() {
         if let ModulePathDetails::FileSystem(path) = module_path.details() {
-            all_ignores.insert(path, ignore);
+            if !ignore.is_empty() {
+                all_ignores.insert(path, ignore);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


pyrefly/lib/error/suppress.rs
```diff
@@ -374,6 +374,9 @@ pub fn remove_unused_ignores(loads: &Errors, all: bool) -> usize {
     let mut all_ignores: SmallMap<&PathBuf, &Ignore> = SmallMap::new();
     for (module_path, ignore) in loads.collect_ignores() {
         if let ModulePathDetails::FileSystem(path) = module_path.details() {
+            if ignore.is_empty() {
+                continue;
+            }
             all_ignores.insert(path, ignore);
```
gmacon: I might structure this with a negative condition and avoid the `continue`, i.e.

```suggestion
            if !ignore.is_empty() {
               all_ignores.insert(path, ignore);
            }
```

self-sasi: Made the change. Thanks for the feedback!

PR #2226, opened 2026-01-25
Handle `self` type annotation when converting `__init__` as a callable

fangyi-zhou:
# Summary

When a constructor is converted as a callable, we should take care of any type annotation on the `self` when setting the return type of the callable. 

In particular, the spec says that https://typing.python.org/en/latest/spec/constructors.html#init-method

> To enable this, type checkers should allow the `self` parameter to be annotated with a type that influences the resulting type of the constructor call.

This diff uses the `self` parameter type to set the return type of the callable to implement this requirement.

NOTE: In the conformance test, the callbable converted from constructor is passed around an identical function that causes overload handing to be incorrect. Therefore, this fix does not fix the conformance test in full.

Help needed: There's a regression on `dataclasses_transform_converter` conformance test, but I'm unable to make sense of the error. I suspect this is due to some mishandling of generics.

Next step: Handle generics correctly when converting constructor into a callable.

# Test Plan

`test.py`


migeed-z:
I will take a look!

PR #2228, opened 2026-01-25
Only show bug reporting URL in non-debug builds

fangyi-zhou:
# Summary

Conformance test expectation files contains the error reporting URL which deviates between fbcode builds (for Meta environements) and open-source environments. This change suppresses some noise due to environment differences.

# Test Plan

Regenerated conformance test expectation files and observed that the bug reporting URL is removed.


yangdanny97:
given there's only a few cases let's just fix the internal error, ideally they shouldn't show up at all in conformance tests

Issue #1122, opened 2025-09-22
Code completion for overloads are incomplete


fangyi-zhou:
### Describe the Bug

```python
from typing import overload, Any

@overload
def foo(x: int) -> int: ...
@overload
def foo(x: str) -> str: ...

def foo(x: Any) -> Any:
    return x

foo(x=)
```

Python: 3.12

The completion for `x=` should show both `x: int` and `x: str`

<img width="567" height="311" alt="Image" src="https://github.com/user-attachments/assets/e350630f-28ea-4f58-ab62-94cdd3509d06" />

Pyright playground completes both cases

<img width="327" height="303" alt="Image" src="https://github.com/user-attachments/assets/9aa4e956-3bb9-4208-81e0-696b31c0572e" />

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN24AbjEpRcqTABo6AQXSkAOuiUABISLESlmGGDphcuABT5EndAwCUdALQA%2BcwzOEXq9aPGZtu-YZNm4BkprezpAymdXZSwfA2NTWXkQhzlSRCU6TLpKGAYAV0p0OnwlJTiTAF5LEEkQPIZoOBJyRBAAYjoAVQaoCCZ9PPQAYwbcdDhSmL0DShpUBgB9dDyabBF-R2SwoPSirJz8wv0FEAA5FbWIumB8AF8TpRBb2tQRiGEAMWgYCjQsPCIZCeQA&version=3.12

fangyi-zhou:
https://github.com/facebook/pyrefly/blob/4ce17e819521e2933c8930267166e5235e6a1dc3/pyrefly/lib/test/lsp/completion.rs#L1219

This test looks a bit suspecious, should give both results.

https://github.com/facebook/pyrefly/blob/4ce17e819521e2933c8930267166e5235e6a1dc3/pyrefly/lib/test/lsp/completion.rs#L1247

This test looks more suspecious, should give a single result. (also should be correctly indented)

stroxler:
cc @SamChou19815 @kinto0 

Also not I added a new `language-server` tag; I haven't had a good way of sorting IDE-specific issues until now

fangyi-zhou:
Looks like we're committing on the closest overload here (L605)
[...]

PR #2241, opened 2026-01-27
Show completions from all overloads in LSP

fangyi-zhou:
## Summary

Resolves https://github.com/facebook/pyrefly/issues/1122

Previously, LSP code completion for overloaded functions only showed parameters from the closest matching overload. This PR improves both completions and signature help for overloaded calls:

**Completions:** Show keyword argument and literal completions from all *compatible* overloads, not just the closest match. A new `filter_compatible_overloads` method narrows the overload list by checking whether the positional arguments already provided are type-compatible with each overload's parameters. This means e.g. `foo(1, ` only suggests params from overloads that accept `int` as the first argument. Results are deduplicated so the same param/literal doesn't appear multiple times.

**Signature help:** All overloads are shown with the best-matching one highlighted via `active_signature`. The `get_callables_from_call` return type now uses `Option<usize>` for the overload index to distinguish "a specific overload was matched" from "overloads were expanded without call-site analysis". It also returns `provided_arg_ranges` so callers can filter overloads by compatibility.

**Bug fix:** Fixed a pre-existing bug in `expected_call_argument_type` where `chosen_overload_index` was used as a direct index without unwrapping the `Option`.

## Test Plan

- Added completion tests for overloaded functions (kwarg completions filtered by compatible overloads)
- Added three signature help tests covering `active_signature` tracking, keyword argument highlighting, and three-overload display
- `python3 test.py`

### DIFF
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -290,9 +290,6 @@ dim.get("")
         report.contains(
             r#"
 Completion Results:
-- (Value) 'x': Literal['x'] inserting `x`
-- (Value) 'y': Literal['y'] inserting `y`
-- (Value) 'z': Literal['z'] inserting `z`
 - (Field) x: int
 - (Field) y: int
 - (Field) z: int
@@ -324,9 +321,6 @@ dim.get(key="")
         report.contains(
             r#"
 Completion Results:
-- (Value) 'x': Literal['x'] inserting `x`
-- (Value) 'y': Literal['y'] inserting `y`
-- (Value) 'z': Literal['z'] inserting `z`
 - (Field) x: int
 - (Field) y: int
 - (Field) z: int
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -1015,13 +1015,20 @@ impl Transaction<'_> {
                         &mut result,
                         in_string_literal,
                     );
-                    self.add_literal_completions(handle, position, &mut result, in_string_literal);
-                    self.add_dict_key_completions(
+                    let dict_key_claimed = self.add_dict_key_completions(
                         handle,
                         mod_module.as_ref(),
                         position,
                         &mut result,
                     );
+                    if !dict_key_claimed {
+                        self.add_literal_completions(
+                            handle,
+                            position,
+                            &mut result,
+                            in_string_literal,
+                        );
+                    }
                     // in foo(x=<>, y=2<>), the first containing node is AnyNodeRef::Arguments(_)
                     // in foo(<>), the first containing node is AnyNodeRef::ExprCall
                     if let Some(first) = nodes.first()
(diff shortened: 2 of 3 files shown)


migeed-z:
qq: why are mypy results differing? I would have assumed it stays the same if it's just lsp?

fangyi-zhou:
> qq: why are mypy results differing? I would have assumed it stays the same if it's just lsp?

I'm not sure, maybe this was introduced by a minor change introduced outside of LSP changes. I'll investigate it later.

yangdanny97:
There might be non-determinism in this project's results (perhaps a new regression?), cc @stroxler

stroxler:
It's almost certainly nondeterminism (the Unknowns are a hint), I'll disable this project

fangyi-zhou:
Thanks for the review @kinto0 , could you elaborate on this?
> it only works for kwarg completions

Do we have any other test cases at the moment for other completions? I probably have missed them.

kinto0:
> if only works for kwarg completions

logic was only added in the `add_kwargs_completions` function. instead, we should make this fix generalizable across everything (see details below)

> it stops working after an argument is provided

```
// When args have been provided that narrow down the overload, use only
            // the closest matching overload.
```

> I wonder if instead we should change get_callables_from_call to return a list of compatible overloads

right now, `add_kwargs_completions` calls `get_callables_from_call` and binds the result to `(callables, chosen_overload_index, active_argument, _)`. if we can also return `compatible_overloads` here, we can implement this issue for every lsp request, not just kwarg completions. namely, signature help (screenshot in original issue). 

> Do we have any other test cases at the moment for other completions?

biggest one is signature help (technically not completions), but other ones could be the custom ones like dictionary key completions.

kinto0:
thanks for the updates! please mark this as ready for review when you're ready for a review

fangyi-zhou:
> thanks for the updates! please mark this as ready for review when you're ready for a review

Will do! I didn't have time to review the AI output yet, will mark this ready for review after I'm confident about the change.

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -290,6 +290,9 @@ dim.get("")
         report.contains(
             r#"
 Completion Results:
+- (Value) 'x': Literal['x'] inserting `x`
+- (Value) 'y': Literal['y'] inserting `y`
+- (Value) 'z': Literal['z'] inserting `z`
 - (Field) x: int
 - (Field) y: int
 - (Field) z: int
```
kinto0: (pre-existing) hmm, I wonder if we can remove whatever was populating the old typeddict completions here? maybe it was a workaround

kinto0 (approved):
looks awesome! let me get a second set of eyes

kinto0:
I will try landing it as-imported once I get another set of eyes. we can open another PR for that next enhancement of removing duplicates (which is pre-existing and not your fault, not really a regression)

kinto0:
@fangyi-zhou thanks for the hard work! if you want to remove duplicates, please open a new PR for [this](https://github.com/facebook/pyrefly/pull/2241/changes/8f9fff3c5c30f8492f4e954abeb3dd3f30d10a17) so it can be reviewed separately (should be easy to rebase)

PR #2242, opened 2026-01-28
Check type argument scope in constructor self annotation

fangyi-zhou:
# Summary

Per spec for constructor calls https://typing.python.org/en/latest/spec/constructors.html#init-method

> Class-scoped type variables should not be used in the self annotation
> because such use can lead to ambiguous or nonsensical type evaluation
> results.

This diff implements this requirement by adding a specific and targeted check. If a type variable is used in the `self` annotation in the `__init__` method, an error will be raised if the type variable is defined in the class scope.

This gives us another file with passing conformance tests.

# Test Plan

Added new test case, checking for type parameters declared via both `Generic[T]` and  syntax.

`python test.py`


rchen152 (approved):

Issue #1555, opened 2025-11-12
variable typed as Callable with TypeVar has wrong type


xhantu:
### Describe the Bug

```python
from typing import *

T = TypeVar('T')
f: Callable[[T], T] = lambda x: x
reveal_type(f)
r: int = f(1)

```

```
INFO sandbox.py:5:12-15: revealed type: (TypeVar[T]) -> TypeVar[T] [reveal-type]
ERROR sandbox.py:6:10-14: `TypeVar[T]` is not assignable to `int` [bad-assignment]
ERROR sandbox.py:6:12-13: Argument `Literal[1]` is not assignable to parameter with type `TypeVar[T]` [bad-argument-type]
```

Python: 3.12

The type of `f` should be `[T](T) -> T` and not `(TypeVar[T]) -> TypeVar[T]`

Because of this the usage of `f` fails with type errors.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRAFToBeOmOYwAaqkoAKAORjVAShFhEdAMKooUVNlgBtc2IC6AGhnWpdEzWyZUdfHvwjKMAG4wRgD6TMQwymDa6JR6rPzSYMoAjJogtiAArgzQcCTkiCAAxHQAqjlQEEx0YJnoAMY5uOhwIiKYMGA1vDSoDMHoma4wKl6c6AyadAC0AHx0cAyxInSrdH4MmZToNUIgAHJD2CN6wPgAvnsi6SBkfmBQpIQMtFAUJQAKpPePCxg4BDo9WakDYWz6EGahBEJQAyjAYHQABYMBjEOCIAD0mLunUehF4bExMHQmMwuHqcExwPQoPBTVJ3UodFQ-lQ0FMsCBIIgYMoEOadFwxAZeREZAYSOa00ClDgkJ20j2AGZCMkAExXdAgc4ZVCNCCBABi0BgFDQWDwRDIOqAA

yangdanny97:
The bug appears to be that we're not checking the type annotation for type variables & inferring a generic type from it.

bluetech:
Real-life examples from django-stubs in case it helps in any way:
https://github.com/typeddjango/django-stubs/blob/f788f6ce9833c6805bad7a3cd19c77893b9f9063/django-stubs/views/decorators/http.pyi

PR #2244, opened 2026-01-28
Fix not inferring legacy TypeVar in a `var: Callable[...]` annotation (without assignment)

### LINKED ISSUE #2118: Decorator function annotated with `Callable` + `TypeVar` cannot be used
### Describe the Bug

This is similar to #1555, however the commit 59098f44a5dba811b645214ba06d6ed157210415 included in 0.48.0 did not resolve the issue as I encounter it. Please see the unexpected error below, I tried to minimize it (see https://github.com/typeddjango/django-stubs/blob/f788f6ce9833c6805bad7a3cd19c77893b9f9063/django-stubs/views/decorators/http.pyi for the real definition of `require_GET`).

```py
from typing import Any, TypeVar

F = TypeVar("F", bound=Callable[..., Any])

require_GET: Callable[[F], F]

@require_GET
def view() -> Any: ...
```

Error:

```
ERROR Argument `() -> Any` is not assignable to parameter with type `TypeVar[F]` [bad-argument-type]
 --> x.pyi:8:1
  |
8 | @require_GET
  | ^^^^^^^^^^^^
  |
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIIANCGQE4xhSmEAuuAtlBQMQAEABVIMmpPmix58fAMa50kAOYBXOqlYR5hADrp%2BAZRgw%2BAC1aticRAHpr9Rs0K46i6zHTXMuGXGtyFECpqGvLWfGDOfKgAbqjQqNiwsvJKquqa6Hy4xCHocDroZKwm8gC00TB0cBl8ALx82iAAzIQAjABMjbrUIERkEBRgdJzJULAyufkJMnwQHMTOrHwAwqhjCbC6QyOspMQQ6Iqz84t8AILopFR8ACp7MABqqHS6ugBidbf3T3QAFI1vRrXPDKLC1VbrRIwADahDh1wupAAugBKV7oBgAR2UEAYAH0AOIAURuiBWaygGxh0LeSOutPRAAEsTj8cSbrpMIw%2BNEIDAAO6-FF8UoAPnOlzJcMIIAAvjRUBMIBU3tAYBQ%2BqQBrKgA

_No response_

bluetech:
Fix #2118. Since this issue is blocking my experimentation with pyrefly, I decided to see if I can figure out the problem. After looking at the previous fix and some debugging I think I found the issue and did the "obvious" fix, although admittedly I don't fully understand the details so it might be wrong. Maybe @asukaminato0721 (the author of 59098f44a5dba811b645214ba06d6ed157210415) would be kind enough to see if it makes sense.

---

The commit 59098f44a5dba811b645214ba06d6ed157210415 fixed issue #1555 but only added the `wrap_callable_legacy_typevars` for `NameAssign` bindings. However, I believe the same thing happens with `AnnotatedType` bindings as can appear in stubs. So the following was fixed:

```py
require_GET: Callable[[F], F] = lambda x: x
```

but not

```py
require_GET: Callable[[F], F] = ...
require_GET: Callable[[F], F]
```

# Summary

Fixes #2118

# Test Plan

I added a test for the original minimal reproducer.

I also ran it on my project which originally showed the issue and it is fixed without visible regressions.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -3846,8 +3846,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     Type::ClassDef(cls.dupe())
                 }
             },
-            Binding::AnnotatedType(ann, val) => match &self.get_idx(*ann).ty(self.stdlib) {
-                Some(ty) => (*ty).clone(),
+            Binding::AnnotatedType(ann, val) => match self.get_idx(*ann).ty(self.stdlib) {
+                Some(ty) => self.wrap_callable_legacy_typevars(ty),
                 None => self.binding_to_type(val, errors),
             },
             Binding::Type(x) => x.clone(),
--- a/pyrefly/lib/test/callable.rs
+++ b/pyrefly/lib/test/callable.rs
@@ -61,6 +61,30 @@ f("hello")  # E: `str` is not assignable to upper bound `int` of type variable `
 );
 
 testcase!(
+    test_callable_annotation_only_typevar,
+    TestEnv::one_with_path(
+        "foo",
+        "foo.pyi",
+        r#"
+from collections.abc import Callable
+from typing import Any, TypeVar
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+require_GET: Callable[[F], F]
+"#
+    ),
+    r#"
+from typing import reveal_type
+from foo import require_GET
+def view() -> None:
+    return None
+reveal_type(require_GET)  # E: revealed type: [F: (...) -> Any](F) -> F
+reveal_type(require_GET(view))  # E: revealed type: () -> None
+"#,
+);
+
+testcase!(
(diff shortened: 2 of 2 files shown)


pyrefly/lib/alt/solve.rs
```diff
@@ -3846,8 +3846,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     Type::ClassDef(cls.dupe())
                 }
             },
-            Binding::AnnotatedType(ann, val) => match &self.get_idx(*ann).ty(self.stdlib) {
-                Some(ty) => (*ty).clone(),
```
bluetech, line 3850: A bit of an unrelated change here: I am not a Rust expert but I don't know a reason to match on a reference and then clone, where it's possible to do a move. But if there's a reason for it I can change it back.

PR #2251, opened 2026-01-28
Model GenericAlias runtime type

fangyi-zhou:
# Summary

`GenericAlias` is the runtime time when a type argument is applied on a generic class (via subscripting)
https://docs.python.org/3.15/library/stdtypes.html#generic-alias-type

While the typing spec does not mention how this runtime time should be treated, we are adding some modelling of this behaviour in this diff, i.e. supporting attribute lookup on `GenericAlias`: the doc specifies a number of special attributes such as `__origin__`, `__args__`, `__parameters__` (https://docs.python.org/3.15/library/stdtypes.html#special-attributes-of-genericalias-objects).

Notably, generic aliases support `|` operator, and can construct a union type with string literals (See discussion in #2048).

# Test Plan

`test.py` with new test case.

### DIFF
--- a/pyrefly/lib/test/simple.rs
+++ b/pyrefly/lib/test/simple.rs
@@ -744,17 +744,14 @@ for a in A:  # E: Type `type[A]` is not iterable
     "#,
 );
 
-// It's debatable whether it is possible to construct a value of `type[A[int]]`.
-// If we consider `A[int]` to be of type `type[A[int]]`, then this test captures the bahavior correctly.
-// If not (based on the philosophy that `type[...]` must be an instance of `type`), then we cannot construct such a value,
-// which makes this test vacuously correct.
 testcase!(
-    test_getitem_can_iterate_generic_class,
+    bug = "Although A[int] is a generic alias and can be iterated over, type[A[int]] is a type and cannot be iterated over.",
+    test_getitem_cannot_iterate_generic_class,
     r#"
 class A[T]:
     def __getitem__(self, i: int) -> int: ...
 def f(x: type[A[int]]):
-    for a in x:
+    for a in x:  # Expect an error here
         pass
     "#,
 );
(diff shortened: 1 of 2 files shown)


pyrefly/lib/test/simple.rs
```diff
@@ -674,7 +674,7 @@ class B[T]:
     pass
 for x in A[str]:
     assert_type(x, int)
-for _ in B[str]:  # E: Type `type[B[str]]` is not iterable
+for _ in B[str]:
```
fangyi-zhou, line 677:
```
$ cat test.py
class B[T]:
    pass
for x in B[str]:
    print(x)
$ python3 test.py
typing.Unpack[__main__.B[str]]
```

This is actually valid 😮

fangyi-zhou:
cc @yangdanny97 over previous discussion on discord

fangyi-zhou:
I think I introduced a regression for comparing generic aliases (now `list[int] == list[str]` is an error)

pyrefly/lib/test/simple.rs
```diff
@@ -744,13 +744,17 @@ for a in A:  # E: Type `type[A]` is not iterable
     "#,
 );
 
+// It's debatable whether it is possible to construct a value of `type[A[int]]`.
+// If we consider `A[int]` to be of type `type[A[int]]`, then this test captures the bahavior correctly.
+// If not (based on the philosophy that `type[...]` must be an instance of `type`), then we cannot construct such a value,
+// which makes this test vacuously correct.
 testcase!(
-    test_getitem_cannot_iterate_generic_class,
+    test_getitem_can_iterate_generic_class,
```
stroxler: I'm actually not sure what part of the PR changes this.

I think it's somewhat mistaken. The type `type[A[int]]` means a `type` instance (a class object) of some class that is a subtype of `A[int]`.

This actually can exist, e.g. if I define
```
class B(A[int]):
    ...
```
then `B` is a plain class object (not a `GenericAlias`) and is a valid value of type `type[A[int]]`.

But note that `B` is not iterable - `B()` is, but the class object is not.

If it's hard to change the behavior, I'm okay with merging as-is if we mark this as a bug and update the comments (I'm not honestly all that worried about false negatives in weird edge cases - at the moment false positives matter more), but I do think we should track the issue

fangyi-zhou:
@stroxler Thanks for the example, that was something I didn't consider. I've added the `bug` annotation and some comments in the code to document this caveat.

stroxler (approved):

Issue #650, opened 2025-07-09
Tracker: Support all type guards from Pyright


yangdanny97:
### Describe the Bug

From https://microsoft.github.io/pyright/#/type-concepts-advanced?id=type-guards

Type guards of  the form `x.y` or `x[y]` are all partially-supported: we currently apply the narrowing to the type of `x[y]`, but do not narrow the type of `x` (if it's something that can be discriminated by that field/index/key)

- [x] `x is None` / `x is not None`
- [x] `x == None` / `x != None`
- [ ] `x is ...` / `x is not ...` / `x == ...` / `x != ...` (where ... is an ellipsis token) *Note: as far as I can tell, this is just for narrowing `object` or `Any` to `Ellipsis`*
- [ ] `x is S` / `x is not S` (where S is a Sentinel) *Note: Sentinel here seems to refer to some `Final` variable - even if `S` is not a literal, if it is final we can use it for type narrowing.*
- [x] `type(x) is T` / `type(x) is not T`  https://github.com/facebook/pyrefly/issues/335
- [x] `type(x) == T` / `type(x) != T` https://github.com/facebook/pyrefly/issues/335
- [x] `x is L` / `x is not L` (where L is an expression that evaluates to a literal type)
- [x] `x is C` / `x is not C` (where C is a class)
- [x] `x == L` / `x != L` (where L is an expression that evaluates to a literal type)
- [x] `x.y is None` / `x.y is not None` (where x is a type that is distinguished by a field with a None)
- [x] `x.y is E` / `x.y is not E` (where E is a literal enum or bool and x is a type that is distinguished by a field with a literal type) https://github.com/facebook/pyrefly/issues/418
- [x] `x.y == LN` / `x.y != LN` (where LN is a literal expression or None and x is a type that is distinguished by a field or property with a literal type) https://github.com/facebook/pyrefly/issues/418
- [x] `x[K] == V` / `x[K] != V` / `x[K] is V` / `x[K] is not V` (where K and V are literal expressions and x is a type that is distinguished by a TypedDict field with a literal type)  https://github.com/facebook/pyrefly/issues/418
- [x] `x[I] == V` / `x[I] != V` (where I and V are literal expressions and x is a known-length tuple that is distinguished by the index indicated by I)  https://github.com/facebook/pyrefly/issues/418
[...]

PR #2252, opened 2026-01-29
impl Sentinel, part of #650

asukaminato0721:
# Summary

Fixes part of #650

Added a sentinel literal type and wired is/is not narrowing to use final-name initializers, so Final sentinels now narrow even when the declared type isn’t literal. Updated literal display/output handling.

# Test Plan

Added narrowing tests for Final bool and object sentinels.

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -262,10 +262,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         if !annotation.annotation.is_final() {
             return None;
         }
-        let Binding::NameAssign { expr, .. } = self.bindings().get(binding_idx) else {
+        let Binding::NameAssign(name_assign) = self.bindings().get(binding_idx) else {
             return None;
         };
-        Some(self.expr_infer(expr.as_ref(), &self.error_swallower()))
+        Some(self.expr_infer(name_assign.expr.as_ref(), &self.error_swallower()))
     }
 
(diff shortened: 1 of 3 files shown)


yangdanny97 (changes requested):
Hmm, I don't like adding a new literal case to represent a potentially non-literal type.

This is the implementation I have in mind:
- at binding phase, if we see a `is <some variable>` we can generate a IsSentinel/IsNotSentinel narrow-op that holds 1) the variable binding idx and 2) the annotation idx
- when that narrow op is solved it looks up the variable, looks up the annotation, and narrows the subject if the annotation is final. if it isn't final then we do nothing

yangdanny97:
There's a potential performance regression here, I'll investigate when I have time.

yangdanny97:
I'm working on merging this one, needs a bit of splitting up to make it reviewable internally. No action needed from you ATM, thanks

yangdanny97:
I was looking at the code here, and I think we may already handle `is`/`is not` narrowing correctly for `Final` variables, as long as it's not explicitly annotated with a wider type. (and if it is, we should respect that wider type).

I think my guidance on #650 was actually wrong re: this being about final variables, looking at the history of that feature in Pyright this seems to actually be for PEP 661, which is not part of the typing spec (yet). The `Sentinel` type was added to `typing_extensions` for implementation, but the PEP is not accepted.

https://github.com/microsoft/pyright/issues/10565

In that case, I'm sorry I sent you down the wrong path here, I really appreciate the work here still

Issue #2207, opened 2026-01-23
Pyrefly IDE Feature Tracker


jvansch1:
### Describe the Bug

This issue should be the main source of truth for the status of all Pyrefly IDE features.

## Core features(Originally tracked by https://github.com/facebook/pyrefly/issues/344)

- [X] Go To Definition
- [X] Go To Type Definition
- [X] Find References
- [X] Document Symbols
- [X] Workspace Symbols
- [X] Hover
- [X] Document Highlights
- [X] Signature Help
- [X] Completion
- [X] Rename
- [X] Call Hierarchy
- [X] Publish Diagnostics
- [X] Semantic Tokens
- [X] Inlay Hints
- [X] Notebook Support
- [X] Implementation

## Refactoring Support(Originally tracked by https://github.com/facebook/pyrefly/issues/364)

- [X] Rename variables
- [X] Rename Classes
- [X] Rename Functions
- [x] Rename Modules
- [ ] Rename Packages

## Code Actions(Originally tracked by https://github.com/facebook/pyrefly/issues/364)

- [X] Extract Method
- [X] Extract Variable
- [x] Inline Method
- [x] Inline Variable
- [x] Inline Parameter
- [ ] Use function([Rope](https://rope.readthedocs.io/en/latest/overview.html#use-function-refactoring))
- [ ] Method to Method Object
- [ ] Convert Local variable to field
- [ ] Organize Imports([Rope](https://rope.readthedocs.io/en/latest/overview.html#sorting-imports))
- [x] Generate Variable
- [x] Generate Function
- [x] Generate Class
- [ ] Implement missing members (protocol, abc)
- [ ] add type ignore
- [ ] add all type ignores (project)
- [ ] import unused (entire module)

## PyCharm Refactoring Features(Originally tracked by https://github.com/facebook/pyrefly/issues/364)

- [x] Invert Boolean([Pycharm](https://www.jetbrains.com/help/pycharm/invert-boolean.html))
- [ ] Change Signature([Pycharm](https://www.jetbrains.com/help/pycharm/change-signature.html))
- [X] Convert to Package and Module([Pycharm](https://www.jetbrains.com/help/pycharm/refactoring-convert.html))
- [X] Copy and Move refactors([Pycharm](https://www.jetbrains.com/help/pycharm/move-refactorings.html))
- [X] Pull members up and down([Pycharm](https://www.jetbrains.com/help/pycharm/pull-members-up.html))
[...]

PR #2254, opened 2026-01-29
impl Completions: Auto add function call parenthesis #2207

asukaminato0721:
# Summary

Fixes part of #2207

 Added optional auto‑parentheses for function/method completion items behind `python.analysis.completeFunctionParens`, with snippet insertion when the client supports it.

Plumbed the setting through LSP config, exposed it in the VS Code settings.

# Test Plan

Added an LSP interaction test that asserts snippet behavior.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -250,6 +250,7 @@ use crate::lsp::non_wasm::will_rename_files::will_rename_files;
 use crate::lsp::non_wasm::workspace::LspAnalysisConfig;
 use crate::lsp::non_wasm::workspace::Workspace;
 use crate::lsp::non_wasm::workspace::Workspaces;
+use crate::lsp::wasm::completion::CompletionOptions as CompletionRequestOptions;
 use crate::lsp::wasm::completion::supports_snippet_completions;
 use crate::lsp::wasm::hover::get_hover;
 use crate::lsp::wasm::notebook::DidChangeNotebookDocument;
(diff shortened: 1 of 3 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -2858,6 +2859,25 @@ impl<'a> Transaction<'a> {
...
+            handle,
+            position,
+            import_format,
+            supports_completion_item_details,
+            false,
+            false,
+        )
+    }
+
+    pub fn completion_with_incomplete_with_function_parens(
```
jvansch1: Any new functionality related to completions should ideally go in https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/lsp/wasm/completion.rs

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -2003,6 +2003,17 @@ impl Server {
             .unwrap_or(false)
     }
 
+    fn supports_snippet_completions(&self) -> bool {
```
jvansch1: Can we move this to wasm/completion.rs? Also I think this function needs a docstring.

pyrefly/lib/state/lsp.rs
```diff
@@ -3103,6 +3138,27 @@ impl<'a> Transaction<'a> {
         (result, is_incomplete)
     }
 
+    fn add_function_call_parens(completions: &mut [CompletionItem], supports_snippets: bool) {
```
jvansch1: This should also go in the completion.rs file.

pyrefly/lib/state/lsp.rs
```diff
@@ -2858,6 +2859,25 @@ impl<'a> Transaction<'a> {
...
+            false,
+            false,
+        )
+    }
+
+    pub fn completion_with_incomplete_with_function_parens(
+        &self,
+        handle: &Handle,
+        position: TextSize,
+        import_format: ImportFormat,
```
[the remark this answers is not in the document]
jvansch1: I think this is good advice. We should probably add something simple here

pyrefly/lib/state/lsp.rs
```diff
@@ -2609,6 +2609,25 @@ impl<'a> Transaction<'a> {
         position: TextSize,
         import_format: ImportFormat,
         supports_completion_item_details: bool,
+    ) -> (Vec<CompletionItem>, bool) {
+        self.completion_with_incomplete_with_function_parens(
```
jvansch1: I feel like this is an unnecessary abstraction here. We now have the outer `completions_with_incomplete` function which then just calls `completion_with_incomplete_with_function_parens`. Could we instead just update the logic in `completions_with_incomplete` directly? It looks like this function is only called in two places so it should be fairly easy to update the call site with new arguments if we need to do that. What do you think?

kinto0 (approved):

PR #2271, opened 2026-01-30
Fix SelfType attribute lookup when base class is Any

### LINKED ISSUE #2269: `Any` as base class; No attribute errors should be reported
### Describe the Bug

The spec [mentions](https://typing.python.org/en/latest/spec/special-types.html#any): 
> Any can also be used as a base class. This can be useful for avoiding type checker errors with classes that can duck type anywhere or are highly dynamic.

But, the following fails with pyrefly:
```
from typing import Any
class MyTest(Any):
  def foo(self):
    self.bar() # missing-attribute error
```
Can pyrefly be similar to other type checkers (pyright, pytype, and ty) here and not report any `missing-attribute` errors if `Any` is in inheritance the hierarchy of the class?

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn0%2BjA0pYvVGKkxsMCAEF0KR9IwoKg4MoALKkAAq8A42x1pF2tV8-NEuVwuG2CygYHt%2BmovuoHrAhGw7j21H0IDkIDIIjEpEIXBYUAoNEEMZFlVk%2BEy6EgWiKBTJcSSkhg7B4HA4xDgiAA9DXo6JxIRYVoazB0DXMOi4DXmDmIHmhAXYjWXRLUEtUNBUNgcn3c-mujzcMQl%2Bg4PF0GRueg6lS4IXqABeMMgADMhAAjAAmXQgcMAX0jUy6KwAYtAYBQM9USOQH0AA

_No response_

tsembp:
# Summary

Fixes #2269.

### What changed
- When looking up attributes from `AttributeBase1::SelfType`, if the underlying class has `Any` in its base hierarchy, fall back to `Any` instead of reporting `missing-attribute`.

### Why
Pyrefly already suppresses missing-attribute errors for other attribute bases when `has_base_any()` is true, but `SelfType` didn’t apply the same rule. This caused a `missing-attribute` error inside methods for classes inheriting from `Any`.

### Tests
- Added `test_any_as_base_class_suppresses_missing_attribute_in_method`
- `cargo test` passes

# Test Plan

- `cargo test`

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -1512,7 +1512,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
             AttributeBase1::SelfType(cls) => match self.get_self_attribute(cls, attr_name) {
                 Some(attr) => acc.found_class_attribute(attr, base),
-                None => acc.not_found(NotFoundOn::ClassInstance(cls.class_object().dupe(), base)),
+                None => {
+                    let metadata = self.get_metadata_for_class(cls.class_object());
+                    if metadata.has_base_any() {
+                        acc.found_type(Type::Any(AnyStyle::Implicit), base)
+                    } else {
+                        acc.not_found(NotFoundOn::ClassInstance(cls.class_object().dupe(), base))
+                    }
+                }
             },
             AttributeBase1::Intersect(bases, fallback) => {
                 // For now, only handle the simplest case: if exactly one base has a successful lookup, use it.
--- a/pyrefly/lib/test/attributes.rs
+++ b/pyrefly/lib/test/attributes.rs
@@ -1054,6 +1054,16 @@ class Test(B):
 );
 
 testcase!(
+    test_any_as_base_class_suppresses_missing_attribute_in_method,
+    r#"
+from typing import Any
+class MyTest(Any):
+    def foo(self):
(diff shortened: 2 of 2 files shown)


migeed-z:
lgtm!

samwgoldman (approved):

samwgoldman:
Thank you!

PR #2277, opened 2026-01-31
fix Overloaded function is not assignable to Callable #2273

### LINKED ISSUE #2273: Overloaded function is not assignable to Callable
### Describe the Bug

pyrefly 0.50.1 says that an overloaded function is incompatible with a Callable that takes a union of the overload argument types and returns (a union of) the overload return types, resulting in `bad-assignment` or `bad-argument-type` errors being reported depending on context.

Minimum reproducible example:

```python
from typing import *

@overload
def foo(x: int) -> str: ...

@overload
def foo(x: str) -> str: ...

def foo(x: int | str) -> str:
    return str(x)

bar: Callable[[int | str], str] = foo
# `Overload[
#   (x: int) -> str
#   (x: str) -> str
# ]` is not assignable to `(int | str) -> str` [bad-assignment]
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRAAVwA3GJSi5UmEZhhg6YXLgAU%2BRJ3QMAlHQC0APjpwGlXYVsjxUmXIVKVajdt2XKRsxas2dqJYbupaOnr8AD7%2BPibm3ogidCl0lDAMAK6U6LHaBvbo2KjWdADCqFBQqNiwANp1rNGxALoANK10ALzuuCIgbSBk6WBQpIQMtFAUAMR0AAqkI2MWGDgEdADGuOiQbNmoDBA7hCJzAMowMHQAFgwMxHCIAPTPwypjhLxszzDoz5hcJs4M9trsIPtKIdjv93JQ6KhJKhoDVYFsdnsDkcdnRcMRseg4Kd0GQGDcdsZpJQ4DDunQhCAAMyEACMACYGf0AL6DVCbI7SABi0BgFDQWDwRDIIC5QA

_No response_

asukaminato0721:
# Summary

Fixes #2273

Adjusted overload-to-Callable subtyping so single-parameter overloads can satisfy a union-parameter Callable (the case in issue 2273), while keeping the existing per‑signature fallback.

# Test Plan

Added a regression test.

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -14,7 +14,6 @@ use pyrefly_types::callable::ArgCounts;
 use pyrefly_types::callable::Param;
 use pyrefly_types::tuple::Tuple;
 use pyrefly_types::types::TArgs;
-use pyrefly_types::types::Union;
 use pyrefly_util::gas::Gas;
 use pyrefly_util::owner::Owner;
 use pyrefly_util::prelude::SliceExt;
@@ -36,11 +35,12 @@ use crate::config::error_kind::ErrorKind;
 use crate::error::collector::ErrorCollector;
 use crate::error::context::ErrorContext;
 use crate::error::context::ErrorInfo;
+use crate::overload_expansion::OVERLOAD_PARAM_EXPANSION_GAS;
+use crate::overload_expansion::expand_type_for_overload;
 use crate::types::callable::Callable;
 use crate::types::callable::FuncMetadata;
 use crate::types::callable::Function;
 use crate::types::callable::Params;
-use crate::types::literal::Lit;
 use crate::types::types::Type;
 
 struct CalledOverload {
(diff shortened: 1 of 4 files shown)


rchen152 (changes requested):
I have a few concerns about this fix:

* It seems overly specific to the exact example in #2273. Why are we only handling unions? Why only in the first parameter?
* Merging callable signatures like this seems fragile and error-prone. If we're going this route, I'd want to see a robust set of tests. But my hope is that we can avoid doing this altogether. (See below.)
* We're only grabbing the first signature from the `want` type. This seems potentially questionable if `want` is an overloaded function type.

The root of the problem seems to be that no single signature of the overloaded function is assignable to the `Callable` type because it unions together the overloads' parameter types. This is similar to the problem solved by "argument type expansion" (https://typing.python.org/en/latest/spec/overload.html#step-3) in overload evaluation, which makes me think that we may be able to fix it the same way. The rough idea would be:

(1) first call `is_subset_eq` on every overload (current behavior), then
(2) if this fails and `u` is a single-signature callable, (a) check that the union of the overloads' return types is assignable to the callable's return type, and (b) turn `u`'s parameters into a fake argument list, expand it, and check if every expansion is accepted by the overloaded function. The existing `ArgsExpander` could likely be reused: https://github.com/facebook/pyrefly/blob/50293ea618f6f91ebb3a949ca92c165b91ccad40/pyrefly/lib/alt/overload.rs#L54.

Argument type expansion is expensive, so we'd have to check pretty carefully that this doesn't regress perf.

asukaminato0721:
now reused

PR #2278, opened 2026-01-31
fix Implement type hierarchy #2264

### LINKED ISSUE #2264: Implement type hierarchy
### Describe the Bug

this involves obtaining the MRO of the type

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2264

 Added LSP type hierarchy support end‑to‑end: prepare/supertypes/subtypes handlers backed by class MRO + reverse‑dep scanning, and injected typeHierarchyProvider into the initialize response so clients can enable the feature.

# Test Plan

added lsp test.

### DIFF
--- a/pyrefly/lib/commands/lsp.rs
+++ b/pyrefly/lib/commands/lsp.rs
@@ -86,17 +86,7 @@ fn initialize_connection(
         return Ok(None);
     };
     let capabilities = capabilities(indexing_mode, &initialize_params);
-    let type_hierarchy_provider = match indexing_mode {
-        IndexingMode::None => None,
-        IndexingMode::LazyNonBlockingBackground | IndexingMode::LazyBlocking => Some(true),
-    };
-    if !initialize_finish(
-        connection,
-        id,
-        capabilities,
-        server_info,
-        type_hierarchy_provider,
-    )? {
+    if !initialize_finish(connection, id, capabilities, server_info)? {
         return Ok(None);
     }
     Ok(Some(initialize_params))
--- a/pyrefly/lib/commands/tsp.rs
+++ b/pyrefly/lib/commands/tsp.rs
@@ -64,12 +64,8 @@ fn initialize_tsp_connection(
         return Ok(None);
     };
     let capabilities = tsp_capabilities(indexing_mode, &initialize_params);
-    let type_hierarchy_provider = match indexing_mode {
-        IndexingMode::None => None,
-        IndexingMode::LazyNonBlockingBackground | IndexingMode::LazyBlocking => Some(true),
-    };
     // Note: TSP doesn't include serverInfo, unlike LSP
-    if !initialize_finish(connection, id, capabilities, None, type_hierarchy_provider)? {
+    if !initialize_finish(connection, id, capabilities, None)? {
         return Ok(None);
     }
     Ok(Some(initialize_params))
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -287,6 +287,7 @@ use crate::state::state::CommittingTransaction;
 use crate::state::state::State;
 use crate::state::state::Transaction;
 use crate::state::subscriber::PublishDiagnosticsSubscriber;
+use crate::types::class::ClassDefIndex;
 
 #[derive(Clone, Copy, Serialize)]
 #[serde(rename_all = "kebab-case")]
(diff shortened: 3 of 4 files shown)


pyrefly/lib/commands/lsp.rs
```diff
@@ -86,7 +86,17 @@ fn initialize_connection(
         return Ok(None);
     };
     let capabilities = capabilities(indexing_mode, &initialize_params);
-    if !initialize_finish(connection, id, capabilities, server_info)? {
+    let type_hierarchy_provider = match indexing_mode {
```
jvansch1: Why don't we just add this to the capabilities function that we are calling on the line above this? It should be very similar to the CallHierarchyProvider in that function.

pyrefly/lib/commands/lsp.rs
```diff
@@ -86,7 +86,17 @@ fn initialize_connection(
...
+        IndexingMode::None => None,
+        IndexingMode::LazyNonBlockingBackground | IndexingMode::LazyBlocking => Some(true),
+    };
+    if !initialize_finish(
+        connection,
+        id,
+        capabilities,
+        server_info,
+        type_hierarchy_provider,
```
jvansch1: If we add this provider to capabilities we won't need to pass this argument anymore.

pyrefly/lib/commands/tsp.rs
```diff
@@ -64,8 +64,12 @@ fn initialize_tsp_connection(
         return Ok(None);
     };
     let capabilities = tsp_capabilities(indexing_mode, &initialize_params);
+    let type_hierarchy_provider = match indexing_mode {
```
jvansch1: Same comment here. We can probably add this to the capabilities function.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -626,14 +639,34 @@ pub fn initialize_start(
 
 // Sends the initialize response and waits for the initialized notification.
 // If the connection is closed, or we receive an exit notification, returns false.
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct ServerCapabilitiesWithTypeHierarchy<'a> {
```
jvansch1: I don't we really need this if we just update the capabilities function to also have a type hierarchy provider.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4246,6 +4316,269 @@ impl Server {
...
+        };
+
+        self.async_find_from_definition_helper(
+            request_id,
+            transaction,
+            handle,
+            &uri,
+            params.item.selection_range.start,
+            FindPreference::default(),
+            move |transaction, handle, definition| {
```
jvansch1, line 4591: Can we try to break this up or move it into a helper function. This is over 100 lines and at this point becoming a bit difficult to read.

pyrefly/lib/tsp/server.rs
```diff
@@ -199,7 +199,7 @@ pub fn tsp_loop(
 pub fn tsp_capabilities(
     indexing_mode: IndexingMode,
     initialization_params: &InitializeParams,
-) -> ServerCapabilities {
+) -> ServerCapabilitiesWithTypeHierarchy {
```
jvansch1, line 202: Sorry maybe I was not clear but this is not exactly what I meant. On line 842 of fbcode/pyrefly/pyrefly/lib/lsp/non_wasm/server.rs we add the call hierarchy provider. We should be able to add the TypeHierarchyProvider in a very similar way. Once we do that we will not need `ServerCapabilitiesWithTypeHierarchy` since type hierarchy will already be included in the existing `ServerCapabilities`

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -626,10 +640,27 @@ pub fn initialize_start(
 
 // Sends the initialize response and waits for the initialized notification.
 // If the connection is closed, or we receive an exit notification, returns false.
-pub fn initialize_finish(
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ServerCapabilitiesWithTypeHierarchy {
```
jvansch1, line 645: I still think we can implement this in a way where these structs are not required. See my other comment.

pyrefly/lib/tsp/server.rs
asukaminato0721: The problem is:

we have https://docs.rs/lsp-types/latest/lsp_types/enum.CallHierarchyServerCapability.html

but we don't have TypeHierarchyServerCapability in lsp_types

jvansch1: Ah good catch! In that case I think this is fine for now.

asukaminato0721: see something very interesting https://github.com/astral-sh/lsp-types/pull/2

astral team patch the dep.

PR #2279, opened 2026-01-31
fix using += on types without __iadd__ will evaluate __getattr__ instead of __add__ #2248

### LINKED ISSUE #2248: using `+=` on types without `__iadd__` will evaluate `__getattr__` instead of `__add__`
for a `.pyi` like

```pyi
import pandas as pd

s: pd.Series[str]
s += "a"
```

, Pyrefly reports

```
`+=` is not supported between `Series[str]` and `Literal['a']`
  Expected `__iadd__` to be a callable, got `str`
```

This is accepted by mypy, pyright, basedpyright, and ty.

I'm guessing this has to do with there not being a `Series.__iadd__` method in the stubs, causing Pyrefly to incorrectly prioritize `Series.__getattr__` (which returns a `str`) over `Series.__add__`.

Used versions:

- `pyrefly==0.50.0`
- `pandas==2.3.3`
- `pandas-stubs==2.3.3.260113`

asukaminato0721:
# Summary

Fixes #2248

Fix scoped operator lookup to ignore `__getattr__` and added a regression test for += when `__iadd__` is missing. This aligns augmented assignment with Python’s runtime semantics for magic methods and prevents `__getattr__` from hijacking `__iadd__`/`__add__` resolution.

# Test Plan

add a test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,8 +231,7 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 [[package]]
 name = "backtrace"
 version = "0.3.76"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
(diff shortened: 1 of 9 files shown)


stroxler (approved):

PR #2284, opened 2026-02-01
fix Narrowing for isinstance and type[T] is unsound #713

### LINKED ISSUE #713: Narrowing for isinstance and type[T] is unsound

asukaminato0721:
# Summary

Fixes #713

Adjusted isinstance negative narrowing so we only subtract when the classinfo is a concrete class object (a class def or alias to one), avoiding the unsound narrowing for type[T].

Adjusted negative isinstance narrowing so we only treat type[...] classinfo as concrete when the classinfo expression is a precise literal (class defs combined with | or tuple literals).

# Test Plan

Added a regression test that mirrors the issue case to ensure the else branch stays at int | str.

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -26,6 +26,7 @@ use ruff_python_ast::Expr;
 use ruff_python_ast::ExprNumberLiteral;
 use ruff_python_ast::Int;
 use ruff_python_ast::Number;
+use ruff_python_ast::Operator;
 use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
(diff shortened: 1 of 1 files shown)


rchen152:
I'll take this one, since I've looked at a couple similar narrowing PRs recently.

PR #2285, opened 2026-02-01
[conformance] Disallow repeated type alias declaration in type stmt

fangyi-zhou:
# Summary

Errors when re-declaring a type alias, per requirement in the conformance test suite.

# Test Plan

`test.py`


migeed-z:
Thank you! I will take a look

PR #2289, opened 2026-02-02
Emit config diagnostics

### LINKED ISSUE #2078: Make some sort of UI signal in VS Code if we fail to parse pyproject.toml
### Describe the Bug

Either a pop-up, or we emit diagnostics for the TOML file.

if the former, we should make sure it doesn't repeatedly re-appear. If the latter, we should make sure it clears once the issue is fixed.

### Sandbox Link

_No response_

_No response_

sgavriil01:
# Summary

This PR adds LSP diagnostics for config file (pyrefly.toml/pyproject.toml) parsing errors. When a config file has invalid syntax, diagnostics now appear in VS Code's Problems panel and clear automatically when the file is fixed.

**Changes:**
- Modified `ConfigError` to include file path information
- Fixed bug in `workspace.rs` where config errors were being discarded
- Added `publish_config_diagnostics()` to emit LSP diagnostics for config files
- Added caching to ensure diagnostics persist across publish cycles

Fixes #2078

# Test Plan

- Added integration test `test_config_file_diagnostics()` that creates invalid TOML and verifies proper handling
- Manually tested by creating syntax errors in pyrefly.toml - diagnostics appear in Problems panel with correct line/column positions
- Verified diagnostics clear when config is fixed
- Ran `./test.py --no-test --no-conformance` - formatting and linting pass

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -1164,9 +1164,9 @@ impl ConfigFile {
 
         if let ConfigSource::File(path) = &self.source {
             configure_errors
-                .into_map(|e| ConfigError::warn(e.context(format!("{}", path.display()))))
+                .into_map(|e| ConfigError::warn(e.context(format!("{}", path.display())), None))
         } else {
-            configure_errors.into_map(ConfigError::warn)
+            configure_errors.into_map(|e| ConfigError::warn(e, None))
         }
     }
 
@@ -1231,7 +1231,7 @@ impl ConfigFile {
                 Ok(Some(config)) => (Some(config), ConfigSource::File(config_path.to_path_buf())),
                 Ok(None) => (None, ConfigSource::Marker(config_path.to_path_buf())),
                 Err(e) => {
-                    errors.push(ConfigError::error_with_path(e, config_path.to_path_buf()));
+                    errors.push(ConfigError::error(e, Some(config_path.to_path_buf())));
                     (None, ConfigSource::File(config_path.to_path_buf()))
                 }
             };
@@ -1247,9 +1247,9 @@ impl ConfigFile {
                     }
                 }
                 None => {
-                    errors.push(ConfigError::error_with_path(
+                    errors.push(ConfigError::error(
                         anyhow!("Could not find parent of path `{}`", config_path.display()),
-                        config_path.to_path_buf(),
+                        Some(config_path.to_path_buf()),
                     ));
                     maybe_config.unwrap_or_else(ConfigFile::default)
                 }
@@ -1258,20 +1258,20 @@ impl ConfigFile {
 
             if !config.root.extras.0.is_empty() {
                 let extra_keys = config.root.extras.0.keys().join(", ");
-                errors.push(ConfigError::warn_with_path(
+                errors.push(ConfigError::warn(
                     anyhow!("Extra keys found in config: {extra_keys}"),
-                    config_path.to_path_buf(),
+                    Some(config_path.to_path_buf()),
                 ));
             }
             for sub_config in &config.sub_configs {
                 if !sub_config.settings.extras.0.is_empty() {
                     let extra_keys = sub_config.settings.extras.0.keys().join(", ");
-                    errors.push(ConfigError::warn_with_path(
+                    errors.push(ConfigError::warn(
                         anyhow!(
                             "Extra keys found in sub config matching {}: {extra_keys}",
                             sub_config.matches
                         ),
-                        config_path.to_path_buf(),
+                        Some(config_path.to_path_buf()),
                     ));
                 }
             }
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -1211,6 +1211,34 @@ impl ConfigFile {
     }
 
     pub fn from_file(config_path: &Path) -> (ConfigFile, Vec<ConfigError>) {
+        /// Extract line and column from toml error by checking if it's a toml::de::Error
+        fn extract_toml_span(error: &anyhow::Error) -> Option<(usize, usize)> {
+            // Check if the error is a toml deserialization error
+            error
+                .downcast_ref::<toml::de::Error>()
+                .and_then(|toml_err| {
+                    // Get the span from toml error
+                    toml_err.span().map(|_span| {
+                        // The span is in bytes, but toml errors also expose line/column in their Display
+                        // For now, we parse the display string which includes "at line X, column Y"
+                        let error_str = toml_err.to_string();
+                        let line = error_str
+                            .split("line ")
+                            .nth(1)
+                            .and_then(|s| s.split(&[',', ' ', '\n'][..]).next())
+                            .and_then(|s| s.parse::<usize>().ok())
+                            .unwrap_or(1);
+                        let column = error_str
+                            .split("column ")
+                            .nth(1)
+                            .and_then(|s| s.split(&[',', ' ', '\n'][..]).next())
+                            .and_then(|s| s.parse::<usize>().ok())
+                            .unwrap_or(1);
+                        (line, column)
+                    })
+                })
+        }
+
         fn read_path(config_path: &Path) -> anyhow::Result<Option<ConfigFile>> {
             let config_str = fs_anyhow::read_to_string(config_path)?;
             if config_path.file_name() == Some(OsStr::new(ConfigFile::PYPROJECT_FILE_NAME)) {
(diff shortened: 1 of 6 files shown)


connernilsen:
Hey @sgavriil01, thanks for making this! It seems like something we definitely need. I'll review early next week

crates/pyrefly_config/src/finder.rs
```diff
@@ -32,36 +33,58 @@ use crate::config::ConfigFile;
...
         }
     }
 
     pub fn warn(msg: anyhow::Error) -> Self {
         Self {
             severity: Severity::Warn,
-            msg,
+            msg: format!("{:#}", msg),
+            file_path: None,
+        }
```
connernilsen: Thoughts on instead making this take an optional path? Then when creating a config error it will be more apparent that it can be tied to a config at a path? I could go either way on this/am open to discussion here though :)

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1916,6 +1923,126 @@ impl Server {
...
+            message: error_message,
+            code: Some(NumberOrString::String("config-error".to_owned())),
+            ..Default::default()
+        }
+    }
+
+    /// Extract line and column numbers from TOML error messages
+    /// Returns (line, column) with 1-based indexing, defaults to (1, 1)
+    fn extract_line_col_from_toml_error(error_msg: &str) -> (usize, usize) {
+        let line = error_msg
```
connernilsen: Instead of trying to parse a position out of an error message, can we instead keep this information as part of the error we construct? TOML deserialization should have [spans](https://docs.rs/toml/latest/toml/de/struct.Error.html#method.span) that will tell you this

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1916,6 +1923,126 @@ impl Server {
...
+    }
+
+    /// Publish diagnostics for config files (pyrefly.toml or pyproject.toml)
+    fn publish_config_diagnostics(&self, transaction: &Transaction) {
+        let mut config_errors = transaction.get_config_errors();
+
+        // If we got fresh errors, deduplicate and update the cache
+        if !config_errors.is_empty() {
+            // Deduplicate by (file_path, message) since the same error can be added multiple times
+            let mut seen = std::collections::HashSet::new();
```
connernilsen: nit: can we use a `SmallSet` here instead? They're more efficient, especially when working with small cardinalities or if you aren't removing things from the set

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1916,6 +1923,126 @@ impl Server {
...
+            config_errors.retain(|err| {
+                let key = (err.file_path().map(|p| p.to_path_buf()), err.get_message());
+                seen.insert(key)
+            });
+            *self.cached_config_errors.lock() = config_errors.clone();
+        } else {
+            // If get_config_errors() returned empty (consumed by mem::take),
```
connernilsen, line 1971: Those errors might be empty because all errors have been fixed. If that's the case, will we not end up clearing diagnostics here?

What we might need to do instead to make sure this is working is to keep a map in the config loader of all config paths to errors that have been loaded since the last time they were taken. Then we can use that to determine if a config has been changed since the last time we published errors. If it's not in the map, then we don't update diagnostics from the previous result. If it is, then completely overwrite with new errors.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1916,6 +1923,126 @@ impl Server {
...
+        let mut tracked_files = self.config_files_with_diagnostics.lock();
+        for previously_erroring_file in tracked_files.iter() {
+            if !current_error_files.contains(previously_erroring_file) {
+                // File no longer has errors, publish empty diagnostics to clear
+                if let Ok(uri) = Url::from_file_path(previously_erroring_file) {
+                    self.connection.publish_diagnostics_for_uri(
+                        uri,
+                        Vec::new(),
+                        None,
+                        DiagnosticSource::DidClose,
```
connernilsen: I don't think we want to use `DidClose` here unless we're sure it's from a close operation.

Instead, this function (`publish_config_diagnostics`) is only called from `publish_for_handles` which takes a `DiagnosticSource` as an arg. I think we should reuse that.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1916,6 +1923,126 @@ impl Server {
...
+        }
+
+        // Publish diagnostics for files with errors
+        for (path, diagnostics) in config_diags {
+            if let Ok(uri) = Url::from_file_path(&path) {
+                self.connection.publish_diagnostics_for_uri(
+                    uri,
+                    diagnostics,
+                    None,
+                    DiagnosticSource::CommittingTransaction,
```
connernilsen: Same here as above

connernilsen (changes requested):
Hey, thanks for doing this! It's looking great so far and I love the motivation. I left a few comments, mostly small things, but I do think there needs to be a different approach for how we figure out which files to update for, which I left in the else branch when we're getting the diagnostics to publish. 

Let me know if you have any questions or disagree, I'm totally down to discuss things or give more thoughts if you want :)

sgavriil01:
Thanks for the review! I've addressed the first 4 points:

1. Using `SmallSet` for deduplication
2. Passing `DiagnosticSource` as a parameter instead of hardcoding it
3. Simplified the `ConfigError` API 
4. Using TOML's built-in error information instead of regex parsing

For point 5 about the caching strategy - could you clarify the intended behavior? Should the `ConfigFinder` track which config files were recently loaded and only publish diagnostics for those, rather than maintaining a separate cache in the `Server`?

connernilsen:
Hey @sgavriil01, sorry for the long delay. I've had to shift my focus internally for a bit to focus on a bug that's been causing a lot of problems. I think it should be resolved soon, so I'll come back and check out your changes here hopefully this week :)

crates/pyrefly_config/src/config.rs
```diff
@@ -1211,6 +1211,34 @@ impl ConfigFile {
...
+                            .and_then(|s| s.split(&[',', ' ', '\n'][..]).next())
+                            .and_then(|s| s.parse::<usize>().ok())
+                            .unwrap_or(1);
+                        let column = error_str
+                            .split("column ")
+                            .nth(1)
+                            .and_then(|s| s.split(&[',', ' ', '\n'][..]).next())
+                            .and_then(|s| s.parse::<usize>().ok())
+                            .unwrap_or(1);
+                        (line, column)
```
connernilsen, line 1237: Instead of trying to parse this, can we convert the span to a line, col pair? You can probably do something like that with

```rust

```

connernilsen, line 1237: Instead of trying to parse this, can we convert the span to a line, col pair? You can probably do something like this

```rust
  use ruff_source_file::LineIndex;

  let index = LineIndex::from_source_text(source);
  let line_col = index.line_column(text_size_offset, source);
  // line_col.line and line_col.column are both 0-indexed
```

connernilsen (changes requested):
Alright, thanks for making the previous changes. And sorry about the wait. I was pulled into a few internal issues over the past few weeks that took up all of my time.

>For point 5 about the caching strategy - could you clarify the intended behavior? Should the ConfigFinder track which config files were recently loaded and only publish diagnostics for those, rather than maintaining a separate cache in the Server?

What I'm thinking is, we have our [`ConfigFinder` implementation](https://github.com/facebook/pyrefly/blob/50fb37b4c29771d53a936841ab6c53b1b2bbdf65/crates/pyrefly_config/src/finder.rs#L112) that handles loading and caching configs. There are times where a config is re-loaded or we clear the config finder's cache, but it kind of knows the state of all configs at all times, so let's use that as our source of truth.

[Your changes above](https://github.com/facebook/pyrefly/pull/2289/changes#diff-5628399aaa09d6f4d9e8f345149abe4a6b018228cd1020825757c27fb64fc76eR39-R42) removing `anyhow::Error` from `ConfigError` make it so we can add `#[derive(Clone)]` to `ConfigError`, so now we [don't need to take errors from the config finder](https://github.com/facebook/pyrefly/blob/50fb37b4c29771d53a936841ab6c53b1b2bbdf65/crates/pyrefly_config/src/finder.rs#L220), we can just clone them when needed. Those errors will also [automatically be cleared](https://github.com/facebook/pyrefly/blob/50fb37b4c29771d53a936841ab6c53b1b2bbdf65/crates/pyrefly_config/src/finder.rs#L215) when they're not longer useful, so we can just:
1. clear all config errors that currently exist
2. clone all errors from the config finder
3. process the errors to turn them into diagnostics
4. push those diagnostics to the IDE

Then all we need to do is make sure we know which configs we've pushed diagnostics to, and for that, we can continue to use the [`config_files_with_diagnostics`](https://github.com/facebook/pyrefly/pull/2289/changes#diff-22e156bf6a1346c50bb7718b2e4b179ef2245ab9906e32115141b49b32f69999R517) you created, but not have to manage complicated state in the server with `cached_config_errors`.

Sorry, I changed a few details from my first comment mentioning the caching changes, so let me know if this makes more sense.

rchen152:

PR #2291, opened 2026-02-03
fix Support bare class names for in-based narrowing #937

### LINKED ISSUE #937: Support bare class names for `in`-based narrowing
### Describe the Bug

```python
from typing import *

def f(x: type[object]):
    if x in (int, float):
        reveal_type(x) # reveals type[object] in Pyrefly, but `type[int] | type[float]` in Pyright
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQCYCmwUwAFAB4Bcsi5A2mAEYBW5AxjAF0AlAyJRxmKnUwooNVDAA01ADZgAhjBFiJukOQBu5dSoD68BOXpCiQA&version=3.12

_No response_

asukaminato0721:
# Summary

Fixes #937

in-based narrowing treats bare class objects in list/tuple/set literals as narrowable values (via ClassDef -> type[...] and direct type[...]), and added a small class-object equality case for not in.

# Test Plan

Added a regression test covering type[object] narrowed by x in (int, float).

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -762,10 +762,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     let mut literal_types = Vec::new();
                     for expr in exprs {
                         let expr_ty = self.expr_infer(&expr, errors);
-                        if matches!(expr_ty, Type::Literal(_) | Type::None) {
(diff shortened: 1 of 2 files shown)


stroxler (changes requested):

Issue #2224, opened 2026-01-25
fastapi support


asukaminato0721:
### Describe the Bug

It's very popular, can try to add some supports.

https://www.jetbrains.com/help/pycharm/2025.3/fastapi-project.html

asukaminato0721:
https://www.jetbrains.com/help/pycharm/2025.3/flask-endpoints.html ~~this also seems useful.~~

rchen152:
@asukaminato0721 Could you expand on what kind of support you're looking for? From what I understand, FastAPI works well with static type hints, so a lot of things should already work out of the box; is there something specific you're missing?

asukaminato0721:
https://www.jetbrains.com/help/pycharm/2025.3/fastapi-project.html#endpoints

But I am not sure whether this need to be supported.

Otto-AA:
In PyCharm, I frequently use the endpoint plugin to navigate to the definition (when I know the endpoint url in the frontend, it's a fast way to go to the definition in the backend).

Another feature I liked with (iirc) resharper for .net, is url autocompletion and go to definition for test API clients. So when I have a FastAPI TestClient, and start typing `client.get("...`, that I can autocomplete the urls. And afterwards, doing Ctrl+click on the url would get me to the definition of the endpoint.

Both are not necessary, but indeed helpful.

asukaminato0721:
> In PyCharm, I frequently use the endpoint plugin to navigate to the definition (when I know the endpoint url in the frontend, it's a fast way to go to the definition in the backend).
> 
> Another feature I liked with (iirc) resharper for .net, is url autocompletion and go to definition for test API clients. So when I have a FastAPI TestClient, and start typing `client.get("...`, that I can autocomplete the urls. And afterwards, doing Ctrl+click on the url would get me to the definition of the endpoint.
> 
> Both are not necessary, but indeed helpful.

<img width="1147" height="491" alt="Image" src="https://github.com/user-attachments/assets/835ec429-6d61-448d-96ea-6dc3841b2109" />

like this?

Otto-AA:
Similar to this. I am using the search method shown in this Video at 1:50 https://youtu.be/xanrdSKV1k4?t=110

My workflow is:
1. See network request in the browser
2. Press shortcut for search endpoint (not sure about the exact term) in PyCharm
3. Start typing the url
4. Select the appropriate endpoint with arrow keys, press enter to go to the definition

[...]

PR #2299, opened 2026-02-04
fix autocomplete endpoints in web framework #2224

asukaminato0721:
# Summary

Part of #2224.

This adds LSP support for Python web framework endpoints. Pyrefly now recognizes decorator-based route definitions, offers endpoint path completions inside HTTP client calls, and lets go-to-definition jump from a client URL literal to the route handler.

# What changed

- Added endpoint discovery
  - Scans Python ASTs for route decorators such as `@app.get(...)`, `@app.post(...)`, `@app.route(...)`, `@app.api_route(...)`, and `@app.websocket(...)`.
  - Extracts the route path, HTTP method, handler name range, and docstring range.
  - Supports explicit methods from `methods=[...]` / `method=...`.
  - Defaults `route` / `api_route` to `GET` when no method argument is present.

- Added endpoint completions for client URL literals.
  - In calls like `client.get("")`, Pyrefly suggests known matching endpoint paths.
  - Completion results are filtered by HTTP method, so `client.get(...)` and `client.post(...)` surface the right handlers.
  - Insert text handles both inside-string and outside-string completion positions.

- Added go-to-definition for endpoint URL literals.
  - From a call like `client.get("/users")`, go-to-definition jumps to the matching route handler.
  - Path matching understands route parameters, including FastAPI-style `{id}` and Flask-style `<id>` segments.

- Added guardrails to avoid false positives.
  - Endpoint completions/definitions only trigger on recognized HTTP client receivers, such as variables constructed from `TestClient(...)`, `httpx.Client(...)`, `httpx.AsyncClient(...)`, or `requests.Session(...)`.
  - This avoids treating unrelated calls like `dict.get("key")` or arbitrary custom `.get(...)` methods as endpoint requests.

- Added a lightweight performance filter.
  - Before loading and walking a module AST for endpoint discovery, Pyrefly first checks whether the source text contains route-decorator-shaped strings such as `.get(`, `.route(`, or `.api_route(`.
  - This keeps the first implementation simple while avoiding a full AST scan for modules that clearly cannot define endpoints.

# Tests

Added LSP tests covering:

- Endpoint completions from `TestClient` calls.
- Endpoint go-to-definition from URL literals.
- Method-specific endpoint matching.
- Path parameter matching.
- Ignoring non-HTTP `.get(...)` calls such as `dict.get(...)`.

Verified locally with:

- `cargo build`
- `cargo test -p pyrefly endpoint_completion_from_client -- --nocapture`
- `cargo test -p pyrefly endpoint_completion_ignores_non_http_get_calls -- --nocapture`
- `cargo test -p pyrefly endpoint_definition_from_client_literal -- --nocapture`
- `cargo test -p pyrefly endpoint_definition_ignores_non_http_get_calls -- --nocapture`

### DIFF
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -1201,6 +1201,12 @@ impl Transaction<'_> {
                                 in_string_literal,
                             );
                         }
+                        self.add_endpoint_completions(
+                            handle,
+                            mod_module.as_ref(),
+                            position,
+                            &mut result,
+                        );
                         // in foo(x=<>, y=2<>), the first containing node is AnyNodeRef::Arguments(_)
                         // in foo(<>), the first containing node is AnyNodeRef::ExprCall
                         if let Some(first) = nodes.first()
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -1335,6 +1335,12 @@ impl Transaction<'_> {
                                 in_string_literal,
                             );
                         }
+                        self.add_endpoint_completions(
+                            handle,
+                            mod_module.as_ref(),
+                            position,
+                            &mut result,
+                        );
                         // in foo(x=<>, y=2<>), the first containing node is AnyNodeRef::Arguments(_)
                         // in foo(<>), the first containing node is AnyNodeRef::ExprCall
                         if let Some(first) = nodes.first()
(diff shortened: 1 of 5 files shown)


kinto0:
i'm sorry - it's hard to figure out what this PR does from the linked issue. would you be able to explain it in the PR with examples?

NathanTempest:
Hi @asukaminato0721 the PR apparently has failing tests could you address them. Also could you explain what the different changes in this PR do somewhere as it is quite a big change, and we would like to have a brief understanding before diving in.

PR #2310, opened 2026-02-05
impl extract-superclass #2207

asukaminato0721:
# Summary

part of #2207

https://www.jetbrains.com/help/pycharm/extract-superclass.html

Added Extract Superclass refactor to create a new base class in the same file, move the selected members, update the original class bases, and insert pass when the original becomes empty.

Wired the new refactor into LSP code actions.

note: to keep it simple, it does not implement the “Make Abstract” yet.

# Test Plan

Added tests.


stroxler (approved):

PR #2311, opened 2026-02-05
impl Completions: Auto import submodules. #2207

asukaminato0721:
# Summary

part of #2207

Added auto-import suggestions for submodules so completions insert `from <parent> import <submodule>` (and corresponding quickfix entries).

# Test Plan

Added a new LSP completion test and fixtures.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2030,11 +2030,11 @@ impl<'a> Transaction<'a> {
                             &mut import_actions,
                             unknown_name,
                         );
-
                         for module_name in self.search_modules_fuzzy(unknown_name) {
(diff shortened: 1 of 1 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -1969,6 +1969,39 @@ impl<'a> Transaction<'a> {
                             if module_name == handle.module() {
                                 continue;
                             }
+                            if let Some((parent_module_str, submodule_name)) =
```
jvansch1: A lot of this logic seems very similar to the logic added in completion.rs. Can we consolidate this into one shared function that we can re-use?

jvansch1:
Seems like this build is failing for all operating systems. I can take a closer look once that is resolved.

PR #2312, opened 2026-02-05
impl  Completions: Auto import with common alias("import numpy as np") #2207

asukaminato0721:
# Summary

part of #2207

Added a common-alias map and helper lookups for auto-imports.

Auto-import completions and unknown-name quick fixes now honor common aliases (inserting `import <module> as <alias>` and using the alias at the cursor) while avoiding duplicate non-aliased suggestions.

# Test Plan

Added a completion test covering alias import behavior for `spio -> scipy.io`

### DIFF
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -118,11 +118,11 @@ fn assign_sort_text(ranked: &mut RankedCompletion) {
     });
 }
 
-fn autoimport_sort_text(module_name: &str) -> &'static str {
+fn autoimport_source(module_name: &str) -> CompletionSource {
     if module_name.split('.').any(|part| part.starts_with('_')) {
-        SORT_AUTOIMPORT_PRIVATE
+        CompletionSource::AutoimportPrivate
     } else {
-        SORT_AUTOIMPORT_PUBLIC
+        CompletionSource::AutoimportPublic
     }
 }
 
@@ -155,7 +155,7 @@ impl Transaction<'_> {
         module_info: &Module,
         identifier_text: &str,
         supports_completion_item_details: bool,
-        completions: &mut Vec<CompletionItem>,
+        completions: &mut Vec<RankedCompletion>,
     ) -> Option<ModuleName> {
         let module_name_str = common_alias_target_module(identifier_text)?;
         let module_name = ModuleName::from_str(module_name_str);
@@ -170,18 +170,23 @@ impl Transaction<'_> {
             new_text: import_text.clone(),
         };
         let auto_import_label_detail = format!(" (import {module_name_str} as {identifier_text})");
-        completions.push(CompletionItem {
-            label: completion_label.clone(),
-            detail: Some(import_text),
-            kind: Some(CompletionItemKind::MODULE),
-            additional_text_edits: Some(vec![import_text_edit]),
-            label_details: supports_completion_item_details.then_some(CompletionItemLabelDetails {
-                detail: Some(auto_import_label_detail),
-                description: Some(module_name_str.to_owned()),
-            }),
-            insert_text: Some(completion_label),
-            sort_text: Some(autoimport_sort_text(module_name_str).to_owned()),
-            ..Default::default()
+        completions.push(RankedCompletion {
+            item: CompletionItem {
+                label: completion_label.clone(),
+                detail: Some(import_text),
+                kind: Some(CompletionItemKind::MODULE),
+                additional_text_edits: Some(vec![import_text_edit]),
+                label_details: supports_completion_item_details.then_some(
+                    CompletionItemLabelDetails {
+                        detail: Some(auto_import_label_detail),
+                        description: Some(module_name_str.to_owned()),
+                    },
+                ),
+                insert_text: Some(completion_label),
+                ..Default::default()
+            },
+            source: autoimport_source(module_name_str),
+            is_incompatible: false,
         });
         Some(module_name)
     }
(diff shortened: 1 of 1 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -1919,15 +1921,48 @@ impl<'a> Transaction<'a> {
...
+                                    title,
+                                    module_info.dupe(),
+                                    range,
+                                    insert_text,
+                                    false,
+                                    is_private_import,
+                                ));
+                                aliased_modules.insert(module_name);
+                            }
+                        }
```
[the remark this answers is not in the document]
asukaminato0721: this remain to be decided?

pyrefly/lib/lsp/wasm/completion.rs
```diff
@@ -408,8 +409,45 @@ impl Transaction<'_> {
                     SORT_AUTOIMPORT_PUBLIC
                 }
             };
-            for (handle_to_import_from, name, export) in
-                self.search_exports_fuzzy(identifier.as_str())
+            let identifier_text = identifier.as_str();
```
connernilsen, line 560: nit: can we pull the common alias part out into a separate function for readability?

pyrefly/lib/state/ide.rs
```diff
@@ -234,19 +234,50 @@ pub fn insert_import_edit(
     )
 }
 
-/// Insert `import <>` import at the top of the file
+/// Common module aliases used for auto-imports.
+/// These are modeled after Pylance's built-in aliases.
+const COMMON_MODULE_ALIASES: &[(&str, &str)] = &[
```
connernilsen: Nit: `SmallMap`?
```suggestion
static COMMON_MODULE_ALIASES: LazyLock<SmallMap<&str, &str>> = LazyLock::new(|| smallmap! {
```

pyrefly/lib/lsp/wasm/completion.rs
```diff
@@ -408,8 +409,45 @@ impl Transaction<'_> {
...
+            let mut aliased_modules = SmallSet::new();
+            if let Some(module_name_str) = common_alias_target_module(identifier_text) {
+                let module_name = ModuleName::from_str(module_name_str);
+                if module_name != handle.module()
+                    && let Some(module_handle) =
+                        self.import_handle(handle, module_name, None).finding()
+                {
+                    let (position, insert_text, insert_name) =
```
connernilsen: What's the difference from `insert_name` and `insert_text` here?

pyrefly/lib/lsp/wasm/completion.rs
```diff
@@ -408,8 +409,45 @@ impl Transaction<'_> {
...
+                        detail: Some(insert_text),
+                        kind: Some(CompletionItemKind::MODULE),
+                        additional_text_edits: Some(vec![import_text_edit]),
+                        label_details: supports_completion_item_details.then_some(
+                            CompletionItemLabelDetails {
+                                detail: Some(auto_import_label_detail),
+                                description: Some(module_name_str.to_owned()),
+                            },
+                        ),
+                        insert_text: Some(insert_name),
```
connernilsen: I think this part might be a bit confusingly named, since we have something named `insert_text` above, but we're using `insert_name` here instead.

pyrefly/lib/state/lsp.rs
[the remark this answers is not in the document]
connernilsen: @asukaminato0721, do you mean you want to have a discussion if the suggestion is valid?

pyrefly/lib/state/lsp.rs
```diff
@@ -1919,15 +1921,48 @@ impl<'a> Transaction<'a> {
...
+                                && let Some(module_handle) =
+                                    self.import_handle(handle, module_name, None).finding()
+                            {
+                                let (position, insert_text, _) = import_regular_import_edit(
+                                    &ast,
+                                    module_handle,
+                                    Some(unknown_name),
+                                );
+                                let range = TextRange::at(position, TextSize::new(0));
+                                let title = format!("Insert import: `{}`", insert_text.trim());
```
connernilsen: nit: Maybe rename this to suggest you're rewriting it to alias the import?
```suggestion
                                let title = format!("Use common alias: `{}`", insert_text.trim());
```

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -1826,6 +1826,86 @@ Completion Results:
...
+        .as_ref()
+        .expect("auto import completion should include label details");
+    assert_eq!(
+        label_details.detail.as_deref(),
+        Some(" (import numpy as np)")
+    );
+    assert_eq!(label_details.description.as_deref(), Some("numpy"));
+}
+
 #[test]
```
connernilsen, line 1909: Can you add a test for the `code_action` part too? If it's not in this file, there may be an `lsp_interaction/code_action.rs` file you can test this in.

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -1826,6 +1826,86 @@ Completion Results:
...
+        .label_details
+        .as_ref()
+        .expect("auto import completion should include label details");
+    assert_eq!(
+        label_details.detail.as_deref(),
+        Some(" (import numpy as np)")
+    );
+    assert_eq!(label_details.description.as_deref(), Some("numpy"));
+}
+
```
connernilsen, line 1908: If we end up adding checks to make sure the suggested alias imports aren't in user-space code, then we should add a test for that too.

connernilsen (changes requested):
Looks good, just a few small things I'd like to see addressed before pulling this in. Let me know if you disagree with anything, down to discuss or change my mind.

connernilsen:
Also adding @kinto0, since he may have more comments than I do.

pyrefly/lib/state/lsp.rs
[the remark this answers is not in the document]
kinto0: as long as there are tests I don't mind either behavior

connernilsen:
I'll pull it in and see what people think

connernilsen:
Looks like GH is down ;(

connernilsen:
Hey @asukaminato0721, looks like there are a few merge conflicts. Would you be able to resolve them when you get a chance? It looks like GH is working again, so I'll pull the changes in afterward.

kinto0 (approved):

PR #2315, opened 2026-02-05
fix [lsp][completion] rank results based on compatibility in context #1653

### LINKED ISSUE #1653: [lsp][completion] rank results based on compatibility in context
### Describe the Bug

in this example:

Image

pyrefly knows that XB is incompatible there. pyrefly should rank it beneath the other suggestions which are compatible

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1653

Implemented compatibility-aware completion ordering for call-argument contexts by computing the expected argument type and demoting incompatible local/attribute suggestions

# Test Plan

Added an LSP interaction test that asserts compatible suggestions are ranked ahead of incompatible ones

### DIFF
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -2631,11 +2631,12 @@ class Other:
 def needs_base(x: Base) -> None:
     pass
 
-exact_val = Base()
-compatible_val = Derived()
-incompatible_val = Other()
+val_exact = Base()
+val_compatible = Derived()
+val_incompatible = Other()
+val = val_exact
 
-needs_base(v)
+needs_base(val)
 #           ^
 "#;
     let (handles, state) = mk_multi_file_state(&[("main", code)], Require::Exports, true);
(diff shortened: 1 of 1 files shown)


pyrefly/lib/test/lsp/lsp_interaction/completion.rs
```diff
@@ -274,6 +274,50 @@ fn test_completion_keywords() {
     interaction.shutdown().unwrap();
 }
 
+#[test]
+fn test_completion_sorts_incompatible_call_argument_last() {
```
kinto0: this doesn't need to be an lsp_interaction test, does it? it can just go in lsp/completion.rs?

kinto0: while you're at it, a test for exact types, compatible types, and incompatible types would be appreciated

pyrefly/lib/test/lsp/lsp_interaction/completion.rs
```diff
@@ -274,7 +274,6 @@ fn test_completion_keywords() {
     interaction.shutdown().unwrap();
 }
 
-#[test]
```
kinto0, line 277: nit: i'll fix this on import

PR #2317, opened 2026-02-05
fix show diff in signature #2154

### LINKED ISSUE #2154: show diff in signature
### Describe the Bug

```py
from abc import ABC
class A():
    def foo(self, a: int, b:int, c:int):
        raise NotImplementedError()

class B(A):
    def foo(self):
        x = 1
        print(x)
```

```
ERROR sandbox.py:15:9-12: Class member `B.foo` overrides parent class `A` in an inconsistent manner [[bad-override](https://pyrefly.org/en/docs/error-kinds/#bad-override)]
  `B.foo` has type `BoundMethod[B, (self: B) -> None]`, which is not assignable to `BoundMethod[B, (self: B, a: int, b: int, c: int) -> Never]`, the type of `A.foo`
```

I want the error message to be

```
def foo(self, a: int, b:int, c:int):
               ------------------
                              |
                               -----------------------these not match
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2154

Added ASCII-style signature diff rendering (ruff_annotate_snippets) to bad-override errors when both sides are single-signature callables. This yields a rustc-like caret/underline block beneath the two signatures.

# Test Plan

add snapshot test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1324,7 +1324,7 @@ fn has_any_abstract(ty: &Type) -> bool {
 fn signature_parts(sig: &str) -> Option<(std::ops::Range<usize>, std::ops::Range<usize>)> {
     let open = sig.find('(')?;
     let close = sig[open..].find(')')? + open;
-    if close <= open + 1 {
+    if close < open + 1 {
         return None;
     }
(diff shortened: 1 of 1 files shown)


asukaminato0721:
the style is my prototype plus ruff's.

PR #2322, opened 2026-02-05
if *args: Any and **kwargs: Any are present, treat signature like `...` for assignability purposes

yangdanny97:
Summary:
This ignores any other parameters (though they are preserved when we eventually call the function)

Pyright accepts this snippet:

```
from typing import override, Any

class Base:
    def method1(self, a: int, /, *args: Any, k: int, **kwargs: Any) -> None: ...

class Derived(Base):
    override
    def method1(self) -> None: ...

class Base2:
    def method1(self) -> None: ...

class Derived2(Base2):
    override
    def method1(self, a: int, /, *args: Any, k: int, **kwargs: Any) -> None: ...
```

Spec says: https://typing.python.org/en/latest/spec/callables.html#meaning-of-in-callable

> If the input signature in a function definition includes both a *args and **kwargs parameter and both are typed as Any (explicitly or implicitly because it has no annotation), a type checker should treat this as the equivalent of .... Any other parameters in the signature are unaffected and are retained as part of the signature

Differential Revision: D89676164

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -2094,28 +2094,6 @@
     },
     {
       "code": -2,
-      "column": 20,
-      "concise_description": "`Proto7` is not assignable to `Proto6`",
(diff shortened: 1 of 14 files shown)


rchen152 (approved):

PR #2324, opened 2026-02-05
impl code action to remove Redundant cast #2316

### LINKED ISSUE #2316: code action to remove Redundant cast
### Describe the Bug

this require type info so ruff can't do it. 

```
WARN Redundant cast: `int` is the same type as `int` [redundant-cast]
   --> core/ops/aliyun_trace/data_exporter/traceclient.py:204:20
    |
204 |         return cast(int, uuid_obj.int)
    |                    -------------------
    |
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2316

Added a quick fix that removes redundant typing.cast(...) by replacing the call with the value expression (with parentheses when needed)

# Test Plan

Added a code action regression test for the redundant cast quick fix.

### DIFF
--- a/pyrefly/lib/test/lsp/code_actions.rs
+++ b/pyrefly/lib/test/lsp/code_actions.rs
@@ -843,6 +843,18 @@ fn redundant_cast_cursor_inside_args() {
 }
 
 #[test]
+fn redundant_cast_preserves_multiplication_precedence() {
+    let code =
+        "from typing import cast\nx: int = 1\ny: int = 2\nz: int = 3\nx * cast(int, y + z)\n";
+    let cursor_offset = code.find("cast(").unwrap();
+    let after = redundant_cast_action_after(code, cursor_offset).unwrap();
+    assert_eq!(
+        "from typing import cast\nx: int = 1\ny: int = 2\nz: int = 3\nx * (y + z)\n",
+        after
+    );
+}
+
+#[test]
 fn test_import_from_stdlib() {
     let report = get_batched_lsp_operations_report_allow_error(
         &[("a", "TypeVar('T')\n# ^")],


asukaminato0721:
Although for simplicity I don't impl whole project fix, but find -exec can do it so it can work in ci.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -3298,6 +3297,32 @@ impl Server {
...
+                    continue;
+                };
+                changes.entry(lsp_location.uri).or_default().push(TextEdit {
+                    range: lsp_location.range,
+                    new_text,
+                });
+            }
+            if !changes.is_empty() {
+                actions.push(CodeActionOrCommand::CodeAction(CodeAction {
+                    title: "Fix all redundant casts".to_owned(),
```
yangdanny97: "Remove all redundant casts"?

pyrefly/lib/state/lsp/quick_fixes/redundant_cast.rs
```diff
@@ -0,0 +1,128 @@
...
+            | Expr::NoneLiteral(_)
+            | Expr::EllipsisLiteral(_)
+            | Expr::Subscript(_)
+            | Expr::Attribute(_)
+            | Expr::Call(_)
+            | Expr::List(_)
+            | Expr::Dict(_)
+            | Expr::Set(_)
+            | Expr::Tuple(_)
+            | Expr::FString(_)
```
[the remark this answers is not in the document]
yangdanny97: I wonder if we want to make this a shared util in the future. Introduce parameter also has a similar concept: https://github.com/facebook/pyrefly/blob/6de02a92804cd8caa41aef281ebed2e44b28da1f/pyrefly/lib/state/lsp/quick_fixes/introduce_parameter.rs#L650

pyrefly/lib/state/lsp/quick_fixes/redundant_cast.rs
```diff
@@ -0,0 +1,128 @@
...
+
+fn wrap_if_needed(expr: &Expr, text: &str) -> String {
+    if expr_needs_parens(expr) {
+        format!("({text})")
+    } else {
+        text.to_owned()
+    }
+}
+
+fn find_enclosing_call(ast: &ModModule, selection: TextRange) -> Option<ExprCall> {
```
yangdanny97, line 20: Can we use this helper? 

https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_python/src/ast.rs#L328

pyrefly/lib/state/lsp/quick_fixes/inline_method.rs
```diff
@@ -247,36 +249,6 @@ fn find_method_def_for_self_call(
 /// Simple expressions (literals, names, subscripts, attributes, calls) don't need
 /// parentheses because they have high precedence. Complex expressions (binary ops,
 /// unary ops, comparisons, etc.) need parentheses to preserve semantics.
-fn expr_needs_parens(expr: &Expr) -> bool {
```
jvansch1, line 250: It looks like you forgot to remove this docstring. Now it looks like this docstring is for `build_param_map`

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -3362,6 +3363,32 @@ impl Server {
...
+            if !changes.is_empty() {
+                actions.push(CodeActionOrCommand::CodeAction(CodeAction {
+                    title: "Remove all redundant casts".to_owned(),
+                    kind: Some(CodeActionKind::SOURCE_FIX_ALL),
+                    edit: Some(WorkspaceEdit {
+                        changes: Some(changes),
+                        ..Default::default()
+                    }),
+                    ..Default::default()
+                }));
```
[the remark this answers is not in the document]
jvansch1: This is something that we need to fix. `source.fixAll ` actions are commonly consumed by editors on save, where the trigger kind is `AUTOMATIC` and the only filter specifies `source.fixAll`. With the current placement, the fix-all code action will never be returned in that scenario because the function returns early

pyrefly/lib/state/lsp.rs
```diff
@@ -1991,12 +1992,25 @@ impl<'a> Transaction<'a> {
...
+                        if error_range.contains_range(range) || call_range.contains_range(range) {
+                            other_actions.push(action);
+                        }
+                    }
+                }
                 _ => {}
             }
         }
 
         // Sort code actions: non-private first, then non-deprecated, then alphabetically
```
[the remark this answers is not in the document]
jvansch1: Lets apply this.

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -624,6 +624,62 @@ my_export
     );
 }
 
+#[test]
+fn redundant_cast_quickfix() {
```
jvansch1, line 734: Can we also add tests for the following scenarios?

- A casted expression which could use parenthsis: `cast(int, a + b)`
- A cast nested within another call: `print(cast(a + b))`
- A call where the cursor is not directly on the cast statement. Maybe we can see what happens the the parenthesis.

jvansch1 (changes requested):
Can we also add tests to make sure this works for the foll

jvansch1:
@asukaminato0721 We are getting close on this one. I have a a concern about a case like `x * cast(int, y + z)`. If we remove the cast this should become `x * (y + z)`. However, I want to ensure that we are not accidentally removing the parenthesis and making the statement actually `x * y + z` since this has a very different meaning. Can you add a test which shows the we properly preserves the parenthesis in this case?

PR #2330, opened 2026-02-06
fix Exception message for incorrect dict literal is too wide #893

### LINKED ISSUE #893: Exception message for incorrect dict literal is too wide
### Describe the Bug

Given:

```python
mp: dict[int, int] = {
    1: 2,
    3: "test"
}
```

This tells me:

```
`dict[int, int | str]` is not assignable to `dict[int, int]`
```

But the error message is reported as the whole dict literal in `{ brackets }`. Would be nice if it told me just about the one that is wrong.

### Sandbox Link

https://pyrefly.org/sandbox/?code=LYBwXABAJglgxgFwNowHYIDQTQguhAXggG8AoCCiARkgCYNzKBmSAIgQFMBnBV0gXyA&version=3.12

_No response_

asukaminato0721:
# Summary

Fixes #893

Narrowed dict-literal assignment errors to the first mismatching key/value/unpack when a dict[...] hint is present, so the caret lands on the wrong entry instead of the whole { ... }.

to decide: only show first error for now. Maybe show full error? But will it too noisy?

# Test Plan

Added a unit test that asserts the error range points at the offending value literal.

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1545,49 +1545,63 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             let mut key_tys = Vec::new();
             let mut value_tys = Vec::new();
             let mut has_type_mismatch = false;
+            let check_items = hint
+                .filter(|hint| hint.errors().is_some())
+                .is_some_and(|hint| hint.types().len() == 1);
+            let infer_with_hint = |expr: &Expr, hint: Option<HintRef>, errors: &ErrorCollector| {
+                let ty = self.expr_infer_with_hint(expr, hint, errors);
+                match hint {
+                    Some(hint) => {
+                        let want = match hint.types() {
+                            [hint] => hint.clone(),
+                            hints => Type::union(hints.to_vec()),
+                        };
+                        let matches_hint = self.is_subset_eq(&ty, &want);
+                        let ty = if matches_hint {
+                            want
+                        } else {
+                            ty.promote_implicit_literals(self.stdlib)
+                        };
+                        (ty, matches_hint)
+                    }
+                    None => (ty.promote_implicit_literals(self.stdlib), true),
+                }
+            };
             items.iter().for_each(|x| match &x.key {
                 Some(key) => {
-                    let key_t = self.expr_infer_with_hint_promote(
-                        key,
-                        key_hint.as_ref().and_then(|key_hint| {
-                            hint.as_ref()
-                                .map(|hint| HintRef::new(key_hint, hint.errors()))
-                        }),
-                        errors,
-                    );
-                    let value_t = self.expr_infer_with_hint_promote(
-                        &x.value,
-                        value_hint.as_ref().and_then(|value_hint| {
-                            hint.as_ref()
-                                .map(|hint| HintRef::new(value_hint, hint.errors()))
-                        }),
-                        errors,
-                    );
-                    if let Some(hint) = hint
+                    let key_hint_ref = key_hint.as_ref().and_then(|key_hint| {
+                        hint.as_ref()
+                            .map(|hint| HintRef::new(key_hint, hint.errors()))
+                    });
+                    let value_hint_ref = value_hint.as_ref().and_then(|value_hint| {
+                        hint.as_ref()
+                            .map(|hint| HintRef::new(value_hint, hint.errors()))
+                    });
+                    let (key_t, key_matches_hint) = infer_with_hint(key, key_hint_ref, errors);
+                    let (value_t, value_matches_hint) =
+                        infer_with_hint(&x.value, value_hint_ref, errors);
+                    if check_items
+                        && let Some(hint) = hint
                         && let Some(check_errors) = hint.errors()
                     {
                         let tcc: &dyn Fn() -> TypeCheckContext =
-                            &|| TypeCheckContext::of_kind(TypeCheckKind::AnnAssign);
+                            &|| TypeCheckContext::of_kind(TypeCheckKind::DictLiteralItem);
                         if let Some(key_hint) = &key_hint
-                            && !self.check_type(
-                                &key_t,
-                                key_hint,
-                                key.range(),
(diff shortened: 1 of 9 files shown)


asukaminato0721:
```
ERROR `dict[int | str, int | str]` is not assignable to `dict[str, str]` [bad-assignment]
 --> /tmp/tmp/main.py:3:5
  |
3 |     1: "2",
  |     ^
  |
 INFO 1 error
```

rchen152 (changes requested):
Do you know why the mypy_primer results changed? That seems concerning to me; I wouldn't expect an improvement to an error message to affect what errors are reported.

asukaminato0721:
In dict_items_infer we call check_type on each key/value and record those errors on the item ranges. That can increase the count and change message text

For example, this shows the range is smaller.

```diff
- ERROR static_frame/test/unit/test_frame.py:1007:43-62
+ ERROR static_frame/test/unit/test_frame.py:1007:44-45
```


In multiline dicts, the error now lands on the offending key/value line, not on the assignment line. Any # type: ignore or pyrefly suppression on the assignment line no longer applies, so extra errors appear.

pyrefly/lib/alt/expr.rs
```diff
@@ -343,12 +343,77 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                let range = self
+                    .dict_literal_error_range(x, hint)
+                    .unwrap_or_else(|| x.range());
+                self.check_and_return_type_info(got, hint, range, hint_errors, tcc)
             }
             _ => self.expr_infer_type_info_with_hint(x, None, errors),
         }
     }
 
+    fn dict_literal_error_range(&self, x: &Expr, hint: &Type) -> Option<TextRange> {
```
rchen152: This basically seems to be redoing the check of `x` against its hint in order to get the right TextRange? That seems potentially expensive, and it's also duplicating a fair bit of code, which seems like a bit of a maintainability risk.

Is there any way to report an error in dict_items_infer (https://github.com/facebook/pyrefly/blob/b00bc92c57012a8044b5766ac7a26665fa2fca94/pyrefly/lib/alt/expr.rs#L915) right when we find a mismatch between a key/value and its hint?

rchen152 (changes requested):
Sorry for the delay in getting you a more in-depth review of this PR. The mypy_primer change being caused by the error range changing makes sense, thanks.

asukaminato0721:
well, this is a bit ... verbose but more precise.

pyrefly/lib/alt/expr.rs
asukaminato0721: done, now mypy_primer report all the single errors.

pyrefly/lib/alt/expr.rs
```diff
@@ -961,6 +963,34 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         value_hint.as_ref().map(|hint| hint.as_ref()),
                         errors,
                     );
+                    if let Some(hint) = hint
```
rchen152: expr_infer_with_hint_promote above already does an `is_subset_eq` check using the hint, and `check_type` is doing a second one here. Could we reuse that first result rather than doing the same check again?

pyrefly/lib/alt/expr.rs
```diff
@@ -961,6 +963,34 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                        && let Some(check_errors) = hint.errors()
+                    {
+                        let tcc: &dyn Fn() -> TypeCheckContext =
+                            &|| TypeCheckContext::of_kind(TypeCheckKind::AnnAssign);
```
rchen152: AnnAssign doesn't seem like the right type check kind for this. If there doesn't exist an appropriate one, could you add a new variant to TypeCheckKind?

pyrefly/lib/alt/expr.rs
```diff
@@ -996,6 +1026,34 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 None => {
                     let ty = self.expr_infer(&x.value, errors);
                     if let Some((key_t, value_t)) = self.unwrap_mapping(&ty) {
+                        if let Some(hint) = hint
+                            && let Some(check_errors) = hint.errors()
+                        {
+                            let tcc: &dyn Fn() -> TypeCheckContext =
+                                &|| TypeCheckContext::of_kind(TypeCheckKind::AnnAssign);
+                            if let Some(key_hint) = &key_hint
+                                && !self.check_type(
```
rchen152: Similar comment to above, this seems to be duplicating work done by `is_subset_eq` at line 1059.

rchen152 (changes requested):
Thanks! This looks like a more promising direction.

Some of the new errors in mypy_primer look like false positives. For example, there's a new error at https://github.com/scrapy/scrapy/blob/ccfa052fa19f712355fb17b863e8ff77f34ff3ac/scrapy/downloadermiddlewares/cookies.py#L188C1-L188C76 but no corresponding old error or `# type: ignore`. Looking at the signature of `Response.__init__`: https://github.com/scrapy/scrapy/blob/ccfa052fa19f712355fb17b863e8ff77f34ff3ac/scrapy/http/response/__init__.py#L63, we seem to be matching the `headers` argument against `Iterable[tuple[AnyStr, Any]]` and not considering the other options in the union.

I'm suspicious that this may be due to a preexisting bug in how union hints are decomposed (https://github.com/facebook/pyrefly/issues/793). Maybe we need to exclude union hints from this new logic until that bug has been resolved.

rchen152:
After digging into this a bit more, I realized that there are several pre-existing problems with how we handle contextual hints during inference, and fixing those first allows for a much simpler fix for this issue. https://github.com/facebook/pyrefly/pull/4275 is the new PR I created - it contains some fixes to `infer_with_decomposed_hint` and `expr_infer_with_hint_promote`, plus a new fix for #893, which now just consists of passing a type checking context to `expr_infer_with_hint_promote`.

Issue #2313, opened 2026-02-05
do check/autocompletion on unittest patch string


asukaminato0721:
### Describe the Bug

unittest patch use string to patch obj.

When doing refactor, I rename or reorder the obj, I will need to know where to update the string.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

example

```
def foo():
    ...

@patch(...)
def test_...
```

when I rename foo -> bar, I want the patch can show error message.

connernilsen:
This seems like an interesting feature to have! I'm not sure how much support we can offer, given the dynamism involved, but definitely worth looking into.

PR #2333, opened 2026-02-06
impl string check for patch #2313

asukaminato0721:
# Summary

Fixes part of #2313

Added a call-time check when the callee is unittest.mock.patch (detected via the _patcher instance type):

finds the longest importable module prefix

resolves the remaining components as attributes

emits the usual missing import / missing attribute errors at the string literal location

# Test Plan

Added an integration test covering the missing-attribute case: with patch("main.foo") now reports

No attribute \foo` in module `main``


pyrefly/lib/alt/call.rs
```diff
@@ -2353,6 +2359,145 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            }
+        }
+
+        let Some(module) = module else {
+            // If we can't resolve the module (or it is present but untyped/ignored), skip patch
+            // validation to avoid spurious failures when patching third-party libraries.
+            // This keeps the check focused on modules Pyrefly can actually analyze.
+            return;
+        };
+        if module != self.module().name() {
```
rchen152: IMO this severely limits the usefulness of this PR. Tests typically patch the code under test, not their own code. I'm hesitant to add this analysis with this limitation.

rchen152 (changes requested):

pyrefly/lib/alt/call.rs
```diff
@@ -2431,6 +2437,137 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            let swallower = self.error_swallower();
+            base_ty = self.type_of_attr_get(
+                &base_ty,
+                &attr_name,
+                range,
+                &swallower,
+                None,
+                "unittest.mock.patch target",
+            );
+            if !swallower.is_empty() {
```
rchen152: This silently discards all lookup errors after the target traversal leaves the module portion. As a result, patch("mod.C.missing") does not report an error even when C is known and definitely has no missing attribute. Class and instance attributes are common patch targets, so this substantially limits the analysis.

pyrefly/lib/alt/call.rs
```diff
@@ -2431,6 +2437,137 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            // validation to avoid spurious failures when patching third-party libraries.
+            return;
+        };
+
+        let mut base_ty = ModuleType::new_as(module).to_type(self.heap);
+        for attr in &parts[module_prefix_len..] {
+            let attr_name = Name::new(*attr);
+            if let Type::Module(module_type) = &base_ty {
+                let module_name = ModuleName::from_parts(module_type.parts());
+                if attr_name.as_str().starts_with('_')
```
rchen152: Why skip all private attributes? Private names are common patch targets, and privacy does not affect runtime attribute lookup.

pyrefly/lib/alt/call.rs
```diff
@@ -2431,6 +2437,137 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                    || self
```
rchen152: Why does this exception exist? Builtin names are not automatically attributes of every module.

pyrefly/lib/alt/call.rs
```diff
@@ -2431,6 +2437,137 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                            range,
+                            ErrorKind::MissingAttribute,
+                            format!("No attribute `{attr_name}` in module `{module_name}`"),
+                        )
+                        .emit();
+                    return;
+                }
+            }
+            let swallower = self.error_swallower();
+            base_ty = self.type_of_attr_get(
```
rchen152, line 2568: `type_of_attr_get` already provides the normal semantic lookup path for modules, submodules, module `__getattr__`, and class or instance attributes. This code currently duplicates a narrower module-only lookup, then calls `type_of_attr_get` with an `ErrorSwallower`, which is why nested missing attributes disappear. My suggestion would be to find the longest importable module prefix, then traverse every remaining component with `type_of_attr_get` using the real error collector. That should remove most of this custom logic as well as the private-name and builtin-name special cases.

pyrefly/lib/alt/call.rs
```diff
@@ -2431,6 +2437,137 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            _ => false,
+        }
+    }
+
+    fn unittest_mock_patch_target_expr<'b>(arguments: &'b Arguments) -> Option<&'b Expr> {
+        match arguments.args.first() {
+            Some(Expr::Starred(_)) | None => {}
+            Some(arg) => return Some(arg),
+        }
+        for kw in &arguments.keywords {
```
rchen152: Arguments::find_keyword can do this for you.

rchen152 (changes requested):
Thanks for addressing the same-module limitation! Here's a more in-depth review. Tl;dr I think you can reuse existing logic more heavily to simplify the code, and I had questions about the carveouts for private attributes, builtins etc.

rchen152 (approved):
Thanks! This looks good to me and I've put it up for internal review. I realized while reviewing the mypy_primer diff that `unittest.mock.patch` indeed automatically sets `create=True` for mock builtins, so you were right to omit builtins from the check. Sorry for leading you astray there! I'll stack a commit to fix this up, so no action needed from you.

grievejia (approved):

rchen152:
This is landing now. FYI we've decided to make this diagnostic a new sub-kind of missing attribute, which will be set to`warn` (not `error`) by default.

PR #2338, opened 2026-02-06
impl MRU Ranked suggestions.

asukaminato0721:
# Summary

Fixes part of #2207

Reworked MRU to match Pyright: in-memory MRU list (max 128), no disk persistence, recorded via standard completionItem/resolve and applied by rewriting sort_text with MRU rank and preselect for the most recent item.

MRU now uses LSP resolve and works with any client that calls resolve.

# Test Plan

Added an LSP interaction test covering MRU ordering across sessions.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2758,7 +2758,7 @@ impl<'a> Transaction<'a> {
         position: TextSize,
         import_format: ImportFormat,
         options: CompletionOptions,
-        mut mru_index: Option<F>,
+        mru_index: Option<F>,
     ) -> (Vec<CompletionItem>, bool)
     where
         F: FnMut(&CompletionItem) -> Option<usize>,
@@ -2771,16 +2771,13 @@ impl<'a> Transaction<'a> {
             }
         }
 
-        let (mut results, is_incomplete) =
-            self.completion_sorted_opt_with_incomplete(handle, position, import_format, options);
-        if let Some(ref mut mru_index) = mru_index {
-            for item in &mut results {
-                let rank = mru_index(item).unwrap_or(9999).min(9999);
-                let base_sort_text = item.sort_text.as_deref().unwrap_or("0");
-                item.sort_text = Some(format!("{base_sort_text}.{rank:04}.{}", item.label));
-                item.preselect = if rank == 0 { Some(true) } else { None };
-            }
-        }
+        let (mut results, is_incomplete) = self.completion_sorted_opt_with_incomplete(
+            handle,
+            position,
+            import_format,
+            options,
+            mru_index,
+        );
         results.sort_by(|item1, item2| {
             item1
                 .sort_text
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -79,39 +79,45 @@ impl RankedCompletion {
 }
 
 /// All completion ranking logic lives here. Assigns `sort_text` to each item
-/// based on its source classification, name prefix, and compatibility.
-fn assign_sort_text(ranked: &mut RankedCompletion) {
+/// based on its source classification, name prefix, compatibility, and MRU rank.
+fn assign_sort_text(ranked: &mut RankedCompletion, mru_rank: Option<usize>) {
     let is_deprecated = ranked
         .item
         .tags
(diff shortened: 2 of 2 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1063,6 +1079,124 @@ impl Server {
...
+        mru.prune(now);
+        let mru_snapshot = mru.clone();
+        drop(mru_map);
+        let path = workspace_mru_path(&workspace_root);
+        if let Err(err) = save_to_path(&mru_snapshot, &path) {
+            info!("Failed to save completion MRU data: {err}");
+        }
+    }
+
+    fn apply_completion_mru(&self, uri: &Url, items: &mut Vec<CompletionItem>) {
```
kinto0: should this be in server.rs? or can it be in completions.rs?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1063,6 +1079,124 @@ impl Server {
...
+    }
+
+    fn now_epoch_secs() -> u64 {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs()
+    }
+
+    fn completion_item_mru_key(item: &CompletionItem) -> Option<String> {
```
kinto0: move to completion.rs?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -525,10 +534,10 @@ pub struct Server {
     /// operations we have yet to process.
     uris_pending_close: Mutex<HashMap<String, usize>>,
     workspaces: Arc<Workspaces>,
```
kinto0, line 538: if completion_mru is workspace scoped, why not put it in the same workspaces struct?

pyrefly/lib/lsp/non_wasm/mru.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Workspace-scoped MRU tracking for completion items.
+#[derive(Clone, Debug)]
+pub struct CompletionMru {
+    entries: HashMap<String, MruEntry>,
+    max_entries: usize,
+    max_age_days: u64,
+}
+
+impl CompletionMru {
```
kinto0, line 27: can you add a comment describing this? why write to disk? why not just store in memory?

lsp/src/extension.ts
```diff
@@ -44,6 +45,63 @@ function requireSetting<T>(path: string): T {
   return ret;
 }
 
+function completionItemPayload(item: vscode.CompletionItem): object {
```
kinto0: is there a way to use a json encoding method on `CompletionItem` to avoid this manual encoding?

lsp/src/extension.ts
```diff
@@ -180,6 +248,21 @@ export async function activate(context: ExtensionContext) {
     clientOptions,
   );
 
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
```
kinto0: did you consider reordering completions in the extension instead of adding a new method to the protocol? since it only works in this extension, is there a reason we need the language server to know about this?

asukaminato0721:
find it https://github.com/microsoft/pyright/blob/5a325e4874e775436671eed65ad696787a1ef74b/packages/pyright-internal/src/languageService/completionProvider.ts#L277

pyrefly/lib/state/lsp.rs
```diff
@@ -2691,6 +2731,14 @@ impl<'a> Transaction<'a> {
 
         let (mut results, is_incomplete) =
             self.completion_sorted_opt_with_incomplete(handle, position, import_format, options);
+        if let Some(ref mut mru_index) = mru_index {
+            for item in &mut results {
+                let rank = mru_index(item).unwrap_or(9999).min(9999);
+                let base_sort_text = item.sort_text.as_deref().unwrap_or("0");
+                item.sort_text = Some(format!("{base_sort_text}.{rank:04}.{}", item.label));
+                item.preselect = if rank == 0 { Some(true) } else { None };
```
kinto0: I still think it'd be better if we keep the completions consolidated in assign_sort_text. what do you think?

kinto0 (changes requested):

PR #2340, opened 2026-02-06
fix Odd warning on shutil.rmtree and ExitStack.callback #2105

### LINKED ISSUE #2105: ParamSpec pins to the first overload incorrectly
### Describe the Bug

Here's the example code:

```python
import shutil
from contextlib import ExitStack


def foo(tmpdir):
    with ExitStack() as resources:
        resources.callback(shutil.rmtree, tmpdir, ignore_errors=True)


def bar(tmpdir):
    shutil.rmtree(tmpdir, ignore_errors=True)
```

and the output of `pyrefly check example.py`:

```
ERROR Missing argument `onerror` in function `contextlib._BaseExitStack.callback` [missing-argument]
 --> example.py:7:27
  |
7 |         resources.callback(shutil.rmtree, tmpdir, ignore_errors=True)
  |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
 INFO 1 error
```

What's odd to me is that pyrefly complains about `onerror` (which clearly is a [default argument](https://docs.python.org/3/library/shutil.html#shutil.rmtree)), but only when called through `ExitStack.callback`.  When called explicitly, pyrefly (correctly, IMHO) accepts the call with the missing `onerror` argument.

```
% pyrefly --version
pyrefly 0.48.0
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2105

Fixed ParamSpec inference for forwarded callbacks to select the overload that matches the forwarded arguments, even when the callback is a bound `__call__` (protocol/class instance).

# Test Plan

Added a regression test for ExitStack.callback(shutil.rmtree, ..., ignore_errors=True).

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -581,6 +581,116 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
         }
         let mut expected_types: HashMap<TextRange, Type> = HashMap::new();
+        if let Some(paramspec_var) = paramspec {
+            // If we're forwarding args to a callable ParamSpec, prefer an overload that
+            // matches those forwarded arguments instead of pinning P to the first overload.
+            let positional_params = params
+                .items()
+                .iter()
+                .enumerate()
+                .filter_map(|(idx, param)| match param {
+                    Param::PosOnly(_, ty, _) | Param::Pos(_, ty, _) => Some((idx, ty)),
+                    _ => None,
+                })
+                .collect::<Vec<_>>();
+            if let Some((callable_param_idx, _param_ty)) =
+                positional_params.iter().rev().find(|(_, ty)| {
+                    let param_callable = match **ty {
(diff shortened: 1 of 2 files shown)


rchen152 (changes requested):
At least conceptually, this seems like another flavor of https://github.com/facebook/pyrefly/issues/105 - we're matching the paramspec to the first overload rather than considering all of them.

I'd be okay with putting in a workaround to address this specific bug, but this PR makes me a bit nervous because (1) it's doing some fiddly tracking of parameters and arguments by idx, and (2) it's doing its own overload evaluation, using different logic from `call_overloads`.

As a more general question, I wonder how pervasive/urgent this issue is. Maybe if we wait until we've addressed some of our problems with type parameter solving more holistically, a better solution will become more obvious.

rchen152:
Hey @asukaminato0721! I'm going to close this PR since Steven is working on a more comprehensive fix for https://github.com/facebook/pyrefly/issues/2105.

PR #2347, opened 2026-02-06
cap literal string size

yangdanny97:
Differential Revision: D92565631


stroxler (approved):

PR #2349, opened 2026-02-06
cap literal union size

yangdanny97:
Differential Revision: D92570874


stroxler (approved):

PR #2351, opened 2026-02-07
promote anonymous typed dicts when unioning

yangdanny97:
Differential Revision: D92586637


samwgoldman (approved):

PR #2353, opened 2026-02-07
fix [feature] error when == or != is used on values that are incompatible types #721

### LINKED ISSUE #721: [feature] error when `==` or `!=` is used on values that are incompatible types
### Describe the Bug

If someone uses `==` and `!=` on two values that are completely incompatible types, it will likely work at runtime but is almost certainly a bug.

```
x: int = 1
y: str = ""

if x == y:
  pass
```

We should give an error in such cases, with a custom error code so that it can be disabled if necessary.

I was thinking we could just check that `is_subset_eq(x, y)` or `is_subset_eq(y, x)` are both false, but there's probably some nuance here in how we want to handle unions.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #721

Added a new incompatible-comparison diagnostic for ==/!= when operand types cannot overlap
Only triggers for built‑in scalar families (numeric, bytes-like, str) and only when the families are disjoint.
Skips user types, TypeVar, and type[...] comparisons.

add `if let guard` for now since or else it will break the compile..

# Test Plan

add tests

### DIFF
--- a/website/docs/error-kinds.mdx
+++ b/website/docs/error-kinds.mdx
@@ -655,8 +655,10 @@ def project[S](func: Callable[[S], S], y: S) -> Callable[[], S]:
 project(f, 1)  # Overload type was not compatible with solved type variables: S = int
 ## incompatible-comparison
 
-This error is raised when an equality (`==`) or inequality (`!=`) comparison is
-made between types that cannot overlap. This is almost always a bug.
+This error is raised when Pyrefly can prove that an equality (`==`) or inequality
+(`!=`) comparison is made between incompatible built-in scalar types (for example,
+`int` versus `str`). This check is currently limited to primitive scalar types and
+does not analyze unions, `None`, or user-defined classes.
 
 ```python
 x: int = 1
--- a/website/docs/error-kinds.mdx
+++ b/website/docs/error-kinds.mdx
@@ -653,6 +653,15 @@ def project[S](func: Callable[[S], S], y: S) -> Callable[[], S]:
     return lambda: y
 
 project(f, 1)  # Overload type was not compatible with solved type variables: S = int
+## incompatible-comparison
+
+This error is raised when an equality (`==`) or inequality (`!=`) comparison is
+made between types that cannot overlap. This is almost always a bug.
+
+```python
+x: int = 1
+y: str = ""
+if x == y: ...  # Comparison `==` between incompatible types `int` and `str` [incompatible-comparison]
 ```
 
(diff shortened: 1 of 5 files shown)


kinto0:
this seems contreversial... I imagine we want to have this default to off? not sure. let's discuss in discord

yangdanny97:
I wonder if we can use disjoint bases for this? there's a disjoint_base helper already in narrow.rs https://peps.python.org/pep-0800/#specification

NathanTempest:
Thanks @asukaminato0721 for your continued effort. After going through this, this PR has a few issues it could improve upon:

- Default-on is too noisy. The latest mypy_primer shows ~555 new errors across 63 projects, most of which are false positives. Both Pyright (reportUnnecessaryComparison) and MyPy (comparison-overlap) deliberately keep this check opt-in / off by default because Python's == accepts object on both sides, making "incompatible" comparisons valid Python.

- The PR description says "limited to built-in scalar families" but comparison_base() accepts any ClassType, which is why we see false positives on dict, deque, tuple, attrs classes, etc.

Could you follow-up on this PR to implement:

- An opt-in via a config flag (default off) so as to not run into the false positives.
- Restrict comparison_base() to a fixed allowlist of built-in scalar types.
- Add explicit handling for numeric tower / bytes-like / set / Decimal compatibility.


We appreciate the work and time you've put into this.

website/docs/error-kinds.mdx
NathanTempest, line 1: Thank you for fixing all the comments, and also updating the documentation

NathanTempest (approved):
Thanks @asukaminato0721 for the substantial revamp! This addresses all the previous concerns:

- Default-off (Severity::Ignore) which matches Pyright's and MyPy's current convention where users opt-in via config.
- Allowlist-based grouping (EqualityCompatibilityGroup) cleanly handles all 5 false-positive families from the previous round. No more false positives on user types, dicts, lists, etc.
- Test coverage explicitly verifies the cross-group-compatible cases don't error (int == float, bytes == bytearray, set == frozenset, int == Decimal).
- Documentation updated with the scope and default severity.

Tagging @stroxler for review and @ndmitchell as they were involved in https://github.com/facebook/pyrefly/issues/721

kinto0:
cc @rchen152 in case intersection types are useful here?

stroxler (approved):
Looks pretty good to me, I'll try to get it merged. Thanks!

yangdanny97 (approved):

PR #2354, opened 2026-02-07
fix Expressions within lambdas should also affect inference #633

### LINKED ISSUE #633: Expressions within lambdas should also affect inference
Consider this example:

```python
def build_list[T](capacity: int, callback: Callable[[list[T]], None]) -> list[T]: ...

built = build_list(10, lambda l: l.append(''))
reveal_type(built)  # list[Unknown]
```

Currently, this is the behaviour of all type checkers ([Mypy](https://mypy-play.net/?mypy=1.16.1&python=3.13&flags=strict%2Cwarn-unreachable&gist=479c666f88db4ce3a2e0e6908f0b8f0d), [Pyright](https://pyright-play.net/?pyrightVersion=1.1.402&pythonVersion=3.14&strict=true&enableExperimentalFeatures=true&code=GYJw9gtgBAxmA28CmMAuBLMA7AzgOgEMAjGKdCABzBFSgGEDFjkBYAKHfYBMlgoiArunhcA%2BvHQ5UAbQAqAXQAUMAhQIx0qAJ4AuMllQAaWI3hF1Aaz0MmRZNOkSpc%2BfOMA5bEnkBKKAFoAPignGQU9PEjODjZBYVoAXn4hEXFJVEUARgAGY3gCCCIuAhC9eEIKCiQsLkUAcjqfH3YQJAA3JEZRbSrFOPhUPygAYhD06QBVLAssMAB3LHl2IA), [Pyrefly](https://pyrefly.org/sandbox/?code=GYJw9gtgBAxmA28CmMAuBLMA7AzgOgEMAjGKdCABzBFSgGEDFjkAoNgEyWCiIFd147APrx0OVAG0AKgF0AFDAIUCMdKgCeALjJZUAGliN4RFQGttDJkWQSJo8dJkyDAOWxIZASigBaAHxQ9pKy2nhhbCx8ArQAvDz8giJiqHIAjAAMBvAEEETsBIHa8IQUFEhY7HIA5FWeniwgSABuSIxCGmVyUfCo3lAAxIHJEgCqWKZYYADuWDIsQA), [ty](https://play.ty.dev/0b1e97dd-b4bb-47c7-ba56-83ffc8d09e54)). As for other languages, [Kotlin](https://pl.kotl.in/jwm_j6MrQ) and [Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7a53d8ae22952cc738754add896798f9) figure things out just fine, while [TypeScript](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwCNksJgBBGGKATwB4AVAPgAowoAHKMLDagLnipkAWwIgYAGnhsIEAlwDWA5gH0B9ANoBdAJTwAvI3gA3HFmA712gNwAoW2DwBnDPFhVqBwsVIUPzAEYABil3GgMjMOoAOnZkJwALZgByZJ0dOwB6TPhc+AA9AH5bIA) also says unknown.

I think Pyrefly can do better here and infer the type of `built` as `list[str]` (or `list[Literal['']]`, but that's too strict).

asukaminato0721:
# Summary

Fixes #633

Fixed the root cause in binding solving: lambda parameter bindings were pinning partial types too early, which collapsed contextual type vars to Any.

# Test Plan

Added a regression test to cover issue #633: lambda body should infer the type parameter from list[T] and propagate to the return type.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -2089,7 +2089,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             self.binding_to_type_info(binding, errors)
         };
         type_info.visit_mut(&mut |ty| {
-            self.pin_all_placeholder_types(ty, true, range, errors);
+            let pin_partial_types = !matches!(binding, Binding::LambdaParameter(..));
+            self.pin_all_placeholder_types(ty, pin_partial_types, range, errors);
             self.expand_mut(ty);
         });
         Arc::new(type_info)
--- a/pyrefly/lib/alt/answers_solver.rs
+++ b/pyrefly/lib/alt/answers_solver.rs
@@ -1587,11 +1587,16 @@ pub struct ThreadState {
     /// which checks before pushing its own frame).
     partial_answers: RefCell<FxHashMap<(Idx<Key>, usize), Arc<TypeInfo>>>,
     /// Solve-time mapping from per-module lambda parameter IDs to the
-    /// thread-local Var that represents that parameter in the current solve.
+    /// thread-local Var that represents that parameter while its owner lambda
+    /// is actively being solved.
     ///
     /// The `ModulePath` is needed to distinguish the in-memory and on-disk
     /// versions of the same module, which can coexist in the IDE (issue #3789).
-    lambda_param_vars: RefCell<FxHashMap<(ModuleName, ModulePath, LambdaParamId), Var>>,
+    active_lambda_param_vars: RefCell<FxHashMap<(ModuleName, ModulePath, LambdaParamId), Var>>,
+    /// The most recent completed Var for a lambda parameter ID. This allows a
+    /// `Binding::LambdaParameter` that forces its owner first to reuse the
+    /// finished parameter Var after the owner lambda returns.
+    completed_lambda_param_vars: RefCell<FxHashMap<(ModuleName, ModulePath, LambdaParamId), Var>>,
     /// Active trace side-effect sink for the current calculation.
     /// Set before `K::solve`, taken after. `None` when tracing is disabled
     /// or between calculations. Saved sinks form a stack to handle recursive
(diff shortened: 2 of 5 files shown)


pyrefly/lib/alt/solve.rs
```diff
@@ -1682,7 +1682,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             let pin_partial_types = !self.solver().infer_with_first_use
                 || !matches!(
                     binding,
-                    Binding::NameAssign { .. } | Binding::PartialTypeWithUpstreamsCompleted(..)
+                    Binding::NameAssign { .. }
+                        | Binding::PartialTypeWithUpstreamsCompleted(..)
+                        | Binding::LambdaParameter(..)
```
stroxler: @asukaminato0721 do you have a strong sense of where the parameter type can flow?

There's a *lot* of work that goes into trying to "trap" partial type Vars from NameAssign so that they cannot leak out into other bindings; this is necessary because pinning a Var is sensitive to order of evaluation, which means it's nondeterministic unless it's been carefully handled in order to not be.

I suspect that LambdaParameter is able to be used in ways that can bubble out into multiple bindings in the graph, which would make this fix nondeterministic

We might be able to do a lot better once recursion handling is improved

stroxler (changes requested):
I suspect that this is a very hard feature to ship without potential nondeterminism. Back to you to think about this - if you can trace through every possible flow of `Binding::LambdaParameter` and work out where the pin happens and why it would be deterministic, then we could potentially ship this (with some comments explaining things) but my current instinct is that it's pretty risky

samwgoldman:
Like I said in the linked issue, I don't think we should pursue this kind of inference _yet_. For now, I want to keep inference of lambda parameters simple and just based on contextual typing from the outside.

Thanks for your work on this PR.

PR #2356, opened 2026-02-07
fix use all usages within class (or within __init__) to infer type, not just initial assignment #1159

### LINKED ISSUE #1159: use all usages within class (or within __init__) to infer type, not just initial assignment
### Describe the Bug

```py
class A:
    a = 1
    def f(self):
        self.a = "a"  # error, expect no error
```

here `a` is `int`, but if we look at all usages in the class then we can infer a much better type

asukaminato0721:
# Summary

Fixes #1159

Implemented the issue 1159 request to use assignments within the class (e.g., a = 1 then self.a = "a") when inferring attribute types, so the method assignment widens the inferred type instead of erroring.

# Test Plan

add tests

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1872,35 +1872,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
         };
 
-        let mut merged_method_assignments = false;
-        if !method_assignments.is_empty()
-            && annotation
-                .as_ref()
-                .and_then(|ann| ann.ty.as_ref())
-                .is_none()
-        {
-            let ignore_errors = self.error_swallower();
-            let mut types = Vec::with_capacity(method_assignments.len() + 1);
-            types.push(value_ty);
-            for assignment in method_assignments {
-                let direct_annotation = assignment
-                    .annotation
-                    .map(|annot| self.get_idx(annot).annotation.clone());
-                let (ty, _, _) = self.analyze_class_field_value(
-                    &assignment.value,
-                    class,
-                    name,
-                    direct_annotation.as_ref(),
-                    true,
-                    assignment.range,
-                    &ignore_errors,
-                );
-                types.push(ty);
-            }
-            value_ty = self.unions(types);
-            merged_method_assignments = true;
-        }
-
         if let Some(annotation) = direct_annotation.as_ref() {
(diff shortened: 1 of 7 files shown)


kinto0:
interesting... it does end up adding more errors than it removes so the argument of "widen type instead of erroring" might not hold. let's see what other people think

rchen152:
Hmm. I'm not convinced this is a good idea. Looking at the proposed new behavior:
```
class A:
    a = 1
    def set_a(self, value: str):
        self.a = value
def f(a: A):
    assert_type(a.a, int | str)
```
this strikes me as potentially surprising behavior (if `set_a` is buried in a long class definition, a user might be quite confused about where the `str` type is coming from) and a little out-of-step with the "infer on first use" rule that pyrefly uses elsewhere.

rchen152 (commented):
(Leaving a comment to clear this out from my review queue.)

yangdanny97:
we should experiment with only allowing this in `__init__`

asukaminato0721:
@yangdanny97 done

yangdanny97:
thanks @asukaminato0721 

odd, the mypy primer delta is more mixed than i would expect

@migeed-z can you run the classifier on this one?

yangdanny97:
wow that's a long comment lol, i'll read tomorrow and see if it's legit

yangdanny97:
I think the regressions are mostly caused by something being initialized both on the class & in the constructor.

I wonder if we should only take the union of assignments in `__init__` if it's not initialized on the class, and take the class's initialization as the single source of truth if it's present.

The downside is that it wouldn't be able to handle things like this. We currently have a special-case for `None` on the class that makes it `Any | None`, but if we restrict this analysis we wouldn't be able to remove that hack.

```
class C:
  x = None

def __init__(self):
  if something:
    self.x = "foo"
```

yangdanny97:
ok, i tried a few tweaks on top of this, but the delta still does not look super good. the LLM classifier output is somewhat nonsensical, which confused me for a bit.

I'm going to attempt a different approach that synthesizes fake variables for each attribute & infers the type based on their flow info at exit points to `__init__`, similar to how we do return type inference.

yangdanny97:
I wasn't able to figure out a balanced solution here that didn't regress mypy primer

yangdanny97:
Thanks for your work prototyping the fix here, the original issue was closed by https://github.com/facebook/pyrefly/pull/3835 so I will close this out

PR #2358, opened 2026-02-08
  Fix unannotated-parameter if first classmethod parameter not named cls (#2327)

### LINKED ISSUE #2327: `unannotated-parameter` if first classmethod parameter not named `cls`
### Describe the Bug

When I enable the `unannotated-parameter` error in scipy-stubs (scipy/scipy-stubs#1262), 17 errors are reported ([gh action log](https://github.com/scipy/scipy-stubs/actions/runs/21730604142/job/62683973208?pr=1262)) for the first parameter of classmethods or `__new__` methods if its name isn't `cls`

Repro

```pyi
class A:
    def __new__(cls, a: int) -> A: ...  # ✔️ 

class B:
    def __new__(_cls, a: int) -> B: ...  # ❌ `__new__` is missing an annotation for parameter `_cls`
```

([sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQCCiAOuvR-ZjGPQPp90MAO4CAFNTgAaeqkT0I6AC4BKegFoAfE3mE9bNtVoMAQq3aduvAUNF8xfSTLkLlarfTP09hNiCkgAK5K0HAk5IggAMT0AKohUBBKpPRggeiUIbjocAZYPKm4AE4AtqhKgoEl2DBFYvjyiqoa2nBKReac9EUwSoFF7GAsIAByVTUd9MD4AL7DfgFkPWBQpIRKuCVQFDEACqTLq-RoWHj49JTZkADm-eUQ2b7oMQDKMDD0ABZKSsRwiAB6AFLHirQjFa4AmDoAGYXCUOAAy7oG53LIwwpFWQAN1Q0FQ2FgFyuEFuRXu2XouGI6LCbDISk%2B2XU2NqcAe7AAvPRhgBmQgARgALPN0GwANq1IrFOAAXTYEBKxESlCS6gwKW5wylxVF6Qw6FwSnKMEw6mIqHJJV6tXoWpAOqKevQBqNJrNPT6AztPIdRWlTpAfhmAVQmQgrIAYtAYBQQSs1hstiAZkA))

I'd rather just name them `cls`, but I'm not able to, because then it wouldn't match the runtime name (i.e., stubtest will complain).

And although it's not an issue for scipy-stubs, I also noticed that this is also the case for `self` parameters of instance methods ([sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQDCiAOuvR-ZjGPQPp8I6CABcBACjgwoYAJT0AtAD56AOVzoYieoV1s21WgwAirdp268BQ0RL5SZ85Wo1adukABoQAVxHQ4EnJEEABiegBVfyhRUnowH3RKfw04fSweeNwAJwBbVDF0H1zsGGzxfG0hEScVOBFss056bJgRH2z2MBYQVWLSxvpgfABfHrYvEDJWsChSQhFcXKgKcIAFUhm5%2BjQsPHx6Sg1IAHMOgogNQjZwgGUYGHoACxERYjhEAHpP6Z45whyJ0%2BMHQn0wuEocE%2BR3Qp3OKVBWWy9FQADdUNBUNhYIdjhAztkLhp6LhiAjAmwyCInhoFKiynBLuwALz0HoAZkIAEYACzjdBsADaZWyOTgAF02BBcsQYpRRAoMHFWT0RTl%2BYkMOhcCICjBMApiKhCbk2mV6CqQGrshr0FqdXqDa12p0LWyrdlRTaQBMRt5UMkIPSAGLQGAUXY4AhBEAjIA)).

---

The opposite also seems to be the case, i.e. non-instance method parameters called `self` or `cls` without annotation, regardless of position, will **not** be reported as unannotated. For example the following will be accepted:

```py
def f(a: str, self, cls, b: int) -> None: ...  # ✔️ 
```

([sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABSqI1wAuATgDQsxRjcBjKHG7ZmEdKwCUNALQA%2BGgDlc6GM0KaQnEAFdW0OCXKIQAYhoBVA1AitSdXegEHVcADrpPVWmFzsAW1RWAH10XQDsGHZ6fHFJGQUWDkRPGnSadhhWXXZ0OncQJQio9mZgfABfQs9tEDIssChSQlZcAKgKCwAFUkbmlgwcAhoBVUgAc1zgiFVCTwsAZRgYGgALVlZiOEQAel2G6mbCfwndmHRdzFwBOF2x9Enp10u6fxpUADdUaFRsWFG4wgU3YM1UNFwxBeRk8ZFYa1Usk%2B0Tgs3yAF4aIUAMyEACMABYal50ABtaLsfxwAC6nggAWItgEdlkGAcmMKFP8xKcGHQuFYwRgmFkxFQoIC2WiNA5IC57B56D5AqFIqyOTyMqxcvYlIVIFqlR0qBcEGRADFoDAKGgsHgiGQQJUgA))

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQCCiAOuvR-ZjGPQPp90MAO4CAFNTgAaeqkT0I6AC4BKegFoAfE3mE9bNtVoMAQq3aduvAUNF8xfSTLkLlarfTP09hNiCkgAK5K0HAk5IggAMT0AKohUBBKpPRggeiUIbjocAZYPKm4AE4AtqhKgoEl2DBFYvjyiqoa2nBKReac9EUwSoFF7GAsIAByVTUd9MD4AL7DfgFkPWBQpIRKuCVQFDEACqTLq-RoWHj49JTZkADm-eUQ2b7oMQDKMDD0ABZKSsRwiAB6AFLHirQjFa4AmDoAGYXCUOAAy7oG53LIwwpFWQAN1Q0FQ2FgFyuEFuRXu2XouGI6LCbDISk%2B2XU2NqcAe7AAvPRhgBmQgARgALPN0GwANq1IrFOAAXTYEBKxESlCS6gwKW5wylxVF6Qw6FwSnKMEw6mIqHJJV6tXoWpAOqKevQBqNJrNPT6AztPIdRWlTpAfhmAVQmQgrIAYtAYBQQSs1hstiAZkA

_No response_

jackulau:
## Summary

  Fixes #2327

  The `unannotated-parameter` error was using a name-based check (`name != "self" && name != "cls"`) to decide whether to skip the first
  parameter of methods. This caused:

  - **False positives**: Methods whose first parameter was named `_cls`, `_self`, `klass`, etc. were incorrectly flagged
  - **False negatives**: Standalone functions with parameters named `self` or `cls` were incorrectly skipped

  The fix changes the check to be position-based: skip only the first non-variadic parameter when the function is a non-static method in a
  class. This correctly handles:

  - Instance methods with any name for the first parameter (e.g. `_self`, `this`)
  - `@classmethod` methods with any name for the first parameter (e.g. `_cls`, `klass`)
  - `__new__` (implicit staticmethod that still takes `cls`)
  - `__init_subclass__` (implicit classmethod)
  - `@staticmethod` methods (no implicit first parameter — all params checked)
  - Standalone functions (no implicit first parameter — `self`/`cls` names are not special)
  - Vararg-only methods like `def method(*args)` (no implicit self parameter)

  ## Test plan

  - Added `test_unannotated_parameter_first_param_by_position` covering all scenarios above
  - Verified existing `test_implicit_any_self_cls_ignored` still passes

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -522,15 +522,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // The first parameter of a non-static method is the implicit self/cls
         // parameter and does not require an annotation, regardless of its name.
         // __new__ is an implicit staticmethod but still takes cls as its first parameter.
+        // If the first parameter is variadic (e.g. *args), self is passed inside it,
+        // so there is no separate implicit parameter to skip.
         let is_dunder_new = def.defining_cls.is_some() && stmt.name.id == dunder::NEW;
-        let has_implicit_self_param =
+        let has_implicit_self_or_cls_param =
             def.defining_cls.is_some() && (!def.metadata.flags.is_staticmethod || is_dunder_new);
-        let mut skip_self_param = has_implicit_self_param;
+        let mut skip_first_param = has_implicit_self_or_cls_param;
         for p in stmt.parameters.iter() {
-            // The implicit self/cls param is always the first non-variadic parameter.
-            if skip_self_param && !p.is_variadic() {
-                skip_self_param = false;
-                continue;
+            if skip_first_param {
+                skip_first_param = false;
+                if !p.is_variadic() {
+                    continue;
+                }
             }
             if p.annotation().is_none() {
                 let name = p.name().as_str();
--- a/pyrefly/lib/test/inference.rs
+++ b/pyrefly/lib/test/inference.rs
@@ -85,9 +85,7 @@ class C:
 // https://github.com/facebook/pyrefly/issues/2327
 testcase!(
     test_unannotated_parameter_first_param_by_position,
-    TestEnv::new()
-        .enable_unannotated_parameter_error()
-        .enable_unannotated_return_error(),
+    TestEnv::new().enable_unannotated_parameter_error(),
     r#"
 class A:
     def __new__(cls, a: int) -> "A": ...
@@ -115,6 +113,9 @@ class C:
 
     # vararg as first param is NOT an implicit self param
     def vararg_method(*args, **kwargs) -> None: ...  # E: `vararg_method` is missing an annotation for parameter `args` # E: `vararg_method` is missing an annotation for parameter `kwargs`
+
+    # keyword-only params after variadic first param should also error
+    def vararg_method2(*args, name) -> None: ...  # E: `vararg_method2` is missing an annotation for parameter `args` # E: `vararg_method2` is missing an annotation for parameter `name`
 
 # self/cls in standalone functions should still error
 def f(a: str, self, cls, b: int) -> None: ...  # E: `f` is missing an annotation for parameter `self` # E: `f` is missing an annotation for parameter `cls`


pyrefly/lib/test/inference.rs
```diff
@@ -82,6 +82,49 @@ class C:
 "#,
 );
 
+// https://github.com/facebook/pyrefly/issues/2327
+testcase!(
+    test_unannotated_parameter_first_param_by_position,
+    TestEnv::new()
+        .enable_unannotated_parameter_error()
+        .enable_unannotated_return_error(),
```
yangdanny97: I don't think we need this line

pyrefly/lib/alt/function.rs
```diff
@@ -519,9 +519,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        // parameter and does not require an annotation, regardless of its name.
+        // __new__ is an implicit staticmethod but still takes cls as its first parameter.
+        let is_dunder_new = def.defining_cls.is_some() && stmt.name.id == dunder::NEW;
+        let has_implicit_self_param =
+            def.defining_cls.is_some() && (!def.metadata.flags.is_staticmethod || is_dunder_new);
+        let mut skip_self_param = has_implicit_self_param;
         for p in stmt.parameters.iter() {
-            let name = p.name().as_str();
-            if p.annotation().is_none() && name != "cls" && name != "self" {
+            // The implicit self/cls param is always the first non-variadic parameter.
```
yangdanny97: This is wrong, if the first parameter is variadic the `self` gets passed before any user-provided arguments

For example, the instance of the class is stored in `*args` here, and `self` is just a regular named argument:

```
>>> class C:
...     def test(*args, self):
...         print(args, self)
>>> x = C()
>>> x.test(1, 2, 3, self="abc")
(<__main__.C object at 0x1059970e0>, 1, 2, 3) abc
```

yangdanny97: This means that if the first parameter is variadic, we should still set `skip_self_param` to false

pyrefly/lib/alt/function.rs
```diff
@@ -519,9 +519,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 format!("`{}` is missing a return annotation", stmt.name),
             );
         }
+        // The first parameter of a non-static method is the implicit self/cls
+        // parameter and does not require an annotation, regardless of its name.
+        // __new__ is an implicit staticmethod but still takes cls as its first parameter.
+        let is_dunder_new = def.defining_cls.is_some() && stmt.name.id == dunder::NEW;
+        let has_implicit_self_param =
```
yangdanny97: nit: probably rename to `has_implicit_self_or_cls_param` & `skip_first_param`

PR #2373, opened 2026-02-10
fix: avoid adding parameters for variables defined in the selection during helper extraction

### LINKED ISSUE #2335: Extract to function refactoring capability sometimes tries to use undeclared variables.
### Describe the Bug

I ran into a bug where when using the Extract to Function we may try to use variables that are not declared.

I was able to create this using the following code example:

```python
def calculate_total_price(prices: list[int]) -> float:
    total = 0
    for price in prices:
        total += price

    with_tax = total * 1.085
    return with_tax
```

If you highlight everything in the function body besides the return statement and try to extract it, the original function will try to pass in `total` even though it does not exist in the scope of that function

```python
def extracted_function(prices, total):
    total = 0
    for price in prices:
        total += price
    
    with_tax = total * 1.085
    return with_tax

def calculate_total_price(prices: list[int]) -> float:
    ## Total does not exist here
    with_tax = extracted_function(prices, total)

    return with_tax
```

https://github.com/user-attachments/assets/6e49730a-aa4e-4cfe-ace8-27bcccdec122

### Sandbox Link

_No response_

_No response_

self-sasi:
# Summary

This PR fixes an issue where the “Extract into helper" functionality incorrectly added variables as parameters to the extracted function when those variables were actually defined inside the selection. 

Specifically, variables on the left-hand side of augmented assignments (e.g., `total` in `total += price`) were unconditionally added to the extracted method's parameter list, even if the variable was initialised earlier in the same selection (e.g., `total = 0`). 

The fix simply checks `store_refs` for a prior assignment of the variable within the selection before adding it as a parameter.

Fixes #2335 

# Test Plan

1. Added a regression test to cover the changed behaviour.
2. Manual testing: 
2.1. Locally built pyrefly to `/some-dir/pyrefly`
2.2. In editor settings, add the following setting: `”pyrefly.lspPath”: ”/some-dir/pyrefly”`
2.3. Manually test the “Extract into helper” functionality to verify fix. See below for reference. 

https://github.com/user-attachments/assets/58b18f2a-262b-468b-8f28-1a6e7ee35ebe

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/extract_function.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/extract_function.rs
@@ -72,8 +72,13 @@ pub(crate) fn extract_function_code_actions(
             continue;
         }
         if ident.synthetic_load {
-            seen_params.insert(ident.name.clone());
-            params.push(ident.name.clone());
+            let defined_earlier_in_selection = store_refs
+                .iter()
+                .any(|store| store.name == ident.name && store.position < ident.position);
+            if !defined_earlier_in_selection {
+                seen_params.insert(ident.name.clone());
+                params.push(ident.name.clone());
+            }
             continue;
         }
         let defs = transaction.find_definition(handle, ident.position, FindPreference::default());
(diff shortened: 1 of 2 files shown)


self-sasi:
@kinto0 could you PTAL?

kinto0:
sorry about the delay! looking now

PR #2374, opened 2026-02-10
consolidate completion ranking

kinto0:
Summary: completion ranking is currently very confusing. it happens in many places and sort_text can be misleading or overwritten at random. this diff moves it into a centralized place, by makin g a RankedCompletion struct and using it everywhere completions are needed to be ranked

Differential Revision: D92792321


stroxler (approved):

PR #2380, opened 2026-02-10
Update search path prefix when getting source db

connernilsen:
Summary: This should fix the sourcedb bug Jia mentioned last week, where sometimes search path prefix entries aren't used correctly. This happens when there's already a sourcedb entry loaded, so we skip appending the config root to the search path as expected.

Reviewed By: kinto0

Differential Revision: D92848242


kinto0 (approved):

Issue #2369, opened 2026-02-09
`type[T]` constructors don't use `TypeVar` bounds for parameter inference


migeed-z:
### Describe the Bug

 When calling a constructor via `type[T]` where `T` is bounded, pyrefly should use the bound's
  `__init__` signature to determine valid parameters. Currently it falls through to
  `object.__init__` and accepts any call.

```
def test[T](cls: type[T]) -> T:
    cls(1)  # should error: no args for object constructor
    return cls()
class A:
    def __init__(self, x: int) -> None: pass
def test2[T: A](cls: type[T]) -> T:
    a1: A = cls()  # should error: missing required arg x
    a2: A = cls(1)
    return cls()
```

Note: this unittest is already available in the codebase. 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABAC7x0DaAKgLoAUAxlHIvaWIxWbAJQ0AtAD4aLRAB10NZTR5wOARnE0AxDTgALXAFcomGjABOl3Jf65sAKxhc6HcXVQBreDXS4aVEsAczhFFRpLGDpjSyU1d0VFHlQ4OBoAQQUlFSpaAH18iHQIOkKOOBgoMAAaGnx%2BYrpxaRoAOVx0GH5iVLD0RTz6RgAmVn4MzjV%2BOkFhdhaZOXCVVA0JmgBeVV53ZT1DEzMLa1t%2BAFsINOLgyJgAR2MIKPMg2-wV5VQRje2ErU%2BdxicR26lEihANRAxjo0DgJHIiBAegAqrCoKVSDQwMZ0K4IJ1%2BoNqNjbOdUGV0MZztgrBwGjQmot9HQ7ICosClGB5CA2tTaXYaMB8ABfHkQqFkKJgKCkQh0XDnKAUPQABVI0tl%2BgwOAIqk6kGCsQpBPQhEUegAyjAYDQDHQ6MQ%2BAB6Z1S6iywi2YLOmDoZ2YXBcODOrgGiBGywmzrO0mWQIAN1Q0FQ2Fg%2BvQhuNsM6NFwxBz6HhijIdCM6AkCascFNWxoPIAzIQNCNxegQCKoah8VWAGLQGAUNBYPBEMgdoA

PR #2381, opened 2026-02-10
[conformance] Handle `type[T]` calls where T is bounded by a class

fangyi-zhou:
# Summary

When types of `type[T]` are called, we currently infer a callable that takes any arguments and returns the type variable T. This is a conformance gap, where we should treat such calls as on `T`. However, the spec made clear that this is not free of problems:

From https://typing.python.org/en/latest/spec/constructors.html#constructor-calls-for-type-t
> It should be noted that such code could be unsafe because the type `type[T]` may represent subclasses of `T`, and those subclasses could redefine the `__new__` and `__init__` methods in a way that is incompatible with the base class. Likewise, the metaclass of `T` could redefine the `__call__` method in a way that is incompatible with the base metaclass.

If `T` is unrestricted, we assume an implicit bound of `object`, and use the callable type `() -> T`.
If `T` is bounded, we use the bound class type to infer the callable signature, but use the type `T` as the return type.

We cannot handle constrained type variables, since we don't have any intersection type support at the call target level, so I'm leaving the current behaviour as is.

Let's assess the impact on open source codebases, since I expect this change will bring some false positives.

#2369

# Test Plan

Ran conformance test suite and fixed 4 failing tests.

### DIFF
--- a/pyrefly/lib/test/generic_basic.rs
+++ b/pyrefly/lib/test/generic_basic.rs
@@ -11,7 +11,6 @@ use crate::test::util::TestEnv;
 use crate::testcase;
 
 testcase!(
-    bug = "conformance: Should use bounds/constraints of type var to determine callable input type for bounded type[T] constructors",
     test_tyvar_constructor,
     r#"
 def test[T](cls: type[T]) -> T:


fangyi-zhou:
Checking the mypy_primer results, I think we're getting some false positives due spec warning and dynamic behaviours.

The error in `static_frame` is a bit strange, not sure where we infer any callable in the constructor

PR #2384, opened 2026-02-11
fix: correctly feature-gate backtrace-on-stack-overflow

SkyBird233:
# Summary

The `backtrace-on-stack-overflow` dependency is currently compiled unconditionally, even though `main.rs` only attempts to use it behind the `debug-stack-overflow` feature flag. This PR registers the `debug-stack-overflow` feature and marks the dependency as optional.

This change prevents compiling unused code for default builds. It also resolves build failures on architectures not supported by the outdated `nix` 0.23 transitive dependency (such as `loongarch64`).

Additionally, the `--all-features` flag has been removed from `build_extension.yml` becaues the crate previously had no features, and the feature added in this PR is not intended for production use.

# Test Plan

Verified using `cargo check` to confirm conditional compilation:
```
$ cargo check -v 2>&1| grep stack-overflow

$ cargo check --features debug-stack-overflow -v 2>&1| grep stack-overflow
       Fresh backtrace-on-stack-overflow v0.3.0
```

The change has also been verified in [this job](https://github.com/loongcodium/pyrefly-loong64/actions/runs/21800834275/job/62895972643) (cross compiling Pyrefly for `loongarch64`).

### DIFF
--- a/.github/workflows/build_extension.yml
+++ b/.github/workflows/build_extension.yml
@@ -111,11 +111,11 @@ jobs:
         with:
           command: build
           target: ${{ matrix.rust_target }}
-          args: "--release --all-features --artifact-dir lsp/bin --manifest-path pyrefly/Cargo.toml -Z unstable-options"
+          args: "--release --artifact-dir lsp/bin --manifest-path pyrefly/Cargo.toml -Z unstable-options"
           toolchain: ${{ env.toolchain }}
       - name: build pyrefly binary
         if: ${{ matrix.rust_target == '' }}
-        run: cargo build --release --all-features --artifact-dir lsp/bin --manifest-path pyrefly/Cargo.toml -Z unstable-options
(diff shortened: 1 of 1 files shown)


stroxler:
@jvansch1 Do you have a sense how valuable the feature is?

We should probably try to get this fix in, but the tricky part is that `autocargo` has to be able to generate this config for us (absolutely no manual edits of Cargo.toml are permitted, which is in place to make sure the buck and cargo builds stay in sync)

If we can figure it out, it's a good fix but I've run into a couple walls dealing with autocargo at times

I wonder if we should either just get rid of this and go back to having a patch that's easy to apply as-needed. Feature flags are pretty hard to maintain with the dual build

jvansch1:
@stroxler Let me investigate this a bit and see if I can get it playing nice with Buck and Cargo. If not then I think it is something that we can just remove for now.

PR #2389, opened 2026-02-12
[conformance] Error if an imported name marked as `Final` is re-assigned

fangyi-zhou:
# Summary

This diff implements a new feature to error when an imported `Final` name is re-assigned.
This requirement stems from the spec section:

https://typing.python.org/en/latest/spec/qualifiers.html#importing-final-variables
> If a module declares a `Final` variable and another module imports that variable in an import statement by name or wildcard, the imported symbol inherits the `Final` type qualifier. Any attempt to assign a different value to this symbol should be flagged as an error by a type checker

Implementation notes:
- Add a new field `is_final` for each Export, to indicate whether the name is final
- Extend export lookup to check for finality, similar to special exports are handled
- Raise an error if an imported final value is assigned to.

# Test Plan

New test case.
Reran conformance test suite and observed 2 fewer errors.

### DIFF
--- a/pyrefly/lib/test/imports.rs
+++ b/pyrefly/lib/test/imports.rs
@@ -1155,3 +1155,13 @@ X = 10  # E: Cannot assign to `X` because it is imported as final
 Y = 10  # E: Cannot assign to `Y` because it is imported as final
 "#,
 );
+
+testcase!(
+    test_modify_imported_as_final_value,
+    env_final_value(),
+    r#"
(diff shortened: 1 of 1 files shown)


fangyi-zhou:
Checked mypy-primer issues, most of them are of form of re-importing a `Final` value twice (potentially once via `*` import, once via direct import)

stroxler (approved):

PR #2391, opened 2026-02-12
optimize can_be_recursive

yangdanny97:
Summary:
D92216744 refactored some of the recursive assumptions code.

We actually only call is_subset_protocol if the RHS is a protocol, the LHS doesn't matter. Avoiding the LHS check means that we can store fewer recursive assumptions, and clone less.

There appears to be a modest performance benefit.

Reviewed By: rchen152

Differential Revision: D93043272


rchen152 (approved):

PR #2397, opened 2026-02-12
add code lens run command for test and main #2361

### LINKED ISSUE #2361: add code lens for test and may be main?
### Describe the Bug

Image

add it to pytest / unittest / `if __name__ == "__main__"` will be handy.

rust analyzer has this.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2361

Implemented LSP CodeLens generation for pytest/unittest-style tests and `if __name__ == "__main__"` with Run/Test commands

Added VS Code extension commands (pyrefly.runMain, pyrefly.runTest) and wiring to execute main files or trigger the VS Code Testing API at the cursor.

# Test Plan

Added LSP interaction coverage and fixtures for CodeLens behavior.

```
> '/usr/bin/python' -m unittest 'main.TestT.test_'
.
----------------------------------------------------------------------
Ran 1 test in 0.000s

OK

> '/usr/bin/python' -m unittest 'main.TestT'
.
----------------------------------------------------------------------
Ran 1 test in 0.000s

OK

> '/usr/bin/python' '/tmp/tmp/main.py'
Hello from tmp!
```

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/pytorch_benchmark.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/pytorch_benchmark.rs
@@ -48,7 +48,7 @@ fn test_pytorch_error_propagation_latency() {
     };
     // Use all available cores for realistic benchmarking
     let mut interaction =
-        LspInteraction::new_with_args(args, NoTelemetry, Some(ThreadCount::AllThreads));
+        LspInteraction::new_with_args(args, NoTelemetry, Some(ThreadCount::AllThreads), None);
     interaction.set_root(pytorch_root.clone());
 
     interaction


lsp/src/extension.ts
```diff
@@ -91,6 +94,154 @@ async function overridePythonPath(
   return newResult;
 }
 
+function shellQuote(value: string): string {
+  if (value.length === 0) {
+    return "''";
+  }
+  if (process.platform === 'win32') {
+    return `"${value.replace(/"/g, '\\"')}"`;
+  }
```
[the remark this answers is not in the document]
kinto0: let's look at how other language servers deal with this and just copy that. do they escape this?

lsp/src/extension.ts
```diff
@@ -92,6 +95,154 @@ async function overridePythonPath(
   return newResult;
 }
 
+function shellQuote(value: string): string {
```
kinto0: can all this stuff go in a separate codeLens.ts file?

lsp/src/extension.ts
```diff
@@ -91,6 +94,154 @@ async function overridePythonPath(
...
+  if (relativePath.startsWith('..')) {
+    return undefined;
+  }
+  if (relativePath.endsWith('.py')) {
+    relativePath = relativePath.slice(0, -3);
+  }
+  return relativePath
+    .split(path.sep)
+    .filter(part => part.length > 0)
+    .join('.');
```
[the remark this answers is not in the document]
kinto0: would there ever be stubs in tests or main? probably not?

lsp/src/extension.ts
```diff
@@ -92,6 +95,154 @@ async function overridePythonPath(
...
+      nodeId = `${nodeId}::${className}`;
+    }
+    if (testName) {
+      nodeId = `${nodeId}::${testName}`;
+    }
+    command = `${shellQuote(interpreter)} -m pytest ${shellQuote(nodeId)}`;
+  }
+
+  terminal.show(true);
+  terminal.sendText(command ?? `${shellQuote(interpreter)} -m pytest`);
```
kinto0: what happens if pytest is not installed? or we're on an old version? should we avoid showing this?

kinto0 (changes requested):
how useful do we think this will be? does the python extension come with it? if so, the pyrefly extension depends on that, so it might duplicate this functionality

asukaminato0721:
image

image

lsp/src/extension.ts
asukaminato0721: moved now

lsp/src/extension.ts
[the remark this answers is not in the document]
asukaminato0721: switched to vscode.ProcessExecution tasks in lsp/src/codeLens.ts

lsp/src/extension.ts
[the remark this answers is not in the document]
asukaminato0721: now .pyi CodeLens suppressed server-side

lsp/src/extension.ts
asukaminato0721: My reasoning was that, we don’t have a reliable cheap signal for that at lens-generation time, and hiding the action based on guesswork would be worse than exposing it and letting execution fail visibly if the environment is wrong.

improved the execution path.

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,205 @@
...
+const TASK_SOURCE = 'pyrefly';
+
+async function interpreterForUri(
+  uri: vscode.Uri,
+  pythonExtension: PythonExtension,
+): Promise<string> {
+  const envPath = await pythonExtension.environments.getActiveEnvironmentPath(
+    uri,
+  );
+  return envPath.path.length > 0 ? envPath.path : 'python';
```
kinto0: is there a way to make this fallback use the same logic as our extension? does our extension do something similar? python as a string seems sketchy to me

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function scopeForUri(uri: vscode.Uri): vscode.WorkspaceFolder | vscode.TaskScope {
+  return vscode.workspace.getWorkspaceFolder(uri) ?? vscode.TaskScope.Workspace;
```
kinto0, line 162: what does this fallback do?

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,205 @@
...
+): Promise<void> {
+  // Use ProcessExecution so VS Code passes argv directly instead of relying on
+  // shell-specific quoting rules.
+  const task = new vscode.Task(
+    definition,
+    scopeForUri(uri),
+    label,
+    TASK_SOURCE,
+    new vscode.ProcessExecution(command, args, {
+      cwd: cwdForUri(uri),
```
kinto0: does this always work? would it be better to use pyrefly's config finding logic?

lsp/src/codeLens.ts
asukaminato0721: now removed python as a string

lsp/src/codeLens.ts
asukaminato0721: “attach this task to the workspace instead of a folder” for loose files / non-folder URIs.

lsp/src/codeLens.ts
asukaminato0721: now resolves the runnable root from Pyrefly’s config finder first, then falls back to the server’s workspace root only if no config root exists.

asukaminato0721:
image

now it would just run and error.

kinto0:
sorry about the delay: two more small requests and then i'm happy with it

- can we add more details to the error message? something like "ensure the correct interpreter is selected in your IDE" for the interpreter not found and something actionable for the pytest not found error?
- should we prompt them to disable this editor setting if it does fail

asukaminato0721:
Missing interpreter now shows a more specific error: “Ensure the correct interpreter is selected in your IDE…”

Pytest runs now do a preflight import pytest check before launching the task.
If that fails, the user gets an actionable error telling them to select the correct interpreter or install pytest in that environment.

I did not add a generic “disable this feature” prompt for every task failure after launch, because with ProcessExecution we do not get a reliable structured reason for arbitrary runner failures.

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,282 @@
...
+  );
+}
+
+export function registerCodeLensCommands(
+  context: ExtensionContext,
+  pythonExtension: PythonExtension,
+): void {
+  context.subscriptions.push(
+    vscode.commands.registerCommand('pyrefly.runTest', async args => {
+      await runTestAtLocation(args as RunTestArgs, pythonExtension);
```
kinto0: can we validate these instead of casting them? as I understand it this will crash within runTestAtLocation if args are wrong

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,282 @@
...
+    }),
+  );
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand('pyrefly.runMain', async args => {
+      await runMainFile(args as RunMainArgs, pythonExtension);
```
kinto0: same. can we validate?

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  let nodeId = uri.fsPath;
+  if (className) {
+    nodeId = `${nodeId}::${className}`;
+  }
+  if (testName) {
+    nodeId = `${nodeId}::${testName}`;
+  }
+  if (!(await canImportModule(interpreter, 'pytest', cwd))) {
+    void showRunnableCodeLensError(
```
kinto0, line 315: I worry this error will happen each time we try showing a codelens. should we have a maximum number of times to show it?

lsp/src/codeLens.ts
```diff
@@ -0,0 +1,282 @@
...
+    await vscode.commands.executeCommand(
+      'workbench.action.openSettings',
+      'python.pyrefly.runnableCodeLens',
+    );
+  } else if (action === DISABLE_RUNNABLE_CODE_LENS) {
+    await disableRunnableCodeLens(uri);
+  }
+}
+
+async function canImportModule(
```
kinto0, line 146: nit: I wonder if we could move all the pytest errors into their own file?

lsp/package.json
```diff
@@ -221,6 +231,12 @@
                     "default": true,
                     "description": "If true, Pyrefly will sync notebook documents with the language server. Set to false to disable notebook support."
                 },
+                "python.pyrefly.runnableCodeLens": {
```
kinto0, line 234: another idea: we should probably let people disable code lenses as an lsp operation with the `Disabled Language Services`

pyrefly/lib/lsp/non_wasm/code_lens.rs
```diff
@@ -0,0 +1,179 @@
...
+
+fn is_name_dunder_name(expr: &Expr) -> bool {
+    matches!(expr, Expr::Name(name) if name.id.as_str() == "__name__")
+}
+
+fn is_main_string(expr: &Expr) -> bool {
+    matches!(
+        expr,
+        Expr::StringLiteral(ExprStringLiteral { value, .. }) if value.to_str() == "__main__"
+    )
```
kinto0, line 244: is there nowhere else in pyrefly we do this we can copy to avoid duplication?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4660,6 +4687,83 @@ impl Server {
...
 
+    fn code_lens(
+        &self,
+        transaction: &Transaction<'_>,
+        params: CodeLensParams,
+    ) -> Option<Vec<CodeLens>> {
+        let uri = &params.text_document.uri;
+        if uri.path().ends_with(".pyi") {
+            return Some(Vec::new());
+        }
```
kinto0: not all code lenses will have this requirement if we add more. can we make it specific to the runnable ones?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4660,6 +4687,83 @@ impl Server {
...
+        let maybe_cell_idx = self.maybe_get_cell_index(uri);
+        let handle = self.make_handle_if_enabled(uri, Some(CodeLensRequest::METHOD))?;
+        let info = transaction.get_module_info(&handle)?;
+        let entries = transaction.code_lens_entries(&handle)?;
+        let cwd = self.runnable_code_lens_cwd(&path);
+
+        let mut lenses = Vec::new();
+        for entry in entries {
```
kinto0, line 4824: can this be in it's own code_lens file

lsp/src/codeLens.ts
asukaminato0721: done

lsp/src/codeLens.ts
asukaminato0721: now deduped the runnable-CodeLens error prompts

pyrefly/lib/lsp/non_wasm/server.rs
asukaminato0721: done

pyrefly/lib/lsp/non_wasm/server.rs
asukaminato0721: moved

pyrefly/lib/lsp/non_wasm/code_lens.rs
asukaminato0721: checked, seems no.

lsp/package.json
asukaminato0721: That broader LSP knob already exists, but it disables all CodeLens.

python.pyrefly.runnableCodeLens is narrower and still useful.

lsp/src/codeLens.ts
asukaminato0721: I think they are now small enough?

PR #2403, opened 2026-02-13
[conformance] Handle generic `__init__` when converting to callable

fangyi-zhou:
# Summary

Fix the issue that generic constructors were not handled correctly when they are used as callabls.

Previously, we override the return type of the `__init__` function, and return the bound method type. Then, the `subset.rs` will then call `bind_function` to bind the method type. In between, generics were re-instantiated, and the returning type was not correct.

In this diff, we consolidate the return type changing, and then bind the constructor with a flag to skip instantiation. As a result, the generics in constructor is handled correctly.

The change is mostly generated by Claude, but I have verified it myself.

LLM summary:
  - Fix constructor-to-callable conversion for generic classes by reusing `bind_function` instead of the ad-hoc `set_callable_return_type_for_constructor` helper. This properly strips `self`, sets the return type,
   and preserves type variables for inference at the call site.
  - Add a `skip_instantiation` flag to `bind_function` so that when converting `__init__` to a callable signature, type variables remain uninstantiated (they get inferred later when the callable is actually
  called).
  - Remove `set_callable_return_type_for_constructor` from `pyrefly_types` — its logic is now handled inline in `call.rs` and by the new `bind_dunder_init_for_callable` method.

# Test plan

  - Added `test_init_to_callable_generics` covering generic classes with `__init__[V](self: C[V], x: V)` converted to `Callable[[V], C[V]]`
  - Updated `test_constructor_callable_conversion`: `Class3` with both `__new__` and `__init__` now correctly passes as a callable (removed false-positive error). Bug marker narrowed to remaining issues with
  overloads and `__new__`.
  - Conformance: 4 fewer differences in constructors_callable.py (79 total, down from 83)

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1308,22 +1308,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         };
         // Check the __init__ method and whether it comes from object or has been overridden
         let (init_attr_ty, overrides_init) = if let Some(t) = self.get_dunder_init(cls, false) {
-            let t = if let Type::BoundMethod(ref method) = t {
-                // Strip self param and set return type to self's type (for generic handling)
-                self.bind_dunder_init_for_callable(method)
-                    .unwrap_or_else(|| {
-                        // Fallback: just set return type without stripping
-                        let ret_type = t
-                            .callable_first_param(self.heap)
-                            .unwrap_or_else(|| class_type.clone());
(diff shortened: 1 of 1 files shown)


fangyi-zhou:
To be honest, I don't understand exactly how this change fixes the issue with `Class3` in the conformance test, but might be a reduced example of the deltas observed in the mypy-primer results.

But I did notice earlier that `Class3` is bit of a particular case. https://github.com/python/typing/issues/2158

Issue #2392, opened 2026-02-12
Invalid output when executed inside github worker


abelcheung:
Any content that allows `pyrefly` to produce output can cause problem in this issue, such as placing following one-liner inside file:

```python
reveal_type(1)
```

To simulate github worker environment, export following environment variable in local shell. (the env var is automatically set inside github workers)

```bash
GITHUB_ACTIONS='true'
```

The issue is that, GitHub formatted annotation is unconditionally printed, even when user requests other output formats. For example, following command creates illegal JSON data because extra content is attached to end of JSON output:

```bash
> pyrefly check --output-format=json --summary=none testbug.py
{
  "errors": [
    {
      "line": 1,
      "column": 1,
      "stop_line": 1,
      "stop_column": 12,
      "path": "testbug.py",
      "code": -2,
      "name": "unknown-name",
      "description": "`reveal_type` must be imported from `typing` for runtime usage",
      "concise_description": "`reveal_type` must be imported from `typing` for runtime usage",
      "severity": "error"
    },
    {
      "line": 1,
      "column": 12,
      "stop_line": 1,
      "stop_column": 15,
      "path": "testbug.py",
      "code": -2,
      "name": "reveal-type",
      "description": "revealed type: Literal[1]",
      "concise_description": "revealed type: Literal[1]",
      "severity": "info"
    }
  ]
}::error file=testbug.py,line=1,col=1,endLine=1,endColumn=12,title=Pyrefly unknown-name::`reveal_type` must be imported from `typing` for runtime usage
::notice file=testbug.py,line=1,col=12,endLine=1,endColumn=15,title=Pyrefly reveal-type::revealed type: Literal[1]
```

I believe this bug is present since 0.47.0.

yangdanny97:
Thanks for reporting!

I saw this in the mypy primer output, and was wondering what this is

yangdanny97:
This will be fixed in the next release

GH actions output now requires `--output-format=github` to be passed in

matdibu:
@yangdanny97 I think the output needs to be tweaked slightly

the `file` attribute in `--output-format=github` output seems to work only when it's an absolute path

I've simulated an error to compare the output between `pyrefly` and other similar tools

`pyrefly` produced this, which did not show up as an error in the `Files changed` tab
```
::error file=src/foo/bar.py,line=19,col=12,endLine=19,endColumn=13,title=Pyrefly bad-return::Returned type `Literal[1]` is not assignable to declared return type `str`
```

[...]

PR #2411, opened 2026-02-13
use absolute path in github output format

yangdanny97:
Summary: followup for https://github.com/facebook/pyrefly/issues/2392

Differential Revision: D93287045


grievejia (approved):

PR #2413, opened 2026-02-14
[conformance] Raise error when `Self` appears in some invalid locations

fangyi-zhou:
# Summary

This diff implements checks mentioned in the [spec for valid locations of `Self`](https://typing.python.org/en/latest/spec/generics.html#valid-locations-for-self), raising errors when `Self` appears in invalid locations:

- Outside a class
- In a static method within a class
- In a metaclass (subclass of `type`), including both method signatures and class attributes

The metaclass check is done at `SelfTypeLiteral` resolution time, which catches all explicit `Self` usages (attributes, method params, return types) and reports the error on the `Self` token itself. The static method check is done during function type decoration, reporting on the function name.

A new `is_metaclass` field is added to `ClassMetadata` to track whether a class is a subclass of `type`.

# Test Plan

- Added unit tests for all three invalid locations
- All existing metaclass tests (30) continue to pass
- Conformance: `generics_self_usage.py` down from 6 differences to 2 (remaining are unrelated)

### DIFF
--- a/pyrefly/lib/test/callable.rs
+++ b/pyrefly/lib/test/callable.rs
@@ -908,7 +908,7 @@ zoo(partial(bar, b=99))
 );
 
 testcase!(
-    bug = "Self in Metaclass should be error, treated as Any. Any in metaclass call should act like no annot.",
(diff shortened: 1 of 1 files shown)


stroxler (approved):

PR #2417, opened 2026-02-14
impl type gen #2357

asukaminato0721:
# Summary

Fixes #2357

Added call-site inference for generated function/class signatures (including `__init__`) and optional variable annotations.

# Test Plan

Updated expectations for TypeVar('T') and added a new call-site inference test.


asukaminato0721:
@MartinBernstorff

stroxler (approved):

PR #2423, opened 2026-02-14
impl pyrefly ignore #2207

### LINKED ISSUE #675: Feature request: Add codeactions (ignore error)

asukaminato0721:
# Summary

Fixes part of #2207
fix https://github.com/facebook/pyrefly/issues/675
Added a quick fix that inserts # pyrefly: ignore [error-code], merging into an existing pyrefly ignore on the same line or comment-only line above when present.

# Test Plan

Added LSP code action tests covering insertion and merge behavior.


migeed-z:
cc @maggiemoss

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,119 @@
...
+use ruff_text_size::TextSize;
+use starlark_map::small_set::SmallSet;
+
+use crate::ModuleInfo;
+use crate::config::error_kind::ErrorKind;
+use crate::error::error::Error;
+
+static PYREFLY_IGNORE_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"#\s*pyrefly:\s*ignore\s*\[([^\]]*)\]").unwrap());
+
```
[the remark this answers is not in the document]
jvansch1: @asukaminato0721 I agree with this comment. Whatever regex we use to match ignores needs to be able to handle all variations of ignores.

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,158 @@
...
+pub(crate) fn add_pyrefly_ignore_code_action(
+    module_info: &ModuleInfo,
+    error: &Error,
+) -> Option<(String, Module, TextRange, String)> {
+    if module_info.is_notebook() || module_info.is_generated() {
+        return None;
+    }
+    if error.error_kind() == ErrorKind::UnusedIgnore {
```
jvansch1: We can consolidate this check with the one on line 30.

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,158 @@
...
+    let line_text = line_text_full
+        .strip_suffix("\r\n")
+        .or_else(|| line_text_full.strip_suffix('\n'))
+        .unwrap_or(line_text_full);
+    let start = module_info.lined_buffer().line_start(line);
+    let end = start + TextSize::from(line_text.len() as u32);
+    Some((TextRange::new(start, end), line_text))
+}
+
+fn pyrefly_ignore_codes(line: &str) -> Option<Vec<String>> {
```
jvansch1: Can we make this function name a little more descriptive? Something like `get_current_pyrefly_ingore_codes`

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,158 @@
...
+        let updated_line = replace_pyrefly_ignore_comment(above_text, &new_comment)?;
+        return Some((title, module_info.dupe(), above_range, updated_line));
+    }
+
+    let insert_range = TextRange::new(line_range.end(), line_range.end());
+    let insert_text = format!("  # pyrefly: ignore [{error_code}]");
+    Some((title, module_info.dupe(), insert_range, insert_text))
+}
+
+fn line_text_and_range(module_info: &ModuleInfo, line: LineNumber) -> Option<(TextRange, &str)> {
```
jvansch1: Can we name this something like `get_line_text_and_range`?

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,158 @@
...
+    let codes = inside
+        .split(',')
+        .map(|code| code.trim())
+        .filter(|code| !code.is_empty())
+        .map(str::to_owned)
+        .collect::<Vec<_>>();
+    Some(codes)
+}
+
+fn pyrefly_ignore_blank(line: &str) -> bool {
```
jvansch1: Can we add a docstring here explaining this function? 

Also maybe a name like `should_ignore_blank_error_codes` is better? It's hard for me to suggest a name because right now I am a little confused about exactly what the intention behind this function is. Are we trying to filter out cases where we have pyrefly ignores without any error code?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -578,6 +578,14 @@ class my_export:
     pass
 my_export
 # ^
+# Title: Add `# pyrefly: ignore [unknown-name]`
```
jvansch1, line 615: What would happen if we have a line that already has a comment after it?

```python
x: int = "abc" # intentional error
``` 

Would we still append the full comment with another `#` character? Maybe we need a test for that.

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
jvansch1: I also feel like if we are trying to detect ignores without any specific error codes, could we just use a regex instead?

pyrefly/lib/state/lsp.rs
```diff
@@ -1931,9 +1931,18 @@ impl<'a> Transaction<'a> {
         let mut generate_actions = Vec::new();
         let mut other_actions = Vec::new();
         for error in errors {
+            let error_range = error.range();
+            if error_range.contains_range(range)
+                && let Some(action) = quick_fixes::pyrefly_ignore::add_pyrefly_ignore_code_action(
+                    &module_info,
+                    &error,
+                )
+                && !other_actions.iter().any(|existing| existing == &action)
```
jvansch1: Do we need to iterate through every other action here? I think in most cases this wont be a huge performance issues but it does not seems very efficient. Could we use a hashmap or set for a quicker lookup?

jvansch1 (changes requested):

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,151 @@
...
+    let line_text = line_text_full
+        .strip_suffix("\r\n")
+        .or_else(|| line_text_full.strip_suffix('\n'))
+        .unwrap_or(line_text_full);
+    let start = module_info.lined_buffer().line_start(line);
+    let end = start + TextSize::from(line_text.len() as u32);
+    Some((TextRange::new(start, end), line_text))
+}
+
+fn get_current_pyrefly_ignore_codes(line: &str) -> Option<Vec<String>> {
```
jvansch1: This function is very similar to `parse_ignore_comment` in lib/error/suppress.rs. It would be a good idea to investigate if we can re-use logic here.

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -806,6 +830,131 @@ class my_export:
...
+x: int = "hello"  # pyrefly: ignore [bad-assignment]
+#         ^
+"#
+        .trim(),
+        report.trim()
+    );
+}
+
+#[test]
+fn quickfix_add_pyrefly_ignore_code_with_existing_comment() {
```
jvansch1, line 916: @maggiemoss How do you think ignores should function in this case?

jvansch1 (changes requested):

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,151 @@
...
+    let Some(captures) = PYREFLY_IGNORE_COMMENT_REGEX.captures(comment_part) else {
+        return false;
+    };
+    let Some(group) = captures.get(1) else {
+        return true;
+    };
+    group.as_str().trim().is_empty()
+}
+
+fn merge_pyrefly_ignore_codes(existing: Vec<String>, new_code: &str) -> Vec<String> {
```
jvansch1: Same thing here. This is very similar to `merge_error_codes` in lib/error/suppress.rs`

jvansch1 (changes requested):

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
```diff
@@ -0,0 +1,127 @@
...
+        .or_else(|| line_text_full.strip_suffix('\n'))
+        .unwrap_or(line_text_full);
+    let start = module_info.lined_buffer().line_start(line);
+    let end = start + TextSize::from(line_text.len() as u32);
+    Some((TextRange::new(start, end), line_text))
+}
+
+/// Returns true if the line already has a `# pyrefly: ignore` comment without any codes.
+fn has_blank_pyrefly_ignore_comment(line: &str) -> bool {
+    let Some(comment_start) = find_comment_start_in_line(line) else {
```
jvansch1: I wonder if it would just be simpler to use a regex here instead of doing all of this logic?

jvansch1 (commented):
Overall looks pretty good to me. There is one outstanding question I would like @maggiemoss to weigh in on before fully approving this.

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
asukaminato0721: could use a regex, but it wouldn’t really be simpler here.

Even with a regex, still need find_comment_start_in_line to avoid matching inside string literals, and then need a fairly careful pattern to distinguish:

`# pyrefly: ignore (blank, should return true)`
`# pyrefly: ignore [bad-assignment] (not blank)`
`# pyrefly: ignore_me (not a valid directive)`

That ends up being roughly the same complexity, just shifted into a regex with negative lookaheads and word‑boundary edge cases.

kinto0 (approved):

pyrefly/lib/test/lsp/code_actions.rs
maggiemoss: I think the simplest solution (that is how the existing logic works) is just to add the suppression on the line above. Keep the comments on a separate line and then we don't have to have all of this additional logic

pyrefly/lib/state/lsp/quick_fixes/pyrefly_ignore.rs
maggiemoss: I think a regex might be easier for folks to understand though. This code is kind of hard to grasp while scrolling through the file.

maggiemoss: Also your first example we'd never need a quick fix for, so I think it can be simplified to just handling the second example when we need to add a new error code.

kinto0:
@jvansch1 any reason this can't be merged? looks like there's been an update

PR #2424, opened 2026-02-14
fix Detect incorrect overload defaults #651

### LINKED ISSUE #651: Detect incorrect overload defaults
Same request as in https://github.com/python/mypy/issues/19411#issue-3215174168, but just the "detect incorrect default" part, as the "missing default" part might not be a good idea after all

(as discussed on discord)

Example:
```py
from typing import Literal, overload, reveal_type

@overload
def foo(a: Literal[True] = ...) -> None: ...
@overload
def foo(a: Literal[False]) -> int: ...
def foo(a: bool = False) -> None | int:
    return 1 if not a else None

reveal_type(foo())
```

I'd like Pyrefly to flag that `a: Literal[True] = ...` is incorrect, as the default (`False`) has no overlap with `Literal[True]`

asukaminato0721:
# Summary

Fixes #651

Implemented overload default compatibility checking so overloads that declare a default now validate against the implementation’s actual default value. This flags the incorrect Literal[True] = ... case from issue #651.

# Test Plan

Added a targeted test.

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -1601,6 +1601,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             sig.ret = self.heap.mk_any_implicit();
             sig
         };
+        let mut overload_default_candidates: SmallMap<Name, Vec<(TextRange, Type, Type)>> =
+            SmallMap::new();
+        let mut overload_default_matches: SmallSet<Name> = SmallSet::new();
         for (range, overload) in overloads.iter() {
             let (overload_tparams, original_overload_func) = match overload {
                 OverloadType::Function(func) => (None, func),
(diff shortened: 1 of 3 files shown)


pyrefly/lib/alt/function.rs
```diff
@@ -1545,6 +1545,40 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                    let (overload_name, overload_ty, overload_required) = match overload_param {
+                        Param::PosOnly(Some(name), ty, required)
+                        | Param::Pos(name, ty, required)
+                        | Param::KwOnly(name, ty, required) => (name, ty, required),
+                        _ => continue,
+                    };
+                    if !matches!(overload_required, Required::Optional(None)) {
+                        continue;
+                    }
+                    let impl_default = impl_params.items().iter().find_map(|param| match param {
```
rchen152: This iterates through all of impl_params for every overload_param. Could we build a name => default map for the implementation signature upfront instead?

pyrefly/lib/test/overload.rs
```diff
@@ -152,6 +152,20 @@ assert_type(f(h(0)), str) # E: Argument `Literal[0]` is not assignable to parame
 "#,
 );
 
+testcase!(
+    test_overload_default_incompatible,
+    r#"
+from typing import Literal, overload
+
+@overload
+def foo(a: Literal[True] = ...) -> None: ...  # E: Default `Literal[False]` is not assignable to overload parameter `a` with type `Literal[True]`
```
rchen152:
```suggestion
def foo(a: Literal[True] = ...) -> None: ...  # E: Inconsistent types for parameter `a`: default `Literal[False]` from implementation is not assignable to annotation `[Literal[True]]` from overload
```

IMO the error message should mention that the default is from the implementation. I'd suggest something like this, but feel free to reword it if you think of something better.

rchen152:
Ah, actually, looking at the mypy_primer results, I think the logic needs to be adjusted a bit. A case like:
```
@overload
def f(x: None = ...): ...
@overload
def f(x: int = ...): ...
def f(x: int | None = None): ...
```
is ok because the `None` default matches the first overload.

Issue #2126, opened 2026-01-15
`includeDeclaration` is not respected when enumerating references


acooper-akamai:
### Describe the Bug

Most LSPs support a parameter `includeDeclaration` when enumerating references which informs the LSP to include or exclude the symbol's declaration in its references. I prefer to have it off, but I don't seem to be able to do so with Pyrefly. 

As a simple example, in the code
```python
def foo() -> None:
    ...

foo()
foo()
```
Opening the references in neovim shows the following:
```
foo.py|1 col 5-8| def foo() -> None:
foo.py|4 col 1-4| foo()
foo.py|5 col 1-4| foo()
```

I'd prefer it to show just the bottom 2 lines, but passing `includeDeclaration = false` to the native LSP interface doesn't seem to do it for me. This does work for me with other LSPs like rust-analyzer. The full command to test is
```
:lua vim.lsp.buf.references({ includeDeclaration = false })
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGLrgBQCUNAtAHw0Byu6MiADroaomsQBOEdABdGg8AwXNhw%2BkxXp1LYSAA0IAK4zocEuUQgAxDQCqJqBBmk6h9AGMTfOKqzU6uBIAtqgyAProhkHYMBKM%2BIg00jKsnDRwMhJCImISMDKGEiJgCtxRMVk0wPgAvgq6BmR5YFCkhDK4QVAUNgAKpM2t6Rg4BDTufJAA5oWhEHyEwjYAyjAwNAAWMjLEcIgA9PtN1K2EgVP7MOj7mLjucPsTWhAzEnN8%2BwESNKgAbqjQVDYWDjSYvWZeES4YiQszCMgyDZ8Ni-WJweYiAC8NAUAGZCABGABM9XQIBqBlQnggqIAYtAYBQ0Fg8EQyOSgA

### (Only applicable for extension issues) IDE Information

```
$ nvim --version
NVIM v0.11.4
Build type: Release
LuaJIT 2.1.1741730670
```

DanielNoord:
Got a PR to fix this at https://github.com/facebook/pyrefly/pull/2427

Could somebody assign this to me? :)

PR #2427, opened 2026-02-14
Take `includeDeclaration` into account

DanielNoord:
# Summary

Closes https://github.com/facebook/pyrefly/issues/2126

Not sure if just passing the `bool` around is the best, but I guess it works? :smile:

The logic for `references` might also not be fully idiomatic. So very open to feedback and suggestions.

# Test Plan

I copied `test_references_cross_file_no_config` from the test file as it seems to use a module that is pretty simple. From the expected test output I removed the declaration, and that test now passes.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -3745,13 +3745,15 @@ impl Server {
         let Some(handle) = self.make_handle_if_enabled(uri, Some(Rename::METHOD)) else {
             return self.send_response(new_response::<Option<WorkspaceEdit>>(request_id, Ok(None)));
         };
+        let path = self.path_for_uri(uri).unwrap();
         self.async_find_references_helper(
             request_id,
             transaction,
             handle,
             uri,
             params.text_document_position.position,
-            true,
+            self.workspaces
+                .get_with(path, |(_, workspace)| workspace.include_declaration),
             move |results| {
                 let mut changes = HashMap::new();
                 for (uri, ranges) in results {
--- a/pyrefly/lib/lsp/non_wasm/workspace.rs
+++ b/pyrefly/lib/lsp/non_wasm/workspace.rs
@@ -65,6 +65,7 @@ pub struct Workspace {
     pub disable_language_services: bool,
     pub disabled_language_services: Option<DisabledLanguageServices>,
     pub display_type_errors: Option<DisplayTypeErrors>,
+    pub include_declaration: bool,
     pub lsp_analysis_config: Option<LspAnalysisConfig>,
 }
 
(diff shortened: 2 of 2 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -3748,6 +3751,7 @@ impl Server {
             handle,
             uri,
             params.text_document_position.position,
+            true,
```
kinto0: we probably need to allow this to be configured via the lsp config `workspace_configuration_response`

kinto0 (changes requested):
looks great! now we just need to let people set this

pyrefly/lib/lsp/non_wasm/server.rs
DanielNoord: Sorry for the late response. I haven't touched the configuration code before.

I think you mean to add it to so it can be set:
https://github.com/facebook/pyrefly/blob/aeb2b1095a16888ede0bd1281eb9e442e430b8ff/pyrefly/lib/lsp/non_wasm/workspace.rs#L283-L291

But I'm not quite sure in which section it should be?
Or should it be in `Workspace`?

Happy to make the change, just don't really know where I should make it 😅

kinto0: if we put it in the PyreflyClientConfig, it'll be accessible by LSP via `configuration` and `initializationOptions`. see the tip in our wiki [here](https://pyrefly.org/en/docs/IDE/#customization) (we probably should make this info easier to understand and find)

DanielNoord: @kinto0 Added it, but I'm not 100% confident in the change 😅 

Not sure where the config belongs to, and I think with the current change it will default to `false` whereas `main` uses `true` considering it doesn't even look at the argument.

Any pointers are welcome! :)

kinto0: I'm sorry, I misunderstood this part! `includeDeclaration` is only part of references params, so the question is what to do with rename. I think the old behavior is the only correct behavior, since renaming must include declarations otherwise the references would reference the wrong thing

kinto0 (approved):
approving the first commit before I suggested changes. let me try importing + resolving merge conflicts

kinto0:
now that i'm looking at it more directly to resolve conflicts, I'm noticing a few other things:
- local references still include declarations (I added tests for this + fixed it by adding the bool to local references function)
- global references calls local references. we only need this logic in local references functionality. and we don't even have to filter, we just need to avoid explicitly adding the declaration [here](https://github.com/facebook/pyrefly/blame/0d0ea90a250a0018094e155b2bf65669792214f7/pyrefly/lib/state/lsp.rs#L2570)

i'm just fixing these up since they're quick and i've already resolved the conflicts

DanielNoord:
Thanks @kinto0 

Still getting used to the codebase. Happy to see they trigger further improvements :)

PR #2428, opened 2026-02-14
[conformance] Handle protocol subtyping checks with instances & objects

fangyi-zhou:
# Summary

This diff addresses a number of issues when whether classes / objects are subtypes of a protocol.

- Checks whether an attribute is a `ClassVar` during subtyping.
  + For class objects, a `ClassVar` attribute corresponds to a non-`ClassVar` attribute in protocol.
  + For class instances, add checks to ensure that attributes are matched by their `ClassVar`-ness.
- `__dataclass_fields` attribute for dataclasses is now synthesised as a `ClassVar` instead of an instance attribute.
- Implicit attributes in protocols are no longer checked during subtyping checks.

# Test Plan

Ran conformance testsuite -> 7 fewer failing tests


stroxler (approved):

Issue #2116, opened 2026-01-15
panic: key lacking binding, module=__unknown__ with fuzzed assignment


correctmost:
### Describe the Bug

Pyrefly crashes when checking this fuzzed code:

```python
@:a=1
```

```
Thread panicked, shutting down: panicked at pyrefly/lib/binding/bindings.rs:311:13:
Internal error: key lacking binding, module=__unknown__, path=/home/user/a.py, key=Key::Definition( 1:2), key-debug=Definition(ShortIdentifier(1..1))
```

### Version info
62e54d65b343

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAAoqgLwCMAOuiADQgCuALtHCeYhABiAAQBVHlAhdSIsB3QBjHrnRxmzTDDBzcAJwC2qLgH10HA9hh6AFPkQiI6LgEoRAWgB8IuFz2JmESCRPRguDj10OUYQADkLK38RYHwAXxjmNhAyULAoUkIuXAMoClEABVJc-J8MHAIRRVVIAHMI4whVQmZRAGUYGBEACy4uYjhEAHpJnO18wn0WyZh0ScxcRThJpvRW9pVV3T0RVAA3VGhUbFhG5og2vQ7VEVxiA75mMi4h1XdT6zgnSitBEMQAzIR6AAmDIsVLsVDKCD-ABi0BgFDQWDwRDIIFSQA

stroxler:
Thanks!

DanielNoord:
Got a PR to fix this at https://github.com/facebook/pyrefly/pull/2435

Can I be assigned to this issue?

PR #2435, opened 2026-02-15
Do not insert Bindings for broken definitions

DanielNoord:
# Summary

Closes https://github.com/facebook/pyrefly/issues/2116

This avoid returning this "expression" here:
https://github.com/facebook/pyrefly/blob/02dcdc40150998bdbd915fcb2f8051c50a1161f6/pyrefly/lib/binding/bindings.rs#L515
Which avoids adding a `Binding` for something that really doesn't exist.

Not sure if this is the best preferred way to handle these panics and bogus code.

# Test Plan

Added a testcase. I did run `test.py` but it did not generate any changes. Therefore, I just added the error codes myselff. That seems to have worked.

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -2518,6 +2518,12 @@ impl ScopeTrace {
         let mut exportables = SmallMap::new();
         let scope = self.toplevel_scope();
         for (name, static_info) in scope.stat.0.iter_hashed() {
+            // Definitions with empty names are not actually accessible and should not be considered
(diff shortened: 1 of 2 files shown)


rchen152 (approved):
Thanks!

rchen152:
> Added a testcase. I did run test.py but it did not generate any changes. Therefore, I just added the error codes myselff. That seems to have worked.

FYI, test.py runs the tests but does not update expectations in them, so updating the expectations manually is the right thing to do. (It *does* update pyrefly's results on our copy of the typing conformance tests in https://github.com/facebook/pyrefly/tree/main/conformance.)

PR #2441, opened 2026-02-16
fix Feature: show source of type narrowing/inference #958

### LINKED ISSUE #958: Feature: show source of type narrowing/inference
### Describe the Bug

Long term we should build the infrastructure to trace where types are narrowed/inferred.

Being able to show the user "this type was narrowed on this line" or "this empty container type was pinned by this usage" will be useful to reduce confusion caused by type inference.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #958

image

Added a hover “Type source” section that reports narrowing and first‑use inference origins with location + snippet.

# Test Plan

Added an LSP hover test covering narrowing and first‑use inference output.

### DIFF
--- a/pyrefly/lib/test/lsp/hover.rs
+++ b/pyrefly/lib/test/lsp/hover.rs
@@ -244,8 +244,15 @@ def f(x: int | None) -> None:
     y
 #   ^
 "#;
-    let report = get_batched_lsp_operations_report(&[("main", code)], get_test_report);
-    let report = normalize_typeshed_paths(&report);
+    let report = get_batched_lsp_operations_report(&[("main", code)], |state, handle, position| {
+        match get_hover(&state.transaction(), handle, position, false) {
+            Some(Hover {
+                contents: HoverContents::Markup(markup),
+                ..
+            }) => markup.value,
+            _ => "None".to_owned(),
+        }
+    });
     assert_eq!(
         r#"
 # main.py
@@ -259,8 +266,6 @@ def f(x: int | None) -> None:
 - Narrowed by condition at 3:13: `x is not None`
 
 
-Go to [int](file:///tmp/pyrefly_bundled_typeshed/builtins.pyi#L420,7)
-
 9 |     y
         ^
 ```python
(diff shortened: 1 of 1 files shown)


pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
...
+}
+
+fn format_inline_code(snippet: &str) -> String {
+    format!("`{snippet}`")
+}
+
+fn format_narrow_op(module: &Module, base_name: &Name, op: &NarrowOp) -> Option<String> {
+    fn subject_with_facet(
+        base_name: &Name,
+        facet: Option<&crate::binding::narrow::FacetSubject>,
```
kinto0: nit: can we `use` this instead of inlining the crate?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
         .map(|id| id.identifier.id.to_string())
 }
 
+fn format_type_source_location(module: &Module, range: TextRange) -> String {
+    format!("{}", module.display_pos(range.start()))
```
kinto0: nit: is this the same as `.to_string()`? can we use that here instead?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
...
+            .nth(max_chars)
+            .map(|(idx, _)| idx)
+            .unwrap_or_else(|| truncated.len());
+        truncated.truncate(end_byte);
+        truncated.push_str("...");
+    }
+    Some(truncated)
+}
+
+fn format_inline_code(snippet: &str) -> String {
```
kinto0: this function doesn't seem useful. should we inline uses instead?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
...
+        .split_whitespace()
+        .collect::<Vec<_>>()
+        .join(" ")
+        .replace('`', "'");
+    if cleaned.is_empty() {
+        return None;
+    }
+    const MAX_LEN: usize = 80;
+    let mut truncated = cleaned;
+    if truncated.len() > MAX_LEN {
```
kinto0: let's use char_indices here too if we use it below on line 356?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
...
+        }
+        match bindings.get(current) {
+            Binding::Forward(next)
+            | Binding::Narrow(next, ..)
+            | Binding::CompletedPartialType(next, ..)
+            | Binding::PartialTypeWithUpstreamsCompleted(next, ..)
+            | Binding::LoopPhi(next, ..) => current = *next,
+            Binding::Phi(_, branches) if !branches.is_empty() => {
+                current = branches[0].value_key;
+            }
```
kinto0, line 609: what happens if different branches define different variables?

pyrefly/lib/test/lsp/hover.rs
```diff
@@ -231,6 +231,42 @@ a: int = "test"  # type: ignore
...
+        return
+    y = []
+    y.append(1)
+    x
+#   ^
+    y
+#   ^
+"#;
+    let report = get_batched_lsp_operations_report(&[("main", code)], get_test_report);
+    assert!(
```
kinto0: nit: can we test for an exact match instead? I think it's easier to understand

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
...
+        seen.insert(current);
+        match bindings.get(current) {
+            Binding::Forward(next) => current = *next,
+            Binding::Narrow(_, op, _) => {
+                let key = bindings.idx_to_key(current);
+                let Key::Narrow(name, op_range, _) = key else {
+                    return None;
+                };
+                let location = format_type_source_location(module, *op_range);
+                let mut msg = format!("Narrowed by condition at {location}");
```
kinto0, line 576: is there a way to add a hyperlink so it can be clicked?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -306,6 +330,344 @@ fn identifier_text_at(
...
+    }
+
+    fn args_snippet(module: &Module, args: &ruff_python_ast::Arguments) -> Option<String> {
+        format_code_snippet(module, args.range())
+    }
+
+    match op {
+        NarrowOp::Atomic(facet, atomic) => {
+            let subject = subject_with_facet(base_name, facet.as_ref());
+            let formatted = match atomic {
```
kinto0, line 406: I wonder if it's easier to just say "narrowed at <location>" to avoid this. if it's clickable / easy to find, it doesn't seem to add much to have this match

kinto0 (commented):
some nits but I think I'd like a review from someone more familiar with the actual. narrowing logic like @stroxler

pyrefly/lib/lsp/wasm/hover.rs
asukaminato0721: the raw op_range doesn’t always include the full condition (we’ve already seen it collapse to None).

pyrefly/lib/lsp/wasm/hover.rs
asukaminato0721: `Binding::Phi` is created per-name; all branches are alternate values for the same name.
Using the first branch is a cheap way to reach a `Key::Definition`.

stroxler (approved):
This looks reasonable to me, aiming to get a second review and merge soon

kinto0 (approved):

PR #2443, opened 2026-02-16
fix Semantic highlighting needs some improvement #1324

### LINKED ISSUE #1324: Semantic highlighting needs some improvement
Hi, first thank you to create Pyrefly, an alternative solution for IDE that are VSCode based.
I am using Positron, and find Pyrefly is extremely good. I understand Pyrefly is still in alpha stage.
First, the performance is relatively slow comparing to Pylance, as the semantic highlighting take a while in order to show.
Secondly, some semantic highlighting needs improvement, see example below:

### Class and method access from module, like pd.DataFrame, pd.read_excel

Pylance

Image

Pyrefly

Image

### selfParameter

Pylance

Image

Pyrefly

Image

### Method builtin

Pylance

Image

Pyrefly

Image

asukaminato0721:
# Summary

Fixes #1324

Improved attribute semantic token classification: class for Type::Type, interface for type aliases, and method vs function based on callable metadata.

Added a selfParameter modifier for self/cls parameters and updated semantic token expectations.

# Test Plan

Adjusted class attribute token in method_and_property_test to be method.

### DIFF
--- a/pyrefly/lib/state/semantic_tokens.rs
+++ b/pyrefly/lib/state/semantic_tokens.rs
@@ -17,6 +17,7 @@ use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_python::symbol_kind::SymbolKind;
 use pyrefly_python::sys_info::SysInfo;
+use pyrefly_types::callable::FunctionKind;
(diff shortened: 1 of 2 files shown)


kinto0 (approved):

grievejia (approved):

PR #2460, opened 2026-02-18
fix Pyrefly extension not supported for VSCode "browser" mode (vscode.dev) #2240

### LINKED ISSUE #2240: Pyrefly extension not supported for VSCode "browser" mode (vscode.dev)
### Describe the Bug

TIL if you use VSCode Web in the browser (e.g. such as when you use `.` to open a GitHub file in the web ui) it's not possible to install the Pyrefly extension there. It looks like it might be something [we could configure](https://code.visualstudio.com/api/extension-guides/web-extensions) from a quick skim of the docs.

I'm not sure how many VSCode users frequently use VSCode Web, or how complicated it would be to implement

Image

cc @kinto0

asukaminato0721:
# Summary

Fixes #2240

Implemented a web worker–based LSP for vscode.dev using the Pyrefly WASM playground API. This enables hover, completion, definitions, inlay hints, semantic tokens, and diagnostics in VS Code Web.

Added web activation that spins up the worker and indexes the workspace (with size limits) to seed the WASM state.

Added a WASM stub + build wiring so real WASM artifacts can be dropped in, and the build copies the .wasm into dist.

# Test Plan

image

image


asukaminato0721:
1) Build the WASM

  cd pyrefly_wasm
  ./build.sh

  2) Copy the JS binding into the extension

  cp pyrefly_wasm/target/pyrefly_wasm.js lsp/wasm/pyrefly_wasm.js

  3) Build the extension

  cd lsp
  npm install
  npm run compile

then `code .`

image

pop a window. Test like normal case.

kinto0:
> 1. Build the WASM
> 
> cd pyrefly_wasm ./build.sh
> 
> 2. Copy the JS binding into the extension
> 
> cp pyrefly_wasm/target/pyrefly_wasm.js lsp/wasm/pyrefly_wasm.js
> 
> 3. Build the extension
> 
> cd lsp npm install npm run compile
> 
> then `code .`
> 
> image
> pop a window. Test like normal case.

I wonder if you could add this to a dev instructions markdown file somewhere?

additionally, I don't really understand how this gets distributed in it's current form. do we need to update the publish-vsix action to publish a wasm?

lsp/package.json
```diff
@@ -257,7 +259,7 @@
         "vsce": "^2.15.0",
         "vscode-languageclient": "9.0.1"
     },
```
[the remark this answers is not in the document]
kinto0: yeah i'm not sure this is correct. I don't think VSCode will install python which is needed for interpreters. is there a way to have a separate package.json for browser?

lsp/src/pyrefly-web-worker.ts
```diff
@@ -0,0 +1,468 @@
...
+    params.position.line + 1,
+    params.position.character + 1,
+  );
+  if (!ranges.length) {
+    return null;
+  }
+  return ranges.map(range => ({
+    uri: params.textDocument.uri,
+    range: toLspRange(range),
+  }));
```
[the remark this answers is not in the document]
kinto0: this is a dealbreaker. maybe it's necessary to improve the sandbox to work cross-file first?

lsp/wasm/pyrefly_wasm.js
```diff
@@ -0,0 +1,645 @@
+let wasm;
```
kinto0: when is this generated? how? can it have a @generated marker?

lsp/tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "compilerOptions": {
         "module": "commonjs",
         "target": "es2019",
-        "lib": ["ES2019"],
+        "lib": ["ES2019", "DOM"],
```
kinto0: looks like this leaks a lot of info

instead, what if we use a separate tsconfig.web.json for the browser sources, or use per-file /// <reference lib="dom" /> directives.

lsp/src/pyrefly-web-worker.ts
```diff
@@ -0,0 +1,468 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+import {
```
kinto0, line 28: could you add a doccomment for how this works at the top?

lsp/src/pyrefly-web-worker.ts
```diff
@@ -0,0 +1,468 @@
...
+    startLineNumber: range.start.line + 1,
+    startColumn: range.start.character + 1,
+    endLineNumber: range.end.line + 1,
+    endColumn: range.end.character + 1,
+  };
+}
+
+function toDiagnosticSeverity(severity: number): DiagnosticSeverity {
+  switch (severity) {
+    case 8:
```
kinto0, line 339: can you leave comments for what the numbers mean and where to find the corresponding values?

lsp/src/pyrefly-web-worker.ts
```diff
@@ -0,0 +1,468 @@
...
+  const state = await ensureWasmState();
+  if (!state) {
+    return;
+  }
+  const filename = uriToFilename(closeParams.textDocument.uri);
+  const content = files.get(filename);
+  if (!content) {
+    return;
+  }
+  state.updateSingleFile(filename, content);
```
kinto0: should we also remove the file from `files`?

kinto0 (changes requested):
looks really great, but I have a few concerns. thanks for picking this up!

lsp/src/extension.ts
```diff
@@ -117,7 +120,22 @@ export async function activate(context: ExtensionContext) {
     process.platform === 'win32' ? 'pyrefly.exe' : 'pyrefly',
   );
 
-  let pythonExtension = await PythonExtension.api();
+  let pythonExtension: PythonExtension | undefined;
```
asukaminato0721: since vscode don't support have two package.json, I show warning to install the extension.

lsp/src/pyrefly-web-worker.ts
```diff
@@ -0,0 +1,511 @@
...
+      return;
+    }
+    const filename = uriToFilename(closeParams.textDocument.uri);
+    const content = files.get(filename);
+    if (!content) {
+      return;
+    }
+    // Intentionally keep `files` content even after close so we can continue to produce
+    // workspace-wide diagnostics and cross-file results for imports.
+    state.updateSingleFile(filename, content);
```
asukaminato0721: I don't remove the files to keep the diagnose.

lsp/src/pyrefly-web-worker.ts
asukaminato0721: https://github.com/facebook/pyrefly/pull/2460#discussion_r2825948956

asukaminato0721:
fixed semantic highlight.

image

lsp/esbuild.js
```diff
@@ -57,6 +103,48 @@ const esbuildProblemMatcherPlugin = {
...
+    fs.copyFileSync(candidate, wasmOutputPath);
+    copiedWasm = true;
+    break;
+  }
+
+  const jsInputPath = path.join(wasmDir, 'pyrefly_wasm.js');
+  const jsOutputPath = path.join(__dirname, 'wasm', 'pyrefly_wasm.js');
+  if (fs.existsSync(jsInputPath)) {
+    const generatedMarker =
+      '// @generated by `./pyrefly_wasm/build.sh` (wasm-pack). Do not edit by hand.\n';
```
asukaminato0721, line 129: here add header

lsp/src/extension.ts
kinto0: hmm, then I wonder if we could make these two separate extensions? I think the Python extension needs to be installed automatically in the IDE. 

otherwise, let's try exploring other options here.

lsp/src/pyrefly-web-worker.ts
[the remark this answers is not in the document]
kinto0: I think we need to hold off until this

kinto0 (changes requested):

kinto0 (changes requested):
Thanks for working on this - I still think we need a solution for the extensionPack change before I can do an in-depth review.

kinto0:
we've removed the extensionDependency on ms-python.python in 1.0.0. I think this unblocks a lot of this workstream

lsp/src/extension-web.ts
```diff
@@ -0,0 +1,132 @@
...
+  const startPromise = client.start();
+  context.subscriptions.push({
+    dispose: () => {
+      void client?.stop();
+    },
+  });
+  await startPromise;
+
+  try {
+    await indexWorkspaceFiles(client);
```
kinto0: I'm surprised this isn't already handled by default. is it because pyrefly doesn't have a filesystem to find these workspace files on? 

I think I'd like to see a doc explaining why this is necessary and why you picked this approach over something else (virtual filesystem within pyrefly?).

for now, it might make sense to drop workspace indexing from this PR and we can add it as an additional PR in the future

lsp/src/extension-web.ts
```diff
@@ -0,0 +1,132 @@
...
+
+async function indexWorkspaceFiles(client: LanguageClient): Promise<void> {
+  const folders = vscode.workspace.workspaceFolders;
+  if (!folders || folders.length === 0) {
+    return;
+  }
+
+  outputChannel?.appendLine('Indexing workspace files for Pyrefly web...');
+  const pyFiles = await vscode.workspace.findFiles(
+    '**/*.{py,pyi}',
```
kinto0: nit: let's make this a const like the other ones

lsp/src/extension-web.ts
```diff
@@ -0,0 +1,132 @@
...
+    INDEX_EXCLUDES,
+    INDEX_MAX_FILES,
+  );
+  const configFiles = await vscode.workspace.findFiles(
+    '**/pyrefly.toml',
```
kinto0: what about pyproject.toml? 

should this be a const too

lsp/src/extension-web.ts
```diff
@@ -0,0 +1,132 @@
...
+    INDEX_EXCLUDES,
+    INDEX_MAX_FILES,
+  );
+
+  const uris = new Map<string, vscode.Uri>();
+  for (const uri of [...pyFiles, ...configFiles]) {
+    uris.set(uri.toString(), uri);
+  }
+
+  const files: Record<string, string> = {};
```
kinto0, line 38: this is scary

lsp/src/extension-web.ts
```diff
@@ -0,0 +1,132 @@
...
+function createWorker(context: vscode.ExtensionContext): Worker {
+  const workerUri = vscode.Uri.joinPath(
+    context.extensionUri,
+    'dist',
+    'pyrefly-web-worker.js',
+  );
+  return new Worker(workerUri.toString(true));
+}
+
+async function indexWorkspaceFiles(client: LanguageClient): Promise<void> {
```
kinto0: if we keep this (I don't think we should in this PR), we would need a docstring with a lot of reasoning for why it's needed. 

i also think it might deserve a different name, since `index` is overloaded and this is really preparing / sending the setWorkspaceFiles request

lsp/src/extension-web.ts
```diff
@@ -0,0 +1,132 @@
...
+    `Indexed ${Object.keys(files).length} files for Pyrefly web.`,
+  );
+}
+
+export async function activate(
+  context: vscode.ExtensionContext,
+): Promise<void> {
+  if (!outputChannel) {
+    outputChannel = vscode.window.createOutputChannel(
+      'Pyrefly language server',
```
kinto0, line 88: nit: we have two output channels in the vscode extension: normal logs (this) + trace. the trace can get noisy if turned on so I think we should have the same model here

lsp/src/extension.ts
```diff
@@ -261,7 +261,10 @@ export async function activate(context: ExtensionContext) {
   await triggerMsPythonRefreshLanguageServersIfInstalled();
 
   vscode.workspace.onDidChangeConfiguration(async e => {
-    if (e.affectsConfiguration(`python.pyrefly.disableLanguageServices`)) {
+    if (
+      pythonExtension &&
```
kinto0: is this necessary? does it crash without these changes?

lsp/src/pyrefly-web-worker.ts
```diff
@@ -0,0 +1,598 @@
...
+ */
+
+/**
+ * Web-worker "server" used by the Pyrefly VS Code Web extension (vscode.dev / github.dev).
+ *
+ * Architecture:
+ * - `lsp/src/extension-web.ts` spins up this Worker.
+ * - This worker speaks LSP over `postMessage` using the browser JSON-RPC transport from
+ *   `vscode-languageserver-protocol/browser`.
+ * - Typechecking is powered by the Pyrefly WASM playground (`pyrefly_wasm`), which exposes a
```
kinto0: no longer the playground after this PR
```suggestion
 * - Typechecking is powered by the Pyrefly WASM (`pyrefly_wasm`), which exposes a
```

pyrefly/lib/playground.rs
```diff
@@ -182,6 +182,12 @@ impl Range {
     }
 }
 
+#[derive(Serialize)]
+pub struct DefinitionLocation {
+    pub filename: String,
```
kinto0, line 176: does this also change the playground to work correctly with multiple files?

kinto0 (changes requested):
lots of good ideas. it looks like you were able to find an API that lets us translate between the sandbox wasm and the new web client. 

I would push back on a few pieces and ask a bigger picture question:
**what kind of editing do we expect to happen on vscode web?**
I would expect the web extension to be capable of full development similar to what would happen on the VSCode app. I wonder if this design would handle that scale. If not, how long until we need to rewrite it completely?

I can see two approaches that would end in a place I'm happy with for this web extension:

*Starting with vscode-web:* simple but always works
- small PR with basic support (no workspaces)
- only open files
- minimal API changes
- benefits of simplicity, obviously, bug-free, and hard to break (likely no system slowdowns by opening a few files, likely no encoding size/latency issues by only sending one message at a time between clients/servers)

*Starting with wasm:* paves the path for a fast system
- a set of PRs improving our wasm APIs in preparation for a vscode-web extension
- decide early if we want to fork our sandbox wasm APIs from the web extension APIs: pros/cons
- benefits of performance, extensibility, longer-term support

How do you think this extension will be used? Happy to discuss more over office hours or discord.

kinto0:
cool! It seems less risky this way. curious about your thoughts on how to test this. how can we make a reproducible way to debug issues like lsp_interaction? how can we ensure these APIs don't break? should we consider e2e tests in addition to integration tests? why / why not?

asukaminato0721:
> cool! It seems less risky this way. curious about your thoughts on how to test this. how can we make a reproducible way to debug issues like lsp_interaction? how can we ensure these APIs don't break? should we consider e2e tests in addition to integration tests? why / why not?

I'd test this in layers, with most coverage below VS Code and only a small e2e smoke test.

1. Add Rust Playground contract tests now for the new APIs, especially cross-file definition with filenames.
1. Add a lightweight worker/protocol test harness as a follow-up if we keep expanding web support.
1. Add one `@vscode/test-web` smoke test in CI once packaging is settled, mainly to prevent works in Node, fails in VS Code Web regressions.

for this pr, Implemented the API contract tests.

I did not add VS Code web e2e in this pass; that needs separate test-web infrastructure and is better as a follow-up rather than coupling this PR to browser automation immediately.

kinto0:
awesome! thanks for the work on this and the prompt replies. it'll take me a bit to test this since I fly out to pycon very soon and won't get a chance to before then. 

still curious about the changes to the sandbox. I imagine this also makes the sandbox work multi-file? are there tests that changed for that?

asukaminato0721:
> awesome! thanks for the work on this and the prompt replies. it'll take me a bit to test this since I fly out to pycon very soon and won't get a chance to before then.
> 
> still curious about the changes to the sandbox. I imagine this also makes the sandbox work multi-file? are there tests that changed for that?

https://github.com/facebook/pyrefly/pull/2460/changes#diff-23aaad95a91940165a87e1b9646ed6879755641d332e2d510583b18afe3fe924R787 I think this count?

`test_cross_file_goto_definition_locations` for cross-file definition returning filename

NathanTempest:
Hello @asukaminato0721 thank you for your persistence on working on this issue. We are grateful for your efforts in improving Pyrefly! I have reviewed this PR and here are my findings:

**Setup + Blockers**
Pulled, built locally (pyrefly_wasm/build.sh + npm run compile in lsp/), ran Run Web Extension against a two-file Python workspace. Reviewing b29862077.

Big picture: the slim-down in 2fe4927 was the right call and the new playground.rs::goto_definition_locations + cross-file contract test is a solid foundation. Two surgical fixes needed:

**1. gotoDefinitionLocations throws on any token without a definition**
pyrefly_wasm/lib.rs:91-99 returns JsValue::NULL when empty; pyrefly-web-worker.ts:520 calls .filter() on it with no guard; pyrefly_wasm.d.ts:73 declares the return as always-array.

Live repro: Cmd-click a variable declaration like bad: str = ... or a literal 42. Dev Tools console:
```
ERR Request textDocument/definition failed:
    Cannot read properties of null (reading 'filter')
    at handleResponse (extension-web.js:2049:40)
```
User-visible toast: Request textDocument/definition failed. Triggered by every "what's this?" / pre-rename click on a declaration.

Fix at the source: return an empty array from WASM:
```
pub fn goto_definition_locations(&mut self, line: i32, column: i32) -> JsValue {
    let results = self.0.goto_definition_locations(Position::new(line, column));
    serde_wasm_bindgen::to_value(&results).unwrap()
}
```
**2. Missing WASM artifact silently disables the extension**
extension-web.ts:47 vscode.workspace.fs.readFile(wasmFileUri) is unguarded. A clone with npm install but without pyrefly_wasm/build.sh produces this state: no toast, no Output channel error, no path forward and Pyrefly just silently does nothing. Wrap in try/catch with a showErrorMessage pointing at build.sh.

Also fires from your own F5 watch flow in esbuild.js:77 only calls copyWasmArtifact() in non-watch builds, so the watch-driven launch config hits this every time the wasm hasn't been rebuilt. Adding copyWasmArtifact() before Promise.all([... .watch()]) fixes both.

NathanTempest:
**Cross-file imports break until every dependency is opened**
The sharpest edge of the "open files only" approach. With lib.py saved next to main.py and only main.py open:
```
Cannot find module `lib`
  Did you mean `zlib`?
  Looked in these locations:
  Build system source database
Pyrefly(missing-import)
```
Opening lib.py in another tab clears it. The worker only knows about files arriving via didOpen. A real project with 20 files looks broken until the user opens every dependency. Two ways to soften without re-introducing speculative full-workspace indexing:
- On-didOpen sibling pull: when a file opens, also readFile siblings in the same folder up to ~100 files / 1 MB and updateSingleFile them.
- Lazy worker-initiated reads: custom request from worker → extension to readFile when state.getErrors() reports a missing import.

**Python version hardcoded to '3.12'**
pyrefly-web-worker.ts:188 new wasmModule.State('3.12'). Projects on 3.8 or with a pinned python_version get silently wrong errors. Pick one: README note (min) / pyrefly.python.version setting (better) / detect from pyrefly.toml/pyproject.toml among open files (best).

**Nits + validation**
- Three Pylance namespace collisions at activation
[meta.pyrefly]: Cannot register 'python.analysis.completeFunctionParens'. ...
[meta.pyrefly]: Cannot register 'python.analysis.diagnosticMode'. ...
[meta.pyrefly]: Cannot register 'python.analysis.importFormat'. ...

Pre-existing in the extension (not this PR), but worth a follow-up. Move to python.pyrefly.* (the convention the rest of the extension already uses) so it composes with Pylance.

- Deprecated wasm-bindgen init call
pyrefly_wasm.js:623 using deprecated parameters for the initialization function
ensureWasmState @ pyrefly-web-worker.ts:181
wasmModule.default(wasmUrl) → wasmModule.default({ module_or_path: wasmUrl }). Warning today, future wasm-bindgen will reject.

### What I tried: 
local build on macOS (clean) → dev host launch → hover, Cmd-click defs, inlay hints, diagnostics all functional alongside Pylance → captured the live Cannot read properties of null trace for blocker #1 → characterized silent failure for #2 by removing the wasm → confirmed cross-file resolution by opening/closing lib.py.

kinto0:
> Two ways to soften without re-introducing speculative full-workspace indexing

This sounds like the biggest dealbreaker. I wonder if we can copy how other web extensions work?

kinto0:
thanks!! do we need to publish a separate binary for pyrefly to work on the web? I'd like this to be "experimental" in some way since it seems like it doesn't support all the features of the native extension

will let @NathanTempest review

PR #2462, opened 2026-02-18
[Pyrefly][github actions][RFC] mypy primer error message classifier

migeed-z:
Adds a classifier that automatically labels mypy_primer diffs as improvement, regression, or neutral. 

  For non-trivial cases, the classifier fetches source code from GitHub and uses a three-category framework (spec says so / pyrefly wrong / too strict) to guide classification.

This diff Includes:
-  a GitHub Actions workflow (manual trigger for now)
-  58 unit tests, and 30 real PR fixtures for validation.

Next steps:
I would want to manually trigger this on a couple of PRs and examine the results before having it run automatically.


grievejia (changes requested):

grievejia (approved):

Issue #2398, opened 2026-02-12
Incorrect error for `numpy.iinfo(numpy.int32)`


jakevdp:
### Describe the Bug

Pyrefly incorrectly errors for the following, with NumPy v2.4.2:
```python
import numpy as np
np.iinfo(np.int32)
```
```
$ pyrefly check test.py
ERROR No matching overload found for function `numpy._core.getlimits.iinfo.__init__` called with arguments: (type[signedinteger[_32Bit]]) [no-matching-overload]
 --> test.py:3:9
  |
3 | np.iinfo(np.int32)
  |         ^^^^^^^^^^
  |
  Possible overloads:
  (int_type: _HasDType[dtype[_IntegerT_co]] | _HasNumPyDType[dtype[_IntegerT_co]] | dtype[_IntegerT_co] | _IntegerT_co | type[_IntegerT_co]) -> None [closest match]
  (int_type: Literal['<n', '=n', '>n', 'int', 'int_', 'intp', 'n', '|n'] | int | type[int]) -> None
  (int_type: _Int8Codes) -> None
  (int_type: _UInt8Codes) -> None
  (int_type: _Int16Codes) -> None
  (int_type: _UInt16Codes) -> None
  (int_type: _Int32Codes) -> None
  (int_type: _UInt32Codes) -> None
  (int_type: _Int64Codes) -> None
  (int_type: _UInt64Codes) -> None
  (int_type: str) -> None
 INFO 1 error
```
This code passes with `mypy`, `pytype`, `ty`, and `pyright`; I'm not sure exactly what the issue is with `pyrefly`, and I have so far been unable to reduce this to a smaller/self-contained reproduction.

### Sandbox Link

Sandbox link not possible because of NumPy requirement.

kinto0:
removing the other overloads leads to a clearer picture of the issue:
```
Argument `type[signedinteger[_32Bit]]` is not assignable to parameter `int_type` with type `Literal['<i4', '=i4', '>i4', 'i4', 'int32', '|i4']` in function `numpy._core.getlimits.iinfo.__init__`Pyrefly[bad-argument-type](https://pyrefly.org/en/docs/error-kinds/#bad-argument-type)
```

Pyright also raises an error here:
```
Argument of type "int32" cannot be assigned to parameter "int_type" of type "_Int32Codes" in function "__init__"
  Type "int32" is not assignable to type "_Int32Codes"
    Type "int32" is not assignable to type "Literal['int32']"
[...]

PR #2471, opened 2026-02-19
Fix union of T | type[T] failing to match the type[T] branch

kinto0:
Summary:
When matching a value against a union containing both bare TypeVars and wrapped TypeVars (e.g. `T | type[T]`), the solver tried them in arbitrary order. If the bare `T` was tried first, it would eagerly pin the TypeVar to a type that violates bounds (e.g. `type[Sub]` instead of `Sub`), and since var-solving always returns Ok even on bound violations, the more specific `type[T]` branch was never tried.

The fix partitions var-containing union members into bare vars (`T`) and wrapped vars (`type[T]`, `list[T]`, etc.), trying wrapped vars first. Wrapped forms are more specific — they extract the inner type from the argument (e.g. `type[T]` extracts the instance type from a class object), producing better TypeVar solutions that are more likely to satisfy bounds.

fixes https://github.com/facebook/pyrefly/issues/2398

Differential Revision: D93779356


samwgoldman (approved):

Issue #2448, opened 2026-02-17
JSONSchema for Pyrefly


connernilsen:
We should add Pyrefly to jsonschema's [schemastore](https://github.com/SchemaStore/schemastore) to provide automatic structural validation for Pyrefly configs through their plugin. It's used by a lot of other type checkers as well, so seems like a good idea.

* Create a schema for Pyrefly (#2100) 
* Autogenerate Pyrefly schema with the [schemars](https://docs.rs/schemars/latest/schemars/) crate
* Make PR to upload Pyrefly schemas to schemastore

connernilsen:
Hey @Prathamesh-tech-eng, if you're interested in continuing to work on this, I can add you as the assignee for this issue. Otherwise, let me know and I'll leave it open

kinto0:
how does this differ from #536?

connernilsen:
Oops forgot I made that one a while ago. We'll consider this one an expansion of the previous issue, since we're expanding the scope a bit and there's a lot of outdated discussion there

Prathamesh-tech-eng:
Hey @connernilsen! I'd be interested in continuing work on this issue.

I see that Task 1 (creating the schemas) was completed in PR #2100. For the remaining tasks:

Task 2 - Autogenerate with schemars:**
The current schemas are manually maintained. I can work on:
- Adding `schemars` as a dependency to `pyrefly_config`
- Deriving `JsonSchema` on the config structs (`ConfigBase`, etc.)
- Creating a schema generation tool/binary that outputs to `schemas/`
- Integrating this into the build process or as a standalone command

Task 3 - SchemaStore submission:**
Once we have schemas hosted at `https://pyrefly.org/schemas/`, I can:
- Create a PR to SchemaStore following the guide in `schemas/SCHEMASTORE.md`
- Register both `pyrefly.toml` and `pyproject.toml` patterns in their catalog

Should I focus on Task 2 (schemars autogeneration) first?

connernilsen:
@Prathamesh-tech-eng alright, I assigned you to the task. And yeah, could you work on task 2 first? That way, keeping schemastore in sync will be easier when we make changes.

Let me know if you have any questions or want to discuss anything around this.

Prathamesh-tech-eng:
Hello @connernilsen , hope you are doing fine.  Just wanted an update on the review

connernilsen:
[...]

PR #2473, opened 2026-02-20
Auto-generate JSON Schema for Pyrefly config using schemars (#2448 Task 2)

Prathamesh-tech-eng:
Summary

Implements Task 2 from #2448: auto-generate the Pyrefly JSON Schema from Rust config types using the [`schemars`](https://docs.rs/schemars/0.8) crate, replacing the need to manually maintain schemas/pyrefly.json.

Changes

- Add schemars = "0.8" as an optional dependency behind a jsonschema feature flag to 4 crates: pyrefly_util, pyrefly_python, pyrefly_build, pyrefly_config
- Add JsonSchema implementations for all config-related types:
  - Derive-based : Tool, Severity, ErrorKind, UntypedDefBehavior, RecursionOverflowHandler, BxlArgs, BuildSystemArgs, BuildSystem
  - Manual impls (for types with custom serde or external types without JsonSchema): Glob, Globs, PythonVersion, PythonPlatform, CustomQueryArgs, ErrorDisplayConfig, ModuleWildcard, ConfigOrigin<T>, ExtraConfigs, ConfigBase, PythonEnvironment, Interpreters, SubConfig, ConfigFile
- Add generate_schema binary in pyrefly_config that outputs Draft-07 JSON Schema to stdout
- All schema code is gated with #[cfg(feature = "jsonschema")] ,  zero impact on normal builds

Usage

cargo run -p pyrefly_config --features jsonschema --bin generate_schema > schemas/pyrefly.json


Prathamesh-tech-eng:
@connernilsen This implements Task 2 from #2448. I had few doubts:

1. Should the generated schema replace the existing schemas/pyrefly.json directly, or should they coexist during a transition period?
2. Would you like me to add a CI step that runs the generator and checks the output matches the checked-in schema (to catch drift)?
3. Cargo.toml files: I noticed the # @generated by autocargo comment , will Meta's internal tooling need to be updated separately for these dependency additions, or is my change sufficient?

connernilsen:
@Prathamesh-tech-eng sorry for the delay.

> 1. Should the generated schema replace the existing schemas/pyrefly.json directly, or should they coexist during a transition period?

I think it would be alright to replace them directly. Is there any specific reason you're thinking of to have them coexist?

> 2. Would you like me to add a CI step that runs the generator and checks the output matches the checked-in schema (to catch drift)?

Yes! This would be amazing!

> 3. Cargo.toml files: I noticed the # @generated by autocargo comment , will Meta's internal tooling need to be updated separately for these dependency additions, or is my change sufficient?

The `@generated` should be fine to leave in there. Our internal tooling handles that comment nicely right now :)

Thanks for working on this, and sorry again for the delay

crates/pyrefly_build/src/query/buck.rs
```diff
@@ -42,6 +42,7 @@ fn use_systemd() -> bool {
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Default, Hash)]
+#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
```
connernilsen, line 45: Can we enable that as a feature everywhere in the `pyrefly_build`, `pyrefly_config`, and `pyrefly_python` crates?

[Here's an example](https://github.com/facebook/pyrefly/blob/50fb37b4c29771d53a936841ab6c53b1b2bbdf65/crates/pyrefly_util/src/lib.rs#L25) of how we did it for another feature in one of our crates

crates/pyrefly_config/src/bin/generate_schema.rs
```diff
@@ -0,0 +1,32 @@
...
+//!   cargo run -p pyrefly_config --features jsonschema --bin generate_schema > schemas/pyrefly.json
+
+use schemars::r#gen::SchemaSettings;
+
+use pyrefly_config::config::ConfigFile;
+
+fn main() {
+    let settings = SchemaSettings::draft07().with(|s| {
+        s.option_nullable = false;
+        s.option_add_null_type = false;
```
connernilsen, line 32: Quick question: what are these? I am struggling to find much documentation on what they do. Can we add comments describing them here?

crates/pyrefly_config/src/bin/generate_schema.rs
```diff
@@ -0,0 +1,32 @@
...
+
+//! Binary to auto-generate JSON Schema for pyrefly.toml / pyproject.toml [tool.pyrefly]
+//! configuration files using the `schemars` crate.
+//!
+//! Usage:
+//!   cargo run -p pyrefly_config --features jsonschema --bin generate_schema
+//!
+//! The output is written to stdout (JSON Schema Draft-07).
+//! Redirect to a file to save:
+//!   cargo run -p pyrefly_config --features jsonschema --bin generate_schema > schemas/pyrefly.json
```
connernilsen, line 16: Awesome!

.gitignore
```diff
@@ -1,3 +1,14 @@
 debug.json
 debug.js
 target/
+
+# Ignore all files in schemas/ except specific JSON schemas and supporting files
+schemas/*
```
connernilsen: Are you running into issues where other schemas are being generated? If not, can we remove this? It seems like something that might just cause headaches with adding other schemas in the future

crates/pyrefly_build/src/query/custom.rs
```diff
@@ -40,6 +40,69 @@ pub struct CustomQueryArgs {
     pub repo_root: Option<PathBuf>,
 }
 
+#[cfg(feature = "jsonschema")]
```
connernilsen: What are all of the extra non-derived schemas adding?

For cases where we need custom behavior for fields, can we instead use `#[schemars(schema_with = "some::function")]` (override specific fields), `#[schemars(transform = some::transform)]` (mutate the schema after auto generating it), or other [macros](https://docs.rs/schemars/latest/schemars/derive.JsonSchema.html) instead of implementing `schemars::JsonSchema` completely for some of the structs below?

connernilsen (changes requested):
Hey @Prathamesh-tech-eng, sorry for the delay in review. I have a few questions on how this works, since I haven't used this before, but I like the look of things so far.

The main question below is about whether we can take advantage of macro settings on fields, variants, and the overall structs instead of manually implementing the `schemars::JsonSchema` trait, since I'm worried we might forget to update it in the future.

Let me know if you have any questions or want to discuss stuff :)

crates/pyrefly_config/src/bin/generate_schema.rs
Prathamesh-tech-eng: Added detailed comments explaining both settings:

option_nullable: Explains this is an OpenAPI 3.0 extension that we disable since JSON Schema Draft-07 doesn't use the "nullable" keyword
option_add_null_type: Explains this would create anyOf unions with null, but we set false so optional fields simply omit the "required" constraint, producing cleaner schemas

.gitignore
Prathamesh-tech-eng: Removed the complex schemas/* wildcard pattern. You're right that it could cause headaches when adding new schemas in the future. The .gitignore now just tracks the standard entries (debug.json, debug.js, target/), and all schema files are tracked normally.

crates/pyrefly_build/src/query/custom.rs
Prathamesh-tech-eng: Refactored CustomQueryArgs to use schemars derive macros as suggested. The 60+ line manual implementation is now replaced with:

#[derive(Debug, Clone, ...)]
#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]
pub struct CustomQueryArgs {
    #[cfg_attr(feature = "jsonschema", schemars(schema_with = "vec1_string_schema"))]
    pub command: Vec1<String>,
    // ...
}

A small helper function vec1_string_schema generates the array schema with minItems: 1. This is much more maintainable and the descriptions come from doc comments automatically.

Prathamesh-tech-eng:
Thanks for the review! I've pushed the changes addressing all feedback.

For the next step (Task 3 - SchemaStore submission), I have a few questions:

1. Should I wait for this PR to be merged before submitting to SchemaStore, or can I work on the SchemaStore PR in parallel?
2. The schema currently has $schema: "http://json-schema.org/draft-07/schema#" but no $id. Should I add $id: "https://pyrefly.org/schemas/pyrefly.json" or similar before SchemaStore submission?
3. Are there any specific SchemaStore requirements or guidelines I should follow? I see there's a schemas/SCHEMASTORE.md file - should I update that with submission instructions?
4. For the SchemaStore catalog entry, what file patterns should pyrefly.json match? I'm thinking:
pyrefly.toml
pyrefly.json
.pyrefly.toml

crates/pyrefly_build/src/query/buck.rs
Prathamesh-tech-eng: Done! The jsonschema feature already propagates through all crates via Cargo.toml:

pyrefly_config  → pyrefly_build/jsonschema, pyrefly_python/jsonschema, pyrefly_util/jsonschema
pyrefly_build   → pyrefly_python/jsonschema, pyrefly_util/jsonschema
pyrefly_python  → pyrefly_util/jsonschema

We use the same #[cfg(feature = "jsonschema")] pattern as in pyrefly_util/src/globs.rs (lines 331-353). Running cargo run -p pyrefly_config --features jsonschema enables the feature across all dependent crates.

connernilsen:
Awesome, thanks for the updates.

>Should I wait for this PR to be merged before submitting to SchemaStore, or can I work on the SchemaStore PR in parallel?

Yeah, I think waiting until we nail down the generation logic is the best approach. It'll help us reduce the likelihood of errors in any schema we push.

>The schema currently has $schema: "http://json-schema.org/draft-07/schema#" but no $id. Should I add $id: "https://pyrefly.org/schemas/pyrefly.json" or similar before SchemaStore submission?

That might be a good idea. Probably something like `https://raw.githubusercontent.com/facebook/pyrefly/refs/heads/main/schemas/pyrefly.json`? What is that `$id` field used for?

>Are there any specific SchemaStore requirements or guidelines I should follow? I see there's a schemas/SCHEMASTORE.md file - should I update that with submission instructions?

I think a lot of the [recommendations here](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#schema-authoring) would be good to follow. Maybe some [validation checks here](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#schemasafe). Other than that, I think once this is committed we should be good to start working on a submission to schemastore.

> For the SchemaStore catalog entry, what file patterns should pyrefly.json match? I'm thinking: ...

We should only match `pyrefly.toml`, since that's the only file name we parse (other than `pyproject.toml`).

crates/pyrefly_config/src/bin/generate_schema.rs
connernilsen: Thanks for adding this!

crates/pyrefly_build/src/query/buck.rs
connernilsen: Oh I think I see what you're saying, but I also think I misread what is going on here. Can we remove the `cfg_attr(feature = "jsonschema"` part? If it's set as an enabled feature in the cargo file, do we also need to gate this here?

crates/pyrefly_build/src/query/custom.rs
```diff
@@ -17,8 +17,30 @@ use vec1::Vec1;
 
 use crate::query::SourceDbQuerier;
 
+/// Generates a JSON schema for Vec1<String>: an array of strings with minItems: 1.
+#[cfg(feature = "jsonschema")]
+fn vec1_string_schema(
```
connernilsen: This is great! Can we instead move it into a new file in `pyrefly_util` for other schema generators to use?

crates/pyrefly_build/src/query/custom.rs
connernilsen: Thanks for fixing that schema generator! It looks much cleaner and will be a lot easier to maintain in the future.

Would you also be able to update the other custom schema generators to attempt to move as much logic as possible into those configuration macros?

crates/pyrefly_config/src/base.rs
```diff
@@ -139,6 +141,24 @@ pub struct ConfigBase {
 #[serde(transparent)]
 pub(crate) struct ExtraConfigs(pub(crate) Table);
 
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for ExtraConfigs {
```
connernilsen, line 174: Is it possible to get schemars to generate something that says this field shouldn't exist? It might just be that we have to mark it as something schemars should skip

schemas/validate_schemas.py
```diff
@@ -1,266 +1,117 @@
 #!/usr/bin/env python3
```
connernilsen, line 1: It looks like the changes to this file are getting rid of tests that validate schemars should fail when given invalid values. Can we add those back?

connernilsen (changes requested):
I love the direction this is going in. Thanks for working on it, and as always, let me know if you have questions

crates/pyrefly_build/src/query/custom.rs
Prathamesh-tech-eng: The `cfg_attr` gating is still needed because the `jsonschema` feature is optional, not a default feature. Looking at the Cargo.toml:

[features]
default = []
jsonschema = ["schemars", ...]

When building pyrefly without `--features jsonschema`, the code should compile without the `schemars` dependency. The `cfg_attr` ensures:
1. The derive macro only runs when the feature is enabled
2. The code compiles cleanly without `schemars` in the dependency tree
3. Users who don't need schema generation don't pay for the extra dependency

Prathamesh-tech-eng: > This is great! Can we instead move it into a new file in `pyrefly_util` for other schema generators to use?

Done! I've moved it to `crates/pyrefly_util/src/schema_helpers.rs` as a reusable helper function. The function is now feature-gated with `#[cfg(feature = "jsonschema")]` and can be used by any crate that needs to generate schemas for Vec1<String>.

crates/pyrefly_config/src/base.rs
Prathamesh-tech-eng: I considered using `#[schemars(skip)]` but kept the current manual implementation because it more accurately represents the actual behavior.

`ExtraConfigs` is a `#[serde(flatten)]` catch-all that captures any unknown fields via `toml::Table`. The current manual impl generates an object schema with `additionalProperties: true`, which correctly tells schema validators that extra properties are allowed.

Using `#[schemars(skip)]` would hide this field from the schema entirely, suggesting only explicitly defined properties are valid. This would cause false validation errors for users with custom/future config fields.

We could use `#[schemars(skip)]` and rely on the parent schema's `additionalProperties: true` setting. Let me know if you'd prefer that approach.

schemas/validate_schemas.py
Prathamesh-tech-eng: Done! I've restored the full test suite with all 23 negative test cases from upstream/main. 

Note: 3 tests are commented out with an explanation - they validate mutual exclusivity between build-system types (buck shouldn't have `command`, custom shouldn't have `extras`/`isolation-dir`). The auto-generated schema doesn't include `additionalProperties: false` in the oneOf variants, so these specific constraints can't be enforced. The comments document this known limitation.

Prathamesh-tech-eng:
Thanks for the review! I've addressed all the feedback:

I kept the manual implementation to accurately represent `additionalProperties: true`. Would you prefer me to use `#[schemars(skip)]` instead? Both approaches work, but have different schema output.

Also fixed a bug I discovered during the review: the `strict-callable-subtyping` field was missing from the ConfigBase manual JsonSchema implementation. This field is now included in the generated schema.

crates/pyrefly_config/src/base.rs
connernilsen: Yeah, I think if that correctly shows the current behavior, that would be great. The less manual implementation we have here the better.

schemas/validate_schemas.py
connernilsen: Oh good point. In that case, we can delete the extra tests instead of just commenting them out.

crates/pyrefly_config/src/environment/environment.rs
```diff
@@ -78,6 +78,62 @@ pub struct PythonEnvironment {
     pub interpreter_stdlib_path: Vec<PathBuf>,
 }
 
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for PythonEnvironment {
```
connernilsen, line 82: Is there a reason we can't generate most of this? It seems like as long as generators for the subschemas exist, we can just derive `#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]` for `PythonEnvironment.

crates/pyrefly_config/src/environment/interpreters.rs
```diff
@@ -45,6 +45,90 @@ pub struct Interpreters {
     pub skip_interpreter_query: bool,
 }
 
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for Interpreters {
```
connernilsen, line 49: Same here, can we derive `#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]` on `Interpreters`?

crates/pyrefly_config/src/base.rs
```diff
@@ -177,6 +197,142 @@ impl PartialEq for ExtraConfigs {
...
+impl schemars::JsonSchema for ConfigBase {
+    fn schema_name() -> String {
+        "ConfigBase".to_owned()
+    }
+
+    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        use schemars::schema::*;
+
+        let mut properties = schemars::Map::new();
+        let required: std::collections::BTreeSet<String> = std::collections::BTreeSet::new();
```
connernilsen, line 210: I don't think this is being used. Could you double check on what this was supposed to do?

crates/pyrefly_config/src/base.rs
```diff
@@ -177,6 +197,142 @@ impl PartialEq for ExtraConfigs {
     }
 }
 
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for ConfigBase {
```
connernilsen, line 201: Is it possible to derive this with `#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]`?

crates/pyrefly_config/src/config.rs
```diff
@@ -80,6 +80,54 @@ pub struct SubConfig {
     pub settings: ConfigBase,
 }
 
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for SubConfig {
```
connernilsen, line 84: Could we try flattening here with the `#[serde(flatten)]` value macro? If so, I'm hoping we can get rid of the manual generator here.

crates/pyrefly_config/src/config.rs
```diff
@@ -566,6 +614,266 @@ pub struct ConfigFile {
...
+impl schemars::JsonSchema for ConfigFile {
+    fn schema_name() -> String {
+        "ConfigFile".to_owned()
+    }
+
+    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        use schemars::schema::*;
+
+        let mut properties = schemars::Map::new();
+        let required = std::collections::BTreeSet::new();
```
connernilsen, line 627: Similar to the SubConfig implementation, if we can just derive everything and use `#[serde(flatten)]` for sub configs, that would be super helpful!

Is there anything else here we would need a manual implementation for?

crates/pyrefly_config/src/error.rs
```diff
@@ -24,6 +24,50 @@ use crate::error_kind::Severity;
...
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for ErrorDisplayConfig {
+    fn schema_name() -> String {
+        "ErrorDisplayConfig".to_owned()
+    }
+
+    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        use schemars::schema::*;
+
+        // Each error code maps to either a bool or a severity string
```
connernilsen, line 36: Does the default derivation for severity enum not automatically allow for `one_of` a severity string or bool?

crates/pyrefly_config/src/module_wildcard.rs
```diff
@@ -114,6 +114,18 @@ impl Serialize for ModuleWildcard {
     }
 }
 
+#[cfg(feature = "jsonschema")]
+impl schemars::JsonSchema for ModuleWildcard {
```
connernilsen, line 118: Can we derive this too?

crates/pyrefly_config/src/util.rs
```diff
@@ -53,6 +53,18 @@ pub(crate) enum ConfigOrigin<T> {
     ConfigFile(T),
 }
 
+/// ConfigOrigin is transparent in schema — it serializes as the inner type T.
+#[cfg(feature = "jsonschema")]
+impl<T: schemars::JsonSchema> schemars::JsonSchema for ConfigOrigin<T> {
```
connernilsen, line 58: This one is somewhat interesting, makes sense why we need a manual implementation here :)

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -147,6 +147,33 @@ impl Serialize for PythonVersion {
...
+            metadata: Some(Box::new(schemars::schema::Metadata {
+                description: Some(
+                    "The Python version to use for type checking. Format: <major>[.<minor>[.<micro>]]"
+                        .to_owned(),
+                ),
+                default: Some(serde_json::json!("3.13.0")),
+                ..Default::default()
+            })),
+            string: Some(Box::new(schemars::schema::StringValidation {
+                pattern: Some(r"^\d+(\.\d+)?(\.\d+)?$".to_owned()),
```
connernilsen, line 168: Can we derive and use `[schemars(regex...)]` for this? https://docs.rs/schemars/latest/schemars/derive.JsonSchema.html#regex

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -229,6 +256,34 @@ impl PythonPlatform {
...
+    fn json_schema(_generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        schemars::schema::SchemaObject {
+            instance_type: Some(schemars::schema::InstanceType::String.into()),
+            metadata: Some(Box::new(schemars::schema::Metadata {
+                description: Some(
+                    "The value used with conditions based on type checking against sys.platform values."
+                        .to_owned(),
+                ),
+                default: Some(serde_json::json!("linux")),
+                examples: vec![
```
connernilsen, line 274: Is there a way to derive all of this, specify that we prefer `linux`, `darwin`, or `win32`, but take anything here?

connernilsen (changes requested):
Hey, thanks for continuing to work on this, and sorry for the delay. We're busy getting ready for our V1 release, so things are hectic right now.

Things are progressing well! For your question about `ExtraConfigs`,
>I kept the manual implementation to accurately represent `additionalProperties: true`. Would you prefer me to use `#[schemars(skip)]` instead? Both approaches work, but have different schema output.

I think if `additionalProperties: true` and `#[schemars(skip)]` get Pyrefly to allow additional properties for now but hide the `extra-configs` value, that would be perfect. If not, feel free to leave the custom implementation for `ExtraConfigs`.

Also, generally, we want to keep the manual schema generation to the absolute minimum we can. I expect config stuff to change over time, so I'd like to avoid a case where we're doing something wrong because someone forgot to update the schema generator when changing the struct. For anything (including possibly `ExtraConfigs`) that you leave custom schema generation for, can you put a comment detailing why this is so that future editors/reviewers understand why things are the way they are now? And for anything that doesn't have a great reason, can you derive `#[cfg_attr(feature = "jsonschema", derive(schemars::JsonSchema))]`? If it's just a few fields that are making it difficult, prefer to use ` schemars(schema_with = <function>)` like we did for `Vec1` types instead of defining schema generation for the whole type.

connernilsen:
I'm going to close this, since there hasn't been activity in a while. If you're interested in continuing to work on it, please reopen the PR/tag me/let me know on Discord (whatever works best for you)!

A lot of what I'm looking for in future work around this task is around automating the generation of the schema as much as possible, and having as few manual implementations as possible (only where absolutely required).

PR #2481, opened 2026-02-21
[FEAT] pyrefly lsp config path customization

yeetypete:
# Summary

Adds support for customizing the config path used by the pyrefly LSP via the
`python.pyrefly.configPath` VSCode setting. This is handled per-workspace
through the standard `workspace/configuration` mechanism, so each workspace
folder can point to a different config file, and changes take effect without
restarting the server.

This is particularly useful in large projects where your `search-path`
configuration includes paths outside the directory tree where the pyrefly LSP
searches for its configuration. In that case, if you open a file in your editor,
the LSP will fall back to showing the file without any of your custom pyrefly
settings.

## Changes

- Added `workspace_config` field to `Workspace` to store the per-workspace config path
- Added `configPath` to `PyreflyClientConfig` so it's parsed from `python.pyrefly.configPath`
- Updated `WorkspaceConfigConfigurer::configure` to load the workspace-specified
  config via `ConfigFile::from_file` instead of auto-discovery when set

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/workspace.rs
+++ b/pyrefly/lib/lsp/non_wasm/workspace.rs
@@ -92,8 +92,12 @@ impl ConfigConfigurer for WorkspaceConfigConfigurer {
             self.0.get_with(dir.to_owned(), |(workspace_root, w)| {
                 if let Some(workspace_config_path) = &w.workspace_config {
                     let (new_config, new_errors) = ConfigFile::from_file(workspace_config_path);
-                    config = new_config;
-                    errors = new_errors;
+                    if new_errors.is_empty() {
+                        config = new_config;
+                        errors = vec![];
+                    } else {
+                        errors = new_errors;
+                    }
                 }
                 if let Some(search_path) = w.search_path.clone() {
                     config.search_path_from_args = search_path;


kinto0:
thanks for the change! and sorry about the delay in review. 

I'm still not sure I understand the exact situation you're running into. a project dependency doesn't have a pyrefly configuration so there's no way to configure that project?

in that case, it sounds like our default configuration does not work correctly there? what are the bugs you are running into? is it editor-related things like showing / hiding type errors (in which case we have global [initialization options](https://pyrefly.org/en/docs/IDE/#customization) which are very similar to what it looks like this PR does). Or is it project setup like adding search paths? if it's project setup, I don't think a global configuration is desired since it affects every project. 

can I ask where these files are? are they site-packages or simply other projects you own? in either of these cases, have you tried adding a pyrefly.toml?

or if it's unrelated to setup, are there bugs we can fix with import resolution to make these external projects work correctly?

yeetypete:
@kinto0 Sure, Happy to add some more context. I took a look at `initializationOptions` and am not sure if any of them will fix the issue unless `extraPaths` does the trick? Unless I am mistaken that specific parameter doesn't seem to be exposed in the VSCode settings.

The problem is pretty simple: I want the pyrefly lsp to use a specific `pyproject.toml` file for its configuration instead of attempting to discover one automatically. It falls back to default settings if no config file is discovered in an active file's parent directory which breaks the pyrefly IDE integration in my project. 

I have the following pyrefly configuration in my `pyproject.toml` file in my project root (parent directory of `src`):
```toml
[tool.pyrefly]
baseline = "baseline.json"
project-includes = ["src", "../../libs/ros2/src"]
search-path = [
    # A few dozen packages from the `libs/ros2` directory
    # packages in the `src` directory 
]

site-package-path = [
    "/opt/ros/jazzy/lib/python3.12/site-packages",
    "/opt/venv/lib/python3.12/site-packages",
    "install/lib/python3.12/site-packages",
]

[tool.pyrefly.errors]
# Pyrefly does not yet have a strict mode, so we manually set all diagnostics
# with "ignore" severity levels to "warn".
implicit-abstract-class = "warn"
implicit-any = "warn"
implicitly-defined-attribute = "warn"
missing-override-decorator = "warn"
missing-source = "warn"
not-required-key-access = "warn"
open-unpacking = "warn"
unannotated-attribute = "warn"
unannotated-parameter = "warn"
unannotated-return = "warn"
untyped-import = "warn"
unused-ignore = "warn"
```

As soon as I open a file in the `../libs/ros2` directory I don't get the proper type checking information visualized in the IDE. A hacky solution would be to duplicate the pyrefly configuration in `../libs/ros2` or move it to a parent directory shared by both projects. However this gets pretty messy since i have a lot of packages in `search-path`.

I'm sure this is not a conventional setup because of how ROS treats python packages but I could imagine a similar situation if you use [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) or just have a libs folder outside the current project's directory tree.

If there's a way to configure the lsp to do what I want with the existing set of exposed parameters that would be great, but otherwise I think something like this may be needed.

kinto0:
> A hacky solution would be to duplicate the pyrefly configuration in ../libs/ros2 or move it to a parent directory shared by both projects

I don't think this is a hacky solution - it seems like this project requires dependencies that are not explicitly set up. if another project depends on this project, they should also have these dependencies. adding a pyrefly.toml/pyproject.toml to that project would accomplish the same goal. 

did you have a similar setup with another LSP? and was the solution there to set a global config?

kinto0:
> I don't get the proper type checking information visualized in the IDE

oh, is this the issue? we have a vscode setting + initializationOption for forcing type errors everywhere

image

if that isn't strict enough, I think the correct thing to do would be to add a vscode config + initializationOption for these type error overrides

yeetypete:
> did you have a similar setup with another LSP? and was the solution there to set a global config?

No. We made the jump to pyrefly from basedpyright at the same time we reorganized packages and only then realized this configuration did not work. For c++ code the lsp can handle this project configuration from a single compile commands json with no issues. 

We primarily switched becasue `baseline.json` and `pyrefly check` worked with paths outside the main project tree. The paths outside the current project become absolute, (see https://github.com/facebook/pyrefly/issues/2547) but in our case is ok cause we work inside a devcontainer. We had the assumption that since `pyrefly check` works perfectly fine with the above project configuration we would expect the lsp to also work.

The vsode setting isn't the issue as we already use the setting you mentioned :)

```json
{
  "pyrefly.lspPath": "/opt/venv/bin/pyrefly",
  "terminal.integrated.cwd": "${workspaceFolder:AMG Workspace}",
  "python.languageServer": "None",
  "python.pyrefly.displayTypeErrors": "force-on"
}
```

I guess in theory we could duplicate the config. However this is a bit strange because the packages are only installed in the project's current directory (`install/lib/python3.12/site-packages`). The `../../libs/ros2` packages are not meant to be used standalone and instead are used in several projects across our repo.

yeetypete:
If we duplicated the config then `site-package-path` would be a bit wonky. Since the packages in `libs/ros2` are never installed to a specific path (the path depends on the project they are used in) a theoretical `pyproject.toml` file with pyrefly configuration in `libs/ros2` would have to point to non-existant site install dirs. I.e.

```toml
# libs/ros2/pyproject.toml
[tool.pyrefly]
site-package-path = [
    "../../project-a/install/lib/python3.12/site-packages",
    "../../project-b/install/lib/python3.12/site-packages", # only one of these site package paths exists at any time
]
```
We also don't use editable installs and instead rely on `search-path` for IDE integration by clicking on a type. Additionally some `.py` files are generated by cmake (ros2 msg stubs, flatbuffer stubs, proto stubs). I guess mixed-language projects is probably not the most common place where pyrefly is used :sweat_smile:

yeetypete:
@kinto0 We did not use it but it does actually seem like basedpyright supports this. See [`basedpyright.analysis.configFilePath`](https://docs.basedpyright.com/dev/configuration/language-server-settings/). In that case I do not think it is too much to get to feature parity between the two here :)

The reasoning basedpyright uses is pretty similar to our use-case here (we also have a monorepo).

kinto0:
thanks for all the info - you've almost convinced me. some questions and thoughts:

when your developers switch from project-1 to project-2, is it expected that they switch their vscode setting to the other project's pyrefly.toml in this vscode setting? 

the most important issue you've pointed out is the fact that your `ros2` lib is depended upon by two separate projects with their own site-packages.I think editable installs will solve this for you since then each project can have it's own version of `ros2`. this would mean developers don't have to switch the setting each time they switch projects

what do you think? if it's a hassle to migrate, we should get something working for you now. my understanding is that a global extraPaths vscode setting which uses the existing global setting for pyrefly.toml is semantically equivalent to a global extraPaths as a vscode setting. you can attempt this right now by adding 
```
    "python.pyrefly.extraPaths": ["/Users/kylei/projects/test/test2"],
```
to your settings.json.

yeetypete:
> when your developers switch from project-1 to project-2, is it expected that they switch their vscode setting to the other project's pyrefly.toml in this vscode setting?

We have separate devcontainers for `project-1` and `project-2`. Instead of having top level vscode settings in our monorepo root, the vscode settings are set in each project individually. We currently use a fork of pyrefly with the changes in this PR. In each project `pyrefly.configPath` refers to the project's pyproject.toml file (Also could be `PYREFLY_CONFIG` env var or in `pyrefly.lspArgs`) In our case since we use devcontainers we set the `PYREFLY_CONFIG` env var in our docker-compose file. No one needs to edit vscode settings when switching projects. We use [code workspaces](https://code.visualstudio.com/docs/editing/workspaces/workspaces) to have VSCode open up all required directories where a developer needs to make code changes when working on a specific project (mainly `libs/ros2` dir and the specific project dir.)

> the most important issue you've pointed out is the fact that your `ros2` lib is depended upon by two separate projects with their own site-packages.

The site-packages directory (`install/lib/python3.12/site-packages`) inside each project is generated during the build. Each project does not have its own `site-packages` until we actually build it. It is most similar to a venv (except that `install` is not editable and contains both c++ and python code). This is also the reason why we must build our project before we get proper type checking. Note that you would also have to do this in the case of a pure python project which contains a `setup.py` or hatch hook that generates stubs for flatbuffers, protobuf, etc.

> I think editable installs will solve this for you since then each project can have it's own version of ros2.

I'm not following how editable installs would fix this for us. How would editable installs fix the issue of the lsp not properly displaying type info in the IDE when I open files in `../../libs/ros2` where there is no parent pyproject.toml or pyrefly.toml? I enabled an editable install (not default in ROS2) and ran into the same issue.

Maybe showing you our monorepo layout in tree view helps:

```
core/
├── libs/
│   └── ros2/
│       └── src/
│           ├── package-a/
│           ├── package-b/
│           └── ...         # a few dozen shared packages here. They are not meant to be used standalone. Only in project-1, project-2, etc.
│
└── projects/
    ├── project-1/
    │   ├── colcon_defaults.yaml  # file specific to ROS2 build system
    │   ├── pyproject.toml        # Contains pyrefly pyproject.toml configuration similar to https://github.com/facebook/pyrefly/pull/2481#issuecomment-4000146075
    │   └── src/
    │       └── ...               # project-specific packages here.
    │
    └── project-2/
        ├── colcon_defaults.yaml  # file specific to ROS2 build system
        ├── pyproject.toml        # Contains pyrefly pyproject.toml similar to https://github.com/facebook/pyrefly/pull/2481#issuecomment-4000146075
        └── src/
            └── ...               # project-specific packages here.
```
In this case all `src` directories contain a mixture of c++ and python packages. `package-a`, `package-b` are referenced in `project-1` and `project-2`. When we build `project-1` or `project-2` any depenencies in `libs/ros2` get pulled into the build. When we open either project we would expect that when we click on a type the IDE goes to `libs/ros2` dir. This already works great without the changes in this PR. However once inside `libs/ros2` everything has red squiggles because pyrefly doesn't know about the original `pyproject.toml` file config inside `project-1` or `project-2`. `pyrefly check` doesn't have any issues because it we run it from inside `project-1` or `project-2` where it sees the right config.

I realize that ROS2 handles python in an unconventional (bad) way compared to the typical python project setup. Unfortunately we are at the mercy of ROS2 there. However I think the existance of the basedpyright `basedpyright.analysis.configFilePath` setting is a good indicator that there are other use-cases where it is beneficial to give the developer control over how the lsp discovers its config.

yeetypete:
Here's an example of what it looks like viewing a `libs/ros2` python file without our custom changes.

image

As expected lots of red squiggles.

kinto0:
> How would editable installs fix the issue of the lsp not properly displaying type info in the IDE when I open files in ../../libs/ros2 where there is no parent pyproject.toml or pyrefly.toml?

I was thinking an editable install would behave like a symlink where you can open the symlinked file. But you're right if you just open the ros2 project it would not behave like this. 

I wonder if you could make a symlink from project-1 to ros2 and project-2 to ros2. per [this issue](https://github.com/facebook/pyrefly/issues/1104) it does seem like we support symlinks natively. so your structure would look like this

```
core/
├── libs/
│   └── ros2/
│       └── src/
│           ├── package-a/
│           ├── package-b/
│           └── ...         # a few dozen shared packages here. They are not meant to be used standalone. Only in project-1, project-2, etc.
│
└── projects/
    ├── project-1/
    │   ├── colcon_defaults.yaml  # file specific to ROS2 build system
    │   ├── pyproject.toml        # point at symlink of ros2 instead of realpath
    │   ├──ros2/ # symlink
    │   └── src/
    │       └── ...               # project-specific packages here.
    │
    └── project-2/
        ├── colcon_defaults.yaml  # file specific to ROS2 build system
        ├── pyproject.toml        # point at symlink of ros2 instead of realpath
        ├──ros2/ # symlink
        └── src/
            └── ...               # project-specific packages here.
```
if you open `project-1/ros2/src/package-a`, I think you should find that it upwards-searches and finds the project-1 pyrefly.toml. then we have two "versions" (each symlink) that each can check with different configs.

my worry with the global / workspace-folder-level config file setting is that it will mangle `ModuleName`s. the way we track a fully-qualified name in Pyrefly is relative to a pyrefly/pyproject.toml (i.e. foo.bar in project-2 is at project-2/foo/bar) and I'm worried we will eventually break this setting accidentally.

yeetypete:
> if you open `project-1/ros2/src/package-a`, I think you should find that it upwards-searches and finds the project-1 pyrefly.toml. then we have two "versions" (each symlink) that each can check with different configs.

We don't like using symlinks in the monorepo as they are messy and hard to follow in the editor :). All of our other tooling supports the project layout we have, we just need to get pyrefly there!

> my worry with the global / workspace-folder-level config file setting is that it will mangle `ModuleName`s. the way we track a fully-qualified name in Pyrefly is relative to a pyrefly/pyproject.toml (i.e. foo.bar in project-2 is at project-2/foo/bar) and I'm worried we will eventually break this setting accidentally.

Just to clarify `project-1` and `project-2` themselves are not pip-installable. The `pyproject.toml` is there just to configure settings for pytest, pyrefly, declare dependencies, etc. Each package inside the project contains a `setup.py` (yes, ROS2 still doesn't support pyproject.toml in 2026). This is sort of similar to [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/). During the build they all get pip-installed individually by [colcon](https://colcon.readthedocs.io/en/released/), the ROS2 meta-build tool.

FYI ty also supports the setting I mentioned in basedpyright. See [configurationFile](https://docs.astral.sh/ty/reference/editor-settings/#configurationfile). It would be a shame if we had to switch or continue to maintain our own fork because of one setting! We really like the pyrefly type-checking philosophy! I understand the implications of having this setting may break some assumptions pyrefly has been making but if 2 / 3 major type checkers support it I think that is good indicator pyrefly should have it as well :).

kinto0:
> > if you open `project-1/ros2/src/package-a`, I think you should find that it upwards-searches and finds the project-1 pyrefly.toml. then we have two "versions" (each symlink) that each can check with different configs.
> 
> We don't like using symlinks in the monorepo as they are messy and hard to follow in the editor :). All of our other tooling supports the project layout we have, we just need to get pyrefly there!
> 
> > my worry with the global / workspace-folder-level config file setting is that it will mangle `ModuleName`s. the way we track a fully-qualified name in Pyrefly is relative to a pyrefly/pyproject.toml (i.e. foo.bar in project-2 is at project-2/foo/bar) and I'm worried we will eventually break this setting accidentally.
> 
> Just to clarify `project-1` and `project-2` themselves are not pip-installable. The `pyproject.toml` is there just to configure settings for pytest, pyrefly, declare dependencies, etc. Each package inside the project contains a `setup.py` (yes, ROS2 still doesn't support pyproject.toml in 2026). This is sort of similar to [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/). During the build they all get pip-installed individually by [colcon](https://colcon.readthedocs.io/en/released/), the ROS2 meta-build tool.
> 
> FYI ty also supports the setting I mentioned in basedpyright. See [configurationFile](https://docs.astral.sh/ty/reference/editor-settings/#configurationfile). It would be a shame if we had to switch or continue to maintain our own fork because of one setting! We really like the pyrefly type-checking philosophy! I understand the implications of having this setting may break some assumptions pyrefly has been making but if 2 / 3 major type checkers support it I think that is good indicator pyrefly should have it as well :).

sounds good, you've convinced me. sorry about the back/forth. @connernilsen has some thoughts on where this setting should be stored but otherwise we should be good to merge

lsp/package.json
```diff
@@ -240,6 +240,12 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Controls whether comment section folding ranges are included in the editor. When true, comments following the pattern '# Section Name ----' (with 4+ trailing dashes) create collapsible regions, similar to R's code section convention."
+                },
+                "pyrefly.configPath": {
+                    "type": "string",
+                    "default": "",
+                    "description": "Path to a pyrefly.toml or pyproject.toml configuration file. When set, the LSP will use this config for all files instead of an upward-filesystem-walk config discovery approach.",
```
connernilsen: @kinto0 thoughts on this wording?
```suggestion
                    "description": "Path to a pyrefly.toml or pyproject.toml configuration file. When set, the LSP will use this config for all files in your workspace instead of the default Pyrefly config-finding logic. Prefer to use default logic wherever possible.",
```

lsp/README.md
```diff
@@ -38,6 +38,9 @@ settings:
   but not find-references.
 - `pyrefly.lspPath` [string: '']: if your platform is not supported, you can
   build pyrefly from source and specify the binary here.
+- `pyrefly.configPath` [string: '']: path to a `pyrefly.toml` or
+  `pyproject.toml` configuration file. When set, the LSP will use this config
+  for all files instead of an upward-filesystem-walk discovery approach.
```
connernilsen:
```suggestion
- `pyrefly.configPath` [string: '']: path to a `pyrefly.toml` or
  `pyproject.toml` configuration file. When set, the LSP will use this config
  for all files in your workspace instead of the default Pyrefly config-finding logic
  wherever possible.
```

website/docs/IDE.mdx
```diff
@@ -42,6 +42,8 @@ The following configuration options are IDE-specific and exposed as VSCode setti
   - `python.pyrefly.displayTypeErrors` [enum: default, force-on, force-off, error-missing-imports]: If `'default'`, Pyrefly will only provide type checking diagnostics in the IDE if your file is covered by a [Pyrefly configuration](../configuration). If `'force-off'`, Pyrefly will never provide type check diagnostics in the IDE. If `'force-on'`, Pyrefly will always provide type check diagnostics in the IDE. If `'error-missing-imports'`, Pyrefly will only show errors for missing imports and missing sources (`missing-import`, `missing-source`, and `missing-source-for-stubs`). Note that syntax errors (`invalid-syntax` and `parse-error`) are shown by default in every mode besides `force-off`.
 - Specify a custom Pyrefly Binary (lspPath)
   - `pyrefly.lspPath` [string: '']: If your platform is not supported, you can build pyrefly from source and specify the binary path with the `lspPath` config.
+- Specify a custom configuration file (configPath)
+  - `pyrefly.configPath` [string: '']: Path to a `pyrefly.toml` or `pyproject.toml` configuration file. When set, the LSP will use this config for all files instead of an upward-filesystem-walk discovery approach.
```
connernilsen:
```suggestion
- Specify a custom configuration file (configPath)
  - `pyrefly.configPath` [string: '']: path to a `pyrefly.toml` or
  `pyproject.toml` configuration file. When set, the LSP will use this config
  for all files in your workspace instead of the default Pyrefly config-finding logic
  wherever possible.
```

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1931,7 +1935,27 @@ impl Server {
...
+
+        let config_finder = if let Some(config_path) = explicit_config {
+            let (mut config_file, mut errors) = ConfigFile::from_file(&config_path);
+            errors.extend(config_file.configure());
+            let finder = ConfigFinder::new_constant(ArcId::new(config_file));
+            finder.add_errors(errors);
+            finder
+        } else {
+            Workspaces::config_finder(workspaces.dupe())
+        };
```
connernilsen: Instead of storing/loading a config here, which will unconditionally override the config for the entire server, I think it would be preferred if we could do this on a per-workspace approach, which would let us update the config being used without having to restart everything.

For that, what we'll need to do is:
1. on `lsp::non_wasm::workspace::Workspace`, add a field to store the workspace-set config, if present
2. in `lsp::non_wasm::workspace::Workspaces::apply_client_configuration`, handle the new `pyrefly.configPath` field you added to store it on the config being updated
3. update `lsp::non_wasm::workspace::PyreflyClientConfig` to support parsing the new `configPath` value
4. update `lsp::non_wasm::workspace::WorkspaceConfigConfigurer::configure` to use the workspace-set config, if present, instead of the found config (example below)

```rust
impl ConfigConfigurer for WorkspaceConfigConfigurer {
    fn configure(
        &self,
        root: Option<&std::path::Path>,
        mut config: ConfigFile,
        mut errors: Vec<pyrefly_config::finder::ConfigError>,
    ) -> (ArcId<ConfigFile>, Vec<pyrefly_config::finder::ConfigError>) {
        if let Some(dir) = root {
            self.0.get_with(dir.to_owned(), |(workspace_root, w)| {
                if let Some(workspace_config_path) = &w.workspace_config {
                    match ConfigFile::from_file(workspace_config_path) {
                        Ok(new_config) => {
                            config = new_config;
                            // clear any errors, since we're using a different config
                            errors = vec![];
                        }
                        // replace the old errors with errors from parsing the requested config, but use whatever we we found previously
                        Err(new_errors) => errors = new_errors,
                    }
                }
                ...
```

yeetypete:
Thank you both for the discussion and the PR comments. No need to apologize, I think its great that you question the necessity of adding new functionality to pyrefly :)

I'll make sure to address them as soon as I get a chance (likely by Monday or Tuesday)!

yeetypete:
Apologies for the delay, I got caught up in some other things. I'll allocate some time to get this done this week.

connernilsen:
Sounds good!

pyrefly/lib/lsp/non_wasm/server.rs
yeetypete: I implemented it as you suggested. The only change I made was to prefix the configuration with `python.*` like the other config options are. So the full configuration option now reads `python.pyrefly.configPath`. I guess if we put the optiion in `PyreflyClientConfig` then `python.pyrefly.configPath` must be the setting name anyways.

yangdanny97 (approved):

PR #2483, opened 2026-02-21
impl code action to convert dict to TypedDict / dataclass / pydantic #2144

### LINKED ISSUE #2144: code action to convert dict to TypedDict / dataclass / pydantic
### Describe the Bug

https://stackoverflow.com/questions/62267544/generate-pydantic-model-from-a-dict

even a website for this

https://jsontopydantic.com/

usage:

I have code like this

```
def f():
    return {"a": 1, "b": [1, 2]}
```

```
class F(TypedDict):
    a: int
    b: list[int]

def f() -> F:
    return {"a": 1, "b": [1, 2]}
```
convert them to handle something like `f()["a"]`

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2144

Implemented dict-to-model refactor code actions (TypedDict, dataclass, Pydantic) and wired them into LSP refactors, with imports and scoped insertion logic.

note: to keep simple, currently nest case is not considered（fallback to dict）. Will impl in the future pr.

# Test Plan

Added coverage for the new actions.

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/dict_definition.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/dict_definition.rs
@@ -1,0 +1,528 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+use std::collections::HashSet;
+
+use dupe::Dupe;
+use lsp_types::CodeActionKind;
+use pyrefly_build::handle::Handle;
+use pyrefly_python::ast::Ast;
+use pyrefly_python::module::Module;
+use ruff_python_ast::AnyNodeRef;
+use ruff_python_ast::Expr;
+use ruff_python_ast::ExprDict;
+use ruff_python_ast::ModModule;
+use ruff_python_ast::Stmt;
+use ruff_python_ast::helpers::is_docstring_stmt;
+use ruff_text_size::Ranged;
+use ruff_text_size::TextRange;
+use ruff_text_size::TextSize;
+
+use super::extract_shared::code_at_range;
+use super::extract_shared::find_enclosing_statement_range;
+use super::extract_shared::is_exact_expression;
+use super::extract_shared::line_indent_and_start;
+use super::extract_shared::selection_anchor;
+use super::extract_shared::split_selection;
+use super::extract_shared::unique_name;
+use super::types::LocalRefactorCodeAction;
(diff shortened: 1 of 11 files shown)


kinto0:
sorry about the delay - happy to review this once merge conflicts are fixed and copilot review resolved. and are there any opportunities to reuse more common functionality to avoid reimplementing?

pyrefly/lib/state/lsp/quick_fixes/dict_definition.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+fn collect_dict_fields(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    stdlib: &Stdlib,
+    dict_expr: &ExprDict,
+    requires_any: &mut bool,
+) -> Option<Vec<DictField>> {
+    let mut fields = Vec::new();
```
kinto0: maybe this should just be a hashmap? then we don't need to track `seen` and we get the correct behavior (last one should be inserted if there are duplicates).

pyrefly/lib/state/lsp/quick_fixes/dict_definition.rs
```diff
@@ -0,0 +1,466 @@
...
+    if !(first.is_alphabetic() || first == '_') {
+        return false;
+    }
+    if !chars.all(|c| c.is_alphanumeric() || c == '_') {
+        return false;
+    }
+    !is_keyword(name)
+}
+
+fn is_keyword(name: &str) -> bool {
```
kinto0: this shouldn't be done manually - can we use something from the AST for this?

kinto0 (changes requested):
seems pretty good. just a few suggestions and I'm curious why it's called `dict_definition`. should it be renamed to something active like `convert_dict`?

pyrefly/lib/state/lsp/quick_fixes/dict_definition.rs
asukaminato0721: For duplicate keys, I did not switch to a plain HashMap because that would lose deterministic field order.

I now keep the `Vec<DictField>` for stable output order and uses a `HashMap<String, usize>` to overwrite the existing field when a duplicate key appears.

pyrefly/lib/state/lsp/quick_fixes/dict_definition.rs
asukaminato0721: now uses Ruff’s `ruff_python_stdlib::identifiers::is_identifier`

yangdanny97 (approved):

PR #2484, opened 2026-02-21
fix Renaming a kwarg does not work across files. #1583

### LINKED ISSUE #1583: Renaming a kwarg does not work across files.
### Describe the Bug

Currently, updating a kwarg will update the definition of the kwarg as well as all of the locations it is used within the same module. However, this currently does not work across modules. We need to ensure that if we rename a kwarg, it will update locations in all modules.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1583

Updated cross-module rename logic so keyword-argument references are included when the parameter definition lives in a different file. This now resolves keyword args by name, then refines against the definition module’s AST to match the exact parameter.

# Test Plan

Added a cross-file rename test and fixtures to cover kwarg renames across modules

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2414,7 +2414,8 @@ impl<'a> Transaction<'a> {
         results
     }
 
-    /// Return the cached AST for a module, parsing it if the module is not open.
+    /// Return the cached AST for a module, parsing its contents if unavailable.
+    /// Files that Pyrefly has not explicitly opened may have module information but no cached AST.
     fn get_ast_or_parse_module(&self, handle: &Handle, module: &ModuleInfo) -> Arc<ModModule> {
         let module_handle = Handle::new(
             module.name(),
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2387,20 +2387,7 @@ impl<'a> Transaction<'a> {
 
         let mut results: Vec<FindDefinitionItem> = Vec::with_capacity(location_count);
         for (module_info, ranges) in modules_to_ranges.into_iter() {
-            let ast = {
-                let handle = Handle::new(
-                    module_info.name(),
-                    module_info.path().dupe(),
-                    handle.sys_info().dupe(),
-                );
-                self.get_ast(&handle).unwrap_or_else(|| {
-                    // We may not have the AST available for the handle if it's not opened -- in that case,
-                    // Re-parse the module to get the AST.
-                    Ast::parse(module_info.contents(), module_info.source_type())
-                        .0
-                        .into()
-                })
-            };
+            let ast = self.get_ast_or_parse_module(handle, &module_info);
 
             for range in ranges.into_iter() {
                 let (metadata, definition_range) = if let Some(param_range) =
@@ -2425,6 +2412,17 @@ impl<'a> Transaction<'a> {
             }
         }
         results
+    }
+
+    /// Return the cached AST for a module, parsing it if the module is not open.
+    fn get_ast_or_parse_module(&self, handle: &Handle, module: &ModuleInfo) -> Arc<ModModule> {
+        let module_handle = Handle::new(
+            module.name(),
+            module.path().dupe(),
+            handle.sys_info().dupe(),
+        );
+        self.get_ast(&module_handle)
+            .unwrap_or_else(|| Ast::parse(module.contents(), module.source_type()).0.into())
     }
 
     fn get_callee_location(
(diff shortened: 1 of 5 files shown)


kinto0 (changes requested):

kinto0:
commented [here](https://github.com/facebook/pyrefly/pull/3290#issuecomment-4836318897)

pyrefly/lib/state/lsp.rs
```diff
@@ -3575,13 +3579,38 @@ impl<'a> Transaction<'a> {
...
+                        || (module.code_at(*def_range) == definition_name()
+                            && module.to_lsp_position(def_range.start()).line == definition_line())
                     {
                         references.push(*ref_range);
                     }
                 }
             }
         }
+
+        let symbol_kind = match definition_metadata {
```
kinto0: why not handle attribute here?

kinto0: might be worth enumerating all matches and comments why you don't include some. it's very unclear what this variable is used for and it has a very generic name

pyrefly/lib/state/lsp.rs
kinto0: this function seems unnecessary now

kinto0, line 3802: would it be cleaner to put the logic here? how it works now, we add special-casing in local_references_from_external_definition and local_references_from_local_definition.

pyrefly/lib/state/lsp.rs
```diff
@@ -3575,13 +3579,38 @@ impl<'a> Transaction<'a> {
...
+            DefinitionMetadata::Variable(kind) | DefinitionMetadata::VariableOrAttribute(kind) => {
+                *kind
+            }
+            _ => None,
+        };
+
+        if matches!(
+            symbol_kind,
+            Some(SymbolKind::Parameter) | Some(SymbolKind::Variable)
```
kinto0: why not just parameter? what variables are relevant?

pyrefly/lib/state/lsp.rs
```diff
@@ -3575,13 +3579,38 @@ impl<'a> Transaction<'a> {
             if attribute_module_path == module.path() {
                 for (def_range, ref_range) in def_and_ref_ranges {
                     if *def_range == definition_range
-                        || module.to_lsp_position(def_range.start()).line == definition_line()
+                        || (module.code_at(*def_range) == definition_name()
+                            && module.to_lsp_position(def_range.start()).line == definition_line())
```
kinto0: should these changes to include name be done separately? what happens with aliases?

kinto0 (changes requested):
seems like a necessary approach to compare string content, but it's somewhat confusing. I wonder if we could improve the ergonomics and make it less special-casey

pyrefly/lib/state/lsp.rs
```diff
@@ -3923,34 +3949,37 @@ impl<'a> Transaction<'a> {
...
-        let definition_module = self.get_module_info(handle)?;
+        let definition_ast = if handle.path() == definition_module.path() {
+            self.get_ast(handle)?
+        } else {
+            let definition_handle = Handle::new(
+                definition_module.name(),
+                definition_module.path().dupe(),
+                handle.sys_info().dupe(),
+            );
+            self.get_ast(&definition_handle).unwrap_or_else(|| {
```
kinto0: this seems necessary for unopen modules. instead of silently copying this precedent, can we factor out a helper and share it between here + find_definition_for_keyword_argument?

pyrefly/lib/test/lsp/lsp_interaction/rename.rs
```diff
@@ -186,6 +186,65 @@ fn test_rename_editable_package_symbols_is_allowed() {
     interaction.shutdown().unwrap();
 }
 
+#[test]
+fn test_rename_kwarg_across_files() {
```
kinto0, line 260: nit: can we add a negative test? like a second foo(name=...) call to an unrelated function in uses.py

kinto0 (changes requested):

pyrefly/lib/state/lsp.rs
```diff
@@ -2387,35 +2387,44 @@ impl<'a> Transaction<'a> {
...
+                    metadata,
+                    definition_range,
                     module: module_info.dupe(),
                 })
             }
         }
         results
     }
 
+    /// Return the cached AST for a module, parsing it if the module is not open.
```
kinto0: I might put more details here. some info like "files that were not explicitly opened by Pyrefly will not have their AST"

kinto0 (approved):

stroxler (approved):

kinto0:
re-running mypy primer since that result doesn't make a lot of sense to me

PR #2485, opened 2026-02-21
fix handling of bool and abs values #924

### LINKED ISSUE #924: fix handling of bool and abs values
### Describe the Bug

```

from typing import *


def foo(expected: Union[bool, int, float, complex]):
    return abs(expected) # this is an error
    

foo(False) # outputs 0 in abs

abs(False) # is fine - only when bool is part of a union
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUCQCYCmwUwYYAFOQB4LkDGM5pAXFAKopIwKANoAjWgBsANJhQwZwCWACG8qK0gIJTALoBKLkSjGoIcjACuIFFGWiAzg2ZsOpPVADEsABZJ7mf2UbchBwECMTEhp6ADFlCXtydy8wCxgENP8ABllbBxI7RziEpM8A6lRyKABaKCEJOCgAd29yG3EwCXKEZTw6qmUoCwEhIA&version=3.12

_No response_

asukaminato0721:
# Summary

Fixes #924

Adjusted subtype checking so unbound type variables can widen across union members, which fixes generic calls like abs(x) on a union without forcing the first member’s specialization.

# Test Plan

Added a regression test that asserts abs(bool|int|float|complex) yields int | float.


maggiemoss:
cc: @stroxler are you best suited to review this one?

stroxler (commented):
Thanks for the fix! This is actually under the umbrella of #105, which Rebecca is working on and I think the more root-cause fix she has will take priority over this, so I'm going to close the PR and assign the task to her. It's good work though!

PR #2488, opened 2026-02-22
WYSIWYG display of generic type arguments with defaults

### LINKED ISSUE #2461: WYSIWYG reporting of generic type arguments default
In recent versions of NumPy, the two `np.ndarray` generic type parameters default to `tuple[Any, ...]` and `numpy.dtype`, respectively. The generic type parameter of `numpy.dtype` defaults to `Any`. 
So writing `_: np.ndarray` is equivalent to writing `_: np.ndarray[tuple[Any, ...], np.dtype[Any]]`. And of course, the opposite is also the case, i.e., `_: np.ndarray[tuple[Any, ...], np.dtype[Any]]` can be written more compactly as `_: np.ndarray`. 

When reporting types, it looks like pyrefly currently always chooses to "fill in" the defaults of type parameters that aren't provided. For example:

```py

import numpy as np

def f(a: np.ndarray, /) -> None:
    reveal_type(a)
```

```
 INFO revealed type: ndarray[tuple[Any, ...], dtype[Any]] [reveal-type]
```

One obvious issue is that this is needlessly verbose: `"ndarray[tuple[Any, ...], dtype[Any]]"` counts 36 characters, while `"ndarray"` requires only 7; that's a factor of >5x.

It can also be confusing to users to see the type they explicitly wrote as `ndarray` to be printed as `ndarray[tuple[Any, ...], dtype[Any]]`. I can imagine that users new to typing in Python would not understand where the `[tuple[Any, ...], dtype[Any]]` came from.

There's also subjective loss of information: In natural language there are many synonyms, but even though they *mean* the same, there are often subtle context-dependent nuances that cause people to use them with specific intent, so that they can express themselves a little bit better. The way I see it, the same can be said about `ndarray`. For example, if one writes `_: ndarray[tuple[Any, ...]]` it highlights that the shape-type is unknown but relevant, whereas writing `_: ndarray` could mean that the shape-type isn't relevant. 

So all things considered, I would like for pyrefly to report types as they are written. So with this [WYSIWYG](https://en.wikipedia.org/wiki/WYSIWYG) approach you'd have

```py
def f(a: np.ndarray, /) -> None:
    reveal_type(a)  # revealed type: ndarray [reveal-type]

def f(a: np.ndarray[tuple[Any, ...]], /) -> None:
    reveal_type(a)  # revealed type: ndarray[tuple[Any, ...]] [reveal-type]
```

and so forth.

Adist319:
## Summary

- When generic types have default type parameters (PEP 696), omit trailing type args that match their defaults in displayed types
- Types are shown as written rather than with all defaults filled in
- Adds `TArgs::display_count()` method that strips trailing default-matching args at display time

### Examples

| Before | After |
|--------|-------|
| `Generator[str, None, None]` | `Generator[str]` |
| `AsyncGenerator[int, None]` | `AsyncGenerator[int]` |
| `ExceptionGroup[Exception]` | `ExceptionGroup` |
| `MyClass[float, bool]` (non-defaults) | `MyClass[float, bool]` (unchanged) |

Closes #2461

## Test plan

- Added 5 new WYSIWYG test cases in `generic_legacy.rs` covering bare generic, partial, full, no-defaults, and invalid-order scenarios
- Updated error message expectations in `yields.rs`, `contextual.rs`, `flow_branching.rs`, `operators.rs`, `constructors.rs`

### DIFF
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -211,16 +211,13 @@ impl TArgs {
     /// Returns the number of type arguments to display, stripping trailing args
     /// that match their parameter defaults (WYSIWYG display per issue #2461).
     pub fn display_count(&self) -> usize {
-        let paired: Vec<_> = self.iter_paired().collect();
-        let mut count = paired.len();
-        for (param, arg) in paired.iter().rev() {
-            if param.default().is_some() && *arg == &param.as_gradual_type() {
-                count -= 1;
-            } else {
-                break;
+        let mut last_non_default = 0;
+        for (i, (param, arg)) in self.iter_paired().enumerate() {
+            if param.default().is_none() || arg != &param.as_gradual_type() {
+                last_non_default = i + 1;
             }
         }
-        count
+        last_non_default
     }
 
     /// Apply a substitution to type arguments.
--- a/pyrefly/lib/test/generic_legacy.rs
+++ b/pyrefly/lib/test/generic_legacy.rs
@@ -515,6 +515,20 @@ U = TypeVar('U', default=int)
 V = TypeVar('V')
 class MyClass(Generic[T, U, V]):  # E: Type parameter `V` without a default cannot follow type parameter `U` with a default
     ...
+def f(x: MyClass[str, int, bool]) -> None:
+    reveal_type(x)  # E: revealed type: MyClass[str, int, bool]
+    "#,
+);
+
+testcase!(
+    test_wysiwyg_explicit_args_match_defaults,
+    r#"
+from typing import Generic, TypeVar, reveal_type
+T = TypeVar('T', default=int)
+U = TypeVar('U', default=str)
+class MyClass(Generic[T, U]): ...
+def f(x: MyClass[int, str]) -> None:
+    reveal_type(x)  # E: revealed type: MyClass
     "#,
 );
 


crates/pyrefly_types/src/types.rs
```diff
@@ -208,6 +208,21 @@ impl TArgs {
         self.0.1.is_empty()
     }
 
+    /// Returns the number of type arguments to display, stripping trailing args
+    /// that match their parameter defaults (WYSIWYG display per issue #2461).
+    pub fn display_count(&self) -> usize {
+        let paired: Vec<_> = self.iter_paired().collect();
```
yangdanny97: you don't need to collect here, you can iterate it directly

pyrefly/lib/test/generic_legacy.rs
```diff
@@ -454,6 +454,70 @@ def f9(c1: C[int, str], c2: C[str]):
...
+);
+
+testcase!(
+    test_wysiwyg_middle_default_not_stripped,
+    r#"
+from typing import Generic, TypeVar, reveal_type
+T = TypeVar('T')
+U = TypeVar('U', default=int)
+V = TypeVar('V')
+class MyClass(Generic[T, U, V]):  # E: Type parameter `V` without a default cannot follow type parameter `U` with a default
```
yangdanny97, line 516: I don't see any assertions on what type is printed

pyrefly/lib/test/generic_legacy.rs
```diff
@@ -454,6 +454,70 @@ def f9(c1: C[int, str], c2: C[str]):
...
+def f(x: MyClass[float]) -> None:
+    reveal_type(x)  # E: revealed type: MyClass[float]
+    "#,
+);
+
+testcase!(
+    test_wysiwyg_fully_specified_generic,
+    r#"
+from typing import Generic, TypeVar, reveal_type
+T = TypeVar('T', default=int)
```
yangdanny97, line 489: We're missing a test case here for when the explicit user-provided type arguments match the defaults

yangdanny97:
@rchen152 noted during review that PR does not exactly implement WYSIWYG

WYSIWYG would mean that even if the user writes explicit type arguments that do match the defaults, we will display it. Right now, this PR makes it so that we do not display any type arguments that match the default.

I actually prefer the approach this PR takes over the suggestion in the original issue since types are displayed consistently & compactly, but want to double check that this is OK w/ @jorenham

jorenham:
> @rchen152 noted during review that PR does not exactly implement WYSIWYG
> 
> WYSIWYG would mean that even if the user writes explicit type arguments that do match the defaults, we will display it. Right now, this PR makes it so that we do not display any type arguments that match the default.
> 
> I actually prefer the approach this PR takes over the suggestion in the original issue since types are displayed consistently & compactly, but want to double check that this is OK w/ @jorenham

Yea I'm ok with that. The downside is that it might be a bit confusing for users that if they write `dtype[Any]` it'd be reported as `dtype`, but like you said, there's also something (or a lot IMO) to be said for optimizing for compactness/readability.

Adist319:
@yangdanny97 @jorenham Thanks for the review! Pushed the latest commit that removed the unnecessary .collect(), added reveal_type assertions where it was missing, and added the missing test case for explicit args matching defaults.

And yes, the WYSIWYG approach isn't "followed" when explicit type arguments match the default. Agreed it could be a little confusing, but my thought process was that they're equivalent types, and distinguishing "user wrote the defaults" from "defaults were filled in" would mean tracking provenance through the type system. Reasonable tradeoff imo, but I'm open to more suggestions before closing this PR.

pyrefly/lib/test/generic_legacy.rs
```diff
@@ -454,6 +454,84 @@ def f9(c1: C[int, str], c2: C[str]):
...
+U = TypeVar('U', default=str)
+class MyClass(Generic[T, U]): ...
+def f(x: MyClass) -> None:
+    reveal_type(x)  # E: revealed type: MyClass
+    "#,
+);
+
+testcase!(
+    test_wysiwyg_partial_generic_one_default,
```
yangdanny97, line 474: I think I'd prefer if we renamed the tests to compact_display or something like that instead of wysiwyg, to avoid it being misleading

yangdanny97: I can patch that after import. the rest of it looks fine to me

PR #2489, opened 2026-02-22
try fix  for non ABC base class, abstract method with NotImplementedError be infer as Never then in an inconsistent manner #2456

### LINKED ISSUE #2456: for non ABC base class, abstract method with NotImplementedError be infer as Never then in an inconsistent manner
### Describe the Bug

I wonder is this by design?

```py
class A:
    @abstractmethod
    def foo(self):
        raise NotImplementedError()

class B(A):
    def foo(self):
        x = 1
        print(x)
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn0%2BjA0pYfkYKkxsMCAEEAEIAYQUqGwcA4QimHDY3Mw%2BkYUFQcGUWtqvmoroAAkaTWbTpaYtaXa6iWUwLhcNsFlAwLtna7Y7GzUp2AA5XAcL6Y2DPJyYP6XWF7BXoW325Q67Za6P6ONB3Kh8MMKMxuOx-DUAC81AAjFXm67iA4Xtt8PT0CA5CAAtA4CRyIgqHMulBHBIwBzTmT0HBCzWQ0IWAU1ugLCxsGkh4gIhx9nUAHzUL1N10iYG%2BMC6EBJ4%2BnoQX4D4AC%2B776GOIBkCIYikIQXAsFAFA0II4EipUsitsw6CQFoRQFBu8QGFIMDsDwHAcMQcCIAA9ORYGiOIhCwlo5EwOg5GYOicDkWhGFYV0sTkbWEqoEsqDQEaOScRAmFmjxvi4MQ0nTvoZDcugdRUnAG7ttQ74AMyEJ2ABMQGjv%2B47mhAKwAGLQDAFDIdUM4gP%2BQA

_No response_

asukaminato0721:
# Summary

Fixes #2456

Treat trivial raise NotImplementedError() bodies as abstract when either `@abstractmethod` is present or the class extends ABC, so inference returns Any instead of Never.

# Test Plan

Added a regression test for non-ABC classes using `@abstractmethod` with NotImplementedError.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -3062,6 +3062,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 &want_class_field,
                 &Instance::of_protocol(parent, self.instantiate(cls)),
             );
+            // If parent return type was inferred as Never (for example a method that only raises),
+            // skip override consistency checks so we preserve reachability behavior without
+            // producing noisy bad-override diagnostics.
+            if !want_class_field.has_explicit_annotation()
+                && want_attribute
+                    .clone()
+                    .as_instance_method()
+                    .and_then(|ty| ty.callable_return_type(self.heap))
+                    .is_some_and(|ret| ret.is_never())
+            {
+                continue;
+            }
             if got_attribute.is_none() {
                 // Optimisation: Only compute the `got_attr` once, and only if we actually need it.
                 got_attribute = Some(self.as_instance_attribute(
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -3117,7 +3117,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 yields,
                 yield_froms,
                 body_is_trivial,
-                body_is_raise_not_implemented_error,
                 class_metadata_key,
             } => {
                 let is_generator = !(yields.is_empty() && yield_froms.is_empty());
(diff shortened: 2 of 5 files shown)


rchen152 (changes requested):
This PR seems to bundle two changes: (1) always infer `Any` when a function body is just `raise NotImplementedError()`, and (2) always treat methods decorated with `@abstractmethod` as abstract even if the class isn't abstract. Could you split these up so we can evaluate them separately?

rchen152:
A couple more thoughts on this, after a re-read of https://github.com/facebook/pyrefly/issues/2456:
* If we implement Jia's suggestion to never infer `Never` as a return type in a method in a base class, I believe that makes the abstractmethod check unnecessary. That is, you can probably just revert the changes related to is_abstract_method in this PR rather than splitting them out; I think your other change is already sufficient.
* Looking at the mypy_primer results, the false positives that this change eliminates are exclusively [bad-override] errors. This makes me wonder if we could take a different approach and disable the override check when the inferred return type is `Never`, rather than changing the return type to `Any`. The advantage of that is that we'd still be able to tell when code is unreachable after a call to a function that raises `NotImplementedError`. On the other hand, if we're not seeing any new false positives related to reachability analysis, maybe it's fine to stick with the current approach, which seems fairly simple and straightforward if we revert the is_abstract_method changes.

Issue #2382, opened 2026-02-11
unbound name when using walrus operator within a ternary expression


jakevdp:
Short reproduction of the issue:
```python
# test.py
def f(i):
  return a if (a := round(i)) - 1 else a + 1
print(f"result is {f(1.0)}") 
```
This executes correctly:
```
$ python test.py
result is 2
```
But pyrefly raises an error (using v0.52.0):
```
$ pyrefly check test.py
ERROR `a` is uninitialized [unbound-name]
 --> test.py:3:40
  |
3 |   return a if (a := round(i)) - 1 else a + 1
  |                                        ^
  |
 INFO 1 error
```

### Sandbox Link

[sandbox link](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABQQCUiAOujTQE4wAuArlw6oaEWvRGIAvN1z8sjJkxoBaGgEYaMKHBg0RAag3tiXCOl70wrEDzj8ovUXBrAG6wgAYmAXxvKQABoQfl5oOBJyRBAAYhoAVTCoCF5SOnkAYzDcdDh2dipaMFwuAFtUXgB9dH5S7BguenxEUQtlFQA%2BGjheLjYOTm4%2BQQ5rEAA5Wvq%2B13w-EHYgkDIeMChSQl5cUqgKOIAFUlX17owcAhoMnMgAc0EKiBzCdjiAZRg9AAteXmI4RAA9ACVtR1oQSjcATB0ADMLgMnAAVd0Ld7tkYXQSvoAG6oaCobCwS7XCB3LgPHI0XDEdEREykXifHIqbENOCPDgyGwAZkI6gATDZFj5gqgshBWQAxaAwChoLB4IhkEA%2BIA)

jakevdp:
I came across what looks like another instance of the same bug, but without the ternary expression:
```python
def f(a):
    if (x := a) > 0:
        pass
    return x
```
```
$ uv run pyrefly check test.py
ERROR `x` may be uninitialized [unbound-name]
 --> test.py:4:12
  |
4 |     return x
  |            ^
  |
 INFO 1 error
```

samwgoldman:
The version with an if-statement still reproduces

Adist319:
Handled the if-statement case in this PR: #2493

yangdanny97:
[...]

PR #2493, opened 2026-02-22
Fix walrus operator in if-statement test not visible after branch

Adist319:
## Summary
- Fix false `unbound-name` errors for variables assigned via walrus operator (`:=`) in if-statement test conditions
- Process the first `if` test expression before `start_fork()` so walrus-defined names are in the base flow and visible after the if-statement
- Mirrors the existing fix for ternary expressions in `expr.rs` (commit eaac26fab)

## Test Cases Added
- `test_walrus_in_if_basic` — basic `if (x := a) > 0: pass; return x`
- `test_walrus_in_if_both_branches` — walrus visible in both if and else
- `test_walrus_in_if_with_narrowing` — walrus with `is not None` narrowing
- `test_walrus_in_elif` — correctly reports `x` may be uninitialized (elif is conditional)
- `test_walrus_in_if_no_else` — walrus visible after if with no else clause
- Updated `test_walrus_on_first_branch_of_if` to reflect correct behavior
- Updated `test_false_and_walrus` with `bug =` annotation for known BoolOp laxness false negative

Fixes https://github.com/facebook/pyrefly/issues/2382

### DIFF
--- a/pyrefly/lib/binding/stmt.rs
+++ b/pyrefly/lib/binding/stmt.rs
@@ -934,10 +934,14 @@ impl<'a> BindingsBuilder<'a> {
                     is_while_true,
                 );
             }
-            Stmt::If(x) => {
+            Stmt::If(mut x) => {
                 let is_definitely_unreachable = self.scopes.is_definitely_unreachable();
                 let mut exhaustive = false;
                 let if_range = x.range;
+                // Process the first `if` test before forking so that walrus-defined names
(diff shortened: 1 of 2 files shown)


yangdanny97:
There was one very strange false positive that we ran into when checking Instagram, that's what's holding this PR up. I'll try my best to finish investigating it this week

yangdanny97:
thanks @Adist319 !

PR #2494, opened 2026-02-23
fix Feature: support @field.default decorator for attrs #2012

### LINKED ISSUE #2012: Feature: support `@field.default` decorator for attrs
### Describe the Bug

Attrs allows setting a default factory for a dataclass field `x` by decorating a function with `@x.default`. I don't think we support this today

https://www.attrs.org/en/stable/init.html#defaults

Looks similar to `x.setter`/`x.deleter` for properties, so I imagine the implementation for that might be a good reference

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2012

Recognize `@<field>.default` on attrs field specifiers in class bodies, suppress the bogus attribute error, and mark the decorated method with an attrs-default flag so it’s treated as a special decorator.

Use that flag when building dataclass/attrs field metadata to treat the field as having a default (so `__init__` no longer requires it).

# Test Plan

Updated the attrs test to remove the bug marker and expected errors.


yangdanny97:
@shobhitmehro this task is technically on your TODO for the internship, you can review the PR and import/merge if it matches how you were planning to do it

PR #2506, opened 2026-02-23
fix Dataclass/Pydantic: go-to-def on named arguments in an __init__ call should go to specific fields #1231

### LINKED ISSUE #1231: Dataclass/Pydantic: go-to-def on named arguments in an `__init__` call should go to specific fields
### Describe the Bug

```
# 1. When using Go To Definition on a model field in `__init__()`, point to the annotated attribute `a`, not the model name:
class Model5(BaseModel):
    a: int

# Go to definition here:
# -----|
#      |
Model5(a=1)
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1231

Updated keyword-argument go-to-definition so class-constructor calls prefer class field annotations and fall back to `__init__` parameters, which aligns with issue #1231.

# Test Plan

Added a class-field keyword-argument definition test

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2497,12 +2497,8 @@ impl<'a> Transaction<'a> {
             let ast = self.get_ast_or_parse_module(handle, &module_info);
 
             for range in ranges.into_iter() {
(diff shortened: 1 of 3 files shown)


stroxler (approved):

PR #2525, opened 2026-02-24
[conformance] fix Passing an unpacked tuple of types as type arguments to the type parameters [T, *Ts] gets error #2363 

### LINKED ISSUE #2363: Passing an unpacked tuple of types as type arguments to the type parameters `[T, *Ts]` gets error
### Describe the Bug

*Memo:
- pyrefly check test.py
- pyrefly 0.46.1
- Python 3.14.0

Passing an unpacked tuple of types as type arguments to the type parameters `[T, *Ts]` gets the error as shown below:

*Memo:
- mypy and ty don't get error.

```python
from typing import Unpack

class Cls[T, *Ts]: ...

v1: Cls[*tuple[int, str, int, str]]        # Error
v2: Cls[Unpack[tuple[int, str, int, str]]] # Error
```

> ERROR Unpacked argument cannot be used for type parameter T 

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2363

Updated specialization matching in targs.rs so TypeVarTuple handling can consume `*tuple[T, ...]` while still satisfying trailing required params.

In the `TypeVarTuple` branch, when the current arg is `Type::Unpack(Tuple::Unbounded(_))`, the matcher now materializes missing trailing args as element type T before computing how many args the tuple var should consume.

# Test Plan

Converted the unbounded test from a bug case with expected errors to a passing regression in `type_var_tuple.rs`.

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -7364,28 +7364,6 @@
     {
       "code": -2,
       "column": 8,
-      "concise_description": "Unpacked argument cannot be used for type parameter T1",
-      "description": "Unpacked argument cannot be used for type parameter T1",
-      "line": 153,
-      "name": "bad-unpacking",
-      "severity": "error",
-      "stop_column": 29,
-      "stop_line": 153
-    },
-    {
-      "code": -2,
-      "column": 16,
-      "concise_description": "assert_type(tuple[Any], tuple[*tuple[int, ...], int]) failed",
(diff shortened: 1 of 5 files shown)


asukaminato0721:
cc @migeed-z

migeed-z:
This issue was fixed upstream. Sorry for the confusion! If there are any conformance issue that you would like to look into, please let me know and I will assign it to you.

Issue #1197, opened 2025-09-28
False positive when calling an imported function that shadows a global/built-in


epatey:
### Describe the Bug

Pyrefly checks a function call against the wrong signature when that function shadows a global/built-in.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEArgC7Rx0Q3G6VN0wBuqKAB10ozDDB0A7gAtcuYnAAUASkSi6WugGI6AFVkQ2AYyFQ6ceQyiZGcGHSazHHLjxh2BQumGr0TBjgmWgB9GlxMGxgAGjpsZjoySkkoUnZ0E24UkyY0ulQ4OAYaeHYmAHI2Z0cE6CYAWgh0PkELDDtSCBhbOA0W7R1NQboAQUoAcxKYdF4AAwAZCCYYSiEAbQBWAF059jZ0XF5CuAgJ9FRsWCdcJNQ10pXKOjm4XAZKExg96WXZJ1IxEccwAQgwwGBVnQAD50ADCkRg%2BkBjlhwUoe2avgYmRYuBac28UDmAAVSCkwGllNhUJgGvcpqVZg0mCjVMMtEMBloicpNuz0CAYiBmKwSORECAJFJefhEBkmKo6A0AHx0ABy%2BJg-W0dyKohAAF9hahchB%2BDAAGLQGAUNBYPBEMhGoA&version=3.12

fangyi-zhou:
Observation 1: Shadowed builtin doesn't need to live in a different file

[...]

PR #2533, opened 2026-02-24
fix Incorrect type inference when overriding zip and other builtins #2385

### LINKED ISSUE #2385: Incorrect type inference when overriding `zip` and other builtins
### Describe the Bug

Here is an example of code that leads pyrefly to a false positive error:
```python
def zip(*args):
  return []

def f(a, b):
  return reversed(zip(a, b))
```
```
ERROR No matching overload found for function `reversed.__new__` called with arguments: (type[reversed[_T]], zip[tuple[@_, @_]] | Unknown) [no-matching-overload]
 --> test.py:5:18
  |
5 |   return reversed(zip(a, b))
  |                  ^^^^^^^^^^^
  |
  Possible overloads:
  (cls: type[reversed[_T]], sequence: Reversible[_T], /) -> Iterator[_T] [closest match]
  (cls: type[reversed[_T]], sequence: SupportsLenAndGetItem[_T], /) -> Iterator[_T]
 INFO 1 error
```
This seems to have to do with shadowing the builtin `zip` name; if you replace `zip` with a differently-named function there is no error:
```python
def _zip(*args):
  return []

def f(a, b):
  return reversed(_zip(a, b))
```
```
 INFO 0 errors
```

### Sandbox Link

[sandbox link](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABAF4TEAUAVKgE4DmcAlIgDroaNdjAAuAV3ZCA2gF1BgqrTBNUAGhrY%2Bg4aMnSRMAG4x2cGJiYNmGrTx4h1ICWOhwS5RCADENAKpuUBBipDRgEugAxm646HCKWNThuOwAtqhiAProEmnYZkz4iDQQ6GI8NAC0AHw0cGLsAkLCRgZCYPwgAHJ5BU00wPgAvl2CTiBkomBQpIRiuGlQFH4ACqTTs-UYOAQ0UXGQnFKZEHGEgn4AyjAwNAAWYmLEcIgA9G9T1LOEqZxvMHQb0wuCicDeB3QRxOsSBKXYNFQxlQ0FQ2Fg%2B0OEGO7FOcRouGIsI8gjIYnucSqpnMZyEAF4aF0AMyEACMACYxugQMNnKgYhBTAAxaAwChoLB4IhkHlAA)

_No response_

asukaminato0721:
# Summary

Fixes #2385

Prevented builtins wildcard imports from shadowing existing definitions during static scope setup, so local overrides (like zip) no longer get merged with builtins.

# Test Plan

Added a regression test to cover shadowing zip and using reversed.

### DIFF
--- a/pyrefly/lib/test/simple.rs
+++ b/pyrefly/lib/test/simple.rs
@@ -222,7 +222,10 @@ if cond:
     def zip(*args) -> list[int]:
         return [42, 42, 42]
 
-reveal_type(zip([1], [2]))  # E: revealed type: list[int] | zip[tuple[int, int]]
+def f():
+    return zip([1], [2])
+
+reveal_type(f())  # E: revealed type: list[int] | zip[tuple[int, int]]
 "#,
 );
 
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -387,6 +387,12 @@ impl Static {
             d.inject_implicit_globals();
         }
 
+        let unconditional_defs: SmallSet<Name> = d
+            .definitions
+            .iter_hashed()
+            .filter_map(|(name, def)| (!def.conditional).then(|| (*name.key()).clone()))
+            .collect();
+
         let mut all_wildcards = Vec::with_capacity(d.import_all.len());
         for (m, range) in d.import_all {
             if let Some(wildcards) = lookup.get_wildcard(m) {
@@ -408,12 +414,12 @@ impl Static {
             self.upsert(name, range, style);
         }
         for (module, range, wildcard) in all_wildcards {
-            // Builtins are a fallback, so they should never shadow an existing definition.
+            // Builtins are a fallback, so they should never shadow unconditional definitions.
             let skip_existing =
                 module == ModuleName::builtins() || module == ModuleName::extra_builtins();
             for name in wildcard.iter_hashed() {
                 // TODO: semantics of import * and global var with same name
-                if skip_existing && self.0.get_hashed(name).is_some() {
+                if skip_existing && unconditional_defs.contains(*name.key()) {
                     continue;
                 }
                 self.upsert(name.cloned(), range, StaticStyle::MergeableImport)
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -100,6 +100,8 @@ pub struct Definition {
     pub style: DefinitionStyle,
     /// A location where the name is defined. Always matches the source of `self.style`.
     pub range: TextRange,
+    /// Whether all definitions for this name are conditional.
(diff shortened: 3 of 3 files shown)


fangyi-zhou:
Presumably this PR also resolves #1197?

rchen152:
It does, thanks!

rchen152 (commented):
Passing along some feedback from @grievejia:

> What would happen if we use conditional overrides like this?
> 
> ```
> if cond():  # or while, try, with, etc.
>     def zip(*args) -> list[int]:
>         return []
>     # Would pyrefly completely disregard the built-in zip here?
> ```
> 
> I'm a little concerned that unconditionally overwriting any built-in names might be too aggressive. What if the user actually wants to reuse the types or names from the built-ins?

asukaminato0721:
```
def f(cond):
      if cond:
          def zip(*args) -> list[int]:
              return []
      return zip([1], [2])
f(True)
f(False)
```

```
> python tmp.py 
Traceback (most recent call last):
  File "/tmp/tmp.py", line 7, in <module>
    f(False)
    ~^^^^^^^
  File "/tmp/tmp.py", line 5, in f
    return zip([1], [2])
           ^^^
UnboundLocalError: cannot access local variable 'zip' where it is not associated with a value
```

asukaminato0721:
> Would pyrefly completely disregard the built-in zip here?

yes
with this change, once a name is defined anywhere in the scope (even conditionally), pyrefly will treat it as a local and not fall back to the builtin. That matches Python’s runtime name resolution.

Why this is correct:
  - In Python, any assignment/def to a name in a scope makes it local for the whole scope.
  - If the conditional branch doesn’t execute, the name is uninitialized, and using it raises UnboundLocalError rather than falling back to the builtin.

asukaminato0721:
add a test to show it.

function scope vs module scope.

pyrefly/lib/test/simple.rs
```diff
@@ -196,6 +196,36 @@ assert_type(y, Literal[1])
...
+    r#"
+from typing import reveal_type
+import random
+
+cond = random.randint(0, 1)
+if cond:
+    def zip(*args) -> list[int]:
+        return [42, 42, 42]
+
+reveal_type(zip([1], [2]))  # E: revealed type: list[int] | zip[tuple[int, int]]
```
rchen152: Could you try putting the `zip(...)` call inside a function? When I was testing this PR, that changed the revealed type (to an incorrect one).

asukaminato0721:
did some check...

update track whether a definition is conditional; propagate this through control‑flow constructs.

only skip builtins wildcard injection when a name is unconditionally defined; conditional defs still merge with builtins.

PR #2534, opened 2026-02-24
Add resolve import message to TSP support

rchiodo:
# Summary

This is the next message in the Type Server Protocol. 

`typeServer/resolveImport` allows a caller to find what file an imported module is from. 

This is inherently a function of the type server because this logic is used when computing types.

# Test Plan

Added a number of unit and integration tests. Ran 'python test.py' as well.

### DIFF
--- a/pyrefly/lib/tsp/requests/resolve_import.rs
+++ b/pyrefly/lib/tsp/requests/resolve_import.rs
@@ -82,7 +82,9 @@ impl<T: TspInterface> TspServer<T> {
         };
 
         // --- 4. Resolve the import via existing infrastructure ---
-        let transaction = self.inner.non_committable_transaction(ide_transaction_manager);
+        let transaction = self
+            .inner
+            .non_committable_transaction(ide_transaction_manager);
         let result = transaction.import_handle(&source_handle, module_name, None);
 
         // --- 5. Convert result to URI string (or null) ---


pyrefly/lib/tsp/requests/resolve_import.rs
```diff
@@ -0,0 +1,265 @@
...
+                self.send_err(id, invalid_params_error("sourceUri is not a file:// URI"));
+                return;
+            }
+        };
+
+        // --- 3. Build source handle and resolve the module name ---
+        let state = self.inner.state();
+        let source_module_path = ModulePath::filesystem(source_path.clone());
+        let source_handle = handle_from_module_path(state, source_module_path);
+
```
kinto0: is this the only API we use from state? if so, can we expose that instead in TspInterface instead of exposing state? exposing state makes it harder to reason about API changes - the more specific we can make the interface the easier it'll be to refactor

pyrefly/lib/tsp/requests/resolve_import.rs
```diff
@@ -0,0 +1,265 @@
...
+            Ok(name) => name,
+            Err(err) => {
+                self.send_err(id, err);
+                return;
+            }
+        };
+
+        // --- 4. Resolve the import via existing infrastructure ---
+        let transaction = ide_transaction_manager.non_committable_transaction(state);
+        let result = transaction.import_handle(&source_handle, module_name, None);
```
kinto0, line 88: this prefers .pyi. is that the intention? when would resolve_import prefer .py files? if that's ever the case, we would want to use `import_handle_prefer_executable` instead

rchiodo: Which does Pyrefly use first internally? If a method is in a type stub, does that take precedence over the real implementation? It does for Pyright.

pyrefly/lib/tsp/requests/resolve_import.rs
rchiodo: I was thinking we'd need this for future usage, but it makes sense to minimize it. I'll refactor.

pyrefly/lib/tsp/requests/resolve_import.rs
kinto0: for what? 

- go-to definition by default uses the .py
- go-to type definition by default uses the .pyi
- types come from the .pyi

rchiodo: Internally I guess. When looking for where an import resolves to, does it have a precedence or is it always operation dependent. Pyright always looks in the stubs first and if it fails there, it looks in the real implementation. Goto def doesn't use this for our TSP situation. It uses this to get the stub file and then finds the real implementation in the corresponding source file.

rchiodo: For pyright the import resolver is used to find the types of stuff, it's why it uses the type stub if it's present.

PR #2536, opened 2026-02-24
Fix false positive type var scoping checks when `Self` is used

### LINKED ISSUE #2526: false positive `invalid-type-var` for `Self` after upgrading to 0.54.0
In several locations in scipy-stubs I'm seeing `invalid-type-var` errors being reported by pyrefly after upgrading to 0.54.0. For example,

```pyi
@final
class _SumLinearOperator(LinearOperator[_SCT1_co | _SCT2_co], Generic[_SCT1_co, _SCT2_co]):
    args: tuple[LinearOperator[_SCT1_co], LinearOperator[_SCT2_co]]

    def __new__(cls, A: LinearOperator[_SCT1_co], B: LinearOperator[_SCT2_co]) -> Self: ...
    def __init__(self, /, A: LinearOperator[_SCT1_co], B: LinearOperator[_SCT2_co]) -> None: ...
    @override
    def _adjoint(self, /) -> Self: ...
```

([src](https://github.com/scipy/scipy-stubs/blob/190dbcc326ee00c288f6a2f4afadc792a0ed8001/scipy-stubs/sparse/linalg/_interface.pyi#L428-L435))

now reports

```
ERROR Type variable `_SCT1_co` is not in scope [invalid-type-var]
   --> scipy-stubs/sparse/linalg/_interface.pyi:432:83
    |
432 |     def __new__(cls, A: LinearOperator[_SCT1_co], B: LinearOperator[_SCT2_co]) -> Self: ...
    |                                                                                   ^^^^
    |
```

This, to me, looks like a false positive, seeing as `_SCT1_co` is most certainly in scope, after all.

See https://github.com/scipy/scipy-stubs/pull/1291 for three other cases like this (there are other issues too there, but I'll open a separate issue for that)

fangyi-zhou:
# Summary

Fixes #2526 

When `Self` is used, there's no need to check for type var scoping
because they don't actually appear in the surface syntax.

# Test Plan

Added a minimised regression test

### DIFF
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -932,6 +932,9 @@ impl Type {
                 //   when visiting Vars. See https://github.com/facebook/pyrefly/issues/2016.
                 Type::ClassType(cls) => recurse_targs(cls.targs()),
                 Type::TypedDict(TypedDict::TypedDict(td)) => recurse_targs(td.targs()),
+                // `Self` is a keyword, not a user-written type variable reference, so we don't
+                // recurse into it when looking for type variable references.
+                Type::SelfType(_) => {}
                 _ => ty.recurse(&mut |ty| visit(ty, f)),
             }
         }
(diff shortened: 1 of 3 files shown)


rchen152:
I think a better fix might be to add a case for `Type::SelfType` here to prevent recursing into it when looking for type variable references: https://github.com/facebook/pyrefly/blob/3209baa517228c8ab1ea2cda11476c077978cea7/crates/pyrefly_types/src/types.rs#L925.

If we go this route, we should also update the similar `transform_raw_legacy_type_variables`: https://github.com/facebook/pyrefly/blob/3209baa517228c8ab1ea2cda11476c077978cea7/crates/pyrefly_types/src/types.rs#L1003.

PR #2566, opened 2026-02-26
negative refinement for type(X) != Y, X.__class__ != Y when Y is final

### LINKED ISSUE #2530: No narrowing on type(...) is T
### Describe the Bug

Reproducer:

```

def f(x: int | bool):  # does not narrow
  if type(x) is bool:
    x = 42
  reveal_type(x)


def g(x: int | None):  # works OK
  if x is None:
    x = 42
  reveal_type(x)
```

Output:

```
 INFO revealed type: bool | int [reveal-type]
 --> /tmp/foo.py:6:14
  |
6 |   reveal_type(x)
  |              ---
  |
 INFO revealed type: int [reveal-type]
  --> /tmp/foo.py:12:14
   |
12 |   reveal_type(x)
   |              ---
   |
 INFO 0 errors
```

### Sandbox Link

_No response_

_No response_

yangdanny97:
Summary:
Previously we did not narrow in the negative case for `type(X)` and `X.__class__`, because it only checks whether something is _exactly_ that type. However,  if the RHS is a final class then we know there can be no subclasses, so it would be safe to narrow it away.

fixes https://github.com/facebook/pyrefly/issues/2530

Differential Revision: D94528111

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -9,6 +9,7 @@ use num_traits::ToPrimitive;
 use pyrefly_config::error_kind::ErrorKind;
 use pyrefly_graph::index::Idx;
 use pyrefly_python::ast::Ast;
+use pyrefly_python::dunder;
 use pyrefly_types::class::Class;
 use pyrefly_types::display::TypeDisplayContext;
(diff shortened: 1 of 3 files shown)


rchen152 (approved):

PR #2567, opened 2026-02-26
always promote literals when resolving comprehensions

yangdanny97:
Summary:
Leaving literal types in comprehensions is too restrictive since the inferred container is invariant

fixes https://github.com/facebook/pyrefly/issues/2517

Differential Revision: D94534921

PR #2568, opened 2026-02-26
Auto-trigger mypy primer classifier 

migeed-z:
Tested the tool on some PRs manually. 
Example run: https://github.com/facebook/pyrefly/pull/2567 

I would like to gather some feedback on the took before tuning it further, so I am proposing to make the workflow automatic.

I added a feedback thumbs up/thumbs down to gather feedback in the last diff. I increased the font size to make it more visible.


grievejia (approved):

PR #2569, opened 2026-02-26
don't infer type[Any] for `|` if either operand is `Any`

### LINKED ISSUE #2498: Incorrect implicit type alias with |
### Describe the Bug

It looks like Pyrefly concludes that `A | B` is a type alias when both `A` and `B` are unknown, e.g.

```
def test(x, y):
  reveal_type(x | y)  # => type[Any]
```

This is not particularly helpful, since it could lead to type errors downstream, continuing the example

```
def f(x: int): ...

def test(x, y):
  f(x | y)
```

which produces

```
ERROR Argument `type[Any]` is not assignable to parameter `z` with type `int` in function `f` [bad-argument-type]
```

### Sandbox Link

_No response_

_No response_

yangdanny97:
Summary: fixes https://github.com/facebook/pyrefly/issues/2498

Differential Revision: D94560390

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -9,6 +9,7 @@ use num_traits::ToPrimitive;
 use pyrefly_config::error_kind::ErrorKind;
 use pyrefly_graph::index::Idx;
 use pyrefly_python::ast::Ast;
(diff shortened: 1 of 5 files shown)


fangyi-zhou (approved):

PR #2571, opened 2026-02-26
support unpack/spread of anonymous typed dicts into dict literals

yangdanny97:
Summary:
Previously, we only inferred an anonymous typed dict if there were no unpackings in the dict literal.

Now, we infer if if all unpacked values are also anonymous typed dicts, supporting a wide range of additional patterns you see in the tests.

This was from grievejia's document of OSS issuest

Differential Revision: D94566803


yangdanny97:
I think this is a net positive, with a few regressions

grievejia (approved):

PR #2572, opened 2026-02-26
[conformance] Model `Annotated` with a distinct Type case

fangyi-zhou:
# Summary

Before this change, we are treating an annotated type `Annotated[T, ...]` as the base type `T`. This is fine in the majority of the places, but does not capture some subtleties according to the spec.

We have failing conformance test cases due to an `Annotated` type being accepted as a `type` or `TypeAlias`, as well as being callable, whereas the spec doesn't allow so. 

https://typing.python.org/en/latest/spec/qualifiers.html#annotated

> As with most [special forms](https://typing.python.org/en/latest/spec/glossary.html#term-special-form), Annotated is not assignable to type or type[T]

> An attempt to call Annotated (whether parameterized or not) should be treated as a type error by type checkers

This diff adds a new Type case for `Annotated`, which contains the base type. The type behaves like the base type, except it cannot be passed to `type` or called.

# Test Plan

Ran conformance testsuite, and observed that `qualifiers_annotated.py` is passing.

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -2241,9 +2241,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
             // At runtime, `Annotated[T, ...]` is an instance of `typing._AnnotatedAlias`,
             // which inherits from `typing._GenericAlias`. We model it as `GenericAlias`.
-            Type::Annotated(_) => {
-                acc.push(AttributeBase1::ClassInstance(self.stdlib.generic_alias().clone()))
-            }
+            Type::Annotated(_) => acc.push(AttributeBase1::ClassInstance(
+                self.stdlib.generic_alias().clone(),
+            )),
             // TODO: check to see which ones should have class representations
             Type::SpecialForm(_)
             | Type::Type(_)


migeed-z:
I'm working on a tool to classify those errors easily. Those look like expected errors to me. I think the LLM classified them as regressions, because it doesn't have the PR title and explanation. It thinks the regressions were not intentional and its feedback is essentially undoing the PR. I will fix that.

migeed-z:
Hmm.. so the diff ended up introducing 100+ errors on IG. cc @rchen152 for opinions on this.

rchen152:
> Hmm.. so the diff ended up introducing 100+ errors on IG. cc @rchen152 for opinions on this.

I only looked at a few errors, but the ones I dug into were caused by pyrefly no longer treating things like `Annotated[int, "metadata"]` as callable. As per https://discuss.python.org/t/is-annotated-compatible-with-type-t/43898, this (pyrefly's new behavior as of this PR) is correct and spec-compliant, and IG was essentially relying on a bug in pyrefly.

fangyi-zhou:
The errors are true positives given a strict reading of the spec, in practice people have been using that in the wild quite often.

See discussion here

https://discuss.python.org/t/is-annotated-compatible-with-type-t/43898

Issue #2564, opened 2026-02-26
Overriding class level functions (without self) fails with bad-override


QEDady:
### Describe the Bug

Pyrefly errors on the following:
```
class A:
    def foo() -> int:
        return 42

class B(A):
    def foo() -> int: # bad-override
        return 100
```
with 
```
ERROR sandbox.py:16:9-12: Class member `B.foo` overrides parent class `A` in an inconsistent manner [[bad-override](https://pyrefly.org/en/docs/error-kinds/#bad-override)]
  `B.foo` has type `() -> int`, which is not assignable to `() -> int`, the type of `A.foo`
```
The above error go away if we add `self`  (or any other parameter) to both functions. Also, annotating them with `@staticmethod` makes the error goes away. Maybe, pyrefly is injecting some thing under the hood to make these fail?

### Sandbox Link

[...]

PR #2577, opened 2026-02-27
Fix false positive bad-override for methods without self parameter

terror:
Resolves https://github.com/facebook/pyrefly/issues/2564

When checking overrides, `BoundMethod` types are compared by first stripping the self parameter via `bind_boundmethod`. This calls `split_first_param`, which returns None when there are no parameters, causing the match guard to fail and fall through to a catch-all error, even when signatures are identical. 

The diff adds a fallback arm that compares the underlying function types directly when self stripping fails.


rchen152 (approved):
Thanks!

PR #2579, opened 2026-02-27
fix More expressive sys.platform gating #795

### LINKED ISSUE #795: More expressive sys.platform gating
### Describe the Bug

```
import sys

assert sys.platform == "win32"
    
import winreg
winreg.OpenKeyEx(winreg.HKEY_LOCAL_MACHINE, key_path)
```
or
```
import sys

if sys.platform != "win32":
    raise EnvironmentError("Couldn't locate HKEY_LOCAL_MACHINE because it's only available on Windows system")
    
import winreg
winreg.OpenKeyEx(winreg.HKEY_LOCAL_MACHINE, key_path)
```

This works with MyPy but doesn't work with Pyrefly.

### Sandbox Link

https://pyrefly.org/sandbox/?code=JYWwDg9gTgLgBAZwJ4IFCoIYIQU1olAOjABsMYAzaEOAXlrgCIB3YAOwGYAmR1OfuKlCR8rNlBwBzVGImTCAeTA42AaRxIAogA8AFLKmEAEqs0BNAPoAZBQGEAglYsBZe7aMBJAHKaANHABrDQswcgALAEpUIA

_No response_

asukaminato0721:
# Summary

Fixes #795

completes the platform-guarded import behavior so modules guarded by assert sys.platform == ... are type-checked and their imports are resolved using the correct platform SysInfo.

# Test Plan

introduces a regression test that a win32-guarded module can import winreg without errors.

### DIFF
--- a/crates/pyrefly_python/src/sys_info.rs
+++ b/crates/pyrefly_python/src/sys_info.rs
@@ -1111,7 +1111,19 @@ impl SysInfo {
 }
 
 pub fn module_platform_guard(body: &[Stmt]) -> Option<PythonPlatform> {
-    body.iter().find_map(platform_guard_from_stmt)
+    let body = match body.split_first() {
+        Some((Stmt::Expr(x), rest)) if x.value.is_string_literal_expr() => rest,
+        _ => body,
+    };
+    for stmt in body {
+        if let Some(platform) = platform_guard_from_stmt(stmt) {
+            return Some(platform);
+        }
+        if !matches!(stmt, Stmt::Import(_) | Stmt::ImportFrom(_)) {
+            return None;
+        }
+    }
+    None
 }
 
 fn platform_guard_from_stmt(stmt: &Stmt) -> Option<PythonPlatform> {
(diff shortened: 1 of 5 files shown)


yangdanny97:
There are panics from this change, you can see it in the mypy primer CI job

Example

> ERROR Thread panicked, shutting down: panicked at pyrefly/lib/state/state.rs:2270:17:
LookupAnswer::get failed to find key, builtins KeyExport(Name("WindowsError")) (concurrent changes?)

yangdanny97:
The primer delta looks unfavorable, so I think this needs some work still. LMK if you want me to help dig into it

asukaminato0721:
help would be great.

NathanTempest (changes requested):
Thanks @asukaminato0721 for the persistence on this, the latest mypy_primer is cleaner. The RwLock-based effective_sys_info cache addressed the cascading inference failures from the previous round.

Tagging @stroxler as the work has moved into type-inference infrastructure territory.

Two outstanding things that are worth considering:

- CI is currently red on all 3 platforms — needs to be green before landing.

- module_platform_guard scans the entire module body (Copilot flagged this earlier):

`
body.iter().find_map(platform_guard_from_stmt)
`

A guard nested deep inside the file (e.g., inside an if __name__ == "__main__": block) would still trigger the override. Restricting to leading top-level statements (allowing optional docstring + imports first) would be safer.

- get_stdlib_for_sys_info shortcut:
```
if self.data.stdlib.len() == 1 {
    return self.data.stdlib.first().unwrap().1.dupe();
}
```
Returns the only cached stdlib regardless of the requested sys_info. With platform-guarded modules creating Handles for non-default platforms, this can silently return a stdlib for the wrong platform. Should at least check the cached entry's SysInfo matches before taking the shortcut.

kinto0 (changes requested):

stroxler:
Heads-up on this @asukaminato0721 - this is actually a pretty difficult feature to do properly. It has implications for both potentially deadlocking and potential incremental update viability.

Your PR was a start, but I'm working on a stack of plumbing commits around it to deal with these issues. I'm not yet sure whether we'll wind up merging it, we need to think through whether the feature is worth the complexity. But don't feel any need to update or rebase - I'm effectively taking over this work for the time being.

Thanks! It took a long time to get to this precisely because we knew that it will be pretty difficult.

PR #2584, opened 2026-02-27
Splitting unknown-name error code

### LINKED ISSUE #2550: Add a way to suppress `unknown-name` diagnostics for `reveal_type` usages
Please consider this a very-low-priority request, and feel free to close it.

When checking a file with pyrefly, `reveal_type` already works fine without importing it from `typing`/`typing_extensions`, but an additional `unknown-name` diagnostic is emitted which makes it hard(er) to find the revealed type:
```
▶ cat test.py
reveal_type(1 + 1)

▶ pyrefly check test.py 
ERROR `reveal_type` must be imported from `typing` for runtime usage [unknown-name]
 --> test.py:1:1
  |
1 | reveal_type(1 + 1)
  | ^^^^^^^^^^^
  |
 INFO revealed type: int [reveal-type]
 --> test.py:1:12
  |
1 | reveal_type(1 + 1)
  |            -------
  |
 INFO 1 error
```

Other type checkers (mypy, pyrefly, ty) all suppress their version of the `unknown-name` diagnostic for `reveal_type` usages, or give users a way to suppress them in the config (without silencing *all* `unknown-name` diagnostics). This is a nice quality-of-life improvement when working with `reveal_type` as a debugging tool, especially if you have multiple `reveal_type`'s in a single file.

Karman-singh15:
# Summary

This PR adds two new error kinds, `UnimportedAssertType` and `UnimportedRevealType`, to make error reporting more specific when `assert_type` or `reveal_type` are used without being imported. Instead of falling back to the generic `unknown-name` error, the binding logic now emits these dedicated error types. Both were added to the `ErrorKind` enum with a default severity of `Error`. I also updated the documentation to describe the new error kinds and adjusted the relevant test expectations to match the new diagnostics, while keeping the existing Rust unit tests unchanged to ensure everything still aligns with the intended error behavior.

Fixes #2550

# Test Plan

I verified that `cargo test test_assert_type_variations` and `cargo test test_reveal_type_variations` both run and pass after the changes. I also made sure the new configuration values are properly sorted and parsed as expected by the existing framework. With this update, users can now suppress these specific diagnostics using `# pyrefly: ignore[unimported-reveal-type]` (or the corresponding assert type rule) without having to ignore all `unknown-name` errors, which keeps suppression more precise and intentional.

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -256,10 +256,8 @@ pub enum ErrorKind {
     UnexpectedKeyword,
     /// An error caused by passing a positional argument for a keyword-only parameter.
     UnexpectedPositionalArgument,
-    /// Attempting to use `assert_type` without importing it from `typing`.
-    UnimportedAssertType,
-    /// Attempting to use `reveal_type` without importing it from `typing`.
-    UnimportedRevealType,
+    /// Attempting to use a type checker directive without importing it from `typing`.
+    UnimportedDirective,
     /// Attempting to use a name that is not defined.
     UnknownName,
     /// Identity comparison (`is` or `is not`) between types that are provably disjoint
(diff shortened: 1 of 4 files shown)


Karman-singh15:
can you review the changes

PR #2587, opened 2026-02-28
[Pyrefly][Init] Add non-interactive mode to pyrefly init

migeed-z:
Currently, mypy primer does not run with any configs, which is wrong. We should invoke pyrefly init and generate a config file for every project. To do this for CI purposes, workflow purposes and experiment purposes, we should also have a non interactive mode. This diff adds this mode, which declines suppressing errors and overriding configs (for CI there are no existing pyrefly configs anyway).


yangdanny97 (approved):

rchen152:
Merged in a791138d087e3a2f5599f5f7ffa2b573c05262e9.

PR #2589, opened 2026-02-28
fix [Docs+feature] Consider how Var should be represented in reveal_type and hover #280

### LINKED ISSUE #280: [Docs+feature] Consider how Var should be represented in `reveal_type` and hover
### Describe the Bug

We already show `@_` on hover if a type is never pinned; I'm in the process of changing `reveal_type` to do the same if it is not *yet* pinned when the reveal_type is analyzed.

We need documentation for what this means, and we might want to also change the printable representation to something more informative (and more searchable in our docs page)

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #280

Added a non‑pinning usage mode in bindings and wired reveal_type calls to use it, so calling reveal_type(x) doesn’t count as a “first use” that pins partial types. This preserves the intended behavior where later real usage (like x.append(1)) still pins the partial.

Updated special‑export handling to recognize `typing.reveal_type/typing_extensions.reveal_type`, enabling special handling in the binder.

Implemented reveal_type display overrides for empty list/dict assignments so that, when the variable is still unpinned, the printed type shows `@_` rather than Unknown. This is display‑only and doesn’t mutate solver state.

# Test Plan

Added a regression test that matches the docs example (initial reveal shows `list[@_]`, subsequent reveal after append(1) shows list[int]) and updated the docs to explain `@_`.


stroxler:
@asukaminato0721 could you rebase this? I've made some changes to partial type inference that might have already been in when you did this, but now I've also implemented a fixpoint for strongly-connected components.

I think we can probably support a non-usage mode for reveal_type, but it would be nice to know the PR is up to date with the latest changes before I consider it too much - I actually think for the most part it should now be very difficult to get a raw var in the display, because I'm isolating the Var much more aggressively than before

It's possible that this isn't worth the rebase, we might want to just close it - I think the original issue is mostly moot at this point because of recent changes

asukaminato0721:
@stroxler done

stroxler:
Great. I'll try to take a look this week, although I have a hackathon so it might take me some time to get to it. I'm a little surprised it works at all to print Vars downstream, given how partial type inference is now handled inline, so I'll have to take a close look and make sure nothing here is a determinism risk.

I would have naively expected to see the pinned value of the Var in the reveal_type the way things are set up, and I'm having trouble understanding why that's not what's happening here. Regardless, thanks for doing this - I expect I'll learn something I didn't know about how Pyrefly works from the review

stroxler:
Thanks for the work! After some consideration, I think this feature is too complicated to be worth it.

The biggest issue is that it's not easy to avoid skipping partial type pins inside a `reveal_type` that actually does something, like
```
x = []
reveal_type(x.append(1))`
```
(silly example, but the point is that it's very tricky to avoid pinning).

We could probably make it work, but I think it's probably not worth the complexity - we've already got pretty fast complexity growth just shipping the most important fixes

PR #2594, opened 2026-02-28
[conformance][RFC] Error on `Final` variables declared without initialization

fangyi-zhou:
## Summary

This diff implements a new feature to error on `Final` declarations with initialization.

https://typing.python.org/en/latest/spec/qualifiers.html#semantics-and-examples

> - There can be at most one final declaration per module or class for a given attribute. There can’t be separate class-level and instance-level constants with the same name.
> - There must be exactly one assignment to a final name.

This implementation adds tracking for initialization for `Final` variables and raise errors if no initialization if founds. The initialisation checks covers different ways of initialization (i.e. direct assignment, `with` operator, and walrus operator). Special consideration is taken for class attributes, where initializations within `__init__` methods are also taken into consideration.

## Test plan

- Add new tests about the feature
- Ran conformance testsuites and observed 2 fewer failing cases


stroxler (approved):

PR #2596, opened 2026-03-01
Fixed notebook quick-fix position

### LINKED ISSUE #2152: Notebooks: insert import quick-fix should insert the import in the current cell, not the top cell
### Describe the Bug

1. verify that this is indeed what Pylance and other language servers do
2. implement the change by remapping quick-fix positions based on notebook cell offsets

### Sandbox Link

_No response_

_No response_

hanzel-sc:
### Summary
This PR fixes the issue where notebook import quick-fixes were incorrectly inserted into the first cell instead of the current active cell. Previously, Pyrefly's auto-import logic calculated a position relative to the entire combined notebook AST. This meant that the best position for an import was always at the very top of the combined source, which corresponds to the first cell.

The fix modifies the code_action handler in server.rs to detect when a quick-fix is triggered from a notebook cell and remaps the insertion position to the top of the currently active cell (line 0, character 0). This matches the behavior of Pylance and other major language servers.

Fixes #2152

### Test Plan

Updated  pyrefly/pyrefly/lib/test/lsp/lsp_interaction/notebook_code_action.rs to include explicit test cases for multiple cells (cell 1, cell 2, and cell 3). Verified that triggering an import in any of these cells correctly targets the respective cell.

Ran: cargo test test_notebook_code_action_import
Result: test result: ok. 1 passed; 0 failed

Manual Verification:
- Built the language server: cargo build --package pyrefly
- Configured VS Code to use the local build.
- Verified that triggering "Insert import" in a later cell (e.g cell 3) correctly inserts the import at the top of cell 3 and not in cell 1.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -3714,24 +3714,83 @@ impl Server {
         let mut actions = Vec::new();
         let server_state = self.telemetry_state();
         let start = Instant::now();
+        // If the code action is triggered from a notebook cell, we need the cell's
+        // index so that import quick-fixes can be redirected to the current cell
+        // instead of always targeting cell 1 (position 0 of the combined AST).
+        let triggered_cell_index = self.maybe_get_cell_index(uri);
         if allow_quickfix
             && let Some(quickfixes) =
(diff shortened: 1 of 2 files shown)


hanzel-sc:
@yangdanny97 
Please let me know if this is satisfactory or if any changes have to be made

Issue #2549, opened 2026-02-25
LSP highlighting not working for variables defined in list comprehension


andy-lz:
### Describe the Bug

```python
[i for i in [1,2,3]]
```

Python: 3.12

Variables defined inside a list comprehension are not highlighted or linked to the variable definition, so goto doesn’t work either. Verified in VSCode and sandbox.

![Image](https://github.com/user-attachments/assets/64d1a3f4-ece8-400d-ac02-9606b6e46529)

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIA2hAARi4BONtE6NlAjADQBMXAzAF1BILiACuAF2hwS5RCADENAKrSoESaTrj0AY2m50cADrozmGGDqMAtqkkB9dONvYYDABT5EzdJIBKGgBaAD4aOEkGRDMaOJoGGElxBjYwExAAOVd3aJpgfABfDLNREDJEsChSQklcWygKZQAFUkrqiIwcAho9I0gAcxSHCCNCM2UAZRgYGgALSUliOEQAelWKq2rCRgHVmHRVzFw9OFW%2B9EHhw0ObJlQAN1RoVGxYXv6IIYYRoxpcYg3WRmMiSOZGYIPDxwUZsAC8NAy-EIHB4JXQIEKYlQBggUIAYtAYBQ0Fg8EQyJigA

pswitchy:
I'd like to work on this, could you assign this to me?

yangdanny97:
assigned, let us know if you have any questions!

pswitchy:
@yangdanny97 could you review the pr #2597

PR #2597, opened 2026-03-01
Fix "Go to Definition" for variables used in list comprehensions(LSP highlighting not working for variables defined in list comprehension #2549)

pswitchy:
**Motivation:**
Currently, when you are working with a list comprehension like `[x for x in items]`, the "Go to Definition" feature works for the variable after the [for](cci:1://file:///e:/pyrefly/pyrefly/lib/binding/bindings.rs:275:4-299:5) (the target `x`), but it fails and returns nothing if you try it on the variable before the [for](cci:1://file:///e:/pyrefly/pyrefly/lib/binding/bindings.rs:275:4-299:5) (the usage `x`). 

This PR fixes that bug so that clicking on the usage `x` will correctly take you to where it is defined.

**Approach:**
The bug happens because of how the Pyrefly language server tries to follow the chain of variable references.
- For variables inside comprehensions, the chain goes through a special step called `IterableValue`.
- The code that searches for definitions ([find_definition_key_from](cci:1://file:///e:/pyrefly/pyrefly/lib/state/ide.rs:54:0-122:1) in [pyrefly/lib/state/ide.rs](cci:7://file:///e:/pyrefly/pyrefly/lib/state/ide.rs:0:0-0:0)) did not know how to handle this `IterableValue` step, so it would give up and return nothing.
- I updated the code to recognize `IterableValue` as a valid stopping point. Now, when the search hits this point, it correctly returns the location of the variable.

**Tests:**
- I added two tests in [pyrefly/lib/test/lsp/definition.rs](cci:7://file:///e:/pyrefly/pyrefly/lib/test/lsp/definition.rs:0:0-0:0) to make sure "Go to Definition" works for both parts of a list comprehension and to prevent this bug from coming back.
- I manually analyzed code bases and tested all the changes to make sure everything works correctly.

**Tests Screenshots**
Screenshot 2026-03-01 184413
Screenshot 2026-03-01 184357

### DIFF
--- a/pyrefly/lib/test/lsp/definition.rs
+++ b/pyrefly/lib/test/lsp/definition.rs
@@ -2083,3 +2083,31 @@ y = f"hello {x:{f()}}"
         "Expected definition to jump to function f, got: {report}"
     );
 }
+
+#[test]
+fn goto_def_list_comprehension_target() {
+    let code = r#"
+items = [1, 2, 3]
+result = [x for x in items]
+#               ^
+"#;
+    let report = get_batched_lsp_operations_report(&[("main", code)], get_test_report);
+    assert!(
+        !report.contains("Definition Result: None"),
+        "Expected goto-def on list comprehension target variable to find a definition, got: {report}"
+    );
+}
+
+#[test]
+fn goto_def_list_comprehension_usage() {
+    let code = r#"
+items = [1, 2, 3]
+result = [x for x in items]
+#         ^
+"#;
+    let report = get_batched_lsp_operations_report(&[("main", code)], get_test_report);
+    assert!(
+        !report.contains("Definition Result: None"),
+        "Expected goto-def on list comprehension usage variable to find a definition, got: {report}"
+    );
+}


yangdanny97:
I've confirmed manually that semantic highlighting works (missing a test)

Go-to-def is fishy, when we have `[x for x in y]` i would expect clicking the first x to jump to the 2nd x, but it jumps to the y instead. It's hard to tell in the test since it only asserts that it jumps somewhere

yangdanny97:
I've made a fairly substantial patch on top of this PR which fixes the issue and added more extensive tests, it ended up requiring some more changes in the binding layer.

Hopefully we can merge it soon, no action needed from you

pswitchy:
> I've made a fairly substantial patch on top of this PR which fixes the issue and added more extensive tests, it ended up requiring some more changes in the binding layer.
> 
> Hopefully we can merge it soon, no action needed from you

Thank you very much @yangdanny97, hoping to contribute to more issues 🚀

PR #2598, opened 2026-03-01
fix Python interpreter gets error, using [int] with a function call while pyrefly doesn't get error #2590

### LINKED ISSUE #2590: Python interpreter gets error, using `[int]` with a function call while pyrefly doesn't get error
### Describe the Bug

*Notes:
- pyrefly check test.py
- pyrefly 0.46.1
- Python 3.14.0

Python interpreter gets the error, using `[int]` with a function call while pyrefly doesn't get error as shown below:

```python
def func[T](x: T) -> T:
    return x
               # python # pyrefly
func[int](100) # Error  # No error
```

> TypeError: 'function' object is not subscriptable

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2590

blocked subscripting generic functions/callables so func[int] now emits UnsupportedOperation, and added a regression test.

# Test Plan


stroxler (approved):

PR #2626, opened 2026-03-03
Fix false positive `missing-attribute` on overridden class variables

grievejia:
Summary:
When a subclass overrides an unannotated class variable with a compatible narrower type, pyrefly was widening the child's type to the parent's type. This caused false positives when the child's narrower type has methods not present on the parent's type. For example, if a parent class has `request_class = Request` and a child overrides it with `request_class = FormRequest`, accessing `FormRequest`-specific methods like `from_response()` would produce a spurious `missing-attribute` error because pyrefly stored the type as `type[Request]` instead of `type[FormRequest]`.

The root cause was in `analyze_class_field_value`: when both parent and child lack annotations and the child's value is compatible with the parent's type, the code discarded the child's inferred type and returned the parent's type. This widening was intended to prevent invariant override errors for read-write attributes, but it loses type precision.

The fix changes the compatible branch to use the child's inferred type and sets `is_inherited = IsInherited::No` to skip the now-redundant override check (compatibility has already been validated by the `errors2.is_empty()` check). The `IsInherited` enum is documented as "only used for efficiency" — setting it to `No` just tells the override checker to skip this field, which is safe.

Differential Revision: D94992218

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2217,7 +2217,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // Otherwise, analyze the value to determine the type
         let (inherited_ty, inherited_annotation) =
             self.get_inherited_type_and_annotation(class, name);
-        let is_inherited = if inherited_ty.is_none() {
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

PR #2630, opened 2026-03-03
fix propagate narrowed types across functions #1201

### LINKED ISSUE #1201: propagate narrowed types across functions
```py
def f(x: object):
    assert isinstance(x, int)

def main(x: object):
    f(x)
    x  # object
```

here is would expect `x` to be `int`, if `f` were to be inlined then `x` would be narrowed, so i see no reason why it shouldn't be

asukaminato0721:
# Summary

Fixes #1201

Implemented assertion-only helper narrowing across call sites, so a def f(x): assert isinstance(x, int) now narrows the caller’s argument after f(x) when used as a statement. This is recorded during binding and applied in Stmt::Expr calls.

# Test Plan

Added a regression test that matches the issue example.

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -31,6 +31,7 @@ use ruff_python_ast::ExprAttribute;
 use ruff_python_ast::Identifier;
 use ruff_python_ast::ModModule;
 use ruff_python_ast::Parameter;
+use ruff_python_ast::Parameters;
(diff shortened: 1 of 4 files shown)


yangdanny97:
it's odd that there's no delta from mypy primer

Per @samwgoldman's comment on the original issue, that feature probably needs to be standardized in the typing spec first before we add it. This PR might be useful as a prototype implementation if/when someone decides to author a PEP, but I don't think we can merge it as-is

PR #2640, opened 2026-03-04
fix part of Exhaustiveness check in match statement #1286

asukaminato0721:
# Summary

Fixes part of #1286

Aligned match-case negative narrowing with Pyright’s control-flow behavior by allowing IsNotInstance from patterns to narrow even for non-final classes, which makes _ branches unreachable for exhaustive class unions.

Added a regression test that mirrors the issue scenario with user-defined classes and assert_never in the wildcard branch.

# Test Plan


asukaminato0721:
```
cargo test -p pyrefly test_match_exhaustive_enum_assign --lib
```

stroxler (approved):

PR #2647, opened 2026-03-04
Seed flow entries for implicit captures at function scope creation

yangdanny97:
Summary:
# The stack:

D91053724 was intended to prevent variables captured from outer scopes getting flagged as uninitialized when they are narrowed in a nested scope.

It was reverted as part of a SEV, this stack is a second attempt at those changes. The detailed plan is described in D95121249 (phase 1)

# This Diff:

Creates an initialized flow info for captured variables when we create a function scope. This fixes the root cause of the issue described in D91053724

Differential Revision: D95121246


stroxler (approved):

PR #2649, opened 2026-03-04
Add support for the TSP typeServer/getPythonSearchPaths message 

rchiodo:
# Summary

Adds support for the next TSP message  = typeServer/getPythonSearchPaths. This message is used to convey basically PYTHONPATH to the other side. In our case, Pylance uses this to find implementations for stubs.

# Test Plan

Added a number of unit tests and ran tests.py. Also refactored some of the existing TSP test code to use common functions.

/cc @kinto0


kinto0:
test failures seem unrelated. importing + rerunning CI

Issue #2419, opened 2026-02-14
Operator < fails on identical constrained TypeVar operands


QEDady:
### Describe the Bug

It seems Pyrefly doesn't handle constrained TypeVar well in the following:

```
from typing import *
N = TypeVar('N', int, float)
class A(Generic[N]):
    a: N
    def foo(self, b: N):
        return self.a < b
```

It fails with 
```
`<` is not supported between `N` and `N` [[unsupported-operation](https://pyrefly.org/en/docs/error-kinds/#unsupported-operation)]
  Argument `N` is not assignable to parameter `value` with type `int` in function `int.__lt__`
```
However, all other type checkers are ok with the above.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroAcnQC8dACrMYANVSUAFAHJRqgDSd0DbWCi5UDAJQiAxlFRw4dAILKA4jHQxKEcwG1RAXROIROiC6VEQ6UUDgzBgwOjBcXGU4GCgwbWww0X9I4ODKGAYAV0p0OmTUwlQ6AB46bBEQTRAyfINSQgZaKAoAYjoABVJWqFIyjBwCOnNcdEg2YuMIGcIRPoBlGBg6AAsGBmI4RAB6I5aYkcJeNiOXI8xcczgj6dmIecpFmaO43hCAN1Q0FQ2FgUxmcwWDCWpVwxChMzgK3QZAY2xmAFo-m44NDJHQhCAAMyEACMACYCQ0AL5NVDmKFYgBi0BgFDQWDwRDIICpQA

kinto0:
[...]

PR #2660, opened 2026-03-05
Promote inferred type to matching constraint for constrained TypeVars

yangdanny97:
Summary:
I'm hoping we could make some progress towards addressing https://github.com/facebook/pyrefly/issues/2419

The typing spec requires that a constrained TypeVar (`T = TypeVar("T", int, str)`) resolves to exactly one of its constraint types — never a subtype like `bool` or `Literal[42]`. For example, calling `f(True)` where `f[T: (int, str)](x: T) -> T` should yield `int`, not `bool`.

Previously, the solver pinned the TypeVar to the raw argument type and validated it against the union of constraints. This meant subtypes like `bool` or user-defined subclasses like `MyStr(str)` leaked through as the resolved type, causing `assert_type` failures and spurious errors when mixing a subclass with its base in the same call (e.g., `concat(m: MyStr, s: str)` would error because `str` is not assignable to `MyStr`).

The fix introduces constraint promotion in the TypeVar solving arms of the solver. When we encounter a constrained TypeVar, instead of recording the raw argument type directly, we search for the narrowest constraint the argument is assignable to and pin the TypeVar to that constraint. "Narrowest" means: among all matching constraints, pick the one that is itself a subtype of all others (e.g., for a class D inheriting from both B and C, prefer C over B if both are constraints). If no constraint matches, we report the existing specialization error.

This also happens to resolve conformance failures in `generics_basic.py` (3 differences → 2; the remaining one is an unrelated `AnyStr` issue).

Differential Revision: D95010963

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -5414,6 +5414,17 @@
     },
     {
       "code": -2,
+      "column": 15,
+      "concise_description": "Argument `bytes` is not assignable to parameter `y` with type `str` in function `concat`",
+      "description": "Argument `bytes` is not assignable to parameter `y` with type `str` in function `concat`",
+      "line": 44,
+      "name": "bad-argument-type",
+      "severity": "error",
+      "stop_column": 16,
+      "stop_line": 44
+    },
+    {
+      "code": -2,
       "column": 18,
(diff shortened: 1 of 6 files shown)


stroxler (approved):

PR #2661, opened 2026-03-05
impl Rename Refactors #2207

asukaminato0721:
# Summary

Fixes part of #2207

Implemented rename refactoring options for comments/strings and text occurrences, aligned with PyCharm’s inline rename behavior and availability cues.

Added LSP linkedEditingRange support for inline rename ranges in-file, and wired the new rename options through workspace config.

# Test Plan

Added LSP interaction tests and fixtures covering comment/string and text-file occurrences, plus linked editing ranges.


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -26,6 +26,7 @@ use crossbeam_channel::Receiver;
 use crossbeam_channel::Sender;
 use dupe::Dupe;
 use dupe::OptionDupedExt;
+use ignore::WalkBuilder;
```
jvansch1: Could we do this logic without adding any new libraries?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4359,24 +4396,153 @@ impl Server {
         let Some(handle) = self.make_handle_if_enabled(uri, Some(Rename::METHOD)) else {
             return self.send_response(new_response::<Option<WorkspaceEdit>>(request_id, Ok(None)));
         };
-        self.async_find_references_helper(
```
jvansch1, line 5350: Why are we getting rid of this? I would think we would still want to use this in cases where we are renaming a symbol?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4359,24 +4396,153 @@ impl Server {
...
-                let mut changes = HashMap::new();
-                for (uri, ranges) in results {
-                    changes.insert(
-                        uri,
-                        ranges.into_map(|range| TextEdit {
-                            range,
-                            new_text: params.new_name.clone(),
-                        }),
-                    );
+            FindPreference {
```
jvansch1: We added a lot of new logic in line here. I think we should either break this up into some smaller functions or try to make it more clear what is happening here. Seeing a large code block like this is hard to follow.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4385,6 +4551,128 @@ impl Server {
         );
     }
 
+    fn text_occurrence_edits_in_workspace(
```
jvansch1: Helper functions like this should probably not go in server.rs. This file is already getting way too large and we should try to find a better place for them.

pyrefly/lib/state/lsp.rs
```diff
@@ -3523,6 +3533,184 @@ impl<'a> CancellableTransaction<'a> {
     }
 }
 
+fn is_identifier_char(ch: char) -> bool {
```
jvansch1: These types of helper functions should not be added to this file. We want to prevent this file from becoming any larger than it needs to be.

pyrefly/lib/state/lsp.rs
```diff
@@ -3523,6 +3533,184 @@ impl<'a> CancellableTransaction<'a> {
...
+            return false;
+        }
+    }
+
+    bytes[start..quote_index]
+        .iter()
+        .any(|b| matches!(b, b'r' | b'R'))
+}
+
+fn comment_and_string_content_ranges(source: &str) -> Vec<TextRange> {
```
jvansch1: We are writing a lot of custom string parsing logic here which is a little concerning. Can we use something like ruff_python_parser to help us determine what strings we need to collect here?

pyrefly/lib/state/lsp.rs
jvansch1: Also it looks like there are no tests for any of these functions. I think if we are going to be adding all of these we need to make sure they are tested.

jvansch1 (changes requested):

pyrefly/lib/lsp/non_wasm/server.rs
asukaminato0721: now becomes async_find_from_definition_helper

kinto0 (changes requested):

asukaminato0721:
cc @rsheftel

jvansch1:
@asukaminato0721 Can you add some videos of how to use this feature or some step by step instructions?

asukaminato0721:
with

```json
    "python.pyrefly.rename": {
        "commentsAndStrings": true,
        "textOccurrences": true
    },
```

in config

[Screencast_20260421_160927.webm](https://github.com/user-attachments/assets/1f0a9aa2-2f95-46b4-ae21-c2a526c2db99)

jvansch1:
@asukaminato0721 I am having a lot of trouble getting this working. When I try to trigger a rename with CommentsAndStrings as well as textOccurrences set to try it does not update comments and strings like intended. Is there any other missing piece I need to consider to get this working?


https://github.com/user-attachments/assets/4168b16f-3c85-4ddb-8a36-298469378fdf

asukaminato0721:
I just put this at `~/.config/Code - Insiders/User/settings.json`

```json
{
    ...
    "python.pyrefly.rename": {
        "commentsAndStrings": true,
        "textOccurrences": true
    }
}
```

and cd lsp/ use F5 to trigger extension develop mode.

kinto0:
looks like some tests are red

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1428,6 +1440,11 @@ struct TypeHierarchyTarget {
     is_object: bool,
 }
 
+struct RenameSearchResults {
```
kinto0: nit: should this be in the rename file?

pyrefly/lib/lsp/non_wasm/server.rs
kinto0: looks like this isn't entirely resolved

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1328,6 +1334,12 @@ pub fn capabilities(
                 }))
             }
         },
+        linked_editing_range_provider: match indexing_mode {
```
kinto0: why do we need this capability for this feature? does this not appear as a refactor?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -5068,6 +5193,45 @@ impl Server {
...
+    fn linked_editing_range(
+        &self,
+        transaction: &Transaction<'_>,
+        params: LinkedEditingRangeParams,
+    ) -> Option<LinkedEditingRanges> {
+        let uri = &params.text_document_position_params.text_document.uri;
+        let handle = self
+            .make_handle_if_enabled(uri, Some(Rename::METHOD))
+            .ok()?;
+        if self.open_notebook_cells.read().contains_key(uri) {
```
kinto0: why doesn't this work in notebooks?

kinto0 (changes requested):

pyrefly/lib/lsp/rename.rs
```diff
@@ -0,0 +1,197 @@
...
+    let mut matches = Vec::new();
+    let word_len = word.len();
+
+    for range in ranges {
+        let start = range.start().to_usize();
+        let end = range.end().to_usize().min(source.len());
+        if start >= end {
+            continue;
+        }
+        let slice = &source[start..end];
```
kinto0: 1. Unchecked byte slicing panics the server on half-typed source — pyrefly/lib/lsp/rename.rs:129

  let slice = &source[start..end];

  start/end come from ruff's content_range() (line 60) and the hand-rolled bytes_literal_content_range (line 91), both of which compute end - closer_len() with no char-boundary guarantee.

  Normally ruff is safe here: closer_len() returns 0 when is_unclosed() is set, and the lexer sets that flag during parse_unchecked recovery. But StringLiteral::invalid(range) / BytesLiteral::invalid(range) 
  discard the flags entirely (ruff_python_ast-0.0.6/src/nodes.rs:1721, :2140) — StringLiteralFlags::empty() means closer_len() == 1 regardless of the real source. The parser uses these during
  implicit-concatenation recovery (ruff_python_parser-0.0.6/src/parser/expression.rs:1551,1572,1575) while keeping the original token range.

  Since module.contents() is the live in-memory buffer, a user mid-keystroke on x = "a" b'𝐁 yields end inside a 4-byte character → byte index N is not a char boundary panic in the LSP.
  bytes_literal_content_range's saturating_sub guards underflow but not boundaries.

  Use source.get(start..end) or clamp both ends to char boundaries before slicing.

pyrefly/lib/lsp/non_wasm/rename.rs
```diff
@@ -0,0 +1,170 @@
...
+    workspace_root: Option<&Path>,
+    old_name: &str,
+    new_name: &str,
+) -> HashMap<Url, Vec<TextEdit>> {
+    let Some(root) = workspace_root else {
+        return HashMap::new();
+    };
+
+    let mut changes = HashMap::new();
+    for path in workspace_file_paths(root) {
```
kinto0, line 49: is it necessary to walk every file in the workspace? this is similar to the perf concern I have in another PR (reference counting). should we consider indexing these? why / why not?

pyrefly/lib/lsp/non_wasm/rename.rs
```diff
@@ -0,0 +1,170 @@
...
+pub(crate) fn append_comment_and_string_occurrences(
+    transaction: &CancellableTransaction<'_>,
+    workspace_root: Option<&Path>,
+    old_name: &str,
+    references: &mut Vec<(ModuleInfo, Vec<TextRange>)>,
+) {
+    let tx = transaction.as_ref();
+    let mut merged = std::mem::take(references);
+
+    for candidate in tx.handles() {
```
kinto0: this order is non-deterministic. it depends what the user has in the transaction. is this correct here?

pyrefly/lib/lsp/non_wasm/rename.rs
```diff
@@ -0,0 +1,170 @@
...
+                == Some(*ext)
+        }) {
+            continue;
+        }
+        if tx.is_third_party_module(&module_info, &candidate)
+            && !tx.is_source_file(&module_info, &candidate)
+        {
+            continue;
+        }
+        let comment_ranges = tx.text_occurrences_in_comments_and_strings(&module_info, old_name);
```
kinto0: do we need to reparse every module on every rename?

kinto0 (changes requested):
AI comments:

  - Was scoping considered for commentsAndStrings? PyCharm restricts comment/string search to the usage search scope; here any comment in any loaded module containing the bare word gets rewritten, including
  modules with their own unrelated foo. That's a lot of unexpected edits behind a single boolean.
  - linkedEditingRange is now advertised whenever indexing_mode != None, which changes behavior for all users who have editor.linkedEditing on — independent of the two new settings. Intentional?

pyrefly/lib/lsp/non_wasm/rename.rs
kinto0: bump. I wonder if we can benchmark this on large projects

asukaminato0721: added a new commit.

PR #2665, opened 2026-03-05
fix ParamSpecs should allow for overload resolution #2642

### LINKED ISSUE #2642: ParamSpecs should allow for overload resolution
### Describe the Bug

Reproducer:

```
import atexit
from typing import overload


@overload
def f(x: int) -> None: ...
@overload
def f(x: str, message: str) -> None: ...

def f(x, message=None): ...

atexit.register(f, "foo")
```

Output:

```
ERROR Argument `Literal['foo']` is not assignable to parameter `x` with type `int` in function `atexit.register` [bad-argument-type]
  --> /tmp/foo.py:12:20
   |
12 | atexit.register(f, "foo")
```

It looks like Pyrefly selects the overload too early, before it inspects the arguments passed to `atexit.register`.

I think the answer to #2355 is "yes" :)

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2642

Added ParamSpec-aware overload resolution when a `Callable[P, ...]` parameter is matched against an overloaded function with additional `*args/**kwargs`, so overload selection uses the call-site arguments before binding P. This fixes the premature overload selection in `atexit.register`.

# Test Plan

Added a regression test that passes a matching overload through atexit.register with ParamSpec args.

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -678,12 +678,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     if !ret.is_error()
                         && let Params::List(chosen_params) = &chosen_sig.params
                     {
-                        let tcc = &|| TypeCheckContext {
-                            kind: TypeCheckKind::CallArgument(
+                        let tcc = &|| {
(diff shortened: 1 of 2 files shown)


rchen152:
The linked issue has been fixed.

PR #2672, opened 2026-03-05
Fix false positive on attribute access for TypeVar bounded by `type`

### LINKED ISSUE #2614: Missing attribute false positive from TypeVar bounded with `type`
### Describe the Bug

Minimal repro:
```python
"""
When a TypeVar has a bound (e.g., bound=type), pyrefly resolves attribute
access on the TypeVar to `object` instead of the bound type. This causes
false positives when accessing attributes that exist on the bound type
but not on `object`.
"""
from typing import TypeVar

# TypeVar bound to `type` -- accessing `__name__`
# `type` has a `__name__` attribute, so cls.__name__ should be valid
# when T is bound to `type`. Pyrefly resolves T as `object` instead.
T = TypeVar("T", bound=type)
def get_class_name(cls: T) -> str:
    return cls.__name__  # FP: Object of class `object` has no attribute `__name__` 
```
No other type checker has this behavior. 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAOlTZegOoAWM6ABKqwCqnEwBqqAE6tGqOO1Z4ArllYAKGIQDmhADSTcMzAF4ALjxgBKdWUEwwUUqzNxcUAG7x2u3YIjYpumHVQBjX-DiuGy6zFwGAsK6uKwABrjYAFYwvrqxrBDocF6omKy4YKyhMBpaRQaEXIwQ4r6oUnDwdGCoUI2sxLhwELoQjuIA7sxsfgFw3ehKzq7unk6hqLqsMPg1S8FFYdKy%2Brx0HkvouOts8UkpaYR01DdU6GCCuAC25cSZUxBPnYJL3LyR13QAGJwv8hKUdjFYrsYOkALRw9j%2BQLvOIAfTR6FQTxgGNidBB0IM6VE4g4sQxWJxeOmbgOMHUtlYvjahEp2NxaNYcEYmigeWwJXsrQgmAJrCGLC4GXE2zy0TiMNilQACqQzBYrDY7P1pWI4glkql0plsjBcld0JxWNpQfwhHJqJxqOo5XoDIY6JhzKwlDBdGiWWI4JiOXIWXBEFxDKw4QA%2BbmuRB0Vip6z%2BqSCNgRtmh6lc1gggBiKqjAHlzql8oUg%2BMDZW0iJ9UdabMvOi85z0iBVCBPNA4CRyIgQCCAKq9KA9KxgGSpCDBOCA72FMC4QRPRaYqRPQWCOT4KOZXQx%2BOJwTJthpsy6TNsMDUAByO73UeA%2BAAvjd0D2QKZzJYhDRE8UAUCCaoapY3IYDgBDMsEkBKJmiwLugloggAyjAJSMC4xCRgA9AR-6aoQ65KARLAEZguC%2BHABG%2BAhEBIYIKHBARrBrsIqDCtAqDYLA8H3MxyG9BsuDEGJWSWmQoTBHCjiCN0Gy2tQADMhAAIwAEzfiAH69n4vSOEW0AwBQaBYHgRBkPpQA

_No response_

yangdanny97:
Summary:
When a TypeVar has `bound=type`, pyrefly internally represents the bound as
`type[Any]`, which resolves to `AttributeBase1::TypeAny` during attribute
lookup. The quantified TypeVar attribute resolution only handled
`AttributeBase1::ClassInstance`, causing any other variant (including `TypeAny`)
to fall back to `object`. This produced false positives like
"Object of class `object` has no attribute `__name__`" when accessing
attributes that exist on the `type` class.

The fix extracts the bound-to-ClassType mapping into a `quantified_bound_class`
helper that handles both `ClassInstance` (the common case) and `TypeAny`
(the `bound=type` case, mapping to `builtins.type`). This is applied to all
four sites where quantified bounds are resolved: both `T` and `type[T]` paths,
for both `Bound` and `Constraints` restrictions.

Fixes https://github.com/facebook/pyrefly/issues/2614

Differential Revision: D95419868

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -1870,6 +1870,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    /// Extract the ClassType to use for attribute lookup on a quantified (TypeVar)
+    /// type from the attribute base of its bound.
+    fn quantified_bound_class(&self, base: AttributeBase1) -> Option<ClassType> {
+        match base {
+            AttributeBase1::ClassInstance(cls) => Some(cls),
+            // Handle `type[Any]`, which happens for TypeVars w/ `bound=type`
+            AttributeBase1::TypeAny(_) => Some(self.stdlib.builtins_type().clone()),
+            _ => None,
+        }
+    }
+
     fn as_attribute_base(&self, ty: Type) -> Option<AttributeBase> {
         let mut acc = Vec::new();
         self.as_attribute_base1(ty, &mut acc);
@@ -1979,7 +1990,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     let mut use_fallback = false;
                     if let Some(base) = self.as_attribute_base(ty.clone()) {
                         for base1 in base.0 {
-                            if let AttributeBase1::ClassInstance(cls) = base1 {
+                            if let Some(cls) = self.quantified_bound_class(base1) {
                                 acc.push(AttributeBase1::ClassObject(ClassBase::Quantified(
                                     (*quantified).clone(),
                                     cls,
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #2676, opened 2026-03-05
Extend correlated constrained-TypeVar branching to method calls

grievejia:
Summary:
The previous commit added correlated constrained-TypeVar branching for binary operators via `try_binop_calls`. This extends it to method calls (e.g., `x.combine(y)` and dunder calls like `a.__lt__(b)`) by adding env-branching to `freeform_call_infer`.

When a method is called on a constrained TypeVar `T: (A, B)`, the attribute lookup produces a union of method types (one per constraint). `expr_call_infer` pre-infers the arguments before distributing over this union. The env-branching in `freeform_call_infer` detects that the pre-inferred argument types contain constrained quantifieds and tries each substitution environment, suppressing errors from environments where the argument doesn't match the method's parameter type.

To avoid the regression seen in the first attempt (where `f(D())` for `f[T: (B, C)](x: T) -> T` would incorrectly pre-solve the callee's own TypeVar), the branching only activates when constrained quantifieds appear in the argument types (which are `TypeOrExpr::Type` when pre-inferred). When the callee is not a union, args remain as `TypeOrExpr::Expr` and no branching occurs, so the callee's own TypeVars are solved normally through inference.

Two previously bug-marked tests are now fixed:
- `test_constrained_typevar_dunder_comparison`: `a.__lt__(b)` where `T: (int, float)` no longer errors
- `test_constrained_typevar_correlated_method_must_pass`: `x.combine(y)` where `T: (A, B)` no longer errors

Differential Revision: D95336216

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -17,6 +17,7 @@ use pyrefly_types::types::CalleeKind;
 use pyrefly_types::types::TArgs;
 use pyrefly_types::types::TParams;
 use pyrefly_types::types::Union;
+use pyrefly_util::owner::Owner;
 use pyrefly_util::prelude::SliceExt;
 use pyrefly_util::prelude::VecExt;
 use ruff_python_ast::Arguments;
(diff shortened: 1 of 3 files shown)


kinto0:
@grievejia is this still being worked on?

PR #2677, opened 2026-03-05
fix special-case NotImplementedType in binop signatures to match runtime semantics #1129

### LINKED ISSUE #1129: special-case `NotImplementedType` in binop signatures to match runtime semantics
### Describe the Bug

```py
from types import NotImplementedType

class A:
    def __add__(self, i: object) -> NotImplementedType: ...


class B:
    def __radd__(self, i: A) -> int: ...

_ = A() + B() # _NotImplementedType
```

binops can `Never` return `NotImplementedType` by cause of runtime semantics, i would expect this to result in `Never` or `int`

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1129

fixed the binop resolution path so NotImplementedType no longer leaks as a possible operator result.

successful dunder calls now strip only the exact NotImplementedType branch, keep searching reflected dunders when needed, and union any concrete results that can actually occur at runtime.

# Test Plan

a regression test covering both pure fallback and mixed `int | NotImplementedType` behavior.

### DIFF
--- a/crates/pyrefly_types/src/stdlib.rs
+++ b/crates/pyrefly_types/src/stdlib.rs
@@ -114,6 +114,8 @@ pub struct Stdlib {
     builtins_type: StdlibResult<ClassType>,
     /// Introduced in Python 3.10.
     ellipsis_type: Option<StdlibResult<ClassType>>,
+    /// Introduced in Python 3.10.
+    not_implemented_type: Option<StdlibResult<ClassType>>,
     /// Moved from `_typeshed` to `types` in 3.10.
     none_type: StdlibResult<ClassType>,
     function_type: StdlibResult<ClassType>,
@@ -248,6 +250,9 @@ impl Stdlib {
             ellipsis_type: version
                 .at_least(3, 10)
                 .then(|| lookup_concrete(types, "EllipsisType")),
+            not_implemented_type: version
+                .at_least(3, 10)
+                .then(|| lookup_concrete(types, "NotImplementedType")),
             none_type: lookup_concrete(none_location, "NoneType"),
             iterable: lookup_generic(typing, "Iterable", 1),
             async_iterable: lookup_generic(typing, "AsyncIterable", 1),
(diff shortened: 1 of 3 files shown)


NathanTempest (changes requested):
Thanks for continuing to push on this, @asukaminato0721  the core mechanism still needs a change before it can land, and I think it's the same point Danny raised earlier https://github.com/facebook/pyrefly/issues/1129#issuecomment-4962408336.

The issue: the current logic special-cases NotImplementedType on the return type by stripping NotImplementedType out of the forward dunder's result and, when the result is a partial T | NotImplementedType, keeps T and then re-invokes the reflected dunder with the whole original argument, unioning whatever comes back:
```
if ret_without_not_implemented != ret {
    successful_ret = self.union(successful_ret, ret_without_not_implemented);
    continue; // ← reflected dunder still gets the full arg
}
```
What's needed instead: split the argument union. For an argument A | B where __add__ returns a concrete type for A but NotImplemented for B, only B should be passed to __radd__ not the whole A | B.

Why it matters: the return-type approach is what's driving the mypy_primer regressions (latest run: +38 across pandas / pandas-stubs / xarray / freqtrade / werkzeug / core; earlier iterations peaked in the thousands on pandas). The classic case is DataFrame.__truediv__(int) -> DataFrame | NotImplementedType: stripping to DataFrame and then still calling int.__rtruediv__(DataFrame) yields float, so the result becomes DataFrame | float → cascading false unsupported-operation and missing-attribute errors. Notably pandas-stubs agrees with both mypy and pyright here, so these are genuine false positives, not stub quirks.

Test coverage gap: test_binop_not_implemented_type_uses_runtime_fallback only exercises single-type operands (A() + B(), C() + D()). It'd be worth adding a case with a union argument where only one member returns NotImplemented, since that's exactly the path that's currently mishandled and what the fix needs to get right.

Happy to help review once the per-union-member dispatch is in place.

NathanTempest (commented):
Thanks for the change in approach as suggested by @yangdanny97 

The C()+D() → int change did fix the partial-NotImplemented over-union, but the primer on dfeafa59 still shows major regressions, and pandas-stubs actually grew (~90 errors, up from ~33). The dominant remaining failure is a different path: when the forward dunder returns pure NotImplemented, we fall back to the reflected dunder and numpy's reflected dunders return Any, so ops the stubs assert as Never now resolve to Any (→ ~70 assert_type(Any, Never) failed, plus unused-call-result cascades). freqtrade/pandas/xarray/core also still regress on date/datetime arithmetic. 

I think the open design question is if an Any-returning reflected dunder be allowed to erase a Never? @yangdanny97 and @samwgoldman's appreciate your input and extra pair of eyes on this PR

PR #2680, opened 2026-03-05
fix Usage of a variable should initialise it #828

### LINKED ISSUE #828: Usage of a variable should initialise it
### Describe the Bug

```python
def f(b):
    x: int
    if b:
        x = 1
    print(x) # E: ERROR 7:11-12: `x` may be uninitialized
    print(x) # E: ERROR 8:11-12: `x` may be uninitialized
```

This gives two errors that `x` might not be initialised. But in the second case either we have crashed in the previous one, or it was initialised. Effectively, usage of a variable should be treated as an initialisation.

### Sandbox Link

https://pyrefly.org/sandbox/?code=FDAmFMDMAJICgEYEoBcxoegDxdAlgHYAu6meMCam120AvNAIykYAOAToUXFki9By48+QA

_No response_

asukaminato0721:
# Summary

Fixes #828

runtime name reads now update the current flow only when the name was Conditionally or DeferredCheck initialized. That means a successful read suppresses the duplicate follow-on "may be uninitialized" error on the surviving path, without changing definitely-uninitialized cases like exception-target cleanup.

# Test Plan

a test

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -1142,6 +1142,8 @@ impl<'a> BindingsBuilder<'a> {
     /// First-use detection happens later in `process_deferred_bound_names`
     /// when all phi nodes are populated.
     pub fn lookup_name(&mut self, name: Hashed<&Name>, usage: &mut Usage) -> NameLookupResult {
+        let may_prove_initialized =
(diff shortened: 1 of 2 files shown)


yangdanny97:
at a high level this sounds useful to have, there might be a few conflicts with ongoing work, we'll have to be careful merging this one in.

PR #2681, opened 2026-03-05
fix Overriding methods on NamedTuple does not work #934

### LINKED ISSUE #934: Overriding methods on NamedTuple does not work
### Describe the Bug

```python
from typing import *

class ReadWrites(NamedTuple):
    field: int

    def __iter__(self) -> Iterator[str]:
        ...

def f(rw: ReadWrites):
    for k in rw:
        reveal_type(k) # reveals int, but should reveal str
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoBUAUEQMYA2AhgM7VQBKAppQCYDqISMj1AFAHKUIjFgBUArgnKMAlAC4iUJVGBJG5FnMwoYJZVBaNgUAPomujEGd7V1wGVAC0APigBJbiEoxcAbWowIAC6Cvr6AHSRJIbGwLwgAO5aTKwcFtTyisrAuFAA1tpQiaFhSiCMAG7M5CbwCIy8eTJEQA&version=3.12

_No response_

asukaminato0721:
# Summary

Fixes #934

changing iteration on NamedTuple instances to honor an explicit `__iter__` override instead of always taking the tuple-element fast path.

overridden `__iter__` now gets called directly and its return type is unwrapped for for-loop element inference.

# Test Plan

 converted the existing regression into a passing test.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -731,6 +731,27 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    fn has_named_tuple_iter_override(&self, cls: &ClassType) -> bool {
+        if self
+            .get_metadata_for_class(cls.class_object())
+            .named_tuple_metadata()
+            .is_none()
+        {
+            return false;
+        }
+        let Some(iter_method) = self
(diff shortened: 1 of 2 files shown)


fangyi-zhou (approved):

PR #2690, opened 2026-03-06
impl Transform a function or a method to a method object. #2207

asukaminato0721:
# Summary

Fixes part of #2207

ref rope.refactor.method_object: Transform a function or a method to a method object.

https://rope.readthedocs.io/en/latest/library.html#list-of-refactorings

# Test Plan


asukaminato0721:

PR #2691, opened 2026-03-06
impl Implemented encapsulate_field #2207

asukaminato0721:
# Summary

Fixes part of #2207

Implemented encapsulate_field as a new local rewrite refactor. It now offers a code action on attribute selections, inserts unique `get_<field> / set_<field>` methods, rewrites reads to getter calls, rewrites assignments and += to setter calls, and rejects unsupported tuple-assignment targets. It’s wired into the transaction and LSP code-action pipeline.

ref https://rope.readthedocs.io/en/latest/library.html#list-of-refactorings

# Test Plan

Added focused tests for read/write rewriting, unique accessor naming, and tuple-target rejection.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -4728,6 +4728,10 @@ impl Server {
                 transaction.extract_field_code_actions(&handle, range)
             );
             timed_refactor_action!(
+                "encapsulate_field",
+                transaction.encapsulate_field_code_actions(&handle, range)
+            );
+            timed_refactor_action!(
                 "extract_variable",
                 transaction.extract_variable_code_actions(&handle, range)
             );
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -3400,6 +3400,14 @@ impl<'a> Transaction<'a> {
         quick_fixes::extract_field::extract_field_code_actions(self, handle, selection)
     }
 
+    pub fn encapsulate_field_code_actions(
+        &self,
+        handle: &Handle,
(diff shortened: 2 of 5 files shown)


asukaminato0721:

jvansch1:
@asukaminato0721 Looks like there are a lot of failing checks. Once these are fixed I can do a review.

jvansch1:
@asukaminato0721  As part of the description of this PR could you also add either a video showing this new functionality or instructions for how to test it?

asukaminato0721:
[Screencast_20260421_012325.webm](https://github.com/user-attachments/assets/d7b2326f-9b4f-4e63-abf7-cd33b72475c4)

rchen152:
Thanks for this change, and sorry for letting it sit for so long. After reviewing it, I think this would need substantial rewrites to address issues with the structure. Unfortunately, I don't have the bandwidth at the moment for an in-depth design review of a PR of this size, so I'm going to close it. Please feel free to open a new one; if you do, I'd recommend starting with a smaller scope and building on it incrementally.

The overall flow makes sense (semantic resolution -> transaction -> LSP code action), but the implementation builds independent source-text edits without statement-level planning. This leads to several correctness issues:
- The hand-written enclosing-statement walk descends only through functions and classes, so writes inside control-flow statements cause the entire action to disappear.
- Assignment RHS text is pasted into a call argument without the necessary grouping; for example, `obj.x = 1, 2` becomes `obj.set_x(1, 2)`.
- Each reference independently produces edits, allowing overlapping ranges.
- Converting `receiver().x = value()` directly into `receiver().set_x(value())` reverses Python’s receiver/value evaluation order.
- Ambiguous resolution results are silently reduced to the first matching definition.

PR #2692, opened 2026-03-06
impl change_signature #2207

asukaminato0721:
# Summary

Fixes part of #2207

Implemented change_signature as a new local refactor.

The current scope is the non-interactive subset that fits Pyrefly’s existing code-action model:

when the cursor is on a simple positional-or-keyword parameter, it now offers Remove parameter, Move parameter left, and Move parameter right, and rewrites matching call sites.

That includes bound and unbound method calls.

# Test Plan

add related test.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -4780,6 +4780,10 @@ impl Server {
                 transaction.introduce_parameter_code_actions(&handle, range)
             );
             timed_refactor_action!(
+                "change_signature",
+                transaction.change_signature_code_actions(&handle, range)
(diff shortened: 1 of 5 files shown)


stroxler (changes requested):

stroxler:
Thanks! This looks pretty good to me, but I think it needs a rebase - the copy pulled internally has build errors.

In addition, I think there may be a gap in test coverage. There are tests for remove, move-left, bound+unbound methods, and variadic rejection, but no explicit end-to-end test for the delta: 1 (move right) path. Consider adding one analogous to change_signature_move_parameter_left_updates_calls.

stroxler (approved):

PR #2694, opened 2026-03-06
impl introduce_factory #2207

asukaminato0721:
# Summary

Fixes part of #2207

The feature adds a static factory such as `Service.create(*args, **kwargs)` and rewrites constructor calls from `Service(...)` to `Service.create(...)`.

Matching now distinguishes actual constructors from calls such as `Service.some_classmethod(...)`; unsupported forms such as `Box[int]()` suppress the action to prevent an incomplete refactor.

Reference collection now uses reverse-dependency-based global references for workspace files.

Only uncovered in-memory modules are checked individually because unsaved editor buffers may not appear in the filesystem dependency graph.

This avoids querying local references for every loaded module while preserving edits in open files.

# Test Plan

Added coverage for same-file rewrites, cross-module rewrites, unique-name fallback, and unsupported-call rejection.

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/introduce_factory.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/introduce_factory.rs
@@ -20,6 +20,7 @@ use ruff_python_ast::StmtClassDef;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
 use ruff_text_size::TextSize;
+use vec1::Vec1;
 
 use super::extract_shared::line_end_position;
 use super::extract_shared::line_indent_and_start;
@@ -54,6 +55,8 @@ pub(crate) fn introduce_factory_code_actions(
             class_def.name.range().start(),
             FindPreference::default(),
         )
+        .map(Vec1::into_vec)
+        .unwrap_or_default()
         .into_iter()
         .find(|def| {
             def.module.path() == module_info.path()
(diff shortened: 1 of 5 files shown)


asukaminato0721:

pyrefly/lib/state/lsp/quick_fixes/introduce_factory.rs
```diff
@@ -0,0 +1,221 @@
...
+    transaction: &Transaction<'_>,
+    definition: &crate::state::lsp::FindDefinitionItemWithDocstring,
+    factory_name: &str,
+) -> Option<Vec<(pyrefly_python::module::Module, TextRange, String)>> {
+    let mut edits = Vec::new();
+    for module_handle in transaction.handles() {
+        let Some(module_info) = transaction.get_module_info(&module_handle) else {
+            continue;
+        };
+        let Some(refs) = transaction.local_references_from_definition(
```
kinto0, line 156: local references on _every_ module? do we need this?

kinto0 (changes requested):
i'd love some details on what this feature is in the summary + some kind of description as to why you chose to implement it this way

pyrefly/lib/state/lsp/quick_fixes/introduce_factory.rs
kinto0: still unclear. is this a performance issue?

PR #2701, opened 2026-03-07
Use the original source text for default values in hover text

### LINKED ISSUE #2697: Hover function with float default parameter shows `...`
### Describe the Bug

When a function has a float number as a parameter with default value, the float value isn't shown in hover, instead showing as `...`

For example, in this function:

```py
def f(y: int = 2, x: float = 0.0) -> None:
    pass
```

The hover text is:

```
(function) f: def f(
    y: int = 2,
    x: float = ...
) -> None: ...
```

Note that it works fine if the float literal is `0`, so the problem seems specifically about the literal value.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABSmI0ToAuNAvDQEwA0NfMzBRcqDtwAMhSQEoaAWgB8NAHK50MRAB10NfTWKo4cXbpABfPiFQBjNhABuMAGLQYFNFjxEyloA

oyarsa:
The previous approach only supported None and some literals, such as
strings, bools and ints. Values without literals like floats, tuples,
dicts, etc. were unsupported and replaced with `...` in the hover text.

This commit keeps the original behaviour for literals, but introduces
displaying the float value.

The changes required are basically just capturing the float value where
it's first encountered in `function::AnswersSolver::get_requiredness`
and formatting it in `callable::Param::fmt_default`.

Note that the number is rendered using Rust's built-in 'display', which
means a `0.0` literal gets rendered as `0`.

# Summary

Fixes #2697 

Example:

```py
def f(
    x: int = 1,
    y: str = "a",
    b: bool = True,
    f: float = 0.0,
    f2: float = 1.0,
    f3: float = 1.5,
    f4: float = -1,
    f5: float = -1.5,
):
    pass
```

The hover on `func` is:

```
```python
(function) f: def f(
    x: int = 1,
    y: str = 'a',
    b: bool = True,
    f: float = 0,
    f2: float = 1,
    f3: float = 1.5,
    f4: float = -1,
    f5: float = -1.5
) -> None: ...
```

# Test Plan

- `test.py` passes. It generated no changes to be commited. There are some compilation warnings, but I believe they are unrelated to the code in this PR.
- Tested as the LSP on neovim 0.11.6 with default lspconfig settings. I copied `target/release/pyrefly` to a directory in `PATH`.

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -35,6 +35,7 @@ use crate::class::Class;
 use crate::class::ClassType;
 use crate::display::TypeDisplayContext;
 use crate::equality::TypeEq;
+use crate::equality::TypeEqCtx;
 use crate::keywords::DataclassTransformMetadata;
 use crate::type_output::TypeOutput;
 use crate::types::Type;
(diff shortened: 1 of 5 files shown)


oyarsa:
New version that renders only floats. I still opted for converting it to String right at the first moment, which allows expanding it to more cases in the future.

yangdanny97:
I looped in some folks with more context on the IDE to review. at first glance, the main doubt i have is whether it's worth all the added complexity to support showing floats

oyarsa:
I'd be more than happy to change to a better approach if I can get some pointers. 

Supporting floats would be interesting for the code that I usually write. And pyright and pylsp both support it:

```
# pylsp                                                                                    
func(f: float = 0) -> None
───────────────────────────────────────── 
# pyright                                                              
(function) def func(f: float = 0) -> None
```

kinto0:
sorry about the delay - I was on PTO for a few days. I think the float hover is pretty worthwhile especially considering frequent hovering from a separate file. I like seeing default arguments. lemme import it and give it a look over

crates/pyrefly_types/src/callable.rs
```diff
@@ -729,10 +780,17 @@ impl Callable {
...
-            Some(Type::None) => "None".to_owned(),
+            Some(DefaultValue {
+                ty: Type::Literal(lit),
+                ..
+            }) => format!("{}", lit.value),
+            Some(DefaultValue { ty: Type::None, .. }) => "None".to_owned(),
+            Some(DefaultValue {
+                display: Some(text),
+                ..
+            }) => text.clone(),
```
kinto0, line 793: should this branch arm should go first? if a defaultvalue has an explicit display string, should it win against literals?

pyrefly/lib/alt/function.rs
```diff
@@ -80,6 +84,27 @@ use crate::types::types::Overload;
 use crate::types::types::OverloadType;
 use crate::types::types::Type;
 
+/// Extract a display string for default values whose types don't preserve the literal value.
+/// Float literals like `3.14` become `ClassType(float)` which loses the actual value.
+fn default_display_for_expr(expr: &Expr) -> Option<String> {
+    // 1. Unwrap the unary minus if it exists
```
kinto0, line 90: nit: i'm not sure we need these numbered steps. this first comment is useful but maybe we can simplify the comments?

pyrefly/lib/alt/function.rs
```diff
@@ -860,22 +892,28 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
         };
-        if let Required::Optional(Some(default)) = required {
+        if let Required::Optional(Some(default_val)) = required {
             // Mark literals as explicit so we don't promote them.
             // This has to happen after the param type has been computed because we do
             // want to promote literals while inferring the type.
-            required = Required::Optional(Some(default.explicit_literals()));
+            required = Required::Optional(Some(DefaultValue {
+                ty: default_val.ty.explicit_literals(),
+                display: default_val.display,
```
kinto0, line 915:
```suggestion
                ..default_val
```
nit: this makes additions to DefaultValue not require updating

kinto0 (approved):
LGTM - let me get another set of eyes on it on import

PR #2703, opened 2026-03-07
Support adjacent __new__.__defaults__ for functional namedtuples

### LINKED ISSUE #2611: Dynamically created `namedtuple` should not be too strict about `__init__`
### Describe the Bug

Minimal repro:
```python
from collections import namedtuple

# The standard pattern for making some namedtuple fields optional is
# to set __new__.__defaults__. Pyrefly ignores this and requires
# all fields to be passed.
Point = namedtuple("Point", ["x", "y", "z"])
Point.__new__.__defaults__ = (0,)  # make z optional
p = Point(1, 2)  # FP: pyrefly reports missing argument 'z' 
```
Going out of the way to figure out the exact structure and mutation in `Point.__new__.__defaults__ ` is probably too impractical. I think an easier solution here is to just mark all namedtuple elements as already having a default (which, from what I've tried, seems to be what other type checkers are doing -- changing the RHS of the `__default__` assignment does not really change the type checking results). 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEAxrlLAwC4S7px0Q3G5KbOulQ0YmNgFdisADroFAYjoAVABYw6cNhkypKmOsVRs2MSujphBdGqgDWEdAHNttLaPGSZsaxBgoTB5cYg4uVCheOGU6NlxtGGEAfWT0GAB3VMJUzBgwVCkoNjhsugAFUkp8qFJeF3RBeDj1CB49OmqARykIapj0FUioyEDguITsLRM4OAlCBXLcZ2EAXhExCWlZGAAKORAllYOAGjoAbQP8U7oD0huDgC8DgF0ASkXl9DYctMzs3L5QrFUrJOjrXYABhObzodBU9gcWkedFC4VEUAUxHBFS%2BbF2AEYzgAmWHwugAMXKiGMVRqdWqAiEPBobTgzjcBhcUnE3zoAHJHvy6CATiApBwoHASORECAVABVSUQNh1MBSdDsTjcBQKPJgayCexsNI8qaUXb4GkrWEAWgAfNo2JREAo4XDqtJLNYDgA5M0WGnAfAAXwOClFIDI1TAtUI8RoUAoKkqMdq2j0eHwjC4kG5lFM2oWgzoAGUYFp1GZiHBEAB6OvR%2BmEQQuOswdB1zC4BhwOtMdB5qQF9F1w2UOioABuqGgqGwfgHQ5H2tRYW10qxpDY6i4tqnFnZXBxBwAzIQCcTw%2BgQCGxagtQeKdAYBQ0FgszLb0A

_No response_

Adist319:
# Summary

Support the canonical `P.__new__.__defaults__ = (val,)` idiom adjacent to functional namedtuple definitions, making trailing fields optional in the constructor.

This is the standard pre-class-syntax way to add defaults to namedtuples:
```python
Point = namedtuple("Point", ["x", "y", "z"])
Point.__new__.__defaults__ = (0,)   # makes z optional
p = Point(1, 2)  # valid — z defaults to 0
```

Works for both `collections.namedtuple` and `typing.NamedTuple` functional forms. Detection uses forward peek in `stmts()` , only the immediately adjacent statement is checked, matching Pyright's behavior.

Things worth mentioning:
- Only tuple/None literal RHS is consumed; non-literal RHS (e.g., variable references) flows through normal type checking
- Right-aligned defaults following Python semantics (N defaults → last N fields optional)
- `__new__.__defaults__` replaces any existing `defaults=` kwarg, matching runtime behavior

Fixes #2611

# Test Plan

- Added 11 test cases covering: basic usage, typing.NamedTuple form, multiple defaults, None clearing, overflow clamping, non-adjacent (should not match), class-syntax unaffected, non-literal RHS fallthrough, replaces kwarg, empty tuple, None overriding kwarg
- All 52 named_tuple tests pass
- Full test suite (4379 tests) passes with no regressions
- Linting and conformance tests pass

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -257,6 +257,9 @@ pub struct BindingsBuilder<'a> {
     lambda_yield_keys: Vec<(TextRange, Box<[Idx<KeyYield>]>, Box<[Idx<KeyYieldFrom>]>)>,
     /// See `BindingsInner::subsequently_initialized`.
     subsequently_initialized: SmallSet<Idx<KeyAnnotation>>,
+    /// Temporary storage for adjacent __new__.__defaults__ expression detected by stmts().
+    /// Set before calling stmt() for a namedtuple assignment, drained by the NamedTuple arms.
+    pub(super) adjacent_namedtuple_defaults: Option<Expr>,
 }
 
 /// An enum tracking whether we are in a generator expression
@@ -512,6 +515,7 @@ impl Bindings {
             deferred_bound_names: Vec::new(),
             lambda_yield_keys: Vec::new(),
             subsequently_initialized: SmallSet::new(),
+            adjacent_namedtuple_defaults: None,
         };
         builder.init_static_scope(&x.body, true);
         if module_info.name() != ModuleName::builtins() {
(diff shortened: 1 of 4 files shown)


yangdanny97:
oh wow nice, I had no idea where to start with that issue

Issue #2607, opened 2026-03-02
Error converting Sequence to tuple


jakevdp:
### Describe the Bug

The following code fails to typecheck with pyrefly v0.54.0:
```python
from typing import Sequence

def f(x: Sequence[int]) -> tuple[int, ...]:
    if isinstance(x, tuple):
        return x
    return tuple(x)
```
```
$ uv run pyrefly check tmp.py 
ERROR Returned type `tuple[*Sequence[int]]` is not assignable to declared return type `tuple[int, ...]` [bad-return]
 --> tmp.py:5:16
  |
5 |         return x
  |                ^
  |
 INFO 1 error
```

### Sandbox Link

[sandbox link](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BlGAEcArjHQBjGAB10czDDB0wACnyJBI8VJgBtVgwC6ASjoBaAHyNRxWAfQMANHUJujiOXW%2BdlEOKxwDBjS6i4MtrAmnug%2BcXSUMBGUsfhePonJsRF2MOomciBOIKIM0HAk5IggAMR0AKplUBBMKqJSZbjocHIKSiq8NKgMAProojTYMJTqmoZmVnRBlDFxmaIpKjIgAHITUyt0wPgAvtuFxWSJYFCkhAy0UBR1AAqk17dLGDgEdJJdkDYG2GEC6hDkdSEMDoAAsGAxiHBEAB6ZFXJS3Qi8NjIiTIzC4SRwZH-dCA4GddDIgaUOioABuqGgqGwsD%2BAIgQMoIK6dFwxEpFTkZAYMK65np0wCvIAvHRtgBmQgARgATOd0CATsVUJIypKAGLQGAUNBYPBEMhaoA)

superbobry:
A similar reproducer

```
from collections.abc import Sequence

def f(x: Sequence[str]) -> Sequence[str]:
  if not isinstance(x, (list, tuple)):
    raise TypeError
  return list(x)
```

which outputs

```
ERROR No matching overload found for function `list.__init__` called with arguments: (list[str] | tuple[*Sequence[str]]) [no-matching-overload]
 --> /tmp/foo.py:6:14
  |
6 |   return list(x)
```

yangdanny97:
[...]

PR #2704, opened 2026-03-07
fix tuple to sequence narrowing (#2704)

yangdanny97:
Summary:

https://github.com/facebook/pyrefly/issues/2607

Differential Revision: D95483019


rchen152 (approved):

PR #2705, opened 2026-03-07
Fix round() narrowing overload

### LINKED ISSUE #2592: round() incorrectly assumes return type of float | int | None when given a float | int
### Describe the Bug

When running this:

```python
def test(value: int | float | None) -> str:
    if value is None:
        return ''

    value = round(value, 4)

    i = int(value)

    return str(i)
```

it incorrectly assumes that value could be None (again?) after the round() call:

```
ERROR sandbox.py:7:12-19: No matching overload found for function `int.__new__` called with arguments: (type[int], float | int | None) [[no-matching-overload](https://pyrefly.org/en/docs/error-kinds/#no-matching-overload)]
      Possible overloads:
      (cls: type[int], x: ConvertibleToInt = 0, /) -> int [closest match]
      (cls: type[int], x: bytearray | bytes | str, /, base: SupportsIndex) -> int
```

The type of value was correct (int | float) before the call because of the narrowing due to "if None -> return". But after the call it forgets the narrowing. Maybe the templated return types of round took the initial type and not the narrowed down one? 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABAC7x0AUAbqlAK4yI0Tp00APjTBRcqAcIByudDACUNALQA%2BGnDoAnRAB10NA71ptOMXnBoy5u-YbuaYdDpv0ByV3r12TXGgF4aTVwOLFZ2LgAaGgAWeU9bAwh-Xn4w0zj0L0MHJxd1LSYIeRAIkA46aDgSckQQAGIaAFUKqAg6UhEQgGMK2Th4qlowXE0AWwkAfXQOUewYTSZ8Hj46RVV87SyDHOd9MB0QKRm57RpgfABfA70SkDIHUVJCOlxRqAoGgAVSB6gOtCweHwNC6skgAHNnBIILJCHoGgBlGBmAAWdDoxDgiAA9Nj7tQ-oQRuDsTB0NjMLgunBsaD0BCob1ySIRjRUGxoKhsLAQWCIJDNNDZDRcMQmVU9GQ6CjZEoWPM4DD9AEDgBmQgARgATNd0CALqVUD0IPKAGLQGAUAE4AjVfVAA

_No response_

oopscompiled:
# Summary

Add a step 3 (argument type expansion) which now also runs even when step 2 already matched. If step 3 successfully resolves all expanded union branches and produces a different type from step 2's result, it is preferred since it is derived from independently matching each union member against the right overload.

Fixes #2592 

# Test Plan

cargo test, python3 test.py

AI were used for partial optimization of step 3 handling process


conformance/conformance_output.py
```diff
@@ -5,6 +5,7 @@
 
 # pyre-strict
 
+from __future__ import annotations
```
rchen152: The changes in this file look unrelated.

oopscompiled:
got it, I'll roll back the overload and work on the hint level

conformance/conformance_output.py
oopscompiled: you're right — that was from local testing. Removed.

oopscompiled:
I fixed the issues with `None` narrowing, still there are some errors regarding double-inference, working on a way to perform unhinted inference without side effects

oopscompiled:
@rchen152 please have a look, I changed forward assignment inference to avoid the previous 3 step double `expr()` behavior. Now it follows the logic to run a single unhinted pass **only** for safe call patterns, and fall back to hinted inference for attribute calls, keyword-heavy calls and other mutating collection methods with cases where unhinted inference may introduce `None` unexpectedly.

rchen152:
Thanks @oopscompiled! I haven't had a chance to take a close look yet, but the mypy_primer delta suggests that we can get some nice wins from being smarter about when we apply the hint, which is exciting.

oopscompiled:
I agree, but the thing is that fixing one bug led to the appearance of two new ones. I am very sorry for having to make so many iterations, and I apologize. I will try to fix the last two errors

oopscompiled:
Dear @rchen152,  I've tried to fix the remaining errors by adding a `# type: ignore` into the error suppressor, but faced a more deeper errors not related to the pr. After some investigation I figured out that these are pre-existing type errors in the colour library, not regressions introduced by this PR. 

Explicitly, the structured dtype arguments are genuinely not assignable to type`[DTypeReal] | None` , which the pyrefly already acknowledged this by suppressing them with` # pyright: ignore`. But since pyrefly doesn't recognize `# pyright: ignore`, these were previously hidden only because hinted inference was broadening the inferred type through the annotation hint. The new unhinted-first strategy correctly surfaces them as true positives. Please let me know if my analysis is faulty

oopscompiled:
Dear @rchen152, I've tried to fix the remaining errors by adding a # type: ignore into the error suppressor, but faced a more deeper errors not related to the pr. After some investigation I figured out that these are pre-existing type errors in the colour library, not regressions introduced by this PR.

Explicitly, the structured dtype arguments are genuinely not assignable to type[DTypeReal] | None , which the pyrefly already acknowledged this by suppressing them with # pyright: ignore. But since pyrefly doesn't recognize # pyright: ignore, these were previously hidden only because hinted inference was broadening the inferred type through the annotation hint. The new unhinted-first strategy correctly surfaces them as true positives. Please let me know if my analysis is faulty

oopscompiled:
@rchen152 please have a look

rchen152:
@oopscompiled sorry for the delay! I checked, and it looks like https://github.com/facebook/pyrefly/issues/2592 is actually already fixed at head - the example no longer errors in the sandbox. We've made some pretty major change to overload evaluation the past couple weeks, so one of them probably resolved this issue. Thanks for your efforts here, and apologies again for not getting back to this sooner.

PR #2710, opened 2026-03-08
fix Narrow types based on collection containment #2706

### LINKED ISSUE #2706: Narrow types based on collection containment
### Describe the Bug

I just ran into https://github.com/python/mypy/pull/20602 and noticed that pyrefly also doesn't type narrow in some of these cases.
```python
from typing import Literal, reveal_type

LitA = Literal["a"]
lit_dict: dict[LitA, int] = {"a": 1}
lit_tuple: tuple[LitA] = ("a",)
lit_list: list[LitA] = ["a"]
lit_set: set[LitA] = {"a"}
lit_frozenset: frozenset[LitA] = frozenset(("a",))

def no_narrowing(key: str):
    if key in lit_dict:
        reveal_type(key)  # str
        lit_dict[key]  # Argument `str` is not assignable to parameter `key` with type `Literal['a']`
    if key in lit_tuple:
        reveal_type(key)  # str
    if key in lit_list:
        reveal_type(key)  # str
    if key in lit_set:
        reveal_type(key)  # str
    if key in lit_frozenset:
        reveal_type(key)  # str
    if key in frozenset(("a",)):
        reveal_type(key)  # str

def narrowing(key: str):
    if key in ("a",):
        reveal_type(key)  # Literal['a']
    if key in ["a"]:
        reveal_type(key)  # Literal['a']
    if key in {"a"}:
        reveal_type(key)  # Literal['a']
    if key in {"a": 1}:
        reveal_type(key)  # Literal['a']
```
https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AZCAxiVUUADR1KMAG4xxAfSbEYAHXQahDAIJ0AvIOGjxAbTUhUFgLoaowxZggBjBojpPXp7TqmsG1gZ0wBZWSHQAjAC%2Bdg4MAK7EsO4JSTDewjqBhgAUoRYSAJSxDIr2cG505QwZutl05pY2JYpwMJVtNT71IU0gMej2pWDUAF4w6J3uI7jjk%2B21WUEzc505eX1FxZo7mDBgdOi4iuiolNQA7qxsOQDWMKTuFZSFiBp0H5wH96Sc6FUOTxud6fUEyeRKFQwO4PQofADEdGeINBHyGjhcNR%2BgToiJ0lDY8RoE34AANnqTOHBDrh%2BKg4HAIGxTthYIxcHRiGdUMSRJQ6KSfpSrgwABaMZgwAXaExQUwAclQ8uspJRXzoPz%2BANKqWSarBcgUUGUkphpDhuKRDEoaog3weWvR1Te-1RH3BRpNqjNFsRyNdHztGodrG1rXaLrd7sNkNNP19VptAfVmtD6JWEym%2Bs%2BHtj3vjCMTtvtv1DGfmDHW%2BRAW0jUdzxqhPsL-o0Gj2B1O51wV3YZqe1texeDpf%2BGzCRTrbobXuhBctMrEcsVyuHqf%2BjTC1inqJnTfniMXZhXtmTQfXwWrUR3BohjbjsMLR%2BXStPoPPIf%2BvTC7miN5zMb3vmj4LsYS4Kq%2BNYgPEDDQHAJDkIgICIgAqrBQy-GA8ToK4EC4JMbZYPsdBgLwNCoKU6BEtgog5Pg7j%2BHCAC0AB8iY7jICSUP8YAWAActRojuMA%2BAxCAGhQWQMhgFApCEAwtBQBQiIAAqkNJslIhgOAEHQzj4ZAhJiLB%2BGEBoiIAMowFKooMAwxBwIgAD0TlSfssmELwbBORMTmYLgzhwE5%2BnoIZ8TGXh6BOSRvB0KgsioNAqCslKIVhRF%2BF0LgxAmZMZnoGQYr4Ux8iUIymWGBYADMhARAATBYElRBIli4fIABi0AwBQaBYHgRBkP0QA

So it seems to narrow when checking against inline collections, but not if we check against typed variables.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2706

containment narrowing now derives finite membership domains directly from typed RHS containers instead of relying only on the generic iterable/mapping decomposition.

That preserves literal precision for typed dict, tuple, list, set, and frozenset values, and the literal fast path now also recognizes inline frozenset(...) calls

# Test Plan

add the test in issue

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -98,6 +98,90 @@ fn synthesize_int_slice(idx: i64) -> Expr {
 const NARROW_ENUM_LIMIT: usize = 100;
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
+    /// Membership narrowing only stays precise when the RHS describes a finite set of values.
+    fn membership_narrow_type(&self, ty: &Type) -> Option<Type> {
+        match ty {
+            Type::Literal(_) | Type::None => Some(ty.clone()),
+            Type::ClassDef(cls) => Some(Type::type_of(self.promote_silently(cls))),
+            Type::Type(inner) if matches!(inner.as_ref(), Type::ClassType(_)) => Some(ty.clone()),
+            Type::Union(union) => {
+                let members: Option<Vec<_>> = union
+                    .members
+                    .iter()
+                    .map(|member| self.membership_narrow_type(member))
+                    .collect();
+                Some(self.unions(members?))
+            }
+            _ => None,
+        }
+    }
+
+    fn membership_narrow_tuple_type(&self, tuple: &Tuple) -> Option<Type> {
+        match tuple {
+            Tuple::Concrete(elts) => {
+                let members: Option<Vec<_>> = elts
+                    .iter()
+                    .map(|elt| self.membership_narrow_type(elt))
+                    .collect();
+                Some(self.unions(members?))
+            }
+            Tuple::Unbounded(elt) => self.membership_narrow_type(elt),
+            Tuple::Unpacked(unpacked) => {
+                let (prefix, middle, suffix) = unpacked.as_ref();
+                let members: Option<Vec<_>> = prefix
+                    .iter()
+                    .chain(std::iter::once(middle))
+                    .chain(suffix.iter())
+                    .map(|elt| self.membership_narrow_type(elt))
+                    .collect();
+                Some(self.unions(members?))
+            }
+        }
+    }
+
+    fn membership_narrow_container_type(&self, ty: &Type) -> Option<Type> {
+        match ty {
+            Type::Var(v) if let Some(_guard) = self.recurse(*v) => {
+                self.membership_narrow_container_type(&self.solver().force_var(*v))
+            }
+            Type::Union(union) => {
+                let members: Option<Vec<_>> = union
+                    .members
+                    .iter()
+                    .map(|member| self.membership_narrow_container_type(member))
+                    .collect();
+                Some(self.unions(members?))
+            }
+            Type::TypedDict(typed_dict) => {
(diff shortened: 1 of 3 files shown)


yangdanny97:
it's weird that mypy primer timed out, i wonder if there is a perf regression somewhere

rchen152:
> it's weird that mypy primer timed out, i wonder if there is a perf regression somewhere

@yangdanny97 I've also started seeing mypy_primer timeouts in the past day or so (but only with the GitHub workflow).

yangdanny97:
@migeed-z the error count diff does not match the analysis for pydantic

pyrefly/lib/alt/narrow.rs
```diff
@@ -99,6 +99,89 @@ fn synthesize_int_slice(idx: i64) -> Expr {
...
+                    .chain(std::iter::once(middle))
+                    .chain(suffix.iter())
+                    .map(|elt| self.membership_narrow_type(elt))
+                    .collect();
+                Some(self.unions(members?))
+            }
+        }
+    }
+
+    fn membership_narrow_container_type(&self, ty: &Type) -> Option<Type> {
```
stroxler: I think there's a problem here. We're not using intersection semantics: obviously we're not producing an intersection type, but in addition I think this code path can probably actually upcast or type-erase (for example if we do `x in l` where `l: list[Any]`, does it change the type of `x` to `Any`?

I think it does, which is not great behavior. Similarly if `l: list[object]` we would upcast `x` to `object` and basically every use of `x` becomes a type error.

stroxler (changes requested):
Let's add some tests for upcasting / type erasure, and improve the behavior if we see a problem

asukaminato0721:
> pydantic: +4630 -59

except this...

PR #2712, opened 2026-03-08
Use release builds for mypy primer

stroxler:
Summary:
We started using pyrefly init to set up, which required moving the builds out of
mypy primer's control.

In the process, we accidentally switched the build mode from dev to release.

Differential Revision: D95741090


rchen152 (approved):

PR #2713, opened 2026-03-08
Restore the old mypy primer blocklist

stroxler:
Summary:
Mypy primer is pretty heavyweight, and github CI is underpowered.

I was looking at runs from before it started timing out, and realized that a lot of things
that take 1-2s to check internally take much longer.

I think it's possible that a few projects (e.g. pandas, sympy) might just be too big to
type check massively in parallel on a tiny, underpowered server.

Working off that assumption, let's try restoring the old "determinism blocklist" - these
projects tend to be some of the biggest, just because the things that trigger nondeterminism
also make type check expensive.

Differential Revision: D95741691


rchen152 (approved):

PR #2714, opened 2026-03-09
Remove stale TODO

stroxler:
Summary: This test is fine now, I think something got fixed but not cleaned up.

Differential Revision: D95749563


yangdanny97 (approved):

PR #2715, opened 2026-03-09
Exclude all CPython subdirectories from github mypy primer

stroxler:
Summary:
I'm having trouble figuring out exactly what the problem is (I cannot repro
slowness locally), but very clearly it's the three CPython subdirectories
that are stuck in CI.

I still have no theory what actually causes this, although I'm much less
concerned now that it's something fixpoint-related, given that a fixpoint
issue would presumably reproduce locally.

I think for now we might want to just disable them - given that it's exactly
three subdirectories of the same project causing us problems (and all three are
very small subprojects), I very strongly suspect something in mypy primer
configuration is messing up pyrefly, as opposed to something in the core type
checker being to blame.

This should allow us to debug at leisure without interrupting CI on PRs.

Differential Revision: D95752934


yangdanny97 (approved):

PR #2718, opened 2026-03-09
Changes to mypy primer scripts should be tested:x

stroxler:
Summary:
Rebecca noted in D95741691 that if you want to test a change to mypy_primer.yml, you need
this. I can't imagine any good reason why we *wouldn't* want to always test changes to the
scripts, this seems like basic hygiene.

Differential Revision: D95759320


rchen152 (approved):

PR #2719, opened 2026-03-09
Re-enable mypy-primer on non-CPython projects

stroxler:
Summary:
I re-enabled the blocklist as part of debugging a CI outage.

Now that the outage is cleared, and cpython is the primary suspect, it makes sense
to see what happens if we revert that change (so that we run everything again, *except*
for cpython which we know has some problems).

Differential Revision: D95759318


rchen152 (approved):

PR #2725, opened 2026-03-09
Fix false positive invalid-type-var for generic functions assigned to attributes (#2204)

### LINKED ISSUE #2204: False positive invalid-type-var from check_and_sanitize_type_parameters
### Describe the Bug

When we assign instance attributes, we need to check for _unbound_ type variables, but the check is incorrectly also flagging type variables which are bound by the right-hand side expression.

```

def f[T](x: T) -> T:
    return x

class C:
    def m[U](self, x: U) -> U:
        return x

class D:
    def __init__(self, c: C):
        self.f = f # Unexpected invalid-type-var
        self.g = c.m # Unexpected invalid-type-var

def test(o: D):
    reveal_type(o.f) # should be [T](x: T) -> T
    reveal_type(o.g) # should be BoundMethod[C, [U](self: C, x: U) -> U]
```

I believe the error is happening here: https://github.com/facebook/pyrefly/blob/6fabfc70c57bbaad325eda052ed1a2030de33823/pyrefly/lib/alt/class/class_field.rs#L2323

We collect the quantified types from the RHS expression, but we only want unbound quantifieds.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRzMMMHTABtACoBdABT5EddQEo6AWgB8BxHLo2BMBgFdK6OvjlyAxlFRw4dAMJWLraKyjSqAKo6cDBQYAA0rvoRxuZ0EUG2WXaOzq7u6F4%2BfgAimSFKdOLirBAM1doxcYke%2Bv6G5dl0TWCEygC8KtZd3bG9HIMehDQFoYzwDNq4%2BiUdwzaCImKSzDBLfcYAxN0AFrgOUJh02DB0Gjp6BqkW6ut2WxJSe7iEbEen50u11uACFzlgALL2M6YVT%2BRKRaJjNqJR4pUwWKIgeIgMiCMBQUiEBi0KAUY4ABVI%2BMJ3QwOAIdA8uHQkDYTlQDAgLMIcmOAGUYLcTgwGMQ4IgAPSSvFKQmEXhsSUwdCSzC4DxwSXM1kQdmUTnc1UqXh0VBCVDQVDYWBMllsjlcll0XDEJ3oOC89BkBhndAmESUOBGuiDGQgADMhAAjAAmcNyEAAXxxqA8XJEADFoDAKGgsHgiGRk0A

_No response_

yangdanny97:
Summary:
fixes https://github.com/facebook/pyrefly/issues/2204

Fixed check_and_sanitize_type_parameters to also exclude Forall-bound tparams within the type of the attribute.

Differential Revision: D95799931

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2431,10 +2431,37 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             return ty;
         }
         let class_tparams = self.get_class_tparams(class);
+        // Exclude quantifieds bound by Forall nodes within the type
+        // (e.g. generic functions assigned to attributes have their own
+        // type parameters that should not trigger invalid-type-var errors).
+        let mut forall_bound: SmallSet<&Quantified> = SmallSet::new();
+        fn collect_forall_tparams<'a>(ty: &'a Type, acc: &mut SmallSet<&'a Quantified>) {
+            match ty {
+                Type::Forall(forall) => {
+                    for q in forall.tparams.iter() {
+                        acc.insert(q);
+                    }
+                }
+                Type::BoundMethod(bm) => {
+                    if let BoundMethodType::Forall(forall) = &bm.func {
+                        for q in forall.tparams.iter() {
+                            acc.insert(q);
+                        }
+                    }
+                }
+                _ => {}
+            }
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

PR #2727, opened 2026-03-09
Fix NamedTuple subclass not assignable to TypeVar bound=NamedTuple (#2455)

### LINKED ISSUE #2455: `NamedTuple` inheriting class isn't assignable when using `TypeVar`
### Describe the Bug

I did notice another `NamedTuple` related bug in #2418 but I think this is separate enough since I'm defining the class instead of just dynamically creating it? If not, I'm happy to fold this into that if it's all related with the explanation of how `NamedTuple` works under the hood.

```python
from typing import TypeVar
from typing import NamedTuple

MyCustomModelT = TypeVar(name="MyCustomModelT", bound=NamedTuple)

class ImplementedModel(NamedTuple):
    name: str
    age: int

def arbitrary_method(_name: str, _age: int, _model: type[MyCustomModelT]) -> MyCustomModelT:
    return _model(**{"name": _name, "age": _age})
o = arbitrary_method("a", 2, ImplementedModel)
```

Python: 3.12

There's two different type issues here that are reported (one of which is similar to `mypy`)
1. There's no keyword argument for name/age within `tuple.__new__` which is I think what `mypy` finds when it says there's no overload variant for this call? Specific error from `mypy` is `call-overload`.
2. I get a `ImplementedModel is not assignable to upper bound NamedTuple of type variable MyCustomModelT`. I assumed that because `MyCustomModelT` is a subclass of `NamedTuple`, I could provide a `NamedTuple` inheriting object but it doesn't look like `pyrefly` agrees with me.

Admittedly, this is somewhere Pydantic could replace all of this but I have an additional constraint where that is not an option.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjABqqSgB10YavSYt2nbr34A5VDRiYBAV2KwpUgLKkAwjrgNaR3JhhQBdALyDhYygAp0GmE4kgTc0trW3sBfwAaOjwdLCd1TW09WABKQ3QAYyhUODg6AEluWE10Bi0bOygPBK1dfRgUxCk6FrpvTUQ6S0l0VrpUNhhO1gZ0uzB%2BymwIBkpxUgB9TQYAC1sPBfahrtmohYHtkb2aUKhO%2BRgAbUCLKxoKsIBdFLoAWgA%2BOhvg%2B9OBJt6rUoMAYOkovSWpw8ACpocB-Ft-J1Nj4ov4Dki6PtBgBfNLoXDOSbTWbzJYgtaYDzoyJ0ABMUUK9RKZUwDygKRAERAOgY0DgJHIiBAAGI6ABVPlQGakOhgWIZPm4dBwMYwCZgXg0VAMTY6GjYGCefDDUovD47SgAvrA0HguX%2BVT6w1WujAfA4-xSLkgMjAsBQUiEO5QChigAKpH9ga6GBwBDoGWVkDYYJ1EGVhCkYoAyjAYHQVgwGMQ4IgAPTlv3qwOEXhscswdDlzC4DJwctJ6QQVNzJXNuW8foAN1Q0FQ2FgieTPbT-bouGI-YFUjIq2Vr2HRrgGd6Ln8AGZCABGOle9AgHHc1CKiBbgBi0BgFDQWDwRDIl6AA

_No response_

yangdanny97:
Summary:
When `NamedTuple` is used as a base class, pyrefly substitutes `NamedTupleFallback`
into the MRO. However, when `NamedTuple` is used as a type annotation (e.g. in a
TypeVar bound), it resolved to the `typing.NamedTuple` class — a completely different
class not in any NamedTuple subclass's MRO. This caused `bad-specialization` errors
when passing NamedTuple subclasses to TypeVar-bounded parameters.

Fix: In `canonicalize_all_class_types`, when `typing.NamedTuple` appears as a
ClassDef in a type annotation context, resolve it to `NamedTupleFallback` — the
class that actually appears in user-defined NamedTuple subclasses' MRO.

fixes https://github.com/facebook/pyrefly/issues/2455

Differential Revision: D95804828

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1411,6 +1411,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         .mk_type_form(self.heap.mk_type_form(self.heap.mk_any_implicit()));
                 } else if cls.has_toplevel_qname("typing", "Any") {
                     *ty = self.heap.mk_type_form(self.heap.mk_any_explicit())
+                } else if cls.has_toplevel_qname("typing", "NamedTuple") {
+                    // When `NamedTuple` is used as a type annotation (e.g. TypeVar bound),
+                    // resolve to `NamedTupleFallback` — the class that actually appears in
+                    // the MRO of user-defined NamedTuple subclasses.
+                    *ty = self.heap.mk_type_form(
+                        self.heap
+                            .mk_class_type(self.stdlib.named_tuple_fallback().clone()),
+                    );
                 } else {
                     // All other classes (including Tensor) get promoted and wrapped in type_form
                     *ty = self.heap.mk_type_form(self.promote(cls, range, errors));
--- a/pyrefly/lib/test/named_tuple.rs
+++ b/pyrefly/lib/test/named_tuple.rs
@@ -622,3 +622,23 @@ Point = collections.namedtuple("Point", ["x", "y"])
 p = Point(x=1, y=2)
     "#,
 );
+
+// https://github.com/facebook/pyrefly/issues/2455
+testcase!(
+    test_named_tuple_typevar_bound,
+    r#"
+from typing import TypeVar, NamedTuple, assert_type
+
+MyCustomModelT = TypeVar(name="MyCustomModelT", bound=NamedTuple)
+
+class ImplementedModel(NamedTuple):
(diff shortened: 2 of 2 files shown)


rchen152 (approved):

PR #2730, opened 2026-03-09
Add type related messages to TSP for fetching types at nodes and declarations

rchiodo:
# Summary

This adds the next set of messages for TSP to Pyrefly. These are the type related messages. These are really the main piece of information in TSP

# Test Plan

Added a number unit and integration tests. Also tested this using Pylance as it's pretty close to being usable :)

Here's it working in Pylance:

image

Same Pylance, using a Pylance specific feature (restructured text support) but with Pyrefly as the backend:

image

/cc @kinto0


yangdanny97:
ignore the primer results...

PR #2754, opened 2026-03-10
Add documentation hover for top-level bindings with docstring

### LINKED ISSUE #2620: Docstrings on variables
### Describe the Bug

Given this code:

```py
VAR = "abc"
"""Some documentation."""
```

With pyright, if I hover `VAR`[1], it shows:

```
(constant) VAR: Literal['abc']
──────────────────────────────
Some documentation
```

While pyrefly shows only:

```
(constant) VAR: Literal['abc']
```

I know docstrings on variables are non-standard, but some tools support them, and they're quite useful. Does pyrefly have any plans to support this, or something else instead?

Versions:
- pyright 1.1.408
- pyrefly 0.54.0
- python 3.12.3
- Ubuntu 24.04.3

[1] `vim.lsp.buf.hover` on neovim 0.11.6

### Sandbox Link

[pyright](https://pyright-play.net/?code=GoQQSgBAvBBECGAjAxrAULTBlA9gWwFMIATHZAV0IDsAXeGgSxyoDpNYg)

[pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAagIIBKABALz0A6Iq2Axmy%2BjyGwDKuALYx6mXJwCuY9ABdU8iLnSF%2BbEABoQ05VDglyiEAGJ6AVX0R5pemGnpOy1XF69MMMPdwAnEUoA%2Buiy2DC%2BABT4iPQQCgCU9AC0AHz0cPK%2BiLz0ufS%2BMPLSvuj2bAByoeExwPgAvjzo2iBkBWBQpITyolAU5gAKpG0d6Rg4BPScqpAA5sVKKmq85oIw4gAW8vLEcIgA9HutXh2EfjN7MOh7kpxwe1Pos-MuVz6%2B9KgAbqjQHLCT0wgc18C1U9FwxBehl4ZHk61USU%2B4TgiyYrBAAGZCABGABMjRAdR0qGcECRADFoDAKGgsHgiGRCUA)

neovim 0.11.6

```
vim.lsp: Active Clients ~
- pyrefly (id: 1)
  - Version: 0.54.0
  - Root directory: ~/dev/genret
  - Command: { "pyrefly", "lsp" }
  - Settings: {}
  - Attached buffers: 1
```

oyarsa:
# Summary

This PR adds attribute docstrings ([PEP 257](https://peps.python.org/pep-0257/)) to hover from assignments, including `type` aliases.

Fixes #2620

# Test Plan

- `test.py` passes. It generated no changes to be commited. There are some compilation warnings, but I believe they are unrelated to the code in this PR.
- Tested as the LSP on neovim 0.11.6 with default lspconfig settings. I copied `target/release/pyrefly` to a directory in `PATH`.

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -257,10 +257,6 @@ pub struct BindingsBuilder<'a> {
     lambda_yield_keys: Vec<(TextRange, Box<[Idx<KeyYield>]>, Box<[Idx<KeyYieldFrom>]>)>,
     /// See `BindingsInner::subsequently_initialized`.
     subsequently_initialized: SmallSet<Idx<KeyAnnotation>>,
-    /// The range of a PEP 257-style variable docstring (a standalone string literal
-    /// immediately following the current assignment statement), if any. Set by `stmts`
-    /// with look-ahead before calling `stmt`, consumed by `bind_single_name_assign`.
-    pub pending_var_docstring: Option<TextRange>,
 }
 
 /// An enum tracking whether we are in a generator expression
@@ -516,7 +512,6 @@ impl Bindings {
             deferred_bound_names: Vec::new(),
             lambda_yield_keys: Vec::new(),
             subsequently_initialized: SmallSet::new(),
-            pending_var_docstring: None,
         };
         builder.init_static_scope(&x.body, true);
         if module_info.name() != ModuleName::builtins() {
@@ -948,17 +943,17 @@ impl<'a> BindingsBuilder<'a> {
         let mut iter = xs.into_iter().peekable();
         while let Some(x) = iter.next() {
             // PEP 257-style variable docstrings: a standalone string literal immediately
-            // following an assignment is the variable's docstring. Set it before `stmt`
-            // so `bind_single_name_assign` can read it; clear it after.
-            if let Stmt::Assign(_) | Stmt::AnnAssign(_) = x
+            // following an assignment is the variable's docstring.
+            let var_docstring = if matches!(x, Stmt::Assign(_) | Stmt::AnnAssign(_))
                 && let Some(Stmt::Expr(e)) = iter.peek()
                 && let Expr::StringLiteral(_) = e.value.as_ref()
             {
-                self.pending_var_docstring = Some(e.range());
-            }
+                Some(e.range())
+            } else {
+                None
+            };
 
-            self.stmt(x, parent);
-            self.pending_var_docstring = None;
+            self.stmt(x, parent, var_docstring);
         }
     }
 
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -257,6 +257,10 @@ pub struct BindingsBuilder<'a> {
     lambda_yield_keys: Vec<(TextRange, Box<[Idx<KeyYield>]>, Box<[Idx<KeyYieldFrom>]>)>,
     /// See `BindingsInner::subsequently_initialized`.
     subsequently_initialized: SmallSet<Idx<KeyAnnotation>>,
+    /// The range of a PEP 257-style variable docstring (a standalone string literal
+    /// immediately following the current assignment statement), if any. Set by `stmts`
+    /// with look-ahead before calling `stmt`, consumed by `bind_single_name_assign`.
+    pub pending_var_docstring: Option<TextRange>,
 }
 
 /// An enum tracking whether we are in a generator expression
@@ -512,6 +516,7 @@ impl Bindings {
             deferred_bound_names: Vec::new(),
             lambda_yield_keys: Vec::new(),
             subsequently_initialized: SmallSet::new(),
+            pending_var_docstring: None,
         };
         builder.init_static_scope(&x.body, true);
         if module_info.name() != ModuleName::builtins() {
@@ -940,8 +945,20 @@ impl<'a> BindingsBuilder<'a> {
     }
 
     pub fn stmts(&mut self, xs: Vec<Stmt>, parent: &NestingContext) {
-        for x in xs {
+        let mut iter = xs.into_iter().peekable();
+        while let Some(x) = iter.next() {
+            // PEP 257-style variable docstrings: a standalone string literal immediately
+            // following an assignment is the variable's docstring. Set it before `stmt`
+            // so `bind_single_name_assign` can read it; clear it after.
+            if let Stmt::Assign(_) | Stmt::AnnAssign(_) = x
+                && let Some(Stmt::Expr(e)) = iter.peek()
+                && let Expr::StringLiteral(_) = e.value.as_ref()
+            {
+                self.pending_var_docstring = Some(e.range());
+            }
+
             self.stmt(x, parent);
+            self.pending_var_docstring = None;
         }
     }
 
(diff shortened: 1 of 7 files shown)


pyrefly/lib/export/definitions.rs
```diff
@@ -414,8 +414,37 @@ impl Definitions {
...
+                && let Some(Stmt::Expr(e)) = xs.get(i + 1)
+                && let Expr::StringLiteral(_) = e.value.as_ref()
+            {
+                self.set_var_docstring(x, e.range());
+            }
+        }
+    }
+
+    /// Set `docstring_range` on all names defined by `assign_stmt`.
+    fn set_var_docstring(&mut self, assign_stmt: &Stmt, docstring_range: TextRange) {
```
kinto0:
```suggestion
    fn set_var_docstring(&mut self, targets: <iterable of targets>, docstring_range: TextRange) {
```

if we map Stmt::Assign/AnnAssign to a collection of targets and `expr_lvalue` each of them, this function does not require invariants on top of the input types. this avoids the caller match -> callee match of the same cases.

pyrefly/lib/test/lsp/hover_docstring.rs
```diff
@@ -351,6 +351,68 @@ Docstring Result: `Test docstring`
...
+# main.py
+4 | print(VAR)
+          ^
+Docstring Result: `Some documentation.`
+"#
+        .trim(),
+        report.trim(),
+    );
+}
+
```
kinto0:
```suggestion
```

seems like a lot of duplicates. I like the names of the tests on the bottom more (<>_variable_docstring), so can we move the "itself" test down there (the only one I can tell that isn't a duplicate)

pyrefly/lib/binding/bindings.rs
```diff
@@ -257,6 +257,10 @@ pub struct BindingsBuilder<'a> {
     lambda_yield_keys: Vec<(TextRange, Box<[Idx<KeyYield>]>, Box<[Idx<KeyYieldFrom>]>)>,
     /// See `BindingsInner::subsequently_initialized`.
     subsequently_initialized: SmallSet<Idx<KeyAnnotation>>,
+    /// The range of a PEP 257-style variable docstring (a standalone string literal
+    /// immediately following the current assignment statement), if any. Set by `stmts`
+    /// with look-ahead before calling `stmt`, consumed by `bind_single_name_assign`.
+    pub pending_var_docstring: Option<TextRange>,
```
kinto0: I understand the rationale for this. however, it adds complicated statefulness when it may not be the only option. 

I think it would be more maintainable and easier to reason about if we instead passed this context through function calls. for example, can we instead add var_docstring to the BindingsBuilder::stmt call? and then pass it through to the bind_<> binding creation?

kinto0 (changes requested):
looks great! just a few suggestions

oyarsa:
I believe these changes address your comments. Sorry if I forgot/misunderstood anything. They're separate commits, so you can review each one individually, but I plan to squash them later (or they can be squashed on merge on GitHub).

pyrefly/lib/export/definitions.rs
oyarsa: Yes, this makes sense. This means we don't get docstrings for type aliases. Is that something we want?

oyarsa: If yes, I implemented that [here](https://github.com/oyarsa/pyrefly/commit/c5bec0938a1fdcab48f0218a475b15409bec5547)

kinto0:
thank you! sorry about the delayed review - I took a few days off last week. let me import it and get another stamp to merge it

grievejia (changes requested):

oyarsa:
@grievejia GitHub says you have requested changes, but I can't see what they are.

pyrefly/lib/binding/stmt.rs
```diff
@@ -690,6 +691,7 @@ impl<'a> BindingsBuilder<'a> {
                             &name,
                             value,
                             Some((&x.annotation, ann_idx)),
+                            var_docstring,
                         ),
                         None => self.bind_definition(
```
kinto0, line 696: jia comment: we silently drop docstrings with no value (e.g. `x: int` instead of `x: int = 5`). we should handle that + add tests since it's probably common to annotate these with a docstring

pyrefly/lib/export/definitions.rs
```diff
@@ -414,8 +414,39 @@ impl Definitions {
 
 impl<'a> DefinitionsBuilder<'a> {
     fn stmts(&mut self, xs: &[Stmt]) {
-        for x in xs {
+        for (i, x) in xs.iter().enumerate() {
             self.stmt(x);
+            // PEP 257-style variable docstrings: a standalone string literal expression
```
kinto0, line 419: jia comment: pep 257-style is a little confusing since it's actually not in pep 257. let's just remove this comment + update the comment on the NameAssign docstring_range

pyrefly/lib/binding/stmt.rs
```diff
@@ -653,6 +653,7 @@ impl<'a> BindingsBuilder<'a> {
                         &Ast::expr_name_identifier(name.clone()),
                         x.value,
                         None,
+                        var_docstring,
                     );
                 } else {
                     self.bind_targets_with_value(&mut x.targets, &mut x.value);
```
kinto0, line 659: jia comment: docstring is silently dropped for multi-target assignments (a = b = 1) and tuple unpacking (a, b = 1, 2). This disagrees with the export side in definitions.rs, which walks all lvalue targets via Ast::expr_lvalue. The two sides should agree: either both handle complex lvalues or both restrict to single names.

As a matter of fact, it's not really clear to me why this diff chooses to calculate this info in both definition phase and binding phase in the first place. Intuitively I would assume that we could just do the var docstring detection in definition phase, and both the binding phase and the IDE hover can simply reuse the same info.

pyrefly/lib/binding/target.rs
kinto0, line 522: jia comment: When the assignment is recognized as a type alias, the docstring_range parameter is silently ignored. So MyType = list[int] followed by a docstring would not show documentation on hover. This could arguably be intentional but may worth documenting.

pyrefly/lib/test/lsp/hover_docstring.rs
```diff
@@ -373,6 +393,78 @@ print(lib)
...
+        test_report_factory(lib),
+    );
+    assert_eq!(
+        r#"
+# main.py
+3 | print(VAR)
+          ^
+Docstring Result: `Some documentation.`
+
+
```
kinto0, line 467: Consider adding:

- Type alias with docstring
- Annotation-only without value (x: int + docstring)
- Function-local assignment with docstring (to document whether this is intended)
- Multi-target assignment (a = b = 1 + docstring)
- Multiple definitions with same name, e.g. (a = 1; 'docstring1'; a = 2 'docstring2').

kinto0 (changes requested):
sorry about the details here. he left comments on my import instead of on github so it didn't show up. here are the comments

oyarsa:
I still intend to finish this. Sorry for the inactivity.

rchen152:

PR #2764, opened 2026-03-10
Resolve ambiguous overload calls by selecting the most general return type among the matched overloads (#2764)

rchen152:
Summary:
Previously, if a call to an overloaded function matched more than one overload and the return types of the matched overloads weren't all equivalent, we fell back to a return type of `Any`, as the spec says to do.

With this diff, we instead try to select the "most general" return type among the matched overloads, by checking if there exists a return type such that all materializations of every other return type are assignable to it. Some examples of what this means:
```python
overload
def f(...) -> A[int]: ...
overload
def f(...) -> A[Any]: ...

# Old return type: Any
# New return type: A[Any]
f(<ambiguous arguments>)
```
```python
overload
def f(...) -> bool: ...
overload
def f(...) -> int: ...

# Old return type: Any
# New return type: int
f(<ambiguous arguments>)
```
```python
overload
def f(...) -> A[int]: ...
overload
def f(...) -> A[str]: ...

# Old return type: Any
# New return type: Any (neither return type is more general than the other)
f(<ambiguous arguments>)
```

We now pass more of numpy and scipy-stubs' assert_type tests:
| project | assert_type failures at head | assert_type failures at previous diff | assert_type failures at current diff |
| numpy | 134 | 186 | 127 |
| scipy-stubs | 49 | 101 | 19 |

The trade-off is that we report a lot of new errors on mypy_primer projects. I investigated a bunch of them, and they're not wrong, per se, but some are arguably low-value. Examples:
```
def f(x: Sequence[str], idx) -> str:
    # The possible types of `x[idx]` are str (if idx is an int)
    # and Sequence[str] (if idx is a slice). str <: Sequence[str],
    # so we pick the latter as the return type.
    # This is technically correct...but idx is probably meant to be an int.
    return x[idx]  # error!
```
```
def f(x: bool, y) -> bool:
    # Similar to above, this is technically correct because y may be an int,
    # but y is probably meant to be a bool.
    return x & y  # error!
```

The LLM classifier seems to have gotten really confused on this one XD The detailed analysis says that Pyrefly is now "stricter than the established ecosystem standard", but we pass more of numpy and scipy-stubs' `assert_type` assertions than before, meaning this diff brings us closer to the ecosystem standard. Because the PR includes the whole stack, it's also complaining about errors introduced in previous diffs.

Reviewed By: grievejia

Differential Revision: D95667476

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -629,7 +629,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     let _ = matched_overloads.split_off(split_point);
                 }
             }
-            let selected_overload = self.disambiguate_overloads_spec_compliant(&matched_overloads);
+            let selected_overload = if self.solver().spec_compliant_overloads {
+                self.disambiguate_overloads_spec_compliant(&matched_overloads)
+            } else {
+                self.disambiguate_overloads(&matched_overloads)
+            };
             if let Some(idx) = selected_overload {
                 let overload = matched_overloads
                     .into_iter()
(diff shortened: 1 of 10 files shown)


grievejia (approved):

rchen152:
cc @migeed-z - some odd-looking classifier results in https://github.com/facebook/pyrefly/pull/2764#issuecomment-4130102767. The detailed analysis describes mixed results (some improvements, some regressions), but they're somehow all reported as regressions in the top-level table.

PR #2767, opened 2026-03-10
fix Pyrefly doesn't catch type errors in fstrings #2669

### LINKED ISSUE #2669: Pyrefly doesn't catch type errors in fstrings
### Describe the Bug

Pyrefly misses type errors in fstrings.

It seems like it's not checking expressions in fstrings at all.

This code...

```python
val1 = len(False)
msg1 = f"{len(False)}"

val2 = len(42)
msg2 = f"{len(42)}"

val3 = 2 + "Foo"
msg3 = f"{2 + 'Foo'}"

arr: list[int] = [1, 2, 3]
val4 = arr["foo"]
msg4 = f"{arr['foo']}"
```

... only produces errors on the `val` assignments, not the `msg` assignments with the same expressions wrapped in fstrings.

```
ERROR Argument `Literal[False]` is not assignable to parameter `obj` with type `Sized` in function `len` [bad-argument-type]
 --> test.py:1:12
  |
1 | val1 = len(False)
  |            ^^^^^
  |
  Protocol `Sized` requires attribute `__len__`
ERROR Argument `Literal[42]` is not assignable to parameter `obj` with type `Sized` in function `len` [bad-argument-type]
 --> test.py:4:12
  |
4 | val2 = len(42)
  |            ^^
  |
  Protocol `Sized` requires attribute `__len__`
ERROR `+` is not supported between `Literal[2]` and `Literal['Foo']` [unsupported-operation]
 --> test.py:7:8
  |
7 | val3 = 2 + "Foo"
  |        ^^^^^^^^^
  |
  Argument `Literal['Foo']` is not assignable to parameter `value` with type `int` in function `int.__add__`
ERROR Cannot index into `list[int]` [bad-index]
  --> test.py:11:8
   |
11 | val4 = arr["foo"]
   |        ^^^^^^^^^^
   |
  No matching overload found for function `list.__getitem__` called with arguments: (Literal['foo'])
  Possible overloads:
  (i: SupportsIndex, /) -> int [closest match]
  (s: slice[Any, Any, Any], /) -> list[int]
 INFO 4 errors
```

### Sandbox Link

[snipped in pyrefly sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAbqlAIwAEAvPbOgBQBiNcMAlADroAtnADmDZmH4hgrTtz4BfaYMHUoAJiYsY7ACwaBwsVsnTZutgd7KQq9OoDM2rQGp60jrlwrjo52YybvQA5F64Ibb2qABOMYgsEHAALgDaEOjJALraqbQANPQahY5ZajR62rExqdJg3tJlfpWBwNWpIfURWbYg%2BSAArsnQcCTkiCAAxPQAqsNQEMmk9GAD6ADGw7jocPaYMGAruDFCqMkA%2BugDQtgwMWz4CRnJvPQAtAB89CnxgvT-9BiMGSAxi6BW0gActdbvF6MB8FF0H0QGQgWAoKRCMlcEIoBRpgAFUjozHfDA4Aj0dbbSCiUFnCDbQiCaYAZRgMHoAAtksliHBEAB6IVog6YwjHURC3RCzC4dZwIU09B0hlbdBCo4xeioajQVDYWDU2kQekxRnbei4Yga0aCMjJbnbN6UO5wJng5jSRyEWgaXwgRT9VCbCBujjQGAUNBYPBEMhBoA)

```
» python --version
Python 3.13.5
» pyrefly --version
pyrefly 0.55.0
```

```toml
[tool.pyrefly]
project-includes = ["**/*.py*", "**/*.ipynb"]
project-excludes = [".venv/**"]
search-path = ["."]
python-interpreter = ".venv/bin/python"
site-package-path = [".venv/lib/python3.13/site-packages"]
```

asukaminato0721:
# Summary

Fixes #2669

F-string interpolations now use the normal error collector, so type errors inside `f"{...}"` are reported instead of being swallowed.

# Test Plan

update test and add test

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -587,12 +587,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 }
             }
             Expr::FString(x) => {
-                // Swallow type errors in f-string interpolations: there is no valid
-                // place to add a suppression comment inside a string literal.
-                let fstring_errors = self.error_swallower();
                 let mut all_literal_strings = true;
                 x.visit(&mut |x| {
-                    let fstring_expr_ty = self.expr_infer(x, &fstring_errors);
+                    let fstring_expr_ty = self.expr_infer(x, errors);
                     if !fstring_expr_ty.is_literal_string() {
                         all_literal_strings = false;
                     }
--- a/pyrefly/lib/error/suppress.rs
+++ b/pyrefly/lib/error/suppress.rs
@@ -13,6 +13,7 @@ use std::sync::LazyLock;
 use anyhow::anyhow;
 use pyrefly_config::error_kind::ErrorKind;
 use pyrefly_python::ast::Ast;
+use pyrefly_python::ignore::find_comment_start;
 use pyrefly_python::ignore::find_comment_start_in_line;
 use pyrefly_python::module::GENERATED_TOKEN;
 use pyrefly_python::module_path::ModulePathDetails;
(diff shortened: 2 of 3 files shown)


yangdanny97:
hi, sorry, i forgot to assign the original issue to myself.

we originally turned off those errors because it was causing our suppression script to insert comments inside of f-strings, which can potentially affect user-visible text.

i have a fix that also updates the entire whole error suppression pipeline to handle suppressions differently for f-string and t-string (basically, if there's a suppression on the line before the f-string, or the last line of the f-string, we suppress errors for the whole f-string. and the suppression script also knows where to add the suppression)

PR #2769, opened 2026-03-10
Mark __all__ as unresolvable when augmented assignment or .extend() RHS can't be statically analyzed

yangdanny97:
Summary:
Pyrefly statically analyzes `__all__` to determine which names a module exports via `from mod import *`. When it encounters patterns like:

```python
__all__ = ["Base"]
__all__ += _impl._distn_names  # computed at runtime
```

it was silently ignoring the unresolvable `+=` while still treating `__all__` as fully resolved. This meant only `["Base"]` was exported, even though `_distn_names` contained additional names at runtime.

With this diff, when the RHS of `__all__ +=` or `__all__.extend()`/`__all__.append()` can't be statically analyzed, the entire `__all__` is marked as unresolvable, falling back to exporting all public module-level names. This matches the existing behavior for `__all__ = generate_all()`.

Differential Revision: D95857423

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -659,11 +659,18 @@ impl DefinitionsBuilder {
             Stmt::AugAssign(x) => {
                 self.named_in_expr(&x.value);
                 if DunderAllEntry::is_all(&x.target) && x.op == Operator::Add {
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #2770, opened 2026-03-10
Resolve bare module names in __all__ = foo.__all__ to fully qualified names

yangdanny97:
Summary:
When a module re-exports another module's `__all__` via a relative import:

```python
from . import _api
__all__ = _api.__all__
```

Pyrefly was creating an unqualified module reference `_api` instead of the fully qualified `pkg.sub._api`. The wildcard lookup then searched for a non-existent top-level module `_api` and found nothing, causing all re-exported names to be missing.

Now, after parsing `foo.__all__` references, Pyrefly resolves the bare name `foo` through the import definitions already in scope to recover the fully qualified module name. This handles `from . import foo`, `import X as foo`, and `from X import foo as foo` patterns.

This fixes false positive `missing-attribute` errors on packages that use `__all__ = submod.__all__` with relative imports (e.g. `scipy.signal`, `scipy.ndimage`).

Differential Revision: D95857420

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -485,6 +485,38 @@ impl DefinitionsBuilder {
         });
     }
 
+    /// Resolve bare module names in `DunderAllEntry::Module` entries to their
+    /// fully qualified names using import definitions in scope.
+    ///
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #2778, opened 2026-03-11
close #923 support pkgutil-style namespace packages

### LINKED ISSUE #923: Legacy pkgutil namespace packages not found
### Describe the Bug

After upgrading pyrefly from 0.27.0 to 0.28.0 it not longer finds imported python packages.
We pass them with the --site-package-path flag in the CLI and they appear correctly in thje logs 'Site package path from user:'.
The imports are also not found if it queries the site packages from the python interpreter.

Let me know if I can provide further information

### Sandbox Link

_No response_

_No response_

mdbernard:
# Summary

This change adds support for legacy pkgutil-style namespace packages, including those using pip's `compat` editable install mode. This is already supported by popular language servers like Jedi and Pyright, and lack of support in Pyrefly makes developers of those repos reluctant to switch despite Pyrefly's much faster speed and lower resource utilization.

Notably, Mypy does not support pkgutil-style packages (see https://github.com/python/mypy/issues/9393), requiring developers of these types of repos to configure massive `MYPYPATH` environment variables to include all the dependency folders, so this change makes Pyrefly not only a great language server for these types of repos, but also makes it work out-of-the-box as a type-checker.

# Example using azure-sdk-for-python

1. Clone the [azure-sdk-for-python](https://github.com/Azure/azure-sdk-for-python/tree/main) repo.
2. Set up a new virtual environment (I used CPython 3.14).
3. Run the following commands to set up an editable install for an arbitrary project in that repo which imports from `azure-core`, and also editably install `azure-core`:
```bash
cd sdk/keyvault/azure-keyvault-secrets
pip install -r dev_requirements.txt -e . --config-settings editable-mode=compat
cd ../../..
cd sdk/core/azure-core
pip install -r dev_requirements.txt -e . --config-settings editable-mode=compat
```

In your virtual environment's site-packages folder, you should see the following:
```bash
__editable__.azure_keyvault_secrets-4.10.1.pth
__editable__.azure_core-1.38.4.pth
```

4. Configure your editor to use the latest Pyrefly release as its language server.
5. Open `sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_client.py`.

In this file, imports from `azure.core` are `Unknown`, despite `azure.core` being installed:

image

The problem here is that when searching for the `azure` package across the search roots, the `find_one_part` function short-circuits after finding a `RegularPackage` named `azure`. It only returns the first root found--in this case the one containing `keyvault`, discarding `azure.core` since it is defined in a different root : https://github.com/facebook/pyrefly/blob/40dc64b99fbeabf4679d122a18cf323c29224d4c/pyrefly/lib/module/finder.rs#L215.

The solution here is to add a new `FindResult` variant for legacy pkgutil-style namespace packages. If `pkgutil.extend_path` is detected in an `__init__.py`, then `find_one_part` shouldn't short-circuit, instead accumulating roots similar to what we do for `NamespacePackage`s. Finally we update the callers of `find_one_part` to search over all roots until a best match is found.

Once compiled, we can verify in our editor that our LSP can find the file:

image

Once compiled, we see that the imports can now be found:

image

# Test Plan

- [X] finder.rs unit tests and functional evidence screenshots from IDE


pyrefly/lib/module/finder.rs
```diff
@@ -39,11 +38,13 @@ enum FindResult {
...
-    /// The ordering of packages should be the same as the order they're found
-    /// in the `includes`.
-    RegularPackage(PathBuf, PathBuf),
+    /// Found regular packages. First field must point to an __init__ file.
+    /// Second field accumulates all roots in priority order. In typical
+    /// cases, the second field will be of length 1, containing only the
+    /// root of the first field. To support legacy pkgutil.extend_path-style
+    /// namespace packages, the second field also contains the remaining roots
+    /// to search for matches if the first does not contain a match.
+    RegularPackage(PathBuf, Vec1<PathBuf>),
```
connernilsen: If we're adding support for namespace package extend path functionality, why is that happening in `RegularPackage`? It seems like that would be more appropriately added to `NamespacePackage`.

The reason I ask is that by design, the *first* result for a `RegularPackage` needs to take priority over all other results. Anything we can find afterward breaks the design of how a `RegularPackage` works in Python, and can cause Pyrefly to fail to error in some cases.

We can make exceptions to this rule, but I think I need to better understand this approach before we can decide on that. It might be that another approach we might want to take would be to make a new `LegacyNamespacePackage(PathBuf, Vec1<PathBuf>)`, which has some of the information of a `RegularPackage`, but is treated as a `NamespacePackage` instead.

pyrefly/lib/module/finder.rs
```diff
@@ -205,16 +209,59 @@ fn find_one_part<'a>(
...
+                debug_assert_eq!(init_roots.len(), 1); // only one root: the one containing init_path
+                match &mut regular_package {
+                    None => {
+                        // First instance of this package found, so initialize the accumulator.
+                        regular_package = Some(FindResult::RegularPackage(init_path, init_roots))
+                    }
+                    Some(FindResult::RegularPackage(_, existing_roots)) => {
+                        // These __init__ files are lower-priority, so just accumulate the
+                        // new roots after the existing roots in case we need to look there.
+                        existing_roots.push(init_roots.into_vec().remove(0));
```
connernilsen: Why do we need to remove the first entry in `init_roots`?

connernilsen (commented):
Hey @mdbernard, thanks for doing this. I think we may still be discussing if/how we want to implement this, but I have a few questions I'm curious about. I haven't looked into this in a while, so my memory on how parts of this works are a little hazy. Let me know if any of my questions are worded poorly or don't represent what you're trying to achieve.

pyrefly/lib/module/finder.rs
mdbernard: Thanks for the review!

My understanding here is that `NamespacePackage` is a shorthand for "implicit namespace package" defined in [PEP 420](https://peps.python.org/pep-0420/). In the case of pkgutil-style packages, the namespaces aren't implicit, since they do use `__init__.py` files. Additionally, once pkgutil-style packages are installed, they look identical to regular packages in the site-packages folder.

For example, this pkgutil structure:
```
foo/
    foobar/
        __init__.py  # contains pkgutil.extend_path
        foo.py
bar/
    foobar/
        __init__.py  # contains pkgutil.extend_path
        bar.py
```

Ends up looking like this in site-packages when it's installed:
```
foobar/
    __init__.py
    foo.py
    bar.py
```

For both of these reasons, I thought it made more sense to treat these as `RegularPackage`s instead of `NamespacePackage`s, but I'm open to your suggestion about `LegacyNamespacePackage` being a distinct enum variant. Two questions though:

> [...] the first result for a `RegularPackage` needs to take priority over all other results. Anything we can find afterward breaks the design of how a `RegularPackage` works in Python, and can cause Pyrefly to fail to error in some cases.

Do you happen to have an example of how this implementation fails? It's a bit unclear to me what scenario you're concerned about. I'd be happy to add more unit tests. In this implementation, we search for matches in all the roots, and then return the match with the highest priority (e.g. `.pyi` > `.py` > `.pyc`) (see the new `test_namespace_package_submodule_prefers_pyi_over_py` unit test for an example). This can be pretty easily changed to just return the match from the first root, regardless of that priority order though.

> [...] we might want to take would be to make a new `LegacyNamespacePackage(PathBuf, Vec1<PathBuf>)`, which has some of the information of a `RegularPackage`, but is treated as a `NamespacePackage` instead.

Currently, pkgutil-style packages are treated as `RegularPackage`s by Pyrefly, since they look the same based on what the finder is looking for (dir name and `__init__.py` file presence). Do you have an idea of how you'd prefer this be done differently?

pyrefly/lib/module/finder.rs
mdbernard: The first one is the folder containing the `init_path`, so we don't want to search that one again.

mdbernard:
Thanks for the review @connernilsen, I've replied to your feedback. I'd also be happy to jump on Discord to chat through it if that's easier!

mdbernard:
@connernilsen just a friendly ping on this!

pyrefly/lib/module/finder.rs
connernilsen: Sorry for the pending word dump 😅

>My understanding here is that NamespacePackage is a shorthand for "implicit namespace package" defined in PEP 420. In the case of pkgutil-style packages, the namespaces aren't implicit, since they do use `__init__.py` files.

The original intention was for `NamespacePackage`s to handle all namespace package types, both implicit and explicit, since their behavior is semantically different from `RegularPackages`. I'll explain it as far as I understand it, but correct me if I've misinterpreted the documentation or my understanding differs from yours.

Both implicit and explicit namespace packages effectively enable Python to continue searching for results for a given package, whereas without being marked as a namespace package, the first result found is immediately returned.

Going back to the example you listed:
```
foo/
    foobar/
        __init__.py
        foo.py
bar/
    foobar/
        __init__.py
        bar.py
```
if our search path is `[foo/, bar/]`
* If the `foo` implementation of `foobar` is not a namespace package (therefore a regular package), then as soon as we find `foo/foobar/__init__.py`, `bar/foobar` is not findable as a package. `foo/foobar` is the only valid result for `from foobar import ...`
* If the `bar` implementation of `foobar` is not a namespace package but `foo/foobar` is, then likewise, `bar/foobar` is the only valid result for `from foobar import ...`
* If both `__init__.py` files are explicit namespace packages (both contain `pkgutil.extend_path`), then we can search for an import in either directory
  * Likewise if neither package has an `__init__.py`, making them both implicit namespace packages

That means, if we have something like this
```
foo/
    foobar/
        __init__.py  # not an explicit namespace package
        foo.py
```
we cannot fall back to looking at other directories afterward. There must always be exactly one entry in `<regular_package>.1`.

> Do you happen to have an example of how this implementation fails? It's a bit unclear to me what scenario you're concerned about.

An example I'm pretty sure will fail (but haven't directly tested myself, so would be good to add a test for) would be that given `search_path = [foo, bar]`, to make sure that:
- `foobar.foo` is importable as long as `bar/foobar` is not the only regular package 
  - i.e. `if (foobar.foo is regular) || (foobar.foo is namespace && foobar.bar is namespace)`
- `foobar.bar` is only importable when `foo/foobar` is not a regular package
  - i.e. `if (foobar.foo is not regular) || (foobar.foo is namespace && foobar.bar is namespace)`

> Currently, pkgutil-style packages are treated as RegularPackages by Pyrefly, since they look the same based on what the finder is looking for (dir name and __init__.py file presence). Do you have an idea of how you'd prefer this be done differently?

I think we'd want to do something similar. What we'd probably do is add logic to determine if an `__init__.py` contains something like `pkgutil.extend_path`, and if so, mark it as a `LegacyNamespacePackage`.

mdbernard: Ohhh thank you for clarifying. I dug around a bit more and found the packaging guide page for [namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/), and yes you're right that these should be treated as namespace packages.

> I think we'd want to do something similar. What we'd probably do is add logic to determine if an __init__.py contains something like pkgutil.extend_path, and if so, mark it as a LegacyNamespacePackage.

I agree with this approach. Would something like checking whether the \_\_init\_\_.py file contains the literal lines
```python
from pkgutil import extend_path
__path__ = extend_path(__path__, __name__)
```
sound ok to you? I'll add some additional unit tests to cover the behavior defined in the packaging guide.

connernilsen: Can users put other things in a legacy namespace package? 

It looks like `__path__ = __import__('pkgutil').extend_path(__path__, __name__)` is another recommended approach. Maybe we can do something like check if the file matches a regex like `__path__ = .*extend_path(__path__, __name__)`? It seems broad enough to catch most cases, but specific enough to determine if we're actually determine if we're looking at a namespace package.

Alternatively, we could parse the file into an AST and look for certain import nodes, but I don't think this logic has to be perfect

mdbernard: Agreed a search for a couple common patterns seems fine.

Are there any performance concerns about building an AST for every \_\_init\_\_.py file here? I would think just checking for those lines should cover pretty much all users.

Speaking from personal experience, yes plenty of people have abused this pattern and shoved a bunch of other stuff in these files haha

connernilsen: > Are there any performance concerns about building an AST for every __init__.py file here?

This can definitely be an issue. If it is a problem, we could stick the parsing behind a config option, but sounds like the couple of common patterns might be the way forward for now.

> Speaking from personal experience, yes plenty of people have abused this pattern and shoved a bunch of other stuff in these files haha

Of course lol 🤦‍♂️

connernilsen:
Sorry, will take a look soon!

mdbernard:
Nothing to look at yet! I pushed a change with what we discussed above, but I need to resolve conflicts with main and test it more still.

mdbernard:
Updated with the things we discussed re. searching for the extend_path regex! Also updated the PR description to accurately describe the change. @connernilsen

samwgoldman (approved):

connernilsen (approved):

PR #2787, opened 2026-03-12
fix Duplicate errors for undefined name in __all__ #2655

### LINKED ISSUE #2655: Duplicate errors for undefined name in `__all__`
### Describe the Bug

If a module's `__all__` attribute contains a name that pyrefly can't resolve, it reports a `bad-dunder-all` error on the `__all__` definition, then a `missing-module-attribute` error on a star-import of the module.

IMO we shouldn't report the latter: it may not be actionable (if you're importing from a third-party library, it's not easy to go in and fix `__all__`), and it's redundant with the `bad-dunder-all` error.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEArgC7Rx0Q3G6VN0BUIADQhmrEuUQgA%2BlNRQoMugF46AbQA6IAMKaAukJBlKMMFFKEmtKBQDEdAAqljp0nTRY8%2BOgGNc6SADmDJSoLH6E6uh2AMowMHQAFkxMxHCIAPTpRiZmhNwB6TDo6Zi43nDpvv4QQSFhxXRg3HSoAG6o0KjYsD5%2BgcGhEH50uMT1cBHoZEwJfgC0rTCUcEPoynSaAMyEAIwATJqRIAC%2BwqjeLIsAYtAwFO44BOInQA

_No response_

asukaminato0721:
# Summary

Fixes #2655

If a module explicitly declares `__all__` and that list contains a missing name, Pyrefly now reports the error on the defining module’s `__all__`, not again on every from foo import * importer.

The importer still gets the wildcard binding as Any(Error), but the duplicate Could not import ... from foo diagnostic is suppressed.

# Test Plan

update test

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -548,7 +548,7 @@ impl Bindings {
         builder.process_deferred_bound_names();
 
         // Validate that all entries in __all__ are defined in the module
-        for (range, name) in exports.invalid_dunder_all_entries(lookup, &module_info) {
+        for (range, name) in exports.invalid_dunder_all_entries(lookup) {
             builder.error(
                 range,
                 ErrorInfo::Kind(ErrorKind::BadDunderAll),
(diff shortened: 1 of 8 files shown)


yangdanny97:
sorry for the churn here, Kyle's review comments weren't exported here - I've patched some stuff on top to simplify things but the behavior is different from this impl, will try to reconcile it before the merge.

Issue #2711, opened 2026-03-08
TypeVar not constrained by default value (false negative


randolf-scholz:
### Describe the Bug

```python
class Getter[T]:
    def get[S=None](self, default: S = None) -> T | S: ...

def test(arg: Getter[str]) -> None:
    result: str = arg.get() 
```

```
ERROR sandbox.py:2:40-44: Default `None` is not assignable to parameter `default` with type `S` [bad-function-definition]
```

I believe this is both a false positive on line 2 and a false negative on line 5. See https://github.com/python/typing/issues/2213

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQDiMALizAE4DaAKgLqIAOunqj6mGGHoBzVlwDKAXgByudDD4AKODChgANOMmoArlBaJ68%2Bovqr1ASnoBaAHz0e9AD5XLhf6IAxPQA8gDSwsISUuxwLJqoHNKWzGycXHEcfE5udmowQiJiHPBmFvSZNvSJ0oSy8U70wQCilpU%2B9jD06Lgs1XQQ0uio2LD0LLgVLBwg%2BiAmLNBwJOSIIMEAqotQECyk9GAm6JSLanCRWJIHuBwAtqgsAProJrfYnJr4lhDoLDnumUKYnoJRYJg4IjAghAyle7w4lmA%2BAAvtDhLMQGQSmAoKRCBNblAKMEAAqkbG4ioYHAEeiUNSQaTgh4QNSEYTBeQwLoACzYxDgiAA9EKsZJcYQbtIhTB0ELMLhKHAhfT0Izmac5dcONUAG6oaAjMaq9UcFlqei4Yia5bCMgsHlqZy6zhwVkiWzQgDMhAAjF60egQMi5qgThAXQAxaAwChoLB4IhkYNAA

Adist319:
I can take a look at this!

grievejia:
This particular issue seems to have been fixed on trunk

PR #2788, opened 2026-03-12
Fix false positive bad-function-definition for TypeVar defaults

Adist319:
# Summary

`def get[S=None](self, default: S = None)` was emitting a false positive `bad-function-definition` because the subset check `None <: S` fails for abstract Quantified types. The fix specializes the annotation by substituting TypeVar defaults before checking the default expression.

Addresses #2711.

**Caveat:** The issue also mentions a false negative where `result: str = arg.get()` doesn't error even though `get()` returns `str | None` when called with no arguments. That's not fixed here. Pyrefly's hint-based instantiation pins `S=str` from the assignment target before argument matching runs, which overrides the TypeVar default. Fixing this would require pinning TypeVars with defaults for omitted parameters during argument matching, similar to what Pyright does. The [upstream typing spec discussion](https://github.com/python/typing/issues/2213) on function TypeVar default semantics is still open, so I wanted to wait on this first before moving forward (in a separate PR).

# Test Plan

- 4 new tests: reproducer, both `infer_with_first_use` modes, mismatched defaults (negative case), scope guard for unchanged behavior

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -839,9 +839,26 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             FunctionParameter::Annotated(idx) => {
                 // If the parameter is annotated, we check the default value against the annotation
                 let param_ty = self.get_idx(*idx).annotation.get_type().clone();
+                // For parameter default validation, substitute TypeVars that have
+                // defaults with their default types. This allows `def f[S=None](x: S = None)`
+                // to pass: the check becomes `None <: None` instead of `None <: S`.
+                let check_ty;
+                let check_ref = if default.is_some() {
+                    check_ty = param_ty.clone().transform(&mut |t| {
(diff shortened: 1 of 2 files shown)


yangdanny97:
I think the LLM analysis on the regressions is correct here

The psycopg example is here:

https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/connection.py#L73

the `.transform` call you added is substituting away `Row` in `RowFactory[Row]`, so now we have a mismatch between `RowFactory[tuple[Any, ...]]` v.s. `RowFactory[Row]`

I'm not sure what the non-hacky way to do this is. Maybe if the un-substituted type fails & there is a quantified w/ default somewhere in the type, we can check the substituted type & if either succeed we don't error?

yangdanny97 (changes requested):
Thanks for the PR! Back to you with a comment

rchen152:
I'm going to close this PR since it's stale. Please feel free to open a new one if you'd like to take another stab at this issue.

PR #2791, opened 2026-03-13
fix Upward Find for venv #759

### LINKED ISSUE #759: Upward Find for venv
I'm not 100% sure on whether this feature would be useful or not yet, but I'm thinking an upward find for venvs could be a good idea (instead of just looking in the same directory as your config file or project root marker)

asukaminato0721:
# Summary

Fixes #759

keeping the existing per-root scan and applying it across project_path plus its ancestors, so Pyrefly now finds the nearest ancestor .venv instead of only searching downward from the current root.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/pytorch_benchmark.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/pytorch_benchmark.rs
@@ -48,7 +48,7 @@ fn test_pytorch_error_propagation_latency() {
     };
     // Use all available cores for realistic benchmarking
     let mut interaction =
-        LspInteraction::new_with_args(args, NoTelemetry, Some(ThreadCount::AllThreads));
+        LspInteraction::new_with_args(args, NoTelemetry, Some(ThreadCount::AllThreads), None);
     interaction.set_root(pytorch_root.clone());
 
     interaction


crates/pyrefly_config/src/environment/venv.rs
```diff
@@ -14,6 +14,7 @@ const CONFIG_FILE: &str = "pyvenv.cfg";
 /// How deep within a project root should we attempt to search for a valid Python executable?
 /// 3 seems like a reasonable default to be able to find something in `.venv/bin/python3`.
 const SEARCH_DEPTH: usize = 3;
+const CANDIDATE_DIRS: &[&str] = &[".venv", "venv", ".env", "env"];
```
connernilsen: I don't think we want to use `.env` or `env` here, since people use that for other environment files and other things. If users are setting the name of their venv to `env`, I think they should have to explicitly configure it.

crates/pyrefly_config/src/environment/venv.rs
```diff
@@ -48,6 +49,28 @@ pub fn find(project_path: &Path) -> Option<PathBuf> {
...
+}
+
+fn search_roots(project_path: &Path) -> impl Iterator<Item = &Path> {
+    project_path
+        .ancestors()
+        .take_while(|path| !path.as_os_str().is_empty())
+}
+
+pub fn find(project_path: &Path) -> Option<PathBuf> {
+    search_roots(project_path).find_map(find_in_root)
```
connernilsen: Doing an upward filesystem walk to then do a recursive downward search can be very expensive, which is why I originally designed this to be very limited in scope. I have two suggestions for how to improve this, but I'm curious about your/others' thoughts.

1. instead of looking at all `project_path.ancestors()`, only look at ancestors where a candidate dir exists (something like `path.ancestors().flatten(|a| CANDIDATE_DIRS.iter().map(|c| a.join(c))).filter(|p| p.exists())`). This I think would be necessary at minimum.
2. stop the upward search at the first config file or marker file (`path.ancestors().take_while(|p| !ConfigFile::CONFIG_FILE_NAMES.chain(ConfigFile::ADDITIONAL_ROOT_FILE_NAMES).any(|c| p.join(c).exists()))`)

asukaminato0721: functionally that minimum is already implemented, though not in exactly that iterator shape.

find_in_root does:

take one ancestor root
build root/.venv and root/venv
require path.join("pyvenv.cfg").exists()
only then call the deeper interpreter walk inside that candidate venv

connernilsen: Oh cool. I see now, definitely missed that during the last review.

crates/pyrefly_config/src/environment/venv.rs
```diff
@@ -48,6 +49,28 @@ pub fn find(project_path: &Path) -> Option<PathBuf> {
...
+        .find_map(|path| find_in_env_dir(&path))
+}
+
+fn search_roots(project_path: &Path) -> impl Iterator<Item = &Path> {
+    project_path
+        .ancestors()
+        .take_while(|path| !path.as_os_str().is_empty())
+}
+
+pub fn find(project_path: &Path) -> Option<PathBuf> {
```
connernilsen, line 70: I think one more thing that might be nice is to do a wide search from the project path we start from. Basically attempt a `find_in_env_dir()` from the first path we're given, even if `CONFIG_FILE` doesn't exist. 

This will give us a more permissive policy for the directory containing a user's config file, so we can attempt to find non-standard venv names, even when it may be a lot more expensive to do at every parent directory level.

crates/pyrefly_config/src/environment/venv.rs
```diff
@@ -14,6 +14,7 @@ const CONFIG_FILE: &str = "pyvenv.cfg";
 /// How deep within a project root should we attempt to search for a valid Python executable?
 /// 3 seems like a reasonable default to be able to find something in `.venv/bin/python3`.
 const SEARCH_DEPTH: usize = 3;
+const CANDIDATE_DIRS: &[&str] = &[".venv", "venv"];
```
connernilsen: Sorry, now that I better understand what you were doing the first time by joining `CONFIG_FILE` to the candidate dirs., we can add back the `env` and `.env` entries here.

grievejia (changes requested):

PR #2792, opened 2026-03-13
fix False positive with @dataclass(frozen=True) and a ClassVar #2789

### LINKED ISSUE #2789: False positive with `@dataclass(frozen=True)` and a `ClassVar`
```python
import dataclasses
from typing import ClassVar

@dataclasses.dataclass(frozen=True)
class _Test:
    x: ClassVar[bool] = True
    def some(self) -> None:
        self.__class__.x = False

```

Produces:

```
» pyrefly check
 INFO Checking project configured at `pyproject.toml`
ERROR Cannot set field `x` [read-only]
  --> test.py:8:9
   |
23 |         self.__class__.x = False
   |         ^^^^^^^^^^^^^^^^
   |
  This field is a frozen dataclass member
```

However, this works in runtime.

asukaminato0721:
# Summary

Fixes #2789

stops ClassVar members from being classified as frozen dataclass fields.

preserves the existing ClassVar write restriction on instances, but no longer incorrectly rejects writes through a class object like `self.__class__.x = False` in a frozen dataclass.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2118,9 +2118,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         {
             return Some(ReadOnlyReason::NamedTuple);
         }
+        // ClassVars in dataclasses are class attributes rather than frozen instance fields.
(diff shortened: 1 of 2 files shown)


rchen152 (approved):
This looks great, thanks! I'll work on getting it merged.

PR #2800, opened 2026-03-14
fix False positive [missing-attribute] for dict.setdefault(key, []).append(val) on unannotated dict #2799

### LINKED ISSUE #2799: False positive [missing-attribute] for `dict.setdefault(key, []).append(val)` on unannotated dict
### Describe the Bug

When using `dict.setdefault(key, []).append(val)` on an unannotated empty dict, pyrefly reports:

> Object of class `NoneType` has no attribute `append`

pyrefly fails to resolve the return type of `setdefault()` — it should return the default value `[]` (a `list`) when the key is not present, but pyrefly infers `None` as a possible return type.

pyright and ty both accept this code without errors.

## Reproducer

```python
d = {}
items = [("news", "token1"), ("sports", "token2"), ("news", "token3")]

for topic, token in items:
    d.setdefault(topic, []).append(token)
```

## Expected behavior

No error. `setdefault()` returns the existing value or the provided default (the `[]` list), never `None` when a default is given.

## Note

Often a good solution here is to use `collections.defaultdict` - I'm adapting our code to use that. But still wanted to mention this here since I think it can be considered a false positive from `pyrefly`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImABALxXAC%2BAOuhAC4wC2ctVA2gAomIdDADucYQBoqwtrgDWMdAEZhAShlCQcYrgBObSSBlzFygEwatw0ROmyQ8pegDMGgLosWYA1XnEEADGMs7KVBDoERzciCxUCVSYhHAwbJgwYKgArlBsAgHBMnwe6oSoxMTKmAXm6OomINls0HAk5IggAMRUAKotUOykVGDZ6EEtuOiS6CwZYCMGnKhsAPro2ZzYMPoC%2BIgR6GzqVAC0AHxUcGz6cVGJ%2BmnZ%2BlFgwgBym9u39PjMICxGmRHmAoKRCPJOFAKD0AAqkEFgq4YHAEKhBKaQADmzxWECmhBYPQAyjAYFQABZsNjEOCIAD09OBmTBhAMWPpynpmFwQTg9Ix6GxuMm6Hpi30VFQADdUNBUNhYOjMRAcfo8VMqLhiKK2iwyGwKVNTtKdnB8VE6MJXIQVFYAegQAwpCBUBMIKaAGLQGAUNBYPBEMhOoA

_No response_

asukaminato0721:
# Summary

Fixes #2799

isolating failed overload attempts so they can’t mutate solver state and leak None into later matches.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -825,6 +825,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let mut overload_ctor_targs = ctor_targs.as_ref().map(|x| (**x).clone());
         let tparams = callable.0.as_deref();
 
+        let checkpoint = self.solver().checkpoint();
         let call_errors = self.error_collector();
         let (res, specialization_errors, expected_types) = self.callable_infer(
             callable.1.signature.clone(),
@@ -843,6 +844,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             hint,
             overload_ctor_targs.as_mut(),
         );
+        if !call_errors.is_empty() {
+            self.solver().restore_checkpoint(checkpoint);
+        }
 
         CalledOverload {
             func: callable,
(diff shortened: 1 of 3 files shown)


asukaminato0721:
```
INFO Checking project configured at `/home/runner/work/pyrefly/pyrefly/primer_base/projects/meson/pyproject.toml`
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/solver/solver.rs:450:24:
Internal error: a variable has leaked from one module to another.
```

need to investigate...

PR #2808, opened 2026-03-16
Add integration tests for stubgen with checked-in fixtures

yangdanny97:
Summary:
Add snapshot-based integration tests for the stubgen command. Each test
case has a checked-in input.py and expected.pyi pair under
pyrefly/lib/test/stubgen/. Test categories:

- functions: params, defaults, async, pos-only, kw-only, varargs
- classes: empty class, inheritance, methods, attributes
- variables: annotated module-level variables with values
- imports: preserved import statements
- mixed: type aliases, classes, and functions together

The test harness supports an update mode via STUBGEN_TEST_WRITE_PATH
env var for regenerating expected output after intentional changes.

Reviewed By: rchen152

Differential Revision: D95791749


rchen152 (approved):

PR #2813, opened 2026-03-17
chore(stubgen):` __all__` correctness

Rayahhhmed:
### Context
When a Python module defines `__all__`, it explicitly declares its public API. Before this change, pyrefly stubgen ignored `__all__` entirely and relied solely on naming conventions (skip _ prefixed names which is correct). 

However, this meant that internal helpers not in `__all__` leaked into the generated stub, and private names explicitly listed in `__all__` could be incorrectly filtered out.

For example, given:

```python3
__all__ = ["public_func", "PublicClass"]
def public_func(x: int) -> str: ...
def unlisted_func() -> bool: ...  # internal, not in __all__
def _explicitly_exported() -> None: ...  # private by convention but in __all__
```

Before: `unlisted_func` appeared in the stub (wrong), `_explicitly_exported` was excluded (wrong). 

After: Only `public_func` and `PublicClass` appear and this is exactly what `__all__` specifies.

### Approach
* Import statements are always preserved regardless of `__all__`, since they may be needed for type resolution in the stub.
* Type aliases (TypeVar calls, old-style aliases, PEP 695 type statements) are now also filtered through should_include_name, which they previously bypassed.
* Threads the result through ExtractionContext into `should_include_name()`, which short-circuits at module level: if `__all__` is present, only listed names pass. Inside classes, `__all__` is irrelevant and the existing convention logic applies unchanged.
* Falls back gracefully to convention-based filtering when `__all__` is absent, unresolvable, or contains cross-module re-exports.

### Test plan
- [x] New snapshot test: dunder_all
- [x] All 11 existing stubgen snapshot tests pass and no regressions for modules without
`__all__`
- [x] `python3 test.py --no-test --no-conformance` passes

### DIFF
--- a/pyrefly/lib/stubgen/extract.rs
+++ b/pyrefly/lib/stubgen/extract.rs
@@ -17,6 +17,7 @@ use std::sync::Arc;
 use pyrefly_build::handle::Handle;
 use pyrefly_python::module::Module;
 use pyrefly_python::short_identifier::ShortIdentifier;
+use pyrefly_python::sys_info::SysInfo;
 use pyrefly_types::callable::Param;
 use pyrefly_types::types::Type;
 use ruff_python_ast::Expr;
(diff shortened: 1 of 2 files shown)


yangdanny97:
@migeed-z why is mypy primer showing a huge delta here? this PR should be a no-op

migeed-z:
> @migeed-z why is mypy primer showing a huge delta here? this PR should be a no-op

The fix landed. It's looking normal again on new PRs.

Rayahhhmed:
Rebased on main again! Thanks @migeed-z :)

stroxler (approved):

PR #2816, opened 2026-03-17
fix Incorrect invalid-type-var check when persisting get method of a fully-annotated dict #2812

### LINKED ISSUE #2812: Incorrect `invalid-type-var` check when persisting get method of a fully-annotated `dict`
### Describe the Bug

Minimal repro:
```python
class ThemeStack:
    def __init__(self) -> None:
        self._entries: list[dict[str, str]] = [{}]
        self.get = self._entries[
            -1
        ].get  # pyrefly FP: Attribute `get` cannot depend on type variable `_T`
```
The error does not make much sense to me: `self._entries` has an explicit annotation and therefore `self._entries[-1]` should have a concrete dict type. Its `get` method shouldn't involve any typevar. 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQAqAFjALYwDKALqpQNaIAOunpj6mGGHoB9GRHQRucgBRwYUMAEp6AWgB89AHK50MYaPFX1mwjJjpuAJwjxE9KBDjcA2pgiUvt5OADT0wQC6EfQAvPQ%2BwAC%2BESJW1hpghADmMNyx4Rl2Ds6ucD6paZW6AIwVlRHZuWIAxPRkTlJQpPQAYgAK7gCC3CXYAK7cMPQABjnc0-SUGOi4eZLEDpj0pvTcpBv0AG6oLqjYsDMyjNMgISAT0HAk5IggrQCq3NBK3WBj6IEIKY4CIRJJpGBcE42KhlOgxmxsDAnCp8O4FNwdAZws4LFYOtwxk5RGAhCAjAikU53MB8IkySJbiB2p1SIRuLg2FAKK0%2BqQOmAuuEMDgCItTJAskTYUD0IQRK1ODApswRsQ4IgAPSalmCtlQrKahyazC4ShwTWUCUQKVOGWmTX0SFOeioY7QM4XK3oSXSr47XDEf3oJ4iMjcZimXSHZFwWX5MkAZkI1QATAz0CBEnc%2BF8Yz1oDAKGgsHgiGQs0A

_No response_

asukaminato0721:
# Summary

Fixes #2812

The cause was: invalid-type-var sanitization already exempted generic Forall methods, but it was not exempting generic overload signatures carried inside Overload and bound-method overloads. I added that handling there, so storing `dict[str, str].get` on an instance attribute no longer falsely reports leaked `_T`.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2466,6 +2466,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // (e.g. generic functions assigned to attributes have their own
         // type parameters that should not trigger invalid-type-var errors).
         let mut forall_bound: SmallSet<&Quantified> = SmallSet::new();
+        fn collect_overload_tparams<'a>(
+            overload: &'a Overload,
+            acc: &mut SmallSet<&'a Quantified>,
+        ) {
+            for sig in overload.signatures.iter() {
+                if let OverloadType::Forall(forall) = sig {
+                    for q in forall.tparams.iter() {
+                        acc.insert(q);
+                    }
+                }
+            }
+        }
         fn collect_forall_tparams<'a>(ty: &'a Type, acc: &mut SmallSet<&'a Quantified>) {
             match ty {
                 Type::Forall(forall) => {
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

PR #2822, opened 2026-03-17
Add snapshot changed notification to TSP

rchiodo:
# Summary

Add a snapshot changed notification to the TSP implementation in Pyrefly. Also fix some of the kind fields to be integers instead of strings as that matches typescript.

I believe this is the last change for Pyrefly to work inside of Pylance as a TSP provider. I ran a bunch of TSP tests using Pylance and it seems to work for all of them.

# Test Plan

Added a bunch of unit and integration tests.

/cc @kinto0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3559,6 +3559,7 @@ dependencies = [
  "lsp-server",
  "serde",
  "serde_json",
+ "serde_repr",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


rchiodo:
I'm not sure how the TSP enum change could affect the rest of Pyrefly? That's what the AI is suggesting.

migeed-z:
> I'm not sure how the TSP enum change could affect the rest of Pyrefly? That's what the AI is suggesting.

This is a bug. We are working through it!

PR #2823, opened 2026-03-17
Fix race condition causing false primer deltas (#2823)

migeed-z:
Summary:

mypy_primer runs old and new pyrefly concurrently via `asyncio.gather` in
the same project directory. The wrapper script (which runs `pyrefly init`
and installs deps via `setup_primer_deps.py`) was not safe for concurrent
execution: both instances would race on writing `pyrefly.toml` and
creating `.venv`, causing one run to get deps and the other to not, which produced massive deltas. We fix the race condition by adding a lock to the wrapper script.

Reviewed By: yangdanny97

Differential Revision: D97007184


yangdanny97 (approved):

Issue #2621, opened 2026-03-02
Handling dual-use decorator


grievejia:
### Describe the Bug

Pyrefly reports "Expected N positional arguments, got M" when a dual-use decorator (usable with or without parentheses) is applied without parentheses.

Minimal repro:
```python
from functools import wraps

# `optional_debug` can be used as @optional_debug or @optional_debug(flag).
# When used bare (@optional_debug), pyrefly infers that `compute` has the
# type of the inner `decorator` function (which takes 1 arg), not the
# original `compute` (which takes 3 args).
def optional_debug(func_or_flag=None):
    """Can be used as @optional_debug or @optional_debug(True)."""

    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            return func(*args, **kwargs)

        return wrapper

    if callable(func_or_flag):
        return decorator(func_or_flag)
    return decorator

@optional_debug
def compute(x: int, y: int, z: int) -> int:
    return x + y + z

r1 = compute(1, 2, 3)  # Pyrefly reports "Expected 1 positional argument, got 3" 
```
Technically speaking, Pyrefly is not wrong: the inferred return type for the decorator is a union type, and applying that union to an arg list that only satisfy one branch of that union is no doubt a type error, e.g.
```python
from typing import Callable
def test(f: Callable[[], int] | Callable[[int], int]) -> int:
    return f(42)  # This should be a type error
```

That said, many libraries use dual-use decorators (click, celery, flask) in practice, and custom decorators in application code frequently follow this pattern. Do we want to somehow add special-casing to support this code pattern better?

### Sandbox Link

[...]

PR #2824, opened 2026-03-18
Dual-use decorator return unions alternative heuristic (#2824)

grievejia:
Summary:

Dual-use decorators often infer a return type that is the union of the final wrapped callable and the leftover decorator factory branch. Preserving that raw union is not useful at decoration sites: every use of the decorated symbol then has to satisfy both branches, even though only the wrapped callable branch can actually represent the final decorated value.

This change normalizes decorator return unions in `apply_function_decorator` before we finalize the decorated type. When a union member still looks like a decorator factory, meaning it accepts the decoratee as its first argument and itself returns another callable, we treat that member as an artifact of the dual-use control flow and prune it away. If pruning leaves a single unannotated passthrough callable, we preserve the original decoratee type; otherwise we keep the remaining normalized union.

This keeps the fix local to decorator application instead of adding a broad type-level heuristic for all `(*args, **kwargs) -> Any` callables. That makes the behavior match the motivating dual-use decorator pattern without changing unrelated callable typing.

NOTE: The other alternative approach to this is D96672235

Differential Revision: D97025245

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -73,6 +73,7 @@ use crate::types::callable::PropertyRole;
 use crate::types::callable::Required;
 use crate::types::class::ClassKind;
 use crate::types::keywords::DataclassTransformMetadata;
+use crate::types::types::AnyStyle;
 use crate::types::types::CalleeKind;
 use crate::types::types::Forall;
 use crate::types::types::Forallable;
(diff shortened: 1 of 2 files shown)


rchen152:
Going to close this since https://github.com/facebook/pyrefly/issues/2621 looks like it's been resolved.

PR #2849, opened 2026-03-22
fix Narrowing inside try is preserved into finally, ignoring type outside of try #2845

### LINKED ISSUE #2845: Narrowing inside `try` is preserved into `finally`, ignoring type outside of `try`
### Describe the Bug

```python
from typing import *


def something_that_might_throw() -> None:
    raise Exception()
        
class Thing():
    def something(self) -> None:
        pass

def blah() -> None:
    x = None
    try:
        something_that_might_throw()
        if x is None:
            x = Thing()
    except Exception:
        raise
    finally:
        reveal_type(x) # revealed type: Thing, but it could be `None`
        x.something()
```

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRIzDDB04tGAwAWrNgH0FqBspoQ28jQuoB3ABQBKOgFoAfHQByudDEQi6LupVQQ4MOgFF8AYxhiBgh7U2dXSJF-KFQ4ODoAFUV2Uyd0SIkpGRo5FLYjLygwMytbe0cIyJdiOLgxLEk6bFj5UwtrOwd0yPw6AF5yhyrGSlIe6pccvKVVeXVNbV05w3CMyc4pPs8hyvWN1z7B5KU1yJgAoP4-QODQ9AnJ908YEch0VCgocZHIyhgAG4wT6qZgwIz4MwAYjcgOBsEwjDBiCS%2BQANM0AK78CD8fy4TFQRHYbwAAy6MFJv0OhGmClOJhAaJAZH%2BYG%2BhAYtCgFBhAAVSGzvtIMDgCHR8ehIGxMe4QvZCCIYQBlGDeXQMYhwRAAeh1rMkHN4bB1MHQOswuH8cB1kulsvU9x1dDAvDoqABHliLW8du0DvlGVwd3scEV6DICns5iBlDg9wGdCEIAAzIQAIwAJmTIgA2jBKNQ4wBdESY9BcHh8GCYcyYCD-fwhIGJgDk2nQvBgrZEnYY5n%2BAEdMQ2a%2BYANYwUjmVD%2BQLxNsGVCUdA99AgAC%2BzNnzZgADFoDAKGgsHgiGRN0A

I'm not sure what the desired behavior is, but clearly we're guessing wrong in this case by taking the refined type from the `try`.

If we take the original type before the `try` then it doesn't account for the fact that the `try` could have succeeded. I thought we created a flow branch at the beginning of the try-block and merging the two branches should have handled this, but something is going wrong.

asukaminato0721:
# Summary

Fixes #2845

try statements now use a dedicated pre-finally merge so branches that have already terminated still contribute type state while the finally body is analyzed, instead of being dropped too early.

but mypy primer seems not so good.

# Test Plan

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -917,7 +917,8 @@ impl FlowStyle {
                         MergeStyle::Loop
                         | MergeStyle::LoopDefinitelyRuns
                         | MergeStyle::Exclusive
-                        | MergeStyle::Inclusive => FlowStyle::PossiblyUninitialized,
+                        | MergeStyle::Inclusive
+                        | MergeStyle::Finally => FlowStyle::PossiblyUninitialized,
                     }
                 }
             }
@@ -3336,6 +3337,10 @@ enum MergeStyle {
     /// Distinct from [Branching] because we have to be more lax about
     /// uninitialized locals (see `FlowStyle::merge` for details).
     BoolOp,
+    /// This is the merge immediately before analyzing a `finally` block.
+    /// Terminating branches still contribute because `finally` executes before
+    /// their control-flow effect is observed outside the statement.
+    Finally,
 }
 
 impl MergeStyle {
@@ -3688,17 +3693,20 @@ impl<'a> BindingsBuilder<'a> {
             return;
         }
 
+        let use_all_branches = matches!(merge_style, MergeStyle::Finally);
         // We normally only merge the live branches (where control flow is not
         // known to terminate), but if nothing is live we still need to fill in
         // the Phi keys and potentially analyze downstream code, so in that case
         // we'll use the terminated branches.
-        let (terminated_branches, live_branches): (Vec<_>, Vec<_>) =
-            branches.into_iter().partition(|flow| flow.has_terminated);
-        let has_terminated = live_branches.is_empty() && !merge_style.is_loop();
-        let flows = if has_terminated {
-            terminated_branches
+        let n_live_branches = branches.iter().filter(|flow| !flow.has_terminated).count();
+        let has_terminated = n_live_branches == 0 && !merge_style.is_loop();
+        let flows = if use_all_branches || has_terminated {
+            branches
         } else {
-            live_branches
+            branches
+                .into_iter()
+                .filter(|flow| !flow.has_terminated)
+                .collect()
         };
         // Determine reachability of the merged flow.
         // For Loop style with empty flows (all branches terminated), the loop body might
@@ -3710,6 +3718,11 @@ impl<'a> BindingsBuilder<'a> {
                 MergeStyle::Loop => base.is_definitely_unreachable,
                 _ => true,
             }
+        } else if use_all_branches && n_live_branches > 0 {
+            flows
+                .iter()
+                .filter(|f| !f.has_terminated)
+                .all(|f| f.is_definitely_unreachable)
         } else {
             flows.iter().all(|f| f.is_definitely_unreachable)
         };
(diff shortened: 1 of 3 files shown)


NathanTempest:
Sorry for the five-month silence @asukaminato0721, you flagged the primer
yourself on day one and nobody followed up. Here's the root cause.

The regression is one pattern, and it traces to the exact line this PR edits:

1. `binding/scope.rs:944`  adding `MergeStyle::Finally` to this arm means any
   name absent from *any* merged branch becomes `FlowStyle::PossiblyUninitialized`
2. `binding/scope.rs:783` `PossiblyUninitialized => InitializedInFlow::Conditionally`
3. `binding/bindings.rs:181`  `Conditionally => "`{name}` may be uninitialized"`

With `use_all_branches` now pulling *terminated* branches into that merge  where
the name genuinely was never bound every instance of this shape breaks:

```python
try:
    proc = subprocess.Popen(...)
except OSError as exc:
    raise InstallationError(...) from exc
use(proc)          # <- "`proc` may be uninitialized"

**Comment by NathanTempest** [comment]
Following up on the root cause: the reason this reaches code *after* the `try`
is structural, not a bad constant.

`binding/stmt.rs:1534-1537`:

```rust
self.finish_exhaustive_fork();   // -> finish_finally_fork()
self.scopes.enter_finally();
self.stmts(x.finalbody, parent);
self.scopes.exit_finally();
```
The merged flow is the current flow for the finally body and stays current after exit_finally(), so it becomes the state for everything following the statement. enter_finally/exit_finally (scope.rs:1705-1712) only bump a finally_depth counter for PEP 765 they don't save or restore flow.

Worth saying that PossiblyUninitialized is correct inside finally:
```
try:
    x = compute()     # can raise
finally:
    print(x)          # genuinely may be unbound
```

It's only wrong once it leaks out. So the fix is two merges, not one merge style: all branches (incl. terminated) for the finally body's entry state, live branches only for the continuation and then install the second with the finally body's bindings applied on top.

NathanTempest:
Out of the three copilot reponses from March, two are correct:

- **[#1, single-branch short-circuit](https://github.com/facebook/pyrefly/pull/2849/changes#r2977480047)
  — correct, verified.** `merge_flow` returns early at `scope.rs:3899`
  (`if !merge_style.is_loop() && branches.len() == 1`). A bare `try/finally` with
  no handlers produces exactly one branch — `start_fork_and_branch`, body +
  orelse, `finish_branch`, handler loop never runs (`stmt.rs:1470-1531`). So
  `MergeStyle::Finally` is **never reached** for `try/finally` without `except`,
  which is the canonical shape of this bug.
- **[#3, missing coverage](https://github.com/facebook/pyrefly/pull/2849/changes#r2977480111)
  — correct**, and follows from #1: the untested case is the one that doesn't work.
- **[#2](https://github.com/facebook/pyrefly/pull/2849/changes#r2977480085)** — I'd skip. It patches one symptom the same root cause and on its own would mark the post-`try` flow reachable when it shouldn't be.

The new testcase checks `assert_type(x, Thing | None)` inside `finally`, which is
the right target but nothing asserts the state *after* the statement, and
that's what regressed. This would have gone red on the first CI run:

```python
def f() -> None:
    try:
        x = Thing()
    except Exception:
        raise
    x.something()   # must NOT error - the except branch never reaches here
```
@stroxler @rchen152   Could one of you confirm the two-merge shape is the direction you'd want before @asukaminato0721 invests in the restructure? #2845 is a real bug; it just needs more than a new MergeStyle.

rchen152 (changes requested):
I don't think it's possible to merge this with this many new mypy_primer errors.

rchen152:
Sorry, I realized I read too quickly and didn't see the question addressed to me!

Looking at the reported bug, this seems pretty tricky to fix without catching a lot of false positives, given that it basically requires assuming that every line in a try can throw an exception. My feeling is that that's too pedantic to be particularly useful in practice, so we'd probably have to add some exceptions for various cases in which we want to relax the rule. I'm not positive we want to pursue this direction unless there's a clear payoff (it doesn't have to be a *big* payoff, but I'd hope that we can show we report some actionable errors in mypy primer or are fulfilling a user request or something).

NathanTempest:
Hey @asukaminato0721 closing this PR for now as the primer diff is pretty hard to fix with the current direction. We highly appreciate your persistence in making Pyrefly a better tool for everyone!

PR #2887, opened 2026-03-24
Emit deprecated tag for deprecated LSP diagnostics

### LINKED ISSUE #2885: Emit deprecated tag for `deprecated` LSP diagnostics
### Describe the Bug

For the `deprecated` error code, we should emit LSP diagnostics w/ the deprecated tag: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag

That way, the editor can render a strikethrough over the function name. We should be careful what position we emit the diagnostic in. I think the strikethrough should only cover the function name, not the arguments.

### Sandbox Link

_No response_

_No response_

mvanhorn:
Emits DiagnosticTag::DEPRECATED on LSP diagnostics for deprecated error codes.

## Why this matters

Editors that support LSP diagnostic tags render strikethrough on deprecated names. Pyrefly was missing this tag. Now deprecated functions get the visual cue.

## Changes

- **pyrefly/lib/error/error.rs**: Added DiagnosticTag import. Changed `tags: None` to emit `DiagnosticTag::DEPRECATED` when `error_kind() == ErrorKind::Deprecated`.

## Testing

The change is a 6-line addition to a single file. ErrorKind derives PartialEq so the comparison is sound.

Fixes #2885

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
@@ -1672,3 +1672,54 @@ fn test_no_diagnostics_for_non_open_files_in_open_files_only_mode() {
         non_open_file.display()
     );
 }
+
+#[test]
+fn test_deprecated_diagnostic_tag() {
+    let test_files_root = get_test_files_root();
+    let mut interaction = LspInteraction::new();
+    interaction.set_root(test_files_root.path().to_path_buf());
+    interaction
+        .initialize(InitializeSettings {
+            configuration: Some(None),
+            ..Default::default()
+        })
+        .unwrap();
+
+    interaction.client.did_change_configuration();
+
+    interaction
+        .client
+        .expect_configuration_request(None)
+        .unwrap()
+        .send_configuration_response(json!([
+            {"pyrefly": {"displayTypeErrors": "force-on"}}
+        ]));
+
+    interaction.client.did_open("deprecated_function.py");
+
+    interaction
+        .client
+        .diagnostic("deprecated_function.py")
+        .expect_response(json!({
+            "items": [
+                {
+                    "code": "deprecated",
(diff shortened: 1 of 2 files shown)


yangdanny97:
This should be tested somewhere. Did you verify the diagnostic ranges are correct for all the locations where this error is emitted?

mvanhorn:
Added an LSP interaction test in 6a21eb5 that opens a file with a `@deprecated` function call and asserts the diagnostic includes `tags: [2]` (DEPRECATED).

On range correctness: the tag is applied in `to_diagnostic()` based on `error_kind() == ErrorKind::Deprecated`, so it's independent of which call site emits the error. All four sites (expr.rs, call.rs, overload.rs, stmt.rs) produce `ErrorKind::Deprecated` errors that flow through the same `to_diagnostic()` path. The range for each comes from the AST node at the call site (call expression, arguments, import statement), which is the same range used for the error message.

yangdanny97:
To clarify: I'm saying we should emit the Deprecated error at the position of the call expression's subject, not the whole expression. That way, if we have something like `deprecated_function(a, b, c)` it's rendered as ~~deprecated_function~~(a, b, c) rather than ~~deprecated_function(a, b, c)~~

mvanhorn:
Looking at the call paths, `freeform_call_infer` already passes `x.func.range()` (the callee range) into `as_call_target_or_error`, so the deprecated diagnostic should already use just the function name for direct calls. The test shows range `0:12` for `old_function()` which is only the name, not the full call expression.

Is there a specific case where you're seeing the range cover the arguments?

yangdanny97:
Understood, I'll try to get this merged. thanks

mvanhorn:
Thanks for pulling this in, @yangdanny97. Appreciate the back-and-forth on the diagnostic ranges.

PR #2891, opened 2026-03-25
fix “Decorator @X can only be used on methods” is too strict #2610

### LINKED ISSUE #2610: “Decorator @X can only be used on methods” is too strict
### Describe the Bug

Pyrefly reports "Decorator @X can only be used on methods" when `@property`, `@classmethod`, or `@staticmethod` are applied to functions outside a class body. This is valid Python -- these descriptors work correctly when the resulting objects are later assigned to class attributes.

Real-world: Django uses this pattern extensively in `ModelAdmin` and `FormSet` classes. Libraries like `attrs` and `SQLAlchemy` also create properties outside class bodies via factory functions.

```python
# Pattern 1: @property factory function
#
# A helper creates a property descriptor, which is later assigned to a class.
# This is common when multiple classes share the same computed attributes.
def make_property(attr_name: str):
    @property
    def prop(self):
        return getattr(self, "_" + attr_name)

    return prop


class Point:
    def __init__(self, x: int, y: int):
        self._x = x
        self._y = y

    x = make_property("x")
    y = make_property("y")


p = Point(1, 2)
assert p.x == 1
assert p.y == 2


# Pattern 2: @staticmethod outside class body
#
# A utility function decorated with @staticmethod at module level,
# then assigned as a class attribute.
@staticmethod
def utility(x: int) -> int:
    return x * 2


class Helper:
    compute = utility


result = Helper.compute(5)


# Pattern 3: @classmethod outside class body
#
# A classmethod created at module level, then assigned to a class.
@classmethod
def from_value(cls, val: int) -> "Container":
    obj = cls()
    obj.value = val
    return obj


class Container:
    value: int = 0
    from_value = from_value  # type: ignore[assignment]


c = Container.from_value(42)
```
The easiest way to address this issue might be to lower the default severity of the problem? 

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2610

top-level `@staticmethod`, `@classmethod`, `@property`, and `@cached_property` are no longer treated as invalid just because they appear outside a class body.

# Test Plan

update test

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/pytorch_benchmark.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/pytorch_benchmark.rs
@@ -48,7 +48,7 @@ fn test_pytorch_error_propagation_latency() {
     };
     // Use all available cores for realistic benchmarking
     let mut interaction =
-        LspInteraction::new_with_args(args, NoTelemetry, Some(ThreadCount::AllThreads));
+        LspInteraction::new_with_args(args, NoTelemetry, Some(ThreadCount::AllThreads), None);
     interaction.set_root(pytorch_root.clone());
 
     interaction


rchen152 (approved):

yangdanny97:
this was merged

PR #2907, opened 2026-03-26
fix Reference Counting #2194

### LINKED ISSUE #2194: Reference Counting
### Describe the Bug

> For some languages ​​, we have references mentioned above of class/function with the number of references that the object is used.
>
> But for Python, I can't find it. This could be a good feature for Python.

https://github.com/microsoft/pylance-release/issues/6660

Image

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2194

The server now advertises `textDocument/codeLens` when indexing is enabled and returns CodeLens entries for class/function/method definitions using the existing global references engine.

Each lens shows N reference(s) and wires to VS Code’s built-in `editor.action.showReferences`, so clicking the count opens the references UI.

# Test Plan

add LSP interaction test


crates/pyrefly_lsp_test/src/object_model.rs
```diff
@@ -538,6 +539,15 @@ impl TestClient {
         }))
     }
 
+    pub fn code_lens(&self, file: &'static str) -> ClientRequestHandle<'_, CodeLensRequest> {
```
kinto0: can all this code lens logic be in its own file and not server.rs?

pyrefly/lib/lsp/non_wasm/workspace.rs
```diff
@@ -284,6 +284,8 @@ pub struct DisabledLanguageServices {
     #[serde(default)]
     pub references: bool,
     #[serde(default)]
+    pub code_lens: bool,
```
kinto0: any chance we can avoid these unrelated changes?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -5024,6 +5095,151 @@ impl Server {
...
     }
 
+    fn code_lens<'a>(
+        &'a self,
+        request_id: RequestId,
+        transaction: &Transaction<'a>,
+        params: CodeLensParams,
+        activity_key: Option<ActivityKey>,
+    ) -> Result<(), EmptyResponseReason> {
+        let uri = &params.text_document.uri;
```
kinto0: let's try splitting this function up and moving it into another module

kinto0 (changes requested):

pyrefly/lib/lsp/non_wasm/server/code_lens.rs
```diff
@@ -0,0 +1,475 @@
...
+            .ok_or(EmptyResponseReason::ModuleInfoNotFound)?;
+        let symbols = transaction
+            .symbols(handle, None)
+            .ok_or(EmptyResponseReason::ModuleInfoNotFound)?;
+        let mut symbol_defs = Vec::new();
+        collect_reference_symbols(&symbols, &mut symbol_defs);
+
+        let mut seen = SmallSet::new();
+        let mut targets = Vec::new();
+        for symbol in symbol_defs {
```
kinto0: I wonder how long this takes on large (10k+ LoC) files?

I wonder if we could only calculate these for visible targets in the IDE. it seems like we can do that with a `codeLens/resolve` resolve message

pyrefly/lib/lsp/non_wasm/server/code_lens.rs
```diff
@@ -0,0 +1,475 @@
...
+                ]),
+                tooltip: None,
+            }),
+            data: None,
+        });
+    }
+    Ok(lenses)
+}
+
+fn runnable_lsp_code_lens(
```
kinto0: we have `runnable_code_lens` and `runnable_lsp_code_lens` in this file. can we organize this into pieces that have clear boundaries with docstrings?

kinto0 (changes requested):

pyrefly/lib/lsp/non_wasm/server/code_lens.rs
```diff
@@ -0,0 +1,475 @@
...
+    /// Build runnable lenses and resolve reference counts on the references queue.
+    pub(super) fn code_lens<'a>(
+        &'a self,
+        request_id: RequestId,
+        transaction: &Transaction<'a>,
+        params: CodeLensParams,
+        activity_key: Option<ActivityKey>,
+    ) -> Result<(), EmptyResponseReason> {
+        let uri = &params.text_document.uri;
+        if self.open_notebook_cells.read().contains_key(uri) {
```
kinto0, line 41: this limits requests within notebooks (this is ok but at some point we should refactor this so the type system can guarantee this is not a notebook handle) but this code fails to prevent references within a notebook to this symbol (defined within a .py file). looks like other places handle that with this:

  if let Some(cell_idx) = info.to_cell_for_lsp(range.start())
      && let Some(path) = to_real_path(info.path())
      && let Some(notebook) = open_notebooks.get(&path)
      && let Some(cell_url) = notebook.get_code_cell_url(cell_idx)
  {
      uri = cell_url.clone();
  }

maybe it's worth making a reusable system for handling this. (looks like `to_lsp_location` might do this?)

can we add a test for both of these cases (references from and to a notebook are blocked) so we don't regress if we refactor this?

PR #2908, opened 2026-03-26
fix(lsp): emit DiagnosticTag::DEPRECATED for deprecated errors

### LINKED ISSUE #2885: Emit deprecated tag for `deprecated` LSP diagnostics

LeC-D:
## Summary

Fixes #2885

When Pyrefly emits a `deprecated` error (e.g. calling a function decorated with `@deprecated`), the LSP diagnostic now includes `DiagnosticTag::DEPRECATED` (tag value 2 per the LSP 3.17 spec). This allows editors to render a strikethrough over the deprecated symbol, consistent with how other type checkers like Pyright behave.

### What changed

In `pyrefly/lib/error/error.rs`, the `to_diagnostic()` method previously always set `tags: None`. It now sets `tags: Some(vec![DiagnosticTag::DEPRECATED])` when `error_kind == ErrorKind::Deprecated`.

The diagnostic range for deprecated calls was already correct: call-site errors pass `x.func.range()` (the callee name only) to `as_call_target_or_error`, so the strikethrough covers only the function name and not the argument list.

### Tests

Two unit tests added to `error::error::tests`:
- `test_deprecated_diagnostic_has_deprecated_tag`: verifies that a `Deprecated` error emits `DiagnosticTag::DEPRECATED`
- `test_non_deprecated_diagnostic_has_no_tag`: verifies that unrelated errors still produce no tags

## Test Plan

```
cargo test -p pyrefly --lib error::tests
```

All 4 tests pass (2 existing + 2 new).

### DIFF
--- a/pyrefly/lib/error/error.rs
+++ b/pyrefly/lib/error/error.rs
@@ -14,6 +14,7 @@ use std::path::Path;
 use itertools::Itertools;
 use lsp_types::CodeDescription;
 use lsp_types::Diagnostic;
+use lsp_types::DiagnosticTag;
 use lsp_types::Url;
 use pyrefly_python::ignore::Tool;
 use pyrefly_python::module::Module;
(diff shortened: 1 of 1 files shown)


yangdanny97:
someone else already made a PR for this, it was just merged

PR #2934, opened 2026-03-26
Fix noisy label timeline events in V1 ranking workflow

migeed-z:
## Summary

The V1 analysis workflow (`issue_ranking.yml`) was using a "clean slate" approach for label management: remove ALL managed labels from ALL issues, then re-apply them from scratch. This generated redundant add/remove timeline events on **every run**, even when labels hadn't changed, cluttering issue timelines with noise like:

> github-actions added `v1-verified`
> github-actions removed `v1-verified`  
> github-actions added `v1-verified`

See https://github.com/facebook/pyrefly/issues/1422 for an example.

**Fix:** Replace the clean-slate approach with diff-based label updates:
1. Fetch the current managed label state across all issues
2. Compute the desired label state from the analysis
3. Only add/remove labels where the state actually differs
4. Leave already-correct labels untouched (no API calls = no timeline events)

## Changes

- `v1_analysis.py`: Replaced `_cleanup_managed_labels()` with `_get_current_managed_labels()`. Rewrote `apply_labels()` to diff current vs desired state.
- `test_issue_ranker.py`: Updated tests, including `test_skips_already_correct_labels` which verifies zero API calls when labels are already correct.

## Test Plan

- `pytest -k "TestGetCurrentManagedLabels or TestApplyLabels or TestRunV1Analysis"` — 10/10 pass


yangdanny97 (approved):

Issue #2831, opened 2026-03-19
`@override` assignability not enforced for `__init__`


randolf-scholz:
### Describe the Bug

```python
from typing import override

class Foo:
    def __init__(self, x: int) -> None: ...
    
    
class Bar(Foo):
    @override
    def __init__(self, y: int) -> None: ...  # no error
```

Typically type checkers do no check that subclass `__init__` is assignable to superclass `__init__`. However, this check should happen when the method is explicitly annotated with `@override` because according to the [typing spec](https://typing.python.org/en/latest/spec/class-compat.html#override):

> When type checkers encounter a method decorated with @typing.override they should treat it as a type error unless that method is overriding a method or attribute in some ancestor class, **and the type of the overriding method is [assignable](https://typing.python.org/en/latest/spec/glossary.html#term-assignable) to the type of the overridden method.**

x-ref: https://github.com/python/typing/issues/2222

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN24AbjEqUImGAB100gMZRUcOHQBiuXIml1tdCWDoB9A6wgMjACjgwoYADR18iTugYBKOgFoAfHQByudBgnQhDpOQUlOgAhVEpzNVxXTXQdOgABIRExCS0dPUNjdFMLKxt7UidWN08ff0Dg0JSdEFsQMkoYMChSQgZaKAoAYjoABVIOrtI6NCw8fDpZAMg2AFdKVAYIAMJpYYBlGBg6AAsGBmI4RAB6K-bO7sJeNiuYdCvMXFk4K8X0ZbWNls3nQwLw6KhBKhoKhsLAFksIKt1psAgJiCj0HAdugyAxjgEPMJKHAgXQALx0SQgADMhAAjAAmKnSADaWV4cAAutIVkVuLwGDBMB5MBAOrJNsJyXQAOSI9C8GAy6QKhgeDoARxWYqFHgA1jBSB5ULJZPBlBSZQB3WLoZXoEAAX1aJslMBU0BgFBmOAIJHIjqAA

kshitijgetsac:
hey @rchen152 can I try to work on this issue and if yes is the expected behavior in this scenario to produce an error while instantiating the base class or something else

yangdanny97:
assigned, thank you. let us know if you have any questions!

[...]

PR #2936, opened 2026-03-26
added override check for init method

kshitijgetsac:
- Previously, pyrefly unconditionally skipped override consistency checks for __init__, __new__, and
  __init_subclass__. This is correct by default, but the typing spec requires that when @override is explicitly
  applied, the type checker must verify the overriding method's type is assignable to the overridden method's
  type.
  - Added an is_override parameter to should_check_field_for_override_consistency so that the skip is only
  applied when @override is not present.

  Test plan

  - Added test_override_init_with_decorator: @override + incompatible signature produces error
  - Added test_override_init_without_decorator: no decorator + incompatible signature produces no error (existing
   behavior preserved)
  - Added test_override_init_compatible: @override + compatible signature produces no error
  - All existing class_overrides tests pass (84 total)


kshitijgetsac:
solves #2831

PR #2937, opened 2026-03-26
fix No narrowing for sequence patterns #2857

### LINKED ISSUE #2857: No narrowing for sequence patterns
### Describe the Bug

Repro:

```

@dataclass
class A: ...
@dataclass
class B: ...


def test(x: A | B, y: A | B):
  match x, y:
    case A(), B():
      reveal_type(y)
```

Output:

```
  --> /tmp/foo.py:13:18
   |
13 |       reveal_type(y)
   |                  ---
   |
```

I suspect you need to special-case subjects/sequence patterns which have statically-known arity?

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2857

now recover per-element subjects from tuple/list match subjects and feed them into fixed-arity sequence-pattern binding, so `match x, y: case A(), B():` narrows x to A and y to B inside the branch.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/binding/pattern.rs
+++ b/pyrefly/lib/binding/pattern.rs
@@ -44,10 +44,24 @@ use crate::export::special::SpecialExport;
 use crate::types::facet::UnresolvedFacetKind;
 
 impl<'a> BindingsBuilder<'a> {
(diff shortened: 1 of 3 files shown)


yangdanny97:
oops, i think this was handled by https://github.com/facebook/pyrefly/commit/1e6ac2558e29fdac2c70e4bbfd8afd8a4525881c

PR #2949, opened 2026-03-27
Fix some bugs found in the TSP implementation while testing with Pylance

rchiodo:
# Summary

Fix some problems found while testing Pyrefly with full Pylance test suite. There were issues with finding declarations and with type flags. 

# Test Plan

Ran unit tests in Pyrefly and ran external tests in Pylance. This work is a precursor to getting a nightly test run for Pylance using TSP.

/cc @kinto0

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -243,7 +243,6 @@ use crate::commands::lsp::IndexingMode;
 use crate::config::config::ConfigFile;
 use crate::error::error::Error;
 use crate::lsp::module_helpers::to_real_path;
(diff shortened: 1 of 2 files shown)


kinto0:
curious how these are tested? does pylance have it's own test suite for this? or should we add some to pyrefly too?

rchiodo:
> curious how these are tested? does pylance have it's own test suite for this? or should we add some to pyrefly too?

Yeah Pylance has a test suite I wrote for Ty that I modified to work for Pyrefly.

I hope to move some of those tests to the https://github.com/microsoft/python-lsp-compare after you ship tsp bits?

yangdanny97 (approved):

PR #2954, opened 2026-03-27
Fix GitHub build

rchen152:
Summary:
pytorch_benchmark.rs is gated with #[cfg(not(fbcode_build))] (line 14), so it's only compiled on GitHub/cargo builds, never internally.

When the thrift_remapper parameter was added to LspInteraction::new_with_args, this call site was missed because it doesn't compile in the internal build.

Fix: Added None as the 4th argument (thrift_remapper) to match the updated function signature, consistent with all other callers.

Differential Revision: D98568416


maggiemoss (approved):

PR #2957, opened 2026-03-28
Add linux ppc64le wheel build to CI

runlevel5:
# Summary

To add wheel build for Linux ppc64le target.

Why? The PPC community has been building pyrefly wheel manually for a while when working with projects that happen to leverage pyrefly for example servo project. It would be a big win in term of convenience/time-saving if there is an official wheel build.


yangdanny97 (approved):

PR #2962, opened 2026-03-30
Emit error when @dataclass is applied to a Protocol class

### LINKED ISSUE #2921: @dataclass applied to Protocol subclass
Applying `@dataclass` to a `Protocol` subclass should be rejected.

```python
from typing import Protocol

class Printable(Protocol):
    def display(self) -> str: ...

dataclass(Printable)
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksuAE74C2y26ALugMZTpxzzIT1S%2BGiyasOXHgB1MYOoxblSETAHN%2Bg4aIAKdFvnb4oMmZ25xkulW1ywAFLvz7DUAJSIZyL0xhgmEOFIucjteKDBXZABaAD5kOBYaFGIUkxxxMx4HGmt0WxhXGRAAXwAaag4WCAA3GAAxaBgqEAwcAhIKEqA)

syhstanley:
## Summary

Fixes #2921

- Adds a check in `class_metadata_of` that emits `InvalidDecorator` when a Protocol class is decorated with `@dataclass` or `@dataclass(...)`.
- Both `mypy` and `pyright` already reject this pattern — `@dataclass` on a Protocol is semantically invalid since Protocol classes define structural interfaces, not data containers.

## Test plan

- [x] Added `test_dataclass_decorator_on_protocol` — bare `@dataclass` on Protocol reports error
- [x] Added `test_dataclass_decorator_with_args_on_protocol` — `@dataclass(frozen=True)` on Protocol reports error
- [x] All 112 existing dataclass tests pass (zero regression)
- [x] All 42 protocol tests pass

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -392,26 +392,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
 
         // @dataclass should not be applied to Protocol classes.
-        if protocol_metadata.is_some() {
-            for (decorator, range) in &decorators {
-                let is_dataclass = match decorator.ty.callee_kind() {
-                    Some(CalleeKind::Function(FunctionKind::Dataclass)) => true,
-                    _ if let Type::KwCall(call) = &decorator.ty
-                        && call.has_function_kind(FunctionKind::Dataclass) =>
-                    {
-                        true
-                    }
-                    _ => false,
-                };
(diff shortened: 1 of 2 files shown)


syhstanley:
> Can we just check if the class has dataclass metadata, instead of looking thru all the decorators? that should have a similar effect.

Thanks for your feedback. New commit pushed.

PR #2963, opened 2026-03-30
fix match-case: false positive [unbound-name] #2932

### LINKED ISSUE #2932: match-case: false positive [unbound-name]
### Describe the Bug

```python

def test(x: int | None, y: int | None) -> None:
    match x, y:
        case None, None:
            raise ValueError
        case int(m), None:
            u = m * 3
            v = m
        case None, int(n):
            u = n
            v = n // 3
        case _, _:
            raise ValueError

    assert_type(u, int)
    assert_type(v, int)
```

```
ERROR sandbox.py:16:17-18: `u` may be uninitialized [unbound-name]
ERROR sandbox.py:17:17-18: `v` may be uninitialized [unbound-name]
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2o4cGHwD6TYjAA66GZhhhG8BgAp8iTun4AfOgDlc6GABo6pDax37DMAJR0AtAD5rRxDLqe6NVAwDGABZ0%2BKbmHl4RfoIwriax7ugRSV6UqBDCdABqqFAArjAAopTUlOHJURmWKjS2pgZuZcleuXQAvN50AFR0AMyNTXQAbm3e-V4VMfVxVei2CQMRLe2yiQuew8t0APRbvWOeE3SipqLza3Sp6THZeYXFvDKNgsJiEjAquaaWtk9CIgziZjvQZfLS2EDGEBkSgKKCkQgMWhQCgAYjoAAVSDCwHC6GgsHh8HQ-IZIGxcqkGBBDIQZGiAMowGIBBgMYhwRA7aGw%2BG8NhbGDoLaYXB%2BOBbEnoMkU3zUoV0MC8ASDNJQVDYWDE0kQcmUuV0XDEKmGOC09BkBgBQwOQYiOD69pSEA9QgARgATE6ZABtEQlOAAXRkuXQXB4fBgmAcmAgML8VNtIwA5Dr0LwYEmZGmGA4YQBHXKxyMOADWMFIDlQfj88DgyYA7qhKOhM%2BgQABfSFVhMwABi0BgFHxOAIJHI7aAA

_No response_

asukaminato0721:
# Summary

Fixes #2932

teaching the binder to treat tuple-subject sequence patterns like `case _, _:` as syntactically exhaustive when the match subject is a tuple expression of compatible shape. 

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/binding/function.rs
+++ b/pyrefly/lib/binding/function.rs
@@ -975,7 +975,12 @@ fn function_last_expressions<'a>(
                 let mut syntactically_exhaustive = false;
                 for case in x.cases.iter() {
                     f(sys_info, &case.body, res)?;
-                    if pattern_is_syntactically_exhaustive_for_subject(&x.subject, &case.pattern) {
+                    if case.guard.is_none()
+                        && pattern_is_syntactically_exhaustive_for_subject(
+                            &x.subject,
+                            &case.pattern,
+                        )
+                    {
                         syntactically_exhaustive = true;
                         break;
                     }
--- a/pyrefly/lib/binding/function.rs
+++ b/pyrefly/lib/binding/function.rs
@@ -63,6 +63,7 @@ use crate::binding::binding::ReturnTypeKind;
 use crate::binding::bindings::BindingsBuilder;
 use crate::binding::bindings::LegacyTParamCollector;
 use crate::binding::expr::Usage;
+use crate::binding::pattern::pattern_is_syntactically_exhaustive_for_subject;
 use crate::binding::scope::FlowStyle;
 use crate::binding::scope::InstanceAttribute;
 use crate::binding::scope::Scope;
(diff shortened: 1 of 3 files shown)


NathanTempest:
@asukaminato0721 Great fix! The approach with `pattern_is_syntactically_exhaustive_for_subject()` is exactly right.

One note on Copilot's suggestion about `guard.is_none()` I checked and the `guard` field is already available and is an `Option<Expr>` type. You can see it's already being used in `pattern.rs`:

```rust
if let Some(mut guard) = guard {
    self.ensure_expr(&mut guard, &mut Usage::Narrowing(None));
    // ...
}
```

So implementing the fix is just a one-liner:
```
let case_is_irrefutable = guard.is_none()
    && pattern_is_syntactically_exhaustive_for_subject(&subject_expr, &pattern);
```

This handles edge cases like:

case _, _ if some_condition:  # Pattern exhaustive, but guard can fail!

Without this check, Pyrefly would incorrectly treat guarded catch-all cases as exhaustive. The same change should be applied in function.rs for the return analysis path.

PR #2966, opened 2026-03-30
fix __bool__ check should look at Never return #1021

### LINKED ISSUE #1021: __bool__ check should look at Never return
### Describe the Bug

```python
from typing import Never

class Foo:
   def __bool__(self) -> Never:
     raise TypeError

if Foo():
    print("Hello World")
```

This works fine, but if I change it to `__bool__ : int` then it gives an error that `Foo` cannot be used as a boolean. The above formulation is accepted by Pyright, and recommended in https://discuss.python.org/t/how-to-get-a-type-check-error-on-bool/74879/7

The code in question is at https://github.com/facebook/pyrefly/blob/54a40386569be3c53f3c760611d7f5b1a1793464/pyrefly/lib/alt/attr.rs#L1695-L1709. We check if the type itself is union or never, then get the call target. We should probably be looking at the return type of the call target.

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoByApgG5EgBQFAxgDYCGAzo1AGJhgBcFUvAJkWBQA%2BsIBGHWqIAUjIrWABKKAFoAfIVLluvXSHpI5UACqIiAURDhKFJEPZhpinboQhUMaQCIAEvNpgUADquLR8XopAA&version=3.12

_No response_

asukaminato0721:
# Summary

Fixes #1021

tightening the `__bool__` guard.

The check still allows a raw Never value in boolean position, but it now rejects a callable `__bool__` whose return type is `Never`

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -47,6 +47,7 @@ use crate::types::callable::Params;
 use crate::types::class::Class;
 use crate::types::class::ClassType;
 use crate::types::facet::FacetKind;
+use crate::types::function::Function;
 use crate::types::function::FunctionKind;
 use crate::types::function::PropertyMetadata;
 use crate::types::function::PropertyRole;
@@ -61,6 +62,7 @@ use crate::types::typed_dict::TypedDict;
 use crate::types::types::AnyStyle;
 use crate::types::types::BoundMethodType;
 use crate::types::types::Overload;
+use crate::types::types::OverloadType;
 use crate::types::types::SuperObj;
 use crate::types::types::Type;
 
@@ -2936,7 +2938,10 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
 
                 if dunder_bool_ty
                     .callable_return_type(self.heap)
-                    .is_some_and(|ret| ret.is_never())
+                    .is_some_and(|ret| {
+                        ret.is_never()
+                            && self.callable_has_explicit_return_annotation(&dunder_bool_ty)
+                    })
                 {
                     self.error(
                         errors,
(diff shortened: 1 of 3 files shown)


stroxler (approved):
I think this makes sense, I'll try to get a second review and merge. Thanks!

fangyi-zhou:
It's interesting why `isna` is triggering new errors for pandas.

> + ERROR pandas/core/arrays/base.py:592:12-42: The `__bool__` method of `NDFrame` returns `Never`, so it cannot be used as a boolean [invalid-argument]

https://github.com/pandas-dev/pandas/blob/1bba48f900275af485e457b056f3e4fc82b38f94/pandas/core/arrays/base.py#L592

> + ERROR pandas/core/arrays/interval.py:1054:12-28: The `__bool__` method of `NDFrame` returns `Never`, so it cannot be used as a boolean [invalid-argument]

https://github.com/pandas-dev/pandas/blob/1bba48f900275af485e457b056f3e4fc82b38f94/pandas/core/arrays/interval.py#L1054

Looking at the definition of `isna` https://github.com/pandas-dev/pandas/blob/main/pandas/core/dtypes/missing.py
It looks like we're getting back an union type with one of the return types with invalid `__bool__` check.

fangyi-zhou:
```
class Foo:
    def __bool__(self):
        raise TypeError
class Bar(Foo):
    def __bool__(self) -> bool:
        return True
bar: Foo = Bar()
if bar: ...   # <-- We're reporting a false positive here
```

I managed to produce a false positive error here, maybe we shall tighten the check?

fangyi-zhou (changes requested):

fangyi-zhou:
Thanks for addressing the issue with the false positive! I'd defer to members of the team to see if we need to do anything with the union issue.

PR #2971, opened 2026-03-30
provide type: fixes and features

KotlinIsland:
- fqn for function type name
- show type var bound, default, and scope
- overload
- include `builtins.` in more cases

i'm not sure about all of the changes, happy for feedback

@yangdanny97

### DIFF
--- a/pyrefly/lib/test/lsp/lsp_interaction/provide_type.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/provide_type.rs
@@ -59,63 +59,7 @@ fn test_provide_type_unopened_file_with_dependencies() {
         .expect_response(json!({
             "contents": [{
                 "kind": "plaintext",
-                "value": "type[bar.Bar]".to_owned()
-            }]
-        }))
-        .unwrap();
-
-    interaction.shutdown().unwrap();
-}
-
-#[test]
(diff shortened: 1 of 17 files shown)


yangdanny97:
Thanks, I'll try my best to get this merged before the next release

KotlinIsland:
> Thanks, I'll try my best to get this merged before the next release

thank you 🙏. in PyCharm we are working on a big, deep integration with Pyrefly, and this would help a lot

pyrefly/lib/test/lsp/lsp_interaction/provide_type.rs
```diff
@@ -162,41 +145,133 @@ fn test_provide_type_from_pyi_file() {
...
+#[test]
+fn test_super() {
+    do_test(37, 10, "builtins.super[builtins.object, bar.S]");
+}
+
+#[test]
+fn test_new_type() {
+    // NOTE: This is a bit of a hack, because it's not actually a `builtins.type`,
+    //  but the client should realize that when they look up the qname
+    do_test(39, 1, "builtins.type[bar.N]");
```
KotlinIsland, line 244: maybe ideally we could say something like `NewType[bar.N]`, but this should be fine for now

PR #2981, opened 2026-03-31
Improve ParamSpec named-arg handling

### LINKED ISSUE #2972: Structural matching incorrectly merges named parameters on Callable Protocol into associated ParamSpec
### Describe the Bug

Easiest to look at the sandbox link, but consider a function decorator like this:

```py
P = ParamSpec('P')

class InstanceMethod(Protocol[K_con, P, T_co]):
    def __call__(_self, self: K_con, *args: P.args, **kwargs: P.kwargs) -> T_co: ...

@overload
def decorator(fn: InstanceMethod[K_con, P, T_co]) -> DecoratorWrapper[P, T_co]: ...
```

When applied to a bare function like this:

```py
@decorator
def bare_fn(a: int, b: str) -> str: ...
```

pyrefly matches against the indicated overload to produce a `DecoratorWrapper[[b: str], str]`, incorrectly pulling off `a` into the `self: K_con` parameter. 

mypy and pyright both correctly skip this overload and proceed to a simpler overload that matches the bare fn more specifically, and only match this overload against an actual class instance method, allowing `decorator` to be used identically on any type of function—classmethod, staticmethod, instance method, or bare fn. I'd love to figure out a way to use pyrefly with this same code, but seems like the structural matching doesn't work the same as mypy or pyright in this case.

Thanks for your time!

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRAFToBeOmOYwAaqkoAKAORjVASnEB9AMa4pMuYpXr9uVQBo6BgG5KIGBpLGUArjG3oACkZ9KqDQAysQwemo%2BWiIA0kayYaZqMdHoMRbo8SZKyRnWtrjoDJSoDpRORa4eXiIielCocHB0AMK8uO4MrDDKAOIw6DDlegDaYgC6mogidLN0mDBgdJTw7lAMynAwUGCadAC0AHwyiHSE57Xo9Y3NAIJwpFeo2LAA6iXEYSo%2B1Ay4BlARj4bBMpjM5jp9KgoFBIacWtCGi8YECQeNLnN5os6DB8OFOjAdI1HhEtjsbAIlGw4KcfIQqXAKQIANYAdwZtMIbIZeyOrXanW6Y3Gp3OhHBswWS1ZEAYAAsMgxcRsybsDsdVPcSc83h8vqiZONVKKLqIzdcmnQAJLoOAMDB6GAAWRg8twmGUP1wfwBI3SBnQNmBMgsk2mmTmUrokL0iMhyh0qpsqtO-sKFI5dDpDKZ3Mo1M5eepvOOYgsJvFZrqDUtLRrcBdbo9Xp9uEBacDWZBobBEcl2JjcZ0CaTtigNOMYT9GXGGfzE%2Bz89z7PnhZXxfVIYMFcu1ZudAAQkoYAAxdCe37-NsGssGMMSrFLQcw%2BOJ7ZgOcFrP0peCFnrhcuQAkst1wHcq3QAABXA7CGKBcFQTARCjYknmRZQHzAdB4XrRs5XdP0gzRKwRBArU0N1VBPiGG8RTOU1Lmg2DKHgxDkOxVDY3QzDsOtW17SuZ1XXwzBCK7Q0SPQMiHgomB3io-VgwmcDGJguCEKQrAOJkrjYGULDTiPFYz1o6TtWReTqMoWiVLNFCdJ1HosLM2TLMUtFw0xEoIC2OgADlvStbhYBoAYlU0kRIM4xz2KWbBjx0LDlFQU5WAYEC7UoWyLWaOsbk8uYooc5EHyjUKm02d8bBSzgigy4oKzoB8ivM2BmrtVAuj0cqRNK7EOq65LUrqzdMvAzEWtk5qcp690%2BqWeo4Fmj1Fuq4b0tGhr6MrEQbiGBgdCYMJlHilZEs7cjdLkvUaJGNLZzoTLJl2pp9sOuRlDypplE0QhlsIXF8SVIkdOSyQABYACZNBsNpqEFQYRie7w9r4d7jq%2BuBCAGiAIlQCHoeTYoUdetGjp6THCEW5awahmHHuJkQQCsEAyBWMAoFIQg-hoKAKAAYizUh2c5x6MBwAgCnQSA2HcEoukKStBeCGAYDoOUGAYYgaQAeh1tnFk5wheDYHWBh1zB-jgHWAxluXOogQodboMBeDoUpUGgRypbt%2BXHcyXBiAV21KzIN10H2Zi4H9owhBAABmQgAEZIbjkQRiGahKDgdF0HcdAuB4PgYEwfZMAgFY9C6WCjFUCA2HQXgYFUERG4YfYVgAR3cCuS-2ZkYFIfZUD0R1LWkVQV3QFv0BAABfFmR%2Br09oBgCg0CwPAiDIeegA

_No response_

dhleong:
# Summary

By storing the full ParamList we are able to more precisely match against `ParamSpec` params used in method signatures that prepend named arguments (unlike `Concatenate` which only prepends types). This should also enable better support for named params in general when captured by `ParamSpec`.

Fixes #2972

This change was mostly one-shot with claude, but I did go through and do some minor cleanup.

# Test Plan

Added a test case matching the problem described in the linked issue and ensured it passes with the proposed changes.

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -680,29 +680,29 @@ impl Callable {
                 output.write_str("(Materialization) -> ")?;
                 write_type(&self.ret, output)
             }
-            Params::ParamSpec(prefix, pspec) => {
+            Params::ParamSpec(args, pspec) => {
                 output.write_str("(")?;
-                for (i, param) in prefix.items().iter().enumerate() {
+                for (i, arg) in args.items().iter().enumerate() {
                     if i > 0 {
                         output.write_str(", ")?;
                     }
-                    write_type(param.as_type(), output)?;
+                    write_type(arg.as_type(), output)?;
                 }
                 match pspec {
                     Type::ParamSpecValue(params) => {
-                        if !prefix.is_empty() && !params.is_empty() {
+                        if !args.is_empty() && !params.is_empty() {
                             output.write_str(", ")?;
                         }
                         params.fmt_with_type(output, write_type)?;
                     }
                     Type::Ellipsis => {
-                        if !prefix.is_empty() {
+                        if !args.is_empty() {
                             output.write_str(", ")?;
                         }
                         output.write_str("...")?;
                     }
                     _ => {
-                        if !prefix.is_empty() {
+                        if !args.is_empty() {
                             output.write_str(", ")?;
                         }
                         output.write_str("ParamSpec(")?;


rchen152 (approved):

yangdanny97:
FYI the version that was merged does not directly store a ParamList, since we felt like that was too permissive and it was hard to enforce the invariant that it shouldn't have Varargs, Kwargs, and Kwonly params.

We instead use a new PrefixParam enum that only has Pos/PosOnly, and maps cleanly to regular Params

dhleong:
@yangdanny97 Sounds good to me, thank you! 🙏

PR #2998, opened 2026-04-03
Prevent panic on malformed empty parameter names during binding solve

### LINKED ISSUE #2984: panic: `key lacking binding` with fuzzed code
### Describe the Bug

```python
def f(
    @
a:b:@or1:c=1
```

```
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/binding/bindings.rs:429:13:
Internal error: key lacking binding, module=__unknown__, path=/home/user/crash.py, key=Key::Definition( 3:9), key-debug=Definition(ShortIdentifier(21..21))
```

### Version info
0.59.0

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABQA66NrNAAs6otou7gCcAjIgDGAXiEgANCDIDqUUoQAuuALZQKAYhoAFUgrBKaaLHnw1RudJADmAVwGoVEG4Wa6AyjBg0AFioqxHCIAPRh8orKgnZhMOhhmLiicGHWthCOzq42YXSCNKgAbqjQqNiwVjb2Ti5uLLjEuehwHuhkKv42ALTFMAJwDTTiNIwgAMyEQgBM48wA2gMCgnAAuswO6BDqxIIqMJg9mBAKoq79IzQA5FnogjDXzPcqPQoAjg6nhz0A1jCkHqoUSieBwK7XADuqAE6Ce6BAAF9ZMCLjAAGLQGAUMw4AgkciIoA

_No response_

Arths17:
### Summary:

Implemented and validated the crash fix for issue #2984, added a regression test for the fuzzed snippet, confirmed the repro no longer panics (it now emits parse/type errors), and completed formatting/lint checks with jsonschema skipped due missing local Python deps.

Fix a crash in the binding/solve pipeline for malformed parser-recovery input where an empty synthesized parameter name could leave a promised definition key without a corresponding binding.

**This PR adds a repair step before solve that:**

Fills only malformed empty-name definition promises with an error binding.

Preserves strict invariant checking by still panicking on any other missing binding.

Also adds a regression test using the fuzzed repro from the issue to ensure this no longer panics and instead reports normal parse/type diagnostics.

Fixes #2984

### **Test Plan**

**Ran targeted regression test:**

cargo test -p pyrefly test_crash_on_fuzzed_empty_parameter_name -- --nocapture

**Re-ran the original fuzzed snippet via CLI:**

cargo run -q -p pyrefly -- snippet $'def f(\n @\na:b:@or1:c=1\n'

Verified: no panic; parse/type errors are reported.

**Ran formatting and lint checks:**

python3 [test.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) --no-test --no-conformance --no-jsonschema

**Note:**

jsonschema checks were skipped locally because required Python packages were not installed in this environment. is this good for a pr


pyrefly/lib/test/simple.rs
```diff
@@ -1766,6 +1766,20 @@ fn test_panic_on_unicode() {
     // We manually ignore the missing expectations, because adding `# E:` would change the test.
 }
 
+#[test]
+fn test_crash_on_fuzzed_empty_parameter_name() {
+    // Regression test for https://github.com/facebook/pyrefly/issues/2984
+    assert!(
```
yangdanny97, line 1772: this test case format seems odd? we can just write a normal test case with `# E:` comments

.gitignore
```diff
@@ -1,3 +1,4 @@
 debug.json
 debug.js
+venv/
```
yangdanny97, line 3: revert

.vscode/tasks.json
```diff
@@ -130,14 +130,7 @@
           }
         }
       },
-      "problemMatcher": {
```
yangdanny97, line 133: revert

pyrefly/lib/binding/bindings.rs
```diff
@@ -817,6 +818,36 @@ fn extract_new_defaults(stmt: &Stmt, name: &str) -> Option<Vec<Expr>> {
...
+        for (idx, key) in self.table.types.0.items() {
+            if self.table.types.1.get(idx).is_none() {
+                missing.push((idx, key.clone()));
+            }
+        }
+
+        for (idx, key) in missing {
+            match key {
+                Key::Definition(short_identifier) if short_identifier.range().is_empty() => {
+                    self.insert_binding_idx(idx, Binding::Any(AnyStyle::Error));
```
yangdanny97, line 837: i believe we panic if we do already have a binding inserted, so i think we need to check that.

yangdanny97:
thanks for the PR, I left some comments but I think no action needed from you ATM since someone else already made a change that fixes this, and we need to decide which fix to use

yangdanny97:
I have some reservations with this approach, since it's kind of a cleanup that hides mistakes in other parts of the system. I think i would prefer something that prevents definitions w/ empty keys from being created at all

Arths17:
Thanks for the review, @yangdanny97! That makes total sense preventing the empty keys from being created in the first place is definitely a cleaner long-term fix.

I saw the note about another potential fix being in the works. I'm happy to hold off for now while you guys decide on the best direction, or if you'd like, I can take a stab at moving the logic earlier into the pipeline to prevent the creation of those definitions. Just let me know how you'd like to proceed!

PR #3001, opened 2026-04-03
fix Pydantic lax-mode rewrites Pattern[str] to Pattern[LaxStr] and rejects re.Pattern[str] #2991

### LINKED ISSUE #2991: Pydantic lax-mode rewrites `Pattern[str]` to `Pattern[LaxStr]` and rejects `re.Pattern[str]`
## Summary

When checking Pydantic models in Pyrefly, constructor parameters for `re.Pattern[str]` fields appear to be rewritten as `Pattern[LaxStr]` in lax mode.  
This then causes valid `re.Pattern[str]` values (from `re.compile(...)`) to be rejected by static checking.

This seems inconsistent with runtime Pydantic behavior and likely indicates a bug/edge case in lax-type conversion for `re.Pattern[...]`.

## Minimal repro

```python
import re
from typing import Literal
import pydantic

class RegexSignature(pydantic.BaseModel):
    type: Literal["regex"] = "regex"
    signature: re.Pattern[str]

_ = RegexSignature(type="regex", signature=re.compile(r"needle"))
```

## Pyrefly output

```
ERROR Argument `Pattern[str]` is not assignable to parameter `signature` with type `Pattern[LaxStr]` in function `RegexSignature.__init__` [bad-argument-type]
```

(From my project this occurs in many callsites with the same shape.)

## Expected behavior

No type error for `re.compile(...)` when field type is `re.Pattern[str]`.

## Actual behavior

Pyrefly reports `Pattern[str]` is not assignable to `Pattern[LaxStr]`.

## Why this looks wrong

- Runtime Pydantic accepts both:
  - `RegexSignature(signature=re.compile("x"))`
  - `RegexSignature(signature="x")` (string pattern coercion)
- Runtime Pydantic **does not** accept bytes for this field (e.g. `signature=b"x"` fails validation), so expanding to `LaxStr` may be over-broad for `Pattern[str]`.
- `pyright` and `mypy` report no errors on the same callsites.

## Environment

- `pyrefly 0.59.0`
- `pydantic 2.12.5`
- `Python 3.14.3`
- Linux

## Additional context

This appears related to Pyrefly’s Pydantic lax-mode conversion behavior:
https://pyrefly.org/en/docs/pydantic/#how-lax-mode-works-in-pyrefly

Potentially, `re.Pattern[T]` may need special handling instead of recursively applying `str -> LaxStr` inside the type parameter.

> [!IMPORTANT]
> Although I encountered and debugged the issue myself, I used AI to create the content of the issue. I then verified the content of the issue and checked that the reproduction is correct.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2991

Pydantic lax conversion now special-cases regex patterns so `re.Pattern[T]` expands to `re.Pattern[T] | T` instead of recursively widening the inner type to LaxStr.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/pydantic_lax.rs
+++ b/pyrefly/lib/alt/class/pydantic_lax.rs
@@ -161,6 +161,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         None
     }
 
+    fn get_regex_pattern_lax_conversion(
+        &self,
+        class_obj: &Class,
+        ty: &Type,
+        targs: &[Type],
+    ) -> Option<Type> {
+        if class_obj.has_toplevel_qname(ModuleName::from_str("re").as_str(), "Pattern")
+            || class_obj.has_toplevel_qname(ModuleName::typing().as_str(), "Pattern")
+        {
+            let pattern_input = targs
+                .first()
+                .cloned()
+                .unwrap_or_else(|| self.heap.mk_any_implicit());
+            return Some(self.union(ty.clone(), pattern_input));
+        }
+        None
+    }
+
     fn get_tuple_element_type(&self, tuple: &Tuple) -> Type {
         match tuple {
             Tuple::Unbounded(elem) => self.expand_type_for_lax_mode(elem),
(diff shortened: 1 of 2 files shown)


migeed-z (approved):

PR #3011, opened 2026-04-04
detect slots layout conflict in multiple inheritance

### LINKED ISSUE #2916: __slots__ layout conflict with same slot names
Even with identical slot names, two separate classes define different C-level layouts, so MI still fails at runtime.

```python
class First:
    __slots__ = ("x",)

class Second:
    __slots__ = ("x",)

class Both(First, Second): ...
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksiAMZTpxzIBiEATnAC6IA6myA5AH0hcKPk5wRyALzIAFDxCElAGgCUfPnQZMAyjBr4cvfoJFiJUobIVKVIDVsw7GyAEISAFvNYdOqsgGRjjqKMQRfCAAvoEg6DScEABuMKywVCAYOAQkFDFAA)

### LINKED ISSUE #2917: __slots__ layout conflict in multiple inheritance
Inheriting from two classes that both define non-empty `__slots__` causes a `TypeError` at runtime.

```python
class Left:
    __slots__ = ("a", "b")

class Right:
    __slots__ = ("c", "d")

class Combined(Left, Right): ...
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksiAMZTpxzIAyMYALogDqbJ-IB9AXCj52cIcgC8yABRcQ6BQBpkC3AoCUPHnQZMAShADmAC049%2Bg4aPGSZ82irUhsWnZj2NkAYXwBbXAhMGGxZVg5VIzN2TRRiBJ4QAF9VRRp2CAA3GAAxaBgqEAwcAhIKFKA)

knQzx:
## summary
- added a check in class metadata computation that detects when two or more base classes define non-empty `__slots__`, which causes a TypeError at runtime in cpython
- removed the `bug` markers from the existing test cases for #2916 and #2917 and added expected error comments
- added new test cases for edge cases - empty slots, both empty, and three bases

## test plan
- ran `cargo test -p pyrefly -- test_slots` - all 27 tests pass
- tested the conflict detection for same slot names (#2916), different slot names (#2917), empty slots (no false positive), and three-base inheritance

fixes #2916
fixes #2917

### DIFF
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -409,9 +409,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             let bases_with_nonempty_slots: Vec<&Class> = bases_with_metadata
                 .iter()
                 .filter(|(_, metadata)| {
-                    metadata
-                        .slots_info()
-                        .is_some_and(|si| !si.names.is_empty())
+                    metadata.slots_info().is_some_and(|si| !si.names.is_empty())
                 })
                 .map(|(base, _)| base)
                 .collect();


yangdanny97 (approved):

PR #3012, opened 2026-04-04
flag ambiguous protocol members without type annotations

### LINKED ISSUE #2925: Ambiguous protocol member with value assignment
Assigning a value (not a type annotation) in a Protocol body is ambiguous. Pyrefly should flag this.

```python
from typing import Protocol

class Ambiguous(Protocol):
    x = None
    y = ...
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksuAE74C2yALuaRJgObIT2n41NkABTpN8AY3xQAOplnio6OHGQBBerggcArvm1wAFCPxjJUAJSJZyG8kLIAvMgBy%2BTDGu3yj5MT%2ByQAF8AGmp0cSYIADcYADFoGCoQDBwCEgogoA)

knQzx:
## summary
- add new `UnannotatedProtocolMember` error kind for protocol members assigned a value without an explicit type annotation
- matches mypy behavior - protocol members must have declared types
- updates existing bug test case to verify the new diagnostic fires correctly

## test plan
- updated `test_protocol_ambiguous_member` to expect the new error on unannotated assignments like `x = None`
- added positive case (`class Ok(Protocol)`) with annotated members to confirm no false positives
- all 34 existing protocol tests still pass

fixes #2925

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1588,7 +1588,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 ..
             } => {
                 let direct_annotation = annot.map(|a| self.get_idx(a).annotation.clone());
-                if metadata.is_protocol() && direct_annotation.is_none() {
+                if metadata.is_protocol()
+                    && direct_annotation.is_none()
+                    && !is_dunder(name.as_str())
+                {
                     self.error(
                         errors,
                         range,


Arths17:
Hey @knQzx, nice work on this. Parity with mypy regarding unannotated protocol members is definitely important for consistency across the ecosystem.

I've looked through the summary, and the test plan seems solid especially verifying that existing protocol tests still pass. One quick question: for members assigned a value like x = None without an annotation, does the new UnannotatedProtocolMember error provide a clear suggestion to the user to add a type hint?

I'm happy to pull this branch and verify it against a few edge cases (like nested Protocols) to make sure there aren't any weird regressions!

knQzx:
thanks, yeah the error points to the unannotated member and suggests adding a type annotation

Arths17:
Appreciate the quick response! Suggesting the type hint directly in the error message is a huge plus for developer experience.

Parity with mypy on unannotated protocol members is a major step for making Pyrefly a solid drop-in replacement. Since the existing 34 tests are passing and the diagnostic is clear, this looks great to me.

lolpack:
I thought we were going to have this error code off by default per the discussion in the original ticket?

PR #3018, opened 2026-04-05
fix Type aliases are expanded in the hover text #2707

### LINKED ISSUE #2707: Type aliases are expanded in the hover text
### Describe the Bug

```py
type Messages = list[str]

def func(msgs: Messages) -> None: pass
```

When hovering `func`, I'd expect it to show `msgs: Messages`, but it shows `msgs: list[str]`. Is this intended or a bug? One of the main reasons for using aliases is documentation. 

This happens in both `type` statements and `TypeAlias`.

pyrefly ([playground](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIALqcTAAQCy8aA5vHQLx1QRyUDafAE4BdADroJmGGDpgArugDGACgC2cFnESNmqNnACUdALQA%2BOgDlc6GDuKo4CAL4AaEKiWUIANxgAxaBgKNCw8IjIQZyA)):
```
(function) func: def func(msgs: list[str]) -> None: ...
```

pyright ([playground](https://pyright-play.net/?code=C4TwDgpgBAshDO8CGBzBUC8UA2BLewA2gQE4C6AUBQCYQBmUdArgHYDGAFALbwrwBcsBMjTwAlFAC0APigA5APYsIgsEkRA))
```
(function) def func(msgs: Messages) -> None
```

asukaminato0721:
# Summary

Fixes #2707

non-call hovers on plain functions render the declared source header instead of the normalized solved signature.

That preserves aliases like Messages in hover text while leaving call-site hover behavior unchanged for constructors, `__call__`, and expanded `**kwargs`.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_types/src/annotation.rs
+++ b/crates/pyrefly_types/src/annotation.rs
@@ -25,6 +25,8 @@ use crate::types::Type;
 pub struct Annotation {
     pub qualifiers: Vec<Qualifier>,
     pub ty: Option<Type>,
+    /// A semantically equivalent type that retains alias names for presentation.
+    pub display_ty: Option<Type>,
 }
 
 impl Display for Annotation {
@@ -49,6 +51,15 @@ impl Annotation {
         Self {
             qualifiers: Vec::new(),
             ty: Some(ty),
+            display_ty: None,
+        }
+    }
+
+    pub fn new_type_with_display(ty: Type, display_ty: Type) -> Self {
+        Self {
+            qualifiers: Vec::new(),
+            ty: Some(ty),
+            display_ty: Some(display_ty),
         }
     }
 
@@ -76,6 +87,7 @@ impl Annotation {
         Self {
             qualifiers: self.qualifiers,
             ty: self.ty.map(|ty| substitution.substitute_into(ty)),
+            display_ty: self.display_ty.map(|ty| substitution.substitute_into(ty)),
         }
     }
 }
@@ -102,7 +114,8 @@ mod tests {
         assert_eq!(
             Annotation {
                 qualifiers: Vec::new(),
-                ty: Some(Type::None)
+                ty: Some(Type::None),
+                display_ty: None,
             }
             .to_string(),
             "None"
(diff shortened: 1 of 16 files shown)


NathanTempest:
Thanks for the PR @asukaminato0721! The approach (using source text instead of resolved type) is the right way to preserve type aliases in hover output. Tagging @yangdanny who owns the LSP hover domain.

A few things worth looking into:

- AST re-parse on hover hot path `Ast::parse(module.contents(), ...)` runs on every function hover. For larger files this could be noticeable. (This is what the copilot reviewer flagged)

- Silent fallback if body has leading comments/blank lines where the slice goes from function_def.range.start() to the first AST statement's start. Comments and blank lines aren't statements, so for a function like:

```
def func(msgs: Messages) -> None:
    # comment
    pass
```

The slice ends with # comment, the ends_with `:` check fails, and you fall back to the (broken) type-based display. Worth scanning to the signature-terminating `:` directly instead.

asukaminato0721:
> Thanks for the PR @asukaminato0721! The approach (using source text instead of resolved type) is the right way to preserve type aliases in hover output. Tagging @Yangdanny who owns the LSP hover domain.
> 
> A few things worth looking into:
> 
> * AST re-parse on hover hot path `Ast::parse(module.contents(), ...)` runs on every function hover. For larger files this could be noticeable. (This is what the copilot reviewer flagged)
> * Silent fallback if body has leading comments/blank lines where the slice goes from function_def.range.start() to the first AST statement's start. Comments and blank lines aren't statements, so for a function like:
> 
> ```
> def func(msgs: Messages) -> None:
>     # comment
>     pass
> ```
> 
> The slice ends with # comment, the ends_with `:` check fails, and you fall back to the (broken) type-based display. Worth scanning to the signature-terminating `:` directly instead.

fixed

stroxler (changes requested):
Back to you for a question: this commit appears to do a same-file-only change where we just use the actual function header as the hover signature.

That's (a) not the same as fixing that type aliases are supposed to be unexpanded, and (b) only works, I think, for same-file hovers.

I'm not necessarily opposed to using the function header, but is there something preventing us from just making type aliases display unexpanded as the original issue requested instead?

stroxler (approved):

rchen152:
This fix unfortunately causes a performance regression (https://github.com/facebook/pyrefly/issues/4769), so I'm backing it out to unblock the upcoming 1.3.0 release.

PR #3019, opened 2026-04-05
fix doesn't understand usage of inline Annotated function parameters #2531

### LINKED ISSUE #2531: doesn't understand usage of inline `Annotated` function parameters
### Describe the Bug

Image

unlike for `T1`, Pyrefly fails to resolve the type of `utils.ValueRange` when used inline as function parameter. when I hover on the symbol it doesn't show anything.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BBdOlwNUDGJgA66MNXqZxqAMZRUcOPE7de-RWNXq4Mrjz50Arg2jH0MgCoBGOgF5Bw0eMkBtVgwA0ltZQcIQAaqhQFjAAShhsMAAUALSOAAyBaQCUALoyMpgwYHRgFujKiQyOiHROgQwATDVCImISmL7oAUE24ZHRcexJjoHpuVk1hFMg-iBWvWQUcrR0%2BipqGlqmuqtKhhr5dkcAAmv7tud0EVGx8TCIMnRPdFC4NX6PzwAWEO9dh89VkUSrwaOIAProCw0bAwSiJfB-BhZOjJAB8dDgDEoD3QgMBlBgDAslDxYCkIAActDYTi6MB8ABfCkzEBkQlgKCkQgMWhQCgAYjoAAVSByuZiMDgCHRlLhZBA2CTxBB5YQZEKAMowGB0L4MBjEOCIAD0JvZRS5hF4bBNMHQJswuGUcBNcoVSsoKvlJpBlDoqAAbqhoKhsLBZfLIJ7vXjcMRrPLQjIyAwvvLkoG4XBVXi3BSAMyERwNCkyECM2YqaxZgBi0BgFDQWDwREWjKAA

_No response_

asukaminato0721:
# Summary

Fixes #2531

teaching hover/definition lookup to recover imported Annotated metadata symbols even when the use site has no type trace.

Attribute definition lookup now falls back to the exact base identifier when utils.

Name inside Annotated[...] lacks a trace, and hover now reuses the resolved definition to recover a displayable type instead of returning None.

# Test Plan

add test


samwgoldman (changes requested):
Thanks for addressing this. I’m requesting changes because I think the PR repairs the missing type information too late, in the LSP layer, rather than fixing where the information is lost.

Binding already visits every Annotated[T, metadata...] metadata expression as an ordinary expression. The inconsistency occurs during solving:

- expr_annotation handles Annotated as a qualifier, solves only T, and skips the metadata.
- The general special-form path infers every metadata expression and constructs Type::Annotated.

Ordinary annotation solving should infer the metadata and produce type traces. That fixes the problem in the right place.

I also don’t think metadata inference should use error_swallower() automatically. PEP 593 allowing arbitrary metadata means there is no required metadata type or compatibility relationship with T. It does not mean ordinary expression errors should be hidden. These should presumably still be diagnosed:

```
Annotated[int, missing_name]
Annotated[int, obj.missing_attribute]
Annotated[int, factory(bad_argument=True)]
Annotated[int, 1 + "x"]
```

The tests should also be strengthened:

- Assert the exact hover result. report.contains("ValueRange") is ineffective because the printed source frame already contains that text.
- Add an exact go-to-definition test, since this PR changes definition lookup.
- Test a chained base such as pkg.utils.ValueRange.
- Test multiple metadata entries.
- Test the intended diagnostics for undefined names, missing attributes, and invalid expressions in metadata.
- Consider directly asserting that metadata expressions receive type traces.

PR #3023, opened 2026-04-05
fix autocomplete for discriminated unions #1442

### LINKED ISSUE #1442: autocomplete for discriminated unions
## Use Cases

```py
from typing import TypedDict, Literal, TypeAlias

class Foo(TypedDict):
    kind: Literal["foo"]
    foo_value: int

class Bar(TypedDict):
    kind: Literal["bar"]
    bar_value: str

type FooBar = Foo | Bar
```

### case 1

```py
item: FooBar = {
    "kind": "|",
}
```

expected completion: `'foo' | 'bar'`

### case 2

```py
item: FooBar = {
    "kind": "foo",
    "|"
}
```

expected completion: `'foo_value'`

asukaminato0721:
# Summary

Fixes #1442

The completion engine now recovers the enclosing expected type for incomplete dict literals, narrows union members from already-entered literal fields like "kind": "foo", and uses that narrowed set for both key suggestions and literal value suggestions.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -220,7 +220,7 @@ impl Transaction<'_> {
     }
 
     /// Adds completion items for literal types (e.g., `Literal["foo", "bar"]`).
-    fn add_literal_completions_from_type(
+    pub(crate) fn add_literal_completions_from_type(
         param_type: &Type,
         completions: &mut Vec<RankedCompletion>,
         in_string_literal: bool,
@@ -1270,6 +1270,12 @@ impl Transaction<'_> {
                             &nodes,
                             &mut result,
                             in_string_literal,
+                        );
+                        self.add_dict_value_literal_completions(
+                            handle,
+                            mod_module.as_ref(),
+                            position,
+                            &mut result,
                         );
                         // `dict_key_claimed` was computed up front; when a dict key was
                         // offered we skip the overload literal completions.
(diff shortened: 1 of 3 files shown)


Arths17:
Hey @asukaminato0721, thanks for putting this together! Discriminated union narrowing for TypedDict is a massive improvement for the LSP experience.

I took a look through the logic and I’d like to second the AI's feedback on two points to help get this 'merge-ready':

-**Performance**: In typed_dict_field_type_in_member, the linear scan on typed_dict_fields could definitely become a bottleneck for large projects since completions fire on every keystroke. Mapping these fields once before the narrowing loop would keep the LSP snappy.

-**Idiomatic Rust**: Switching the BTreeMap<String, ()> to a BTreeSet<String> in dict_literal_present_keys would be much cleaner since we only care about membership.

Regarding the mypy_primer failure: I noticed check (7) is failing. If you need help debugging whether that’s a regression or just a flake, let me know I’m happy to pull the branch and help investigate!

asukaminato0721:
> Hey @asukaminato0721, thanks for putting this together! Discriminated union narrowing for TypedDict is a massive improvement for the LSP experience.
> 
> I took a look through the logic and I’d like to second the AI's feedback on two points to help get this 'merge-ready':
> 
> -**Performance**: In typed_dict_field_type_in_member, the linear scan on typed_dict_fields could definitely become a bottleneck for large projects since completions fire on every keystroke. Mapping these fields once before the narrowing loop would keep the LSP snappy.
> 
> -**Idiomatic Rust**: Switching the BTreeMap<String, ()> to a BTreeSet in dict_literal_present_keys would be much cleaner since we only care about membership.
> 
> Regarding the mypy_primer failure: I noticed check (7) is failing. If you need help debugging whether that’s a regression or just a flake, let me know I’m happy to pull the branch and help investigate!

fixed.

stroxler (approved):

PR #3029, opened 2026-04-06
fix imports using symlinked paths are not updated #1336

### LINKED ISSUE #1336: imports using symlinked paths are not updated
### Describe the Bug

When importing a python module through its symlinked path, changes are not picked up.

To reproduce:
```
echo 'def hello(name):\n    print(f"Hello {name}")' > test_module.py
ln -s test_module.py sym.py
echo 'from sym import hello\nhello("John")' > test_import.py
```

Now open both files in your editor with an LSP client and add/remove/rename an argument of hello in `test_module.py`. When requesting texDocument/hover or diagnostics the the data is not updated. The data is updated correctly if hello is imported with `from test_module import hello`, so it's an issue specific to symlinks.

Not sure if this is related to #1104 or not.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1336

The change has two parts.

Import discovery now canonicalizes symlinked filesystem hits, so import sym resolves through the real target path in the finder.

for the live LSP bug, imported modules now prefer the open in-memory file if the resolved path itself or its canonical target is open, which makes unsaved edits in the real file propagate through symlinked imports.

also invalidate import finding when a watched symlink path changes so retargeted symlinks do not stay cached.

# Test Plan

add test


stroxler (changes requested):

stroxler:
Back to you for a look at the CI failures before we review in more detail

stroxler (approved):

samwgoldman:
I reverted this change internally. This PR seemed to include an incidental change (the "live LSP bug" mentioned in the summary) which changed intended behavior. Pyrefly intentionally avoids propagating changes until didSave to avoid performing expensive cross-file checks on every keystroke.

stroxler:
I reopened the issue, since I think the actual symlink problem we meant to solve is not connected to the undesired didChange behavior

PR #3033, opened 2026-04-06
Fix LSP keyword rename on constructors; FailedParse config source

### LINKED ISSUE #2496: Renaming dataclass property renames class name instead
### Describe the Bug

When renaming a property of a dataclass (or Pydantic Schema / ...) pyrefly renames the class name instead:

```python
@dataclass
class Foo:
    a: int
    b: str


Foo(
    a=123,
    b='abc', # <- press F2 or "Rename Symbol" here, enter 'c'
)
```

After renaming:

```python
@dataclass
class c:
    a: int
    b: str


c(
    a=123,
    b='abc',
)
```

### Sandbox Link

_No response_

VSCode with pyrefly v0.53.0

Louisvranderick:
# Summary

Fixes #2496

Renaming a keyword at a call site (e.g. dataclass `Foo(a=...)`) could resolve the rename symbol to the class instead of the field parameter, so the client offered a rename that produced no edits. When load/parse failed for an on-disk `pyrefly.toml`, we still labeled the source as `File`, which misclassified config roots.

The LSP path adds `keyword_argument_fallback_to_callee` (on for go-to-definition, off for rename and find-refs), refines keyword targets against callee signatures (positional-only params and explicit `__init__` on class calls), and returns null from `prepare_rename` when a keyword argument has no definitions. Config adds `ConfigSource::FailedParse` and treats it like a real config path for roots, discovery, and messaging.

# Test Plan

- Added/updated LSP unit tests (rename, definition, lsp_interaction prepareRename) and `test_failed_parse_on_invalid_toml` in `pyrefly_config`.
- Ran `python3 test.py --no-test --no-conformance --no-jsonschema` and targeted `cargo test` for the new cases locally.

### DIFF
--- a/crates/pyrefly_types/src/literal.rs
+++ b/crates/pyrefly_types/src/literal.rs
@@ -13,7 +13,6 @@ use pyrefly_derive::TypeEq;
 use pyrefly_derive::Visit;
 use pyrefly_derive::VisitMut;
 use pyrefly_util::assert_words;
-use pyrefly_util::visit::VisitMut;
 use ruff_python_ast::ExprBooleanLiteral;
 use ruff_python_ast::ExprBytesLiteral;
 use ruff_python_ast::ExprFString;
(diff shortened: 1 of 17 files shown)


Arths17:
Hey @Louisvranderick, thanks for the hard work on this #2496 fix. I see you’re actively pushing commits and working through the logic as we speak, so I’ll stay tuned while the CI settles!

Just as a heads-up: I’m not a maintainer or official contributor yet just a developer following the project closely and looking to provide some extra eyes on the LSP side of things.

I did notice the recent revert on the enum literal member types; it’ll be interesting to see how that interacts with the new keyword_argument_fallback_to_callee flow. If that change was to avoid a regression elsewhere, I'm happy to help verify it against some complex TypedDict or Enum edge cases once the branch is stable.

crates/pyrefly_types/src/literal.rs
```diff
@@ -58,27 +57,15 @@ pub enum Lit {
...
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[derive(Visit, TypeEq)]
+#[derive(Visit, VisitMut, TypeEq)]
 pub struct LitEnum {
     pub class: ClassType,
     pub member: Name,
     /// Raw type assigned to name in class def.
     /// We store the raw type so we can return it when the value or _value_ attribute is accessed.
-    /// NOTE: Intentionally excluded from VisitMut so that type transformations like
```
kinto0, line 67: did you mean to do this?

pyrefly/lib/test/enums.rs
```diff
@@ -907,111 +908,3 @@ class Foo(str, Enum):
 Foo.from_dict({})
     "#,
 );
-
-// Regression test for https://github.com/facebook/pyrefly/issues/2583
```
kinto0, line 911: why remove these tests?

kinto0 (changes requested):

kinto0:
hey @Louisvranderick  - curious if you're still working on this. it seems like there's a few unrelated changes in here. did you mean to make those changes?

rchen152:
I'm going to close this PR, since it's stale. Please feel free to open a new one if you'd like to take another stab at this issue.

PR #3034, opened 2026-04-06
[TSP] Don't synthesize TSP types if you can avoid it.

rchiodo:
# Summary 

## Why
The TSP type conversion layer was emitting SynthesizedType with generated Python stub content for most pyrefly Type variants. This made the protocol output opaque to consumers — they received stringified stubs instead of structured, navigable type information with proper source declarations.

## What
Replace synthesized stub generation with declaration-based TSP types for all 49 Type variants. Every variant now maps to its closest TSP protocol equivalent (ClassType, FunctionType, UnionType, ModuleType, DeclaredType, BuiltInType, or OverloadedType) with real source declarations pointing to definition sites.

## Key conversions:

-Function/BoundMethod/Callable = FunctionType with declaration + return type
-ClassType/ClassDef/Tuple/Tensor/NNModule/TypedDict = ClassType with RegularDeclaration pointing to the class definition
-Overloaded = OverloadedType with per-overload FunctionType entries
-TypeVar/ParamSpec/TypeVarTuple = DeclaredType (TypeVar kind)
-Union = UnionType with recursively converted members
-Module = ModuleType with module name
-Literals = ClassType with literalValue and LITERAL flag
-Solver-internal types (Var, Materialization, Args, etc.) = BuiltInType with representative names
-Wrapper types (Forall, Annotated, TypeAlias, TypeGuard, etc.) = unwrapped to their inner type

Additionally, FuncId now carries a name_range: TextRange field so that function declarations (including individual overload signatures) emit real source locations instead of zero ranges.

## Why it works
SynthesizedType and its stub generation code (collect_type_names, generate_function_stub) are removed from production paths — synthesized() is now #[cfg(test)] only.
Each Type variant is explicitly matched with no catch-all fallback, so new variants will cause a compile error rather than silently degrading.
FuncId.name_range is populated from def.name.range at construction time in the binding phase, and excluded from identity/hashing so it doesn't affect type equality.

# Tests
Added 22 integration tests covering: tuple, dict, set, module import, bound method, overloaded functions (return type correctness + distinct declaration lines), function return type, function/class declarations with name and URI, class instances with INSTANCE flag, type aliases, Annotated unwrapping, union subtypes, Callable, TypedDict, Optional, lambda, generic class type args, and literal values.

/cc @kinto0

### DIFF
--- a/pyrefly/lib/tsp/type_conversion.rs
+++ b/pyrefly/lib/tsp/type_conversion.rs
@@ -35,6 +35,7 @@ use std::sync::atomic::Ordering;
 
 use lsp_types::Url;
 use pyrefly_types::callable::Callable;
+use pyrefly_types::callable::FuncId;
 use pyrefly_types::callable::FunctionKind;
 use pyrefly_types::class::Class;
 use pyrefly_types::class::ClassType as PyreflyClassType;
@@ -42,6 +43,7 @@ use pyrefly_types::literal::Lit;
 use pyrefly_types::types::BoundMethodType;
 use pyrefly_types::types::Forallable;
 use pyrefly_types::types::Type as PyreflyType;
+use ruff_text_size::TextRange;
 use tsp_types::BuiltInType;
 use tsp_types::ClassType as TspClassType;
 use tsp_types::Declaration;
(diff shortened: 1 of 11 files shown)


pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -128,37 +161,66 @@ pub fn convert_type(ty: &PyreflyType) -> TspType {
...
-                        convert_function(&f.body.signature, &f.body.metadata.kind)
-                    }
+        PyreflyType::Overload(overload) => convert_overload_to_tsp(overload, None),
+
+        // --- Forall (generic functions/callables) — unwrap body ---
+        PyreflyType::Forall(forall) => match &forall.body {
+            Forallable::Function(f) => convert_function(&f.signature, &f.metadata.kind, None),
+            Forallable::Callable(c) => {
+                let ret = convert_type(&c.ret);
+                TspType::Function(TspFunctionType {
```
kinto0: nit: duplicates code on line 110

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -277,157 +410,118 @@ fn convert_literal(lit: &pyrefly_types::literal::Literal) -> TspType {
...
-/// stub declarations.
-fn collect_callable_type_names(callable: &Callable, names: &mut HashSet<String>) {
-    match &callable.params {
-        Params::List(params) => {
-            for param in params.items() {
-                collect_type_names(param.as_type(), names);
+/// Convert a pyrefly `Overload` to a TSP `OverloadedType`.
+fn convert_overload_to_tsp(
+    overload: &pyrefly_types::types::Overload,
+    bound_to_type: Option<Box<TspType>>,
```
kinto0: nit: can this be an `arc` so we don't have to clone bound_to_type on every overload?

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -128,37 +161,66 @@ pub fn convert_type(ty: &PyreflyType) -> TspType {
...
                 })
-                .collect();
-            TspType::Overloaded(TspOverloadedType {
-                flags: TypeFlags::CALLABLE,
+            }
+            Forallable::TypeAlias(ta) => match ta {
+                pyrefly_types::type_alias::TypeAliasData::Value(alias) => {
+                    convert_type(&alias.as_type())
+                }
+                pyrefly_types::type_alias::TypeAliasData::Ref(r) => builtin(&r.name.to_string()),
```
kinto0: nit: I don't think you need this to_string

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -175,8 +237,89 @@ pub fn convert_type(ty: &PyreflyType) -> TspType {
...
+            TspType::Var(make_typevar_declared(qname))
+        }
+        PyreflyType::TypeVarTuple(tvt) => {
+            let qname = tvt.qname();
+            TspType::Var(make_typevar_declared(qname))
+        }
+
+        // --- Quantified / QuantifiedValue (type params during solving) ---
+        PyreflyType::Quantified(q) | PyreflyType::QuantifiedValue(q) => {
+            builtin(&q.name.to_string())
```
kinto0: nit: to_string() unnecessary

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -175,8 +237,89 @@ pub fn convert_type(ty: &PyreflyType) -> TspType {
...
+
+        // --- NNModule → ClassType from class ---
+        PyreflyType::NNModule(m) => convert_class_type(&m.class, TypeFlags::INSTANCE),
+
+        // --- TypeAlias → unwrap to the aliased type, or builtin for refs ---
+        PyreflyType::TypeAlias(ta) | PyreflyType::UntypedAlias(ta) => match ta.as_ref() {
+            pyrefly_types::type_alias::TypeAliasData::Value(alias) => {
+                convert_type(&alias.as_type())
+            }
+            pyrefly_types::type_alias::TypeAliasData::Ref(r) => builtin(&r.name.to_string()),
```
kinto0: nit: to_string unnecessary

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -175,8 +237,89 @@ pub fn convert_type(ty: &PyreflyType) -> TspType {
...
+        PyreflyType::Unpack(inner) => convert_type(inner),
+
+        // --- TypeForm(X) → convert inner ---
+        PyreflyType::TypeForm(inner) => convert_type(inner),
+
+        // --- Intersect → convert the fallback type ---
+        PyreflyType::Intersect(pair) => convert_type(&pair.1),
+
+        // --- ElementOfTypeVarTuple → builtin with name ---
+        PyreflyType::ElementOfTypeVarTuple(q) => builtin(&q.name.to_string()),
```
kinto0: nit: to_string unnecessary

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -175,8 +237,89 @@ pub fn convert_type(ty: &PyreflyType) -> TspType {
...
+
+        // --- ParamSpec-related internal types → builtin with name ---
+        PyreflyType::Args(q)
+        | PyreflyType::Kwargs(q)
+        | PyreflyType::ArgsValue(q)
+        | PyreflyType::KwargsValue(q) => builtin(&q.name.to_string()),
```
kinto0: nit: to_string unnecessary

Issue #2859, opened 2026-03-23
Generated typestubs are not found in an editable install


jbcoe:
When typestubs are generated for a Python extension, they are not found in an editable install, although they are installed to the appropriate location in site-packages along with the DSO for the extension.

When the same package is installed in a non-editable context, typestubs are found correctly and type errors reported as expected.

The linked repository https://github.com/jbcoe/xyz-generated-typestubs has a minimal (but still not tiny) reproduction where two script can be run to illustrate the problem.

`./check_in_project.sh` incorrectly returns type errors because it cannot find type stubs.

`./check_out_of_project.sh` correctly returns type errors in a non-project file that tries to use a function with the wrong types.

I suspect that pyrefly’s import discovery logic for editable installs needs to be updated to account for generated typestubs located alongside the extension's binary.

## Note
A similar bug was raised with `ty` https://github.com/astral-sh/ty/issues/2911. That was fixed with a simple (if somewhat non-obvious) config change.

migeed-z:
cc @connernilsen do you have context on this issue?

NathanTempest:
We created a minimal reproduction that demonstrates this bug:

https://github.com/facebook/pyrefly/pull/3035

**Setup:**
- `mypackage/_core.py` — module with NO type hints (simulates compiled extension)
- `stubs/mypackage/_core.pyi` — stub defining `add(a: int, b: int) -> int`
- `pyrefly.toml` — config with `search-path = ["stubs"]`

**To reproduce: copy the changes in the PR and then run these commands**
```bash
cd test/repro/stub_overlay_2859
pyrefly check test_script.py

Expected: 4 errors (2 for add("hello", "world") + 2 for sub("foo", "bar")) Actual: 2 errors (only sub() errors — stub not used for add())
```
The root cause is that Pyrefly finds the .py source file but doesn't overlay the companion .pyi stub types, even with search-path configured.

cc @connernilsen  @jvansch1

PR #3035, opened 2026-04-06
Add minimal reproduction for stub overlay issue #2859

NathanTempest:
## Summary

This PR adds a minimal reproduction test for issue #2859: Generated typestubs are not found for untyped modules.

## The Problem

When a module has no inline type hints (common for compiled extensions like `.so`/`.pyd` files), Pyrefly doesn't use companion `.pyi` stub files to supplement type information, even when the stub directory is configured in `search-path`.

## Test Description

The test creates:
- `mypackage/_core.py` - module with NO type hints (simulates compiled extension)
- `mypackage/__init__.py` - re-exports `add()` and defines typed `sub()` function
- `stubs/mypackage/_core.pyi` - stub file with type hints for `add(a: int, b: int) -> int`
- `test_script.py` - calls both `add("hello", "world")` and `sub("foo", "bar")`

**Current (buggy) behavior:**
- Only 2 errors reported (for `sub()` calls which has inline type hints)
- `add("hello", "world")` does NOT error even with `search-path = ["stubs"]`

**Expected behavior when fixed:**
- 4 errors reported (2 for `add()` using stub types + 2 for `sub()` using inline types)

## Test Plan

```bash
PYREFLY=./target/release/pyrefly scrut test test/stub_overlay.md

### DIFF
--- a/test/repro/stub_overlay_2859/README.md
+++ b/test/repro/stub_overlay_2859/README.md
@@ -1,0 +1,49 @@
+# Pyrefly Issue #2859 - Minimal Reproduction
+
+This directory contains a minimal reproduction for issue #2859: **Generated typestubs are not found for untyped modules**.
+
+**Issue:** https://github.com/facebook/pyrefly/issues/2859
+
+## The Bug
+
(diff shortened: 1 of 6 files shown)


NathanTempest:
This PR is created as a minimal repro for the issue - https://github.com/facebook/pyrefly/issues/2859, this does not need to be merged into the repo and is just acting like a reference.

PR #3039, opened 2026-04-07
[LSP] use source.fixAll.pyrefly for fix-all code actions

### LINKED ISSUE #3024: [LSP] `source.fixAll` code action should be `source.fixAll.pyrefly`
### Describe the Bug

Pyrefly advertizes support for `source.fixAll` code action in the `initialize` response:

```json
{
  "capabilities": {
    "callHierarchyProvider": true,
    "codeActionProvider": {
      "codeActionKinds": [
        "quickfix",
        "refactor.extract",
        "refactor.rewrite",
        "refactor.delete",
        "refactor.move",
        "refactor.inline",
        "source.fixAll"
      ]
    },
  },
  // ...
}
```

This should actually be `source.fixAll.pyrefly` which then would allow the editor to only target purefly-specific fixes with:

```json
    "editor.codeActionsOnSave": {
        "source.fixAll.pyrefly": true
    },
```

Otherwise it's only possible to use:

```json
    "editor.codeActionsOnSave": {
        "source.fixAll": true
    },
```

which then can trigger fixes from other servers that might be running on the file.

### Sandbox Link

_No response_

_No response_

Arths17:
## Summary
- advertise `source.fixAll.pyrefly` in server capabilities instead of generic `source.fixAll`
- emit fix-all actions with the pyrefly-specific kind
- for `context.only`, run pyrefly fix-all only when the request includes either `source.fixAll` or a `source.fixAll.pyrefly...` kind

Fixes #3024.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -4174,9 +4174,7 @@ impl Server {
         let allow_fix_all = only_kinds.is_none_or(|kinds| {
             kinds.iter().any(|kind| {
                 kind == &CodeActionKind::SOURCE_FIX_ALL
-                    || kind
-                        .as_str()
-                        .starts_with(CodeActionKind::SOURCE_FIX_ALL.as_str())
+                    || kind.as_str().starts_with(SOURCE_FIX_ALL_PYREFLY)
             })
         });
         let allow_refactor = only_kinds.is_none_or(|kinds| {
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -681,7 +681,11 @@ fn format_diagnostic_message_for_markdown(message: &str) -> String {
 
 #[cfg(test)]
 mod tests {
+    use lsp_types::CodeActionKind;
+
+    use super::SOURCE_FIX_ALL_PYREFLY;
     use super::format_diagnostic_message_for_markdown;
+    use super::is_fix_all_code_action_kind_requested;
 
     #[test]
     fn test_format_diagnostic_message_for_markdown() {
@@ -721,6 +725,26 @@ mod tests {
     #[test]
     fn test_format_only_special_characters() {
         assert_eq!(format_diagnostic_message_for_markdown("***"), "\\*\\*\\*");
+    }
+
+    #[test]
+    fn test_fix_all_kind_filter_matches_supported_kinds() {
+        assert!(is_fix_all_code_action_kind_requested(
+            &CodeActionKind::SOURCE_FIX_ALL
+        ));
+        assert!(is_fix_all_code_action_kind_requested(&CodeActionKind::new(
+            SOURCE_FIX_ALL_PYREFLY,
+        )));
+    }
+
+    #[test]
+    fn test_fix_all_kind_filter_rejects_pyrefly_suffix_kinds() {
+        assert!(!is_fix_all_code_action_kind_requested(
+            &CodeActionKind::new("source.fixAll.pyrefly.foo",)
+        ));
+        assert!(!is_fix_all_code_action_kind_requested(
+            &CodeActionKind::new("source.fixAll.pyreflyyyyyy",)
+        ));
     }
 }
 
@@ -1253,6 +1277,10 @@ pub enum ProcessEvent {
 
 const PYTHON_SECTION: &str = "python";
 const SOURCE_FIX_ALL_PYREFLY: &str = "source.fixAll.pyrefly";
+
+fn is_fix_all_code_action_kind_requested(kind: &CodeActionKind) -> bool {
+    kind == &CodeActionKind::SOURCE_FIX_ALL || kind.as_str() == SOURCE_FIX_ALL_PYREFLY
+}
 
 struct TypeHierarchyTarget {
     def_index: ClassDefIndex,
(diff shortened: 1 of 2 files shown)


Arths17 (commented):
Force-push update: this PR now contains only the intended LSP fix for #3024 (2 files changed: `server.rs` and the corresponding LSP interaction test). The earlier Copilot comments on `perf/scale_test/*` refer to outdated threads from a prior diff and are no longer applicable to the current head commit.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -4171,9 +4172,10 @@ impl Server {
         let allow_quickfix = only_kinds
             .is_none_or(|kinds| kinds.iter().any(|kind| kind == &CodeActionKind::QUICKFIX));
         let allow_fix_all = only_kinds.is_none_or(|kinds| {
-            kinds
-                .iter()
-                .any(|kind| kind == &CodeActionKind::SOURCE_FIX_ALL)
+            kinds.iter().any(|kind| {
+                kind == &CodeActionKind::SOURCE_FIX_ALL
+                    || kind.as_str().starts_with(SOURCE_FIX_ALL_PYREFLY)
```
rchl: I think that the second condition should also match `source.fixAll.pyrefly` exactly rather than using `starts_with` because `source.fixAll.pyreflyyyyyy` shouldn't match, nor should `source.fixAll.pyrefly.foo`

NathanTempest:
The PR correctly updates all 4 locations (constant, capability registration, action tagging, and test). Nice clean change.

Agree with @rchl's comment on the filter matching — starts_with(SOURCE_FIX_ALL_PYREFLY) should be an exact match instead:

```
kinds.iter().any(|kind| {
    kind == &CodeActionKind::SOURCE_FIX_ALL
        || kind.as_str() == SOURCE_FIX_ALL_PYREFLY
})
```

starts_with would incorrectly match `source.fixAll.pyreflyyyyyy` or `source.fixAll.pyrefly.foo` The first condition (SOURCE_FIX_ALL) correctly handles the hierarchical parent case where a client requests all fix-all actions generically.

After making this change, the PR should be good to merge.

Arths17:
Addressed the review feedback on fix-all kind matching and pushed an update.

What changed:
- switched `context.only` fix-all filtering from prefix matching to exact matching for `source.fixAll.pyrefly`
- kept support for generic `source.fixAll`
- added regression tests covering accepted kinds (`source.fixAll`, `source.fixAll.pyrefly`) and rejected kinds (`source.fixAll.pyrefly.foo`, `source.fixAll.pyreflyyyyyy`)

Validation run locally:
- `cargo test test_fix_all_kind_filter`

Latest commit: `bd89c04cd`

Arths17:
@rchl @NathanTempest re-requesting review after the latest push (`bd89c04cd`) that applies the exact-match fix for `source.fixAll.pyrefly` filtering and adds regression coverage for invalid suffix kinds. Could you take another look?

NathanTempest (changes requested):
The PR looks solid to merge but I have a few nitpicks - 

1. Missing negative test for unrelated kinds

The unit tests verify that source.fixAll and source.fixAll.pyrefly are accepted, and that source.fixAll.pyrefly.foo / source.fixAll.pyreflyyyyyy are rejected. But they don't test that completely unrelated kinds like quickfix or refactor.extract are rejected. Adding one line would improve confidence:

```
assert!(!is_fix_all_code_action_kind_requested(
    &CodeActionKind::QUICKFIX
));
```

2. No integration test for the actual code action response kind

The `basic.rs` test verifies the server advertises `source.fixAll.pyrefly` in capabilities, but no test verifies that when a fix-all action is actually returned, it carries kind: "source.fixAll.pyrefly" (not the old "source.fixAll"). A lightweight LSP interaction test that triggers a redundant cast scenario and asserts the returned code action kind would close this gap. This is pre-existing technical debt, but since the PR is touching this exact area, it would be the ideal time to add it.

3. Function naming

`is_fix_all_code_action_kind_requested` is verbose. Since it takes a single CodeActionKind and checks if it matches fix-all semantics, something like is_fix_all_kind or matches_fix_all_kind would be more concise while equally clear. This can be ignored but it is just nicer naming.

NathanTempest (approved):

NathanTempest:
Approved to merge, but would be elated if there is a follow up PR that would address the first two nits.

kinto0 (changes requested):

NathanTempest:
Hey @Arths17, the code changes look correct and address #3024 well. Before we merge, could you flesh out the PR description with some additional context? This helps future contributors understand the "why" behind the change.

A brief explanation of the LSP code action kind hierarchy (source.fixAll → source.fixAll.<tool>) and why tool-specific namespacing matters. Why does this matter specifically for source.fixAll but not for other kinds like `quickfix` or `refactor.*` ? (i.e., source.fixAll is auto-triggered on save, so users need per-tool control)

Was there consideration of advertising both source.fixAll and source.fixAll.pyrefly in capabilities?

**Test plan**
Could you add a short video or step-by-step showing the end-to-end behavior in VS Code?
A file with redundant casts showing pyrefly diagnostics
Triggering the code action and confirming it appears with the correct kind
Configuring "source.fixAll.pyrefly": "always" in codeActionsOnSave and demonstrating fix-on-save
This complements the unit tests by showing the change works at the editor level.
The implementation itself is solid — this is just about making the PR self-documenting. Thanks for the contribution!

To make your test plan easy or to give you a good reference, please check my comment - https://github.com/facebook/pyrefly/issues/3024#issuecomment-4290851057

rchl:
Man, there is some bureaucracy in Meta/this project...
Don't envy ones who submit patches here.

kinto0 (approved):

PR #3044, opened 2026-04-07
fix Regression when using type alias on Enum #3042

### LINKED ISSUE #3042: Regression when using type alias on Enum
### Describe the Bug

The following type-checked fine with Pyrefly 0.59.1:

```py
from typing import Protocol, Sequence, reveal_type, Literal
from enum import Enum

class MyEnum(Enum):
    a = 0


a = MyEnum.a
A = Literal[MyEnum.a]

def func(var: str | A) -> str:
    if var is not MyEnum.a:
        return var  # this is fine...
    if var is not a:
        return var  # ...but pyrefly complains about this one
    return 'foo'
```

But, since 0.60.0, it raises
```
ERROR Returned type `MyEnum` is not assignable to declared return type `str` [bad-return]
  --> t.py:15:16
   |
11 | def func(var: str | A) -> str:
   |                           --- declared return type
12 |     if var is not MyEnum.a:
13 |         return var
14 |     if var is not a:
15 |         return var
   |                ^^^
   |
 INFO 1 error
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AFag1wBjXFAA0dAMowAjgFcY6UTGmUYANxiooAfSbF1dADIQGMSnoA66MNXorF9Ljz50Aouhd27oqFQ4ODoAWVJvFwAKSJoASkQ7OmS6VDoAXjoABj90OzTM8NjCVDsAQQyzCys9AG0inxoSgF1czBgwOjBFVSitVEpEOjgGSjoAHzoyuLoAWgA%2BYdHE9BTOTv6xiBD0XH4GlxKVtbXNBkVKVc2klIgNgc4dvdTjk%2BSzi6uBm-eYc8u6AByMC4XCA3IgSQgMiaMBQUiEEQ0KAUADEglIsPhwwwOAIdHE9ggbAuqAYEFw6EIdnRchgdAAFgwGMQ4IgAPTsmEdeGEXhsdkqdmYMRwdmEyAk6zkynsrq8VL9aCobCwAmUyWkmWrXDEbVwanoMgMBmU2Y6ShwCmrTI2EAAZkIAEYAEx2uy1KzUS2tHzoNy8SyYWaYCCaUTknSVQHE3aacHoXYMWaaJRhmDBgDWMFIs1QojUwWjAHcBugEyAAL5Q-ORmAAMWgMAoaCweCIZCrQA

_No response_

asukaminato0721:
# Summary

Fixes #3042

narrowing now infers comparison RHS names from their underlying binding instead of the function-scope promoted view, so a module alias like a = E.X still behaves as the literal enum member during is/== narrowing.

keeps the existing promotion behavior out of the general expression path and only makes narrowing more precise where it matters.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -12,6 +12,7 @@ use pyrefly_config::error_kind::ErrorKind;
 use pyrefly_graph::index::Idx;
 use pyrefly_python::ast::Ast;
 use pyrefly_python::dunder;
+use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_types::class::Class;
 use pyrefly_types::display::TypeDisplayContext;
 use pyrefly_types::facet::FacetChain;
@@ -127,6 +128,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 })
                 .collect::<Vec<_>>(),
         )
+    }
+
+    fn narrow_expr_infer(&self, expr: &Expr, errors: &ErrorCollector) -> Type {
+        match expr {
+            Expr::Name(name) if !Ast::is_synthesized_empty_name(name) => self
+                .get(&Key::BoundName(ShortIdentifier::expr_name(name)))
+                .arc_clone_ty(),
+            _ => self.expr_infer(expr, errors),
+        }
     }
 
     /// Return the most specific disjoint base for a type per PEP 800.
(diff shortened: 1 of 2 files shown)


kinto0:
do we understand which commit caused the regression? and why? I think that's an important step in understanding if this fix is the correct one or not

stroxler:
I commented on the original issue - the change was caused by us treating (lower-cased) globals as mutable

That was intentional, and I think we should consider our options for solutions carefully

stroxler:
Since Marco said our behavior isn't really problematic, I think we should probably just stick with the status quo for now. Thanks for the PR!

PR #3050, opened 2026-04-07
check non-callable __bool__ in assert statements

### LINKED ISSUE #2914: Non-callable __bool__ attribute
When `__bool__` is defined as a non-callable attribute (e.g., `int`), using the object in a boolean context should error.

```python
class BadBool:
    __bool__: int = 3

assert BadBool()
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksiAMZTpxzIBC62z%2B%2BUiAOpsgOQB9IQS4iUETABdkAXmQBmPnwZwYAJ1mt2nKAAoAlHxABfADTV0NaRABuMAGLQYVEBhwESFM0A)

knQzx:
## summary
- assert statements now check that __bool__ is callable on the test expression
- previously only if/while/ternary checked this, assert was missing the check
- matches behavior of pyright and mypy which flag this

## test plan
- updated existing bug test case to expect the error
- all 4678 tests pass

fixes #2914

### DIFF
--- a/pyrefly/lib/binding/stmt.rs
+++ b/pyrefly/lib/binding/stmt.rs
@@ -120,7 +120,9 @@ impl<'a> BindingsBuilder<'a> {
         self.ensure_expr(&mut test, &mut Usage::Narrowing(None));
         let narrow_ops = NarrowOps::from_expr(self, Some(&test));
         let static_test = self.sys_info.evaluate_bool(&test);
+        let test_clone = test.clone();
         self.insert_binding(Key::Anon(test_range), Binding::Expr(None, Box::new(test)));
+        self.insert_binding(KeyExpect::Bool(test_range), BindingExpect::Bool(test_clone));
         if let Some(mut msg_expr) = msg {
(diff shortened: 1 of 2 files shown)


NathanTempest (approved):
Verified the change matches the established pattern: if (around stmt.rs:1052-1062) and while (around stmt.rs:968-978) both insert a BindingExpect::Bool to trigger the existing __bool__ callability check. The new code in assert does exactly the same. The test.clone() is necessary because the original test is moved into Binding::Expr(None, Box::new(test)) on the next line.

The test conversion (removing the bug = "..." marker and adding the expected error annotation) is appropriate — the existing test now actively verifies the fix.

mypy_primer is clean, behavior matches pyright/mypy, and the fix is 2 lines of code.

I'll import this internally and assign it to myself for review.

stroxler (approved):

PR #3051, opened 2026-04-07
check non-callable __bool__ on in operator result

### LINKED ISSUE #2913: Non-callable __bool__ in membership tests
When `__contains__` returns an object with a non-callable `__bool__`, the `in` operator should produce a bool-conversion error.

```python
class BadBool:
    __bool__: int = 3

class Container:
    def __contains__(self, item: object) -> BadBool:
        return BadBool()

10 in Container()
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksiAMZTpxzIBC62z%2B%2BUiAOpsgOQB9IQS4iUETABdkAXmQBmPnzoMmAYXwz0UmACde-QdhhhhQmtum7McEQAo4MKGAA0yCNJgBbFPlwAKxgaaQBKZABaAD4WNg4uI0Fk5H0YaQBXfX5Wdk4oBzCVTABGAAZPfi0dPX1CvhAAXw8QdFCIADcYADFoGCoQDBwCEgomoA) | [Pyright](https://pyright-play.net/?code=MYGwhgzhAEBCYBNYHtkgFwCho%2BgfTwCNUQD1oBLAOwBdoBeaAZk01EhgGFlaxqBTAE5Zc0BPwBm%2BPMB40%2BVCAQAUEfiAkAaSjX4BbcskIArfsBoBKaAFoAfHEQo0I0bkH8aAV0FUHSEsoWrACMAAyUvty8AoKBmEA)

knQzx:
## summary
- in/not in operator now checks that __bool__ is callable on the __contains__ return type
- also returns bool instead of the raw __contains__ return type, matching python semantics

## test plan
- updated existing bug test case to expect the error
- all 4678 tests pass

fixes #2913

### DIFF
--- a/pyrefly/lib/alt/operators.rs
+++ b/pyrefly/lib/alt/operators.rs
@@ -767,7 +767,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                                     errors,
                                     Some(&context),
                                 ) {
-                                    ret
+                                    self.check_dunder_bool_is_callable(&ret, x.range, errors);
+                                    self.heap.mk_class_type(self.stdlib.bool().clone())
                                 } else {
                                     let iteration_errors = self.error_collector();
                                     let iterables = self.iterate(
--- a/pyrefly/lib/test/operators.rs
+++ b/pyrefly/lib/test/operators.rs
@@ -1068,3 +1068,20 @@ class Container:
 10 in Container()  # E: The `__bool__` attribute of `BadBool` has type `int`, which is not callable
 "#,
 );
+
+testcase!(
+    test_in_operator_returns_bool,
+    r#"
+from typing import assert_type
+
+class Truthy:
+    def __contains__(self, item: object) -> int:
+        return 1
+
+assert_type(10 in Truthy(), bool)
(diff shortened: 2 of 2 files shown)


NathanTempest (changes requested):
The PR looks good to merge, but can you address the missing test for the return type change. This changes the return type of x in y from the raw __contains__ return type to bool. This is a semantic change beyond just the error check, but there's no test that validates the result type. For example:
```
class Truthy:
    def __contains__(self, item: object) -> int:
        return 1

x: bool = 10 in Truthy()  # Should now pass (returns bool, not int)
y: int = 10 in Truthy()   # Should this also pass? (bool is a subtype of int)
```

Before this PR, x in y where __contains__ returns int would type the expression as int. After this PR, it types as bool. This is correct per Python semantics, but the behavioral change deserves a test.

knQzx:
@NathanTempest added a test for the return type, thanks for the review

NathanTempest (approved):

samwgoldman (approved):

PR #3071, opened 2026-04-08
Support CodeClimate issues in `check --output-format`

### LINKED ISSUE #3049: Add GitLab Code-Quality report format
### Describe the Bug

GitLab can read a _Code Quality_ file in JSON format, which can be used to put the findings from `pyrefly` into the context of a merge request and use that as a quality gate: https://docs.gitlab.com/ci/yaml/artifacts_reports/#artifactsreportscodequality

Most other type checkers either provide an option to directly generate the required format **or** some external tool can be used to convert some other machine readable format into the required JSON:
- `mypy`: https://pypi.org/project/mypy-gitlab-code-quality/
- `pyright` https://www.npmjs.com/package/pyright-to-gitlab-ci
- `ty --output-format gitlab`

I would appreciate a similar feature for `pyrefly`.

The format understood by GitLab is a subset of [CodeClimate report format](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types). The GitLab dialect is described at https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format

### Sandbox Link

_No response_

_No response_

coclique:
# Summary

Add a new option `codeclimate` for `pyrefly check --output-format`, which outputs a JSON array of [CodeClimate issues](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#issues). This is useful primarily for GitLab CI/CD to pick up a [Code Quality Report artifact](https://docs.gitlab.com/ci/testing/code_quality).

Fixes #3049

# Test Plan

A scrut test for this option was added.


connernilsen (approved):
Hey @re-xyr, thanks for making this. One minor change I'll make on our end is to rename `OutputFormat::Codeclimate` to `OutputFormat::CodeClimate`. Let me know if that can't be done for some reason and we can follow up to change the naming scheme.

grievejia (changes requested):

crates/pyrefly_config/src/config.rs
```diff
@@ -133,6 +133,8 @@ pub enum OutputFormat {
     Json,
     /// Emit GitHub Actions workflow commands
     Github,
+    /// Emit CodeClimate issues in a JSON array (e.g. for GitLab Code Quality reports)
+    Codeclimate,
```
connernilsen: Can you make this caps case (`CodeClimate`)?

pyrefly/lib/error/codeclimate.rs
```diff
@@ -0,0 +1,119 @@
+/*
```
connernilsen: Can you rename this file to `code_climate.rs`?

pyrefly/lib/error/codeclimate.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::path::Path;
+
+use pyrefly_config::error_kind::Severity;
+use pyrefly_util::prelude::SliceExt;
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::error::error::Error;
+
+pub(crate) fn severity_to_str(severity: Severity) -> String {
```
connernilsen: Can you rename this to `severity_to_code_climate_str` or make it private to this crate? This might be confusing with `legacy::severity_to_str`, so swapping the imports would break things elsewhere.

Can you also have it return a `&'static str`?

pyrefly/lib/error/codeclimate.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl CodeClimateIssue {
+    pub fn from_error(relative_to: &Path, error: &Error) -> Self {
+        let error_range = error.display_range();
+        let error_path = error.path().as_path();
+
+        let mut hasher = DefaultHasher::new();
+        error.hash(&mut hasher);
+        let fingerprint = format!("{:x}", hasher.finish());
```
connernilsen: This is non-deterministic across runs, so the fingerprint won't be consistent. This is because the `Error` struct uses an `ArcId`, which is based on memory addresses.

Also, there's no guarantee the hasher used here will be stable across runs, so let's use MD5 or SHA-256

test/snippet.md
```diff
@@ -119,6 +119,38 @@ $ $PYREFLY snippet "x: int = 'hello'" --output-format=json
...
+          "column": 10
+        },
+        "end": {
+          "line": 1,
+          "column": 17
+        }
+      }
+    },
+    "severity": "major",
+    "fingerprint": "*" (glob)
```
connernilsen: We should get rid of the glob here, since we want to make sure the fingerprint isn't changing between runs.

pyrefly/lib/error/codeclimate.rs
```diff
@@ -0,0 +1,119 @@
...
+            content: error.msg_details().map(|details| CodeClimateIssueContent {
+                body: details.to_owned(),
+            }),
+            categories: vec!["Bug Risk".to_owned()],
+            location: CodeClimateIssueLocation {
+                path: error_path
+                    .strip_prefix(relative_to)
+                    .unwrap_or(error_path)
+                    .to_string_lossy()
+                    .into_owned(),
```
connernilsen: This won't show the location in a notebook, and `legacy.rs` will append a `#<cell>` with `path_string_with_fragment()` in `error.rs`. Can we either do that, or switch to `line_within_file()`?

This will also keep `\\` path separators on windows. Can we normalize this like we do with [`github_actions_path`](https://github.com/facebook/pyrefly/blob/9d758b762bdce6081ec0c0ac0fba86cb83762667/pyrefly/lib/commands/check.rs#L537-L543)? Maybe pull that into `pyrefly_util` somewhere.

pyrefly/lib/error/codeclimate.rs
```diff
@@ -0,0 +1,119 @@
...
+        let fingerprint = format!("{:x}", hasher.finish());
+
+        Self {
+            issue_type: "issue".to_owned(),
+            check_name: format!("Pyrefly/{}", error.error_kind()),
+            description: error.msg_header().to_owned(),
+            content: error.msg_details().map(|details| CodeClimateIssueContent {
+                body: details.to_owned(),
+            }),
+            categories: vec!["Bug Risk".to_owned()],
```
connernilsen: `reveal_type`/`assert_type` are always displayed. This will surface those as errors in code climate, and might be pretty noisy. Can we either not emit errors for those or give them `categories: ["Clarity"], severity: "info"`?

connernilsen (changes requested):
@grievejia had a few requests internally. Would you be able to update this? Afterward, I'll re-pull this in and review.

crates/pyrefly_config/src/config.rs
coclique: This changes the CLI option value to `code-climate`. Let me know if this is nonideal.

pyrefly/lib/error/codeclimate.rs
coclique: I used the more performant XxHash3; this adds a `xxh3` flag to the `xxhash-rust` crate. If you'd prefer not change the build configuration, I can switch to FxHash instead.

pyrefly/lib/error/codeclimate.rs
coclique: Factored out `pyrefly_util::unix_path::path_to_unix_string()`.

pyrefly/lib/error/codeclimate.rs
coclique: I'm not quite sure about `assert_type`; doesn't it only fire when the assertion fails (therefore an actual failure)? In addition, `reveal_type` is already `info` by default.

Currently I left the severity untouched and changed the category to `"Clarity"` on directives. Please advise.

coclique:
Sorry about the delay in follow up, I was busy with other matters. I addressed the review comments to the best of my ability.

grievejia (approved):

PR #3076, opened 2026-04-08
fix str not assignable to LiteralString attribute #3057

### LINKED ISSUE #3057: str not assignable to LiteralString attribute
### Describe the Bug

Minimal repro:
```python
from typing import Literal

class SessionMiddleware:
    def __init__(
        self, same_site: Literal["lax", "strict", "none"] = "lax", domain: str | None = None
    ) -> None:
        self.security_flags = "httponly; samesite=" + same_site
        if domain is not None:
            self.security_flags += f"; domain={domain}"  # FP: str not assignable to LiteralString
```
No other type checker wants to error here

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AZCAxiVUUADropAYyio4cOgGV4cCLnQBZCJkywA7qkoxEUuhbqYYYOgH07rYQ4AU5yx7gwoYADR00Ghg7dRFEQWFRcQBtCRB5fDj-OLgGSggZBiS6OPRNGDiAXToAXhz41ESQf0xaVFZw1Mo6AB86ADl80o789wsASjoAWgA%2BHvRTPo8A7zBCLxkAV3SmOzB5NiUyuIALBgYedChSAG4A1CDQmBK4ugBqc6CQyKmPCFtamnr0TiU8-k6EzMP2moK8PnmMCWK1Iaw2SjuZTAcTOn2%2BJWAaNYAF9bnQAMR0ABiAAVGmk6P86Ap1Gx0KhsLBGLgIiIxFBlGlWGxqiBFgxoHASORECBCQBVAVQYSkOhgRboTIadBwKRSay2MC8L4MOzoRY0bCiFz4cKsBiDUYBNLAjwmBjLH7IkDtA1GyjhYD4XEgKS8sgmdakQgMWhQCiEkmkQPHc5YPD4OgyTSQNjLVACzSEKSE1QwOh7A5wRAAehLAZsx0IvDYJZg6BLtRkcBLyfQqfTmYbct41IAbvV5Iz822O2Iu3RcMQu0KpGQGDtNEM%2B6J1JpunEAMyEACMACY4n7sb4QKglSuidAYBQ0PGCMKQNigA

_No response_

asukaminato0721:
# Summary

Fixes #3057

stops freezing a method-defined instance attribute at its first self.attr write and instead tracks later writes, including self.attr += ..., then widens the inferred field type across those writes.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -63,9 +63,9 @@ use crate::binding::binding::ClassFieldDefinition;
 use crate::binding::binding::ExprOrBinding;
 use crate::binding::binding::KeyClassField;
 use crate::binding::binding::KeyClassSynthesizedFields;
-use crate::binding::binding::MethodFieldWrites;
 use crate::binding::binding::MethodSelfKind;
 use crate::binding::binding::MethodThatSetsAttr;
+use crate::binding::scope::is_constant_name;
 use crate::config::error_kind::ErrorKind;
 use crate::error::collector::ErrorCollector;
 use crate::error::context::ErrorContext;
@@ -1631,7 +1631,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 )
             }
             ClassFieldDefinition::DefinedInMethod {
-                values,
+                value,
                 method,
                 annotation: annot,
                 ..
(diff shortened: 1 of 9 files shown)


rchen152 (changes requested):
This seems to be going in the same direction as https://github.com/facebook/pyrefly/pull/2356, which Danny wasn't able to merge due to mypy_primer regressions. That suggests to me that collecting all writes to infer a type is likely the wrong approach. What happens if you just try promoting LiteralString to str when inferring attribute types? We might need special-casing for ALL_CAPS constants, which we already have elsewhere to avoid promoting constant module attributes.

rchen152:
After looking into this more carefully, we already promote implicit literals in class_field.rs (https://github.com/facebook/pyrefly/blob/abf9da23c9cb0980b52871f380c9c48c2d22524f/pyrefly/lib/alt/class/class_field.rs#L1801), and the reason LiteralString wasn't being promoted is that pyrefly incorrectly marked it as an "explicit" literal. The fix is to extend the string addition case here: https://github.com/facebook/pyrefly/blob/abf9da23c9cb0980b52871f380c9c48c2d22524f/pyrefly/lib/alt/operators.rs#L470-L474.

No action needed on your end; I've patched this internally.

migeed-z (approved):

Issue #2199, opened 2026-01-22
Literal key `in` narrowing for TypedDict w/o extra items


yangdanny97:
### Describe the Bug

```python
from typing import *

class TD(TypedDict):
    x: int

def test(td: TD):
    if "y" in td:
        td["y"] # unexpected error
```

Python: 3.12

If we explicitly check the key exists we should not give this error. Naively we could narrow to `Any`, but this would apply to the whole dict, and would interact weirdly when there is a union

```python
from typing import *

class TD(TypedDict):
    x: int

class TD2(TypedDict):
    x: int
    y: int

def test(td: TD | TD2):
    if "y" in td:
        td["y"] # should this be Any or int | Any ?
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRAYyio4cOgBUAIgApZzGJnkQxDAJSIRdA3XyJO6BiJGYYYRvAaKGmEwt37DEG0JCkvpxk7dDQ0cAbS8fEABdEAAaEDJKayhSQgZaKAoAYjoABVJEsGS6NCw8fDoxXHRINgBXSlQGCCrCEWyAZRgYOgALBgZiOEQAemGEpJTeNmGYdGHMXDE4YcrqiDqGpqrhujBeOlQAN1RoVGxYCqqa%2Bsbm9DpcYi30OFb0MgYeqoBaQ5hKOB3OgAXjoXgAzIQAIwAJi8IhAAF84qgtBA-gAxaAwCglHAEEjkRFAA

rchen152:
Open typed dicts (ones that don't have `closed=` or `extra_items=`) are approximated as having extra items `ReadOnly[object]` (https://typing.python.org/en/latest/spec/glossary.html#term-open), so IMO it would make sense to give `td["y"]` a type of `object` in the above examples.

grievejia:
Would `Any` be a better choice in practice than `object` (at least in default mode), given that the former is better at controlling cascading false positives?

That said, silently introducing an `Any` to the program without an error/warning feels like undesirable if user wants very strict type checking.

yangdanny97:
I originally thought `Any` would be better because `hasattr`-narrowing introduces a new attribute w/ type `Any`, and this check feels conceptually similar to hasattr

[...]

PR #3082, opened 2026-04-09
fix in check does not narrow TypedDict Union #3079

asukaminato0721:
# Summary

Fixes #3079

The solver now narrows the subject type itself for `"key" in x` / `"key" not in x` when x is a TypedDict union, instead of only recording a key facet, so `Foo | Bar` now becomes Foo in the positive branch and Bar in the negative branch.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -21,6 +21,7 @@ use pyrefly_types::facet::UnresolvedFacetKind;
 use pyrefly_types::simplify::intersect;
 use pyrefly_types::simplify::simplify_tuples;
 use pyrefly_types::type_info::JoinStyle;
+use pyrefly_types::typed_dict::ExtraItems;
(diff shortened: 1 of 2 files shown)


rchen152:
There's some discussion about how to fix this in https://github.com/facebook/pyrefly/issues/2199. In particularly, we can't naively narrow the TypedDict union because a non-closed TypedDict can have extra keys we can't see via subtyping.

rchen152:
Thanks for this PR! I kept your narrow_key_membership pruning logic and tests, but reworked how we mark the key accessible. Caching the key's value as a facet narrow masks errors on mixed unions: e.g.
```
def f(v: TD | list[str]):
    if "x" in v:
        v["x"]
```
was wrongly accepted even though a list can't be indexed by a str. I instead added a `present` bit to facets so we can record whether a key is present without overwriting the value, so the TypedDict resolves the key while the list still errors.

No action needed on your end - I made these changes to the imported copy of the PR.

PR #3084, opened 2026-04-09
Feat/import autocomplete with no imports

### LINKED ISSUE #2999: Reddit - Python: Module Import autocomplete suggestions in nvim
I just moved from coc-pyright to the native lsp with nvim-cmp and moved to pyrefly as my python LSP. One thing that I cannot get working is import suggestions.

For example, if I type

from foo import bar

I'd like after I type import to hit <tab> and have a list of elements I can import from `foo`. I can hit <ctrl>-<tab> and the auto-complete menu pops up, but the elements are just general suggestions and not specific to classes,methods or variables that I can import from that module.

Is there a config needed in nvim-cmp to get this working?

https://www.reddit.com/r/neovim/comments/1saj1r3/python_module_import_autocomplete_suggestions/

NarxPal:
# Summary

earlier exported symbols were appearing only when we had importedName for the module.
for eg.
`from math import sq` 

here , `sq` is the `importedName`

since it has chars for importedName the completion suggestion will show symbols for it, but in case when 

`from math import`  

there is no importedName the completion suggestion won't work since  `from_stmt_import_from_name` fn in `pyrefly/lib/state/lsp.rs` require importedName. 

so I added `importedNameEmpty` to match the condition for `from math import |` and trigger completion when there's no name. after that i made sure that `ThisModule` entries appear first, so i marked them `Local` in completion source. and did `Reexport` for `OtherModule` entries.

up until now i was able to trigger completion and symbols are appearing as expected but pyrefly/test are failing due to difference in ordering in the suggestions, so i would like to discuss should we change the ordering inside tests or is there other way out?

one test error eg. 

```
2 | from foo import func
  |                 ^^^^
  |

thread 'test::lsp::completion::from_import_deprecated' (22054) panicked at pyrefly/lib/test/lsp/completion.rs:769:5:
assertion failed: `(left == right)`

Diff < left / right > :
 # main.py
 2 | from foo import func
                ^
 Completion Results:
 
 2 | from foo import func
                         ^
 Completion Results:
<- (Variable) deprecated
 - (Function) func_ok
>- (Variable) deprecated
```

Fixes #2999 

# Test Plan

  - Manually verified in VS Code using a local Pyrefly build from this branch.
  - Opened a Python file with `from math import |` and triggered completion.
  - Confirmed symbol-like `math` exports are shown (for example `sqrt`, `sin`, `cos`) and unrelated helper suggestions are shown at bottom.
  - Also verified `from math import s|` still returns expected symbol completions.


NarxPal:
so i was facing two tests fail before which have now been resolved, below is the complete detail regarding the error in test and how i resolved them:

each two test failure were seen in - 
pyrefly / test (macOS-latest) 
pyrefly / test (ubuntu-latest)
pyrefly / test (windows-latest)

 ### 1. test failure for `from_import_empty_test` 

<details>
<summary> test failure code: </summary>

```
---- test::lsp::completion::from_import_empty_test stdout ----
DEBUG Running with 3 threads (10 MiB stack size)
DEBUG Running epoch 1 of run 0
DEBUG Committing transaction
ERROR Parse error: Expected one or more symbol names after import [parse-error]
 --> main.py:2:16
  |
2 | from foo import
  |                ^
3 | #              ^
  |

thread 'test::lsp::completion::from_import_empty_test' (22058) panicked at pyrefly/lib/test/lsp/completion.rs:735:5:
assertion failed: `(left == right)`

Diff < left / right > :
 # main.py
 2 | from foo import
                    ^
 Completion Results:
>- (Variable) imperial_guard
>- (Constant) __annotations__
>- (Constant) __builtins__
>- (Constant) __cached__
>- (Constant) __debug__
>- (Constant) __dict__
>- (Constant) __doc__
>- (Constant) __file__
>- (Constant) __loader__
>- (Constant) __name__
>- (Constant) __package__
>- (Constant) __path__
>- (Constant) __spec__
 
```
</details>

At the marked cursor (^) in `from foo import|` completion request was being sent which made completion suggestion appear in `Completion Results` while our test file `pyrefly/lib/test/lsp/completion.rs` expects no `Completion Results` which can be seen in following code: 

```
fn from_import_empty_test() {
    let foo_code = r#"
imperial_guard = "cool"
"#;
    let main_code = r#"
from foo import
#              ^
"#;
    let report = get_batched_lsp_operations_report_allow_error(
        &[("main", main_code), ("foo", foo_code)],
        get_default_test_report(),
    );
    assert_eq!(
        r#"
# main.py
2 | from foo import
                   ^
Completion Results:


# foo.py
"#
```


#### Solution:
in `empty_imported_name_from_covering_nodes` fn 

 ```rust
       if pos <= slot_start || pos > line_end {
            return None;
        }
```

this avoided unnecessary completion request being sent during `from foo import|`


### 2. test failure for `from_import_deprecated`

<details>
<summary>test failure code</summary>

```
---- test::lsp::completion::from_import_deprecated stdout ----
DEBUG Running with 3 threads (10 MiB stack size)
DEBUG Running epoch 1 of run 0
DEBUG Committing transaction
ERROR Could not import `func` from `foo` [missing-module-attribute]
 --> main.py:2:17
  |
2 | from foo import func
  |                 ^^^^
  |

thread 'test::lsp::completion::from_import_deprecated' (22054) panicked at pyrefly/lib/test/lsp/completion.rs:769:5:
assertion failed: `(left == right)`

Diff < left / right > :
 # main.py
 2 | from foo import func
                ^
 Completion Results:
 
 2 | from foo import func
                         ^
 Completion Results:
<- (Variable) deprecated
 - (Function) func_ok
>- (Variable) deprecated
 - (Constant) __annotations__
 - (Constant) __builtins__
 - (Constant) __cached__
 - (Constant) __debug__
 - (Constant) __dict__
 - (Constant) __doc__
 - (Constant) __file__
 - (Constant) __loader__
 - (Constant) __name__
 - (Constant) __package__
 - (Constant) __path__
 - (Constant) __spec__
 - (Function) [DEPRECATED] func_not_ok
 
 
 # foo.py
```

</details>

Here, `Completion Results` required `(Variable) deprecated` to be positioned below `(Function) func_ok` but due to my changes in file - `pyrefly/lib/lsp/wasm/completion.rs` :

```rust 
                    let exports = self.get_exports(&handle);
                    for (name, export) in exports.iter() {
                        let (is_deprecated, kind, source) = match export {
                            ExportLocation::ThisModule(export) => (
                                export.deprecation.is_some(),
                                export
                                    .symbol_kind
                                    .map_or(CompletionItemKind::VARIABLE, |k| {
                                        k.to_lsp_completion_item_kind()
                                    }),
                                CompletionSource::Local,
                            ),
                            ExportLocation::OtherModule(_, _) => {
                                (false, CompletionItemKind::VARIABLE, CompletionSource::Local)  //  <----- Failure: mismatch occured when `CompletionSource::Reexport` was used here previously
                            }
                        };
```

when i used `CompletionSource::Reexport` in `ExportLocation::OtherModule` , those items got a lower rank bucket than Local, so deprecated moved below direct module symbols like `func_ok`. this caused ordering mismatch with the test file. so i changed it to `CompletionSource::Local` to prefer test file ordering. 


 ## Test Plan

  - `CARGO_TARGET_DIR=/work/pyrefly-target cargo test from_import_empty_test`
    - Passed

  - `CARGO_TARGET_DIR=/work/pyrefly-target cargo test from_import_deprecated`
    - Passed

  - `CARGO_TARGET_DIR=/work/pyrefly-target python3 test.py --no-test --no-conformance --no-jsonschema`
    - Passed
    
  ### Notes

  - Used `CARGO_TARGET_DIR=/work/pyrefly-target` to avoid build-path issues from spaces in the workspace path.

 ## Manual Verification (VS Code)

  - Verified in VS Code using a local Pyrefly binary built from this branch.
  - Opened a Python file with `from math import |` and triggered completion.
  - Confirmed math symbol completions are shown (for example: `sqrt`, `sin`, `cos`).
  - Also verified `from math import s|` still returns expected symbol completions.

Screenshot from 2026-04-11 00-01-48

Screenshot from 2026-04-11 02-26-01

NarxPal:
test failures are resolved and from..import autocomplete implementation now works as expected , pr is ready to merge!

cc @stroxler

kinto0:
looks like this approach is working within the AST we already have. I wonder if we can simply make an upstream change to the parser to support this natively? 

I wonder what `ty` does since they use the same parser as us? (maybe our paser just needs to be updated?)

NarxPal:
i reviewed the version we are using for ruff- `474b00568ad78f02ad8e19b8166cbeb6d69f8511` . So, ty basically have a token centric approch to detect the `from..import` statement. for eg. `from math import |` , the "state-machine" as they call it, will look backwards from the cursor position and find tokens such as `import` or `from` based upon which the state machine takes decision in `detect` fn inside - `ruff/crates/ty_ide/src/completion.rs` . 

```rust
    if let Some(import) = ImportStatement::detect(db, file, &parsed, tokens, typed.as_deref()) {
        import.add_completions(db, file, &mut completions);
    } else {
```
further `add_completions` fn handles the match block :

```rust
                FromImportKind::Attribute => {
                    completions.extend(model.from_import_completions(ast)); 
                }
```

while in pyrefly we look at the nodes and make decision in the math block, and since empty name wasn't considered so it returned `None` 

`pyrefly/lib/state/lsp.rs`

```rust
    fn identifier_from_covering_nodes(
        covering_nodes: &[AnyNodeRef],
    ) -> Option<IdentifierWithContext> {
        match (
            covering_nodes.first(),
            covering_nodes.get(1),
            covering_nodes.get(2),
            covering_nodes.get(3),
        ) {

```


i think this approach of position matching `covering_nodes.first` `covering_nodes.get()` should be changed.

pyrefly/lib/lsp/wasm/completion.rs
```diff
@@ -892,29 +892,95 @@ impl Transaction<'_> {
...
-                            ..Default::default()
-                        }))
+                            source,
+                            is_incompatible: false,
+                        })
+                    }
+                }
+            }
+
+            Some(IdentifierWithContext {
```
kinto0: this branch has a bunch of duplication with the previous branch. is there a way to consolidate them?

pyrefly/lib/state/lsp.rs
```diff
@@ -672,14 +680,99 @@ impl<'a> Transaction<'a> {
...
+            Self::empty_imported_name_from_covering_nodes(&covering_nodes, source, position)
+        {
+            return Some(identifier);
+        }
+
+        // If the cursor is in trailing whitespace after `import` for eg. `from math import    |`, probe left on
+        // the same line until we hit a non-whitespace character and reuse that
+        // AST context while validating the original cursor position.
+        let mut probe = position.to_usize();
+        while probe > 0 {
```
kinto0: this is pretty concerning. it seems like it could reach unrelated statements pretty easily. what if we instead handle this case as part of `empty_imported_name_from_covering_nodes` by trimming whitespace?

pyrefly/lib/state/lsp.rs
```diff
@@ -313,6 +314,13 @@ pub(crate) enum IdentifierContext {
         /// ex: For `from ... import x`, the name is `x`. For `from ... import x as y`, the name is `y`.
         name_after_import: Identifier,
     },
+
+    ImportedNameEmpty {
```
kinto0: nit: if we keep this, can we add a docstring?

pyrefly/lib/state/lsp.rs
```diff
@@ -313,6 +314,13 @@ pub(crate) enum IdentifierContext {
         /// ex: For `from ... import x`, the name is `x`. For `from ... import x as y`, the name is `y`.
         name_after_import: Identifier,
```
kinto0: I wonder if we should instead make Identifier here optional? it doesn't seem like we need to duplicate the ImportedName fields

pyrefly/lib/state/lsp.rs
```diff
@@ -672,14 +680,99 @@ impl<'a> Transaction<'a> {
...
+
+        if !import_from.names.is_empty() {
+            return None;
+        }
+
+        let stmt_start = import_from.range().start().to_usize();
+        let stmt_end = import_from.range().end().to_usize();
+        let stmt_text = source.get(stmt_start..stmt_end)?;
+
+        let import_keyword_end = stmt_text.find(" import")? + " import".len();
```
kinto0: what if instead of string matching, we re-lex with `ruff_python_parser::lexer::lex()`? then we would have access to the tokens for the relevant statements and we can perform a token-centric approach similar to what I think you're describing in your comment

kinto0 (changes requested):
thanks for the change! and I'm really sorry about the delayed review. It's been very busy recently. feel free to keep tagging me if I am not responsive enough. 

the general approach seems good. left a few comments but I can tell you've put a lot of thought into this, thank you. 

a few other general things:
- there's a test in test/lsp/completion.rs that should cover this case, but it looks like it isn't changed: from_import_empty_test
- could you add more tests relating to the lexing / matching logic you are changing?
- unrelated cargo.lock - I don't think that should change (that's the only merge conflict)

NarxPal:
first of all, thank you for your kind words @kinto0. also sorry from my side for delayed changes , past few weeks were busy, which is part of the reason i wanted to do it earlier asap

all suggested changes have been considered and will reply to each comment stating what changes i made once i add few test for the empty slot case.

pyrefly/lib/lsp/wasm/completion.rs
NarxPal: so since i unified the use of importedName and importedNameEmpty variants into one variant called `importedName` it has consolidated our approach here and have added the code inside a fn called `add_imported_name_completions`

pyrefly/lib/state/lsp.rs
NarxPal: you are right! since we have unified both importedName and importedNameEmpty , i have now made it optional since importedNameEmpty doesn't require name_after_import

pyrefly/lib/state/lsp.rs
NarxPal: not sure why i didn't came over this fn before, and yes you are right that this fn could help us in lexing but since `current_range` fn which helps in providing the byte_range of the token, which in our case we require byte_range for token - `import` , can't be used since it's not public and limited to it's crate.

so i started finding different way to use lexer and i came over this `parse_unchecked_source`  which basically parses the source file and find `import` token from `from..import` statement

NarxPal:
> a few other general things:
> 
> * there's a test in test/lsp/completion.rs that should cover this case, but it looks like it isn't changed: from_import_empty_test
> * could you add more tests relating to the lexing / matching logic you are changing?
> * unrelated cargo.lock - I don't think that should change (that's the only merge conflict)

from_import_empty_test covers the case when cursor is placed right after import for eg. `from math import|`  which is different case from our empty slot case so there's no completion suggestion listed.

for our empty slot case i have added two tests each covering both single trailing whitespace and multiple trailing whitespace.

cargo.lock file was automatically changed not sure when maybe during `cargo build` or it happened due to my `rust analyzer` extension . i did reverted back to main branch's `cargo.lock` file which than led to few error in my editor (but no error are being captured when i run cli commands like `cargo check` or `cargo fmt --check)` so please ensure you don't see any error in file `pyrefly/lib/lsp/wasm/completion.rs` 

changes are ready to be reviewed, @kinto0

stroxler (approved):

PR #3087, opened 2026-04-09
Explicitly declare VS Code extension workspace trust capabilities

samwgoldman:
The extension already does not run in untrusted workspaces and lspPath can already be overridden in trusted workspaces. These changes make that behavior explicit by declaring untrustedWorkspaces support as disabled via the capabilities API and adding machine-overridable scope to the lspPath and lspArguments settings.

# Test Plan

Confirmed that we get a dialog when opening an untrusted workspace:
Screenshot 2026-04-09 at 11 04 00 AM

In an untrusted workspace, the settings.json is not used (Pyrefly is not enabled at all), no PWNED file:
Screenshot 2026-04-09 at 11 22 49 AM

The Extension summary includes the description of why restricted mode is not supported:
Screenshot 2026-04-09 at 11 34 14 AM

In trusted mode, a malicious actor can still cause point lspPath to a malicious binary. This is the same as rust-analyzer and TypeScript extensions.
Screenshot 2026-04-09 at 11 43 05 AM

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -36,6 +36,12 @@
         "onLanguage:python",
         "onNotebook:jupyter-notebook"
     ],
+    "capabilities": {
+        "untrustedWorkspaces": {
+            "supported": false,
+            "description": "Pyrefly can be configured to execute binaries. A malicious actor could exploit this to run arbitrary code on your machine."
+        }
+    },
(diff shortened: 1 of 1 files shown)


grievejia (approved):

PR #3093, opened 2026-04-10
fix Respect VERSIONS file in typeshed #2535

### LINKED ISSUE #2535: Respect `VERSIONS` file in typeshed
### Describe the Bug

Typeshed contains a [`VERSIONS` file](https://github.com/python/typeshed/blob/main/stdlib/VERSIONS) indicating stdlib modules' version availability. For example, the `distutils` package is only available from Python 3.0 to 3.11, so importing it in Python 3.12+ is considered an error that gets reported by all of mypy, pyright, and ty:
```python
import distutils  # Should be an error in python 3.12+, but pyrefly is silent
```
Pyrefly probably needs to include this `VERSIONS` file into its bundled typeshed, and adjust what `BundledTypeshed` contains dynamically based on the python version. 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIEAtsbgE4AuABJhHAwK4PQIA0IXHiXKIQAYiYBVblAgNSTMB3QBjbrnRwAOuh2YYYRfSqoGAfXQcq2GHQAU%2BREwjoGASiYBaAHxN2dRB0mYKY6GE46dEUtEAA5KxsApmB8AF8YnRB%2BMjCwKFJCBlwqKAoJAAVSXPy-DBwCJhUNSABzDjpTCA1CHQkAZRgYJgALBgZiOEQAeimcg3zCehapmHQpzFwVOCmm9Fb2zo0pozomVAA3VGhUbFhG5og2jvUo3GIXuB70MgZhjU9zrY4F0ogBeJgxADMhAAjAAmDLoECpfioNQQQEAMWgMAoaCweCIZGRQA

_No response_

asukaminato0721:
# Summary

Fixes #2535

making bundled stdlib resolution respect typeshed’s stdlib/VERSIONS metadata.

bundled stdlib modules are now gated by the active python_version, so import distutils is rejected on Python 3.12+, and removed stdlib names are also filtered out of prefix results instead of leaking back in through third-party stubs.

# Test Plan

add test


pyrefly/lib/module/typeshed.rs
```diff
@@ -12,32 +12,64 @@ use std::path::PathBuf;
...
+    fn parse(range: &str) -> anyhow::Result<Self> {
+        let (min, max) = range
+            .split_once('-')
+            .with_context(|| format!("Invalid typeshed version range `{range}`"))?;
+        Ok(Self {
+            min: min.parse()?,
+            max: if max.is_empty() {
+                None
+            } else {
+                Some(max.parse()?)
```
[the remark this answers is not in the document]
connernilsen: This is a good point. @asukaminato0721, could you add a way to patch-insensitive compare versions?

crates/pyrefly_bundled/src/lib.rs
```diff
@@ -115,17 +118,31 @@ fn extract_pyi_files_from_archive(filter: PathFilter) -> anyhow::Result<SmallMap
...
+    extract_files_from_archive(PathFilter::Stdlib, |path| {
+        path.extension().is_some_and(|ext| ext == "pyi")
+    })
+}
+
+pub fn bundled_typeshed_versions() -> anyhow::Result<String> {
+    let mut versions =
+        extract_files_from_archive(PathFilter::Stdlib, |path| path == Path::new("VERSIONS"))?;
+    versions
+        .shift_remove(Path::new("VERSIONS"))
```
connernilsen: very minor nit: maybe `into_iter().find()` instead of shift removing?

pyrefly/lib/module/finder.rs
```diff
@@ -742,8 +743,19 @@ pub fn find_import_internal(
...
+    let unavailable_stdlib_module = matches!(style_filter, Some(ModuleStyle::Interface) | None)
+        && typeshed().is_ok_and(|ts| {
+            ts.has_module(module)
+                && !ts.is_available_for_python_version(module, config.python_version())
+        });
+    let typeshed_third_party_result = if unavailable_stdlib_module {
+        None
+    } else {
+        find_third_party_stub(module, style_filter)
+    };
```
connernilsen, line 762: I'm not sure if I really understand why we're limiting which third party stubs we're returning here. Could you explain and addin a comment above here why we're doing this?

pyrefly/lib/module/finder.rs
```diff
@@ -794,7 +806,14 @@ pub fn find_import_internal(
                     err, module,
                 )))
             },
-            |ts| ts.find(module).map(FindingOrError::new_finding),
+            |ts| {
+                (if bundled_typeshed_origin {
```
connernilsen, line 817: Might also be nice to add a comment above this explaining that if we're importing within typeshed, we don't want to return an error on a module not existing

pyrefly/lib/module/finder.rs
```diff
@@ -2417,6 +2444,28 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_find_import_prefixes_respects_stdlib_versions() {
```
connernilsen, line 2452: Can you also add a test similar to this for non-prefix imports?

connernilsen (changes requested):
Hey @asukaminato0721, thanks for adding this functionality! Just a few requests and then I think we should be good to import.

crates/pyrefly_bundled/src/lib.rs
asukaminato0721: fixed

pyrefly/lib/module/finder.rs
asukaminato0721: added

pyrefly/lib/module/finder.rs
asukaminato0721: added

pyrefly/lib/module/typeshed.rs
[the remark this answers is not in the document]
asukaminato0721: added

pyrefly/lib/module/finder.rs
asukaminato0721: added above

connernilsen:
The changes look good, but seems like there's a failing test now. Can you fix? I'll pull in afterward

connernilsen (changes requested):
Requesting changes for the broken test

asukaminato0721:
```
cargo test test::simple::test_tstring
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running unittests lib/lib.rs (target/debug/deps/pyrefly-fc4c0f0a68e04b9c)

running 1 test
test test::simple::test_tstring ... ok
```

connernilsen:
Hmm, maybe it's some error on the base revision of your PR. I'll try pulling this in to see if I still see it.

pyrefly/lib/module/finder.rs
```diff
@@ -792,8 +793,23 @@ pub fn find_import_internal(
...
+        origin.is_some_and(|path| matches!(path.details(), ModulePathDetails::BundledTypeshed(_)));
     let origin = origin.map(|p| p.as_path());
-    let typeshed_third_party_result = find_third_party_stub(module, style_filter);
+    // If a stdlib module exists in bundled typeshed but is unavailable for the active
+    // Python version, do not fall through to third-party bundled stubs with the same
+    // top-level name. That would turn a removed stdlib import like `distutils` on 3.12
+    // into a misleading third-party stub result from `setuptools`.
+    let unavailable_stdlib_module = matches!(style_filter, Some(ModuleStyle::Interface) | None)
+        && typeshed().is_ok_and(|ts| {
+            ts.has_module(module)
```
connernilsen, line 805: Actually, this line might be redundant. Doesn't `is_available_for_python_version`, basically do this already?

crates/pyrefly_bundled/src/lib.rs
```diff
@@ -129,18 +130,36 @@ fn extract_pyi_files_from_archive(
...
-    extract_pyi_files_from_archive(PathFilter::Stdlib).map(|(files, _)| files)
+    extract_files_from_archive(PathFilter::Stdlib, |path| {
+        path.extension().is_some_and(|ext| ext == "pyi")
+    })
+    .map(|(files, _)| files)
+}
+
+pub fn bundled_typeshed_versions() -> anyhow::Result<String> {
+    let (versions, _) =
+        extract_files_from_archive(PathFilter::Stdlib, |path| path == Path::new("VERSIONS"))?;
```
connernilsen, line 141: Actually this will decompress the archive twice on startup. Can we instead return something like `anyhow::Result<(SmallMap<PathBuf, String>, String)>` from `bundled_typeshed` so that we don't have to extract and iterate over the archive twice here?

connernilsen (changes requested):
Sorry, two more things I caught in a last review

yangdanny97:
I'm doing a typeshed update right now, so I'll try to pull this in and get it over the finish line

pyrefly/lib/module/finder.rs
```diff
@@ -440,7 +441,17 @@ pub fn find_import_internal(
     timing: Option<&TransactionTimingCounters>,
 ) -> FindingOrError<ModulePath> {
     let mut namespaces_found = vec![];
+    let bundled_typeshed_origin =
+        origin.is_some_and(|path| matches!(path.details(), ModulePathDetails::BundledTypeshed(_)));
     let origin = origin.map(|p| p.as_path());
+    // If a stdlib module exists in bundled typeshed but is unavailable for the active
+    // Python version, do not fall through to third-party bundled stubs with the same
+    // top-level name. That would turn a removed stdlib import like `distutils` on 3.12
+    // into a misleading third-party stub result from `setuptools`.
```
connernilsen: Actually, I'm not fully sure I fully understand this use case. Thanks for adding the comment, but can you explain why we wouldn't want to see the third-party stub result here?

I can understand for a module like `distutils`, which is specifically named, but there are other modules like `chunk`, which might be a valid third (or first) party import.

Can you let me know a little more about what you were intending?

connernilsen (changes requested):
Sorry for the delay on this @asukaminato0721. I'm looking at reviewing OSS PRs again, so I'll be more attentive here. Looks like there are some test failures from the most recent changes. mypy_primer also looks like it's failing from a panic in these changes (there's a missing version I believe), which I'll use to double check that things are working as expected.

Re-request review once that's fixed and I'll take another look

connernilsen:
Putting up a fix for the codspeed failures

stroxler (approved):

connernilsen (approved):
Hey Asuka, sorry for all of the churn with this PR, it's landing right now, so should hopefully be merged in soon. Thanks for all your work here!

grievejia (approved):

PR #3101, opened 2026-04-10
Fix(#2918): `NotImplemented` being incorrectly accepted as a callable

### LINKED ISSUE #2918: Calling NotImplemented (a constant, not a class)
`NotImplemented` is a singleton constant, not callable. `NotImplemented()` is always a mistake (they mean `NotImplementedError()`).

```python
def broken():
    raise NotImplemented()

def also_broken():
    raise NotImplemented("not yet done")
```

**Sandbox:** [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXAAlHAhgOwCYCMD2AHgHQAOAnksiNjGMrgE74DWMmAFAJSIA6myQckboIcGMgBy%2BAC4BJALalYC9jJjZu-frXroocfAH0mrdtz4ChIsROnylKtRo68QmWcnIwZybPkwYNy5%2BEABfABpqdABjGQgANxgAMWgYKhAMHAISCnCgA)

iamPulakesh:
Fixes [#2918 ](https://github.com/facebook/pyrefly/issues/2918)

# Summary 
Resolves the issue by showing a helpful NotCallable error when users mistakenly try to call the NotImplemented singleton constant as a function.

# Changes
- pyrefly/lib/alt/call.rs: Added the has_qname match checks for builtins._NotImplementedType and types.NotImplementedType inside expr_call_infer to return an early NotCallable error.
- pyrefly/lib/test/calls.rs: Removed the bug marker from test_call_not_implemented_constant and updated the standard test framework expectations to assert our new error cleanly.

# Tests
`cargo test test_call_not_implemented_constant -p pyrefly --lib` - Passed
`python test.py --no-test --no-conformance --no-jsonschema` - Passed

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1748,7 +1748,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 kws = x.arguments.keywords.map(CallKeyword::new);
             }
 
-            let result = self.distribute_over_union(&callee_ty, |ty| match ty.callee_kind() {
+            let result = self.distribute_over_union(&callee_ty, |ty| {
+                // NotImplemented is a singleton constant, not a callable class.
+                if matches!(ty, Type::ClassType(cls) if cls.is_builtin("_NotImplementedType") || cls.has_qname("types", "NotImplementedType"))
+                {
+                    return self.error(
+                        errors,
+                        x.func.range(),
+                        ErrorInfo::Kind(ErrorKind::NotCallable),
+                        "`NotImplemented` is not callable. Did you mean `NotImplementedError`?".to_owned(),
+                    );
+                }
+                match ty.callee_kind() {
                 Some(CalleeKind::Function(FunctionKind::AssertType)) => self
                     .call_assert_type(
                         &x.arguments.args,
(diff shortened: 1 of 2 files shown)


connernilsen (approved):
Hey @iamPulakesh, thanks for doing this. Looks great to me!

iamPulakesh:
> Hey @iamPulakesh, thanks for doing this. Looks great to me!

Hey @connernilsen, thank you so much for the review and Could you please approve those 3 workflows awaiting maintainer approval?

pyrefly/lib/alt/call.rs
```diff
@@ -1603,6 +1603,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
         }
 
+        // NotImplemented is a singleton constant, not a callable class.
+        if matches!(&callee_ty, Type::ClassType(cls) if cls.has_qname("builtins", "_NotImplementedType") || cls.has_qname("types", "NotImplementedType"))
```
fangyi-zhou: Could we move this check down within the `distribute_over_union` call? So that we can make sure we can catching the error when they occur within a union.

```python
def f(): ...
if ...:
    x = f
else:
    x = NotImplemented
x() # should error here
```

fangyi-zhou: nit: `cls.has_qname("builtins", "_NotImplementedType")` -> `cls.is_builtin(...)`

fangyi-zhou (changes requested):

pyrefly/lib/alt/call.rs
iamPulakesh: Hey, @fangyi-zhou I've moved the check inside distribute_over_union to properly handle union cases, added the test you suggested and applied the is_builtin nit. could you please take another look? Thanks for the review.

fangyi-zhou (approved):
Thanks for the fix!

iamPulakesh:
> Thanks for the fix!


Happy to Contribute, thanks for the review.

PR #3107, opened 2026-04-10
Preserve decorated function type through @wraps-style unions

migeed-z:
Summary:
When a decorator returns a union where one branch is an unannotated
*args/**kwargs wrapper, return the original function's type. 

The heuristic: if a decorator's inferred return type is a union, and any
member of that union is a function with only *args/**kwargs parameters
(plus optional self) and no return type annotation, it's recognized as a
functools.wraps-style passthrough wrapper. In that case the decorated
function keeps its original signature instead of getting the union type.

The fix adds `Callable::is_args_kwargs_wrapper()` to detect wrapper
functions, and uses it in `decorate_returned_callable` to handle unions
containing such wrappers.

for issue https://github.com/facebook/pyrefly/issues/2621

Reviewed By: stroxler

Differential Revision: D100251940

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -39,6 +39,7 @@ use crate::equality::TypeEq;
 use crate::equality::TypeEqCtx;
 use crate::keywords::DataclassTransformMetadata;
 use crate::type_output::TypeOutput;
+use crate::types::AnyStyle;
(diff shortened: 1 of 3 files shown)


stroxler (approved):

grievejia (changes requested):

grievejia (approved):

PR #3111, opened 2026-04-12
Add faling test for overloads for returning Self

fangyi-zhou:
Summary:
If a method returns `Self`, then an overload returning the current class should not be accepted as an overload.

This is on par with Pyright: https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDKApgDbAA0UYAbkSCWAIYAmAUKwMYmMDOPUAYQBcrKGKgABGnQYtR45kWBQuYFEQAUPUhSgAPIZhQwAlFAC0APkGGAdPfliptekzbioi5avVadlAygeGBAzKxsoe1tHTyUVBl9tMgCw62IyO3sgA

Differential Revision: D100428648

### DIFF
--- a/pyrefly/lib/test/typing_self.rs
+++ b/pyrefly/lib/test/typing_self.rs
@@ -564,3 +564,18 @@ class Base:
         self.field = "hello"
 "#,
 );
+
+testcase!(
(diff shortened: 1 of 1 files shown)


rchen152 (approved):

PR #3112, opened 2026-04-12
Skip self/cls in overload consistency checks (#3112)

fangyi-zhou:
Summary:

Skip the self/cls parameter when checking overload input signature consistency.
Without this, overloads with narrowed self types (e.g. `self: C[int]`) produce
false positive `inconsistent-overload` errors because `C[int] <: Self@C` fails
after removing the unsound `ClassType <: SelfType` subtyping rule.

This is a deliberate relaxation: stripping self loses type variable constraints
that flow through the self parameter, but in practice the self type is validated
separately (must be a superclass of the defining class, D97390197), and callers don't pass
self explicitly.

This is preparatory work for removing the unsound `ClassType <: SelfType`
subtyping rule.

Reviewed By: grievejia

Differential Revision: D99748452

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -8918,17 +8918,6 @@
     {
       "code": -2,
       "column": 9,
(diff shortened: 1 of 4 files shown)


grievejia (approved):

PR #3115, opened 2026-04-12
fix impl unwrap block #2207

asukaminato0721:
# Summary

part of #2207

impl unwrap block

https://github.com/astral-sh/ruff/issues/24552

# Test Plan

add test


pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -4101,6 +4129,58 @@ class Child(Base):
...
+    for item in items:
+        print(item)
+    else:
+        print("done")
+"#;
+    let selection = find_nth_range(code, ":", 3);
+    assert_no_unwrap_block_action(code, selection);
+}
+
 #[test]
```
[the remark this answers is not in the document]
connernilsen: This seems like a reasonable recommendation

pyrefly/lib/state/lsp/quick_fixes/unwrap_block.rs
```diff
@@ -0,0 +1,295 @@
...
+            if let Some(target) = find_unwrap_block_target_in_body(&try_stmt.body, source, position)
+            {
+                return Some(target);
+            }
+            for handler in &try_stmt.handlers {
+                if let Some(handler) = handler.as_except_handler()
+                    && let Some(target) =
+                        find_unwrap_block_target_in_body(&handler.body, source, position)
+                {
+                    return Some(target);
```
connernilsen, line 151: This would just take the first exception handler. Is there a case where we'd want to take all/none exception handlers?

Can you add some tests for the `try` case?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -4101,6 +4129,58 @@ class Child(Base):
     assert!(apply_first_safe_delete_action(code).is_none());
 }
 
+#[test]
```
connernilsen, line 4132: It looks like we're testing a lot of stuff that works here, but can we test a few examples where we try to break the logic/cause things to fail and see that `None` gets returned?

pyrefly/lib/state/lsp/quick_fixes/unwrap_block.rs
```diff
@@ -0,0 +1,295 @@
...
+        if !line.trim().is_empty() {
+            let indent_len = line
+                .chars()
+                .take_while(|ch| *ch == ' ' || *ch == '\t')
+                .map(char::len_utf8)
+                .sum::<usize>();
+            if line.get(..indent_len) == Some(outer_indent) && line.rfind(':').is_some() {
+                let colon = line.rfind(':').expect("checked above");
+                return TextSize::try_from(line_start + colon).ok();
+            }
```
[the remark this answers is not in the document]
connernilsen: This makes sense. One of the recommendations below adds a test for this

connernilsen (changes requested):
Hey @asukaminato0721, this is looking good, but a few things (mostly requests for more tests). Let me know if you have questions

pyrefly/lib/state/lsp/quick_fixes/unwrap_block.rs
```diff
@@ -0,0 +1,384 @@
...
+    next_clause_start: Option<TextSize>,
+    source: &str,
+    position: TextSize,
+) -> Option<UnwrapBlockTarget> {
+    let first_stmt = body.first()?;
+    let (to_indent, _) = line_indent_and_start(source, outer_range.start())?;
+    let (from_indent, first_stmt_line_start) =
+        line_indent_and_start(source, first_stmt.range().start())?;
+    let colon_position = find_suite_colon(source, header_start, first_stmt_line_start, &to_indent)?;
+    if position != colon_position && position != colon_position + TextSize::new(1) {
```
connernilsen, line 301: This might be hard to discover. It would mean we would only show the unwrap target if the user is on the colon itself, right? Should we instead look at the whole line?

pyrefly/lib/state/lsp/quick_fixes/unwrap_block.rs
```diff
@@ -0,0 +1,384 @@
...
+                '\\' => escaped = true,
+                current if current == quote_char => quote = None,
+                _ => {}
+            }
+            continue;
+        }
+        match ch {
+            '\'' | '"' => quote = Some(ch),
+            '#' => break,
+            ':' => last_colon = Some(index),
```
connernilsen, line 374: This part won't handle string blocks (`"""`/`'''`) or code in f-strings. Could we reference the AST's text ranges for this section along with the text in the line to handle that? Also please add tests for this

pyrefly/lib/state/lsp/quick_fixes/unwrap_block.rs
```diff
@@ -0,0 +1,384 @@
...
+            if let Some(target) =
+                find_unwrap_block_target_in_body(&try_stmt.orelse, source, position)
+            {
+                return Some(target);
+            }
+            if let Some(target) =
+                find_unwrap_block_target_in_body(&try_stmt.finalbody, source, position)
+            {
+                return Some(target);
+            }
```
connernilsen, line 163: In the `For` and `While` cases above, we return `None` if these are present. Is there a reason we're not doing that here?

pyrefly/lib/test/lsp/code_actions.rs
```diff
@@ -4101,6 +4129,191 @@ class Child(Base):
...
```
connernilsen, line 4132: Two more test requests I thought of on the re-review:
1. Can we add a test for unwrapping `elif`
2. Can we attempt unwrapping inner and outer nested statements?

connernilsen (changes requested):
A few more things I found

connernilsen (approved):
Thanks for the fixes! I'll try to pull it in and get a review

kinto0 (changes requested):

connernilsen (changes requested):
Here's Kyle's feedback:

1. Dedent corrupts multi-line string literals (behavior-changing refactor).

`unwrap_block_code_actions` dedents the body via reindent_block (extract_shared.rs:432-458), which rewrites the indentation of every non-blank line textually, with zero string/token awareness. Any body line that sits inside a triple-quoted string and starts with from_indent gets its leading whitespace stripped — silently changing the string's runtime value.

Failure scenario:
```
if cond:
    x = """
    hello
    """
```

Unwrapping the if produces `x = "\nhello\n"` instead of `x = "\n    hello\n    "`. Same corruption for any unwrapped block containing a function/class with a docstring, or implicit string concatenation across lines. This is precisely the case MichaReiser flagged on the ty PR ("Ruff has the adjust_indentation helper… dedenting is very complicated"); ty avoids it entirely by only editing INDENT/DEDENT token span, which never touch string interiors. 

Note this helper is shared: `reindent_block`/`reindent_statement` have the same latent bug in extract-superclass, move-members, and move-module today — but this diff adds a new user-facing action that routinely hits it (unwrapping around docstring'd defs is common). Recommend a token-aware dedent (see below), which would also de-risk the existing callers.

2. Colon-finding re-implements a lexer instead of using the token stream. 

`find_suite_colon_in_header` + `QuoteState` + `triple_quote_delimiter` (~90 lines) manually re-tokenize the header to skip colons inside strings/comments/f-strings. Pyrefly has the token stream available — `Transaction::get_parsed_module(handle)?.tokens()` (populated at `Require::Everything`, which is exactly the open-file + test context where code actions run; confirmed steps.rs:79-100, used already in `semantic_tokens.rs`). Switching to the token stream deletes the hand-rolled scanner and its edge cases (PEP 701 nested same-quote f-strings, raw triple-quoted strings) that the textual version gets wrong. This directly serves the repo's KISS / "extract semantic info early, don't re-parse text" guidance, and matches the sibling impl.

stroxler (changes requested):

PR #3125, opened 2026-04-14
strip "-stubs" suffix from FQN in `pyrefly report` #3094

### LINKED ISSUE #3094: `pyrefly report`: FQNs for `-stubs` packages should not include "-stubs"
For example for `scipy-stubs` every FQN starts with `scipy-stubs.*`, instead of `scipy.*`:

```json
{
  "kind": "function",
  "name": "scipy-stubs.stats._crosstab.CrosstabResult.__new__",
```

Pyrefly 0.60.1

jorenham:
# Summary

This fixes the fully-qualified names reported with `pyrefly report` for stubs-only packages according to [PEP 561](https://peps.python.org/pep-0561/) by removing the top-level `-stubs` suffix.

Fixes #3094

# Test Plan

added a unit test

### DIFF
--- a/crates/pyrefly_python/src/module_name.rs
+++ b/crates/pyrefly_python/src/module_name.rs
@@ -493,6 +493,12 @@ impl ModuleName {
                     out.push(x.to_string_lossy());
                 }
             }
+            // strip `-stubs` from the top-level directory (e.g. `scipy-stubs/` -> `scipy`)
+            if let Some(first) = out.first_mut()
+                && let Some(stripped) = first.strip_suffix("-stubs")
+            {
+                *first = stripped.to_owned().into();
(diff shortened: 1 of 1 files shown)


rchen152:
Thanks for all the pyrefly report fixes, @jorenham!

jorenham:
I see there's a warning, but not what it is exactly :thinking:

stroxler (approved):

rchen152:
> I see there's a warning, but not what it is exactly 🤔

Oh, you can ignore the signal from "Facebook Internal". I sort of wish we didn't export that, but I don't think we can hide it XD (In this case, the warning is just for a pre-existing skipped test.)

Issue #3102, opened 2026-04-10
`pyrefly report`: should external symbols count towards type coverage?


jorenham:
Typestats basically ignores symbols defined outside of the projects (except for a select few from `typing[_extensions]` and `_typeshed`). The reasoning here is that if those symbols are untyped, then there's nothing the user can do about it. That's why typestats doesn't include those in the reported counts and coverage. This avoids situations similar to how `pyright --verifytypes`,  will never be 100% if you have an `import numbers` somewhere in your code. Pyright has an `--excludeexternal` flag for this to ignore such cases, but as far as I've seen, everyone always enables this flag.

I guess there's also something to be said for reporting transitively untyped symbols, but I can see how it's easier to reason about type-coverage if it's limited to the target project. 

I suppose that a `--excludeexternal` flag like pyright has would be most flexible, but it's probably easier to always ignore external dependencies, like typestats currently does.

jorenham:
Actually, it looks like the issue is that `pyrefly report` treats external symbols as `Any`, rather than treating them as implicitly typed like typestats does. So #3127 will also fix this, and this can be considered as a duplicate of #3096.

PR #3127, opened 2026-04-14
consider partially unknown types as typed in `pyrefly report` #3096

### LINKED ISSUE #3096: `pyrefly report`: `"n_any"` is too high
In typestats, the thought process behind the `"n_any"` counts, is b `x: Any` is equivalent to an untyped `x`. So in that sense it's (strictly speaking) not "typed". So typestats only considers something "any" when it's "100% any". 

However, `pyrefly report` seems to also count e.g. `x: np.floating`, which is equivalent to `np.floating[Any]`, as "any". The consequence is that in `scipy-stubs` the strict coverage is below 50%, even though the non-strict coverage is 100%. And according to typestats, the strict coverege of scipoy-stubs is 99.7%: https://jorenham.github.io/typestats/dashboard/

So I propose that `pyrefly report` only counts an annotations as any if it's "100% any", like typstats does.

jorenham:
# Summary

This ensures that annotations like `_: numpy.floating[Any]` are now treated as "typed" rather than "any".

Fixes #3096 (and #3102)

# Test Plan

Added a unit test and integration tests

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -505,9 +505,7 @@ impl ReportArgs {
                     let is_type_known = annotation_text.is_some()
                         && answers
                             .get_idx(*annot_idx)
-                            .and_then(|awt| {
-                                awt.annotation.ty.as_ref().map(Self::is_type_fully_known)
-                            })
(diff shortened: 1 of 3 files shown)


stroxler (approved):

PR #3129, opened 2026-04-14
fix property typable counts in `pyrefly report` #3099

### LINKED ISSUE #3099: `pyrefly report`: property setters/getters/deleters are counted separately
If a property has a getter and a setter, it's still one property. But `pyrefly report` (0.60.1) would currently count that as 2 properties.

Also, the typables are counted as follows in typestats:

- a getter has 1 typable (the output)
- a setter has 1 typable (the input), because the `None` return type is trivial
- a deleter has no typables; the `None` return is trivial (so deleters can effectively be ignored)

So I propose that `pyrefly report` also implements it this way.

jorenham:
# Summary

This special-cases the typable counting logic of property setters and deleters to ignore their trivial `-> None` return types.

Fixes #3099

# Test Plan

Relevant tests updated

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -22,6 +22,7 @@ use pyrefly_python::module::Module;
 use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::nesting_context::NestingContext;
 use pyrefly_python::short_identifier::ShortIdentifier;
+use pyrefly_types::callable::PropertyRole;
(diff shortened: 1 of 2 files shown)


connernilsen:
~~Hey @rchen152, I noticed you reviewed a similar PR. Would you be able to look at this one too?~~ Never mind, this was the diff I was thinking about, somehow missed that it was imported lol

stroxler (approved):

jorenham:
Thanks for the review :)

PR #3131, opened 2026-04-14
always ignore trivial dunder return types in `pyrefly report` #3100

### LINKED ISSUE #3100: `pyrefly report`: ignore trivial return types of special methods and attributes
For example, the return type of `__init__` is always `None`, so annotated or unannotated, it makes no difference for users. And although some type-checkers require it, I've also seen a couple of libraries that never annotate the return type of `__init__`.

See https://github.com/jorenham/typestats/issues/353 for the full list that typestats uses.

For example, with

```py
def __init__(self, /, name: str, values: Sequence[str]): ...
```

`pyrefly report` shows

```json
{
  "kind": "function",
  "name": "scipy-stubs.io.arff._arffread.NominalAttribute.__init__",
  "n_typable": 2,
  "n_typed": 2,
  "n_any": 0,
  "n_untyped": 0,
  "location": {
    "line": 53,
    "column": 5
  }
},
```

but when I add `-> None`, it shows

```json
{
  "kind": "function",
  "name": "scipy-stubs.io.arff._arffread.NominalAttribute.__init__",
  "n_typable": 3,
  "n_typed": 3,
  "n_any": 0,
  "n_untyped": 0,
  "location": {
    "line": 53,
    "column": 5
  }
}
```

So the `-> None` is only ignored if not provided. Instead, it should always be ignored

jorenham:
# Summary

This changes `pyrefly report` to always ignores trivial types of special methods and attributes, instead of selectively ignoring it when not provided. This avoids inflating the type coverage %.

Fixes #3100

# Test Plan

Corresponding tests were updated


stroxler (approved):

PR #3134, opened 2026-04-14
"merge" overloads in `pyrefly report` #3098

### LINKED ISSUE #3098: `pyrefly report`: typable counts for overloads is too high
In case of

```py
@overload
def f(val: str, /) -> str: ...
@overload
def f(val: bytes, enc: str) -> str: ...
```

there should be 3 typables:

- there is 1 `val` parameter
- there is 1 `enc` parameter
- there is one return type

But Pyrefly currently counts each overload separately and sums the result.
So in this example it'll count `val` twice, `enc` once, and the return twice, for a total of 5 typables.

This significantly inflates the number of typables. 
In scipy-stubs, for example, Pyrefly reports 50,251 typables, whereas typestats reports 13,428 typables.

So typestats deduplicates the parameters (and trivially, the return types) of overloaded signatures. See https://github.com/jorenham/typestats/blob/5fb1144/src/typestats/analyze.py#L335-L376 for the implementation. It's a bit complex, because for positional-only arguments names are irrelevant, but for pos-or-kw params it both *does* and *doesn't* matter, depending on context. To be honest, I'm not 100% certain that this implementation is correct in the most complicated of cases (e.g. in case some overloads have `*args`/`**kwargs`, some have pos-or-kw params, and some others have pos-only + kw-only params), but those are probably rare in practice.

jorenham:
# Summary

This ports the typestats overload merging logic (https://github.com/jorenham/typestats/blob/5fb1144/src/typestats/analyze.py#L335-L376) to dedupe the parameters and callable signatures that make up overloads.

Fixes #3098

# Test Plan

Added tests and updated existing ones

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -140,6 +140,53 @@ impl SlotCounts {
     }
 }
 
+/// Annotation quality for a single slot, ordered worst-to-best so that
+/// `min()` gives "worst-wins" semantics.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+enum SlotRank {
+    Untyped,
+    Any,
+    Typed,
+    Skip,
+}
+
+impl SlotRank {
+    /// Map (has_annotation, is_type_known) to a rank.
+    fn classify(has_annotation: bool, is_type_known: bool) -> Self {
+        match (has_annotation, is_type_known) {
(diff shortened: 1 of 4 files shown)


connernilsen:
Hey @rchen152, would you be able to take a look at this (or redirect it to someone who might be a better reviewer)?

stroxler (approved):

jorenham:
Thanks for reviewing :)

PR #3135, opened 2026-04-14
Preserve function signatures through unannotated decorators (#3135)

migeed-z:
Summary:

When a decorator is fully unannotated, Pyrefly infers Unknown for all return paths, producing a union like ((dispatch_target: Unknown) -> Unknown) | Unknown. This union carries no useful type information but replaces the original function's signature, causing cascading false positives.

In this diff, we extend the heuristic as follows: if every member of the decorator's return union is "fully unknown" (either Unknown itself, or a callable where all parameters and the return type are Unknown), preserve the original function signature.

Reduces TensorFlow errors from 21,007 to 16,213 (~23% reduction, ~4,794 fewer false positives).

Differential Revision:
D100840454

https://github.com/facebook/pyrefly/issues/2621

D100840454

Reviewed By: grievejia

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -74,6 +74,22 @@ impl Callable {
             _ => false,
         }
     }
+
+    /// Returns true if this callable carries no real type information: all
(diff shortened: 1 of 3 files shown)


grievejia (approved):

PR #3140, opened 2026-04-15
Add unnecessary-type-conversion lint for str/int/float (fixes #3109)

### LINKED ISSUE #3109: lint unnecessary type convert
### Describe the Bug

```py
def f(x: str):
   y = str(x) # hint here
```

### Sandbox Link

_No response_

_No response_

ahaanbohra:
# Summary

Adds a new unnecessary-type-conversion lint that warns when str(), int(), or float() is called on an argument that is already of that exact type, making the conversion redundant.

```
def f(x: str) -> None:
    y = str(x)  # unnecessary-type-conversion
```
bool() is intentionally excluded since it is commonly used as an explicit truthiness check rather than a type conversion.

Fixes #3109

# Test Plan
- Added pyrefly/lib/test/unnecessary_type_conversion.rs with test cases covering: redundant str/int/float conversions, non-redundant conversions (int(x: bool)), Any arguments, and no-arg calls
- All existing tests pass (cargo test)

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1007,7 +1007,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         range: TextRange,
         errors: &ErrorCollector,
     ) {
-        let builtin_names = ["str", "int", "float"];
+        let builtin_names = ["str", "int", "float", "bytes"];
         if !builtin_names
             .iter()
             .any(|name| cls.has_qname("builtins", name))
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1000,6 +1000,37 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    fn check_unnecessary_type_conversion(
+        &self,
+        cls: &ClassType,
+        args: &[CallArg],
+        range: TextRange,
+        errors: &ErrorCollector,
+    ) {
+        let builtin_names = ["str", "int", "float"];
+        if !builtin_names
+            .iter()
+            .any(|name| cls.has_qname("builtins", name))
+        {
+            return;
+        }
+        if let Some(arg_ty) = self.first_arg_type(args, errors) {
+            let target_ty = self.heap.mk_class_type(cls.clone());
+            if !arg_ty.is_any() && arg_ty == target_ty {
+                self.error(
+                    errors,
+                    range,
+                    ErrorInfo::Kind(ErrorKind::UnnecessaryTypeConversion),
+                    format!(
+                        "Unnecessary `{}()` call; argument is already of type `{}`",
+                        cls.name(),
+                        arg_ty.deterministic_printing(),
+                    ),
+                );
+            }
+        }
+    }
+
     fn call_infer_with_callee_range(
         &self,
         call_target: CallTarget,
@@ -1092,6 +1123,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         }
                     }
                 };
+                self.check_unnecessary_type_conversion(&cls, args, arguments_range, errors);
                 let constructed_type = self.construct_class(
                     cls,
                     constructor_kind,
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1007,7 +1007,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         range: TextRange,
         errors: &ErrorCollector,
     ) {
-        let builtin_names = ["str", "int", "float", "bytes"];
+        let builtin_names = ["str", "int", "float", "bool", "bytes"];
         if !builtin_names
             .iter()
             .any(|name| cls.has_qname("builtins", name))
--- a/pyrefly/lib/test/unnecessary_type_conversion.rs
+++ b/pyrefly/lib/test/unnecessary_type_conversion.rs
@@ -65,6 +65,32 @@ def f(x: bytes) -> None:
 );
 
 testcase!(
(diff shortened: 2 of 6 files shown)


pyrefly/lib/alt/call.rs
```diff
@@ -1000,6 +1000,37 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
     }
 
+    fn check_unnecessary_type_conversion(
+        &self,
+        cls: &ClassType,
+        args: &[CallArg],
+        range: TextRange,
+        errors: &ErrorCollector,
+    ) {
+        let builtin_names = ["str", "int", "float"];
```
NathanTempest: bool is mentioned as a redundant data type in error_kind.rs, but it's not included in builtin_names here. bool is immutable, so bool(x) when x: bool is always redundant. Consider adding "bool" (and potentially "bytes") to the list.

pyrefly/lib/test/unnecessary_type_conversion.rs
```diff
@@ -0,0 +1,65 @@
...
+"#,
+);
+
+testcase!(
+    test_no_args_ok,
+    r#"
+def f() -> None:
+    y = str()  # OK - no argument
+"#,
+);
```
NathanTempest, line 99: Good test coverage but could you also add few cases:

bool(x) where x: bool — should warn (if bool is added to the checked types)
bytes(x) where x: bytes — same reasoning

NathanTempest (changes requested):

pyrefly/lib/alt/call.rs
```diff
@@ -1000,6 +1000,37 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
     }
 
+    fn check_unnecessary_type_conversion(
+        &self,
+        cls: &ClassType,
+        args: &[CallArg],
+        range: TextRange,
+        errors: &ErrorCollector,
+    ) {
+        let builtin_names = ["str", "int", "float", "bytes"];
```
NathanTempest: Bool should be added to the built_in names here

pyrefly/lib/test/unnecessary_type_conversion.rs
NathanTempest: bytes is added but bool is not, we are just trying to be as comprehensive as possible for our typechecker

website/docs/error-kinds.mdx
```diff
@@ -1372,6 +1372,17 @@ def test1(x: object) -> None:
 
 This check is relatively conservative and only warns on limited cases where the comparison is highly likely to be redundant.
 
+## unnecessary-type-conversion
+
+This warning is raised when a builtin type constructor (`str`, `int`, `float`, or `bool`) is called on a value that is already of that type, making the conversion redundant.
```
NathanTempest: add `bytes` to this built in types

NathanTempest (changes requested):

pyrefly/lib/alt/call.rs
ahaanbohra: I looked into adding `bool`, but I dont think it is always redundant. For example, in `test_bool_special_exports_bug`, `bool(x)` where `x: bool` is used in a conditional to drive type narrowing (e.g., narrowing to `Literal[True]` / `Literal[False]`). In this case, the call affects control flow, so flagging it as an unnecessary conversion would likely be incorrect.

Because of that, I’ve removed `bool` from the set of checked builtins and redundant data types for now. Would be glad to revisit if we're sure thats the direction we want, either by refining the lint to be context-aware (e.g., skipping conditionals) or by adjusting expectations if that’s the intended direction.

pyrefly/lib/alt/call.rs
ahaanbohra: Adding bool here causes test_bool_special_exports_bug to fail, where bool(x) is used in a conditional for type narrowing so it does not seem to be redundant. Happy to include it if we want to change that behavior, but wanted to flag this first.

NathanTempest: I think bool should be included in the lint. I think that the narrowing argument doesn't hold up. The test test_bool_special_exports_bug in generic_legacy.rs:820 shows:
```
def f(x: bool):
    if bool(x):
        assert_type(x, Literal[True])
```

But this narrowing comes from the if statement, not from bool(). Writing `if x:` produces the exact same narrowing because bool(x) is a no-op here. The lint flagging it as unnecessary would actually be correct since the user could just write `if x:`

At runtime, bool(x) where x: bool is always a no-op , bool.__new__ returns the same object unchanged. This is unlike bool(x) where x: str | None, which performs meaningful truthiness coercion and would not be flagged by this lint (since str | None != bool).

No other tool catches this today. Neither mypy, pyright, ruff, nor pylint flag bool(x) where x: bool. Adding it to pyrefly would be novel but correct and being comprehensive is a reasonable goal for a type checker.

One edge case to consider: bool(x) where x: Literal[True] is arguably not redundant since Literal[True] != bool the call widens the type. The lint should only fire when the argument type is exactly bool, not a Literal bool. Worth adding a test for this:
```
from typing import Literal
def f(x: Literal[True]) -> None:
    y = bool(x)  # OK - Literal[True] is not exactly bool
```
How about we implement this and check the mypy_primer results to see if any tests change. If the primer comes back clean, we have strong evidence that this doesn't cause false positives in real-world code. If it does flag something, we can use those concrete examples to decide whether to exclude bool or refine the lint.

ahaanbohra:
Added bool to the checked types and included the three test cases you suggested. Also updated test_bool_special_exports_bug in generic_legacy.rs to expect the lint, since bool(x) there is a no-op and if x produces the same narrowing.

The Literal[True] case is handled by the existing equality check and does not fire.

NathanTempest (approved):

kinto0 (changes requested):

kinto0 (approved):

PR #3143, opened 2026-04-15
fix bad-argument-count false positive with unpacking #3138

### LINKED ISSUE #3138: bad-argument-count false positive with unpacking
### Describe the Bug

Reproducer:

```
def f(x, y, z): ...

def test(xs: list[int]):
  f(*xs[-4:-2], 42)
```

Output:

```
ERROR Expected 3 positional arguments, got 4 in function `f` [bad-argument-count]
 --> /tmp/foo.py:4:17
  |
4 |   f(*xs[-4:-2], 42)
  |                 ^^
  |
 INFO 1 error
```

Even though `xs[-4:-2]` has no more than 2 elements (with indices -4 and -3).

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3138

Starred slice arguments like `*xs[-4:-2]` now carry a finite upper bound when the slice bounds are statically known, and the positional matcher reserves slots for later concrete positional arguments instead of greedily consuming everything.

just in case:

```
>>> xs = [1,2,3]
>>> xs[-4:-2]
[1]
```

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -24,6 +24,7 @@ use pyrefly_util::prelude::SliceExt;
 use pyrefly_util::prelude::VecExt;
 use pyrefly_util::visit::VisitMut;
 use ruff_python_ast::Expr;
+use ruff_python_ast::ExprSubscript;
 use ruff_python_ast::Identifier;
 use ruff_python_ast::Keyword;
 use ruff_python_ast::name::Name;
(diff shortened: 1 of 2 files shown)


rchen152:
This is an interesting one. None of [pyright](https://pyright-play.net/?pyrightVersion=1.1.408&reportUnreachable=true&code=CYUwZgBGAUAeA0ECeiBeBKAXBAdHgsAFBGiQAuIAzmXJdgDYCW1A2owHZkC6WREU0AFSxKLALQAWTGIBMXRBJnogA), [mypy](https://mypy-play.net/?gist=5c2473c8a698bb381df784ceacca2148), or [ty](https://play.ty.dev/58b716e0-5ad1-403b-aecf-6805e5b4be9f) supports the pattern that #3138 is asking for. From what I recall, it was intentional that we matched the behavior of other type checkers when deciding how to handle unpacking. If we want to track concrete slice indices, why only the upper bound? Don't we know the exact size? On the other hand, if we track the exact size, does that mean we should error if we now think a slice is contributing the wrong number of arguments?

Implementation-wise, I'd rather we convert `*xs[-4:-2]` to a fixed-length tuple, after which I think the existing logic would handle everything correctly. But this may need some more discussion first about desired behavior.

rchen152:
Closing this PR, since I think there are some open design questions we should work through first.

PR #3144, opened 2026-04-15
fix Recognize explicitly constructed properties as such #3141

### LINKED ISSUE #3141: Recognize explicitly constructed properties as such
### Describe the Bug

Reproducer:

```
class A:
  @property
  def p(self): ...


class B(A):
  p = property(lambda self: None)
```

Output:

```
ERROR Class member `B.p` overrides parent class `A` in an inconsistent manner [bad-override]
 --> /tmp/foo.py:7:3
  |
7 |   p = property(lambda self: None)
  |   ^
  |
  `B.p` and `A.p` must both be descriptors
```

It looks like Pyrefly does not treat `@property` and `property(...)` in the same way.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3141

The change normalizes `property(...)` assignments in class bodies into Pyrefly’s internal property representation instead of leaving them as generic descriptors.

That makes override checking treat `p = property(lambda self: None)` the same way as `@property`.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2172,7 +2172,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             return None;
         };
         let call = expr.as_call_expr()?;
-        let callee_errors = self.error_collector();
+        let callee_errors = self.error_swallower();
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #3145, opened 2026-04-15
fix improve enum error suggestion #3128

### LINKED ISSUE #3128: improve enum error suggestion
### Describe the Bug

```
from enum import StrEnum

class T(StrEnum):
    A = "a"

def f(t: T):
    ...
f("a")
```

current: 
Argument `Literal['a']` is not assignable to parameter `t` with type `T` in function `f` [[bad-argument-type](https://pyrefly.org/en/docs/error-kinds/#bad-argument-type)]

hope:
function f require `t: T`, but get "a", do you mean `T.A` ?

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3128

The change teaches call-argument diagnostics to suggest an enum member when the provided raw literal exactly matches a single expected enum value.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/answers_solver.rs
+++ b/pyrefly/lib/alt/answers_solver.rs
@@ -72,6 +72,7 @@ use crate::dispatch_anyidx;
 use crate::error::collector::ErrorCollector;
 use crate::error::context::ErrorInfo;
 use crate::error::context::TypeCheckContext;
+use crate::error::context::TypeCheckKind;
 use crate::error::style::ErrorStyle;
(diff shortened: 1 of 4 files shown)


rchen152 (changes requested):
The Copilot feedback looks legitimate here - specifically, the concerns about union handling and code duplication.

rchen152 (approved):
Thanks! I'll work on getting this merged.

stroxler (approved):

Issue #3154, opened 2026-04-15
Incorrect `bad-context-manager` error raised when encountering `sklearn.config_context`


fhoehle:
### Describe the Bug

When applying `pyrefly` to code using `scikit-learn`'s context manager `sklearn.config_context`, the false positive error `bad-context-manager` is raised.
Minimal example:
```python
# test_sklearn_cfg_ctx.py
import sklearn

with sklearn.config_context():
    pass
```
Executing `pyrefly`:
```bash
pyrefly --version
# pyrefly 0.61.0
pyrefly check test_sklearn_cfg_ctx.py
# ERROR Cannot use `Iterator[None]` as a context manager [bad-context-manager]
#  --> test_sklearn_cfg_ctx.py:3:6
#   |
# 3 | with sklearn.config_context():
#   |      ^^^^^^^^^^^^^^^^^^^^^^^^
#   |
#   Object of class `Iterator` has no attribute `__enter__`
#   Did you mean `__iter__`?
# ERROR Cannot use `Iterator[None]` as a context manager [bad-context-manager]
#  --> test_sklearn_cfg_ctx.py:3:6
#   |
# 3 | with sklearn.config_context():
#   |      ^^^^^^^^^^^^^^^^^^^^^^^^
#   |
#   Object of class `Iterator` has no attribute `__exit__`
#   Did you mean `__next__`?
#  INFO 2 errors


### Sandbox Link

The above example is not executable in the sandbox because the `scikit-learn` package is missing.

PR #3155, opened 2026-04-15
Add test and fix stub for `sklearn.config_context`

fhoehle:
# Summary

This PR adds a new test asserting that the `sklearn.config_context` context manager is correctly understood. 
The initial failing test is fixed by correcting the corresponding `sklearn.config_context` type hint and adding the `contextlib.contextmanager` decorator. 

Issue #3154 describes the observed error. The originating cause is the incorrect `sklearn-stub` in the `pyrefly_bundled` crate.

The deprecation warning regarding `Iterator[None]` is mitigated by replacing it with `Generator`

# Test Plan

- Executing the new (failing) test at `03b707d7bb6aeb55d9c5f2b6dd7b6e7367fe43a4`, which failed with the above-described `bad-context-manager` error.
  `cargo test -p pyrefly --lib test_sklearn`
- Successful execution at HEAD of this branch
- Successful run of `test.py`


rchen152 (changes requested):
Thanks for the PR! Pyrefly's bundled sklearn stubs are mirrored from here: https://github.com/facebook/pyrefly/blob/b20fa1ff099d546528eeb0f3df7b38835c13b2af/crates/pyrefly_bundled/third_party/stubs_metadata.json#L41. So rather than editing our bundled copy, it would be better to fix the stubs upstream and then bundle a fresh copy.

(If you get pushback on or lack of response to an upstream PR, we can consider doing a local patch instead. We'd need to record the patch somewhere, so we know what we've modified.)

fhoehle:
Many thanks for pointing out the mirroring I oversaw it. After updating the upstream source what are the concrete steps to propagate the change? The `update.py` script seems to update the `python/typeshed` stubs. I assume the stubs listed in `stubs_metadata.json` are updated manually so I will include the update here after the upstream change is merged. Is that correct?

rchen152:
@jvansch1 originally added the bundled stubs and can correct me if I'm wrong, but I believe we just manually copied them over.

jvansch1:
@fhoehle Like @rchen152 pointed out the process currently is to just manually copy them into the Pyrefly repo. Instead of doing this we could replace the existing stubs with a new version of the stubs.

fhoehle:
Current status: Waiting for an upstream maintainer to review the proposed changes.

rchen152:
@fhoehle Thank you for opening a PR with microsoft/python-type-stubs. It looks like the repo may be unmaintained now, which is unfortunate.

@jvansch1 Do you have any thoughts on what to do here? I was originally thinking that if we diverge from the upstream stubs, we should maintain a stack of patches to make it easier to cleanly reapply our changes on top of a new version, but maybe that's unnecessary if the original repo is inactive.

fhoehle:
@rchen152 I opened an [upstream issue](https://github.com/microsoft/python-type-stubs/issues/395) asking about maintainership and also pinged a Pylance maintainer who handles their releases. I'd propose we wait a little to see if anyone responds and clarifies the maintenance.

fhoehle:
Hi @rchen152 — the upstream PR is now merged, so I've bumped the vendored `sklearn-stubs` to that commit.
I also changed the added test `test_sklearn_config_context_is_context_manager` to exercise the real bundled stubs rather than an inline `sklearn-stub` — i.e., no hand-written copy. It ignores missing-source-for-stubs because `sklearn` isn't installed in the test env.

fhoehle:
@rchen152 @jvansch1 I'm happy for any feedback and comments.

yangdanny97:
@NathanTempest let's get this merged

NathanTempest (approved):
The fix looks to be correct and minimal, the test is well-designed, and the wiring matches conventions. Importing and merging

yangdanny97 (approved):

rchen152:
@fhoehle Thank you for seeing this through! Sorry for the lack of response earlier; I just got back from some time off.

PR #3156, opened 2026-04-15
also consider the `.py` of `.pyi` from stubs-only packages in  `pyrefly report` #3105

### LINKED ISSUE #3105: `pyrefly report`: For stubs-only packages, is the source package taken into account?
For example:

```py
# pkg/__init__.py
spam: str = "ham"
ham: int = 42
```

```pyi
# pkg-stubs/__init__.pyi
ham: int
```

Running `typestats check pkg-stubs` (or `typestats report pkg-stubs`) will report 50% coverage, because `spam` is missing in the stubs, and type-checkers will therefore not know that it's there (because they'll only look at the `.pyi` here, following the typing spec, so the annotation of `spam` in the `.py` will not be "seen").

Will `pyrefly report` report 100% or 50% coverage?
I.e., does `pyrefly report` combines the symbol tree of both `pkg` and `pkg-` packages, or does it *only* index the `.pyi`? 

I realize this is a bit paradoxical, because type-checkers ignore a `.py` if there's also a corresponding `.pyi`. 
But we're measuring type coverage, so we can't assume that a `.pyi` is always complete.

Relevant typestats docs: https://jorenham.github.io/typestats/reference/implementation/#stubs-overlay-merge

Assuming that pyrefly currently doesn't do this, we could also do this manually in typestats by running `pyrefly report` on both `pkg` and `pkg-stubs`. But I can imagine that this combined approach can be useful outside of typestats too.

jorenham:
# Summary

`pyrefly report` will now also take the source package sources into account when generating a report for stubs-only packages. Fo for e.g. `scipy-stubs` it'll now also look for `scipy` in the site packages and "overlay" the `.pyi` symbols onto the `.py` ones. This way, symbols missing the stubs will be reported as missing.

Fixes #3105 (which I accidentally closes already)

# Test Plan

Tests added

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -1514,9 +1514,14 @@ impl ReportArgs {
                     .as_ref()
                     .config_finder()
                     .python_file(h.module_kind(), h.path());
-                if let Some(py_module_path) =
-                    find_import_filtered(&config, h.module(), None, Some(ModuleStyle::Executable))
-                        .finding()
+                if let Some(py_module_path) = find_import_filtered(
+                    &config,
+                    h.module(),
+                    None,
+                    Some(ModuleStyle::Executable),
+                    None,
+                )
+                .finding()
                 {
                     let py_handle = config.handle_from_module_path(py_module_path);
                     external_handles.push(py_handle.clone());
@@ -1954,6 +1959,7 @@ mod tests {
             ModuleName::from_str("test"),
             None,
             Some(ModuleStyle::Executable),
+            None,
         )
         .finding()
         .expect("should discover test.py in site-packages");
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -20,6 +20,7 @@ use pyrefly_python::ignore::Ignore;
 use pyrefly_python::ignore::Tool;
 use pyrefly_python::module::Module;
 use pyrefly_python::module_name::ModuleName;
+use pyrefly_python::module_path::ModuleStyle;
 use pyrefly_python::nesting_context::NestingContext;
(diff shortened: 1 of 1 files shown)


pyrefly/lib/commands/report.rs
```diff
@@ -1909,6 +1935,39 @@ mod tests {
         compare_snapshot("partial_stub.expected.json", &report);
     }
 
+    /// `find_import_filtered` discovers .py files by module name in site-package-path.
+    #[test]
+    fn test_report_external_stub_discovery() {
+        use pyrefly_config::config::ConfigFile;
+        use pyrefly_python::module_path::ModuleStyle;
+
+        use crate::module::finder::find_import_filtered;
```
connernilsen: I don't think we should need all of these imports again, they should already be handled by the `use super::*` statement at the start of the test module

pyrefly/lib/commands/report.rs
```diff
@@ -1909,6 +1935,39 @@ mod tests {
         compare_snapshot("partial_stub.expected.json", &report);
     }
 
+    /// `find_import_filtered` discovers .py files by module name in site-package-path.
+    #[test]
+    fn test_report_external_stub_discovery() {
```
connernilsen: This test doesn't really seem to be validating the changes to the behavior above. Can you add a test for that? Can this be tested with `build_stub_module_report`?

connernilsen (changes requested):
Hey @jorenham, thanks for doing this. A few followups

connernilsen:
Let me know when it's ready for another round of review by re-requesting me :)

jorenham:
@connernilsen would you mind taking another look?

jorenham:
@connernilsen friendly bump

migeed-z (approved):

PR #3162, opened 2026-04-16
public symbol filtering with `pyrefly report --public-only` #3104

### LINKED ISSUE #3104: `pyrefly report`: (re-)exports
Typestats only considers public symbols for the stats. The way it determines whether a symbol is public, is by following all (re-)exports (`__all__`, or all non-private names if there's no `__all__`, and `import a as a`), and marking those as public. 

For example:

```py
# pkg/a.py
from ._b import _f
__all__ = ("_f",)

def g(): ...
```

Here, `pkg._b._f` is marked as public, but `g` is only marked public iff. it's publically re-exported in another public module.
With "public module" I mean one that's explicitly exported as public module, is called `__init__` or `__main__` and lives within a public package, or is a submodule of a public package with and has a non-private name (i.e. the module name does not start with `_`).

`pyrefly report` currently reports *all* symbols, and does not distinguish between public/private. AFAIK that's perfectly fine. 
However, for typestats, I'll have to detemrine when it's public. And for that I need to know the export graph, which I can't determine by looking at the `pyrefly report` at the moment. To be precise, for each module, I need to know which symbols it exports. I can figure that out if I know:

- the statically evaluated `__all__` names (so after "flattening" operations like `+=` and `.append`)
- the `import {name} as {name}` implicit re-exports (also if there's an `__all__`)

If there is no `__all__`, then I can piece together the public names by looking at the `names`of the module in the report json.
Note that I need to be able to distinguish between a module with an `__all__ = ()` and a module that has no `__all__`.

Would that be possible to include in `pyrefly report`?

It's not part of `typestats report`, but that's because those reports already only show the public symbols.

jorenham:
# Summary

In #3104 I proposed that `pyrefly report` also includes an export graph so we'd be able to figure out in `typestats` what the public symbols are as post-processing step.

With this change, that is no longer needed. Because now, `pyrefly report --public-only` will report only the public symbols, using the cross-module tracing algorithm from typestats. 

Right now this is opt-in. But we could also make this the default, and instead add a `--include-private` flag or something. Thoughts?

Fixes #3104

# Test Plan

Tests added

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -291,6 +291,35 @@ enum SymbolReport {
     },
 }
 
+impl SymbolReport {
+    fn name(&self) -> &str {
+        match self {
+            Self::Attr { name, .. }
+            | Self::Function { name, .. }
+            | Self::Class { name, .. }
+            | Self::Property { name, .. } => name,
+        }
+    }
+
+    fn name_mut(&mut self) -> &mut String {
+        match self {
+            Self::Attr { name, .. }
+            | Self::Function { name, .. }
+            | Self::Class { name, .. }
+            | Self::Property { name, .. } => name,
+        }
+    }
+
+    fn slots(&self) -> &SlotCounts {
+        match self {
+            Self::Attr { slots, .. }
+            | Self::Function { slots, .. }
+            | Self::Class { slots, .. }
+            | Self::Property { slots, .. } => slots,
+        }
+    }
+}
+
 #[derive(Debug, Serialize)]
 struct ModuleReport {
     /// Fully-qualified module name (e.g. "mypackage.submodule").
@@ -361,6 +390,10 @@ pub struct ReportArgs {
     /// from its filesystem layout.
     #[clap(long)]
     module: Option<String>,
+
+    /// Only report symbols reachable from public modules via re-export chains.
+    #[clap(long)]
+    public_only: bool,
 }
 
 impl ReportArgs {
@@ -376,6 +409,7 @@ impl ReportArgs {
             config_finder,
             self.prefer_stubs,
             self.module,
+            self.public_only,
             thread_count,
         )
     }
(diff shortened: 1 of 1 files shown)


connernilsen:
Hey @stroxler, are you familiar with Pyrefly report? Would you be able to take a look at this?

connernilsen:
Hey @stroxler, are you familiar with Pyrefly report? Would you be able to take a look at this?

pyrefly/lib/commands/report.rs
```diff
@@ -427,10 +461,123 @@ impl ReportArgs {
...
+            },
+        );
+        report.slots = slots;
+        report.coverage = slots.coverage();
+        report.strict_coverage = slots.strict_coverage();
+        report.n_methods = meth;
+        report.n_functions = fun;
+        report.n_classes = cls;
+        report.n_attrs = attr;
+        report.n_properties = prop;
```
connernilsen, line 578: should `n_function_params` and `n_method_params` also be recalculated here?

pyrefly/lib/commands/report.rs
```diff
@@ -427,10 +461,123 @@ impl ReportArgs {
...
     }
 
+    /// A module is public when every dotted-path component is public.
+    fn is_public_module(module: ModuleName) -> bool {
+        module.as_str().split('.').all(Self::is_public_name)
+    }
+
+    /// True if `fqn` or any class-level prefix of it is in the public set.
+    fn is_public_fqn(fqn: &str, module_prefix: &str, public_fqns: &HashSet<String>) -> bool {
+        let start = module_prefix.len();
```
connernilsen, line 471: Should we check if `fqn.starts_with(module_prefix)`? Otherwise, this might panic

pyrefly/lib/commands/report.rs
```diff
@@ -2270,4 +2416,196 @@ mod tests {
...
+            .map(String::from)
+            .collect();
+        ReportArgs::filter_module_report_to_public(&mut report, &public_fqns);
+
+        assert_eq!(report.names, vec!["pkg.Foo", "pkg.bar"]);
+        assert_eq!(report.symbol_reports.len(), 3); // Foo, Foo.method, bar
+        assert_eq!(report.n_functions, 1);
+        assert_eq!(report.n_methods, 1);
+        assert_eq!(report.n_classes, 1);
+        assert_eq!(report.n_attrs, 0);
```
connernilsen, line 2501: nit: can you assert param counts here too to be thorough?

connernilsen (changes requested):
Hey @jorenham, thanks for doing this! I have a few recommendations, but can you rebase and re-request review once you're ready again?

pyrefly/lib/commands/report.rs
jorenham: I rewrote it so that it avoids indexing altogether

pyrefly/lib/commands/report.rs
```diff
@@ -543,10 +579,126 @@ impl ReportArgs {
...
+        let module_prefix = format!("{}.", report.name);
+
+        report
+            .symbol_reports
+            .retain(|sym| Self::is_public_fqn(sym.name(), &module_prefix, public_fqns));
+        report
+            .names
+            .retain(|n| Self::is_public_fqn(n, &module_prefix, public_fqns));
+
+        // recompute all aggregates in a single pass
```
migeed-z, line 695: why is `n_type_ignores` skipped from the recalculation? could we add a comment to explain? sending back just for that point. Otherwise, everything LGTM. Thank you!

migeed-z (changes requested):

pyrefly/lib/commands/report.rs
jorenham: Because type ignores are unrelated to API visibility. But sure, I'll add a comment.

grievejia (changes requested):

jorenham:
@grievejia sorry but I can't see your review

migeed-z:
also adding @grievejia's remarks to the review

pyrefly/lib/commands/report.rs
```diff
@@ -543,10 +579,129 @@ impl ReportArgs {
...
+    }
+
+    /// True if `fqn` or any class-level prefix of it is in the public set.
+    fn is_public_fqn(fqn: &str, module_prefix: &str, public_fqns: &HashSet<String>) -> bool {
+        public_fqns.contains(fqn)
+            || std::iter::successors(fqn.rsplit_once('.').map(|(prefix, _)| prefix), |prefix| {
+                prefix.rsplit_once('.').map(|(parent, _)| parent)
+            })
+            .take_while(|prefix| prefix.starts_with(module_prefix))
+            .any(|prefix| public_fqns.contains(prefix))
```
migeed-z, line 607: `is_public_fqn` treats any member of a public class as public by accepting any class-level prefix in `public_fqns`, but it never checks whether the intermediate nested class/member names are themselves public. As a result, `--public-only` will still keep symbols like `pkg.Public._PrivateInner.method` whenever `pkg.Public` is exported, which means private nested classes and their members can leak into the filtered report.

pyrefly/lib/commands/report.rs
```diff
@@ -543,10 +579,129 @@ impl ReportArgs {
...
+
     /// Module-level dunders that typestats always excludes from the report.
     const EXCLUDED_MODULE_DUNDERS: &'static [&'static str] =
         &["__all__", "__dir__", "__doc__", "__getattr__"];
 
+    /// Collect origin FQNs of all publicly exported names across public modules.
+    fn compute_public_fqns(handles: &[Handle], transaction: &Transaction) -> HashSet<String> {
+        handles
+            .iter()
+            .filter(|h| Self::is_public_module(h.module()))
```
migeed-z, line 649: Public filtering seeds from `handle.module()` (the derived name), but `--module` only rewrites names afterward at line 1925. If the derived module name has a private segment (e.g. file lives under _`private/foo.py` and the user passes `--module public.foo --public-only`), is_public_module excludes the file here before the rename ever happens, and the report ends up empty even though the canonical name is public. Either thread the effective module name through the public-only path, or reject the` --module + --public-only `combination until the semantics are defined.

pyrefly/lib/commands/report.rs
```diff
@@ -1784,6 +1936,14 @@ impl ReportArgs {
             }
         }
 
+        if public_only {
+            let public_fqns = Self::compute_public_fqns(&handles, transaction);
```
migeed-z, line 1966: `compute_public_fqns` rewrites public re-exports to their origin FQN (line 641), but module_reports is only built from the originally requested handles (commands/check.rs:561 builds handles solely from the requested file set). On a file-scoped run like pyrefly report `pkg/__init__.py --public-only`, a re-export `pkg._internal.Foo` lands in `public_fqns as pkg._internal.Foo`, but no module report carries that prefix, so `__init__.pys` report empties out and gets dropped at line 1922. Either include traced origin handles in the report set, or keep the public alias when the origin module is outside the requested inputs.

migeed-z (commented):
Jia's comments

pyrefly/lib/commands/report.rs
jorenham: Good catch! I chose the path of least resistance so that it'll reject this combo :P

grievejia (approved):

PR #3163, opened 2026-04-16
include a schema version in `pyrefly report` #3103

### LINKED ISSUE #3103: `pyrefly report`:  schema versioning
There might be a reason to change the `pyrefly report` json schema in the future, but doing so will be a breaking change. A way around this is to include a schema version number in the json. That way, downstream tools, such as the typestats dashboard upload page (https://jorenham.github.io/typestats/dashboard/report/), can take into account these schema versions, and could even support multiple versions simultaneously.

For inspiration: `typestats report ...` currently uses a `{major}.{minor}` versioning scheme. Increments to the minor version are backwards compatible, but will trigger a CI job to regenerate all reports again. Increments to the major version are backwards incompatible, so if you upload an outdated report json to https://jorenham.github.io/typestats/dashboard/report/, it will kindly tell you that you need to update typestats, instead of silently breaking.

jorenham:
# Summary

This adds a `"{major}.{minor}"` field to the `FullReport`schema so that downstream projects are better able to deal with future schema changes. 
Typestats does the same thing.

Fixes #3103

# Test Plan

Test added

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -65,6 +65,9 @@ use crate::state::require::Require;
 use crate::state::state::State;
 use crate::state::state::Transaction;
 
+/// `(major, minor)` version for the report JSON schema.
+const REPORT_SCHEMA_VERSION: (u32, u32) = (0, 1);
+
 /// Slot-level annotation counts for a symbol.
 ///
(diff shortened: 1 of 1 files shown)


connernilsen (approved):
Hey @jorenham, thanks for doing this. I think this looks good

rchen152 (approved):

PR #3164, opened 2026-04-16
fix vararg unpacking bug in overload context #3161

### LINKED ISSUE #3161: false positive `bad-argument-type` for overloaded function with `*args: *tuple[...]`
### Describe the Bug

Here's the repro.zip:

```py
from typing import overload

@overload
def f(*args: *tuple[int]) -> int: ...
@overload
def f(*args: *tuple[int, int]) -> tuple[int, int]: ...

f(1)  # ❌ 
```

```
ERROR sandbox.pyi:8:2-5: Unpacked argument `tuple[Literal[1]]` is not assignable to parameter `*args` with type `tuple[int, int]` in function `f` [bad-argument-type]
```

Remove the second overload, or replace the `*args` with pos-only params, and the error disappears.

Context: 
We ran into this in https://github.com/scipy/scipy-stubs/pull/1552#discussion_r3088001120, and I managed to work around it in https://github.com/scipy/scipy-stubs/pull/1556.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIIANCGQE4xhSmEAuuAtlBWawBa50AWgBuMOnAiCABAF5pAHRABmQgEYALEoXpqINFjxEyECmDqdprUsQjoA5tIgdiuOq2m4xdKLlSYdHQABL3Fffx1MRmkwAAoAKlQ6ezhEaXjWAFdiWABtO1YAXQBKaSEAPid0VjTCOuDQnz8ArGi4xOTU9KycmHzqqiqi0oqrbLyCwYLC2vr0HTi1Yp0QAF8aVABjVggxADFoGAoDHAISUlNVoA

_No response_

jorenham:
# Summary

The type check for unpacked varargs was sending errors to `arg_errors` rather than to `call_errors`, which lead to false positives in an overload context.

Fixes #3161

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -874,7 +874,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 &unpacked_args_ty,
                 &unpacked_param_tuple,
                 arguments_range,
-                arg_errors,
+                call_errors,
                 &|| {
                     TypeCheckContext::of_kind(TypeCheckKind::CallVarArgs(
                         true,
--- a/pyrefly/lib/test/overload.rs
+++ b/pyrefly/lib/test/overload.rs
@@ -1863,3 +1863,21 @@ def test(value: int | float | None) -> str:
     return str(i)
     "#,
 );
+
+// Regression test for https://github.com/facebook/pyrefly/issues/3161
+testcase!(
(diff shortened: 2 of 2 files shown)


rchen152 (approved):
Good catch, thank you!

fangyi-zhou (approved):

PR #3165, opened 2026-04-16
fix Sequence pattern in match/case does not narrow tuple out of a union #3147

### LINKED ISSUE #3147: Sequence pattern in `match`/`case` does not narrow `tuple` out of a union
### Describe the Bug

```python


def f(value: float | tuple[float, float]) -> None:
    match value:
        case (_, _):
            pass
        case float():
            pass
        case _ as unreachable:
            assert_never(unreachable)

```

`unreachable` should be narrowed to `Never` after the sequence and class patterns have matched all arms of the union.

```
ERROR Argument `tuple[float, float]` is not assignable to parameter `arg` with type `Never` in function `typing.assert_never` [bad-argument-type]
  --> repro.py:11:26
   |
11 |             assert_never(unreachable)
   |                          ^^^^^^^^^^^
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2o4cGHwD66GADcRAHXRy5mGGDpgAFJNRQArjESqouVPwA%2BjbcVgBtMIeMAaA0YYBdAJR0AtAD46AOVwJRDk6ULoaYwBjAAs6TR09ELDkyMEYOjVRR1E3YPRkgrDiQTgkwtThJ2M1XLLCopK6lLS6UQE4Om10ShhUGNRsWDz6gpKRBnEpETUunr7ogdg3ORB7EDIe21JCBlooCgBiOgAFUk2oUjo0LDx8OkjAyDZtSmMIQMI5I4BlGHTohgMYhwRAAelBG2UF0IvDYoJg6FBmFwkTgoIe6CeLzegVBql4Ak00EW6QxWNeDHe%2BVwxEpgTgn3QZAY0UCnmklDgVLoAF46DIQABmQgARgATAK5FYRNROS45F0uDw%2BDBMJ5MBAepFKdJeXQAOQQNjoXgwfVyE0MTw9ACO2k1qs8AGsYKRPH1IvAOnz9QB3VCUdDm9AgAC%2Baz6OpgADFoDAKNccAQSORQ0A

_No response_

asukaminato0721:
# Summary

Fixes #3147

Sequence patterns like (_, _) were picking up Placeholder narrows from wildcard and other irrefutable subpatterns, and those placeholders intentionally block negative narrowing.

In a later case, that prevented the fallback arm from collapsing to Never.

now strip those placeholders when every sequence subpattern is irrefutable, so the carried negation stays precise.

# Test Plan

add test


asukaminato0721:
seems fixed by https://github.com/facebook/pyrefly/pull/3428 , though the same fix and even later lol.

PR #3168, opened 2026-04-16
LSP `unused-ignore` support #3167

### LINKED ISSUE #3167: `unused-ignore` not reported by LSP
### Describe the Bug

With `unused-ignore = "error"` in the pyrefly config and `pyrefly check` reports an unused ignore, it doesn't show up in the Pyrefly VSCode extension. I looked at the code a bit and it seems to be an issue with the LSP, rather than the VSCode extension.
As far as I can tell, the issue is that the the detection of unused ignore errors happens in a separate pass, but the LSP doesn't perform that pass.

### Sandbox Link

_No response_

_No response_

jorenham:
# Summary

The Pyrefly LSP will now report `unused-ignore` diagnostics when configured to do so.

Fixes #3167

# Test Plan

Tests added

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -2963,10 +2963,10 @@ impl Server {
                 diags.insert(handle_path_buf, Vec::new());
             }
         }
-        let collected = transaction.get_errors(handles).collect_errors();
-        let mut output_errors = collected.ordinary;
-        output_errors.extend(collected.directives);
-        for e in output_errors {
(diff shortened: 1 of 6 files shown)


connernilsen (changes requested):
Hey @jorenham, thanks for doing this. Seems like a useful thing to have.

Could you also have a test that checks what happens when no config is supplied? I'm not sure if that's something we have special logic to handle, but it would be great to document in a test

connernilsen (approved):
Thanks, it looks good!

rchen152 (approved):

PR #3173, opened 2026-04-17
count bare annotations as typed in `pyrefly report` #3172

### LINKED ISSUE #3172: `pyrefly report` counts `pi: Final = 3.14` as any
With Pyrefly 0.61.1, running `pyrefly report` on scipy-stubs shows the following

```json
        {
          "kind": "attr",
          "name": "scipy.constants._constants.golden",
          "n_typable": 1,
          "n_typed": 0,
          "n_any": 1,
          "n_untyped": 0,
          "location": {
            "line": 180,
            "column": 1
          }
        },
        {
          "kind": "attr",
          "name": "scipy.constants._constants.golden_ratio",
          "n_typable": 1,
          "n_typed": 0,
          "n_any": 1,
          "n_untyped": 0,
          "location": {
            "line": 181,
            "column": 1
          }
        },
```

for ([src](https://github.com/scipy/scipy-stubs/blob/c4933931f1c04ea40c9dda713ee88047f9355214/scipy-stubs/constants/_constants.pyi#L180-L181)):

```pyi
golden: Final = 1.618033988749895
golden_ratio: Final = golden
```

When I `reveal_type` it, it shows `float` ([sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AxVqigAddOIAKASUSDhUOgF46AZkIBGACwaArAE4ATADZdq3QA59Adn2rxlGADcYIgPpNiMABQyAlHQgADQgZI5gUKSEDLRQFADEdJKk4ZF0aFh4%2BHQAxrjokGwArpSoDBD5hOKJAMowMHQAFgwMxHCIAPQdYTARUbxsHTDoHZi4OXAdeQUQxaXl%2BR10YLx0qE6o0KjYsLn5hSVlFeh0uMQL6HBV6GQMjfkAtC6UcMfKdKIg6hqGn%2BIA2jBKNQXgBdcRFdBcHh8GCYB6YCCOHLlFzvADks3QvBg6PE2IYD0cAEciki4Q8ANYwUgPVA5HLwOAYgDuqEo6Dx6GCIAyOAIJFIEAoYGo9E8rA40N4-CE6BE4nELDkcpE72%2BOgMJjMlhsdgczlcUA8zB8LD84hAAF8QvTUTAhLAKHysoLhVagA)), so it's a bit strange that `pyrefly report` infers it as `Any`.

According to the typing spec ([url](https://typing.python.org/en/latest/guides/writing_stubs.html#module-level-attributes)) recommends omitting the explicit type in cases like `golden` where the type is obvious from context.

For e.g. `kibi: Final[int] = ...` (i.e. with explicit `Final` type) it is correctly counted as typed.

---

I'll see if I can fix this.

jorenham:
# Summary

This changes `typestats report` so that it will treat e.g. bare `Final` attrs and `_: list` as typed instead of any, in line with typestats.

Fixes #3172

# Test Plan

Added unit tests and integration tests

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -537,6 +537,16 @@ impl ReportArgs {
         }
     }
 
+    /// Classify an annotation slot from resolver output. Bare qualifiers
+    /// (e.g. `Final`) have unresolved annotation types but still count as typed.
+    fn classify_annotation_rank(has_annotation: bool, resolved_is_known: Option<bool>) -> SlotRank {
+        match resolved_is_known {
+            Some(is_known) => SlotRank::classify(true, is_known),
+            None if has_annotation => SlotRank::Typed,
+            None => SlotRank::Untyped,
+        }
+    }
+
     /// Returns true if the name is public: does not start with `_`, or is a dunder (`__x__`).
     /// Matches typestats `is_public_name`.
     fn is_public_name(name: &str) -> bool {
@@ -648,12 +658,12 @@ impl ReportArgs {
                         }
                         _ => None,
                     };
-                    let is_type_known = annotation_text.is_some()
-                        && answers
-                            .get_idx(*annot_idx)
-                            .and_then(|awt| awt.annotation.ty.as_ref().map(Self::is_type_known))
-                            .unwrap_or(false);
-                    let slots = SlotRank::classify(annotation_text.is_some(), is_type_known).into();
+                    let resolved_ty = answers
+                        .get_idx(*annot_idx)
+                        .and_then(|awt| awt.annotation.ty.as_ref().map(Self::is_type_known));
+                    let slots =
+                        Self::classify_annotation_rank(annotation_text.is_some(), resolved_ty)
+                            .into();
                     variables.push(Variable {
                         name: qualified_name,
                         annotation: annotation_text,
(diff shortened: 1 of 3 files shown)


connernilsen (approved):
I think this is good. I'll bring it in and see if there are comments internally

connernilsen:
Actually could you rebase first @jorenham? Then I'll pull it in. Just re-request a review from me

connernilsen (changes requested):
marking as changes requested for rebase

pyrefly/lib/commands/report.rs
```diff
@@ -711,15 +711,13 @@ impl ReportArgs {
...
-                    })
-                    .unwrap_or(false);
-            let slots = Self::classify_slot(annotation_text.is_some(), is_type_known);
+            let resolved_ty = annotation_idx.and_then(|idx| {
+                answers
+                    .get_idx(idx)
+                    .and_then(|awt| awt.annotation.ty.as_ref().map(Self::is_type_known))
+            });
+            let slots =
+                Self::classify_slot(resolved_ty.or(annotation_text.is_some().then_some(true)));
```
connernilsen, line 720: Should we check the annotation text here to be sure it's a `Final`? Would this classify something else like `value: Callable` as typed, even when its parameters are empty?

pyrefly/lib/commands/report.rs
connernilsen, line 860: Should we be doing `let resolved_ty = ...` here too?

connernilsen (changes requested):
rebase + two other things I found on an extra pass

pyrefly/lib/commands/report.rs
jorenham: I think it's ok for a `_: Callable` to count as typed, as it's equivalent to `_: Callable[..., Any]`. Because the idea behind `n_any` only counting things that resolve to `_: Any` (in typestats at least) is that this is equivalent to `_`, i.e. no type annotation at all.

kinto0 (approved):

PR #3174, opened 2026-04-18
fix Conformance: Missing type alias validation #2451

### LINKED ISSUE #2451: Conformance: Missing type alias validation
### Describe the Bug

All of the following validation is missing:

```
BadTypeAlias1 = eval("".join(map(chr, [105, 110, 116])))
BadTypeAlias6 = (lambda: int)()
BadTypeAlias7 = [int][0]
BadTypeAlias8 = int if 1 < 3 else str
BadTypeAlias12 = list or set

def bad_type_aliases(
    p1: BadTypeAlias1,  # should error: eval result is not a valid type
    p6: BadTypeAlias6,  # should error: lambda call is not a valid type
    p7: BadTypeAlias7,  # should error: list subscript is not a valid type
    p8: BadTypeAlias8,  # should error: conditional expr is not a valid type
    p12: BadTypeAlias12,  # should error: 'or' expr is not a valid type
):
    pass
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAQqpgCqnEwCCUEqcAjAAQC8XMAN1RQAFAB0QEwgCtcEdCIC2qYiIDGACwBOAGi4BtDgAYArHo7HzHAGwBdAJSOx6anQbNW7a7y4ioqRWxMVEQueQAXexF7Z1d6RhY2OAB2H30I230jW1iaeI8kgA4fCLCwLm4AHi4AZn4oOBguOHCtXLcEz04AJh9WFq5cLWaYcOdnTBhy7BoAfXD3WeEk%2BHF0Lg2uYg5QuPdE9g49LgBiZo1cAFcoTH4tLSHQwWEuLXhr8LC4LnRcT9QuEJWLcFoxnJsttZdnl9l1rMcznALtdbjB7o8uP5AsEuGphFAvj8-lwAUCICD3ODNsRktCOgV2MkEecrjc7g8tKF%2Bp84JdsHA1FoIMRPhBvr9-oDlhSwetqYU6fkDnBCsykazUejObjcFgIOEILqXjB8MRhmKiZKyTKYFSNttuorYUkON01ci2WiOaEAORDH38U3m8XE0nSrig23oeyIO1bdgIHQgS4GhokciIEBnACqqf1pC4YEu6DUBt1cHGWCmhaGynCs3Ql0CaJE%2BFCEXsXAAtAA%2BZqtWNyjZvcKXLTrMASAByTewaNCwHwAF8JM4QEmyG8wFBSIRwrhFFAKGcAAqkLc75oYHAEHXoSAAczHqDL6EIzjOAGUYE0NOFwsQcCIAA9MBm5TDuhBDA%2BwEwOgwGYLgahwMBai6o%2Bz6vsBNbDKgQjQKg2CwHeGFaC%2BhrrLgIoUXA77oGQ4QXOgXYCGicAUT4Eg1IQrqrugIBLkmqClhArEAGLQDAFBoFgeBEGQAlAA

_No response_

asukaminato0721:
# Summary

Fixes #2451

factored the syntax check into a reusable helper.

detect NameAssign bindings that were never real implicit aliases and surface the syntax error at the annotation use site.

preserves normal runtime-value typing while fixing the implicit-alias false negatives.

# Test Plan

update test

### DIFF
--- a/crates/pyrefly_types/src/type_info.rs
+++ b/crates/pyrefly_types/src/type_info.rs
@@ -90,7 +90,7 @@ pub struct TypeInfo {
 #[derive(Debug, Clone, PartialEq, Eq, Visit, VisitMut, TypeEq)]
 pub enum NameAssignTypeForm {
     RuntimeTypeValue,
-    InvalidImplicitAlias(String),
+    InvalidImplicitAlias(Box<str>),
 }
 
 impl TypeInfo {
--- a/crates/pyrefly_types/src/type_info.rs
+++ b/crates/pyrefly_types/src/type_info.rs
@@ -25,7 +25,7 @@ use crate::facet::FacetKind;
 use crate::types::AnyStyle;
 use crate::types::Type;
 
-assert_bytes!(TypeInfo, 40);
+assert_bytes!(TypeInfo, 64);
 
 /// The style of a phi.
 ///
@@ -84,17 +84,29 @@ impl<T> JoinStyle<T> {
 pub struct TypeInfo {
     ty: Type,
     facets: Option<Box<NarrowedFacets>>,
+    name_assign_type_form: Option<NameAssignTypeForm>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Visit, VisitMut, TypeEq)]
+pub enum NameAssignTypeForm {
+    RuntimeTypeValue,
+    InvalidImplicitAlias(String),
 }
 
 impl TypeInfo {
     pub fn of_ty(ty: Type) -> Self {
-        Self { ty, facets: None }
+        Self {
+            ty,
+            facets: None,
+            name_assign_type_form: None,
+        }
     }
 
     pub fn with_ty(self, ty: Type) -> Self {
         Self {
             ty,
             facets: self.facets,
+            name_assign_type_form: self.name_assign_type_form,
         }
     }
 
@@ -102,7 +114,17 @@ impl TypeInfo {
         Self {
             ty: f(self.ty),
             facets: self.facets,
+            name_assign_type_form: self.name_assign_type_form,
         }
+    }
+
+    pub fn with_name_assign_type_form(mut self, x: NameAssignTypeForm) -> Self {
+        self.name_assign_type_form = Some(x);
+        self
+    }
+
+    pub fn name_assign_type_form(&self) -> Option<&NameAssignTypeForm> {
+        self.name_assign_type_form.as_ref()
     }
 
     pub fn record_key_completion(&mut self, facets: &Vec1<FacetKind>, ty: Option<Type>) {
@@ -128,10 +150,12 @@ impl TypeInfo {
             Some(NarrowedFacet::WithoutRoot { facets, .. }) => Self {
                 ty: fallback(),
                 facets: Some(Box::new(facets.clone())),
+                name_assign_type_form: self.name_assign_type_form.clone(),
             },
             Some(NarrowedFacet::WithRoot(ty, narrowed_facets)) => Self {
                 ty: ty.clone(),
                 facets: Some(Box::new(narrowed_facets.clone())),
+                name_assign_type_form: self.name_assign_type_form.clone(),
             },
         }
     }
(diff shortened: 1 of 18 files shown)


pyrefly/lib/alt/solve.rs
```diff
@@ -623,6 +644,36 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         false
     }
 
+    fn implicit_type_alias_syntax_problem(&self, x: &Expr) -> Option<String> {
+        let Expr::Name(name) = x else {
+            return None;
+        };
+        let key = Key::BoundName(ShortIdentifier::expr_name(name));
+        let mut idx = self.bindings().key_to_idx_hashed_opt(Hashed::new(&key))?;
+        for _ in 0..100 {
```
migeed-z, line 653: why are we special casing to 100 here?

pyrefly/lib/test/type_alias.rs
```diff
@@ -1078,11 +1077,11 @@ BadTypeAlias8 = int if 1 < 3 else str
 BadTypeAlias12 = list or set
```
migeed-z, line 1077: Can we add a test on what happens if we import a bad alias such as:

from mod_a
import BadAlias

Is the approach extensible to that case?

pyrefly/lib/test/type_alias.rs
```diff
@@ -1078,11 +1077,11 @@ BadTypeAlias8 = int if 1 < 3 else str
...
 
 def bad_type_aliases(
-    p1: BadTypeAlias1,  # should error: eval result is not a valid type
-    p6: BadTypeAlias6,  # should error: lambda call is not a valid type
-    p7: BadTypeAlias7,  # should error: list subscript is not a valid type
-    p8: BadTypeAlias8,  # should error: conditional expr is not a valid type
-    p12: BadTypeAlias12,  # should error: 'or' expr is not a valid type
+    p1: BadTypeAlias1,  # E: Function call cannot be used in annotations
```
migeed-z, line 1080: Could you also add cases for ParamSpec and TypeVarTuple? They appear to be implicitly handled in the code

migeed-z (changes requested):

pyrefly/lib/test/type_alias.rs
asukaminato0721: fixed

pyrefly/lib/test/type_alias.rs
asukaminato0721: currently not

migeed-z:
Hmmm. I am not sure this approach is the right one. The reason is that something like this:

```
DynClass = type("DynClass", (), {})
def f(x: DynClass) -> None:  # E: Function call cannot be used in annotations
    pass
```

will fail and introduce false positives. The reason is that we are interpreting type(..) as a function call and rejecting it, but we should interpret the type of the value instead. So I think the approach needs to be modified to do that instead.

asukaminato0721:
so... should i also support mypy_extensions.TypedDict ?

migeed-z:
> so... should i also support mypy_extensions.TypedDict ?

I will take a look tomorrow :)

migeed-z:
I think this is a false positive. I don't know that we should support mypy_extensions.TypedDict directly. I believe the approach may have a gap. The allowlist in `implicit_alias_call_uses_runtime_type` needs to enumerate every valid callable, and any omission becomes a false positive.

Could we consider classifying at the definition site rather than tracing through bindings at every use site, so that calls that produce class types (like TypedDict, regardless of source module) work automatically without needing to be explicitly listed?

crates/pyrefly_types/src/type_info.rs
```diff
@@ -25,7 +25,7 @@ use crate::facet::FacetKind;
 use crate::types::AnyStyle;
 use crate::types::Type;
 
-assert_bytes!(TypeInfo, 40);
+assert_bytes!(TypeInfo, 64);
```
migeed-z: Do we need to store `name_assign_type_form` or would it be better to lookup this info on the fly?

migeed-z (changes requested):

rchen152:
Hey @asukaminato0721 - it looks like #2451 was recently fixed by https://github.com/facebook/pyrefly/commit/9fe64c22646d26a9d6b5d55da0976b65305fcf33. Sorry for the crossed wires!

PR #3178, opened 2026-04-19
fix Support functools.partial #3175

### LINKED ISSUE #3175: Support functools.partial
### Describe the Bug

ref https://github.com/astral-sh/ty/issues/1536

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3175

Implemented functools.partial support for concrete callable signatures.

When Pyrefly sees functools.partial applied to a normal list-parameter callable, it now rewrites the result to the remaining callable signature and validates the bound arguments up front.

cases like partial(f, 1) behave like a one-argument callable and over-applied constructors like partial(f, 1, "a", 2, "b", 3, "c", 4, "d") report an arity error.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -18,7 +18,6 @@ use pyrefly_types::types::CalleeKind;
 use pyrefly_types::types::NNModuleType;
 use pyrefly_types::types::TArgs;
 use pyrefly_types::types::TParams;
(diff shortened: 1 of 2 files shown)


yangdanny97:
Will close this for now since @shobhitmehro is working on an implementation, but we may end up reopening/merging this one if the final design is very close. thanks for your work here

PR #3179, opened 2026-04-19
fix remove or downgrade the already appear branch in match case #3015

### LINKED ISSUE #3015: remove or downgrade the already appear branch in match case
### Describe the Bug

Image

I think BB should appear first

```py
from enum import StrEnum, auto
class A(StrEnum):
    AA = auto()
    BB = auto()

def f(a: A):
    match a:
        case A.AA:
            ...
        case A.
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3015

now detect when attribute completion is happening inside a match value pattern, collect enum members already covered by earlier case arms, and demote those members in the completion ranking.

The ranking hook is wired and applied in the attribute-completion path.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -29,6 +29,7 @@ use ruff_python_ast::AnyNodeRef;
 use ruff_python_ast::ExprContext;
 use ruff_python_ast::Identifier;
 use ruff_python_ast::ModModule;
+use ruff_python_ast::Pattern;
 use ruff_python_ast::StmtImportFrom;
 use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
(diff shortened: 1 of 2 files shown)


rchen152 (changes requested):
Thanks for the change! This is wired into the right LSP completion path, but `deprioritize_previously_matched_enum_members` duplicates match analysis already represented by `Key::PatternNarrow`. Could you retrieve the enclosing case’s residual type from that key and feed it into the existing compatibility ranking instead? That should fix some edge cases that this PR currently doesn't get right, like guarded cases.

As a side note, **please request review again** once this PR is ready for re-review. You've probably noticed that we've been struggling a bit to keep up with PR review load - explicitly requesting review makes it much easier for me to tell when to take another look.

rchen152 (approved):

rchen152:
LGTM, thanks!

PR #3180, opened 2026-04-19
fix Support single-file scripts with PEP 723 metadata #3176

### LINKED ISSUE #3176: Support single-file scripts with PEP 723 metadata
### Describe the Bug

ref https://github.com/astral-sh/ty/issues/691

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3176

teaching config resolution to apply a per-file post-processing override for PEP 723 scripts instead of forcing a separate project config.

# Test Plan

add tests

### DIFF
--- a/crates/pyrefly_config/src/environment/environment.rs
+++ b/crates/pyrefly_config/src/environment/environment.rs
@@ -129,7 +129,10 @@ impl PythonEnvironment {
         command.arg("-c");
         command.arg(Self::query_script());
 
-        Self::get_env_from_command(command, interpreter.display().to_string())
+        Self::get_env_from_command(
+            command,
+            format!("Python interpreter (`{}`)", interpreter.display()),
+        )
     }
 
     /// Query the environment that `uv` would create for a PEP 723 script.
--- a/crates/pyrefly_config/src/environment/environment.rs
+++ b/crates/pyrefly_config/src/environment/environment.rs
@@ -125,7 +125,39 @@ impl PythonEnvironment {
             );
         }
 
-        let script = "\
+        let mut command = Command::new(interpreter);
+        command.arg("-c");
+        command.arg(Self::query_script());
+
+        Self::get_env_from_command(command, interpreter.display().to_string())
+    }
+
+    /// Query the environment that `uv` would create for a PEP 723 script.
+    pub fn get_env_from_uv_script(script: &Path) -> anyhow::Result<PythonEnvironment> {
+        let Some(script_dir) = script.parent() else {
+            return Err(anyhow!(
+                "Unable to query PEP 723 environment for `{}` because it has no parent directory",
+                script.display()
+            ));
+        };
+
+        let mut command = Command::new("uv");
+        command.arg("run");
+        command.arg("--directory");
+        command.arg(script_dir);
+        command.arg("--no-project");
+        command.arg("--isolated");
+        command.arg("--with-requirements");
+        command.arg(script);
+        command.arg("python");
+        command.arg("-c");
+        command.arg(Self::query_script());
+
+        Self::get_env_from_command(command, format!("uv PEP 723 script {}", script.display()))
+    }
+
+    fn query_script() -> &'static str {
+        "\
 import json, sys, sysconfig
 platform = sys.platform
 v = sys.version_info
@@ -133,26 +165,24 @@ version = '{}.{}.{}'.format(v.major, v.minor, v.micro)
 stdlib_paths = [p for p in [sysconfig.get_path('stdlib')] if p is not None]
 site_package_path = [p for p in sys.path if p != '' and '.zip' not in p and not p.endswith('/lib-dynload') and p not in stdlib_paths]
 print(json.dumps({'python_platform': platform, 'python_version': version, 'site_package_path': site_package_path, 'stdlib_paths': stdlib_paths}))
-";
+"
+    }
 
-        let mut command = Command::new(interpreter);
-        command.arg("-c");
-        command.arg(script);
-
+    fn get_env_from_command(
+        mut command: Command,
+        command_description: String,
+    ) -> anyhow::Result<PythonEnvironment> {
         let python_info = command.output()?;
 
         let stdout = String::from_utf8(python_info.stdout).with_context(|| {
-            format!(
-                "while parsing Python interpreter (`{}`) stdout for environment configuration",
-                interpreter.display()
-            )
+            format!("while parsing `{command_description}` stdout for environment configuration")
         })?;
         if !python_info.status.success() {
             let stderr = String::from_utf8(python_info.stderr)
                 .unwrap_or("<Failed to parse STDOUT from UTF-8 string>".to_owned());
             return Err(anyhow::anyhow!(
-                "Unable to query interpreter {} for environment info:\nSTDOUT: {}\nSTDERR: {}",
-                interpreter.display(),
+                "Unable to query {} for environment info:\nSTDOUT: {}\nSTDERR: {}",
+                command_description,
                 stdout,
                 stderr
             ));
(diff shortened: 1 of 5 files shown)


crates/pyrefly_config/src/environment/environment.rs
```diff
@@ -131,34 +131,64 @@ impl PythonEnvironment {
             );
         }
 
-        let script = "\
+        let mut command = Command::new(interpreter);
+        command.arg("-c");
+        command.arg(Self::query_script());
+
+        Self::get_env_from_command(command, interpreter.display().to_string())
```
connernilsen: I think it might be nice to add a little bit of context like in the env from uv script approach

```suggestion
        Self::get_env_from_command(command, format!("Python interpreter (`{}`)", interpreter.display()))
```

crates/pyrefly_config/src/environment/environment.rs
```diff
@@ -131,34 +131,64 @@ impl PythonEnvironment {
...
         let stdout = String::from_utf8(python_info.stdout).with_context(|| {
-            format!(
-                "while parsing Python interpreter (`{}`) stdout for environment configuration",
-                interpreter.display()
-            )
+            format!("while parsing `{command_description}` stdout for environment configuration")
         })?;
         if !python_info.status.success() {
             let stderr = String::from_utf8(python_info.stderr)
                 .unwrap_or("<Failed to parse STDOUT from UTF-8 string>".to_owned());
```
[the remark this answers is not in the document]
connernilsen: +1

pyrefly/lib/commands/config_finder.rs
```diff
@@ -136,6 +138,90 @@ pub fn default_config_finder_with_overrides(
...
+    }
+
+    fn script_path(path: &ModulePath) -> Option<&Path> {
+        match path.details() {
+            ModulePathDetails::FileSystem(path) | ModulePathDetails::Memory(path) => Some(path),
+            ModulePathDetails::Namespace(_)
+            | ModulePathDetails::BundledTypeshed(_)
+            | ModulePathDetails::BundledTypeshedThirdParty(_)
+            | ModulePathDetails::BundledThirdParty(_) => None,
+        }
```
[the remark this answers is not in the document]
connernilsen: This is a good point. Unless we have the in-memory contents, this will probably fail at runtime. If you can find a way to pass in the (I believe it should already be) pre-read file contents cleanly, it might be worth using here.

Otherwise, we should probably only look at disk contents, even if it may be stale.

pyrefly/lib/commands/config_finder.rs
```diff
@@ -186,6 +188,90 @@ pub fn default_config_finder_with_overrides(
...
+            match line.trim_start() {
+                "# /// script" if !in_script_block => {
+                    in_script_block = true;
+                    metadata.push(line);
+                }
+                "# ///" if in_script_block => {
+                    metadata.push(line);
+                    return Some(metadata.join("\n"));
+                }
+                _ if in_script_block => metadata.push(line),
```
connernilsen, line 335: Can we also return early here for the first non-comment line? I get the feeling that might signify the end of the script block.

pyrefly/lib/commands/config_finder.rs
```diff
@@ -186,6 +188,90 @@ pub fn default_config_finder_with_overrides(
     )
 }
 
+struct Pep723ScriptOverrides {
+    cache: Arc<Mutex<SmallMap<(PathBuf, String), ArcId<ConfigFile>>>>,
```
connernilsen: What if we changed this to
`Arc<Mutex<SmallMap<PathBuf, (String, ArcId<ConfigFile>)>>>`? Then we can get the config for a path and replace the old config + string instead of keeping a bunch of extra copies of the metadata around.

I'm thinking that editing this script directly might result in a bunch of slightly different metadata strings + configs stored in this cache, which might be worth clearing out more often than config invalidation.

Thoughts?

pyrefly/lib/commands/config_finder.rs
```diff
@@ -136,6 +138,90 @@ pub fn default_config_finder_with_overrides(
...
+            return Ok(config);
+        };
+        let Some(metadata) = Self::read_pep723_metadata(script_path)? else {
+            return Ok(config);
+        };
+
+        let cache_key = (script_path.to_path_buf(), metadata);
+        if let Some(config) = self.cache.lock().get(&cache_key) {
+            return Ok(config.dupe());
+        }
```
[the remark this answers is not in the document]
connernilsen: Kind of depends on the outcome of [this comment](https://github.com/facebook/pyrefly/pull/3180/changes#r3677438739)

pyrefly/lib/commands/config_finder.rs
```diff
@@ -136,6 +138,90 @@ pub fn default_config_finder_with_overrides(
...
+        script_environment.site_package_path = config.python_environment.site_package_path.clone();
+
+        let mut config = config.as_ref().clone();
+        config.interpreters.disable_query();
+        config.python_environment = script_environment;
+
+        let config = ArcId::new(config);
+        self.cache.lock().insert(cache_key, config.dupe());
+        Ok(config)
+    }
```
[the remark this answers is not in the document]
connernilsen: +1, would be good to add failed attempts to the cache (maybe some kind of result wrapping the `ArcId<ConfigFile>`?

pyrefly/lib/commands/config_finder.rs
```diff
@@ -262,6 +352,16 @@ pub fn standard_config_finder(
...
+        Box::new(move |_, path, config| {
+            pep723_script_overrides
+                .override_config(path, config)
+                .map_err(|err| {
+                    err.context(format!(
+                        "While resolving PEP 723 metadata for `{}`",
+                        path.as_path().display()
+                    ))
+                })
+        }),
```
connernilsen, line 433: I haven't looked much into it yet, but is it possible to move this logic into the part of `config.configure()` where we set up the environment? Then we won't need an extra post-configure override.

crates/pyrefly_config/src/finder.rs
```diff
@@ -273,6 +286,14 @@ impl ConfigFinder {
             ModulePathDetails::BundledTypeshed(_) => f(None),
             ModulePathDetails::BundledTypeshedThirdParty(_) => f(None),
             ModulePathDetails::BundledThirdParty(_) => f(None),
+        };
+
+        match (self.after)(name, path, config.dupe()) {
```
connernilsen, line 298: This worries me a little bit, since now we're producing a possibly different config here. That means some behaviors (like build systems) might not work correctly if we're expecting some files to have the same config.

I'm wondering if, instead of doing this here, we did it in `ConfigFile::get_sys_info()`, which would avoid producing an entirely new config here.

connernilsen (changes requested):
Hey @asukaminato0721, thanks for looking into this. I'm thinking it may be worth moving a lot of this into `ConfigFile::get_sys_info` rather than messing with the config finder to produce an entirely different config, since it might mess up some behavior around configs we have set up elsewhere.

Let me know your thoughts on this.

crates/pyrefly_config/src/finder.rs
asukaminato0721: added comment

pyrefly/lib/commands/config_finder.rs
asukaminato0721: fixed, has test

pyrefly/lib/commands/config_finder.rs
asukaminato0721: `ConfigConfigurer::configure()` runs before config selection and does not receive the script path.

Since PEP 723 is per-script and affects dependency search paths, I kept the override after selection.

PR #3181, opened 2026-04-20
impl moving items to new files

asukaminato0721:
# Summary

part of #2207

cc @rsheftel

Implemented the new refactor.move code action for “move symbol to new file”.

The new server-side action creates `<symbol>.py` beside the source file, moves the selected top-level symbol there, updates the source file, and rewrites direct consumer imports like from source import test to from test import test.

pushed the consumer-import rewrite into the shared move-symbol logic, fixing the reverse-dependency walk to use filesystem handles so workspace updates actually see other files.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/mod.rs
+++ b/pyrefly/lib/lsp/non_wasm/mod.rs
@@ -15,6 +15,7 @@ pub mod external_provider;
 pub mod folding_ranges;
 pub mod lsp;
 pub mod module_helpers;
+pub mod move_symbol_new_file;
 mod mru;
 pub mod protocol;
 pub mod queue;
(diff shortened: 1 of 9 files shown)


kinto0 (changes requested):
looks like tests are failing on windows

kinto0:
looks like it's still failing

NathanTempest:
Re-running the test to rule out the windows test failure bug that was fixed (https://github.com/facebook/pyrefly/pull/3518)

NathanTempest:
Hey @asukaminato0721 looks like the failure on windows was not due to an earlier bug and maybe due to the new test.

`Test: test_move_symbol_to_new_file_code_action`

Why this fails only on Windows ?  Due to platform-specific file URI handling. When file paths are converted to URIs:


Linux/macOS: file:///home/user/project/source.py
Windows: file:///C:/Users/user/project/source.py (with different encoding of separators and drive letters)

The test directly compares URI objects (lines 47, 101), but Windows encodes paths differently, causing the URI comparison to fail even though they represent the same files.

Fix would be to normalize URIs before comparison by converting them to file paths.

asukaminato0721:
@NathanTempest fixed

kinto0 (approved):

PR #3189, opened 2026-04-21
fix(suppression): handle ignores on multi-line implicit concatenations

### LINKED ISSUE #3188: Errors in multiline implicit string concatenation not properly suppressed
### Describe the Bug

When you write a `# pyrefly: ignore` comment above a multi-line implicit string concatenation, only the first error is covered, even though this is a single construct.

```python
# lol.py                                                                                     
x = (                                                                                                     
    # pyrefly: ignore [bad-index]                                                                         
    f"{d['foo']}\n"
    f"{d['bar']}"
)
```

In `pyrefly@0.60.0`, this results in an error in `pyrefly check`, even though the LSP does not error:

```bash
❯ uv run pyrefly check lol.py
ERROR Cannot index into `Literal['str']` [bad-index]
 --> lol.py:6:8
  |
6 |     f"{d['bar']}"
  |        ^^^^^^^^
  |
  No matching overload found for function `str.__getitem__` called with arguments: (Literal['bar'])
  Possible overloads:
  (key: SupportsIndex | slice[SupportsIndex | None, SupportsIndex | None, SupportsIndex | Unknown | None], /) -> LiteralString [closest match]
  (key: SupportsIndex | slice[SupportsIndex | None, SupportsIndex | None, SupportsIndex | Unknown | None], /) -> str
 INFO 1 error (1 suppressed)
```

I would expect one of the following:
- The multiline string is treated as a single construct, 
- When running `uv run pyrefly suppress lol.py`, each component of the multiline string receive their own suppressions. However, this does not happen.

```
❯ uv run pyrefly suppress lol.py

 INFO 1 error (1 suppressed)
 INFO Inserting error suppressions...
 INFO Finished suppressing errors in 1/1 files

❯ cat lol.py
─────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────
     │ File: lol.py
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ d = "str"
   2 │
   3 │ x = (
     │
   4 │     # pyrefly: ignore [bad-index]
   5 │     f"{d['foo']}\n"
   6 │     f"{d['bar']}"
   7 │ )
─────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────
```

### Sandbox Link

_No response_

_No response_

rexledesma:
# Summary
We did not respect ignore comments for the entirety of a multi-line implicit concatenation. This only happened in `pyrefly check`.

This occurred because we had thrown away the AST by the time we collect and check for errors. Instead, just reparse it.

Fixes https://github.com/facebook/pyrefly/issues/3188.

# Test Plan

- `cargo test`
- Manually (e.g. `cargo run --check`), with the following file:

```python
# lol.py
d = "lol"

x = (
    # pyrefly: ignore [bad-index]                                                                                 
    f"{d['foo']}\n" 
    f"{d['bar']}"
)
```

### DIFF
--- a/pyrefly/lib/state/state.rs
+++ b/pyrefly/lib/state/state.rs
@@ -33,6 +33,7 @@ use enum_iterator::Sequence;
 use fxhash::FxHashMap;
 use itertools::Itertools;
 use pyrefly_build::handle::Handle;
+use pyrefly_python::ast::Ast;
 use pyrefly_python::ignore::parse_ignore_all;
 use pyrefly_python::module::Module;
 use pyrefly_python::module_name::ModuleName;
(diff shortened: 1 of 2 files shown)


kinto0:
thank you for the fix! and sorry about the delayed review.

interesting. if this is true, we might have other bugs like this. I don't think that test covers it because the test would not evict. cc @maggiemoss

rexledesma:
Hi @kinto0 @maggiemoss, any update on this? If this is a wider issue, I'm happy to close my PR for you all to address it.

kinto0 (approved):

PR #3199, opened 2026-04-22
[Prototype] Add multi-connection IPC support to pyrefly TSP

heejaechang:
Teach the server to advertise and manage extra read-only channels for new TSP capabilities.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -31,6 +31,14 @@ use crossbeam_channel::Receiver;
 use crossbeam_channel::Sender;
 use dupe::Dupe;
 use dupe::OptionDupedExt;
+use interprocess::TryClone;
+#[cfg(unix)]
+use interprocess::local_socket::GenericFilePath;
+use interprocess::local_socket::ToFsName;
+use interprocess::local_socket::prelude::LocalSocketStream;
+use interprocess::local_socket::traits::Stream;
+#[cfg(windows)]
+use interprocess::os::windows::local_socket::NamedPipe;
 use itertools::Itertools;
 use lsp_server::ErrorCode;
 use lsp_server::RequestId;
@@ -494,6 +502,7 @@ pub struct Connection {
 pub enum MessageReader {
     Channel(Receiver<Message>),
     Stdio(BufReader<Stdin>),
+    Stream(BufReader<Box<dyn std::io::Read + Send>>),
 }
 
 impl MessageReader {
@@ -504,6 +513,7 @@ impl MessageReader {
         match self {
             MessageReader::Channel(r) => r.recv().ok(),
             MessageReader::Stdio(r) => read_lsp_message(r).ok().flatten(),
+            MessageReader::Stream(r) => read_lsp_message(r).ok().flatten(),
         }
     }
 }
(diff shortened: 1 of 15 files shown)


pyrefly/lib/tsp/server.rs
```diff
@@ -295,7 +548,11 @@ pub fn tsp_capabilities(
     indexing_mode: IndexingMode,
     initialization_params: &InitializeParams,
 ) -> ServerCapabilitiesWithTypeHierarchy {
-    // Use the same capabilities as LSP - TSP server supports the same features
-    // but will only respond to TSP protocol requests
-    capabilities(indexing_mode, initialization_params)
+    let mut result = capabilities(indexing_mode, initialization_params);
+    result.base.experimental = Some(serde_json::json!({
```
kinto0, line 552: instead of exposing all of base, can we instead add a `enable_multi_connection` API and call it here?

pyrefly/lib/tsp/server.rs
```diff
@@ -228,10 +299,192 @@ impl<T: TspInterface> TspServer<T> {
...
+        }
+
+        let (connection, mut reader, _io_thread) =
+            Connection::ipc(&pipe_name).map_err(|error| {
+                crate::tsp::validation::internal_error(&format!(
+                    "Failed to connect to IPC endpoint {pipe_name}: {error}"
+                ))
+            })?;
+
+        let extra_server = Self::with_connection(
```
kinto0, line 355: clever to use an `Arc` server in order for both servers to use the same memory. If we keep this approach (I wonder about instead abstracting around connections), we should document it clearly

pyrefly/lib/tsp/server.rs
```diff
@@ -228,10 +299,192 @@ impl<T: TspInterface> TspServer<T> {
...
+
+        extra_connections.insert(
+            pipe_name.clone(),
+            ExtraConnectionHandle {
+                close_tx: close_tx.clone(),
+            },
+        );
+        drop(extra_connections);
+
+        std::thread::spawn(move || {
```
kinto0, line 374: nit: can we abstract this out of here?

pyrefly/lib/tsp/server.rs
```diff
@@ -228,10 +299,192 @@ impl<T: TspInterface> TspServer<T> {
...
+            }
+        } else {
+            ConnectionRequestResult {
+                success: true,
+                message: Some(format!("Extra IPC connection already closed: {pipe_name}")),
+            }
+        }
+    }
+
+    fn get_pipe_name(&self, params: &ConnectionRequestParams) -> Result<String, ResponseError> {
```
kinto0, line 446: this would be a lot cleaner within the same module as `ConnectionRequestParams`. but I guess the protocol generation script doesn't allow that?

pyrefly/lib/tsp/server.rs
```diff
@@ -228,10 +299,192 @@ impl<T: TspInterface> TspServer<T> {
...
+            .filter(|pipe_name| !pipe_name.is_empty())
+            .cloned()
+            .ok_or_else(|| {
+                crate::tsp::validation::invalid_params_error(
+                    "Connection request args must include the IPC pipe name",
+                )
+            })
+    }
+
+    fn handle_extra_request<'a>(
```
kinto0, line 466: it's sketchy that a non-extra server can still call this function. maybe we need two server structs with two impls

kinto0 (commented):
some initial thoughts, but checking it out now to look closer

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -544,6 +554,47 @@ impl Connection {
         )
     }
 
+    pub fn ipc(pipe_name: &str) -> std::io::Result<(Self, MessageReader, IoThread)> {
+        #[cfg(windows)]
```
kinto0, line 558: I didn't realize this is OS-dependent. did we consider TCP? it seems like it could be simpler + just as fast in practice?

pyrefly/lib/tsp/server.rs
kinto0: I am testing an alternate approach here: https://github.com/facebook/pyrefly/pull/3203/changes/9e45b560ec2fb062398236970a1e8c87cc617d5b

what do you think?

pyrefly/lib/lsp/non_wasm/server.rs
heejaechang: chose IPC so we don't need to worry about exposing the connection to remote? (also things like vscode will auto detect tcp port being opened, and try to open forward channel by default and etc for remote vscode case) we could support tcp as well, but if we are not planning to support remote case yet, I think ipc is simpler.

and when ipc is used, client and server is on the same machine so we don't need to worry about client/server on different OSes and etc.

Issue #3187, opened 2026-04-21
Panic in report mode when @no_type_check decorator is used


amjames:
### Describe the Bug

Minimal repro below, real example was running `pyrefly report` pointed at `torch/testing/_internal/common_fsdp.py` ([codepointer](https://github.com/pytorch/pytorch/blob/f71088361f3c3325cc31736765b29e9260d8fd61/torch/testing/_internal/common_fsdp.py#L1067)) codex minified this down to the @no_type_check decorator. 

```python
# Repro command:
#   uvx pyrefly report repro.py

from typing import no_type_check


@no_type_check
def f(x: int):
    pass
```

Log
```
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/binding/bindings.rs:429:13:
Internal error: key not found, module `repro`, path `/.../repro.py`, key Hashed { hash: StarlarkHashValue(3454569666), key: Annotation(ShortIdentifier(114..115)) }
Backtrace:
   0:        0x102fe2450 - <std[2f7cd32bf4beae4b]::backtrace::Backtrace>::create
   1:        0x102e64ba8 - <std[2f7cd32bf4beae4b]::sync::once::Once>::call_once::<pyrefly_util[4cac38fcd6c06356]::panic::print_panic::{closure#0}>::{closure#0}
   2:        0x103155a58 - <std[2f7cd32bf4beae4b]::sys::sync::once::queue::Once>::call
   3:        0x102e64a88 - pyrefly_util[4cac38fcd6c06356]::panic::print_panic
   4:        0x102e65de8 - pyrefly_util[4cac38fcd6c06356]::panic::exit_on_panic::{closure#0}
   5:        0x102fbe248 - std[2f7cd32bf4beae4b]::panicking::panic_with_hook
   6:        0x102fddfe0 - std[2f7cd32bf4beae4b]::panicking::panic_handler::{closure#0}
   7:        0x102fddf9c - std[2f7cd32bf4beae4b]::sys::backtrace::__rust_end_short_backtrace::<std[2f7cd32bf4beae4b]::panicking::panic_handler::{closure#0}, !>
   8:        0x102fde57c - __rustc[f6c708c0d8e4bdd6]::rust_begin_unwind
   9:        0x1030d2b70 - core[1fab243bf906de66]::panicking::panic_fmt
[...]

PR #3211, opened 2026-04-23
Fix @no_type_check panic

kinto0:
Summary:
Add a regression test that reproduces the panic when running `pyrefly report` on functions decorated with `no_type_check`. The test currently asserts the panic behavior using `#[should_panic]`.

When running `pyrefly report` on a function like:
```python
from typing import no_type_check
no_type_check
def f(x: int):
    pass
```

The report command panics at `pyrefly/lib/binding/bindings.rs:429` in `key_to_idx_hashed` with the error: "Internal error: key not found, module `<module>`, path `<path>`, key Annotation(ShortIdentifier(...))"

Root cause: The binding pass skips creating `KeyAnnotation` entries for parameters of `no_type_check` decorated functions (since their bodies are not analyzed), but `parse_functions` in `report.rs` unconditionally tries to look them up.

A follow-up diff will fix the panic and update this test to assert correct behavior instead of the panic.

Differential Revision: D102023871

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -33,6 +33,7 @@ use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
 use serde::Serialize;
+use starlark_map::Hashed;
 use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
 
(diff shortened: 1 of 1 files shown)


jorenham:
Will symbols in `@no_type_check` functions/classes now be included in the report, or will they be ignored?

kinto0:
> Will symbols in `@no_type_check` functions/classes now be included in the report, or will they be ignored?

good question. I imagine it shouldn't. what do you think? cc @maggiemoss

jorenham:
> > Will symbols in `@no_type_check` functions/classes now be included in the report, or will they be ignored?
> 
> good question. I imagine it shouldn't. what do you think? cc @maggiemoss

Yea same, I think that those symbols should be ignored then.

kinto0:
> > > Will symbols in `@no_type_check` functions/classes now be included in the report, or will they be ignored?
> > 
> > 
> > good question. I imagine it shouldn't. what do you think? cc @maggiemoss
> 
> Yea same, I think that those symbols should be ignored then.

did you still want to take this over? or do you want me to finish it

jorenham:
> > > > Will symbols in `@no_type_check` functions/classes now be included in the report, or will they be ignored?
> > > 
> > > 
> > > good question. I imagine it shouldn't. what do you think? cc @maggiemoss
> > 
> > 
> > Yea same, I think that those symbols should be ignored then.
> 
> did you still want to take this over? or do you want me to finish it

I guess that depends on how quickly you are able to get this in. Does next week sound realistic?

kinto0:
closed from other fix, referenced in #3187

PR #3212, opened 2026-04-23
correlated condition analysis for simple guards #3185

### LINKED ISSUE #3185: Could `unbound-name` be made less noisy?
### Describe the Bug

```py
def main(a: bool) -> int:
    if a:
        b = 3
    
    c = 5

    if a:
        return b
    return 9
```

I get
```
ERROR sandbox.py:8:16-17: `b` may be uninitialized [[unbound-name](https://pyrefly.org/en/docs/error-kinds/#unbound-name)]
```

If `b` is defined within an `if a` block, and then only ever used in subsequent `if a` blocks, and `a` is never modified, then could the `unbound-name` not be reported?

---

side-note: I got an internal error the first time I tried putting into into the pyrefly sandbox 🤷 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABALaoToAUqiNeuUAlDQLQA%2BGkwAuiADroa04bTaSZijjQC8NAMwKZW6QGNVNAKySdsmvKlLpAJxgiArtanZTth05oBOEABoQZWzAoUkIRXDooCgBiGgAFUkDgmjQsPHwaXVx0SABzR1QRCCzCSRiAZRgYGgALERFiOEQAeiaA6mDCXGscpph0JsxcXTgmzOyIPOsCov6aMC7zADdGKFRsWAys3PzCrJpcYl30OBL0MhFqrL5FmGs4GYNxEHVCAEYAJifJAG1b6y64ABdST2dAQOjELoiGCYPiYCC2XSFG4GADkE3QXRgqMkmJEfFsAEd7AiYXwANYwUh8VC6XTwOBogDuqCcOPQIAAvn5aciYAAxaAwCgpHAEEjkTlAA

_No response_

jorenham:
# Summary

I had Claude Opus 4.7 take a stab at #3185, and to my (untrained) eyes it actually looks pretty decent. Since I'm not very familiar with this bit of the codebase, I opened it as a draft, with the intention of wanting to see mypy primer has to say about it, and to see if it's actually as easy

Fixes #3185

# Test Plan

Added some tests, and 
the failing test from https://github.com/facebook/pyrefly/commit/54cfb04b832ceefb389dc9030dfe23a5ca95eb9 now passes.

### DIFF
--- a/pyrefly/lib/test/flow_branching.rs
+++ b/pyrefly/lib/test/flow_branching.rs
@@ -1902,6 +1902,79 @@ def f(x: int) -> str:
 "#,
 );
 
+// Correlated-condition analysis: a name defined only on the truthy branch of
+// `if <guard>:` (with no else) is considered initialized inside any later
+// `if <guard>:` block where the guard has not been reassigned.
+testcase!(
+    test_correlated_if_simple,
+    r#"
+def main(a: bool) -> int:
+    if a:
+        b = 3
+    c = 5
+    if a:
+        return b
+    return 9
+"#,
+);
+
+testcase!(
+    test_correlated_if_no_match_without_guard,
+    r#"
+def main(a: bool) -> int:
+    if a:
+        b = 3
+    return b  # E: `b` may be uninitialized
+"#,
+);
+
+testcase!(
+    test_correlated_if_broken_by_reassignment,
+    r#"
+def main(a: bool, other: bool) -> int:
+    if a:
+        b = 3
+    a = other
+    if a:
+        return b  # E: `b` may be uninitialized
+    return 9
+"#,
+);
+
+testcase!(
+    test_correlated_if_does_not_cross_else,
+    r#"
+def main(a: bool) -> int:
+    if a:
+        b = 3
+    else:
+        pass
+    if a:
+        return b  # E: `b` may be uninitialized
+    return 9
+"#,
+);
+
+// When the guard branch contains a NoReturn call, the `MaybeInitialized` termination
+// keys are preserved through `InitializedIfGuardTruthy`, so reads in unrelated code
+// still correctly suppress the error via the Never-check.
+testcase!(
+    test_correlated_if_preserves_noreturn_fallback,
+    r#"
+from typing import NoReturn
+
+def raises() -> NoReturn:
+    raise Exception()
+
+def main(a: bool) -> int:
+    if a:
+        b = 3
+    else:
+        raises()
+    return b
+"#,
+);
+
 testcase!(
     test_declared_variable_with_noreturn_else_false_positive,
     r#"
--- a/pyrefly/lib/test/flow_branching.rs
+++ b/pyrefly/lib/test/flow_branching.rs
@@ -1902,45 +1902,8 @@ def f(x: int) -> str:
 "#,
 );
 
-// Correlated-condition analysis: a name defined only on the truthy branch of
-// `if <guard>:` (with no else) is considered initialized inside any later
-// `if <guard>:` block where the guard has not been reassigned.
-testcase!(
-    test_correlated_if_simple,
-    r#"
-def main(a: bool) -> int:
-    if a:
-        b = 3
-    c = 5
-    if a:
-        return b
-    return 9
-"#,
-);
-
-testcase!(
-    test_correlated_if_no_match_without_guard,
-    r#"
-def main(a: bool) -> int:
-    if a:
-        b = 3
-    return b  # E: `b` may be uninitialized
-"#,
-);
-
-testcase!(
-    test_correlated_if_broken_by_reassignment,
-    r#"
-def main(a: bool, other: bool) -> int:
-    if a:
-        b = 3
-    a = other
-    if a:
-        return b  # E: `b` may be uninitialized
-    return 9
-"#,
-);
-
+// An `else` branch means the name's uninitialized-ness survives the merge via
+// paths that are not controlled by `<guard>`, so the guarded upgrade is skipped.
 testcase!(
     test_correlated_if_does_not_cross_else,
     r#"
@@ -2480,7 +2443,6 @@ for x in range(10):
 // `if a:`, the variable is guaranteed to be initialized because the same
 // condition guards both the definition and the use.
 testcase!(
-    bug = "false positive: b is always initialized when a is truthy",
     test_guarded_initialization_basic,
     r#"
 def f(a: bool) -> int:
@@ -2488,7 +2450,7 @@ def f(a: bool) -> int:
         b = 3
     c = 5
     if a:
-        return b  # E: `b` may be uninitialized
+        return b
     return 9
     "#,
 );
@@ -2518,7 +2480,6 @@ def f(a: bool, c: bool) -> int:
 );
 
 testcase!(
-    bug = "false positive: b and c are always initialized when a is truthy",
     test_guarded_initialization_multiple_variables,
     r#"
 def f(a: bool) -> int:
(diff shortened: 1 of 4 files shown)


jorenham:
cc @MarcoGorelli

jorenham:
Hmm there seems to be an issue with ~mypy_primer~ github:

```
Collecting git+https://github.com/hauntsaninja/mypy_primer.git
  Cloning https://github.com/hauntsaninja/mypy_primer.git to /tmp/pip-req-build-bbd7y0eh
  Running command git clone --filter=blob:none --quiet https://github.com/hauntsaninja/mypy_primer.git /tmp/pip-req-build-bbd7y0eh
  error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
  fatal: expected 'packfile'
  error: subprocess-exited-with-error
  
  × git clone --filter=blob:none --quiet https://github.com/hauntsaninja/mypy_primer.git /tmp/pip-req-build-bbd7y0eh did not run successfully.
  │ exit code: 128
  ╰─> No available output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
ERROR: Failed to build 'git+https://github.com/hauntsaninja/mypy_primer.git' when git clone --filter=blob:none --quiet https://github.com/hauntsaninja/mypy_primer.git /tmp/pip-req-build-bbd7y0eh
```

https://www.githubstatus.com/

jorenham:
This is what I managed to salvage from the 4/15 primer jobs that succeeded:

rotki (https://github.com/rotki/rotki)
```diff
- ERROR rotkehlchen/db/drivers/gevent.py:211:117-127: `identifier` may be uninitialized [unbound-name]
- ERROR rotkehlchen/db/drivers/gevent.py:482:98-108: `identifier` may be uninitialized [unbound-name]
- ERROR rotkehlchen/tests/api/test_database.py:55:29-45: `backup2_contents` may be uninitialized [unbound-name]
- ERROR rotkehlchen/tests/api/test_database.py:57:29-45: `backup1_contents` may be uninitialized [unbound-name]
```

static-frame (https://github.com/static-frame/static-frame)
```diff
- ERROR static_frame/core/db_util.py:576:32-44: `query_create` may be uninitialized [unbound-name]
- ERROR static_frame/core/frame.py:1013:29-38: `fields_dc` may be uninitialized [unbound-name]
- ERROR static_frame/core/frame.py:6147:68-77: `drop_mask` may be uninitialized [unbound-name]
- ERROR static_frame/core/frame.py:6149:62-71: `drop_mask` may be uninitialized [unbound-name]
- ERROR static_frame/core/frame.py:9580:36-49: `columns_names` may be uninitialized [unbound-name]
- ERROR static_frame/core/type_blocks.py:175:48-57: `drop_mask` may be uninitialized [unbound-name]
- ERROR static_frame/core/type_blocks.py:180:45-54: `drop_mask` may be uninitialized [unbound-name]
- ERROR static_frame/core/type_blocks.py:271:48-57: `drop_mask` may be uninitialized [unbound-name]
- ERROR static_frame/core/type_blocks.py:276:45-54: `drop_mask` may be uninitialized [unbound-name]
- ERROR static_frame/core/util.py:1554:27-38: `is_not_none` may be uninitialized [unbound-name]
- ERROR static_frame/core/util.py:1563:20-31: `is_not_none` may be uninitialized [unbound-name]
- ERROR static_frame/core/util.py:3142:40-44: `cols` may be uninitialized [unbound-name]
- ERROR static_frame/core/util.py:3149:40-44: `cols` may be uninitialized [unbound-name]
- ERROR static_frame/core/util.py:3177:48-52: `cols` may be uninitialized [unbound-name]
- ERROR static_frame/core/util.py:3210:61-65: `cols` may be uninitialized [unbound-name]
```

pandas (https://github.com/pandas-dev/pandas)
```diff
- ERROR pandas/core/common.py:603:32-41: `old_value` may be uninitialized [unbound-name]
- ERROR pandas/tests/reshape/test_get_dummies.py:112:72-82: `fill_value` may be uninitialized [unbound-name]
```

mypy (https://github.com/python/mypy)
```diff
- ERROR mypy/test/testpep561.py:159:23-30: `program` may be uninitialized [unbound-name]
- ERROR mypyc/codegen/emitclass.py:419:64-82: `shadow_vtable_name` may be uninitialized [unbound-name]
- ERROR mypyc/irbuild/ll_builder.py:1093:28-32: `skip` may be uninitialized [unbound-name]
- ERROR mypyc/irbuild/ll_builder.py:2787:37-48: `false_block` may be uninitialized [unbound-name]
```

meson (https://github.com/mesonbuild/meson)
```diff
- ERROR mesonbuild/backend/ninjabackend.py:717:20-52: `captured_compile_args_per_target` may be uninitialized [unbound-name]
- ERROR mesonbuild/modules/_qt.py:760:38-45: `results` may be uninitialized [unbound-name]
- ERROR unittests/allplatformstests.py:5462:27-29: `cc` may be uninitialized [unbound-name]
```

core (https://github.com/home-assistant/core)

```diff
- ERROR homeassistant/components/bayesian/config_flow.py:557:25-34: `sub_entry` may be uninitialized [unbound-name]
- ERROR homeassistant/components/bayesian/config_flow.py:572:70-79: `sub_entry` may be uninitialized [unbound-name]
- ERROR homeassistant/components/opnsense/__init__.py:118:50-65: `interfaces_resp` may be uninitialized [unbound-name]
- ERROR homeassistant/components/recorder/util.py:126:49-60: `timer_start` may be uninitialized [unbound-name]
- ERROR homeassistant/components/tplink/config_flow.py:459:27-29: `un` may be uninitialized [unbound-name]
- ERROR homeassistant/components/tplink/config_flow.py:459:34-36: `pw` may be uninitialized [unbound-name]
- ERROR homeassistant/components/tplink/entity.py:678:45-64: `has_parent_entities` may be uninitialized [unbound-name]
- ERROR homeassistant/helpers/entity.py:1375:17-28: `update_warn` may be uninitialized [unbound-name]
```

spack (https://github.com/spack/spack)
```diff
- ERROR lib/spack/spack/config.py:728:34-42: `comments` may be uninitialized [unbound-name]
```

kornia (https://github.com/kornia/kornia)
```diff
- ERROR kornia/feature/disk/disk.py:117:39-40: `h` may be uninitialized [unbound-name]
- ERROR kornia/feature/disk/disk.py:117:43-44: `w` may be uninitialized [unbound-name]
```

pyrefly/lib/test/flow_branching.rs
```diff
@@ -1902,6 +1902,42 @@ def f(x: int) -> str:
 "#,
 );
 
+// An `else` branch means the name's uninitialized-ness survives the merge via
```
migeed-z, line 1905: Thank you! 

I think a sound solution for this is tricky. For context, this PR https://github.com/facebook/pyrefly/pull/3253 has all the unittests with the corner cases and the expected result. I think we should add all of them here first since that might reveal some gaps.

According to what I tried in the PR, supporting all these cases adds a lot of complexity to the code, so it's unclear to me if we should or shouldn't support the feature. If we can find an elegant solution that passes all of these tests and doesn't add complexity nor increase the runtime complexity of the code, we should add the feature.

migeed-z (changes requested):

pyrefly/lib/test/flow_branching.rs
jorenham: Hm yea, I can see how a complete solution would require a lot of bookkeeping. But looking at the mypy_primer diff, it might be worth it either way. 

Maybe something like second backwards pass would be easier to implement? I.e. keep things as they are in the first pass, then backtrack for each unbound variable and see if the relevant flow paths (and only those) justify its use.

jorenham: Either way, your modal-logic-based approach seems to be a lot more robust than this hacky stuff. Should I close this, or do you think there's anything worth salvaging here?

migeed-z: Yeah, I think we might want to shelve this for now. I think neither solution works, so we should discuss more

jorenham:
Closing because https://github.com/facebook/pyrefly/pull/3212#discussion_r3170627041

PR #3218, opened 2026-04-23
Implement multi-connection architecture

kinto0:
Summary:
Multi-connection architecture for the TSP (Type Server Protocol)

The commit splits the single monolithic TspConnection into a three-tier structure:
1. TspServer<T> — shared core holding the Arc<T> LSP server, the snapshot counter, and a registry of extra connections. All connections share this via Arc.
  2. TspMainConnection<T> — wraps a TspConnection (via Deref) and owns the main stdio event loop. Only this type can:
    - Process LSP lifecycle events (delegating to the inner server)
    - Handle typeServer/connection requests (open/close extra connections)
    - Broadcast snapshotChanged notifications to all connections
  3. TspExtraConnection<T> — also wraps TspConnection (via Deref). Spawned on demand over IPC named pipes. Can handle TSP query requests (getComputedType, resolveImport, etc.) but explicitly rejects connection management and LSP methods.

Key mechanics:
  - Extra connections are opened via a new ConnectionRequest with type: "open" and an IPC pipe name. Each gets its own reader thread + select loop, terminating on a close signal or pipe disconnect.
  - snapshotChanged notifications are broadcast from the main connection to all extra connections, so every client stays in sync.
  - The server advertises typeServerMultiConnection in its experimental capabilities.
  - All TSP request handlers now call self.inner() (method) instead of self.inner (field), since the inner server is accessed through the shared TspServer.

The result: multiple IDE clients can query the same pyrefly process concurrently over IPC without spawning separate server instances.

Differential Revision: D102228875

### DIFF
--- a/crates/tsp_types/protocol_generator/tsp.json
+++ b/crates/tsp_types/protocol_generator/tsp.json
@@ -66,6 +66,20 @@
     ],
     "requests": [
         {
+            "method": "typeServer/connection",
+            "typeName": "ConnectionRequest",
+            "messageDirection": "clientToServer",
+            "documentation": "Main-connection-only request used to open or close an extra TSP transport. Extra transports must remain read-only and must not be used for LSP traffic.",
+            "params": {
+                "kind": "reference",
+                "name": "ConnectionRequestParams"
+            },
+            "result": {
+                "kind": "reference",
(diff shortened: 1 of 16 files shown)


maggiemoss (approved):

PR #3223, opened 2026-04-24
stubgen: emit reduced field metadata in dataclass / Pydantic stubs

### LINKED ISSUE #3221: bug(stubgen): dataclass fields with only `field(...)` / non-literal defaults lose `= ...` in generated stubs, breaking `__init__` checking
## Summary

### Input

```python

@dataclass
class A:
```

### Expected result

```python

@dataclass
class A:
```

OR

```python

@dataclass
class A:
    name: str | None = None
```

OR

```python

@dataclass
class A:
    name: str | None = ...

        ...
```

### Actual result

```python

@dataclass
class A:
    name: str | None
```

### Notes

tobyh-canva:
## Context

As identified in #3221, stub generation for class body annotated assignments only handled **simple** RHS values (e.g. literals, `...`). `dataclasses.field` / Pydantic `Field` calls are not "simple" text, so stubs either dropped the real `__init__` / field behavior or were forced into lossy workarounds. Callers that care about **`kw_only`**, **defaults**, and **field metadata** need something closer to the real declaration without copying the whole module.

## Intent

Make `.pyi` output for `field(...)` and `Field(...)` **type-checking oriented**: keep a **reduced** call that still reflects init-altering options, strip runtime-only or validation noise, and avoid echoing **heavy** `default_factory` bodies. **Optional** Pydantic schema doc strings on `Field` should follow the same `include_docstrings` switch as other stub doc content.

## Changes

- **Stub extract (`extract.rs`)**  
  - For class `AnnAssign` RHS, after simple literals, try a **structured strip** of `field` / `Field` (by callee name, including `dataclasses.field` / `pydantic.Field`-style attributes).  
  - **Whitelist** kwargs per "kind" (dataclass vs Pydantic `Field`): defaults, `default_factory`, `init`, `kw_only` (dataclass), aliasing-related kwargs (Pydantic), etc.; drop the rest (e.g. `metadata`, `min_length`, …).  
  - **`default_factory`:** keep only **simple callable refs** (bare name or `a.b` chains); otherwise emit **`default_factory=lambda: ...`**.  
  - With **`include_docstrings`**, Pydantic `Field` also keeps **`description`** and **`title`**.  
  - **Fallback:** if the RHS is not a re-printable `field`/`Field` (e.g. `*args`), use the solver plus **`ClassField::dataclass_flags_of`** to emit **`= ...`** when the solved field’s init parameter has a default, matching the existing init-default notion (with the documented Pydantic caveat for dual name/alias params).  
- **`ClassField`:** `dataclass_flags_of` is **`pub(crate)`** for stubgen.  
- **Tests (`stubgen/mod.rs`):** dataclass + Pydantic coverage for stripping, `default_factory`, complex factory placeholder, mixed fields, and docstring retention under config.

## Testing

- `cargo test -p pyrefly stubgen::tests::`

---

Fixes #3221.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1110,7 +1110,7 @@ impl ClassField {
         }
     }
 
-    fn dataclass_flags_of(&self, heap: &TypeHeap) -> DataclassFieldKeywords {
+    pub(crate) fn dataclass_flags_of(&self, heap: &TypeHeap) -> DataclassFieldKeywords {
         match &self.0 {
             ClassFieldInner::Property { .. } => DataclassFieldKeywords::new(),
             // Descriptors are always initialized in the class body (otherwise they wouldn't
--- a/pyrefly/lib/stubgen/extract.rs
+++ b/pyrefly/lib/stubgen/extract.rs
@@ -18,8 +18,11 @@ use pyrefly_build::handle::Handle;
 use pyrefly_python::module::Module;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_python::sys_info::SysInfo;
+use pyrefly_graph::index::Idx;
 use pyrefly_types::callable::Param;
+use pyrefly_types::class::ClassDefIndex;
 use pyrefly_types::display::TypeDisplayContext;
+use pyrefly_types::keywords::DataclassFieldKeywords;
 use pyrefly_types::types::Type;
 use ruff_python_ast::Expr;
 use ruff_python_ast::Operator;
@@ -33,7 +36,9 @@ use starlark_map::Hashed;
 
 use crate::alt::answers::Answers;
 use crate::alt::types::decorated_function::DecoratedFunction;
+use crate::binding::binding::BindingClass;
 use crate::binding::binding::Key;
+use crate::binding::binding::KeyClass;
 use crate::binding::binding::KeyClassField;
 use crate::binding::binding::KeyDecoratedFunction;
 use crate::binding::bindings::Bindings;
@@ -127,6 +132,15 @@ pub fn extract_module_stub(
 
     let dunder_all = resolve_dunder_all(&ast.body, &module_info);
 
+    let class_field_by_def_and_name: HashMap<(ClassDefIndex, Name), Idx<KeyClassField>> =
+        bindings
+            .keys::<KeyClassField>()
+            .map(|idx| {
+                let k = bindings.idx_to_key(idx);
+                ((k.0, k.1.clone()), idx)
+            })
+            .collect();
+
     let mut ctx = ExtractionContext {
         bindings: &bindings,
         answers: &answers,
@@ -136,6 +150,8 @@ pub fn extract_module_stub(
         uses_self: false,
         function_map: &function_map,
         dunder_all: &dunder_all,
+        class_field_by_def_and_name,
+        class_def_index_stack: Vec::new(),
     };
 
     let items = extract_stmts(&ast.body, &mut ctx, false);
@@ -158,6 +174,9 @@ struct ExtractionContext<'a> {
     /// When `__all__` is explicitly defined, only these names are exported
     /// at module level. `None` means no explicit `__all__` — use convention.
     dunder_all: &'a Option<HashSet<Name>>,
+    class_field_by_def_and_name: HashMap<(ClassDefIndex, Name), Idx<KeyClassField>>,
+    /// Innermost class is at the end (nested `class` bodies).
+    class_def_index_stack: Vec<ClassDefIndex>,
 }
 
 fn extract_stmts(stmts: &[Stmt], ctx: &mut ExtractionContext, in_class: bool) -> Vec<StubItem> {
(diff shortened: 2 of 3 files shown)


tobyh-canva:
I'm not 100% sure if this is the best approach, I considered emitting any field with a default as `= ...`, but unfortunately this causes information like `kw_only = True` and `init = False` to be lost. Plus, keeping the field description when `--include-docstrings` is enabled is kinda neat.

I guess an alternative approach would be for stubgen to emit an explicit `__init__()` constructor, derived from the fields. Presumably, this would make the type-checker perform better when run on these stubs, since more "inference" work is done ahead of time and materialised in the stubs.

Please let me know what approach you would prefer :)

NathanTempest:
Hey @tobyh-canva thanks for starting contributions to our repo, a necessary check before importing this into our repo is to sign the CLA - https://code.facebook.com/cla

NathanTempest:
Also a quick heads-up, this diff is also trying to achieve the same goal - https://github.com/facebook/pyrefly/pull/3225

Maybe this could be a collab between you both, I would suggest to reach out to them on discord to prevent any redundant commits / conflicts

tobyh-canva:
Thanks @NathanTempest, I'll try and get myself added to Canva's corporate CLA tomorrow

tobyh-canva:
I can see a few issues on #3225 and it hasn't been progressing, so I'm going to assume the author isn't pursuing that one further. 

Following on from my [previous comment](https://github.com/facebook/pyrefly/pull/3223#issuecomment-4310588703), I've made an alternative (draft) fix for the issue in #3306 which changes the stub output. In that version, dataclass/model attributes are emitted with an elided default `= ...`, and an explicit `def __init__` constructor is emitted to retain the signature which would otherwise be inferred during type-checking. If you would prefer one approach over the other, please let me know, and I can focus on just one of them :) 

I think I personally prefer #3306 as the stubs are simpler and the type-checker has less work to do due to the explicit constructors, so type-checking performance should be better when using these stubs for dependencies.

NathanTempest:
Thank you @tobyh-canva for you continued efforts, and also acknowledging the alternative #3306. The fallback strategy, and the test cases look good enough for me to stamp. However the bigger question is which direction we want stubgen to go. This PR (preserve field(...) calls with reduced kwargs) vs #3306 (emit = ... and synthesize an explicit __init__). Both fix #3221 but this PR keeps the stub close to the user's source, low-risk, and works today. Hardcoded to field/Field by name. #3306 is more uniform, would naturally extend to all PEP 681 dataclass_transform specifiers, but more invasive.

Tagging @rechen for a steering call. Which of the two approaches do you prefer ?

rchen152:
Sorry, I responded to @NathanTempest in an internal chat last week and forgot to copy my comment over:

> I'm not very familiar with the stubgen/ directory, so it's sort of hard for me to compare the approaches, given that I don't know which one is more feasible/fits more naturally into the existing design, but a couple high-level comments on each:
> * #3223 seems to be detecting field specifiers by looking for hardcoded field / Field names. That's fine for plain dataclasses, but for dataclass transforms, it would be nice if we could check what field specifiers the dataclass transform declares.
> * #3306 seems to be implementing its own logic to construct a synthesized __init__ for a dataclass? Again, I don't know how stubgen hooks into the rest of Pyrefly, so maybe this is necessary for technical reasons, but it's re-implementing logic that already exists in Pyrefly, which doesn't seem ideal

From a quick look at the commit history, I think @yangdanny97 might know the most about how stubgen works.

tobyh-canva:
Yeah from a user / output perspective it just comes down to how we want dataclass / pydantic stubs to look.

Either like this, basically very similar to actual source code (this PR):

```py
@dataclass
class Person:
    name: str
    age: int = field(default=0, kw_only=True)
```

Or like this, more explicit and stub-like (#3306):
```py

@dataclass
class A:
    name: str
    age: int = ...

    def __init__(name: str, *, age: int = ...):
        ...
```

I personally prefer the latter. My actual implementation in #3306 likely needs work so don't pay too much attention to the code itself, but rather the intent 😅

yangdanny97:
@tobyh-canva I'm curious to learn more about why you prefer the second approach

tobyh-canva:
@yangdanny97 Some reasons I prefer the 2nd approach:

1. Performance - when type-checking your source code against these stubs (assuming you're bringing these stubs for your dependencies), the type-checker has to do less work to synthesise the `__init__` constructors at runtime, compared.
    - At Canva, with the way we've integrated Pyrefly with Bazel, we get Bazel to generate stubs for all libraries used in a target, and those serve as inputs for the type-checking action. These stubs are cached and only need to be re-generated when those external libraries change (not very often), whereas type-checking needs to run every time your own source code changes (very frequent). Hence for us, stubs are a form of cached type inference and have huge performance benefits. Though I think the same principle should apply to non-Bazel applications :)
2. As a general rule I think stubs should aim to be an explicit set of method declarations for simplicity, performance, and user understanding. Users can inspect stubs to see how Pyrefly has synthesised the constructor of dataclasses / pydantic models and give a very clear picture of the constructor signature.

I'm open to either approach though, and I'm keen to hear your thoughts too :)

yangdanny97:
Thanks for sharing your thoughts, here are a few of my own:

In favor of skipping `__init__`: Pyrefly synthesizing the `__init__` is fairly cheap, and skipping it would reduce the line count of the generated stubs (if compactness matters).

In favor of emitting the `__init__`: This requires less custom/additional logic in Pyrefly, since the logic to generate the signature was already pre-existing. Users reading the stub may also benefit from the constructor being explicit.

I'm happy to proceed with emitting `__init__` in the stub for now, we can always revisit that decision in the future.

tobyh-canva:
Sounds like a plan. I'll close this PR and spend some today getting #3306 ready to review :)

PR #3225, opened 2026-04-24
Fix #3221: preserve dataclass field defaults in stubgen

### LINKED ISSUE #3221: bug(stubgen): dataclass fields with only `field(...)` / non-literal defaults lose `= ...` in generated stubs, breaking `__init__` checking

Arths17:
Summary
Stubgen was dropping non-literal dataclass field defaults from generated stubs, which caused synthesized `__init__` signatures to look required when they were actually optional.

Fix
Treat annotated class fields inside dataclasses as having a default sentinel when the source expression is not a simple literal. This preserves the public constructor shape in generated `.pyi` output without changing literal defaults.

Files changed
- `pyrefly/lib/stubgen/extract.rs`
- `pyrefly/lib/stubgen/mod.rs`
- `pyrefly/lib/test/stubgen/dataclasses/input.py`
- `pyrefly/lib/test/stubgen/dataclasses/expected.pyi`

Testing
- `cargo test -p pyrefly stubgen::tests::test_stubgen_dataclasses -- --exact`

Notes
- The unrelated local workspace edit in `.vscode/tasks.json` was left untouched and not included in the commit.


kinto0 (changes requested):
is there a way to avoid the huge binaries being added to the codebase?

kinto0:
looks like someone else already started this in #3223. next time please comment on the issue when you're working on it so we can assign accordingly! if you want to collaborate with them, ping them in the discord. I'll leave this open until we discuss this more

PR #3234, opened 2026-04-24
Fix TSP server crash on Jupyter notebook cell URIs (#3234)

kinto0:
Summary:

Pyrefly's TSP server crashed when Pylance sent `vscode-notebook-cell:` URIs for Jupyter notebook cells, failing with "Invalid params: URI must use the file:// scheme".

The root cause was that `parse_file_uri` rejected any non-`file://` scheme as an error, and `get_type_at_position` called `url.to_file_path()` which fails for non-file schemes. Additionally, `get_type_at_position` hardcoded `notebook_cell: None`, so even if the URI were resolved, position offsets would be wrong for multi-cell notebooks.

This diff:
- Adds `parse_uri` which accepts any valid URI scheme (replacing `parse_file_uri` in all handlers)
- Adds `resolve_uri_to_path` and `maybe_get_cell_index` to `TspInterface`, leveraging the existing LSP notebook infrastructure (`open_notebook_cells`, `path_for_uri_or_notebook_cell`) that is already live inside the shared `Server` instance
- Updates all 5 TSP handlers (`resolveImport`, `getDeclaredType`, `getComputedType`, `getExpectedType`, `getPythonSearchPaths`) to resolve notebook cell URIs
- Fixes `get_type_at_position` to properly look up cell indices for position offsetting in multi-cell notebooks

Reviewed By: rchen152

Differential Revision: D102376882

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -474,6 +474,17 @@ pub trait TspInterface: Send + Sync {
         &self,
         func_id: &pyrefly_types::callable::FuncId,
     ) -> Option<TextRange>;
+
+    /// Resolve a URI to a filesystem path.
+    ///
+    /// Handles both `file://` URIs (via [`Url::to_file_path`]) and notebook
+    /// cell URIs (via the `open_notebook_cells` map). Returns `None` when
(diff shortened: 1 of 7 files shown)


rchen152 (approved):

rchiodo (commented):
Thanks looks great :)

PR #3243, opened 2026-04-25
fix assert_type with metaclasses #3228

### LINKED ISSUE #3228: assert_type with metaclasses
### Describe the Bug

Pyrefly is ok with the following:
```
class meta(type):
    ...
class A(metaclass=meta):
    ...
reveal_type(A) # type[A]
# The following is ok with Pyrefly.
# Other type checkers: expected "meta" but received "type[A]"  
assert_type(A, meta) 
```
However, erroring is more on line with the spec since `type[A]` is not equivalent to `meta`

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2o4cGHwD6TYjAA66AMZRBcOjRgNUACgkwAlIhl0DdQsZnzFdAILqVas0IC8N1Lv2HjhGZRgA3GKijizDDqFtp0AMSMQQDaFgC6MpEAKgAWMHRguFBQuADurBwQSrgA1nT5DCl0AAqkXmBQpB7okQDylSJRknSyabIlInCIdDD4krIMMJh0UiBOs3TYAK78XrIwEL7Ts1qxCSAGMooiDIGSIQA0yqrOM%2BgyIBcgZPWNhAy0UBSRta%2BkdGgsHh8D1cOhIGwlpRUAwIGDmpEAMowdIpBgMYhDAD0WJeMAaTV4bCxMHQWMwuFkcCxsjBEKhMLhZIyvAE3lQ0FQ2FgoPBEEh0NhYLouGIQvQcGaZEqYIAtL5KHAmXR7DMQABmQgARgATLMZNERNRFQl0Et0FweHwprLMBA1rDfCq6AByfnoXgwF0yD0MWVeACOS3tNoGpFlqFk6yEzpduVQlHQ3vQIAAvk9I46YAAxaAwCiAnAEEjkVNAA

_No response_

asukaminato0721:
# Summary

Fixes #3228

assert_type no longer accepts a class object as exactly equal to its non-type metaclass.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/special_calls.rs
+++ b/pyrefly/lib/alt/special_calls.rs
@@ -57,7 +57,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 TypeFormContext::FunctionArgument,
                 errors,
             ));
-            if !self.is_equivalent(&a, &b) {
+            let matches_metaclass_instance = match (&a, &b) {
+                (Type::ClassDef(got), Type::ClassType(want)) => {
+                    !want.is_builtin("type") && self.type_order().has_metaclass(got, want)
+                }
+                (Type::Type(box Type::ClassType(got)), Type::ClassType(want)) => {
+                    !want.is_builtin("type")
+                        && self.type_order().has_metaclass(got.class_object(), want)
+                }
+                _ => false,
+            };
+            if matches_metaclass_instance || !self.is_equivalent(&a, &b) {
                 self.error(
                     errors,
                     range,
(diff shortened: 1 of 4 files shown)


grievejia (changes requested):

kinto0:
from @grievejia 

> I'm not certain this is the correct location to address the issue. [assert_type(A, B)](https://typing.python.org/en/latest/spec/directives.html#assert-type) is defined in the spec to check for equivalence—meaning mutual subtyping for static types. The diff seems to alter this behavior, which would reduce our spec compliance.
The real problem appears to be in the shared subset/equivalence logic. It's concerning that we currently treat Meta as a subtype of type[A]; this may be the root cause. The assert_type() issue is likely just a symptom.

yangdanny97 (approved):

PR #3249, opened 2026-04-27
fix Language server does not expose read/write metadata for symbols #3246

### LINKED ISSUE #3246: Language server does not expose read/write metadata for symbols
### Describe the Bug

I followed the [installation docs](https://pyrefly.org/en/docs/IDE/#zed) for Zed although I assume this affects all editors.

I'm using the Night Owl Dark theme which [defines](https://github.com/elGusto/night-owlz/blob/29b5fedfee923b2598d97520962215d26d51d67d/themes/NightOwl.json#L92-L93) the default read/write background highlights as the following.

```json
"editor.document_highlight.read_background": "#ffffff08",
"editor.document_highlight.write_background": "#5f7e9779",
```

If you override those to make reads have maximum transparency (`00`) and writes have maximum opacity (`FF`) as follows then you'll notice that highlighting goes away completely.

```json
"theme_overrides": {
  "Night Owl Dark": {
    "editor.document_highlight.read_background": "#ffffff00",
    "editor.document_highlight.write_background": "#5f7e97FF"
  }
},
```

If you remove Pyrefly as the language server then you'll notice that the default server properly highlights only instances of assignment.

The tracking issue https://github.com/facebook/pyrefly/issues/2207 marks document highlighting as complete but I think it's not until this is fixed.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3246

Document highlights now set LSP DocumentHighlightKind::WRITE for assignment/declaration-style references and READ for normal uses.

# Test Plan

add test


pyrefly/lib/test/lsp/lsp_interaction/document_highlight.rs
```diff
@@ -0,0 +1,59 @@
...
+use lsp_types::Url;
+use lsp_types::request::DocumentHighlightRequest;
+use serde_json::json;
+
+use crate::object_model::InitializeSettings;
+use crate::object_model::LspInteraction;
+use crate::util::get_test_files_root;
+
+#[test]
+fn document_highlight_includes_read_write_kind() {
```
kinto0: I wonder if instead this test could be a `test/lsp/document_highlight.rs` test? since it's only a single file. then we don't need to write to disk

pyrefly/lib/test/lsp/lsp_interaction/document_highlight.rs
```diff
@@ -0,0 +1,59 @@
...
+    let root = get_test_files_root();
+    let path = root.path().join("document_highlight.py");
+    std::fs::write(&path, "x = 1\ny = x\n").unwrap();
```
kinto0: for these lsp_interaction tests, we really should just write the file to disk in the files directory. then the tests don't do IO and we can actually open them in the IDE to have real-world tests

pyrefly/lib/test/lsp/lsp_interaction/document_highlight.rs
```diff
@@ -0,0 +1,59 @@
...
+                    "end": { "line": 0, "character": 1 }
+                },
+                "kind": 3
+            },
+            {
+                "range": {
+                    "start": { "line": 1, "character": 4 },
+                    "end": { "line": 1, "character": 5 }
+                },
+                "kind": 2
```
kinto0: it would be nice to have DocumentHighlightKind::READ/WRITE here

pyrefly/lib/state/lsp.rs
```diff
@@ -2914,6 +2917,29 @@ impl<'a> Transaction<'a> {
         .concat()
     }
 
+    pub fn local_reference_is_write(&self, handle: &Handle, range: TextRange) -> bool {
```
kinto0: instead of this function, can we have the following setup?

- caller (document_highlight) calls self.identifier_at
- IdentifierContext has a `impl` for `is_write` that is called from document_highlight
- we can avoid this overly-specific `local_reference_is_write` function

kinto0 (changes requested):
approach looks great, thanks for picking this up!

just a few code style suggestions

kinto0 (approved):
thank you!! very clean

connernilsen (approved):

PR #3250, opened 2026-04-27
[TSP] Add module uri to module tsp types

rchiodo:
# Summary

As found in these two issues:
https://github.com/microsoft/pylance-release/issues/8013
https://github.com/microsoft/pylance-release/issues/7999

Pyrefly wasn't returning a URI for module types. This prevents Pylance from finding the actual module and prevents completions and other things from working.

Here's a picture of Pylance now showing completions for sys:

image

I also updated the 'literal' types to use the real builtins.

# Test Plan

Unit tests and manual test with Pylance.

/cc @kinto0

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -475,7 +475,7 @@ pub trait TspInterface: Send + Sync {
         func_id: &pyrefly_types::callable::FuncId,
     ) -> Option<TextRange>;
 
-    /// Resolve an importable module to its backing file URI using the
+    /// Resolve an importable module to its backing filesystem path using the
     /// import-resolution context of `source_uri`.
     ///
     /// Returns `None` when the source URI is invalid, cannot be mapped to a
@@ -484,7 +484,7 @@ pub trait TspInterface: Send + Sync {
         &self,
         source_uri: &str,
         module: &pyrefly_types::module::ModuleType,
-    ) -> Option<String>;
+    ) -> Option<PathBuf>;
 
     /// Resolve a URI to a filesystem path.
     ///
@@ -6327,7 +6327,7 @@ impl TspInterface for Server {
         &self,
         source_uri: &str,
         module: &pyrefly_types::module::ModuleType,
-    ) -> Option<String> {
+    ) -> Option<PathBuf> {
         let url = Url::parse(source_uri)
             .ok()
             .or_else(|| Url::from_file_path(source_uri).ok())?;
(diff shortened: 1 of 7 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -475,6 +475,13 @@ pub trait TspInterface: Send + Sync {
         func_id: &pyrefly_types::callable::FuncId,
     ) -> Option<TextRange>;
 
+    /// Resolve an importable module name to its backing file URI using the
+    /// import-resolution context of `source_uri`.
+    ///
+    /// Returns `None` when the source URI is invalid, cannot be mapped to a
+    /// path, or the target module cannot be resolved.
+    fn resolve_module_uri(&self, source_uri: &str, module_name: &str) -> Option<String>;
```
kinto0: to be consistent with resolve_uri_to_path?
```suggestion
    fn resolve_module_uri(&self, source_uri: &str, module_name: &str) -> Option<PathBuf>;
```

kinto0: might make sense to clarify what this does. does the `module_name` and `source_uri` have to match? or is it equivalent to importing `module_name` from within `source_uri`?

kinto0: looks like it's the latter

kinto0: 1) what happens when there's an alias? does pylance always request the fully qualified module_name?

2) how does pylance know which module_name to use when import roots are not always known? example:
```
src/
    foo.py
    pyrefly.toml
```
pyrefly might import `foo` but pyright might import `src.foo`. does this API require a stable root?

kinto0 (changes requested):
general approach seems to make sense, but back for questions about edge cases

pyrefly/lib/lsp/non_wasm/server.rs
rchiodo: This isn't called externally. It's used by Pyrefly internally when creating module TSP types. Which come from module nodes. That's where the module name comes from.

rchiodo: This would make sense if this was a client facing URI, but this is internal. It's supposed to return a string.

rchiodo: It's more like the resolve_func_def_range function that was used to find declarations.

rchiodo: Maybe I should change it to take the module type itself instead of a name. That would make it more explicit.

rchiodo: I switched it. It does look more uniform that way.

kinto0 (approved):

maggiemoss (approved):

PR #3251, opened 2026-04-27
Fix notebook cell index mismatch

connernilsen:
Summary:
LspNotebook::get_cell_index returned the index among all cells (code + markdown), but callers pass it to Notebook::cell_offsets() which is indexed by valid code cells only. For notebooks with markdown cells, this mismatch caused from_lsp_position to resolve to the wrong byte offset — or panic — because the all-cells index would hit the wrong entry in cell_offsets (e.g., the trailing sentinel instead of the target cell's start offset).

Fix get_cell_index to translate the all-cells index to a code-cell index by counting preceding code cells. Returns None for non-code cells.

Differential Revision: D102678315


kinto0 (approved):

PR #3261, opened 2026-04-28
Fix false positive bad-class-definition for class methods

### LINKED ISSUE #3259: `bad-class-definition` due to Class Method Assignments when using `dataclass_transform()`
### Describe the Bug

Pyrefly incorrectly picks up fields from any class method assignments (e.g., inside `@classmethod` or `__init_subclass__`). At runtime, these assignments are ignored by Python's `dataclasses.dataclass` mechanism.

This leads to a false positive `bad-class-definition` error when a subclass defines a field without a default value.

Other static type checkers (Pytype and MyPy) are OK with the above and do not emit any errors.

### Reproduction Example

```python
import typing
@typing.dataclass_transform()
class ModuleBase:
  ...

class Module(ModuleBase):
  @classmethod
  def foo(cls) -> None:
    cls.field: typing.Any = None

class SmallModule(Module):
  x: int
```
Pyrefly reports:
```
ERROR sandbox.py:21:3-4: Dataclass field `x` without a default may not follow dataclass field with a default [[bad-class-definition](https://pyrefly.org/en/docs/error-kinds/#bad-class-definition)]
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxAAQDqMUAxrgLYzUAuunAFuwAqkATjDBRS1AMoYcBADS8IcakuqpqxVAHMYAHXQ0A7nxHVSuAK7VGGaoaEQOMBXB6pTM6rHfo1eCxycpMQQ6FoqvoIcPLjohPqUCdQAEjAeHtaWxLDUYDCoHBYiyrhgispoWHj4iEkAtNQA4jCBIqhQdRwQbEHE7Ix8jADWoeF5MDiow-XUAKpw7CIAbvlQAPocwTAAFACUnNyhcH2MgaF5QiKYvfAzMPh9DmzogYaOPLzsAJIAIgCiuXyhWKvAKdnYcAsxGIuCEgW2MEIWkIagCuGYLGyLWc1C0uE6%2BMwolCjggsQUMRWQgULUYuxmtAcTjUWGoQgskVI0VimSJ1EwEBEp3EETMliE1GwQlwhgWQn03RhcN6o30%2BgAApsQmFCJgClMoKg4HANkIMHAwLCWHt9IxDcbqABZXCYCywABCRpgtV81EI-rV6DtRuUztdsG2YbdME9C12Puo1HVweNbG5mH0iaJZUtuG2drg%2BzqAD5qAA5WLezOJ6xQOCESAMTCIFU6gCC6AkAF5y5XAynlJIWO0oFGI2OYPHqzUIhx9CA5CAyCIxKRCFwWFAKDRBCuRZVZPhMuhIFoigUyXEkpIYOweBwOMQ4IgAPQv5eicSEWFaF8wdAvpg6JwC%2BzAnhAZ5ml0sQvrksJqEsqDQKg2A5GBp7ntBvi4MQWH1voZDcugdRUnAl7UD2uggAAzIQACMABMVH6AA2mk0pCHAAC6%2BgcoqsJOJgdQCkKXQrBR1AAOQQegsIwJJ%2BiyRwdQiAAjhYgoTHUQwwKQdRTIw8DKD2kmGD4CnoCAAC%2Bi5TGJMAAGLQDAFAHtUJDkFZQA

_No response_

QEDady:
# Summary

Pyrefly incorrectly extracts dataclass fields from class method assignments (e.g., inside `@classmethod` or `__init_subclass__`). At runtime, these assignments are ignored by Python's `dataclasses.dataclass` mechanism.

This leads to a false positive `bad-class-definition` error when a subclass defines a field without a default value.

To fix this, I introduced the new variant `ClassMethod` to the `ClassFieldInitialization` enum to explicitly track fields initialized inside class methods. These fields are now filtered out during dataclass field extraction, which aligns Pyrefly's behavior with standard Python dataclass semantics.

Fixes #3259

# Test Plan

I added two new integration tests to `pyrefly/lib/test/dataclass_transform.rs`:
1. `test_class_method_field_ignored_by_dataclass`: Verifies that fields defined in regular class methods are ignored by dataclass synthesis.
2. `test_init_subclass_field_ignored_by_dataclass`: Verifies that fields defined in `__init_subclass__` are ignored by dataclass synthesis.

I ran `test.py` and it didn't have any failure.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -242,9 +242,22 @@ pub enum ClassFieldInitialization {
     /// If this is a dataclass field, DataclassFieldKeywords stores the field's
     /// dataclass flags (which are options that control how fields behave).
     ClassBody(Option<Box<DataclassFieldKeywords>>),
-    /// This field is initialized in a method. Note that this applies only if the field is not
-    /// declared anywhere else.
+    /// This field is initialized in an instance method (e.g. `self.x = 1`).
+    ///
+    /// Note that this applies only if the field is not declared anywhere else.
+    ///
+    /// At runtime, this creates an instance attribute. Therefore:
+    /// 1. It is not visible when accessed on the class object (e.g. `Class.x` yields a no-access error).
+    /// 2. It is ignored by dataclass field extraction (unless also declared in the class body).
     Method,
+    /// This field is initialized in a class method (e.g. `cls.x = 1` inside `@classmethod`).
+    ///
+    /// Note that this applies only if the field is not declared anywhere else.
+    ///
+    /// At runtime, this creates a class attribute. Therefore:
+    /// 1. It is visible when accessed on the class object (e.g. `Class.x` is valid).
+    /// 2. It is ignored by dataclass field extraction (unless also declared in the class body).
+    ClassMethod,
     /// The field is not initialized at the point where it is declared. This usually means that the
     /// field is instance-only and is declared but not initialized in the class body.
     Uninitialized,
@@ -259,6 +272,7 @@ impl Display for ClassFieldInitialization {
         match self {
             Self::ClassBody(_) => write!(f, "initialized on class body"),
             Self::Method => write!(f, "initialized in method"),
+            Self::ClassMethod => write!(f, "initialized in class method"),
             Self::Uninitialized => write!(f, "initialized on instances"),
             Self::Magic => {
                 write!(f, "not initialized on class body/method")
@@ -741,6 +755,7 @@ impl ClassField {
             ClassFieldInner::ClassAttribute { ty, .. } => match self.initialization() {
                 ClassFieldInitialization::ClassBody(_) => Some(ty),
                 ClassFieldInitialization::Method
+                | ClassFieldInitialization::ClassMethod
                 | ClassFieldInitialization::Uninitialized
                 | ClassFieldInitialization::Magic => None,
             },
@@ -846,6 +861,7 @@ impl ClassField {
             ClassFieldInner::ClassAttribute {
                 initialization:
                     ClassFieldInitialization::Method
+                    | ClassFieldInitialization::ClassMethod
                     | ClassFieldInitialization::Uninitialized
                     | ClassFieldInitialization::Magic,
                 ..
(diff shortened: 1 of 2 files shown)


NathanTempest:
Welcome to the project and thank you for your contribution @QEDady!

The fix correctly identifies the bug (Python's @dataclass ignores cls.x = ... assignments inside class methods, but pyrefly was treating them as fields). The 2 test cases are well-chosen.

Tagging @rechen to review this PR. Here are a some key changes to consider while reviewing this PR:

- is_assigned_in_method at class_field.rs (around line 1125) still uses matches!(_, Method) and isn't updated for the new ClassMethod variant. Worth checking if this getter should consider class methods as "assigned in a method" too.

- Literal promotion at line ~1782 currently only triggers for ClassFieldInitialization::Method. After this PR, fields like cls.x: Literal["foo"] = "foo" inside a classmethod won't get literal preservation. Likely a minor edge case but worth confirming.

- Design: adding a new variant means every match site has to handle it, and a few got missed. An alternative would be to check the method kind at the dataclass extraction site itself (a more localized change). Curious which approach you'd prefer.

QEDady:
Thanks for the feedback!

I have updated `is_initialized_in_method` to include ClassMethod as you suggested. Somehow, I missed that.

Regarding literal promotion, I think fields assigned in class methods should still be treated like class body assignments. Since they are class-level attributes (and often used for static metadata), changing this would diverge from how we treat class body fields, which only promote top-level literals. LMK if you prefer otherwise. 

For the design, I preferred adding the new ClassMethod variant because it seemed more semantically representative. It allows us to preserve the information about where the field was defined across the different analysis phases. Carrying this info in the enum is much cleaner than trying to reconstruct it later during dataclass extraction, and Rust's exhaustiveness checking ensures we don't accidentally miss handling it in any other critical places. However, I am also open to other ways. Let me know what you think!

Also, can we run the workflows? It says "This workflow requires approval from a maintainer. "

rchen152 (approved):
Looks good, thank you!

stroxler (approved):

PR #3275, opened 2026-04-30
Symlink pyrefly/README.md to pyrefly/pyrefly/README.md

rchen152:
Summary: We had a stub README.md in the inner pyrefly directory for pyproject.toml to pick up. Symlinking the LICENSE file (D98305784) seems to have worked, so let's try symlinking the real README.md as well.

Differential Revision: D103236976


kinto0 (approved):

PR #3278, opened 2026-04-30
Fix false positive unbound-name for walrus operator in while condition (#3278)

### LINKED ISSUE #3272: False positive `unbound-name` when using walrus operator
### Describe the Bug

```py
from typing import Callable, Any

class Cat:
    def equals(self, other: Any) -> bool:
        return False

def main(f: Callable[[], Cat]) -> None:
    while (a := f()).equals(1):
        break

    print(a)

main(lambda: Cat())
```

```console
All checks passed!
ERROR `a` may be uninitialized [unbound-name]
  --> t.py:11:11
   |
11 |     print(a)
   |           ^
   |
 INFO 1 error
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BhVFCipssADR0AgulIAddIoDGouHEGoGiRXT11MMMHRgBHAK7C4ACjgwoYKbgYALGJUQy5ASjoBaAD46PFwoHXR9SLpKGAZzSgiAMSsYRUVDYxpUVmswTyERMVgAbWKAXSkhBjLfQLoAOVx0GHDIgHcXaBg6a1Q6RABeOjBrb29CM0soGwBGb1aovWwY1ABrNIj9YkpWBl7vDayc0RpsTFR8rVGD9BAJEDIYsChSQgZaKAoAYjoABVIni86GgsHh8HRlE1IGx4loIE1CIofgBlGDdFwMBjEOCIAD0uMeRhehF4bFxMHQuMwuGUcFxkPQ0NhDHhlOGvDoqAAbtlROJugymZQ4U06LhiCymnBEegyK4mn4ue44Ky6EN5CAAMyEGYAJg1imK7molDgZUU5nQXB4fBgmD8mAgMWULKVaroAHIIGx0LwYB7FL6GH4YhYnXa-KsYKQ-KhlMp4Bohh62qgEgHbgBfe5x10wRJdCggnAEEjkTNAA

_No response_

kinto0:
Summary:

When a walrus operator (:=) assigns a variable in a `while` condition, pyrefly incorrectly reports the variable as possibly uninitialized after the loop. The `while` condition always executes at least once, so the variable is guaranteed to be defined.

The root cause is that `setup_loop` saves the current flow as the loop's base *before* the test expression is evaluated. Walrus-defined names are introduced into the in-loop flow but not the base, so `teardown_loop` merges base (no variable) with loop flow (has variable) and concludes it's possibly uninitialized.

The fix adds `propagate_new_flow_entries_to_loop_base`, analogous to the existing `propagate_new_flow_entries_to_fork_base` used for elif walrus bindings. After evaluating the while test, any newly introduced names are copied into the loop's base flow so they are treated as unconditionally defined after the loop.

fixes https://github.com/facebook/pyrefly/issues/3272

Reviewed By: stroxler

Differential Revision: D103302224

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -2136,6 +2136,28 @@ impl Scopes {
     /// Also updates base entries that are uninitialized (e.g. `x: int` with no
     /// assignment) when the branch has an initialized binding, so that the
     /// merge does not falsely report "may be uninitialized".
+    /// Propagate names introduced by walrus operators in a `while` condition
+    /// into the loop's base flow. The condition always executes at least once,
+    /// so these names are guaranteed to be defined after the loop.
+    pub fn propagate_new_flow_entries_to_loop_base(&mut self) {
+        let scope = self.current_mut();
+        let loop_ = scope
+            .loops
+            .last_mut()
+            .expect("propagate_new_flow_entries_to_loop_base called outside of a loop");
+        for (name, info) in scope.flow.info.iter() {
+            if let Some(existing) = loop_.base.info.get(name) {
+                if matches!(existing.initialized(), InitializedInFlow::No)
+                    && !matches!(info.initialized(), InitializedInFlow::No)
+                {
+                    loop_.base.info.insert(name.clone(), info.clone());
+                }
+            } else {
+                loop_.base.info.insert(name.clone(), info.clone());
+            }
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


migeed-z:
the fix is landing right now

PR #3282, opened 2026-05-01
Warn when class instance without __bool__ or __len__ is used as boolean condition

### LINKED ISSUE #868: [feature] warn on using classes that don't define `__bool__` in a conditional
### Describe the Bug

We currently support a number of redundant conditional checks: https://github.com/facebook/pyrefly/blob/c7e38ff18b7bd76eaf6f7b4d4bfe0d351025fbc6/pyrefly/lib/alt/expr.rs#L1780

One more we could add is flagging whenever we use a class that does not define `__bool__` in a conditional (if a class does not define `__bool__` it is always truthy)

We'll need to be careful about potentially causing noisy false positives if type stubs are incomplete and missing `__bool__` for a class that actually defines it.

### Sandbox Link

_No response_

_No response_

knQzx:
# Summary

closes #3282 closes #868

adds `ConditionRedundantReason::InstanceAlwaysTruthy` - fires when a class instance is used in a boolean context but the class defines neither `__bool__` nor `__len__` (so instances are unconditionally truthy)

```python
class NoBool:
    pass

x = NoBool()
if x:   # warning: NoBool defines neither __bool__ nor __len__, instances are always truthy
    pass
```

## skipped to keep noise down

- `object` itself
- protocols and ABCs (`extends_abc`) - concrete subclasses can define the dunders at runtime
- custom metaclasses inheriting from `ABCMeta` via MRO (covers HA's `ABCCachedProperties(CachedProperties, ABCMeta)`)
- classes from bundled stubs (typeshed stdlib + third-party stubs) - many stub-only classes have dynamic runtime behavior not in stubs (`datetime`, `asyncio.Future`, `Lock`, sqlalchemy `Session`, etc)
- dataclasses
- classes with `__bool__` / `__len__` / `__getattr__` / `__getattribute__` / `__get__` anywhere in MRO (descriptors, dynamic attribute interception)

## remaining warnings on home-assistant primer

23 warnings, three patterns:

- **dead defensive checks (~16)**: `device = random.choice(...)` returns non-optional T then code does `if device:`. dead by types, lint working as intended
- **subclass narrowing (~2)**: base declares `attr: T | None`, subclass redeclares `attr: T`. after redeclaration `if obj.attr:` is dead by types
- **lying types via `# type: ignore[assignment]` (~5)**: `platform: EntityPlatform = None  # type: ignore[assignment]`. attr annotated T but assigned None with the assignment error suppressed. `if self.platform:` is meaningful at runtime but dead by declared types. proper fix is `T | None`. not suppressed because doing so needs initializer tracking on `ClassField` to hide what are arguably real type-annotation bugs

none look like the "incomplete stubs" failure mode the issue author flagged

# Test Plan

tests in `flow_branching.rs` cover plain classes, inheritance, descriptors, `__getattr*__`, dataclasses, stub-only classes, protocols, ABCs, custom-metaclass-via-MRO

`cargo test -p pyrefly` clean

### DIFF
--- a/crates/pyrefly_python/src/dunder.rs
+++ b/crates/pyrefly_python/src/dunder.rs
@@ -159,6 +159,7 @@ pub const INIT_SUBCLASS: Name = Name::new_static("__init_subclass__");
 pub const INVERT: Name = Name::new_static("__invert__");
 pub const ITER: Name = Name::new_static("__iter__");
 pub const LE: Name = Name::new_static("__le__");
+pub const LEN: Name = Name::new_static("__len__");
 pub const LT: Name = Name::new_static("__lt__");
 pub const MATCH_ARGS: Name = Name::new_static("__match_args__");
 pub const NE: Name = Name::new_static("__ne__");
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -1197,6 +1197,29 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         false
     }
 
+    /// Returns `true` if `cls` or any of its non-object ancestors defines `__bool__`, `__len__`,
+    /// or `__getattr__` (which can intercept `__bool__` dynamically).
+    /// Used to detect class instances that are always truthy.
+    pub fn class_has_bool_or_len(&self, cls: &Class) -> bool {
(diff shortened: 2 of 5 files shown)


rchen152:
Sorry for the delay! At a glance, this looks pretty good to me, but I'll import it tomorrow and get a second review.

stroxler (approved):

PR #3287, opened 2026-05-01
Fix TSP extra IPC connection shutdown hang

heejaechang:
## Summary

This draft PR investigates a hang in Pyrefly's TSP multi-connection IPC support.

Current Pyrefly main can cause clients to hang when they open an extra IPC connection for read-only TSP requests and later close that extra connection. This affects clients using the advertised `typeServerMultiConnection` IPC capability.

## What is happening

When a client closes an extra TSP IPC connection, Pyrefly sends a close signal to the thread that owns that connection.

Today, that thread can notice the close signal but exit before fully receiving it. Because the close signal uses a bounded channel, the main thread can stay blocked while trying to send the close signal.

When that happens, Pyrefly never responds to the client's close request, so the client waits forever.

There are two other issues in the same area:

- the extra IPC reader can keep reading more messages while shutdown is trying to happen
- IPC writes are treated as successful once they are queued, even if the writer thread has not actually written them to the pipe yet

## What changed

- Receive the close signal before the extra connection thread exits.
- Stop the extra IPC reader from getting ahead of request processing.
- Make IPC sends wait until the writer thread actually writes the message or reports an error.
- Keep existing in-memory test connections working the same way as before.
- Send `snapshotChanged` notifications only on the main TSP connection.

## Why

Extra IPC connections are used for read-only TSP requests. They should be safe to open, use, and close.

This change makes close handling finish properly and makes IPC write errors visible to the caller.

## Validation

Built Pyrefly locally:

- `cargo build --release -p pyrefly`

Then tested these scenarios with the patched binary:

- open an extra IPC connection and run a read-only request
- run read-only requests after snapshot changes
- open and close extra IPC connections repeatedly
- handle concurrent open requests
- use multiple extra IPC connections from separate clients

All targeted scenarios passed with this patch.


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -502,13 +502,89 @@ pub trait TspInterface: Send + Sync + 'static {
...
+
+                match result_receiver.recv() {
+                    Ok(Ok(())) => Ok(()),
+                    Ok(Err(_)) | Err(_) => Err(crossbeam_channel::SendError(original_message)),
+                }
+            }
+        }
+    }
+
+    #[allow(clippy::result_large_err)]
```
kinto0: curious what this is from?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -502,13 +502,89 @@ pub trait TspInterface: Send + Sync + 'static {
...
+                        crossbeam_channel::SendTimeoutError::Timeout(original_message),
+                    ),
+                }
+            }
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct WriterMessage {
```
kinto0: nit: for consistency, should this be `MessageWriter`? was that already taken?

heejaechang: Fixed in 929ed3d4c. Renamed `WriterMessage` to `MessageWriter` for consistency.

pyrefly/lib/lsp/non_wasm/server.rs
heejaechang: Addressed in 929ed3d4c. This was from Clippy's `result_large_err`: `SendError<Message>` / `SendTimeoutError<Message>` carry the unsent `Message` by value, which is large. I changed the suppression to `#[expect(...)]` with a reason, and added the matching expectation on `send` after Clippy pointed out the same API-shape issue there too.

samwgoldman (approved):

PR #3290, opened 2026-05-02
Rename parameters across files in keyword arguments

asukaminato0721:
# Summary

Part of #2207.

This PR fixes parameter rename across files when the external use is a keyword argument. For example, renaming `param` in:

```python
# fileA.py
def test(param):
    return param
```

now also updates the caller-side keyword in another file:

```python
# fileB.py
from fileA import test

result = test(param="hello")
```

The implementation adds keyword-argument references to the external reverse-dependency reference path. It does not rename keyword textually: each matching `param=` candidate resolves its callee, refines that callee back to the actual parameter definition, and only edits the keyword when the refined parameter range matches the parameter being renamed.

## Relationship to branch `1583`

Branch `1583` overlaps with this PR. It targets the same cross-file parameter-keyword rename behavior and adds a similar LSP regression fixture. It is broader than this PR because it also carries fallback logic around external-definition range/line matching.

This PR is the narrower version: it only wires parameter keyword-argument references into the existing external reverse-dependency lookup. If `1583` lands first, this branch should be rebased or dropped rather than landed independently with duplicate coverage.

# Test Plan

`cargo test rename::test_rename`

`python3 test.py --no-test --no-conformance --no-jsonschema`

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -3542,6 +3542,7 @@ impl<'a> Transaction<'a> {
     fn local_references_from_external_definition(
         &self,
         handle: &Handle,
+        definition_metadata: &DefinitionMetadata,
         definition_range: TextRange,
         module: &Module,
     ) -> Option<Vec<TextRange>> {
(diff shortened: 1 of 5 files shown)


kinto0:
I think this is #1583 and I think you already have a PR open for this? #2484

asukaminato0721:
oh, this also handle the name in like comments, strings.

kinto0:
oops - missed this. would you mind making the issue title / summary more clear as to why these are both necessary and if they don't overlap?

kinto0:
so it looks like this branch will be rebased / dropped? can you combine these PRs or decide which one you want us to review?

asukaminato0721:
suppressed by #2484

PR #3296, opened 2026-05-03
fix match case with type check #3295

### LINKED ISSUE #3295: match case with type check
### Describe the Bug

```py
def foo(x: list[int]):
    match x:
        case 1:
            ...
        case 2:
            ...
```

maybe give some warning.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

The match-case binder now records a deferred reachability check for each typed case pattern, and the solver warns when applying that case’s narrowing makes the current subject type Never.

The warning uses the existing unreachable kind, e.g. Case pattern can never match subject of type `list[int]`.

also skipped top-like object subjects to avoid false positives for runtime-matchable values.

Fixes #3295

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -1932,6 +1932,71 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         );
     }
 
+    pub fn check_match_case_reachability(
+        &self,
+        subject_idx: &Idx<Key>,
+        narrowing_subject: &NarrowingSubject,
+        narrow_ops_for_case: &(Box<NarrowOp>, TextRange),
+        case_range: &TextRange,
+        errors: &ErrorCollector,
+    ) {
+        let (op, narrow_range) = narrow_ops_for_case;
+        if !Self::is_value_match_case_op(op) {
+            return;
+        }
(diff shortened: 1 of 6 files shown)


NathanTempest (changes requested):
Solid implementation by cleanly reusing the existing narrowing machinery via a deferred BindingExpect::MatchCaseReachability, and is appropriately conservative about when to fire (skips Any, Never, object, and non-value patterns). Reusing ErrorKind::Unreachable rather than adding a new error kind is the right call.

A few suggestions before landing:

1. Expand test coverage. Currently only one test case in pattern_match.rs. Worth adding:

- Or-pattern: case 1 | 2: against list[int] (verifies is_value_match_case_op Or-handling)
- Subject is a union including the literal type: x: int | list[int] with case 1: (should NOT warn)
- Any subject: x: Any with case 1: (should NOT warn — verifies the early exit guard)
- Facet chain: match obj.field: case 1: where field has incompatible type
- case None: against a non-Optional type (should warn)
- Negative case: confirm case SomeClass(): against an incompatible type does NOT trigger this check (not a value pattern)

2. Update existing placeholder test. pyrefly/lib/test/flow_branching.rs:478-488 currently has cases marked "should be impossible to match" without asserting errors. Those should now be updated to assert the new diagnostic.

cc @stroxler for the final call.

NathanTempest (approved):

stroxler (approved):

PR #3297, opened 2026-05-03
fix support check more enum exhaustive #3294

### LINKED ISSUE #3294: support check more enum exhaustive
### Describe the Bug

this works

```py
from typing import Literal
def foo(x: Literal[1,2,3]):
    match x:
        case 1:
            ...
        case 2:
            ...
```

but this not

```py
def foo(x: int):
    match x:
        case 1:
            ...
        case 2:
            ...
```

the same case: str, list, ...

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3294

# Test Plan


stroxler (changes requested):
Back to you more for discussion than over the code, which looks okay to me from a skim.

My question is: what's the intent here?

My understanding is that we had a very limited set of types for which we enabled the warning with the idea that in certain scenarios a user "probably" intended to match everything.

We also already have an explicit, spec-mandated way to explicitly mark a match as exhaustive with `assert_never`.

I would, I think, support a change where users can opt into *all* matches being checked for exhaustiveness; that makes sense to me as a warning or even a type error if a team wants that to be their style convention.

But this change feels like something awkwardly in the middle: why would we want to check exhaustiveness on a limited collection of types that have infinite cardinality, while *not* checking exhaustiveness for other types?

-----

I marked the issue as "needs discussion" for this reason, I don't think I'm clearly seeing the rationale for this change.

stroxler:
Great, this looks like a clean feature that makes a lot more sense to me. Thanks!

stroxler (approved):

PR #3304, opened 2026-05-04
Ignore attribute types of schema classes in `pyrefly report`

### LINKED ISSUE #3301: `pyrefly report`: ignore schema classes (`@dataclass`, `TypedDict`, `NamedTuple`, `Enum`, etc.)
These "schema classes" are, by definition, always have typed attributes. So counting these as typed will inflate the type coverage %.

This is something that typestats does, but as far as I can tell, `pyrefly report` doesn't yet.

jorenham:
# Summary

This ports the typestats behavior where it ignores attributes of schema types like `TypedDict` and `NamedTuple`, which are always annotated, so that they no longer inflate the type coverage%.

Fixes #3301

# Test Plan

Tests added

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -745,8 +745,8 @@ impl ReportArgs {
     }
 
     /// Returns the ClassBinding if the class owning this field is a schema class (dataclass, enum,
-    /// TypedDict, or NamedTuple). Fields of schema classes are IMPLICIT — they have
-    /// 0 typable slots because the class definition itself governs their types.
+    /// TypedDict, NamedTuple, pydantic model, or attrs class). Fields of schema classes are
+    /// IMPLICIT — they have 0 typable slots because the class definition itself governs their types.
     fn get_cls_binding_if_schema_class_field<'a>(
         bindings: &'a Bindings,
         answers: &Answers,
@@ -762,6 +762,8 @@ impl ReportArgs {
             || metadata.is_enum()
             || metadata.is_typed_dict()
             || metadata.named_tuple_metadata().is_some()
+            || metadata.is_pydantic_model()
+            || metadata.is_attrs_class()
         {
             Some(cls_binding)
         } else {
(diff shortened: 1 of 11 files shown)


pyrefly/lib/commands/report.rs
```diff
@@ -762,6 +762,8 @@ impl ReportArgs {
             || metadata.is_enum()
             || metadata.is_typed_dict()
             || metadata.named_tuple_metadata().is_some()
+            || metadata.is_pydantic_model()
```
migeed-z, line 765: Why is django not included in this check? we also support other frameworks such as factory_boy and marshmallow_schema. If we add any of those, we should add tests, if not, I would love to learn why not.

pyrefly/lib/commands/report.rs
```diff
@@ -2904,4 +2916,44 @@ def g(x: int) -> int:
...
+    #[test]
+    fn test_report_schema_classes_inherited() {
+        let report = build_module_report_for_test("schema_classes_inherited.py");
+        compare_snapshot("schema_classes_inherited.expected.json", &report);
+    }
+
+    /// Pydantic BaseModel fields are IMPLICIT (0 typable).
+    #[test]
+    fn test_report_schema_classes_pydantic() {
+        let path = std::env::var("PYDANTIC_TEST_PATH").expect("PYDANTIC_TEST_PATH must be set");
```
migeed-z, line 2945: could you use pydantic_env() instead?

pyrefly/lib/commands/report.rs
```diff
@@ -2904,4 +2916,44 @@ def g(x: int) -> int:
...
+        let env = TestEnv::new_with_site_package_paths(&[&path]);
+        let report = build_module_report_for_test_with_env("schema_classes_pydantic.py", env);
+        compare_snapshot("schema_classes_pydantic.expected.json", &report);
+    }
+
+    /// attrs @define and @attr.s(auto_attribs=True) fields are IMPLICIT (0 typable).
+    #[test]
+    fn test_report_schema_classes_attrs() {
+        let path = std::env::var("ATTRS_TEST_PATH").expect("ATTRS_TEST_PATH must be set");
```
migeed-z, line 2954: can we use attrs_env()?

migeed-z (changes requested):

pyrefly/lib/commands/report.rs
jorenham: Django's model fields don't need to be typed, so I'm not sure if we should also ignore them.
I'm not familiar with `factory_boy` or `marshmallow_schema`, but I could look into it.

jorenham: Thinking about Django's case a bit more, I don't think we should include it here. Because even though Pyrefly implements Django support, other type-checkers don't. So annotating the model fields has value in that sense.

jorenham: It looks like the fields in `factory_boy`, `marshmallow_schema`, and `glean` are also not required to be annotated, so I don't think we should special-case them here either.

pyrefly/lib/commands/report.rs
jorenham: Hmm I don't think I can, unfortunately. Because when I use `crate::test::pydantic::util::pydantic_env` it complains that

```
module `pydantic` is private
```

pyrefly/lib/commands/report.rs
jorenham: That one also lives in a private module, I'm afraid.

rchen152 (approved):

PR #3306, opened 2026-05-05
stubgen: emit synthesized `__init__` for dataclass / Pydantic models

tobyh-canva:
## Context

As in #3221 / #3223, stub output for dataclass and Pydantic models needs to reflect **how instances are constructed** (defaults, `kw_only`, validation aliases, etc.) without turning `.pyi` files into a dump of `field()` / `Field()` kwargs.

## Intent

**Alternative to #3223.** Rather than primarily re-emitting a **reduced** `field()` / `Field()` on the class line, this approach **injects a synthesized `__init__`** derived from the checker’s existing dataclass / Pydantic init plumbing, then **aligns annotations with the authored class body** where parameter names match (including alias-only init).

## Changes

- **`extract.rs`:** Resolve synthesized `__init__` from `ClassSynthesizedFields`, map `Param` → `StubParam`, merge **AST annotation text** for init params (skipping `InitVar`), drop `**kwargs` from emitted synthetic inits, preserve / improve class-body ordering and `field` / `Field` elision behavior, import pruning for unused Pydantic helpers.
- **`emit.rs`:** Multiline `def __init__` when the stub has many parameters.
- **`dataclass.rs`:** Validation-**alias-only** init parameters are **keyword-only** (`*, id: ...`), matching runtime and stub tests.
- **`pydantic.rs` / `class_field.rs`:** Treat `pydantic.BaseModel` re-exports like other BaseModel bases; expose `ClassField::dataclass_flags_of` to stubgen.
- **`mod.rs` tests:** Minimal on-disk `pydantic` shim + glob only `input.py` so temp package imports do not overwrite stubgen output.

## Testing

- `cargo test -p pyrefly stubgen::tests::`

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1115,7 +1115,7 @@ impl ClassField {
         }
     }
 
-    fn dataclass_flags_of(&self, heap: &TypeHeap) -> DataclassFieldKeywords {
+    pub(crate) fn dataclass_flags_of(&self, heap: &TypeHeap) -> DataclassFieldKeywords {
         match &self.0 {
             ClassFieldInner::Property { .. } => DataclassFieldKeywords::new(),
             // Class-body-initialized descriptors have a default value (the descriptor instance).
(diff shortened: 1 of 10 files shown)


tobyh-canva:
Closing in favour of #3586

PR #3315, opened 2026-05-06
[website] Center error line in sandbox when scrolling from results panel

### LINKED ISSUE #338: When clicking on an error in the sandbox output, scroll so that the affected line is in the middle of the screen
### Describe the Bug

When clicking on an error line that is not in the current pane, the sandbox currently scrolls so that the highlighted line is at the top or bottom of the screen (I assume based on which direction it has to scroll). It would be a nicer user experience to put it as close to the centre of the screen as possible (if it's at the beginning/end of the code you can't do much, of course),

### Sandbox Link

_No response_

_No response_

markselby9:
## Summary

Switches the SandboxResults error-click handler from `editor.revealRange(range)` to `editor.revealRangeInCenterIfOutsideViewport(range)` so the target line lands with surrounding context instead of flush against the viewport edge.

## Why this API, not `revealRangeInCenter`?

The issue title says "in the middle of the screen", but the body clarifies the actual complaint is the offscreen case — when the line being scrolled to ends up at the very top or bottom with no context around it. `revealRangeInCenter` would forcibly recenter on every click, including when the line is already comfortably visible, which is jarring and unexpected for a "go-to-error" action. `revealRangeInCenterIfOutsideViewport` only recenters when the line is offscreen and leaves it alone otherwise — matching standard editor "jump-to" behavior (VS Code, IntelliJ, etc.).

## Test plan

- [x] `tsc --noEmit` passes
- [x] Existing test suite passes
- [ ] Manual smoke test in the sandbox recommended for visual UX confirmation: trigger an error far offscreen (line should center), then trigger one already visible (no scroll should occur)

Closes #338

### DIFF
--- a/website/src/sandbox/Sandbox.tsx
+++ b/website/src/sandbox/Sandbox.tsx
@@ -878,7 +878,7 @@ export default function Sandbox({
             endColumn,
         };
 
-        editor.revealRange(range);
+        editor.revealRangeInCenterIfOutsideViewport(range);
         editor.setSelection(range);
     };
 


rchen152 (approved):

PR #3319, opened 2026-05-06
fix No narrowing for unions in match when the subject is a call #2858

### LINKED ISSUE #2858: No narrowing for unions in `match` when the subject is a call
### Describe the Bug

Repro:

```

@dataclass
class A: ...
@dataclass
class B: ...


def f(x: A | B) -> A | B:
  return x


def test(x: A | B):
    match f(x):
      case A(): ...
      case B(): ...
      case y: assert_never(y)
```

Output:

```
ERROR Argument `A | B` is not assignable to parameter `arg` with type `Never` in function `typing.assert_never` [bad-argument-type]
  --> /tmp/foo.py:18:28
   |
18 |       case y: assert_never(y)
   |                            ^
   |
```

Interestingly, using `x` instead of `f(x)` works.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2858

match subjects without a real narrowing name, like match f(x):, still carry direct fallthrough narrows across cases.
The final capture now forwards to the already-evaluated subject narrowed by previous cases, without synthesizing a user-visible name.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/binding/pattern.rs
+++ b/pyrefly/lib/binding/pattern.rs
@@ -19,6 +19,7 @@ use ruff_python_ast::Pattern;
 use ruff_python_ast::PatternKeyword;
 use ruff_python_ast::StmtMatch;
 use ruff_text_size::Ranged;
+use ruff_text_size::TextRange;
 
 use crate::binding::binding::Binding;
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #3327, opened 2026-05-06
support vscode-python-environments extension (#3327)

kinto0:
Summary:

# This stack
with the python-environments extension, there are now multiple places Pyrefly needs to look for interpreter info. This stack serves to support the new extension.

Steps:
- Abstract away API
- remove dependency on ms-python
- add support for python-environments

# This diff
add support for python-environments:
- access API, prefer to ms-python.python
- watch for extension changes in case someone installs another one

note that our "nothing is installed" warning still says ms-python.python and offers to install ms-python.python. I will keep this until ms-python.python is deprecated

Differential Revision: D104096556

### DIFF
--- a/lsp/package-lock.json
+++ b/lsp/package-lock.json
@@ -7,7 +7,7 @@
         "": {
             "name": "pyrefly",
             "version": "0.0.1",
-            "license": "Apache2",
+            "license": "MIT",
             "dependencies": {
                 "@vscode/python-extension": "^1.0.5",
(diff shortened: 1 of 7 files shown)


kinto0:
cc @eleanorjboyd, we are starting to use the new APIs in case they change. no extension dependency, contrary to the readme (because legacy users still use ms-python.python)

samwgoldman (approved):

PR #3333, opened 2026-05-07
Fix Pyrefly regression with imported TypeVars via attribute access

### LINKED ISSUE #3332: False postive invalid-type-var when importing type variables from another module in v0.64
### Describe the Bug

We have 
```
#pkg/lib.py
from typing import TypeVar
MyTypeT = TypeVar('MyTypeT')
```
Then, we have:
```
from typing import Generic, List, Protocol
from pkg import lib

class MyProto(Protocol[lib.MyTypeT]):  # E
  pass

class MyGeneric(Generic[lib.MyTypeT]): # E
  pass

def my_func(x: lib.MyTypeT) -> lib.MyTypeT: # 2 E
  return x


MyAlias = List[lib.MyTypeT]  # Fine
```

I get:

```
ERROR Expected a type variable, got `TypeVar[MyTypeT]` [invalid-type-var]
 --> b.py:4:7
  |
4 | class MyProto(Protocol[lib.MyTypeT]): 
  |       ^^^^^^^
  |
ERROR Expected a type variable, got `TypeVar[MyTypeT]` [invalid-type-var]
 --> b.py:7:7
  |
7 | class MyGeneric(Generic[lib.MyTypeT]): 
  |       ^^^^^^^^^
  |
ERROR Type variable `MyTypeT` is not in scope [invalid-type-var]
  --> b.py:10:16
   |
10 | def my_func(x: lib.MyTypeT) -> lib.MyTypeT:
   |                ^^^^^^^^^^^
   |
ERROR Type variable `MyTypeT` is not in scope [invalid-type-var]
  --> b.py:10:32
   |
10 | def my_func(x: lib.MyTypeT) -> lib.MyTypeT:
   |                                ^^^^^^^^^^^
   |
 INFO 4 errors
```
What is weird is this only happens when using `from pkg import lib`. It doesn't happen with `from pkg.lib import MyTypeT` or with ` import pck.lib.MyTypeT`.

This is a regression in v0.64. It worked in v0.63.1.

### Sandbox Link

_No response_

_No response_

QEDady:
# Summary

This PR fixes a regression introduced in v0.64 where Pyrefly reports `invalid-type-var` errors when a `TypeVar` is imported from another module using the `from pkg import lib` syntax and then accessed as an attribute (e.g., `lib.MyTypeT`).

The root cause was that `make_legacy_tparam` in `bindings.rs` was too strict for `LegacyTParamId::Attr` (handling dotted names like `lib.T`). It only allowed the base name to be bound to `Binding::Module`. However, when a module is imported via `from pkg import lib`, it results in `Binding::Import` rather than `Binding::Module`. This caused Pyrefly to fail to recognize the imported symbol as a potential legacy type variable holder.

This PR updates `make_legacy_tparam` to also allow `Binding::Import` for `LegacyTParamId::Attr`, allowing the solver to correctly resolve the type variable.

Fixes #3332

# Test Plan

Added a new unit test `test_class_generic_typevar_from_imported_module` to `pyrefly/lib/test/generic_legacy.rs` that sets up a package structure and verifies that `lib.T` is correctly recognized as a type variable. The test passes with this fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2867,6 +2867,7 @@ dependencies = [
 name = "rustversion"
 version = "1.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -2150,7 +2150,7 @@ impl<'a> BindingsBuilder<'a> {
                 Some(_) => None,
             },
             LegacyTParamId::Attr(_, attr) => match binding {
-                Some(Binding::Module(..)) | None => Some((
+                Some(Binding::Module(..) | Binding::Import(..)) | None => Some((
                     KeyLegacyTypeParam(ShortIdentifier::new(attr)),
                     BindingLegacyTypeParam::ModuleKeyed(original_idx, Box::new(attr.id.clone())),
                 )),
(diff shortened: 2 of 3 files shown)


samwgoldman (approved):

PR #3339, opened 2026-05-08
fix Ignored issubclass(categories, enum.Enum) check #3283

### LINKED ISSUE #3283: Ignored `issubclass(categories, enum.Enum)` check
### Describe the Bug

```py
from typing import Iterable
import enum

def main(categories: Iterable[str] | type[enum.Enum]) -> None:
    cached_categories: tuple[str, ...]
    if isinstance(categories, type) and issubclass(categories, enum.Enum):
        cached_categories = tuple(member.value for member in categories)
    else:
        cached_categories = tuple(categories)
```

```console
(scratch) mgorelli@marcoslaptop:~/scratch$ mypy t.py
All checks passed!
(scratch) mgorelli@marcoslaptop:~/scratch$ zuban check t.py
ERROR `tuple[Enum | str, ...]` is not assignable to variable `cached_categories` with type `tuple[str, ...]` [bad-assignment]
 --> t.py:9:29
  |
5 |     cached_categories: tuple[str, ...]
  |                        --------------- declared type
6 |     if isinstance(categories, type) and issubclass(categories, enum.Enum):
7 |         cached_categories = tuple(member.value for member in categories)
8 |     else:
9 |         cached_categories = tuple(categories)
  |                             ^^^^^^^^^^^^^^^^^
  |
 INFO 1 error
```

Spotted in Narwhals

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BJBjEqpssADrouPPnRjoArjSlTMMMHRqpWACgDGqYW14R4iQcNHiYAbTgNKAXToAfRszsLlhAKJKaJwBKOgBaAD46ADlcdBhEKTokukN9AAsYTAB9Q2NTc0ZFYlh7RwAaOkIqp0TkiE0IOFYHDH0YAyMYE0ozOAqmYhgQjExOODhFbH0oVHGOvJ74Cu8aPwCghPRk7ZTUdMyczu7eugBeQuL2mhgabBFCADdUKEUYOjBeLRu7yk4t3K6%2BTgQVqSRgUDg8VBO1SGWyAOO8DOF1g80Bi2BIDKIDIlA0UFIhAYtCgFAAxHQAAqkPFgAl0NBYPD4FKxSBsRSiBgQWKEKQUgDKMDeaQYDGIcEQAHopbj8YTeGwpQopZhcPo4FL9GyIByuTz0FL3p9UE9oGJYKz0OzOUYDXRcMRubE4Hz0GQGGlYqEHiImrFkRIQABmQgARgATEGpLYRNRKHAakppNxeMJMKFMBA8fpub7kQByXXoXgwAtSEsMUJ4gCOimzmVCAGsYKRQns2uNCwB3VCUdDl9AgAC%2B2L2eZgADFoDAKIycAQSORh0A

_No response_

asukaminato0721:
# Summary

Fixes #3283

isinstance(x, type) preserves known class-object types so negative issubclass narrowing distributes over union members instead of giving up when the union also contains non-class values.

# Test Plan

add test


NathanTempest (changes requested):
Thanks @asukaminato0721, the core idea is right and the improvements are real (bokeh, hydpy, static-frame, beartype all get FPs removed). But the latest primer run still has two regressions we need to clear before import:

- materialize type[Never] not iterable (mzcompose.py:172) this is the same isinstance(x, type) and issubclass(x, enum.Enum) + iterate-enum pattern your fix targets, so the positive branch is narrowing to type[Never] for some union shapes even though test_isinstance_type_and_issubclass_else passes. Could you add a regression test matching materialize's shape and fix the positive-branch narrowing? The negative-branch code returning Never (from subtract(Enum, Enum)) looks like it may be leaking into the positive branch as type[Never], but worth confirming.

- pydantic Expected class object, got Any (v1/fields.py:637)  isinstance(self.type_: Any, type) no longer yields something usable as a class object. The isinstance-type special-case only matches Type(_) | ClassDef(_), so Any falls through worth checking whether it should produce type[Any].


Once those clear on a fresh primer run I'll import and loop in @rechen / @yangdanny97 for the narrowing review.

PR #3353, opened 2026-05-09
fix False negative: Pyrefly does not emit an unpacking error in for k, v in dict #3352

### LINKED ISSUE #3352: False negative: Pyrefly does not emit an unpacking error in `for k, v in dict`
### Describe the Bug

```python
for k, v in {"x": 1}: # expect unpack error
    pass
```

this fails at runtime. mypy emits an error (str cannot be unpacked) but pyright does not

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4uATgAQDWANDQG40To3AA6I%2BPiNAIwBfAQGIaMfMRgBjAC40ArumKpZdSVSrUuHGgbVwEDEGSowwUUoXm4AtlAoSACqQtXSNNFjz4asrjokADmSlSo8hBBhHoSAMowMDQAFvLyxHCIAPTZ5pbWhNQh2TDo2Zi4snDZgcEQYRFRQdk0YNQ0qMyo0KjYsAFBoeGR0Ry4xM3ocLGqpPIpQQC0zDBUcGM0ALw0PADMhIIATDx6ANprOusAunoqEPbE1PIwmEuYEBYKEKvbNADkDXQ1Bg-z0wPkSwsAEclJ9Xks6DBSEt1LJ4HA-v8AO6oKjoMHoEDCUzqKKrABi0BgFB8OAIJHIwiAA

_No response_

asukaminato0721:
# Summary

Fixes #3352

unpack-length checking rejects string-like unpack sources (str, string literals, LiteralString) with bad-unpacking, which catches for k, v in {"x": 1} because dict iteration yields keys.

# Test Plan

update &&  add test


yangdanny97:
the added errors for apprise look fishy, both of them are typed as `list[tuple[str, str]]` which should be legal

https://github.com/caronc/apprise/blob/dc93c2a3747ce595032949db75e9dc48a05296d8/apprise/plugins/xmpp/base.py#L202

https://github.com/caronc/apprise/blob/dc93c2a3747ce595032949db75e9dc48a05296d8/apprise/plugins/xmpp/adapter.py#L232

asukaminato0721:
```
x: list[ (str, str) ]= []
```

```
Expected 1 type argument for `list`, got 2 [[bad-specialization](https://pyrefly.org/en/docs/error-kinds/#bad-specialization)]
```

asukaminato0721:
Looks like the problem with tuple handling.

PR #3359, opened 2026-05-11
Fix cross-file references lost when CRLF/LF line endings differ #3237 (#3359)

### LINKED ISSUE #3237: textDocument/references misses cross-file references when the target module starts with a docstring
### Describe the Bug

I tried to integrate pyrefly lsp into [serena](https://github.com/oraios/serena), but [one of the test case](https://github.com/oraios/serena/blob/v1.1.2/test/solidlsp/python/test_python_basic.py#L38) failed. After some digging, I found an issue with textDocument/references in pyrefly lsp.

Repro repo: https://github.com/0x1997/pyrefly-issue
pyrefly version: 0.62.0
OS: Windows 11 x64

### Steps

1. git clone https://github.com/0x1997/pyrefly-issue
2. cd pyrefly-issue
3. python repro.py

### Observed

- For class A (defined in a.py, file starts with a module docstring), references from services.py are not returned.
- For class B (defined in b.py, no module docstring), references from services.py are returned correctly.

Example output:

- A: services.py references=0
- B: services.py references=3

### Expected

Both A and B should return references from services.py.

### Sandbox Link

_No response_

_No response_

kinto0:
Summary:

When files on disk use CRLF but the editor sends LF-normalized content via did_open, byte offsets diverge between the on-disk exports and the in-memory definition. The exact byte-range comparison in local_references_from_external_definition fails for any definition after a line break, dropping cross-file references.

The fix adds a line-number fallback to the comparison. When the exact byte range doesn't match, it compares the line number of the export location against the definition's line number. Line numbers are invariant across CRLF/LF differences, so this correctly matches definitions even when byte offsets differ. The fallback preserves the exact-range fast path for the common case and only activates when content encoding differs.

fixes #3237

Reviewed By: yangdanny97, jvansch1

Differential Revision: D104303086

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -3172,6 +3172,9 @@ impl<'a> Transaction<'a> {
         .concat()
     }
 
+    /// Find references to an external definition within the given handle's module.
+    /// When the exact byte-range comparison fails (e.g. CRLF/LF differences),
+    /// falls back to comparing line numbers, which are encoding-invariant.
     fn local_references_from_external_definition(
         &self,
         handle: &Handle,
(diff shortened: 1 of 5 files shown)


yangdanny97 (approved):

PR #3371, opened 2026-05-13
fix add quickfix for Unnecessary str() call; #3342

### LINKED ISSUE #3342: add quickfix for Unnecessary `str()` call;
### Describe the Bug

Image

now only ignore fix

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3342

Added a local quick fix for UnnecessaryTypeConversion diagnostics so str(x) on an already-str value now offers Remove unnecessary str() call and rewrites it to x.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2674,6 +2674,20 @@ impl<'a> Transaction<'a> {
                         }
                     }
                 }
+                ErrorKind::UnnecessaryTypeConversion => {
+                    if let Some(action) =
+                        quick_fixes::unnecessary_type_conversion::unnecessary_type_conversion_code_action(
+                            &module_info,
+                            &ast,
+                            error_range,
+                        )
+                    {
+                        let call_range = action.2;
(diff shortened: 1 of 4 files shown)


NathanTempest (approved):
Thanks @asukaminato0721 for implementing this quick fix.

The implementation looks solid:
- Correctly locates the Call node from the error range using AST traversal
- Handles both simple names (`str`) and attributes (`builtins.str`)
- Properly validates the call structure (single arg, no kwargs, no starred exprs)
- Uses `wrap_if_needed` for correct parenthesization when replacing
- Good test coverage with the golden test showing before/after

I verified the implementation is generic and should work for all builtin type 
constructors (`int()`, `bool()`, `float()`, `bytes()`) that trigger 
UnnecessaryTypeConversion, not just `str()`. The dynamic extraction of 
`conversion_name` and the generic replacement logic handle all cases.

This will be a nice quality-of-life improvement for users. Approved and will 
import!

kinto0 (approved):

PR #3386, opened 2026-05-14
fix Missing Cargo.lock checksum for rustversion

### LINKED ISSUE #3383: Cargo.lock: Missing checksum for `rustversion`
### Describe the Bug

The dependency was last updated on https://github.com/facebook/pyrefly/commit/0a39dbd0af07505de654c7f14cf802d7058612dc, but it seems that the checksum line has vanished since.

### Sandbox Link

_No response_

_No response_

EpicEric:
# Summary

With an incomplete Cargo.lock, the project fails to build or be vendored with `--locked`.

Fixes #3383 

# Test Plan

Running `cargo check` or most other Cargo commands fixes the Cargo.lock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2866,6 +2866,7 @@ dependencies = [
 name = "rustversion"
 version = "1.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"


samwgoldman:
Thanks for your contribution! I need to investigate this a bit more. Our Cargo.lock file is generated from our internal build infra, where we use a patched version of rustversion. That's why we don't have a checksum.

I'm going to try to see if we can fix our internal infra here, since we definitely want to be able to build with `--locked`, per your original issue. Stay tuned!

grievejia (approved):

PR #3391, opened 2026-05-14
fix false positive in forward type reference #3387

### LINKED ISSUE #3387: false positive in forward type reference
### Describe the Bug

This should not give an error

```
class Foo[T]: ...
class Bar:
    foo = Foo[Bar]()
```

reported in https://www.reddit.com/r/Python/comments/1tbyd7m/comment/olot8wx/

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3387

`Foo[Bar]` under `from __future__ import annotations` binds Bar as static type information when Foo is a known class, avoiding the false Bar is uninitialized diagnostic.

# Test Plan

add test


rchen152 (approved):

PR #3415, opened 2026-05-15
Fix negative narrowing for isinstance tuple targets

### LINKED ISSUE #3412: `if isinstance(val, (str, bytes))` ignored
### Describe the Bug

If I run this:

```py
from typing import reveal_type, Any, Iterable


def arrlen(obj: Any) -> int | None:
    ...

def main(a: float | str | bytes | Iterable[Any]) -> None:
    if isinstance(a, (str, bytes)):
        pass
    elif arrlen(a) is not None:
        reveal_type(a)
```

I get

```console
/home/mgorelli/scratch/t.py
  /home/mgorelli/scratch/t.py:11:21 - information: Type of "a" is "float | int | Iterable[Any]"
0 errors, 0 warnings, 1 information
info[revealed-type]: Revealed type
  --> t.py:11:21
   |
11 |         reveal_type(a)
   |                     ^ `int | float | (Iterable[Any] & ~str & ~bytes)`
   |

Found 1 diagnostic
 INFO revealed type: Iterable[Any] | bytes | float | str [reveal-type]
  --> t.py:11:20
   |
11 |         reveal_type(a)
   |                    ---
   |
 INFO 0 errors
```

Note how pyrefly includes `str` and `bytes` - shouldn't it rule them out, based on the first `if` statement?

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAGjoBBdKTkBJBjEqpssADro9ezDDB1UlSrHQAKXNgBWiBUoCUdALQA%2BTun4AfOgByuOgwiHp0EXSE0QZYxnQ0qKxWqI5gULiofnRwDJR0-tik6nAFdGoaWrAA2oqkALqunoHBoeGRECYQcKy5GADGMClyVrmUckUlzs5h6JHzdMSocHDtETBQnabmlimu3XTouPxBIbMLkYIiYpLMQ6jOIDIgZILppIQMtFAUAMR0AAVSG8oKQchgcAQ6P1gpA2ABXTQMCDBQh6f4AZRgMDoAAsGAxiHBEAB6EmvYygwi8NgkmDoEmYXD9OAkmHoOGIrIohl0MC8UxCJJQKo49mcpE8ui4YjI4JwNHoMgMXHBNwiSg9YJ0AC8dB0IAAzIQAIwAJgNemqGmomvqenh6C4PD4MEwbkwEEE-WRIl1dAA5BA2EdBAG9EcGG5BABHeFet1uADWMFIblQ-UGK39AYA7mZ0OH0CAAL7PDO%2BmAAMWgMAoaCweCIZFLQA

_No response_

LeSingh1:
Fixes #3412.

This fixes negative narrowing for `isinstance` checks where the target is a tuple of runtime classes, such as `isinstance(x, (str, bytes))`.

## Summary

- For `isinstance(x, (A, B))`, `narrow_is_not_instance` previously subtracted each tuple element from the full input type independently and then intersected the per-element results.
- When the input contained a member that overlapped with multiple tuple elements (e.g. `Iterable[Any]` is a supertype of both `str` and `bytes`), the intersection step would reintroduce the subtracted alternatives through subset-based intersection. Concretely, `intersect_impl(Iterable[Any], str)` returns `str` because `str <: Iterable[Any]`.
- The fix walks the input union once and subtracts every tuple element from each union member sequentially, so the per-element scope doesn't reintroduce the overlap. Targets that don't allow negative narrowing (e.g. a non-final `type[C]` variable) are skipped, preserving existing conservative behavior.

## Repro

```py
from typing import reveal_type, Any, Iterable

def arrlen(obj: Any) -> int | None: ...

def main(a: float | str | bytes | Iterable[Any]) -> None:
    if isinstance(a, (str, bytes)):
        pass
    elif arrlen(a) is not None:
        reveal_type(a)
```

Before: `Iterable[Any] | bytes | float | str`
After:  `Iterable[Any] | float`

## Test plan

- [x] Added `test_isinstance_tuple_negative_with_overlap` in `pyrefly/lib/test/narrow.rs`.
- [x] `cargo test -p pyrefly --lib` — 5152 passed, 0 failed.
- [x] All 40 `test_isinstance_*` tests continue to pass, including the existing `test_isinstance_type_negative_partial_narrow` which exercises the not-final `type[C]` skip path.
- [x] Verified the repro file against the locally built `target/debug/pyrefly`.

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -432,39 +432,44 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         errors: &ErrorCollector,
     ) -> Type {
         let force_allow_negative = matches!(source, NarrowSource::Pattern);
-        let mut res = Vec::new();
-        for (right, allows_negative_narrow) in self.expr_as_class_info(right_expr, errors) {
-            res.push(self.distribute_over_union(left, |l| {
-                let allows_negative_narrow = allows_negative_narrow || force_allow_negative;
+        let class_infos = self.expr_as_class_info(right_expr, errors);
+        // Subtract each class info from each union member of `left` in turn,
+        // rather than computing per-target results and intersecting them.
+        // Intersecting `subtract(L, A)` with `subtract(L, B)` reintroduces
+        // any union member that overlaps with both A and B (e.g. `Iterable[Any]`
+        // overlaps with both `str` and `bytes`), so `isinstance(x, (str, bytes))`
+        // would leave the negative branch unchanged. Sequential per-element
+        // subtraction avoids that.
+        self.distribute_over_union(left, |l| {
+            let mut result = l.clone();
+            for (right, allows_negative_narrow) in &class_infos {
+                let allows_negative_narrow = *allows_negative_narrow || force_allow_negative;
                 if !allows_negative_narrow {
-                    return l.clone();
+                    continue;
                 }
-                if let Some((tparams, right)) = self.unwrap_isinstance_target(l, &right) {
+                if let Some((tparams, right)) = self.unwrap_isinstance_target(&result, right) {
                     let (vs, right) = self
                         .solver()
                         .fresh_quantified(&tparams, right, self.uniques);
                     // For constrained TypeVars, subtract from the concrete constraints
                     // so that e.g. isinstance(x, (int, float)) with T(int, str, float)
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #3417, opened 2026-05-16
Fix annotation only descriptors not having descriptor semantics

### LINKED ISSUE #3405: False Positive Asymmetric descriptors do not seem to work
### Describe the Bug

Code example from the Sandbox:
```py
from typing import Any, Generic, TypeVar

T = TypeVar( "T" )
U = TypeVar( "U" )

class InstantiatingAttr( Generic[ T, U ] ):
   def __set__( self, instance: Any, value: U | None ) -> None:
      ...

   def __get__( self, instance: Any, owner: Any = None ) -> T | None:
      ...

class Pistol:
  barrelLength: int
  def __init__( self, barrelLength: int, / ) -> None:
    ...

class Cowboy:
   holster: InstantiatingAttr[ Pistol, tuple[ int ] ]


c = Cowboy()
c.holster = ( 6, )
assert c.holster is not None
print( c.holster.barrelLength )
```

There are actually a couple of complex things going on here, but it seems that Pyrefly is unable to determine either the `__get__` or `__set__` type of `c.holster`.  It produces a false positive error on the line `c.holster = ( 6, )` and `print( c.holster.barrelLength )`. While this example might seem contrived, it is an extremely common pattern in the Arista codebase, with over 10,000 individual occurrences.

For what it is worth, this bug report is almost copied verbatim from [Pyright#11233](https://github.com/microsoft/pyright/issues/11233). Pyright can infer the values for `__get__` and `__set__`, but it has a bug around type narrowing when a set happens before a get in a particular scope.

CC: @stroxler 

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BBdKQA0dAOIx0MShADGYgCrMYANVSUAOum2K6AXjrLiajQAo6mkIqt0AlNoCqBoyvWULVx7Yc70cqFQ4ODoASXQ4BgwGCFQY9gEGBg8JKRl5AG0jMWcAXXtEbTpizBgwOgB9CrgYBiqLGqgwMVZIjDkYREFhMQA3VCgAV066ZwAfOgA5XGl7OgBaAD4pmc6i4o3CLe11ulLyqrZa%2BrpG5s4IqP8RoVE6XAB3aUou25dp2bsF5b0Jj7X0BtNts-AEgiEAAoQSK4KCFQF0bAaSgwKAAGSkbAYAAsuqwGOt9pUKqwIHUKg1UeckZQUejMTi8egGGIAPRzJYraTwoFbQg7fyBYJ0ADCjzwpB5xWxsMiMi64TazNi8TYiWSWShMKgYgYg2IsCy%2BLo%2BVyAu0chcYoeErMvjkhBlUDllBcFgAbGJfOCZPwHU6XZwQuhcPx-tpiLJmRZ-bKGDJCDS6Rj2Dj7NoQCIQGQUWAoKRCAxaFAKABiOgQ0i5-OnDA4Ah0OQzSBsQaUOIQGb89DlgDKMBgdGxSWIcEQrNZObK%2BcIvDYrKkrMwuDkcFZTfQLbbHZm7LAvDoqH60FQ2FgjebEFb7ZiM3uxFvEW7ZBxM3mvRkcE7gMMVgAzIQACMABMVgRiiNQMPoViRLIcgEiA2gZDI1CUHAZroIM6BcDwfAwJg8yYBAKLwRAH4uAA5FeIYohR2ghgw8wogAjoMxH4fMADWMCkPMqByB0wqGBRDwaOgdHoCAAC%2BWb8TEH4AGLQDAFBoFgeBEGQ0lAA

_No response_

mfish33:
## Summary
Fixes #3405 — annotation-only fields with descriptor types were not recognized as descriptors, causing false positives for asymmetric descriptors.

## Description
Annotation-only fields with descriptor types (classes defining `__get__` and/or `__set__`) were not recognized as descriptors because descriptor detection only checked `ClassBody` and `Magic` initializations. This caused false positives for any descriptors created on the class body. With this change, we were able to remove the special case of `sqlalchemy.orm.Mapped`.

**Notable Changes**
- Add `Uninitialized` to the set of initializations that trigger descriptor detection, so annotation-only fields follow the descriptor protocol.
- Store `initialization` in the `Descriptor` struct so downstream logic (dataclass defaults, non-data-descriptor checks) can distinguish class-body descriptors from annotation-only ones.
- Fix conformance bug in dataclass generic descriptor test.
- In `__init__`/`__new__`, accept both the raw descriptor type and the `__set__` value type, since `__init__` may be initializing the descriptor object itself. This choice is not entirely sound, as there is no easy way to check whether the descriptor class has already been initialized. This is because in many libraries, descriptors may be reflected and initialized via metaclasses with the `__attributes__` dunder.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2859,6 +2859,7 @@ dependencies = [
 name = "rustversion"
 version = "1.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -149,17 +149,10 @@ impl ClassAttribute {
         match self {
             Self::ReadWrite(ty) => Self::ReadOnly(ty, reason),
             Self::Property(getter, _, cls) => Self::Property(getter, None, cls),
-            Self::Descriptor(
+            Self::Descriptor(descriptor, base) => Self::Descriptor(
                 Descriptor {
-                    range, cls, getter, ..
-                },
-                base,
-            ) => Self::Descriptor(
-                Descriptor {
-                    range,
-                    cls,
-                    getter,
                     setter: false,
+                    ..descriptor
                 },
                 base,
             ),
@@ -223,6 +216,10 @@ pub struct Descriptor {
     getter: bool,
     /// Does `__set__` exists on the descriptor? Similar considerations to `getter` apply.
     setter: bool,
+    /// How the descriptor field was initialized. Used to distinguish class-body
+    /// descriptors (which have an actual object on the class) from annotation-only
+    /// descriptors (which rely on metaclass or other runtime machinery).
+    initialization: ClassFieldInitialization,
 }
 
 #[derive(Clone, Debug)]
@@ -516,7 +513,7 @@ impl ClassField {
     fn initialization(&self) -> ClassFieldInitialization {
         match &self.0 {
             ClassFieldInner::Property { .. } => ClassFieldInitialization::ClassBody(None),
-            ClassFieldInner::Descriptor { .. } => ClassFieldInitialization::ClassBody(None),
+            ClassFieldInner::Descriptor { descriptor, .. } => descriptor.initialization.clone(),
             ClassFieldInner::Method { .. } => ClassFieldInitialization::ClassBody(None),
             ClassFieldInner::NestedClass { .. } => ClassFieldInitialization::ClassBody(None),
             ClassFieldInner::ClassAttribute { initialization, .. } => initialization.clone(),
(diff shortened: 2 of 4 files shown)


stroxler:
Thanks for the PR! As discussed on the issue I think this is debatable, but let's see what the mypy primer runs tell us. I think there's a good chance they will show things are better in the ecosystem.

mfish33:
PR is now up to date with the conversation on the issue and is ready for review

stroxler:
Build errors are almost certainly related to the move to stable over the weekend (which I believe was something you said would help you). I'm going to try to patch it up internally, I suspect it's not substantial

stroxler (approved):
I think we should move forward with this. I imported it and it builds cleanly, I think the auto-rebase eliminated issues we're seeing in github CI so don't worry about that

Will merge once I get a second review internally. Thanks again!

grievejia (approved):

PR #3426, opened 2026-05-16
Fix false positives when assigning super().method to self.attribute (#3414)

### LINKED ISSUE #3414: `missing-argument` on `self`, `bad-override`, `bad-arguments` when overwriting class method with other classmethod
### Describe the Bug

Overwriting a method on a class with give a false positive `missing-argument` on `self` even in cases where the function correctly has `self`. 
```python
class Parent:
    def meth(self): ...


class BadChild(Parent):
    def ___init__(self) -> None:
        self.meth = super().meth


class GoodChild(Parent):
    def meth(self):
        return super().meth()

if __name__ == "__main__":
    Parent().meth()
    BadChild().meth()
    GoodChild().meth()
```
gives

```
ERROR Class member `Child.meth` overrides parent class `Parent` in an inconsistent manner [bad-override]
 --> tmp.py:7:14
  |
7 |         self.meth = super().meth
  |              ^^^^
  |
  `Child.meth` has type `(self: Child) -> None`, which is not assignable to `(self: Child) -> None`, the type of `Parent.meth`
ERROR `(self: Self@Child) -> None` is not assignable to attribute `meth` with type `(self: Self@Child) -> None` [bad-assignment]
 --> tmp.py:7:21
  |
7 |         self.meth = super().meth
  |                     ^^^^^^^^^^^^
  |
ERROR Missing argument `self` in function `Parent.meth` [missing-argument]
  --> tmp.py:17:17
   |
17 |     Child().meth()
   |                 ^^
   |
```
pyright does not complain and it works fine at runtime. It seems related to https://github.com/facebook/pyrefly/issues/1448.
I'm not claiming this is a particular good pattern but it came up when some code wrapped parent methods inside `__init__`.

It think the main issue is the `bad-assignment` error saying the two identical types are not assignable and the `missing-argument` which implies a runtime error that does not happen. I'm not sure if I think it should complain about the override/assignment but I would expect it not to give an error on `missing-argument`

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQAKqATjOgC6IA669-9TDDD0AtjA4ALABRwYUMAEpE9Qmt4b01WgwBCqTAGFJ0TNOZtOy3gMHD6AfScR0EDk9nyl9ALQA%2BegA5XHQYHj5bATkFQnEpegBeejgAV2IYFmlFWIlJTV5tOnoAcVxcIxMoMwt2DmsI-iEROJlopXDIgTYOFJY%2BVPTM7Jas-NcRJ3RUcSdEpO4QJ1FUFycFjoEaziycqVGG%2Bn0K0x2RxRsBUvLjE%2BHcrJAAGhAyNjAoUkIOXFEoCgBiJikN4fZIYHAEeiUEKQADmvVQHAgIUIvEBAGUYDB6JIOBxiHBEAB6ImvYQfQi4FiwonsImYXCUOBE6HoOEIpEhIn0MBU%2BioABuKxo2FgUJhEHhLERyL4uGInPQcFR6DIUhCPgFGTgssS9AWAGZCABGABMC14AG0MiwqXAALq8FKuUTEKkcGCYHyYCBsShIrV6gDkkvQVJgQd4YY4PjYAEcUr7PT4ANYwUg%2BVCUSjwBhJIMAd1Y6Ej6BAAF9nlmAzAAGLQGAUNBYPBEMgVoA

_No response_

lordhaa123:
## Summary

Fixes https://github.com/facebook/pyrefly/issues/3414.

`self.attr = X` inside a method, when a parent class already has `attr`, was unconditionally substituting the inherited unbound type for the RHS's inferred type. For `self.meth = super().meth` this discarded the `BoundMethod` wrapper, leaving the field stored as a raw `Function(self) -> None` and producing three false positives:

- `bad-override` on the assignment target
- `bad-assignment` on the RHS
- `missing-argument self` at the call site

In `analyze_class_field_value`, the `inferred_from_method` branch now infers the RHS first and keeps it whenever it produced a `BoundMethod`, falling back to the inherited type otherwise. The existing stability behavior for plain attribute assignments (e.g. `self.x = 5` against an inherited `x: int`) is unchanged because those RHS types are not `BoundMethod`s.

This works because a `BoundMethod` already encodes the receiver, so its call signature is the parent method minus `self` — exactly what the runtime sees, and what makes the assignment, override, and call all type-check.

## Test plan

- [x] New regression test `test_super_method_assigned_to_self_attribute` in `pyrefly/lib/test/class_super.rs`
- [x] Original repro now reports 0 errors
- [x] `cargo test --lib -p pyrefly class_super class_overrides class_subtyping attributes` all pass
- [x] `python3 test.py --no-test --no-conformance --no-jsonschema` passes
- [x] Full lib suite: the 9 unrelated `config_finder` / `tsp_interaction` failures reproduce on baseline `main` and are not caused by this change

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2429,7 +2429,20 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     Some(inherited_ty) if inferred_from_method => {
                         // Inherit the previous type of the attribute if the only declaration-like
                         // thing the current class does is assign to the attribute in a method.
-                        inherited_ty
+                        //
+                        // Exception: if the assigned value is a `BoundMethod` (e.g.
+                        // `self.meth = super().meth`), the bound form is callable *without*
+                        // `self`. Using the inherited unbound function type here would make
+                        // the field appear to require `self` at call sites, and would also
+                        // make the assignment itself look type-incompatible (bound vs
+                        // unbound). Preserve the bound-method type instead.
+                        let swallowed = self.error_swallower();
+                        let inferred = self.attribute_expr_infer(e, None, name, &swallowed);
+                        if matches!(&inferred, Type::BoundMethod(_)) {
+                            inferred
+                        } else {
+                            inherited_ty
+                        }
                     }
                     Some(inherited_ty)
                         if inherited_annotation.is_none() && direct_annotation.is_none() =>
--- a/pyrefly/lib/test/class_super.rs
+++ b/pyrefly/lib/test/class_super.rs
@@ -144,6 +144,21 @@ class Triangle(Shape):
 );
 
 testcase!(
+    test_super_method_assigned_to_self_attribute,
+    r#"
+class Parent:
+    def meth(self) -> None: ...
+
+class Child(Parent):
+    def __init__(self) -> None:
+        self.meth = super().meth
+
+Parent().meth()
+Child().meth()
+"#,
+);
+
+testcase!(
     test_illegal_location,
     r#"
 class A:


lordhaa123:
i have done the CLA , check now

stroxler (approved):

PR #3438, opened 2026-05-17
Fix TypeVar resolution through arbatrary multi-level module attribute access (e.g. `Foo.Bar.ImportedT`). Closes #3431

### LINKED ISSUE #3431: Multiple Module TypeVar redirection not recognized
### Describe the Bug

Foo/__init__.py: empty
Foo/Bar.py
```python
from typing import TypeVar

ImportedT = TypeVar( "ImportedT" )
```

main.py:
```python
import Foo.Bar

def myFunc( t: Foo.Bar.ImportedT ) -> Foo.Bar.ImportedT:
    return t
```

It seems like the TypeVar in this case is not resolved through double module-level indirection.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIEAtsbgE4AuABAGK66EBCqdAOuv0wwwTKqRYBXdAGMAFEwaJW7Lj0IBJGvQYxMAFSYBKJgFoAfMo7c6GrY117E-Ji6Z0YDCXXQL%2BIADQgZO5gUKSEDLhUUBQAxEwACqQhYUxoWHj4TNK46JAA5l6oDBC5hPzxAMowMEwAFgwMxHCIAPStwcJhhPT5rTDorZi40nCtOXkQhXTFpYNMYPRMqABuqNCo2LDZuQVFJblMuMQH6HDl6GQMdbkmKzB0cHNMALxMvCAAzIQAjABMH34xHccA8Lw%2BcAYdAg0gYgPQAG0HnR6HAALr8KTUWj2TAmTAQdywiD3V5MADkU3Q9Bg5P41IYJncAEcJITdCYANYwUgmVDSaTwOBk8kAdx46Dp6ACIDYuFaAH0FRB0BAGEqSORECAZXLWtZNRQwCiqApSMQVfkmNjtEw9OaYAA1Hj8fiaHE6fRk%2B3EJ08eQfd3aBwfIx%2BAC%2BgX5JXuLGgMAo6RwBEN4aAA

_No response_

mfish33:
Previously, legacy TypeVar interception only handled single-level dotted access like `mod.T`. When a TypeVar was accessed through deeper module chains (e.g. `Foo.Bar.ImportedT`), pyrefly reported "Type variable is not in scope" because the binding phase only matched `Expr::Attribute` where the value was directly an `Expr::Name`.

**Notable Changes**

**`binding/expr.rs`** — Added `chase_static_attr_chain`, a recursive helper that walks nested `Expr::Attribute` nodes to collect the full dotted chain (base name + attributes).

**`binding/bindings.rs`** — `LegacyTParamId::Attr` changed from `(Identifier, Identifier)` to `(Identifier, Vec1<Identifier>)` to support arbitrary-depth attribute chains.

**`binding/binding.rs`** — `BindingLegacyTypeParam::ModuleKeyed` changed from `(Idx<Key>, Box<Name>)` to `(Idx<Key>, Vec1<Name>)` to store the full attribute chain.

## Testing
Added two tests, one for classes and another for functions.

## Tradeoffs
The current implementation uses `Vec1<T>` for readability and to enforce the non-empty invariant at the type level. As such, this change increases the memory size of bound legacy type vars from 16 bytes to 32. An alternative would be to pack everything into a single `Box<[T]>` (16 bytes), with the first element as the base and the rest as attributes, saving 16 bytes per instance. I opted for clarity over compactness here, but let me know if you would like the more performant option.

### DIFF
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -93,11 +93,9 @@ fn chase_static_attr_chain_impl(
     }
 
     match expr {
-        Expr::Attribute(ExprAttribute {
-            value: box Expr::Name(name),
-            attr,
-            ..
-        }) => Some((name.clone(), vec1![attr.clone()])),
+        Expr::Attribute(ExprAttribute { value, attr, .. }) if let Expr::Name(name) = &**value => {
+            Some((name.clone(), vec1![attr.clone()]))
+        }
         Expr::Attribute(ExprAttribute { value, attr, .. }) => {
             chase_static_attr_chain_impl(value, gas).map(|(name, mut attrs)| {
                 attrs.push(attr.clone());
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -9,6 +9,7 @@ use pyrefly_graph::index::Idx;
 use pyrefly_python::ast::Ast;
 use pyrefly_python::module_path::ModuleStyle;
 use pyrefly_python::short_identifier::ShortIdentifier;
+use pyrefly_util::gas::Gas;
 use pyrefly_util::visit::VisitMut;
 use ruff_python_ast::AtomicNodeIndex;
 use ruff_python_ast::BoolOp;
@@ -33,6 +34,8 @@ use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
 use starlark_map::Hashed;
+use vec1::Vec1;
+use vec1::vec1;
 
 use crate::binding::binding::Binding;
 use crate::binding::binding::BindingDecorator;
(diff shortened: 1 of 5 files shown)


pyrefly/lib/binding/expr.rs
```diff
@@ -71,6 +74,40 @@ fn is_special_name(name: &str) -> bool {
...
+    expr: &Expr,
+    gas: &mut Gas,
+) -> Option<(ExprName, Vec1<Identifier>)> {
+    if gas.stop() {
+        return None;
+    }
+
+    match expr {
+        Expr::Attribute(ExprAttribute {
+            value: box Expr::Name(name),
```
rchen152: This `box` needs to be removed.

rchen152 (changes requested):
Thanks! The code looks pretty good to me, but it looks like there's a compile error due to a change that just landed to make pyrefly use stable rust, which means we can no longer use the experimental `box` feature. Could you get rid of the `box` in `expr.rs`?

pyrefly/lib/binding/expr.rs
mfish33: Fixed

rchen152 (approved):

samwgoldman:
Thanks, Max!

PR #3452, opened 2026-05-18
Fix up the CI

ndmitchell:
Summary: Change required after our move to stable and deleting the stable toolchain file.

Differential Revision: D105551908


fangyi-zhou (approved):

PR #3459, opened 2026-05-18
fix option to ban explicit Any #3442

### LINKED ISSUE #3442: option to ban explicit `Any`
it would be nice if there was an option for pyrefly to report errors here:

```py
from typing import Any


def fn(value: Any):  # no error
    print(value.asdfsdf)  # no error
```

basedpyright and mypy both have rules for this:

- basedpyright: [`reportAny`](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportany) / [`reportExplicitAny`](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportexplicitany)
- mypy: [`--disallow-any-expr`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-expr) / [`--disallow-any-explicit`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-explicit)

asukaminato0721:
# Summary

Fixes #3442

Added ignored-by-default ErrorKind::ExplicitAny, Emitted it for explicit Any in annotations and type alias bodies when enabled.

Wired migration support for: mypy disallow_any_explicit, mypy explicit-any, pyright/basedpyright reportExplicitAny

Updated error-kind and migration docs.

# Test Plan

Added regression tests

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -153,6 +153,8 @@ pub enum ErrorKind {
     Deprecated,
     /// Division, floor division, or modulo by a literal zero value.
     DivisionByZero,
+    /// Explicit usage of `typing.Any` in an annotation.
+    ExplicitAny,
     /// Raised when a class implicitly becomes abstract by defining abstract members without
     /// inheriting from `abc.ABC` or using `abc.ABCMeta`.
     ImplicitAbstractClass,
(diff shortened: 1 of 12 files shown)


asukaminato0721:
opposite reviews.

https://github.com/facebook/pyrefly/pull/3459#issuecomment-4478660421
https://github.com/facebook/pyrefly/pull/3459#issuecomment-4479901934

yangdanny97:
@migeed-z why is the diff showing two different generated TOML files

stroxler:
> @migeed-z why is the diff showing two different generated TOML files

I think I understand - the wrapper script had to change because it's brittle, if we add new config options it crashes old pyrefly

But we need to figure out how to get rid of the extraneous deltas, which are a side-effect of how the wrapper script tries to avoid this problem. I'll split it off into a new commit and try to fix the issue

yangdanny97 (approved):

PR #3464, opened 2026-05-19
Use per-binary config in primer wrapper to avoid cross-version config crashes

stroxler:
Summary:
When a new pyrefly binary adds error codes (e.g. `explicit-any`) and `pyrefly init` writes them into a shared config, the old binary from the base revision crashes because it cannot parse the unknown error code. Since mypy_primer runs old and new binaries concurrently in the same directory, sharing a single generated config causes spurious primer failures whenever a new error code is introduced.

This change gives each binary its own generated config file (tagged by a checksum of the binary path) and passes it via `--config`. The shared `pyrefly init` call is removed from the lock-protected section since config generation is now per-binary.

Because the per-binary config filenames differ between old and new runs, warnings that reference the config path would show up as spurious diffs. The output is piped through `sed` to normalize the config filename to `pyrefly.toml` so primer compares actual checker behavior.

Differential Revision: D105679430

-------

Exporting to verify that the sed piping works, the original version of this logic incorrectly gave deltas due to the two different filenames

### DIFF
--- a/scripts/pyrefly_primer_wrapper.sh
+++ b/scripts/pyrefly_primer_wrapper.sh
@@ -4,12 +4,12 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Wrapper script for mypy_primer that runs `pyrefly init --non-interactive`
-# before forwarding arguments to the real pyrefly binary.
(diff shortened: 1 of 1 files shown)


yangdanny97 (approved):

PR #3467, opened 2026-05-19
fix support multiple values for python-platform #3440

### LINKED ISSUE #3440: support multiple values for `python-platform`
it's very common for python projects to be run on multiple platforms (especially if you're developing a library). `python-platform` should support specifying an array of values eg:

```toml
python-platform = ["linux", "win32"]
```
and maybe an `"all"` option as well, which both [ty](https://docs.astral.sh/ty/reference/configuration/#python-platform) and [pyright](https://docs.basedpyright.com/latest/configuration/config-files/#:~:text=pythonPlatform) support

asukaminato0721:
# Summary

Fixes #3440

PythonPlatform now can represent one platform, multiple platforms, or all.

Platform checks now only fold to true/false when all configured platforms agree; mixed results stay unknown so both branches are type checked.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -1823,7 +1823,21 @@ mod tests {
         .unwrap();
         assert_eq!(
             config.python_environment.python_platform,
-            Some(PythonPlatform::new_many(vec!["linux".to_owned(), "win32".to_owned()]).unwrap())
+            Some(PythonPlatform::new_many(vec![
+                "linux".to_owned(),
+                "win32".to_owned()
+            ]))
+        );
+
+        let config = ConfigFile::parse_config(
+            r#"
+            python-platform = "linux"
+            "#,
+        )
+        .unwrap();
+        assert_eq!(
+            config.python_environment.python_platform,
+            Some(PythonPlatform::linux())
         );
 
         let config = ConfigFile::parse_config(
@@ -1837,13 +1851,26 @@ mod tests {
             Some(PythonPlatform::All)
         );
 
-        assert!(
-            ConfigFile::parse_config(
-                r#"
-                python-platform = ["all", "linux"]
-                "#,
-            )
-            .is_err()
+        let config = ConfigFile::parse_config(
+            r#"
+            python-platform = ["all", "linux"]
+            "#,
+        )
+        .unwrap();
+        assert_eq!(
+            config.python_environment.python_platform,
+            Some(PythonPlatform::All)
+        );
+
+        let config = ConfigFile::parse_config(
+            r#"
+            python-platform = []
+            "#,
+        )
+        .unwrap();
+        assert_eq!(
+            config.python_environment.python_platform,
+            Some(PythonPlatform::new_many(Vec::new()))
         );
     }
 
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -1814,6 +1814,40 @@ mod tests {
     }
 
     #[test]
+    fn deserialize_python_platform_list() {
+        let config = ConfigFile::parse_config(
+            r#"
+            python-platform = ["linux", "win32"]
+            "#,
+        )
+        .unwrap();
+        assert_eq!(
+            config.python_environment.python_platform,
+            Some(PythonPlatform::new_many(vec!["linux".to_owned(), "win32".to_owned()]).unwrap())
+        );
+
+        let config = ConfigFile::parse_config(
+            r#"
+            python-platform = "all"
+            "#,
+        )
+        .unwrap();
+        assert_eq!(
+            config.python_environment.python_platform,
+            Some(PythonPlatform::All)
+        );
+
+        assert!(
+            ConfigFile::parse_config(
+                r#"
+                python-platform = ["all", "linux"]
+                "#,
+            )
+            .is_err()
+        );
+    }
+
+    #[test]
     fn deserialize_pyrefly_config_snake_case() {
         let config_str = r#"
              project_includes = ["tests", "./implementation"]
--- a/crates/pyrefly_python/src/sys_info.rs
+++ b/crates/pyrefly_python/src/sys_info.rs
@@ -310,19 +310,6 @@ impl PythonPlatform {
         Self::Platforms(SmallSet1::new("darwin".to_owned()))
     }
 
-    /// Return the `os.name` value corresponding to this platform.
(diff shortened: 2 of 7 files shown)


website/docs/configuration.mdx
```diff
@@ -315,12 +315,17 @@ against
 [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform)
 values.
 
-- Type: string
+- Type: string or list of strings
 - Default: result from [Environment Autoconfiguration](#environment-autoconfiguration), or
   "linux" if the Python interpreter cannot be queried
-- Flag equivalent: `--python-platform`
+- Flag equivalent: `--python-platform` (single value only; the list form is config-file only)
```
yangdanny97: shouldn't we just allow passing `--python-platform` multiple times, just like how we can do `--ignore` or `--error` multiple times for error codes?

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -194,40 +185,146 @@ impl PythonPlatform {
...
+    pub fn new_many(platforms: Vec<String>) -> anyhow::Result<Self> {
+        Self::new_platforms(platforms.into_iter().map(|platform| Self::new(&platform)))
+    }
+
+    pub fn new_platforms(platforms: impl IntoIterator<Item = Self>) -> anyhow::Result<Self> {
+        let platforms = platforms.into_iter().collect::<Vec<_>>();
+        if platforms.is_empty() {
+            return Err(anyhow::anyhow!(
+                "`python-platform` list must contain at least one value"
+            ));
```
connernilsen: Instead of making this a function that can return an error, can we instead just infer the platform from the current machine type (which I believe we did previously)?

crates/pyrefly_config/src/config.rs
```diff
@@ -1744,6 +1744,40 @@ mod tests {
         );
     }
 
+    #[test]
+    fn deserialize_python_platform_list() {
```
connernilsen, line 1817: Will this parse the current form of python-platform = "linux"? Can you add a test below validating parsing from a toml string?

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -194,40 +185,146 @@ impl PythonPlatform {
...
+                }
+                Self::Platforms(platforms) => platforms,
+            })
+            .collect::<Vec<_>>();
+        if found_all {
+            if platforms.is_empty() {
+                return Ok(Self::All);
+            }
+            return Err(anyhow::anyhow!(
+                "`all` cannot be combined with other `python-platform` values"
```
connernilsen: Instead, I think for now let's just short circuit and return `all` if it's found, even if there are other values. We can maybe output a warning if you want

I don't love the idea of platforms now being something we might error on

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -162,19 +161,11 @@ impl PythonVersion {
...
-    Hash,
-    Serialize,
-    Deserialize,
-    Display
-)]
-pub struct PythonPlatform(String);
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum PythonPlatform {
+    All,
+    Platforms(Vec<String>),
```
connernilsen: If this needs to have at least one item, please make it a `Vec1`

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -347,10 +444,99 @@ impl<'de> Deserialize<'de> for SysInfo {
...
+    if values.all(|value| value == first) {
+        Some(first)
+    } else {
+        None
+    }
+}
+
 #[derive(Debug, PartialEq, PartialOrd, Clone)]
 enum Value {
     Tuple(Vec<Value>),
```
connernilsen, line 600: Why not reuse the existing tuple setup here?

This isn't necessarily a recommendation to do that, just want to understand what's going on here a little better

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -347,10 +444,99 @@ impl<'de> Deserialize<'de> for SysInfo {
...
+                .map(|(left, right)| match op {
+                    CmpOp::Eq => left == right,
+                    CmpOp::NotEq => left != right,
+                    CmpOp::Lt => left < right,
+                    CmpOp::LtE => left <= right,
+                    CmpOp::Gt => left > right,
+                    CmpOp::GtE => left >= right,
+                    _ => unreachable!("unsupported string comparison operator checked above"),
+                }),
+        )
```
connernilsen, line 566: I don't think I really understand what these cartesian product comparisons passed into `aggregate_bools` do, both here and below. Can you describe them a bit (and add comments)

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -194,40 +185,146 @@ impl PythonPlatform {
...
+        if found_all {
+            if platforms.is_empty() {
+                return Ok(Self::All);
+            }
+            return Err(anyhow::anyhow!(
+                "`all` cannot be combined with other `python-platform` values"
+            ));
+        }
+        platforms.sort();
+        platforms.dedup();
```
connernilsen: This seems like a good reason to use `SmallSet` instead of `Vec` (you can even use `SmallSet1`, which should be an internal struct)

crates/pyrefly_python/src/sys_info.rs
```diff
@@ -194,40 +185,146 @@ impl PythonPlatform {
...
-        match self.0.as_str() {
+    pub fn os_name(&self) -> Option<&str> {
+        let Self::Platforms(platforms) = self else {
+            return None;
+        };
+        let [platform] = &**platforms else {
+            return None;
+        };
+        Some(Self::os_name_for_platform(platform))
+    }
```
connernilsen: What is this function being used for? Its signature changed, so I would expect that to affect things elsewhere in the codebase?

connernilsen (changes requested):
Hey Asuka, thanks for making this. I have a few changes, but a few questions remain on some of the implementation details

crates/pyrefly_python/src/sys_info.rs
asukaminato0721: The existing `Value::Tuple` is meant to model an actual Python tuple/list/set literal, like `("linux", "win32")`.

The new `StringValue::Set` is different: it models one scalar string expression whose runtime value may be one of several strings depending on the configured platform.

connernilsen (approved):

PR #3468, opened 2026-05-19
fix no error on invalid call to `__init_subclass__` #3447

### LINKED ISSUE #3447: no error on invalid call to `__init_subclass__`
### Describe the Bug

```python
class Foo:
    def __init_subclass__(cls, asdf: int) -> None:
        pass


class Bar(Foo, asdf=1): ...
class Baz(Foo, asdf=""): ... # no error
class Qux(Foo): ... # no error
```

it looks like pyrefly doesn't support `__init_subclass__` at all. [`bad-typed-dict`](https://pyrefly.org/en/docs/error-kinds/#bad-typed-dict) special-cases it for `TypedDict`, but ideally that rule should be removed and replaced with a more general rule that checks `__init_subclass__` calls on any class

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuiAOuvV-ZjGPQH0BEdBAAuAuAFds1WnCEAKanAA09WpjCJ6IsQEp6AWgB89AHK50Mdp273i8js-Ry69AEKoAToua51TTAAXgBGfR1CKI43Bi8ALz8WQLgtYLYQDIj6KMIYmncARSl8JNxs3JBVEDJvPihSQjFcAFsoCjIxAAsrIwA3GG84CCt6YPoMgGZCUIAmDI4AbUHvXCGAXQ4MdFwxVDER9CMWiDgW-cousfoAchW1m44QAF9q1EoDgcZoGAo0LDwRDILyAA

_No response_

asukaminato0721:
# Summary

Fixes #3447

Added `__init_subclass__` lookup excluding `object.__init_subclass__`

Class metadata now checks class-header keywords against the first real base class’s `__init_subclass__` via the existing callable checker 

Preserved keyword identifiers through binding so diagnostics can point at class-header keyword names.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -4795,6 +4795,45 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         self.get_dunder_init_helper(&Instance::of_class(cls), get_object_init)
     }
 
+    /// Get the class's `__init_subclass__` method, excluding `object.__init_subclass__`.
+    pub fn get_dunder_init_subclass(
+        &self,
+        cls: &ClassType,
+        include_ancestors: bool,
+    ) -> Option<Type> {
+        if cls.class_object().is_builtin("object") {
+            return None;
+        }
+        let init_subclass_member = if let Some(field) = self
+            .get_non_synthesized_field_from_current_class_only(
+                cls.class_object(),
+                &dunder::INIT_SUBCLASS,
+            ) {
+            WithDefiningClass {
+                value: field,
+                defining_class: cls.class_object().dupe(),
+            }
+        } else if !include_ancestors {
+            return None;
+        } else {
+            let mro = self.get_mro_for_class(cls.class_object());
+            self.get_field_from_ancestors(
+                cls.class_object(),
+                mro.ancestors_no_object().iter(),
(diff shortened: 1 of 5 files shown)


NathanTempest (commented):
Thanks @asukaminato0721, and nice work! The feature itself is a real gap vs. mypy/pyright, so it's worth landing.

This looks good to import but can you be an extra reviewer ? @stroxler :

- MRO resolution: the check resolves __init_subclass__ from bases_with_metadata.first() (walking that base's ancestors). For multiple inheritance where a later base defines __init_subclass__, that method won't be validated, it'd be more correct to follow the new class's full MRO.
- Metaclass skip breadth: if !keywords.is_empty() && metaclass.is_some() { return; } turns off all keyword validation whenever a custom metaclass exists. That's what fixed manticore, but it also silences real errors for keywords the metaclass doesn't consume. Worth considering filtering only the metaclass-consumed keywords instead.

NathanTempest (approved):

yangdanny97 (changes requested):

kinto0 (changes requested):

stroxler (changes requested):

stroxler (approved):

PR #3469, opened 2026-05-19
Add musllinux wheels for Alpine and other musl-based distros

### LINKED ISSUE #3432: Add musllinux-* wheels
### Describe the Bug

I can't use pyrefly because there are no usable wheels and I'm not willing to compile from source.

Could you add [musllinux-*](https://github.com/pypa/manylinux#musllinux_1_2-alpine-linux-322-based-313-compatible) builds along with `manylinux-*` ones (like [ty does](https://pypi.org/project/ty/#files))

In particular, I'm using Alpine Linux version 3.22.2.

```
$ uv venv .venv
Using CPython 3.12.12 interpreter at: /usr/bin/python3
Creating virtual environment at: .venv
$ uv pip install pyrefly==1.0.0 --no-build
  × No solution found when resolving dependencies:
  ╰─▶ Because pyrefly==1.0.0 has no usable wheels and you require
      pyrefly==1.0.0, we can conclude that your requirements are
      unsatisfiable.
      hint: Wheels are required for `pyrefly` because building from source is
      disabled for all packages (i.e., with `--no-build`)
```

### Sandbox Link

_No response_

_No response_

samwgoldman:
Summary:
Pip only picks a wheel whose platform tag matches the running system's libc. Alpine (and other musl distros) advertise `musllinux_*` tags and reject `manylinux_*` ones, so users there saw "no usable wheels" and were forced to build from source.

Add a `musllinux` job to the wheel-building workflow that produces `musllinux_1_2` wheels for `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl`, mirroring the existing manylinux matrix. The aarch64 build reuses the `JEMALLOC_SYS_WITH_LG_PAGE=16` workaround from the gnu aarch64 build. The x86_64 wheel is smoke-tested inside an `alpine:latest` container; aarch64 is build-only, matching the existing aarch64-gnu pattern. The new job is added to `merge.needs` so the wheels land in the consolidated `dist` artifact that gets published to PyPI.

Fixes https://github.com/facebook/pyrefly/issues/3432

Differential Revision: D105720763

### DIFF
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -201,6 +201,63 @@ jobs:
           name: wheels-${{ matrix.target }}
           path: pyrefly/dist
 
+  musllinux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+          - x86_64-unknown-linux-musl
+          - aarch64-unknown-linux-musl
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+          persist-credentials: false
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          architecture: x64
+      - name: Build wheels
+        if: ${{ matrix.target == 'x86_64-unknown-linux-musl' }}
+        uses: PyO3/maturin-action@v1
+        with:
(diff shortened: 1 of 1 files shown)


rchen152 (approved):

PR #3487, opened 2026-05-20
Convert `type_check_program` from panics to collected errors (#3487)

stroxler:
Summary:

**This stack**

Reworks tensor shape operations so that instead of being hardcoded in tensor_ops_registry.rs, only DSL primitives are
directly hardcoded into Pyrefly; the actual operations and the association with "normal" (non DSL) stubs all lives in user-space
stub files. This allows iterating on the ops without rebuilding Pryefly, and is 100% essential for actually building out full stubs
for pytorch (and even more so if we want to extend to other libraries like numpy and jax).

The DSL itself is unchanged but we will use a decorator to indicate when a stub function is a DSL function; we use a different
decorator to actually register a DSL function as the "shape transform" associated with some normal function (e.g. to
associate the DSL function `reshape_ir` with a `torch.reshape` function).

Details of the plan are in https://github.com/stroxler/pyrefly-docs/blob/main/tensor-shapes-in-stubs/v2-doc.md

**This commit**

Replace ~20 panic sites in the DSL type checker with error collection,
so type errors in `shape_dsl_function` stubs produce diagnostics instead
of crashing the type checker.

`type_check_program` now returns `Result<(), Vec<String>>`, threading
errors through `check_body`, `check_expr`, `infer_expr`, `infer_call`,
and the narrowing/joining helpers. `validate_shape_dsl_functions`
propagates these errors to the solver, which emits them as
`InvalidArgument` diagnostics on the function definition.

Eval-time panics in `eval_dsl_body` / `eval_dsl_expr` are intentionally
left as panics — they are correctness assertions that should be
unreachable for type-checked programs.

Differential Revision: D105783604


rchen152 (approved):

PR #3516, opened 2026-05-21
Support multiple glob patterns in sub-config matches

kavix:
This change lets [[[sub-config]]](http://vscodecontentref/1) blocks target more than one path pattern by accepting a list of glob strings in [matches](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). The matching logic now applies a sub-config when any pattern matches, which makes large config migrations much less repetitive.

It also keeps existing single-string configs working for compatibility, updates MyPy and Pyright migration paths to emit the new shape, and refreshes the schema, examples, and docs to reflect the new syntax.

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -52,6 +52,8 @@ use pyrefly_util::telemetry::TelemetrySourceDbRebuildInstanceStats;
 use pyrefly_util::telemetry::TelemetrySourceDbRebuildStats;
 use pyrefly_util::watch_pattern::WatchPattern;
 use serde::Deserialize;
+use serde::de::SeqAccess;
+use serde::de::Visitor;
 use serde::Serialize;
 use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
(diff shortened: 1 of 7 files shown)


NathanTempest:
Hey Kavix, thank you for signing the Meta CLA license and welcome aboard to the team! Let me review this PR and get back to you. This is a useful feature that will reduce config duplication.

Tagging @kinto0 for config ownership review.

Put up some comments on the follow-up, once these are addressed, this should be good to land!

crates/pyrefly_config/src/config.rs
```diff
@@ -1538,9 +1592,13 @@ impl ConfigFile {
...
                     let extra_keys = sub_config.settings.extras.0.keys().join(", ");
+                    let matches = sub_config
+                        .matches
+                        .iter()
+                        .map(|glob| glob.to_string())
+                        .join(", ");
                     errors.push(ConfigError::warn(anyhow!(
-                        "Extra keys found in sub config matching {}: {extra_keys}",
-                        sub_config.matches
+                        "Extra keys found in sub config matching {matches}: {extra_keys}"
```
NathanTempest, line 1601: Please verify this compiles and the output is readable. The change from Glob to Vec<Glob> means we can't use {} directly on sub_config.matches anymore. The new code computes a joined string, which is the right approach, but I want to make sure the anyhow! macro correctly captures the matches and extra_keys variables. A test case for this warning path would be great.

schemas/pyrefly.json
```diff
@@ -291,11 +382,23 @@
...
                 "type": "object",
-                "required": ["matches"],
+                "required": [
+                  "matches"
+                ],
                 "properties": {
                   "matches": {
-                    "description": "A filesystem glob pattern detailing which files the config applies to.",
-                    "type": "string"
+                    "description": "One or more filesystem glob patterns detailing which files the config applies to.",
```
NathanTempest, line 390: This file has a lot of formatting changes unrelated to the matches field (enum values expanded to multi-line, arrays expanded, etc.). Could you regenerate using the standard tooling to minimize diff noise? The actual schema change for matches (allowing string or string[]) is correct.

NathanTempest (changes requested):
The implementation is solid, the serde deserializer correctly handles both string and array inputs, and the migration code properly emits arrays. I especially like that the mypy migration now consolidates multiple modules into a single sub-config with multiple globs.

Before we merge, a few things to address:

Please verify the error message formatting in crates/pyrefly_config/src/config.rs. The change from sub_config.matches (a Glob) to computing a matches string and using it in the format string looks correct, but please confirm this compiles and consider adding a test that triggers the "Extra keys found in sub config" warning to ensure the formatting works as expected.

Schema file formatting. The schemas/pyrefly.json diff includes a lot of whitespace changes unrelated to the matches field (enum values expanded to multi-line, arrays expanded, etc.). Could you regenerate this file using the project's standard schema generation tooling to avoid formatting drift? If the schema is hand-maintained, please keep the formatting consistent with the existing file.

crates/pyrefly_config/src/migration/pyright.rs
```diff
@@ -40,7 +40,7 @@ impl ExecEnv {
             ..Default::default()
         };
         Ok(SubConfig {
-            matches: Glob::new(self.root)?,
+            matches: vec![Glob::new(self.root)?],
```
kinto0, line 43: did pyright allow both single glob and vec of glob? can we have tests for what happens in both cases?

website/docs/configuration.mdx
```diff
@@ -1187,15 +1187,15 @@ and would like to see the option supported.
 
 A SubConfig has two or more entries:
 
-- a `matches` key, with a [Filesystem Glob](#filesystem-globbing) detailing which files the config
-  applies to.
+- a `matches` key, with a [Filesystem Glob](#filesystem-globbing) or list of globs detailing which
```
kinto0, line 1190: if it should be both a single glob or a list, do we have test coverage for both?

schemas/pyrefly.json
kinto0: so do the .toml files underneath

rchen152:

PR #3517, opened 2026-05-21
Fix `DeprecationWarning` from `jsonschema` in `test.py`

jorenham:
# Summary

This addreses the following deprecation warning while running `test.py`:

```
/home/joren/Workspace/pyrefly/schemas/validate_schemas.py:58: DeprecationWarning: jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the https://github.com/python-jsonschema/referencing library, which provides more compliant referencing behavior as well as more flexible APIs for customization. A future release will remove RefResolver. Please file a feature request (on referencing) if you are missing an API for the kind of customization you need.
  resolver = jsonschema.RefResolver.from_schema(schema, store=store)
```

The `referencing` library is already a required dependency of `jsonschema` (https://github.com/python-jsonschema/jsonschema/blob/c6dc09209c0224af200f0663c52cecaea063de36/pyproject.toml#L43), so there's no need to include it in the `pip install` error message.

# Test Plan

N/A

### DIFF
--- a/schemas/validate_schemas.py
+++ b/schemas/validate_schemas.py
@@ -25,6 +25,7 @@ from pathlib import Path
 
 try:
     import jsonschema
+    import referencing
     import toml
 except ImportError:
     print("Error: Required packages not installed.")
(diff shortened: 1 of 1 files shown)


rchen152 (approved):
LGTM, thank you!

yangdanny97 (approved):

PR #3518, opened 2026-05-21
Fix Windows CI on main: normalize `.mdx` EOL

jorenham:
# Summary

This addresses an issue was introduced in https://github.com/facebook/pyrefly/commit/25ac9f6e185782d6c722c7ead9e19b8bb8e1fcab causing the Windows CI job to fail on main. 

See for example https://github.com/facebook/pyrefly/actions/runs/26207799459/job/77111366325

It might be interesting to use this `.gitattributes` for other things as well, such as marking of certain files as generated to clean up github diffs.

# Test Plan

I guess that'll be waiting to see if the Windows CI job no longer fails now ;)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.mdx text eol=lf


NathanTempest:
Thanks @jorenham for tracking this down and fixing it!

The root cause analysis is spot-on. Commit 25ac9f6 introduced .mdx files with CRLF endings, and without a .gitattributes rule, Git on Windows checks them out with CRLF, which breaks the tests that expect LF.

The fix is minimal and correct:
- Adds *.mdx to .gitattributes with eol=lf
- This ensures consistent line endings across all platforms

Good call on mentioning the .gitattributes for generated files too as that would definitely clean up diffs. Might be worth a follow-up PR.

Approved and will import.

NathanTempest (approved):

rchen152 (approved):

PR #3521, opened 2026-05-21
Fix inherited stub method.counts in `pyrefly coverage report`

### LINKED ISSUE #3519: `pyrefly report` counts inherited typed stub methods as untyped in NumPy
For `__init__.pyi`, `pyrefly report` outputs

```json
        {
          "kind": "function",
          "name": "numpy.ndarray.__array_wrap__",
          "n_typable": 4,
          "n_typed": 4,
          "n_any": 0,
          "n_untyped": 0,
          "location": {
            "line": 2138,
            "column": 5
          }
        },
```

So it correctly counts `ndarray.__array_wrap__` as fully typed.

However, for `core/defchararray.pyi`, it outputs

```json
        {
          "kind": "function",
          "name": "numpy._core.defchararray.chararray.__array_wrap__",
          "n_typable": 4,
          "n_typed": 0,
          "n_any": 0,
          "n_untyped": 4,
          "location": {
            "line": 582,
            "column": 5
          }
        },
```

The `chararray` here is a subclass of `ndarray`. Because the `chararray.__array_wrap__` method has the same signature as `ndarray.__array_wrap__`, there is no need to override it in the [stubs](https://github.com/numpy/numpy/blob/b832a09cf2a169c833dd2371e7c07aa00b293242/numpy/_core/defchararray.pyi#L95). 
Yet `pyrefly report` counts `chararray.__array_wrap__`, which is statically the very same method as `ndarray.__array_wrap__`, as fully **un**typed.

I suspect that the issue is that `pyrefly report` looks at the `.py`, sees that there's a `chararray.__array_wrap__` method. It then, without considering the superclass, "requires" that `chararray.__array_wrap__` must also exists in the stubs.

jorenham:
# Summary

This fixes an issue where inherited methods in stubs from being incorrectly counted as untyped. This affects NumPy in several places, causing its type coverage to be too low.

Fixes #3519

# Test Plan

This includes unit- and integration tests.

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -24,7 +24,7 @@ use pyrefly_python::module_path::ModuleStyle;
 use pyrefly_python::nesting_context::NestingContext;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_types::callable::PropertyRole;
-use pyrefly_types::class::ClassDefIndex;
+use pyrefly_types::class::{ClassDefIndex, ClassType};
 use pyrefly_types::types::Type;
 use pyrefly_util::forgetter::Forgetter;
 use pyrefly_util::includes::Includes;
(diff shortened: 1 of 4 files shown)


yangdanny97:
failing test seems unrelated, maybe trunk is broken

stroxler (approved):

PR #3527, opened 2026-05-21
Fix stub function counts of reexports in `pyrefly coverage report`

### LINKED ISSUE #3524: `pyrefly report` counts stub functions as untyped if they're re-exported in the `.py` module
For example, in NumPy, pyrefly reports

```json
        {
          "kind": "attr",
          "name": "numpy._core.multiarray.array",
          "n_typable": 1,
          "n_typed": 0,
          "n_any": 0,
          "n_untyped": 1,
          "location": {
            "line": 36,
            "column": 49
          }
        },
```

for the fully typed `array` **function** (so not an `attr`): https://github.com/numpy/numpy/blob/b832a09cf2a169c833dd2371e7c07aa00b293242/numpy/_core/multiarray.pyi#L458-L517

This seems to be related to the fact that this `array` function is not directly defined in the corresponding `.py`; it instead re-exports it (from a C extension module).

jorenham:
# Summary

I confirmed that this fixes the issue in NumPy in all 21 cases, increasing strict coverage from 87.0% -> 94.3%.

Fixes #3524

# Test Plan

This includes unit- and integration tests.

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -1701,26 +1701,27 @@ impl ReportArgs {
         py_variables: Vec<Variable>,
         py_classes: Vec<ReportClass>,
     ) {
-        let stub_func_names: SmallSet<String> =
-            stub_functions.iter().map(|f| f.name.clone()).collect();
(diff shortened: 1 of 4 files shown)


stroxler (approved):

PR #3539, opened 2026-05-22
Fix crashes / internal errors induced by parser error recovery

### LINKED ISSUE #3343: panic: Name '' not found in static scope of module `__unknown__` with fuzzed code
### Describe the Bug

Starting with commit 44020084, Pyrefly crashes when checking this fuzzed code:

```python
@:[
```

```
ERROR Thread panicked, shutting down: panicked at pyrefly/lib/binding/bindings.rs:1681:17:
Name `` not found in static scope of module `__unknown__`.
```

### Bisection log
44020084 - Crash starts here
020a041f - No crash
0.64.1 - No crash

### Version info
The crash is still reproducible with the latest main (2cfa1c5a).

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAAogNogA0IZATjGFKYQC64C2UFAMQACAAqlW7UsLRY8%2BYQGNc6SAHMArs1RcIKwgB10IgMowYwgBZcuxOIgD0Dlmw6FczNQ5joHmXIpwDsqqEJrauioOwmAewqgAbqjQqNiwSirqWjp66MK4xJHocIboZFyWKgC0CTDMcLnCALzCBiAAzIQAjABMbUY0dcwecAC6RhroELzEHlwwmFWYEKyKurXNwgDkYegeMFtGe1xVrACOGisLVQDWMKRVqIqK8HCbWwDuqMzoh%2BggAF9GE91jAAGLQGAUWQ4AgkcgAoA

_No response_

fangyi-zhou:
# Summary

Fixes #3343 

The fix is similar to 4980ddf9d3f46f99b6e581279a57b289bd453a63 / D104345441, just need to apply it to a few more places.

Asked GPT 5.5 to find a few other places where issues may occur and fix them as well, hopefully we're good.

P.S. I was testing the pre-patch behaviour using the web sandbox. It turns out that if the sandbox crashes, the text input is cached in local storage and makes the web sandbox unuseable without clearing local storage!

# Test Plan

New regression tests added

### DIFF
--- a/pyrefly/lib/binding/pattern.rs
+++ b/pyrefly/lib/binding/pattern.rs
@@ -103,9 +103,15 @@ impl<'a> BindingsBuilder<'a> {
                 // If there's no name for this pattern, refine the variable being matched
                 // If there is a new name, refine that instead
                 let original_subject = match_subject.clone();
-                let alias_name = p.name.as_ref().map(|name| name.id.clone());
+                let alias_name = p
+                    .name
+                    .as_ref()
+                    .filter(|name| !Ast::is_synthesized_empty_identifier(name))
+                    .map(|name| name.id.clone());
                 let mut subject = match_subject;
-                if let Some(name) = &p.name {
+                if let Some(name) = &p.name
+                    && !Ast::is_synthesized_empty_identifier(name)
+                {
                     self.bind_definition(name, Binding::Forward(subject_idx), FlowStyle::Other);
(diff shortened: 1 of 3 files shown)


yangdanny97:
That sandbox behavior sounds kind of bad lol. We should allow the text input to be edited and update local storage independently of the WASM bits, that way after a crash the user can fix their code and reload or something.

stroxler (approved):

PR #3545, opened 2026-05-22
fix show variance when hovering over generics #3540

### LINKED ISSUE #3540: show variance when hovering over generics
```py
class Foo[T]:
    def foo(self) -> T: ...
```

both ty and pyright do this:

Image

pyrefly only shows this:

Image

asukaminato0721:
# Summary

Fixes #3540

now class type-parameter hovers find the owning class, query the existing variance solver, and render `T@Foo (covariant)` without the redundant `(type parameter) T: ...` prefix.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -22,6 +22,7 @@ use pyrefly_python::docstring::parse_parameter_documentation;
 use pyrefly_python::ignore::Ignore;
 use pyrefly_python::ignore::Tool;
 use pyrefly_python::ignore::find_comment_start_in_line;
+use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_python::symbol_kind::SymbolKind;
 use pyrefly_types::callable::Callable;
 use pyrefly_types::callable::FunctionKind;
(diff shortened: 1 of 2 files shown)


pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -359,6 +363,61 @@ fn expand_callable_kwargs_for_hover<'a>(
...
+        // Bivariant is an internal inference result. User-facing behavior
+        // treats it like invariant, matching assignability.
+        Variance::Invariant | Variance::Bivariant => "invariant",
+    }
+}
+
+fn type_parameter_hover_display(
+    solver: &AnswersSolver<TransactionHandle<'_>>,
+    type_: &Type,
+    owner: &Class,
```
yangdanny97, line 519: can't the type param be owned by a function definition too?

pyrefly/lib/test/lsp/hover.rs
```diff
@@ -1345,6 +1345,24 @@ Box[str]("hello")
...
+#         ^
+    def foo(self) -> T: ...
+"#;
+    let report = get_batched_lsp_operations_report(&[("main", code)], get_test_report);
+    assert!(
+        report.contains("T@Foo (covariant)"),
+        "Expected class type parameter hover to show inferred variance, got: {report}"
+    );
+    assert!(
+        !report.contains("(type parameter) T: T"),
```
yangdanny97, line 1391: our format seems somewhat standardized with the kind of symbol in parentheses in front of the type, i agree that maybe it's redundant, but I wonder if this is introducing inconsistencies

pyrefly/lib/lsp/wasm/hover.rs
yangdanny97: and perhaps another question is, is it important/useful to display the owner at all? we didn't do it previously, and IDK how much people understand the `@` notation, even if Pyright does use it

rchen152 (approved):

PR #3549, opened 2026-05-22
Fix `bad-specialization` error for generic overloaded callback protocols #3548

### LINKED ISSUE #3548: 0.64 regression: false positive `bad-specialization` in scipy-stubs
### Describe the Bug

After upgrading Pyrefly in scipy-stubs to 0.64.0, Pyrefly reported a new false positive `bad-specialization` error. At first I thought it might be the same issue as #3329, but after more investigation, I think it's a separate issue.

Here's the repro.zip:

```py
from typing import Callable, Protocol, overload

class Base: ...

class HasCall(Protocol):
    @overload
    def __call__[T: Base](self, arg: T) -> T: ...
    @overload
    def __call__(self, arg: float) -> float: ...

def takes(f: Callable[[float], float]) -> None: ...

def repro(p: HasCall):
    takes(p)  # E: bad-specialization
```

This is still on issue on main.

Blocked scipy-stubs PR: https://github.com/scipy/scipy-stubs/pull/1593

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BhVFCipssADR0ACtQa4AxrihTcANxiUouVJgA66AwtFw4dAEKo4MRHUL2DRk2YASVoSIAUs3PKVQASkQDOlC6AAF1TW1dELDMGDA6AH1khWEoVIBtABVbS2sAXU9rKDApVEo2WxyAugBaAD46PLsHdDCIqK0dfQ74xJS0jNSSmDKKqtswGIY6proZnQZbe0JHLEGGVABreE8wWw9RcRgsrKXUBkKpS%2Bv55oA5XHQbNvXDTaTKGGJqT2ItjccGOQTioW2ezgALqdAAxHQAKK2bC6epwYgwBQQYQQABeVwgLwMIAkIDIPxmpEI8hoUAoAFd0ExMZh6gkwPVsDAABaoNREyh0AC8dD0IAUPKxO3qGDZrDAmnqPwYDMo6Fl6FI4pAAF8yagFAwIBoAGLQGAUNBYPBEMh6oA

_No response_

jorenham:
# Summary

Fixes #3548

# Test Plan

Integration test added

### DIFF
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -1764,6 +1764,27 @@ impl Solver {
         vs.0.iter().any(|v| lock.contains_key(v))
     }
 
+    /// Snapshot/restore pair for `instantiation_errors`, used by `try_overload_arms` to drop
+    /// recordings from speculatively-tried overload arms that didn't cleanly match.
+    pub fn instantiation_error_keys(&self) -> SmallSet<Var> {
+        self.instantiation_errors.read().keys().copied().collect()
+    }
+
+    /// Returns whether any entries were removed.
+    pub fn discard_new_instantiation_errors(&self, snapshot: &SmallSet<Var>) -> bool {
+        let mut errors = self.instantiation_errors.write();
+        let to_remove: Vec<Var> = errors
+            .keys()
+            .copied()
+            .filter(|v| !snapshot.contains(v))
+            .collect();
+        let removed = !to_remove.is_empty();
+        for v in to_remove {
+            errors.shift_remove(&v);
+        }
+        removed
+    }
+
     /// Have these vars picked up any new instantiation errors since they were snapshotted?
     pub fn has_new_instantiation_errors(&self, snapshot: &VarSnapshot) -> bool {
         let lock = self.instantiation_errors.read();
--- a/pyrefly/lib/solver/subset.rs
+++ b/pyrefly/lib/solver/subset.rs
@@ -29,6 +29,7 @@ use pyrefly_types::typed_dict::ExtraItems;
 use pyrefly_types::typed_dict::TypedDict;
 use pyrefly_types::typed_dict::TypedDictField;
 use pyrefly_types::types::Overload;
+use pyrefly_types::types::OverloadType;
 use pyrefly_types::types::Var;
 use pyrefly_util::owner::Owner;
 use ruff_python_ast::name::Name;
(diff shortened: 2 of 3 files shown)


jorenham:
I had Claude (Opus 4.7 xhigh) take a look at the mypy_primer output, and I thought that its analysis was worth sharing:

---

Here's my read of each affected project in the mypy_primer diff:

## Removed errors (improvements from the fix)

- **scipy-stubs** — the original target. True positive removal. ✓
- **static-frame** (4 errors on `_AnyDT64Arg`) — same shape as scipy-stubs (`TypeVar` bound + overloaded match). True positive removals.
- **spark** (`SupportsRichComparisonT` bound) — the spurious bound-check error is replaced by `bad-argument-type` from the genuine overload-resolution failure at the same call site. Net improvement — the underlying call is actually invalid.
- **more-itertools** (`incompatible-overload-residual`) — almost certainly fallout of error isolation; the fix gives the residual mechanism a cleaner view. Likely correct (worth a quick look but not a blocker).

## Added errors

- **spark** (`bad-argument-type` for `bisect.bisect_left`) — **true positive**, not a regression. The previous `bad-specialization` was hiding the real problem (the code passes `list[K]` to a function expecting `SupportsLenAndGetItem[K]` for an unbounded `K`, and the bound check was a confused proxy for that). The new error is the right error.

- **pandas-stubs** (8 `assert-type` failures in `test_add.py`/`test_mul.py`) and **pandas** (1 `no-matching-overload`) — **real regressions**, but they expose a pre-existing pyrefly bug, not a defect in this fix. I traced it:
  - Before the fix, on `Index[bool] + np.ndarray[..., bool]`, overload arm 7 `(self: Index[T_COMPLEX], other: np_ndarray_bool | Index[bool]) -> Index[T_COMPLEX]` matched and produced `Index[bool]`.
  - With the fix, an earlier arm 5 (`Supports_ProtoAdd[T_contra, S2] -> Index[S2]`) matches and produces `Index[<union of every numpy dtype>]`. The `Supports_ProtoAdd` structural match itself is buggy — `S2` should resolve to `bool`, not that union.
  - Before, arm 5 was *accidentally* failing because leaked `instantiation_errors` from arm 1's probe were "carrying over" and breaking arm 5's bind. The right answer fell out by luck. The fix correctly isolates per-arm side effects, which uncovers the broken `Supports_ProtoAdd` resolution.

## Recommendation

Don't gate this PR on pandas-stubs/pandas. The regressions are honest exposures of a separate `Supports_ProtoAdd`/protocol-overload bug — file as follow-up. The other diffs are net positive (scipy-stubs/static-frame/spark all become more correct). If reviewers push back, the alternative is to also snapshot/restore `deferred_quantified_vars` per-arm (like the witness path does for `captured_vars`) so failed arms leave *no* state behind, not just no errors — that would likely paper over the pandas regression by restoring the previous lucky-fail of arm 5. But it would also be a much bigger change in semantics.

jorenham:
marking as draft; because there's probably a better way to do this

EDIT: Eh, nevermind; that'd require a big refactor, which would be out of scope for this PR

rchen152 (commented):
Thanks for the PR! Reverting specialization errors when we don't select an overload branch is the right idea, but implementation-wise, I'd rather reuse the solver's existing snapshot/restore mechanism (https://github.com/facebook/pyrefly/blob/e993c40e66b2758b0a1797c860aa0fa212daf9b4/pyrefly/lib/solver/solver.rs#L676) than add another one. I've implemented that in https://github.com/facebook/pyrefly/pull/3592/changes/38d8624865441e7e2b39cac4f96bca432c79e501. (Sorry for swiping this bug fix XD That commit depends on a stack of 9 other in-flight commits, so I figured me taking this over would be the least messy option.)

jorenham:
> Thanks for the PR! Reverting specialization errors when we don't select an overload branch is the right idea, but implementation-wise, I'd rather reuse the solver's existing snapshot/restore mechanism (
> 
> https://github.com/facebook/pyrefly/blob/e993c40e66b2758b0a1797c860aa0fa212daf9b4/pyrefly/lib/solver/solver.rs#L676
> 
> ) than add another one. I've implemented that in [38d8624](https://github.com/facebook/pyrefly/commit/38d8624865441e7e2b39cac4f96bca432c79e501). (Sorry for swiping this bug fix XD That commit depends on a stack of 9 other in-flight commits, so I figured me taking this over would be the least messy option.)

Yea no problem; I'm not doing this for the fame haha

PR #3555, opened 2026-05-22
Fix document symbol search for editors without hierarchical support

### LINKED ISSUE #2945: Document symbol search not working on helix
### Describe the Bug

needs repro

https://www.reddit.com/r/HelixEditor/comments/1s4jod9/no_document_symbol_search_spaces_with_pyrefly_or/

### Sandbox Link

_No response_

_No response_

mmatz-101:
# Summary

Editors like Helix do not declare `hierarchicalDocumentSymbolSupport`
in their client capabilities. Previously pyrefly treated this as a
signal to return nothing, so document symbol search was completely
broken in Helix.

Check the client capability at request time and return a flat
`SymbolInformation` list when hierarchical support is absent, falling
back to the existing nested `DocumentSymbol` response for clients like
VS Code that support it. The flat list is produced by recursively
unwrapping the symbol tree, preserving dotted container names
(e.g. `Outer.Inner`) so clients can still display symbol hierarchy
contextually.

Fixes #2945 

# Test Plan

Added unit tests for `flatten_to_symbol_information` covering:
- Basic class with methods flattened correctly with container_name
- Deep nesting with dotted container names (e.g. Outer.Inner)
- Empty file returns empty list

Run with: cargo test test_flatten

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/document_symbols.rs
+++ b/pyrefly/lib/lsp/non_wasm/document_symbols.rs
@@ -287,3 +287,42 @@ fn recurse_stmt_adding_symbols<'a>(
     };
     symbols.append(&mut recursed_symbols);
 }
+
+pub fn flatten_to_symbol_information(
+    symbols: Vec<DocumentSymbol>,
+    uri: &lsp_types::Url,
+) -> Vec<lsp_types::SymbolInformation> {
+    let mut results = Vec::new();
+    flatten_recursive(symbols, uri, None, &mut results);
+    results
+}
+
+fn flatten_recursive(
+    symbols: Vec<DocumentSymbol>,
+    uri: &lsp_types::Url,
+    container_name: Option<String>,
+    result: &mut Vec<lsp_types::SymbolInformation>,
+) {
+    for sym in symbols {
+        let children = sym.children.unwrap_or_default();
+        let qualified_name = match &container_name {
+            Some(parent) => format!("{}.{}", parent, sym.name),
+            None => sym.name.clone(),
+        };
+
+        #[allow(deprecated)]
+        result.push(lsp_types::SymbolInformation {
+            name: sym.name,
+            kind: sym.kind,
+            tags: sym.tags,
+            deprecated: sym.deprecated,
+            location: lsp_types::Location {
+                uri: uri.clone(),
+                range: sym.range,
+            },
+            container_name: container_name.clone(),
+        });
+
+        flatten_recursive(children, uri, Some(qualified_name), result);
+    }
+}
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -5115,16 +5115,24 @@ impl Server {
         }) {
             return Err(EmptyResponseReason::LanguageServicesDisabled);
         }
-        let handle = self.make_handle_if_enabled(uri, Some(DocumentSymbolRequest::METHOD))?;
-        let symbols = transaction.symbols(&handle, maybe_cell_idx);
-        let supports_hierarchical = self
+
+        // Avoid creating a handle when the client doesn't support document symbols
+        let document_symbols_caps = self
             .initialize_params
             .capabilities
             .text_document
             .as_ref()
-            .and_then(|t| t.document_symbol.as_ref())
+            .and_then(|t| t.document_symbol.as_ref());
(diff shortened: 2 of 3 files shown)


Arths17:
Good direction overall this correctly restores symbol visibility for non-hierarchical clients like Helix.

I think the key thing to validate is that flattening is purely a presentation-layer transformation and does not leak back into any semantic symbol resolution logic.

If that separation holds, this is a clean compatibility fix. Might be worth adding a regression test for mixed-capability clients (hierarchical true vs false) to ensure identical symbol sets, just different shapes.

yangdanny97:
Is this flattened representation part of the LSP & supported in other editors that don't support hierarchical document symbols, or is this something specific to Helix?

I'm not sure what impact (if any) there is on other editors

I'll let @kinto0 make the call here

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -5101,34 +5100,38 @@ impl Server {
...
-            .workspaces
-            .get_with(path, |(_, workspace)| workspace.disable_language_services)
-        {
+        if self.workspaces.get_with(path, |(_, workspace)| {
+            workspace.disabled_language_services.is_some()
+        }) {
             return Err(EmptyResponseReason::LanguageServicesDisabled);
         }
-        let Some(true) = self
+        let handle = self.make_handle_if_enabled(uri, Some(DocumentSymbolRequest::METHOD))?;
```
kinto0: nit: we should avoid making the handle if they don't have any support for any document symbols

pyrefly/lib/lsp/non_wasm/document_symbols.rs
```diff
@@ -287,3 +287,42 @@ fn recurse_stmt_adding_symbols<'a>(
...
+}
+
+fn flatten_recursive(
+    symbols: Vec<DocumentSymbol>,
+    uri: &lsp_types::Url,
+    container_name: Option<String>,
+    result: &mut Vec<lsp_types::SymbolInformation>,
+) {
+    for sym in symbols {
+        let children = sym.children.unwrap_or_default();
```
kinto0, line 307: can you double check there's no better way to flatten `DocumentSymbol`s through the lsp_types library?

kinto0 (changes requested):
looks great! thanks for working on it. I'd like a few more tests like lsp_interaction tests to make sure the behavior works end-to-end. and is there a way to make every document_symbols test test both flat + nested symbols? then we won't need to add any new tests

mdomke:
I would be really great if we could push this a forward, because it makes using `pyrefly` in Helix a bit of a downgrade atm. I enjoy the rest of the LSP functionality, but it would be marvellous to be able to use the symbol picker (again).

mmatz-101:
@kinto0, thanks for your time reviewing.

> looks great! thanks for working on it. I'd like a few more tests like lsp_interaction tests to make sure the behavior works end-to-end. and is there a way to make every document_symbols test test both flat + nested symbols? then we won't need to add any new tests

Commit (4b9a0d1ec8236478b38cc2c6e286d94378980dba)
- Unified document symbol tests to assert both hierarchical `DocumentSymbol` and flat `SymbolInformation` output from the same test inputs.
- Added combined-report/section parsing helpers so assertion-based tests validate both modes consistently.

> nit: we should avoid making the handle if they don't have any support for any document symbols

Commit (12a6cf200b916aa4928307b71479a40287d8d9c4)
- Skips `make_handle_if_enable` if client doesn't support document symbols

> can you double check there's no better way to flatten DocumentSymbols through the lsp_types library?

The lsp_types crate doesn't provide this utility because SymbolInformation requires the URI context.
- DocumentSymbol is self-contained - it only has ranges relative to the document being queried.
- SymbolInformation is the problem. It requires a Location which bundles a uri together with the range.

pyrefly/lib/lsp/non_wasm/server.rs
mmatz-101: fixed

pyrefly/lib/lsp/non_wasm/document_symbols.rs
mmatz-101: addressed

connernilsen (approved):

mmatz-101:
Thank you for your review and your time!

@kinto0 @connernilsen

yangdanny97:
Thanks for your contribution @mmatz-101 !

PR #3557, opened 2026-05-22
Fix pydantic private attribute handling in field inference

### LINKED ISSUE #3544: pydantic - protected attributes are incorrectly treated as fields
### Describe the Bug

```py
from pydantic import BaseModel


class Foo(BaseModel):
    a: int
    _a: int

    def initialize(self):
        self._a = 1


foo = Foo(a=1) # error: Missing argument `_a` in function `Foo.__init__` [missing-argument]
```

in pydantic, an attribute prefixed with `_` is not considered a field. this code works at runtime

### Sandbox Link

_No response_

_No response_

Arths17:
Problem: _ attributes incorrectly treated as fields
Root cause: field extraction logic does not exclude private attributes
Fix: add rule to exclude attributes starting with _
Testing: cargo test + reproduction verified fixed

Closes #3544


yangdanny97:
thanks!

rchen152 (approved):

PR #3558, opened 2026-05-23
Add safe mode recovery for sandbox crashes

fangyi-zhou:
## Summary
Quick follow up from #3539 regarding sandbox crashes, vibed coded a crash helmet called "safe mode"
- add a sandbox-specific crash fallback with a safe-mode recovery button
- preserve the user's sandbox files while safe mode pauses wasm-backed type checking
- add a resume action and results-panel messaging for paused type checking

## Test Plan
- yarn test sandbox --runInBand
- built and served the production website locally with the generated wasm artifacts

https://github.com/user-attachments/assets/55f89b99-69c6-4562-949b-5d358f9fe4b0

### DIFF
--- a/website/src/pages/sandbox.tsx
+++ b/website/src/pages/sandbox.tsx
@@ -12,6 +12,7 @@ import BrowserOnly from '@docusaurus/BrowserOnly';
 import Layout from '@theme/Layout';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import * as stylex from '@stylexjs/stylex';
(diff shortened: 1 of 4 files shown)


kinto0 (approved):

yangdanny97 (approved):

kinto0:
landed [here](https://github.com/facebook/pyrefly/commit/2d0124727aadc34f587b1eeb3e13bacf0e25a201), closing

I think it didn't auto-close because the imported summary was changed since importing. I think you need to leave the metadata in the summary for it to auto-close

PR #3565, opened 2026-05-24
fix pydantic - Strict types are not treated as strict mode #3543

### LINKED ISSUE #3543: pydantic - `Strict` types are not treated as strict mode
### Describe the Bug

```py
from pydantic import BaseModel, StrictInt


class Foo(BaseModel):
    a: StrictInt = 1


foo = Foo(a="1")  # no error, crashes at runtime
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3543

detects Pydantic Strict metadata on Annotated aliases like StrictInt, then carries that metadata into per-field strict flags, including assigned defaults.

# Test Plan

add test


rchen152 (approved):

grievejia (approved):

PR #3570, opened 2026-05-24
fix: improve decorator error messages for injected parameter mismatches

### LINKED ISSUE #3569: decorator error message is very confusing
### Describe the Bug

Argument `(self: Self@ExternalApiTemplateListApi) -> tuple[dict[str, bool | int | list[ExternalKnowledgeApiDict] | None], int]` is not assignable to parameter `view` with type `(@_, str, ParamSpec(@_)) -> @_` in function `controllers.console.wraps.with_current_tenant_id`

all I need to do is

```py
def get(self): ...
```

to

```py
def get(self, current_tenant_id: str): ...
```

so the error message should tell me I missing a str parameter at definition.

maybe the definition also can help a bit.

```py
    @wraps(view)
    return decorated
```

### Sandbox Link

_No response_

_No response_

Arths17:
This improves decorator-related diagnostics when a `Callable[Concatenate[..., ...]]` decorator injects a parameter into the wrapped function.

Before, the checker surfaced a low-level callable mismatch such as:

`Argument (self: Self@ExternalApiTemplateListApi) -> ... is not assignable to parameter view with type (@_, str, ParamSpec(@_)) -> @_ in function with_current_tenant_id`

After this change, the same case reports a function-level diagnostic that points to the wrapped function:

`Function \`get\` is missing required parameter of type \`str\` injected by decorator \`with_current_tenant_id\``

This keeps ParamSpec / Concatenate semantics unchanged and only improves the diagnostic layer for decorator application.

Fixes #3569

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -736,7 +736,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         .forall(tparams);
         ty = self.move_return_tparams_of_type(ty);
         for (x, range) in def.decorators.iter().rev() {
-            ty = self.apply_function_decorator(x.clone(), ty, &def.metadata, *range, errors);
+            ty = self.apply_function_decorator(
+                x.clone(),
+                ty,
+                &def.metadata,
+                &stmt.name,
+                *range,
+                errors,
+            );
         }
         Arc::new(ty)
     }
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/function.rs
```diff
@@ -1596,11 +1603,59 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            .visit_toplevel_func_metadata(&|meta| Some(meta.kind.function_name().to_string()))?;
+        let expected_decoratee = decorator.callable_first_param(self.heap)?;
+        let expected_signature = expected_decoratee
+            .callable_signatures()
+            .into_iter()
+            .find(|signature| {
+                matches!(&signature.params, Params::ParamSpec(prefix, _) if !prefix.is_empty())
+            })?;
+
+        let actual_signature = match decoratee {
```
rchen152, line 1622: The right way to get signatures from a type is [`callable_signatures`](https://github.com/facebook/pyrefly/blob/39c36d7869f9d5b3896863148a5c0fbcece82126/crates/pyrefly_types/src/types.rs#L1848-L1852), which also handles types with multiple signatures (e.g., overloaded functions) correctly.

pyrefly/lib/alt/function.rs
```diff
@@ -1596,11 +1603,59 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            return None;
+        }
+
+        let missing = prefix.get(actual_params.len())?;
+        let missing_ty = match missing {
+            PrefixParam::PosOnly(_, ty, Required::Required)
+            | PrefixParam::Pos(_, ty, Required::Required) => ty,
+            PrefixParam::PosOnly(_, _, Required::Optional(_))
+            | PrefixParam::Pos(_, _, Required::Optional(_)) => return None,
+        };
```
rchen152, line 1644: This portion of the code seems to be making a number of brittle assumptions:
* It fires based on the actual param count being one less than the expected. What if multiple params are missing? What if an expected argument is optional? (Pyrefly's internal representation is more expressive than type annotations, and `PrefixParam` has a `Required` field: https://github.com/facebook/pyrefly/blob/39c36d7869f9d5b3896863148a5c0fbcece82126/crates/pyrefly_types/src/callable.rs#L392-L395)
* It assumes the missing parameter is the last one in the list. That may be true thanks to the earlier length check and restrictions on what can be expressed via `Concatenate`, but correctness is not obvious and could be easily broken in the future.

IMO what would be more robust is somehow incorporating this check into the regular error logging flow, so it can fire based on the information we've already collected about what kind of call failure we've seen and which parameter it's associated with, rather than having a parallel check that has to guess.

pyrefly/lib/test/decorators.rs
```diff
@@ -419,6 +419,23 @@ g(f)
     "#,
 );
 
+testcase!(
```
rchen152, line 422: This test coverage is rather thin. I'd want to see what happens when things are concatenated in a different order and in various failure conditions to ensure that this check only fires when it's supposed to and always produces an appropriate error message.

rchen152 (changes requested):

rchen152:
I'm going to close this PR given my significant reservations about this approach, but please feel free to open another PR if you'd like to take another stab at this.

PR #3571, opened 2026-05-24
Fix if-branch pruning for Final bool constants

### LINKED ISSUE #3537: always `True` & always `False` variables are incorrectly treated as if they could be either
### Describe the Bug

```python
from typing import reveal_type, Final, Literal

asdf: Final = False
reveal_type(asdf) # revealed type: Literal[False]

if asdf:
    foo = 1
else:
    foo = 2

reveal_type(foo) # revealed type: Literal[1, 2] (wrong)

if False:
    bar = 1
else:
    bar = 2

reveal_type(bar) # revealed type: Literal[2] (correct)

```

Python: 3.14

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAGjoAxVmLkAZCAxiUxAHXS7UcTGEQKlUOgF4FYuDF2CRYycxgAKA0YCUdAMQDhorCYjC4mahpaUADa8jYwALq6uhBgdB7GunRZdGC4uJZ0AIy6MFC2iJnZuflWAExJ6A6BztKu1d5%2BTWIwwVIwYeqaYlGFcrXxdK4A7tTsng0p1mX9lVnYqJQFxeil5at065t1DV0Sfa6HHf6OQSHSAxHD45MAxryCLwzz6CAyIGSCMBQUiEBi0KAUMgMAAWuHQAFoRJQ4BA4QVtCAAMyEQoAFgxuiimmoyMSWBgxA%2BqA0wSsGOJvAxdGS3CgEBe6nhqGwcAYWk%2B8JeUAMcHRIAZlAJ6C4xDZHIY8JlvH4dPFlBJTJZsvZ6mB8MwMEgO0wXIYfIg2AArhoxRKpTQIHAUex4XBcJbKC8YLb1YyQLp0HD4TB8NDUJbeRARPCaNSXtCfRr-ehAwrBABHS0QQQmgDWMFIXJeXqdib9ulw0gRlvQxFQL1zrA4qrtIGZjR6NcwGAVLwMKuZaqT7ZzXZ7grhmHUqPQZclbd0NZ2JbQlELb24G0daJbvvn7aXMBXG0LfQn6CRKJ3g9bB479bD2Fgc816Brggf3NgMbj0MFBm9Xdh0XDs3SgIQvxgfUu00LkoHMIC-TvPoTSVPgXwXN831sVC2EDQQMPbCDKAgDAvRjR1YwYeNCOSC8xAgE0DTeLQwSOG89wxEAAF8-nrBgoxgRRYAoNAsDwIgyB4oA

_No response_

LeSingh1:
## Summary

Fixes #3537.

When a variable is declared `Final` and assigned a bool literal (`False` or `True`), control-flow tests like `if flag:` now prune unreachable branches the same way as literal `if False:` / `if True:`.

Before this change, both arms of an `if/else` could contribute to merged types even when the condition was statically known to be always false via `Final = False`.

## Test plan

- [x] Added `test_final_bool_unreachable_if_branch` — `Final = False` → else-only assignment type
- [x] Added `test_final_bool_unreachable_else_branch` — `Final = True` → if-only assignment type
- [x] `cargo test test_final_bool_unreachable`

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -172,6 +172,9 @@ pub struct Definitions {
     /// The `Option<String>` is `Some` when the RHS is a string literal (e.g. `X: Final = "x"`),
     /// used to resolve Final variable references in synthesized class field names.
     pub final_names: SmallMap<Name, Option<String>>,
+    /// Names marked `Final` with bool literal values, e.g. `FLAG: Final = False`.
+    /// Used to statically evaluate `if FLAG:` / `if not FLAG:` control flow.
+    pub final_bool_values: SmallMap<Name, bool>,
     /// Special exports defined in this module
     pub special_exports: SmallMap<Name, SpecialExport>,
     /// Names that are read (not just defined) in this scope.
@@ -699,6 +702,14 @@ impl DefinitionsBuilder {
                 } else {
                     None
                 };
+                let final_bool_value = if has_final_annotation {
+                    match x.value.as_deref() {
+                        Some(Expr::BooleanLiteral(b)) => Some(b.value),
+                        _ => None,
+                    }
+                } else {
+                    None
+                };
                 match &*x.target {
                     Expr::Name(x) => {
                         self.add_name(
@@ -713,6 +724,9 @@ impl DefinitionsBuilder {
                             self.inner
                                 .final_names
                                 .insert(x.id.clone(), final_string_value);
+                            if let Some(value) = final_bool_value {
+                                self.inner.final_bool_values.insert(x.id.clone(), value);
+                            }
                         }
                     }
                     _ => self.expr_lvalue(&x.target),
--- a/pyrefly/lib/test/narrow.rs
+++ b/pyrefly/lib/test/narrow.rs
@@ -108,6 +108,59 @@ def f(x: str | None):
 );
 
 testcase!(
+    test_final_bool_unreachable_if_branch,
+    r#"
+from typing import Final, Literal, reveal_type
+asdf: Final = False
+if asdf:
+    foo = 1
+else:
+    foo = 2
+reveal_type(foo)  # E: revealed type: Literal[2]
+"#,
+);
+
+testcase!(
+    test_final_bool_unreachable_else_branch,
+    r#"
+from typing import Final, Literal, reveal_type
+flag: Final = True
+if flag:
+    bar = 1
+else:
+    bar = 2
+reveal_type(bar)  # E: revealed type: Literal[1]
+"#,
+);
+
+// Regression test: when a function's trailing statement is an `if` gated by a
+// `Final` bool constant, the implicit-return scan (`function_last_expressions`)
+// must prune the same branch the binding traversal prunes. Otherwise it records
+// a `LastStmt` inside the pruned branch whose binding is never created, and we
+// panic at solve time with "key lacking binding".
+testcase!(
+    test_final_bool_if_branch_implicit_return_no_panic,
+    r#"
+from typing import Final
+FLAG_FALSE: Final = False
+FLAG_TRUE: Final = True
+def prune_if() -> None:
+    if FLAG_FALSE:
+        print("pruned")
+    else:
+        print("kept")
+def prune_else() -> None:
+    if FLAG_TRUE:
+        print("kept")
+    else:
+        print("pruned")
+def prune_if_no_else() -> None:
+    if FLAG_FALSE:
+        print("pruned and last")
+"#,
+);
+
+testcase!(
     test_is_subtype,
     r#"
 from typing import assert_type
--- a/pyrefly/lib/test/narrow.rs
+++ b/pyrefly/lib/test/narrow.rs
@@ -160,6 +160,63 @@ def prune_if_no_else() -> None:
 "#,
 );
 
+// Same regression, but for `Final` bools declared inside the function. These are
+// only visible once `init_static_scope` has populated the function scope, so the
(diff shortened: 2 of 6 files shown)


Arths17:
The idea here makes sense, but the CI failures suggest this change may be leaking beyond Final-declared constants into general boolean control-flow inference.

I’d double-check the branch pruning condition — it should ideally only trigger on Final-backed literals, not inferred constant truthiness.

Might be worth adding a regression test for:

non-Final bool literal assignment
mixed union bool narrowing in if/else joins

to ensure no accidental over-pruning.

pyrefly/lib/test/narrow.rs
```diff
@@ -107,6 +107,59 @@ def f(x: str | None):
...
+    if FLAG_TRUE:
+        print("kept")
+    else:
+        print("pruned")
+def prune_if_no_else() -> None:
+    if FLAG_FALSE:
+        print("pruned and last")
+"#,
+);
+
```
connernilsen, line 162: Can we add a test to validate that a closer, non-final bool definition doesn't prune branches too early?

Probably something like this:

```python
from typing import Final

flag: Final = True
# this should hit that early return case where a local scope might override a more global scope
def foo(flag: bool):
  if flag:
      bar = 1
  else:
      bar = 2
  reveal_type(bar)  # E: revealed type: Literal[1] | Literal[2] (I'm not sure if this would be literal or collapse to just `int`)
```

pyrefly/lib/test/narrow.rs
```diff
@@ -107,6 +107,59 @@ def f(x: str | None):
...
+def prune_else() -> None:
+    if FLAG_TRUE:
+        print("kept")
+    else:
+        print("pruned")
+def prune_if_no_else() -> None:
+    if FLAG_FALSE:
+        print("pruned and last")
+"#,
+);
```
connernilsen, line 161: Add one more check in this test for function-level finals, since those might crash at runtime due to the branch being dropped

```python3
def prune_if_no_else_local() -> None:
  FLAG_FALSE: Final = False
  if FLAG_FALSE
    print("pruned and last")
def prune_else_local() -> None:
  FLAG_FALSE: Final = False
  if FLAG_FALSE
    print("pruned")
  else:
    print("kept")
def prune_if_local() -> None:
  FLAG_TRUE: Final = False
  if FLAG_TRUE
    print("kept")
  else:
    print("pruned")
```

You may need to fix this by computing `implicit_return` after the function scope's static state is initialized (inside or after `init_static_scope`) so both `function_last_expressions -> evaluate_bool_for_control_flow` and `stmt.rs:1265` see the same `final_bool_values`.

pyrefly/lib/export/definitions.rs
```diff
@@ -172,6 +172,9 @@ pub struct Definitions {
     /// The `Option<String>` is `Some` when the RHS is a string literal (e.g. `X: Final = "x"`),
     /// used to resolve Final variable references in synthesized class field names.
     pub final_names: SmallMap<Name, Option<String>>,
+    /// Names marked `Final` with bool literal values, e.g. `FLAG: Final = False`.
+    /// Used to statically evaluate `if FLAG:` / `if not FLAG:` control flow.
```
connernilsen, line 176: You might need to fix the evaluator here (or add a test validating that it does this). I don't think the current logic can handle `not FLAG`, since we only special case a bare `Expr::Name` below.

connernilsen (changes requested):
Hey @LeSingh1, thanks for working on this! Here are a few questions/recommendations real quick

LeSingh1:
Thanks for the review @connernilsen — all three addressed in 56f4645.

**Function-level finals did crash**, exactly as you suspected. Your snippet reproduces it:

```
Internal error: key lacking binding, module=main, key=Key::StmtExpr(...)
```

Your diagnosis was right down to the mechanism. `implicit_return` already prunes through the scopes, but it was being called *before* `function_body_scope` pushed the function scope, so it could only ever see enclosing scopes — a `Final` declared in the function body was invisible to it while `stmts` pruned on it. I moved the call inside `function_body_scope`, after `init_static_scope`, so both see the same `final_bool_values`, and recorded the ordering as an invariant on `implicit_return` so it doesn't get reintroduced.

**`not FLAG` was also unhandled** — you were right that the bare `Expr::Name` case wasn't enough. `SysInfo::evaluate_bool` only negates tests it can evaluate itself, so the negation now wraps the scope lookup:

```rust
Expr::UnaryOp(ExprUnaryOp { op: UnaryOp::Not, operand, .. }) =>
    Some(!self.evaluate_bool_for_control_flow(sys_info, operand)?),
```

Recursing rather than matching one level also covers nested negation.

**The closer-binding case already behaved correctly** — `lookup_final_bool_value` bails as soon as a scope's `stat` contains the name, so the parameter stops the walk before reaching the module-level `Final`. Added it as a regression test since nothing was pinning that behavior; it reveals `Literal[1, 2]` as you'd expect.

Tests added to `narrow.rs`: `test_final_bool_function_scoped_implicit_return_no_panic`, `test_final_bool_not_operator`, `test_final_bool_shadowed_by_closer_binding`. The first two fail on the previous commit (panic and wrong reveal respectively) and pass now.

Full suite is green locally — 6376 lib + 311 LSP + 18 laziness, plus `test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` clean with no new lints.

One note: this is branched off the same base as before, so it'll need a rebase onto current main before it can land — happy to do that whenever you'd like, just say the word.

PR #3572, opened 2026-05-24
fix: skip unannotated-return check for @no_type_check functions

### LINKED ISSUE #3561: false positive `unannotated-return` for `@no_type_check` function
### Describe the Bug

```python
from typing import no_type_check

@no_type_check
def f() -> None: ...  # E: missing a return annotation [unannotated-return]
```

removing `-> None` makes no difference

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN3pcAfSbEYQgMYALGBIDWAHXRKAAoJHNx02YqwwwdMAAoAlHQC0APjoA5XOhiI6hFyAA0IMpX1RShBrRQFADaMJTUlHAAukoArugYggyoDDCY5t4MsZTodAC8dAogYRFFIAC%2BHqgSDBAAbjAAYtAwFGhYeERkFUA

_No response_

Arths17:
Before
`@no_type_check` functions could still trigger `unannotated-return`.

After
Functions marked with `@no_type_check` are excluded from return-annotation validation.

Guarantee
Normal function behavior is unchanged.
This only affects the decorated skip logic for `@no_type_check` functions.

Fixes #3561

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -607,6 +607,8 @@ pub struct FuncFlags {
     pub is_overload: bool,
     pub is_staticmethod: bool,
     pub is_classmethod: bool,
+    /// A function decorated with `@no_type_check`.
+    pub has_no_type_check: bool,
     /// A function decorated with `@deprecated`
     pub deprecation: Option<Deprecation>,
     /// Metadata for `@property`, `@foo.setter`, and `@foo.deleter`.
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -453,6 +453,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let mut flags = FuncFlags {
             is_staticmethod: is_dunder_new,
             is_classmethod: is_dunder_init_subclass,
+            has_no_type_check: false,
             is_async: def.is_async,
             placeholder_body_kind,
             is_return_inferred,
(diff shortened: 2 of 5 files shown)


yangdanny97 (changes requested):
Can you clarify the changes in this PR? They seem much larger than I would expect is needed for implementing this feature.

I expected `def.metadata.flags.has_no_type_check` to be saved in some variable at the top, and then a check to that variable to be added to the gate of the unannotated-return, like so

```
        if !has_return_annotation && !no_type_check {
            self.error(
                errors,
                stmt.name.range(),
                ErrorKind::UnannotatedReturn,
                format!("`{}` is missing a return annotation", stmt.name),
            );
        }
```

I'm not sure why a lot of other logic was changed in function.rs

Arths17:
You’re right. I trimmed this PR down to the minimal @no_type_check fix: it now only skips the UnannotatedReturn check, and I removed the unrelated decorator changes. The branch has been updated.

pyrefly/lib/alt/function.rs
```diff
@@ -753,6 +765,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             return Some(SpecialDecorator::PropertyDeleter(&decorator.ty));
         }
         match decorator.ty.callee_kind() {
+            Some(CalleeKind::Function(FunctionKind::Def(func_id)))
+                if func_id.name.as_str() == "no_type_check"
```
yangdanny97, line 769: Probably should add a FunctionKind for this instead of hardcoded strings

pyrefly/lib/alt/function.rs
```diff
@@ -1596,11 +1621,59 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    fn decorator_missing_injected_parameter_message(
```
yangdanny97, line 1624: This is an unrelated change

pyrefly/lib/alt/function.rs
```diff
@@ -736,7 +741,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         .forall(tparams);
         ty = self.move_return_tparams_of_type(ty);
         for (x, range) in def.decorators.iter().rev() {
-            ty = self.apply_function_decorator(x.clone(), ty, &def.metadata, *range, errors);
+            ty = self.apply_function_decorator(
+                x.clone(),
+                ty,
+                &def.metadata,
+                &stmt.name,
```
yangdanny97, line 748: this is an unrelated change

pyrefly/lib/alt/function.rs
```diff
@@ -568,113 +569,117 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-        for (i, p) in stmt.parameters.iter().enumerate() {
-            // Skip first param if it's implicit self/cls and not variadic
-            if i == 0 && has_implicit_self_or_cls_param && !p.is_variadic() {
-                continue;
-            }
-            if p.annotation().is_none() {
-                let name = p.name().as_str();
+        let contains_self_type = |ty: &Type| ty.any(|t| matches!(t, Type::SelfType(_)));
+
+        if !def.metadata.flags.has_no_type_check {
```
yangdanny97, line 577: Copilot is right, gating the entire block of code behind the flag means that a lot of other things are skipped, which is not the scope of the original task.

I'd prefer for this to be saved in a variable, and checked before emitting each error, like in the code snippet I gave as an example.

yangdanny97 (changes requested):
There are still unrelated changes in this PR, and I left additional feedback about code structure

crates/pyrefly_types/src/callable.rs
```diff
@@ -607,6 +607,8 @@ pub struct FuncFlags {
     pub is_overload: bool,
     pub is_staticmethod: bool,
     pub is_classmethod: bool,
+    /// A function decorated with `@no_type_check`.
+    pub has_no_type_check: bool,
```
yangdanny97, line 611: maybe easier to just call it `no_type_check`

pyrefly/lib/test/untyped_def_behaviors.rs
```diff
@@ -492,6 +492,17 @@ assert_type(f(0), Any)
...
 
+testcase!(
+    test_no_type_check_without_return_annotation,
+    r#"
+from typing import no_type_check
+
+@no_type_check
+def f():
+    pass
+"#,
```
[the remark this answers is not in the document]
yangdanny97: This comment should be addressed

pyrefly/lib/test/decorators.rs
```diff
@@ -419,6 +419,23 @@ g(f)
     "#,
 );
 
+testcase!(
+    test_decorator_missing_injected_parameter,
```
yangdanny97, line 423: This is completely unrelated

pyrefly/lib/alt/function.rs
```diff
@@ -1610,6 +1683,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         {
             return decoratee;
         }
+        let decorator_for_message = decorator.clone();
```
yangdanny97, line 1686: This and below are all unrelated changes

yangdanny97:
closing because the PR is stale

PR #3574, opened 2026-05-25
feat: suppress implicit-any-parameter for underscore-prefixed params

### LINKED ISSUE #3542: support underscore variables starting with (`_`) to indicate intentionally unused variables
it's common to use a single `_` to indicate an unused argument:

```py
def foo(_): # error: `foo` is missing an annotation for parameter `_` [implicit-any-parameter]
    print(_) # should be an error here instead, because it should be unused
```

MD-Mushfiqur123:
## Summary

Parameters starting with a single underscore that do not end with underscore are conventionally used to indicate intentionally unused variables. This PR suppresses the `implicit-any-parameter` error for such parameters, reducing noise for a common Python pattern.

Fixes #3542

## Changes

- `pyrefly/lib/alt/function.rs`: Skip emitting `ImplicitAnyParameter` for parameters where the name starts with exactly one underscore and does not end with underscore (e.g. `_`, `_x`)
- `pyrefly/lib/test/inference.rs`: Add test cases covering all underscore patterns

## Underscore patterns

| Pattern | Behavior |
|---------|----------|
| `_` | No error (intentionally unused) |
| `_abc` | No error (intentionally unused) |
| `__abc` | Error (dunder prefix) |
| `_abc_` | Error (sunder, ends with underscore) |
| `__abc__` | Error (dunder) |
| `x` | Error (no underscore prefix) |

## CLA

I will submit the CLA if needed.

_🤖 This PR was created with AI assistance_

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -593,17 +593,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
             if p.annotation().is_none() {
                 let name = p.name().as_str();
-                // Skip parameters starting with single `_` that don't end with `_`
-                // as they indicate intentionally unused variables.
+                // Skip parameters that indicate intentionally unused variables.
                 // `_`, `_abc` → skip; `__abc`, `_abc_`, `__abc__` → don't skip.
-                let is_underscore_unused = {
-                    let b = name.as_bytes();
-                    b.len() >= 1
-                        && b[0] == b'_'
-                        && (b.len() == 1 || b[1] != b'_')
-                        && !name.ends_with('_')
-                };
-                if !is_underscore_unused {
+                if !Ast::is_intentionally_unused(&name) {
                     self.error(
                         errors,
                         p.name().range(),
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -593,15 +593,27 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
             if p.annotation().is_none() {
                 let name = p.name().as_str();
-                self.error(
-                    errors,
-                    p.name().range(),
-                    ErrorKind::ImplicitAnyParameter,
-                    format!(
-                        "`{}` is missing an annotation for parameter `{name}`",
-                        stmt.name
-                    ),
-                );
+                // Skip parameters starting with single `_` that don't end with `_`
+                // as they indicate intentionally unused variables.
+                // `_`, `_abc` → skip; `__abc`, `_abc_`, `__abc__` → don't skip.
+                let is_underscore_unused = {
+                    let b = name.as_bytes();
+                    b.len() >= 1
+                        && b[0] == b'_'
+                        && (b.len() == 1 || b[1] != b'_')
+                        && !name.ends_with('_')
+                };
+                if !is_underscore_unused {
+                    self.error(
+                        errors,
+                        p.name().range(),
+                        ErrorKind::ImplicitAnyParameter,
+                        format!(
+                            "`{}` is missing an annotation for parameter `{name}`",
+                            stmt.name
+                        ),
+                    );
+                }
             }
         }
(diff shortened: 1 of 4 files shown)


pyrefly/lib/alt/function.rs
```diff
@@ -593,15 +593,27 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-                    format!(
-                        "`{}` is missing an annotation for parameter `{name}`",
-                        stmt.name
-                    ),
-                );
+                // Skip parameters starting with single `_` that don't end with `_`
+                // as they indicate intentionally unused variables.
+                // `_`, `_abc` → skip; `__abc`, `_abc_`, `__abc__` → don't skip.
+                let is_underscore_unused = {
+                    let b = name.as_bytes();
```
yangdanny97: `name.starts_with("_") && (name.len() == 1 || !name.ends_with("_"))` is easier IMO, none of the other dunder/sunder checking in Pyrefly converts to bytes so I don't think we should break that pattern

yangdanny97 (changes requested):
Back to you with one comment, once that's addressed we can merge

pyrefly/lib/alt/function.rs
yangdanny97: also, can we refactor this to be a helper inside 

https://github.com/facebook/pyrefly/blob/main/crates/pyrefly_python/src/ast.rs#L414

with a function name like `maybe_intentionally_unused` that way we can reuse the check in the LSP?

yangdanny97:
Oh one more thing. We should avoid emitting unused-variable and unused-parameter diagnostics for these names in the language server as well

https://github.com/facebook/pyrefly/blob/ef0fcfe90900bd56521bbd49416c062393004235/pyrefly/lib/lsp/non_wasm/server.rs#L5250

https://github.com/facebook/pyrefly/blob/ef0fcfe90900bd56521bbd49416c062393004235/pyrefly/lib/lsp/non_wasm/server.rs#L5298

You can reuse the helper function for that

MD-Mushfiqur123:
Friendly ping — this PR is ready for review. All CI checks are passing.

MD-Mushfiqur123:
The LSP suppression has been implemented in the latest commit (suppress unused-parameter and unused-variable diagnostics for underscore-prefixed names). Could you please re-review?

yangdanny97 (changes requested):
You didn't address my comments in function.rs. The logic in server.rs is also inconsistent with function.rs - it should be the same, which is why I suggested you make a helper function.

MD-Mushfiqur123:
@yangdanny97 friendly ping — changes have been pushed addressing the LSP suppression feedback. Could you take another look?

yangdanny97:
@MD-Mushfiqur123 I took another look - you didn't address all the comments from my initial review

MD-Mushfiqur123:
@yangdanny97 All review comments addressed:

1. **function.rs**: Check simplified to `Ast::is_intentionally_unused()` — uses `name.starts_with(_) && (name.len() == 1 || !name.ends_with(_))` as you suggested
2. **ast.rs**: Added the helper function `is_intentionally_unused` next to `is_mangled_attr`
3. **server.rs**: Both `unused_parameters()` and `unused_variables()` loops now use the same helper, matching function.rs logic

Could you take another look?

pyrefly/lib/alt/function.rs
```diff
@@ -593,15 +593,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-                    errors,
-                    p.name().range(),
-                    ErrorKind::ImplicitAnyParameter,
-                    format!(
-                        "`{}` is missing an annotation for parameter `{name}`",
-                        stmt.name
-                    ),
-                );
+                // Skip parameters that indicate intentionally unused variables.
+                // `_`, `_abc` → skip; `__abc`, `_abc_`, `__abc__` → don't skip.
```
yangdanny97, line 597: some form of this should be the docstring of `is_intentionally_unused`

stroxler (approved):

Issue #3562, opened 2026-05-24
Update typeshed


yangdanny97:
### Describe the Bug

pull in PEP 661 changes

PR #3575, opened 2026-05-25
Add Sentinel support (PEP-661)

### LINKED ISSUE #3207: Support Sentinel types in pydantic models
### Describe the Bug

```
import pydantic
from pydantic.experimental.missing_sentinel import MISSING

class UpdateUserPayload(pydantic.BaseModel):
    phone_number: str | MISSING = MISSING 
    # Errors with "Expected a type form, got instance of `type[str] | UnionType`". Should not error.
```

This is an experimental feature, but required if you have a `PATCH` model and the generated schema shouldn't erroneously indicate that the value is nullable.

https://pydantic.dev/docs/validation/latest/concepts/experimental/#missing-sentinel

### Sandbox Link

_No response_

_No response_

hrolfurgylfa:
# Summary

This PR adds support for PEP 661 Sentinel types.

Fixes #3207

# Test Plan

I've run this as an LSP for a bit on a bigger private repo and it seems to work, but the main testing was the tests added to `test/narrow.rs` and `test/sentinel.rs`.

# Some assumptions/things I didn't fully understand and might deserve more attention:

- I'm using NestingContext::toplevel() when creating a Sentinel struct, this is similar to TypeVar/TypeAlias/a few others, I left the same TODO in place as they had. I looked through some of the other types and couldn't find any other way it was being done, so I left it like this for now, but if anyone has any pointers to how it should be done better, please let me know and I can fix it.

- For the cinderx support, I added the sentinel to it in the TODO section of variables that get set to Any. I haven't really diven into what cinderx is and how the support for it is implemented, but if it is important to get it working I can go ahead and look into that.

- I'm using bind_legacy_type_var_or_typing_alias to bind the Sentinel type, I think this should be ok as the PEP is pretty strict in how it can be assigned. I've done some testing and only found one edge case so far that fails because of this, but I don't think Final's should be mixed with Sentinel objects like this anyway, here is the test case that fails:

```py
from typing import Any, Final
from typing_extensions import Sentinel

MISSING: Final[Any]
MISSING = Sentinel("MISSING")
```

It fails because the `Final` doesn't notice the assignment, it still thinks the variable was never assigned to.

# Status

This PR is mostly ready for review, but I'm keeping it in Draft form until #3562 is addressed, since until then I can't finalize Python 3.15 support.

### DIFF
--- a/crates/pyrefly_types/src/sentinel.rs
+++ b/crates/pyrefly_types/src/sentinel.rs
@@ -24,6 +24,21 @@ use crate::equality::TypeEqCtx;
 use crate::heap::TypeHeap;
 use crate::types::Type;
 
+#[derive(Clone, Copy, Dupe, Debug, PartialEq, Eq, Hash, Ord, PartialOrd, TypeEq)]
+pub enum SentinelKind {
+    Builtins,
+    TypingExtensions,
+}
+
+impl SentinelKind {
+    pub fn name(&self) -> &str {
+        match self {
+            Self::Builtins => "sentinel",
+            Self::TypingExtensions => "Sentinel",
+        }
+    }
+}
+
 /// Used to represent Sentinel calls. Each Sentinel is unique, so use the ArcId to separate them.
 #[derive(Clone, Dupe, Debug, PartialEq, Eq, Hash, Ord, PartialOrd)]
 pub struct Sentinel(ArcId<SentinelInner>);
@@ -48,14 +63,20 @@ impl Display for Sentinel {
 #[derive(Debug, PartialEq, TypeEq, Eq, Ord, PartialOrd)]
 struct SentinelInner {
     qname: QName,
+    kind: SentinelKind,
 }
 
 impl Sentinel {
-    pub fn new(name: Identifier, module: Module) -> Self {
+    pub fn new(name: Identifier, module: Module, kind: SentinelKind) -> Self {
         Self(ArcId::new(SentinelInner {
+            kind,
             // TODO: properly take parent from caller of new()
             qname: QName::new(name, NestingContext::toplevel(), module),
         }))
+    }
+
+    pub fn kind(&self) -> SentinelKind {
+        self.0.kind
     }
 
     pub fn qname(&self) -> &QName {
--- a/crates/pyrefly_types/src/sentinel.rs
+++ b/crates/pyrefly_types/src/sentinel.rs
@@ -67,11 +67,15 @@ struct SentinelInner {
 }
 
 impl Sentinel {
-    pub fn new(name: Identifier, module: Module, kind: SentinelKind) -> Self {
+    pub fn new(
+        name: Identifier,
+        nesting_context: NestingContext,
+        module: Module,
+        kind: SentinelKind,
+    ) -> Self {
         Self(ArcId::new(SentinelInner {
             kind,
-            // TODO: properly take parent from caller of new()
-            qname: QName::new(name, NestingContext::toplevel(), module),
+            qname: QName::new(name, nesting_context, module),
         }))
     }
 
--- a/crates/pyrefly_types/src/sentinel.rs
+++ b/crates/pyrefly_types/src/sentinel.rs
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+use std::fmt;
+use std::fmt::Display;
+use std::hash::Hash;
+
+use dupe::Dupe;
+use pyrefly_derive::TypeEq;
+use pyrefly_python::module::Module;
+use pyrefly_python::nesting_context::NestingContext;
+use pyrefly_python::qname::QName;
+use pyrefly_util::arc_id::ArcId;
+use pyrefly_util::visit::Visit;
+use pyrefly_util::visit::VisitMut;
+use ruff_python_ast::Identifier;
+
+use crate::equality::TypeEq;
+use crate::equality::TypeEqCtx;
+use crate::heap::TypeHeap;
+use crate::types::Type;
+
+/// Used to represent Sentinel calls. Each Sentinel is unique, so use the ArcId to separate them.
+#[derive(Clone, Dupe, Debug, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct Sentinel(ArcId<SentinelInner>);
+
+// This is a lie, we do have types in the bound position
+impl Visit<Type> for Sentinel {
+    const RECURSE_CONTAINS: bool = false;
+    fn recurse<'a>(&'a self, _: &mut dyn FnMut(&'a Type)) {}
+}
+
+impl VisitMut<Type> for Sentinel {
+    const RECURSE_CONTAINS: bool = false;
+    fn recurse_mut(&mut self, _: &mut dyn FnMut(&mut Type)) {}
+}
+
+impl Display for Sentinel {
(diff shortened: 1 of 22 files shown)


yangdanny97:
Thanks for the PR, wow! that was fast, the PEP just got accepted recently and the conformance tests haven't even been added yet

I see you added lots of tests already, but maybe worth looking at the proposed conformance test to see if there's anything missing
https://github.com/python/typing/pull/2277/changes

I can give this a more detailed review once I've unblocked you on https://github.com/facebook/pyrefly/issues/3562

hrolfurgylfa:
Thanks, I didn't know those conformance tests existed, I'll review those once I get back home after the weekend, don't have my laptop with me.

And yeah, a proper review should probably be held until I can add the Python 3.15 support, since there is a bit of extra code to differentiate sentinel lowercase builtin and Sentinel uppercase from typing extensions, might be adding another 30-50 lines, don't remember the exact number though.

I'm hoping to have both of those ready on Tuesday/Wednesday evening (UTC+0).

yangdanny97:
No rush at all. The typeshed update just got merged, so once you rebase it should be there

hrolfurgylfa:
This PR should now be ready for review.

It now passes the [conformance tests](https://github.com/python/typing/pull/2277/changes), the only change was just no longer emit an error when a sentinel is assigned to a variable with a different name like `MISSING = sentinel("<MISSING>")`. This is something the PEP prohibited but in the [open conformance PR to python/typing](https://github.com/python/typing/pull/2277) that prohibition was lifted, so type checkers are no longer supposed to emit an error there.

I'm still not perfectly sure on the query interface and tsp implementation for sentinel, I haven't really figured out how to test the tsp stuff properly and while I have looked at the output of the shape query with sentinels I'm not perfectly sure how it should look, as such I haven't added any automated tests for it.

crates/pyrefly_types/src/stdlib.rs
```diff
@@ -257,6 +263,9 @@ impl Stdlib {
             param_spec_kwargs: lookup_concrete(standardised(3, 10), "ParamSpecKwargs"),
             type_var_tuple: lookup_concrete(standardised(3, 11), "TypeVarTuple"),
             type_alias_type: lookup_concrete(standardised(3, 12), "TypeAliasType"),
+            // sentinel: lookup_concrete(typing_extensions, "Sentinel"),
+            sentinel_builtin: lookup_concrete(builtins, "sentinel"),
```
rchen152: How different are `builtins.sentinel` and `typing_extensions.Sentinel`? If they're pretty similar, I'd suggest combining them into just one `sentinel`, which is looked up from builtins in 3.15+ and from typing_extensions otherwise, and modifying the rest of the PR to likewise stop tracking the distinction between the two.

In general, we treat the `typing` and `typing_extensions` versions of things as the same without worrying about the minor differences between their runtime definitions. And distinguishing between `builtins.sentinel` and `typing_extensions.Sentinel` seems to be adding a fair bit of complexity to this PR.

crates/pyrefly_types/src/sentinel.rs
```diff
@@ -0,0 +1,93 @@
...
+struct SentinelInner {
+    qname: QName,
+    kind: SentinelKind,
+}
+
+impl Sentinel {
+    pub fn new(name: Identifier, module: Module, kind: SentinelKind) -> Self {
+        Self(ArcId::new(SentinelInner {
+            kind,
+            // TODO: properly take parent from caller of new()
```
rchen152: You can grab the nesting context from the Scopes stack when constructing Binding::Sentinel - see comment in stmt.rs.

pyrefly/lib/binding/stmt.rs
```diff
@@ -281,6 +282,36 @@ impl<'a> BindingsBuilder<'a> {
...
+            self.ensure_expr(expr, static_type_usage);
+        }
+        for kw in call.arguments.keywords.iter_mut() {
+            self.ensure_expr(&mut kw.value, static_type_usage);
+        }
+        // Like legacy type var, Sentinel can only be created with a single Sentinel binding to a
+        // single variable (https://peps.python.org/pep-0661/#typing). Thus we bind it in the same
+        // way legacy type vars are bound.
+        self.bind_legacy_type_var_or_typing_alias(name, |ann| {
+            Binding::Sentinel(Box::new((
```
rchen152, line 301: Here you should have access to `self.scopes`, from which you can get the current nesting context: https://github.com/facebook/pyrefly/blob/4d820b2eb582b4c50de23f6e2d9998427d59e6a0/pyrefly/lib/binding/scope.rs#L1513.

website/docs/error-kinds.mdx
```diff
@@ -885,6 +885,20 @@ class TD(TypedDict):
...
 ```
 
+## invalid-sentinel
+
+An error caused by incorrect definition of a Sentinel. A few examples:
+
+```python
+from typing_extensions import Sentinel
+
+# Sentinels must be assigned to a matching variable.
```
rchen152: This doc example looks incorrect; the implementation (correctly) allows the name to not match the variable name.

rchen152 (changes requested):
Thank you for the PR! This is great. I left a few comments, mostly minor. The one larger question I have is whether we can drop the distinction between `builtins.sentinel` and `typing_extensions.Sentinel`, which seems like it would simplify the code a fair bit if we can get away with it.

crates/pyrefly_types/src/stdlib.rs
hrolfurgylfa: The main reason I differentiated between the two was the `__name__` attribute that is only available on the builtins sentinel, not the typing extensions Sentinel. If we merge them it would mean that in Python 3.15+ pyrefly would say that accessing `__name__` on typing_extensions Sentinel is allowed and returns a string, while raising a runtime error if run. If that is fine I can go ahead and remove this logic.

Here is how I confirmed this difference if you want to test it with the python 3.15 beta:
```bash
uv run -p 3.15 --with 'typing_extensions' python -c 'from typing_extensions import Sentinel; MISSING = Sentinel("MISSING"); print(MISSING.__name__)'
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>     from typing_extensions import Sentinel; MISSING = Sentinel("MISSING"); MISSING.__name__
>                                                                            ^^^^^^^^^^^^^^^^
> AttributeError: 'Sentinel' object has no attribute '__name__'. Did you mean '.__ne__' instead of '.__name__'?

uv run -p 3.15 --with 'typing_extensions' python -c 'MISSING = sentinel("MISSING"); print(MISSING.__name__)'
> MISSING
```

Also, if this ends up being kept, I should probably add a comment explaining the reasoning better, and remove that commented out line, not sure how that one got past me.

crates/pyrefly_types/src/sentinel.rs
hrolfurgylfa: Thanks, fixed

website/docs/error-kinds.mdx
hrolfurgylfa: Good catch, forgot to change this when I removed the error on different variable name, fixed

crates/pyrefly_types/src/stdlib.rs
rchen152: Thanks for the explanation! If the presence/absence of the `__name__` attribute is the main difference, then I'm in favor of merging them.

rchen152 (approved):
Looks good, thank you! I'll get a second review and get this merged.

hrolfurgylfa:
Thanks :smile:

yangdanny97 (approved):

yangdanny97:
Great work on this PR @hrolfurgylfa, thanks!

PR #3580, opened 2026-05-25
Include private names from `__all__` in `pyrefly coverage report`

### LINKED ISSUE #3578: `pyrefly report` should include private names listed in  `__all__`
Using a leading `_` to indicate a private name is but a convention, not a hard rule. So if a project explicitly lists such a "private" name in the `__all__`, then I think it's reasonable to assume that it's intended to be part of the public API. Because of that, I think it should then also be treated as such in  pyrefly coverage reports.

I don't have a concrete example of this, but I saw in the code that "private" names are currently always skipped.

jorenham:
# Summary

Fixes #3578

Merge conflicts with #3578 are unavoidable, but I tried to avoid them from getting to big (didn't check though). 
I don't have a particular preference for merge order.

# Test Plan

Served with integration- and unit tests.

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -608,11 +608,12 @@ impl ReportArgs {
         module.as_str().split('.').all(Self::is_public_name)
     }
 
-    /// True if `fqn` or a class-level prefix is public, and every segment below the module is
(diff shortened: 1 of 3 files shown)


kinto0 (approved):

PR #3586, opened 2026-05-26
stubgen: emit synthesized `__init__` for dataclasses (#3586)

### LINKED ISSUE #3221: bug(stubgen): dataclass fields with only `field(...)` / non-literal defaults lose `= ...` in generated stubs, breaking `__init__` checking

yangdanny97:
Summary:

closes https://github.com/facebook/pyrefly/issues/3221

Dataclasses (and similar patterns like NamedTuples) generate their
`__init__` at type-check time rather than declaring it in source.
Previously, stubgen would omit the `__init__` entirely for these
classes, producing stubs that lost the constructor signature.

Look up `KeyClassSynthesizedFields` in the answers to find the
synthesized `__init__` type (a `Type::Function` with the full
`Callable` signature), then convert its `Param` list into stub
parameters. Only emitted when no explicit `__init__` exists in the
source.

Differential Revision: D106327984

### DIFF
--- a/pyrefly/lib/stubgen/extract.rs
+++ b/pyrefly/lib/stubgen/extract.rs
@@ -19,6 +19,8 @@ use pyrefly_python::module::Module;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_python::sys_info::SysInfo;
 use pyrefly_types::callable::Param;
+use pyrefly_types::callable::Params;
+use pyrefly_types::callable::Required;
 use pyrefly_types::display::TypeDisplayContext;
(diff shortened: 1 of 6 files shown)


rchen152 (approved):

PR #3592, opened 2026-05-27
Make `is_subset_overload` revert specialization errors

rchen152:
Summary: Specialization errors from overload branches we didn't select should be reverted.

Differential Revision: D106459333


yangdanny97 (approved):

PR #3594, opened 2026-05-27
fix(lsp): display ** prefix for ParamSpec in hover type parameter lists

### LINKED ISSUE #3568: paramspec parameter don't show difference in hover
### Describe the Bug

```py
    @wraps(view)
    return decorated
```

hover shows

```
(function) with_current_tenant_id: def with_current_tenant_id[T, P, R](view: (T, str, ParamSpec(P)) -> R) -> (T, ParamSpec(P)) -> R: ...
```

`[T, **P, R]` becomes `[T, P, R]`

### Sandbox Link

_No response_

_No response_

aadi-novice:
## Summary

When hovering over a generic function that uses ParamSpec (e.g., **P), the type parameter list in the hover was showing [T, P, R] instead of [T, **P, R] - the ** prefix was missing.

## Changes

- **quantified.rs**: Added `**` prefix in `Quantified::display_with_bounds()` for ParamSpec type parameters, matching Python syntax where ParamSpec is declared as `**P`.
- **display.rs**: Added test `test_display_param_spec` to verify the fix works in both default and LSP hover display modes.

## Testing

Added a unit test that creates a generic bound method with type parameters [T, **P, R] and verifies:
- Default display: [T, **P, R](self: Any, x: Any, y: Any) -> None
- LSP hover display: def foo[T, **P, R](...) -> None: ...

Fixes #3568

### DIFF
--- a/crates/pyrefly_types/src/display.rs
+++ b/crates/pyrefly_types/src/display.rs
@@ -2103,6 +2103,34 @@ pub mod tests {
         );
     }
 
+n    #[test]
+    fn test_display_param_spec() {
+        let tparams = fake_tparams(vec![
+            fake_tparam(0, "T", QuantifiedKind::TypeVar),
+            fake_tparam(1, "P", QuantifiedKind::ParamSpec),
+            fake_tparam(2, "R", QuantifiedKind::TypeVar),
+        ]);
+        let method = fake_generic_bound_method(
+            "foo",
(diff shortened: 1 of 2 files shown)


kinto0 (approved):
looks great! thank you!!

yangdanny97 (approved):

PR #3615, opened 2026-05-30
Prepare semantic tokens to include syntax highlighting

magic-akari:
# Summary

Pyrefly's semantic pipeline currently models only identifier-level tokens. Future support for a full semantic stream therefore requires a syntax-token source covering keywords, comments, strings, numbers, and operators.

This change introduces a lightweight syntax path derived from Ruff's parse results and configures the LSP encoder to split multiline ranges into line-local tokens. To ensure the new path remains unadvertised, this server capability is gated to clients that already augment syntax tokens.

This enables proper representation of triple-quoted strings and preserves current semantic-overlay behavior during the gated rollout.

# Test Plan

- `cargo test`

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -1191,17 +1191,21 @@ pub fn dispatch_lsp_events(server: &Server, reader: &mut MessageReader) {
     let _ = server.lsp_queue().send(LspEvent::Exit);
 }
 
-pub fn capabilities(
-    indexing_mode: IndexingMode,
-    initialization_params: &InitializeParams,
-) -> ServerCapabilitiesWithTypeHierarchy {
-    let augments_syntax_tokens = initialization_params
+fn client_augments_syntax_tokens(initialization_params: &InitializeParams) -> bool {
+    initialization_params
         .capabilities
         .text_document
         .as_ref()
         .and_then(|c| c.semantic_tokens.as_ref())
         .and_then(|c| c.augments_syntax_tokens)
-        .unwrap_or(false);
+        .unwrap_or(false)
+}
+
+pub fn capabilities(
+    indexing_mode: IndexingMode,
+    initialization_params: &InitializeParams,
+) -> ServerCapabilitiesWithTypeHierarchy {
+    let augments_syntax_tokens = client_augments_syntax_tokens(initialization_params);
 
     // Parse syncNotebooks from initialization options, defaults to true
     let sync_notebooks = initialization_params
@@ -1286,6 +1290,9 @@ pub fn capabilities(
             // syntax highlighting for tokens we don't provide, it will be a regression
             // (e.g. users might lose keyword highlighting).
             // Therefore, we should not produce semantic tokens if the client doesn't support `augments_syntax_tokens`.
+            // We now have an implementation path for a full semantic token stream that fills in
+            // syntax tokens, but we do not advertise that capability to non-augmenting clients yet.
+            // Keep the provider gate here until that full-stream behavior is ready to expose.
             Some(SemanticTokensServerCapabilities::SemanticTokensOptions(
                 SemanticTokensOptions {
                     legend: SemanticTokensLegends::lsp_semantic_token_legends(),
@@ -5085,10 +5092,11 @@ impl Server {
         let uri = &params.text_document.uri;
         let maybe_cell_idx = self.maybe_get_code_cell_index(uri);
         let handle = self.make_handle_if_enabled(uri, Some(SemanticTokensFullRequest::METHOD))?;
+        let include_syntax_tokens = !client_augments_syntax_tokens(&self.initialize_params);
         Ok(Some(SemanticTokensResult::Tokens(SemanticTokens {
             result_id: None,
             data: transaction
-                .semantic_tokens(&handle, None, maybe_cell_idx)
+                .semantic_tokens(&handle, None, maybe_cell_idx, include_syntax_tokens)
                 .unwrap_or_default(),
         })))
     }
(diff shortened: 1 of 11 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1286,6 +1290,9 @@ pub fn capabilities(
             // syntax highlighting for tokens we don't provide, it will be a regression
             // (e.g. users might lose keyword highlighting).
             // Therefore, we should not produce semantic tokens if the client doesn't support `augments_syntax_tokens`.
+            // We now have an implementation path for a full semantic token stream that fills in
+            // syntax tokens, but we do not advertise that capability to non-augmenting clients yet.
+            // Keep the provider gate here until that full-stream behavior is ready to expose.
```
kinto0, line 1295: this comment is relative to the PR, not the future of the file. i.e. it speaks to this change explicity

let's just remove it?

kinto0:
this looks great. let me do some memory testing

pyrefly/lib/lsp/wasm/semantic_tokens.rs
```diff
@@ -23,13 +23,21 @@ impl Transaction<'_> {
...
         let module_info = self.get_module_info(handle)?;
         let ast = self.get_ast(handle)?;
         let legends = SemanticTokensLegends::new();
         let disabled_ranges = disabled_ranges_for_module(ast.as_ref(), *handle.sys_info());
         let mut builder = SemanticTokenBuilder::new(limit_range, disabled_ranges);
 
+        if include_syntax_tokens {
+            let tokens = self
+                .get_syntax_tokens(handle)
+                .expect("syntax tokens must be cached with the AST");
```
kinto0: risky: we have to clear ast + syntax tokens together. can we bind their storage together?

we can probably make a new `ParsedModule` and impl `Deref` for it to avoid updating callsites:
```
impl Deref for ParsedModule {                                                                                                                                                                                           
    type Target = ModModule;                                                                                                                                                                                            
                                                                                                                                                                                                                        
    fn deref(&self) -> &ModModule {                                                                                                                                                                                     
        &self.module                                                                                                                                                                                                    
    }                                                                                                                                                                                                                   
}
#[derive(Debug, Clone)]                                                                                                                                                                                                 
pub struct ParsedModule {                                                                                                                                                                                               
    module: ModModule,                                                                                                                                                                                                  
    tokens: Tokens,                                                                                                                                                                                                     
  }
```

pyrefly/lib/state/module.rs
```diff
@@ -367,6 +372,7 @@ impl PostComputeGuard<'_> {
             "evict_ast called before answers computed"
         );
         self.state.steps.ast.store(None);
+        self.state.steps.syntax_tokens.store(None);
```
kinto0: if we want to keep the separation between AST/syntax_tokens, maybe we add a comment here saying this ordering is required to avoid a crash

kinto0 (changes requested):

pyrefly/lib/lsp/wasm/semantic_tokens.rs
kinto0: as a follow-up we probably should eventually update callsites

pyrefly/lib/lsp/non_wasm/server.rs
magic-akari: That's a good point. 

I added the syntax tokens so that semantic tokens are available whether they are augmented or not. 
However, I didn't loosen the gate because I feel that decision should be left to the pyrefly team. 
That's why I felt the need to add a comment here.

Should we just change the gate now and remove the comment entirely?

kinto0: i'm happy removing the gate as long as we have full support. do you think we're at that point after this?

pyrefly/lib/lsp/wasm/semantic_tokens.rs
kinto0: sorry for the back/forth: even better, what if we just re-lexed on-demand in the AST when include_syntax_tokens is true?

we wouldn't have to worry about storing extra on the AST since we wouldn't store anything there.

magic-akari: That was actually the approach in my initial commit, 7057f8c. I moved away from it because it would re-parse the module when `include_syntax_tokens` is true, and I was worried about the performance impact of doing that repeatedly.

My preference is still to cache the tokens from the original parse, as in the current direction, since that avoids the extra parse and keeps the tokens tied to the same parser invocation as the AST.

That said, if you’d prefer the simpler on-demand re-lex/re-parse approach to avoid storing extra data, I can revert back toward the original implementation. Which direction would you prefer?

kinto0: it depends how long it takes to lex and how much storage the cache uses. let me find the biggest file I can and benchmark it

kinto0: benchmarks were pretty good for both: 5ms to parse (we can't just lex, we have to parse) and 800kb extra storage on a very very big file. 

it looks like we evict the ast for any non-open files (by virtue of require::exports) and we keep the ast for open files. 

since parsing is O(semantic token requests) and ast is O(open files), I think your approach here by caching it in the ast is the best solution

pyrefly/lib/state/steps.rs
```diff
@@ -69,13 +71,47 @@ pub struct Context<'a, Lookup> {
...
+}
+
 #[derive(Debug, Default, Dupe, Clone)]
 pub struct Steps {
     /// The last step that was computed.
     /// None means no steps have been computed yet.
     pub last_step: Option<Step>,
     pub load: Option<Arc<Load>>,
-    pub ast: Option<Arc<ModModule>>,
+    pub parsed: Option<Arc<ParsedModule>>,
```
kinto0, line 114: one final nit: I don't think we need to rename the `ast` step to `parsed`. If we want to, we can do that in another PR

pyrefly/lib/lsp/non_wasm/server.rs
kinto0: we can leave it for now and do it later. adding a todo comment here

kinto0 (approved):

pyrefly/lib/state/steps.rs
kinto0: I'll fix this up on import, no need to resubmit for it

kinto0:
I'm also adding a follow-up optimization to avoid computing these tokens at all when we don't require::everything. this makes it so we never request it for CLI or dependencies, limiting peak memory

yangdanny97 (approved):

PR #3616, opened 2026-05-31
Recognize walrus targets in decorator expressions

### LINKED ISSUE #3300: unknown-name false positive when a name is introduced via a walrus in a decorator
### Describe the Bug

Reproducer:

```
def when(cond):
  def wrapper(fn):
    if cond:
      fn()
  return wrapper


@when(x := 42)
def _(): ...

x
```

Output:

```
ERROR Could not find name `x` [unknown-name]
  --> /tmp/foo.py:11:1
   |
11 | x
   | ^
   |
```

### Sandbox Link

_No response_

_No response_

terror:
Resolves https://github.com/facebook/pyrefly/issues/3300

This diff records `named-expression` targets while collecting definitions from function and class decorator
expressions, so names introduced by `@decorator(x := ...)` are visible after the decorated definition.

Function decorators are also bound before entering the function annotation scope, matching where Python
evaluates decorators and matching the existing class decorator behavior. This prevents walrus targets in
decorators from being treated as unknown or uninitialized in the enclosing scope.

### DIFF
--- a/pyrefly/lib/binding/function.rs
+++ b/pyrefly/lib/binding/function.rs
@@ -880,11 +880,11 @@ impl<'a> BindingsBuilder<'a> {
             None
         };
 
+        let decorators = self.decorators(mem::take(&mut x.decorator_list), def_idx.usage());
+
         self.scopes.push(Scope::annotation(x.range));
         let (return_ann_with_range, legacy_tparams) =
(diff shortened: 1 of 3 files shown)


yangdanny97:
nice, thanks

stroxler (approved):

PR #3617, opened 2026-05-31
fix Pyrefly allows unquoted forward references in Python ≤3.13 even though they error at runtime #357

### LINKED ISSUE #357: Pyrefly allows unquoted forward references in Python ≤3.13 even though they error at runtime
### Describe the Bug

Right now, this code typechecks correctly, but will error out at runtime. 
```
from typing import Optional, Union

@dataclass
class Leaf:
    value: int

@dataclass
class Node:
    left: Optional[Tree]
    right: Optional[Tree]

type Tree = Union[Leaf, Node]
```

The issue is that the annotations `left: Optional[Tree]` and `right: Optional[Tree]` are invalid until [PEP649](https://peps.python.org/pep-0649/) (deferred annotations) is introduced in Python 3.14. 

So this code is incorrect in Python ≤3.13 and correct in ≥3.14.  Right now it passes pyrefly -- is this the expected behavior? 

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDyCMSYKAhgDYA0UAqiqSgFCiRQAm5M5AxpeQDOggKaDM2XPi49+Qwc2YABGXwHDmc4VAAyI8sABczKKagA3KgFcRhzChiKV3NfM3rxAOTAdbJs5QiwDB2RCRkVADaACogIiIAuv6mIEhoABYhhMRMUbHxSczwCCJQ+aUAvPSMZJF6BrTevglAA

_No response_

asukaminato0721:
# Summary

Fixes #357

runtime-evaluated annotations now report uninitialized forward names for executable Python < 3.14 unless `from __future__ import annotations` is active.

preserved initialization state through legacy type-parameter lookup so function annotations get the same check.

# Test Plan

added regression tests for the dataclass/tree case, function annotations, Python 3.14, and future annotations.

### DIFF
--- a/pyrefly/lib/binding/expr.rs
+++ b/pyrefly/lib/binding/expr.rs
@@ -400,10 +400,6 @@ impl<'a> BindingsBuilder<'a> {
                         self.error(name.range, ErrorKind::UnboundName, error_message);
                     }
                 }
-                let is_enclosing_class_name = self
-                    .scopes
-                    .enclosing_class_name()
-                    .is_some_and(|class_name| class_name.id == name.id);
                 if is_runtime_evaluated_annotation
                     && matches!(
                         usage,
@@ -411,7 +407,6 @@ impl<'a> BindingsBuilder<'a> {
                             is_annotation: true
                         }
                     )
-                    && !is_enclosing_class_name
                     && self.module_info.path().style() == ModuleStyle::Executable
                     && !self.sys_info.version().at_least(3, 14)
                     && !self.scopes.has_future_annotations()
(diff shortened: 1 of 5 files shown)


pyrefly/lib/binding/expr.rs
```diff
@@ -387,6 +400,25 @@ impl<'a> BindingsBuilder<'a> {
                         self.error(name.range, ErrorKind::UnboundName, error_message);
                     }
                 }
+                let is_enclosing_class_name = self
```
stroxler: Any particular reason for this edge case?

It is in fact a runtime error to have code like this:
```
class C:
     c: C
```
without `__future__` annotations or 3.14+ semantics

If that causes a lot of problems in mypy primer maybe an exception makes sense, but if so we should probably document the rationale

yangdanny97 (approved):

PR #3620, opened 2026-05-31
Report invalid-variance errors on the offending overload

### LINKED ISSUE #3536: `invalid-variance` error always appears on the first overload
### Describe the Bug

```py
from typing import TypeVar, Generic, Iterable, overload

_T_co = TypeVar("_T_co", covariant=True)

class Foo(Generic[_T_co]):
    @overload
    def foo(self) -> None: ... # error: invalid-variance
    @overload
    def foo(self, instance: _T_co) -> None: ... # should error here instead
    def foo(self, instance: _T_co | None = ...) -> None: ... # (and here, but that's a different issue: #3535)
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVZjABqqSgBo6AcRjoYlCAGMpASQYLU2WFNwA3BVFypMAHXTmA%2BgMtLcdALyDhYygApTIa7dyepdvXEIDAYHAUoAVxgASnNzJShUODg6ADFcXDdZeUUlAG1vOwBdaMRzOgq6AAF9Q2MzdEq6TBgwOjAMtzgYKDBougBaAD46ADlceUQ6Qhm6AGI6BWpKKdZAqAhMAcDFDCUYcsqag0ojE0OKlraOzO7eqVY4Bj2YKcLcfuGxidfp2YW4AALXARKCYRaUZZ0QEKGCcdBPGDnRqVK7tTp3MAPBHPdD7N42Ox0AA%2B33kjj%2BhE%2BI3Gk0p8zobgw4JhlBgUmwEX4DEBqAYAHIUqhmhAwGBYeh%2BBBklEpnMAMwAViV0RAEhAZDZYCgpEIDFoUAoZB5E22CjgEAmFM88sIAEYACyecx5Ja8OBFcwtYhspR8mDgpyeN2UTx0cxcYgbJQQBgDLRPSioJRxhJJFJBkAh53oSPR2MDSO8fiZ7MgcO57j5hg6gZXVgB%2BMMBiKTkaa1ZyG8HM0aUW9gDOAgyj7Dtl8zoU0wfC8iJPCAGAY0PlKQFjruhkAT3BxtkARwiEDZWwA1jBSPGlPtkuvljncMQ5AMIuhiMmT6wOKWN2HzMeX5gIQDH6Ty3t25Z-gGAFAXYWCxpajTfneEHoC%2B8jXmglAXnY3BBEOiHhp2yEVmyBioFAAxMI%2BYGbgyrAdOYaEwBh4gXlRMDARMJwWlaSHgRWL5ssmvLaHCfG0Yx6BCauWiwEuK6AsBSRiYRZYCVJ8C4FAgSiXWAEKPGUBQDRv6oZKwhbEWfAmShL5zo2EBsJObI2RWOzBHiHG9nAy4MKurkRug6ybHWzG8HyvA2SAAC%2B6rJgwC4wKk0AwBQaBYHgRBkDFQA

_No response_

mikeleppane:
Fixes #3536

### What

For an overloaded method on a generic class, the `invalid-variance` error
(covariant/contravariant TypeVar used in the wrong position) was always
reported on the **first** overload, regardless of which signature actually
misused the TypeVar. This moves the error onto the offending overload.

```python
from typing import Generic, TypeVar, overload

_T_co = TypeVar("_T_co", covariant=True)

class Foo(Generic[_T_co]):
    @overload
    def foo(self) -> None: ...                    # <- error was reported here
    @overload
    def foo(self, instance: _T_co) -> None: ...   # <- error belongs here (covariant TypeVar in a parameter)
    def foo(self, instance: _T_co | None = ...) -> None: ...
```

Before: error on the first overload. After: error on the signature that uses
`_T_co` in a contravariant position — matching pyright (`Covariant type variable
cannot be used in parameter type`) and mypy (`Cannot use a covariant type
variable as a parameter`), both of which point at the second overload.

### Why

The violation was *detected* on the correct signature; only its reported
location was wrong. `check_class_violations` looked up a single
`field_decl_range` for the whole member and `check_method_shallow` reused that
one range for every signature in the `Type::Overload`, so all overload
violations landed on the first `def`.

The per-overload `def` ranges do exist while the overload type is built
(`extract_signatures` returns `Vec1<(TextRange, OverloadType)>`), but they are
dropped when the `Type::Overload` is constructed. Pyrefly types are
position-free by design (they are interned, hashed, and compared for
incremental reuse, so they cannot carry source ranges).

### How

Recover each overload arm's range at check time from the identity the type
still carries, instead of putting positions back into the type:

- `check_method_shallow` now special-cases `Type::Overload` and resolves a range
  **per signature** from that signature's **own** inner `FuncMetadata`
  (`def_index` → existing `KeyUndecoratedFunctionRange` binding →
  `ShortIdentifier::range()`) - the same `def_index → def range` bridge the LSP
  already uses in `resolve_func_def_range`.
- It reads each arm's own metadata, **not** the merged overload-group metadata,
  which points at the implementation (with an implementation) or the first
  overload (without one).
- All other callable shapes keep flowing through `visit_toplevel_callable` with
  the field range - single-def methods were already located correctly, so their
  behavior is unchanged.
- The lookup is `Option`-returning and local to the current module (variance
  only checks a class's own fields, so the `def_index` is always local).
  Synthesized/metadata-only callables with no `def_index` fall back to the field
  range.

No change to the `Type` representation; no new bindings.

### Test plan

New/updated tests in `pyrefly/lib/test/variance_inference.rs`:

- Un-`bug`-marked `test_variance_error_in_overload` (the existing
  known-wrong-location test), with the expectation moved to the offending arm.
- `test_variance_error_in_overload_issue_3536` — the exact repro from the issue.
- `test_variance_error_in_overload_first_arm` — violation in the *first* arm, to
  prove the error tracks the offender rather than a fixed index.
- `test_variance_error_in_overload_both_arms` — both arms violate; each is
  reported on its own line.
- `test_variance_error_in_overload_contravariant_return` — contravariant TypeVar
  in a return (covariant) position, located on the returning arm.
- `test_variance_error_in_overload_generic_arm` — the offending arm is itself
  generic (`OverloadType::Forall`), exercising the Forall branch of the lookup.
- `test_variance_overload_no_false_positive` — correct positions across overloads
  must not over-report.
- `test_variance_error_in_single_method` — regression guard that single
  (non-overloaded) methods are still flagged.

Verification:

- `cargo test -p pyrefly --lib` — full suite green (5293 passed, 0 failed).
- `python3 test.py --no-test --no-conformance --no-jsonschema` — format + lint clean.
- Snippet confirmed against the release binary: `invalid-variance` now reported on
  the offending overload, matching pyright and mypy.
- `mypy_primer` (wide local scope, 26 typing-heavy projects incl. pydantic, attrs,
  trio, anyio, sympy, pandera, scipy, sphinx, mypy): **no diff** — no regressions
  and no spurious changes, as expected for a fix that only relocates an existing
  diagnostic.

### DIFF
--- a/pyrefly/lib/alt/class/variance_inference.rs
+++ b/pyrefly/lib/alt/class/variance_inference.rs
@@ -18,19 +18,26 @@ use pyrefly_types::dimension::SizeExpr;
 use pyrefly_types::heap::TypeHeap;
 use pyrefly_types::tensor::TensorShape;
 use ruff_python_ast::name::Name;
+use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
+use starlark_map::Hashed;
 use starlark_map::small_map::SmallMap;
 
 use crate::alt::answers::LookupAnswer;
 use crate::alt::answers_solver::AnswersSolver;
 use crate::alt::class::class_field::ClassField;
 use crate::alt::types::class_bases::ClassBases;
+use crate::binding::binding::KeyUndecoratedFunctionRange;
+use crate::types::callable::Callable;
+use crate::types::callable::FuncMetadata;
+use crate::types::callable::FunctionKind;
 use crate::types::callable::Params;
 use crate::types::class::Class;
 use crate::types::quantified::Quantified;
 use crate::types::tuple::Tuple;
 use crate::types::type_var::PreInferenceVariance;
 use crate::types::type_var::Variance;
+use crate::types::types::OverloadType;
 use crate::types::types::TParams;
 use crate::types::types::Type;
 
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

rchen152:
@mikeleppane This is really nice work, thank you!

PR #3627, opened 2026-06-01
Add per-target `config` and `root` fields to build system response (#3627)

kinto0:
Summary:

External users with custom build systems (e.g., Arista Networks) need the
build system to be the single source of truth for both import resolution
and type checking settings. Currently the build system JSON response can
provide `srcs` and `deps` per target, but has no way to communicate
per-target type checking configuration or per-target source roots.

This adds two new optional fields to `PythonLibraryManifest`:
- `config`: raw JSON config overrides (errors, preset, etc.) that are
  deserialized into `ConfigBase` by `pyrefly_config` at resolution time.
  Stored as `serde_json::Value` to avoid a circular crate dependency.
- `root`: per-target override of the top-level root, used for both
  absolutizing `srcs` paths and as the import resolution root.

Also makes `deps` and `buildfile_path` optional via `#[serde(default)]`
so build systems that don't map cleanly to per-target deps can omit them.

The config resolution chain becomes:
  sub-config -> target config -> root config -> preset defaults

`ErrorConfig` is simplified to own its `ErrorDisplayConfig` (removing the
lifetime parameter) so that target configs deserialized from JSON can
participate in the resolution chain without lifetime issues.

Differential Revision: D107113346

### DIFF
--- a/crates/pyrefly_build/src/query/mod.rs
+++ b/crates/pyrefly_build/src/query/mod.rs
@@ -269,15 +269,25 @@ pub trait SourceDbQuerier: Send + Sync + fmt::Debug {
 
 #[derive(Debug, PartialEq, Eq, Deserialize, Clone)]
 pub(crate) struct PythonLibraryManifest {
+    #[serde(default)]
     pub deps: SmallSet<Target>,
     pub srcs: SmallMap<ModuleName, Vec1<InternedPath>>,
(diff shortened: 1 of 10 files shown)


connernilsen (changes requested):

connernilsen (approved):

Issue #3624, opened 2026-06-01
`pyrefly report` includes `if __name__ == "__main__"` declarations


jorenham:
For example, in NumPy, for `numpy/f2py/crackfortran.py`, which starts as

```py
if __name__ == "__main__":
    files = []
    funcs = []
    f = 1
    f2 = 0
    f3 = 0
    showblocklist = 0
    # --snip--
```

([src](https://github.com/numpy/numpy/blob/b6e8ad83299fc6f87107492df0039dadfb5b9b98/numpy/f2py/crackfortran.py#L3654-L3727))

is reported by `pyrefly coverage report` as

```json
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.files",
  "n_typable": 0,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 0,
  "location": {
    "line": 3655,
    "column": 5
  }
},
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.funcs",
  "n_typable": 0,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 0,
  "location": {
    "line": 3656,
    "column": 5
  }
},
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.f",
  "n_typable": 1,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 1,
  "location": {
    "line": 3657,
    "column": 5
  }
},
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.f2",
  "n_typable": 1,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 1,
  "location": {
    "line": 3658,
    "column": 5
  }
},
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.f3",
  "n_typable": 1,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 1,
  "location": {
    "line": 3659,
    "column": 5
  }
},
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.showblocklist",
  "n_typable": 1,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 1,
  "location": {
    "line": 3660,
    "column": 5
  }
},
{
  "kind": "attr",
  "name": "numpy.f2py.crackfortran.l",
  "n_typable": 1,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 1,
  "location": {
    "line": 3661,
[...]

PR #3630, opened 2026-06-01
fix(exports): condition the export of definitions under `TYPE_CHECKING`

### LINKED ISSUE #3629: Definitions under `TYPE_CHECKING` should not be importable
### Describe the Bug

In trying to make our codebase generally compliant with [quoted annotations](https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_quote-annotations), we ran into the issue of identifying modules that were unintentionally importing definitions as a re-export, rather from the file of definitional source of truth.

When converting to quoted annotations, these re-exports become nested in a conditional, and should no longer be importable to other modules. However, pyrefly does not flag this.

Similar to https://github.com/facebook/pyrefly/issues/3624, we should not include these declarations.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AVAJoAFAKIB9AMIAJMVIDSASQByAcQA66LRDCDRk2fOXrEWuhbr46AXjoBGEABoQ2EuURVadNFjzWuHj4rZxAyShgwKFJCBlooCgBiOhFSCKjSHwwcAjoAY1x0SDYAV0pUBghCwi1kgGUYGDoACwYGYjhEAHou8Mjowl42Lph0LsxcPLgugqKIUvLKwq66MF46VAA3VGhUbFh8wuKyiqr0OlxiJfQ4GvQyBmbCgFpNmEo4M9s6DRAAZkI9gATL8tABtd7UD4AXS0JXQgV4DBgmGemAgETylTe3wA5PN0LwYLitISGM8IgBHEoYlHPADWMFIz1QeTy8DgeIA7qhKOgSegQABfFys7EwABi0BgFDcZGFQA

_No response_

rexledesma:
# Summary

Fixes https://github.com/facebook/pyrefly/issues/3629

# Test Plan

`cargo test`

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -110,6 +110,9 @@ pub struct Definition {
     /// definition, this equals `range`. Used to determine if a variable is
     /// reassigned after a given point (e.g. after a nested function definition).
     pub last_range: TextRange,
+    /// True while every definition site is inside an `if TYPE_CHECKING:` body.
+    /// Such names resolve in-module but are not importable at runtime.
+    pub type_checking_only: bool,
 }
 
 impl Definition {
@@ -120,7 +123,8 @@ impl Definition {
         }
     }
 
-    fn merge(&mut self, other: DefinitionStyle, range: TextRange) {
+    fn merge(&mut self, other: DefinitionStyle, range: TextRange, in_type_checking: bool) {
+        self.type_checking_only &= in_type_checking;
         // To ensure binding code cannot produce invalid lookups, we ensure that
         // `self.style` and `self.range` always match.
         if other < self.style {
@@ -274,6 +278,7 @@ struct DefinitionsBuilder {
     is_init: bool,
     sys_info: SysInfo,
     inner: Definitions,
+    in_type_checking: bool,
 }
 
 fn is_private_name(name: &Name) -> bool {
@@ -327,6 +332,7 @@ impl Definitions {
             sys_info,
             is_init,
             inner: Definitions::default(),
+            in_type_checking: false,
         };
         builder.stmts(x);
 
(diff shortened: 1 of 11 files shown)


pyrefly/lib/export/definitions.rs
```diff
@@ -110,6 +110,9 @@ pub struct Definition {
     /// definition, this equals `range`. Used to determine if a variable is
     /// reassigned after a given point (e.g. after a nested function definition).
     pub last_range: TextRange,
+    /// True while every definition site is inside an `if TYPE_CHECKING:` body.
+    /// Such names resolve in-module but are not importable at runtime.
+    pub type_checking_only: bool,
```
rexledesma, line 115: We could probably have a better diff to squash the problem of " no conditional exports", which would also address https://github.com/facebook/pyrefly/issues/3624, but want to keep this diff scoped to get immediate feedback first ✌️

jorenham:
The primer errors are probably related to my comment at  https://github.com/facebook/pyrefly/issues/3629#issuecomment-4595887524

rexledesma:
@jorenham hmm ok I guess that's fine. I'm not a fan of these re-exports in the TYPE_CHECKING block because it adds some indirection around the source of source for these type annotations.

I'll switch my implementation to be more liberal about the exports here, but maybe we can have a rule to emit a warning/error for this sort of indirection?

jorenham:
> but maybe we can have a rule to emit a warning/error for this sort of indirection?

feels more like a job for a linter like `ruff` tbh; it's unrelated to type-safety, after all.

rexledesma:
The proposed rule would be very similar to this rule that already exists: https://pyrefly.org/en/docs/error-kinds/#implicit-import

PR #3633, opened 2026-06-01
fix: show hover in notebook when kernal switches

### LINKED ISSUE #3598: [VSCode][ipynb] Changing kernels causes hover to stop working
### Describe the Bug

reproduction:
- VSCode
- create test.ipynb with `x = 5` line
- select kernel, reload window: hover works
- change to another kernel: hover stops working
```
[Trace - 10:32:54 AM] Sending request 'textDocument/hover - (38)'.
Params: {
    "textDocument": {
        "uri": "vscode-notebook-cell:/Users/kylei/projects/numpy/test.ipynb#W0sZmlsZQ%3D%3D"
    },
    "position": {
        "line": 0,
        "character": 3
    }
}


[Trace - 10:32:54 AM] Received response 'textDocument/hover - (38)' in 2ms.
No result returned.

```

### Sandbox Link

_No response_

VSCode Version: 1.121.0 (Universal)

NarxPal:
# Summary

Cell content metdata was being updated when there are changes made to cells inside the notebook. LspEvent's didChange event runs the `notebook_document_did_change` fn which earlier updated the cell metdata of notebook only when there are changes made to the cell and since kernel switching doesn't cause any cell changes, the existing cells didn't get inserted into `notebookDocument` struct's cell field. this caused hover to stop working on the existing cells after kernel switching.

Fixes #3598 

# Test Plan

- Manually verified in VS Code using a local Pyrefly build from this branch.
- created a `test.ipynb` file, created new cell in it.
- wrote `x=5` , and switched kernel and hovered over `x`

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -3669,15 +3669,18 @@ impl Server {
             notebook_document.metadata = Some(metadata.clone());
         }
         notebook_document.version = version;
+
+        // Track existing cell contents.
+        // Track cell content even when there are no changes to cells, so that cells inside NotebookDocument struct stay updated after kernal switching.
+        for cell in &notebook_document.cells {
+            let cell_contents = original_notebook
+                .get_cell_contents(&cell.document)
+                .unwrap_or_default();
+            cell_content_map.insert(cell.document.clone(), cell_contents);
+        }
+
         // Changes to cells
         if let Some(change) = &params.change.cells {
-            // Track existing cell contents
-            for cell in &notebook_document.cells {
-                let cell_contents = original_notebook
-                    .get_cell_contents(&cell.document)
-                    .unwrap_or_default();
-                cell_content_map.insert(cell.document.clone(), cell_contents);
-            }
             // Structural changes
             if let Some(structure) = &change.structure {
                 let start = structure.array.start as usize;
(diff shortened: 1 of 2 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -3669,15 +3669,18 @@ impl Server {
             notebook_document.metadata = Some(metadata.clone());
         }
         notebook_document.version = version;
+
+        // Track existing cell contents.
```
kinto0: nit: seems like this comment is duplicated

kinto0 (changes requested):
thanks! 

is there any way to test this behavior with a lsp_interaction test that failed before your changes?

pyrefly/lib/lsp/non_wasm/server.rs
NarxPal: combined the two comments into one.

NarxPal:
previously `test_notebook_hover_basic` fn was not handling the hover over cell_content after kernel change, so i added that step in another fn called `test_notebook_hover_after_kernel_change` and since kernel change causes an lsp didChange event we use the `change_notebook` fn to stimulate the behaviour of kernel change.

kinto0 (approved):
thank you! clean fix with great test coverage

yangdanny97 (approved):

PR #3635, opened 2026-06-01
Fix TSP getComputedType to preserve declarations for callees (pylance-release#8020)

### LINKED ISSUE #8020: [TSP] Parameter hints don't work with TSP
1. Enable Pyrefly as a TSP server
2. Add `print("foo", sep=)` to a python file
3. Trigger parameter hints on the `sep=`

Expected result:
Parameter hints appear and sep is bolded.

Actual result:
Nothing happens

rchiodo:
## Summary

Fixes [microsoft/pylance-release#8020](https://github.com/microsoft/pylance-release/issues/8020) — _[TSP] Parameter hints don't work with TSP_.

When Pyrefly is used as a TSP server (e.g. with Pylance as the TSP client), parameter hints / signature help on a call like `print("foo", sep=)` never appear. The TSP client relies on `typeServer/getComputedType` to return the function's declared type so it can re-resolve the signature(s) (parameters, overloads, defaults) from source on its side. Today, the TSP endpoint funnels the request through `Transaction::get_type_at`, which — when the identifier is in call position — does one of two things:

1. Substitutes the **return type of the chosen overload** (via `get_chosen_overload_trace_for_surface`), or
2. Coerces the type to a synthesized `Callable` (`coerce_type_to_callable`).

Both paths discard the original `Declaration::Regular` link that points back at the function's source (e.g. `builtins.pyi` for `print`). With no `Regular` declaration on the wire, the TSP client has nothing to resolve back to and signature help silently fails.

This behavior is correct for the LSP hover path (you want the call's effective type / return type for hover), but wrong for TSP `getComputedType`, which is specified to return the raw declared type of the queried node.

## Fix

Split `get_type_at_impl` into a parameterized helper and add a public `get_type_at_preserving_declaration` entry point that skips the callee-coercion branch entirely. The TSP `getComputedType` handler in `lsp/non_wasm/server.rs` now calls the new method, while existing LSP consumers continue to use `get_type_at` and see no behavior change.

The change in `pyrefly/lib/state/lsp.rs` is purely a refactor + new flag:

- `get_type_at_impl` now delegates to `get_type_at_impl_with_options(..., coerce_callees: true)` — preserving today's LSP behavior.
- The new `get_type_at_impl_with_options` gates the call-position rewrite (chosen-overload trace + `coerce_type_to_callable`) on the `coerce_callees` flag.
- `get_type_at_preserving_declaration` is a thin public wrapper that passes `coerce_callees: false` (and `for_display: false`, since TSP serializes the type rather than rendering it).

## Tests

Three new regression tests in `pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs`, all driving the real TSP endpoint via `get_computed_type_ok`:

1. **`test_get_computed_type_builtin_print_at_call_position_preserves_declaration`** — the direct repro for #8020. Sends `print("foo", sep="x")`, queries the `print` identifier at the call site, and asserts the result is a `Function`/`Overloaded` whose first declaration is `DeclarationKind::Regular` (kind = 0) with a `node.uri` pointing at `builtins.pyi`. Before the fix this came back as a synthesized `Callable` / return-type with no usable declaration, which is exactly what broke Pylance's parameter hints.

2. **`test_get_computed_type_overloaded_callee_returns_overloaded_not_return_type`** — defines a two-overload `process` and queries it at `process(1)`. Asserts the returned kind is `Overloaded` and the `overloads` array has both signatures. Pre-fix, the chosen-overload trace collapsed this to the matched overload's return type (`str`), which would prevent the client from showing the alternate signature in signature help.

3. **`test_get_computed_type_local_def_callee_preserves_declaration`** — covers the non-builtin path: defines `greet` locally and queries it at `greet("world")`. Asserts kind is `Function`, the declaration is `Regular`, the declaration `name` is `"greet"`, and the declaration's `node.uri` matches the source file. This guards against regressions where coercion would still fire for user-defined functions.

Together these lock in: (a) the fix actually resolves #8020 for the canonical `print(...)` case, (b) overloads are preserved end-to-end for call sites, and (c) the behavior is correct for user code, not just typeshed.

## Risk

Scoped to the TSP `getComputedType` handler — all existing LSP callers continue to go through `get_type_at` unchanged.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -6315,7 +6315,11 @@ impl TspInterface for Server {
         let module_info = transaction.get_module_info(&handle)?;
         let position =
             module_info.from_lsp_position(lsp_types::Position { line, character }, notebook_cell);
-        transaction.get_type_at(&handle, position)
+        // For TSP, return the raw declared type without coercing callees in
+        // call position. This keeps the function's `Declaration::Regular`
+        // intact on the wire, which TSP clients need to re-resolve the
+        // signature (parameters, overloads) from source.
+        transaction.get_type_at_preserving_declaration(&handle, position)
     }
 
     fn resolve_func_def_range(
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -984,6 +984,16 @@ impl<'a> Transaction<'a> {
         position: TextSize,
         for_display: bool,
     ) -> Option<Type> {
+        self.get_type_at_impl_with_options(handle, position, for_display, true)
+    }
+
+    fn get_type_at_impl_with_options(
+        &self,
+        handle: &Handle,
+        position: TextSize,
+        for_display: bool,
+        coerce_callees: bool,
+    ) -> Option<Type> {
         match self.identifier_at(handle, position) {
             Some(IdentifierWithContext {
                 identifier: id,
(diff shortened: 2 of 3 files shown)


yangdanny97 (approved):

PR #3637, opened 2026-06-02
Fix callHierarchy/incomingCalls returning no callers for cross-file c…

### LINKED ISSUE #3636: callHierarchy/incomingCalls returning no callers for cross-file callsites
### Describe the Bug

pyrefly language server cannot find a cross-file callsite unless the file containing the callsite was explicitly opened via didOpen.

### Reproduction
Create a project with two files, where one imports and calls a function from the other:

test/main.py:
```python
from base1 import t1
import sys


def main():
    t1(sys.argv[1])
```
test/base1.py:
```python
import os

def t1(a1):
    os.system(a1)
```
With the following test script:
```python
import subprocess, json, time, select

def test(name, cmd):
    p=subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    def s(m):
        d=json.dumps(m).encode()
        p.stdin.write(f'Content-Length: {len(d)}\r\n\r\n'.encode()+d)
        p.stdin.flush()
    def rcv(eid):
        while True:
            h=b''
            while not h.endswith(b'\r\n\r\n'):
                b=p.stdout.read(1)
                if not b: return None
                h+=b
            l=int(h.decode().split(':')[1].strip())
            msg=json.loads(p.stdout.read(l))
            if msg.get('id')==eid: return msg

    caps={
        'textDocument':{
            'callHierarchy':{'dynamicRegistration':False},
            'documentSymbol':{'hierarchicalDocumentSymbolSupport':True}
        }
    }
    s({'jsonrpc':'2.0','id':1,'method':'initialize','params':{
        'processId':None,'rootUri':'file:///path/to/test/',
        'workspaceFolders':[{'uri':'file:///path/to/test/','name':'test'}],
        'capabilities':caps
    }})
    rcv(1)
    s({'jsonrpc':'2.0','method':'initialized'})
    time.sleep(3)
    
    while True:
        ready,_,_=select.select([p.stdout],[],[],0.3)
        if not ready: break
        rcv(None)

    # Open base1.py + docSymbol 
    with open('/path/to/test/base1.py') as f:
        txt=f.read()
    s({'jsonrpc':'2.0','method':'textDocument/didOpen','params':{
        'textDocument':{'uri':'file:///path/to/test/base1.py','languageId':'python','version':0,'text':txt}
    }})
    s({'jsonrpc':'2.0','id':10,'method':'textDocument/documentSymbol','params':{
        'textDocument':{'uri':'file:///path/to/test/base1.py'}
    }})
    r=rcv(10)
    syms=r.get('result',[]) if r else []
    
    if not syms:
        print(f'{name}: docSymbol empty')
        s({'jsonrpc':'2.0','method':'exit'})
        p.wait(timeout=3)
        return 0
    
    sel=syms[0]['selectionRange']['start']
    
    # prepareCallHierarchy
    s({'jsonrpc':'2.0','id':2,'method':'textDocument/prepareCallHierarchy','params':{
        'textDocument':{'uri':'file:///path/to/test/base1.py'},
        'position':sel
    }})
    r=rcv(2)
    got=r and r.get('result')
    if not got:
        print(f'{name}: prepare failed')
        s({'jsonrpc':'2.0','method':'exit'})
        p.wait(timeout=3)
        return 0
    
    # incomingCalls
    s({'jsonrpc':'2.0','id':3,'method':'callHierarchy/incomingCalls','params':{'item':r['result'][0]}})
    r=rcv(3)
    calls=r.get('result',[]) if r else []
    print(f'{name}: {len(calls)} callers found')
    
    s({'jsonrpc':'2.0','method':'exit'})
    p.wait(timeout=3)
    return len(calls)

print('Test: open base1.py, docSymbol, prepareCallHierarchy, incomingCalls')
print()
a=test('pyrefly', ['/path/to/pyrefly','lsp'])
b=test('basedpyright', ['/path/to/basedpyright-langserver','--stdio'])

```
Output:
```
pyrefly     : 0 callers found
basedpyright: 1 callers found
```

### Sandbox Link

_No response_

_No response_

d34db3ff:
Fixes #3636 

# Summary

When `didOpen` triggers `populate_all_project_files_in_config`, discovered project files are processed at `Require::Indexing` level. 

Later, when `callHierarchy/incomingCalls` processes the reverse dependencies to find call sites, the `process_fn` closure needs the AST of the caller module containing the callsite. Since the populated files are at Require::Indexing level (thus no AST), calling `get_ast` on their handles returns `None`, and the caller is silently dropped from the results.

This patch fixes the issue by running all candidates to `Require::Everything`, after computing candidate handles and before returning them in `compute_transitive_rdeps_for_definition_impl`.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -3984,7 +3984,6 @@ fn compute_transitive_rdeps_for_definition_impl<T: RdepTransaction>(
                 sys_info,
             );
             let rdeps = transaction.transitive_rdeps(definition_handle.dupe());
-            transaction.run_for_handles(&[definition_handle], Require::Everything)?;
             rdeps
         }
     };
@@ -4002,10 +4001,12 @@ fn compute_transitive_rdeps_for_definition_impl<T: RdepTransaction>(
     {
         transitive_rdeps.remove(&fs_counterpart_of_in_memory_handles);
     }
-    let candidate_handles = transitive_rdeps
+    let candidate_handles: Vec<Handle> = transitive_rdeps
         .into_iter()
         .sorted_by_key(|h| h.path().dupe())
-        .collect::<Vec<_>>();
+        .collect();
+
+    transaction.run_for_handles(&candidate_handles, Require::Everything)?;
 
     Ok(candidate_handles)
 }
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -4815,7 +4815,6 @@ fn compute_transitive_rdeps_for_definition_impl<T: RdepTransaction>(
                 sys_info,
             );
             let rdeps = transaction.transitive_rdeps(definition_handle.dupe());
-            transaction.run_for_handles(&[definition_handle], Require::Everything)?;
             rdeps
         }
     };
@@ -4833,10 +4832,12 @@ fn compute_transitive_rdeps_for_definition_impl<T: RdepTransaction>(
     {
         transitive_rdeps.remove(&fs_counterpart_of_in_memory_handles);
     }
-    let candidate_handles = transitive_rdeps
+    let candidate_handles: Vec<Handle> = transitive_rdeps
         .into_iter()
         .sorted_by_key(|h| h.path().dupe())
-        .collect::<Vec<_>>();
+        .collect();
+
+    transaction.run_for_handles(&candidate_handles, Require::Everything)?;
 
     Ok(candidate_handles)
 }


yangdanny97:
There may be a perf impact from Require::Everything rather than Require::Indexing, someone needs to profile this one before merging

stroxler:
Thanks for the PR @d34db3ff!

Sorry it took a while to take a look, our review backlog is pretty big. Would you mind rebasing this?

DarkNightForge:
I hit this bug independently and wrote a fix before finding this PR, so rather than open a duplicate: here is the profiling @yangdanny97 asked for in June, plus one observation about placement.

## Profile

Release builds, 189-file Python project, LSP driven programmatically (opens the file it queries and nothing else), warm, median of 5 runs. `refs` are `textDocument/references`, ground-truth result counts in parentheses.

| | refs (73) | refs (36) | incomingCalls (32) | server peak RSS |
|---|---|---|---|---|
| `main` @ 43467e6 | 73.8 ms | 27.9 ms | 29.8 ms | 314 MB |
| this PR | 68.8 ms | 29.2 ms | 29.2 ms | 302 MB |

**No measurable regression** — the differences are inside run-to-run noise, and peak RSS is if anything slightly lower. At this project size the `Require::Everything` escalation does not cost anything detectable.

Caveat worth stating: one project, ~190 modules. A repo with a much larger reverse-dependency fan-out could look different, since what gets escalated here scales with the number of reverse dependencies rather than with the number of references.

## One observation about placement

The escalation lands in `compute_transitive_rdeps_for_definition_impl`, which is shared by `find_global_references_from_definition` and the implementation finder as well as call hierarchy. Those two are answered from the index and never need an AST, so they now pay for one they will not read.

That is exactly why it is invisible in the numbers above — the cost is small — but it does mean the blast radius is wider than the bug being fixed.

A narrower option, if it is ever worth it: keep the walk index-only, and inside the call-hierarchy path escalate just the files that turned out to reference the target. Two passes — ask the index which rdeps reference the target at all, then `run` that subset at `Require::Everything` before attributing call sites. Two things I got wrong on the way to that, in case they save someone time:

- ordering matters for cost — fetching the AST *before* checking whether the file has any references pulls up every rdep anyway;
- batching matters for correctness — escalating one file at a time from inside the walk leaves every file after the first still without an AST. That version returned 1 caller instead of 32, which looks like a working fix until you count.

I measured that variant too: 71.2 ms / 303 MB, i.e. also indistinguishable from both rows above. So on this evidence it is not worth complicating this PR for — I mention it only because the shared-helper placement is a real difference, not because the numbers justify changing course.

## Test coverage

`test_incoming_call_hierarchy_basic` cannot catch this regression: it opens both `callee.py` and `caller.py`, and does not set `IndexingMode::LazyBlocking`. A version that opens only the callee does catch it — on `main` it fails with `got: []`, and passes with this PR's change applied.

Happy to contribute that test here if useful, or leave it to you — your PR, your call. Either way this looks correct to me and the perf question looks answered.

d34db3ff:
> Would you mind rebasing this?

Hi, thank you for your reply! I have rebased this on the upstream.

d34db3ff:
> ## Profile
> 
> Release builds, 189-file Python project, LSP driven programmatically (opens the file it queries and nothing else), warm, median of 5 runs. `refs` are `textDocument/references`, ground-truth result counts in parentheses.
> 	refs (73) 	refs (36) 	incomingCalls (32) 	server peak RSS
> `main` @ [43467e6](https://github.com/facebook/pyrefly/commit/43467e64e36550f232a18e89f24fda79b1020b6b) 	73.8 ms 	27.9 ms 	29.8 ms 	314 MB
> this PR 	68.8 ms 	29.2 ms 	29.2 ms 	302 MB

Hi, thank you for your comment and for the profile!

> Happy to contribute that test here if useful, or leave it to you — your PR, your call. Either way this looks correct to me and the perf question looks answered.

Sure, thank you so much if you are willing to kindly add a regression test for this.

stroxler (approved):

DarkNightForge:
@d34db3ff — following up on the regression test you asked for above: this merged about four
hours after your comment, so I have opened it as #4643 against `main` instead.

`test_incoming_call_hierarchy_unopened_caller_file` opens only `callee.py` and runs with
`IndexingMode::LazyBlocking`, so `caller.py` is reached through the index.
`test_incoming_call_hierarchy_basic` opens both files, which is why it could never catch this.
I confirmed the new test passes against your fix on unmodified `main`, and fails against the
code as it stood before it.

Two other things in that PR that are worth flagging here rather than burying:

- I also went back and answered @yangdanny97's June perf question properly. **My profile in
  this thread was measured wrong** — median of 5 warm reps, and the escalation is a one-time
  state transition the transaction retains, so reps 2..5 measured the already-escalated steady
  state. It structurally could not see the cost being asked about, so it was not evidence that
  the escalation is cheap. Re-measured with a fresh server per rep, `textDocument/references`
  goes 65.4 ms -> 3.4 ms cold and retains 81 MB -> 2 MB, because `references` is answered
  entirely from the index and never reads an AST.
- That is not a criticism of the fix — it is correct, and #4643 keeps it. It just moves the
  escalation out of the shared helper into the two consumers that actually read ASTs.

Thanks for the original diagnosis; the `Require::Indexing` / `get_ast` framing in your summary
is what made the rest of this easy to find.

PR #3644, opened 2026-06-02
Treat `__getattribute__` param as implicitly typed in `pyrefly coverage report`

### LINKED ISSUE #3640: `pyrefly report` should treat the `__getattribute__`  parameter as implicitly typed
Currently (1.1.0.dev0), `pyrefly coverage report` shows the following for NumPy:

```json
{
  "kind": "function",
  "name": "numpy.ma.mrecords.MaskedRecords.__getattribute__",
  "n_typable": 2,
  "n_typed": 0,
  "n_any": 0,
  "n_untyped": 2,
  "location": {
    "line": 180,
    "column": 5
  }
},
```

That's technically correct because the `__getattribute__` method doesn't exist in the stubs ([src](https://github.com/numpy/numpy/blob/251958543e61d1cc8fe4b18fbd92e592eb17fa42/numpy/ma/mrecords.pyi)), but exists at runtime ([src](https://github.com/numpy/numpy/blob/251958543e61d1cc8fe4b18fbd92e592eb17fa42/numpy/ma/mrecords.py#L180-L225)).

The problem is that it reports **2** typables, even though the parameter is always a `str`. So `pyrefly[ coverage] report` should, just like it does for other dunders like `__getattr__`, not require an explicit type annotation for this `__getattribute__` parameter. The return type of `__getattribute__` isn't trivial (even though it's *usually* `Any`), so I think it's fine that that's counted as untyped here.

jorenham:
# Summary

This adds `__getattribute__` to the list of dunder methods whose parameter should be treated as implicitly typed (it's always `str`, after all).

Fixes #3640

# Test Plan

Updated the relevant test file.

### DIFF
--- a/pyrefly/lib/commands/report.rs
+++ b/pyrefly/lib/commands/report.rs
@@ -1463,8 +1463,8 @@ impl ReportArgs {
             // __exit__/__aexit__(self, exc_type, exc_val, exc_tb)
             "__exit__" | "__aexit__" => non_self_param_pos <= 2,
             // First non-self param is protocol-fixed (str, int, or memoryview)
-            "__getattr__" | "__delattr__" | "__setattr__" | "__format__" | "__buffer__"
-            | "__release_buffer__" => non_self_param_pos == 0,
+            "__getattr__" | "__getattribute__" | "__delattr__" | "__setattr__" | "__format__"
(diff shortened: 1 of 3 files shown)


grievejia (approved):

PR #3646, opened 2026-06-02
Trim dependency debug info in dev builds to shrink `target/`

jorenham:
# Summary

Most of `target/debug` is debug info, which usually isn't of interest when doing a `dev` build. 
This removes this debug info in `dev` builds, and instead adds a new `dbg` profile for debugging purposes.

# Test Plan

Running `cargo build -q -p pyrefly` before and after in a clean env showed that `target/debug` went from 4.3G to 2.6G. So that's a -1.7G / -39% reduction.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,6 +152,10 @@ and inspect variables. Many code editors,
 for breakpoints and variable watch. You can also use the command line debuggers
 like [lldb](https://docs.rs/lldb/latest/lldb/#installation):
 
+`dev` builds carry only line tables to keep `target/` small. When you need to step
+through code in a debugger, build with the `dbg` profile (`cargo build --profile dbg`)
+for full debug info.
+
 ## Making a Pull Request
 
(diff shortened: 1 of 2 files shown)


yangdanny97:
@samwgoldman

samwgoldman:
Hey @jorenham, thanks for the PR!

I don't have any problem with this. If 4.3G is big, I think 3.8G is also still pretty big.

Our internal build system distinguishes between dev (default) and dbg builds, with full debug information only included in dbg builds. Maybe we should go even further here and exclude full debug info for everything, while providing a dbg build profile for people who need it?

But if this change is enough to solve a problem for you, I'm OK taking this as-is.

jorenham:
> Maybe we should go even further here and exclude full debug info for everything, while providing a dbg build profile for people who need it?

Yea sounds good, I'll look into it

samwgoldman:
Great, thank you!

yangdanny97 (approved):

PR #3652, opened 2026-06-02
Only stop listing ancestors of directory relative search path if they're children of up_to

connernilsen:
Summary:
It was reported in the build system chat on Discord that supplying a config that isn't a parent of the files that are being checked results in the fallback search path working incorrectly.

It looks like a while ago, I didn't account for the fact that we might want to use the fallback search path on a file that exists outside of the config. This diff fixes that, and only uses `up_to` if the file is under the config.

Differential Revision: D107301498


kinto0 (approved):

PR #3662, opened 2026-06-03
Don't pin an empty container's element from a covariant read

### LINKED ISSUE #1584: Passing an empty list to `sum` results in an odd type being inferred for it
### Describe the Bug
```
x = []
y = sum(x)
# revealed type: list[Literal[-20, -19, -18, -17, -16, -15, -14, -13, -12, -11, -10, -9, -8, -7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25] | bool]
reveal_type(x)
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSI%2BABALxUDaAugDrqm1VwCuAtgBT4AlCAA0ILgBdocEuUQgAxFQCqUqBAnswXdAGMpudHFatMMMFTC4ATj1QSA%2Bul7YY1gYioR0EwVQC0AHycEtaIrFSRVNYwElzW6JbMIAByLm6ewPgAvsmsoiBkMWBQpIQSuDxQFEoACqTFpZwYOARUuoaQAObx9hCGhKxKAMowMFQAFhISxHCIAPTzRealhDZd8zDo85i4unDzHejdvQbbljZUqABuqNCo2LDtnRA91n2GVLjEZzKsZBIJoZ-Nc3HB%2Bok6MkAMyEACMACY8ugQNkxKh9BBQQAxaAwChoLB4IhkVFAA

_No response_

mikeleppane:
Fixes #1584

### What
Passing an empty list to `sum` (or any covariant reader) no longer pins the list's element type to the parameter's element.

```python
x = []
y = sum(x)
reveal_type(x)
# before: list[Literal[-20, -19, …, 24, 25] | bool]   (typeshed's sum-overload union)
# after:  list[Unknown]                                (y is int, unchanged)
```

### Why
`x = []` is a list whose element type isn't known yet. pyrefly's first-use inference looks at the first use to infer it — but it was treating a *covariant read* as evidence. Passing `x` where `Iterable[E]` is expected only asserts "whatever is in the list is `<: E`"; it says nothing about what the list **contains** (you can't write into an `Iterable`). So pinning the element to `E` is unsound — and for `sum`, `E` is typeshed's artificial `bool | _LiteralInteger` (= `Literal[-20..25]`) hack, which is how the nonsense union ended up on `x`. pyright, mypy, and ty all leave the element unsolved (`list[Unknown]` / `list[Any]`) — they are the conformance bar.

An empty container's element should come only from **lower** bounds: values written in (`append`), the equality an **invariant** parameter (`list[E]`) forces, or a generic type var solved against a value. Two code paths were pinning it from a read; both are fixed under that one principle.

### How
- **`pyrefly/lib/solver/solver.rs`** — in `is_subset_eq_var`, the `PartialContained` **upper-bound** arm (a covariant read) no longer pins; it leaves the placeholder unsolved. The **lower-bound** arm (writes / invariance) is untouched, so `append` and `list[int]`-style params still pin.
- **`pyrefly/lib/alt/overload.rs`** — `is_equivalent`/`is_subset_eq` are *side-effecting* (they record bounds and pin/unify vars). Overload disambiguation ran them **outside** any snapshot, leaking a pin onto the element while choosing which `sum` overload to materialize. A new `probe` helper runs those comparisons under `snapshot_vars`/`restore_vars` so disambiguation only observes; the authoritative pin still happens in the selected overload's contextual call.

**Deliberately unchanged:** writes (`append`), invariant parameters, and generic cross-argument inference (`gen[T](Iterable[T], y: T); gen([], 1)` → `list[int]`) all still pin via lower bounds. **Scope:** the fix covers empty container *literals* (`[]`, `{}`); `set()` is a constructor call on a different placeholder path and is left as a documented `bug`-marked gap.

### Test plan
| Test | Protects |
|------|----------|
| `delayed_inference::test_sum_empty_list` | the repro: `x` is `list[Any]`, `y` is `int` |
| `delayed_inference::test_empty_list_invariant_param` | invariant `list[int]` param still pins `list[int]` |
| `delayed_inference::test_empty_list_generic_cross_arg` | generic cross-arg `gen([], 1)` still pins `list[int]` (regression guard) |
| `delayed_inference::test_empty_list_covariant_read_stays_any` | covariant read → `list[Any]` (intended change) |
| `delayed_inference::test_empty_list_covariant_read_then_write` | read-then-write → `list[Any]` (existing first-use limitation; was a false positive before) |
| `delayed_inference::test_empty_dict_covariant_read_value_unpinned` | `{}` value unpinned, key pinned via Mapping's invariant key |
| `delayed_inference::test_empty_set_covariant_read_still_pins` | `bug=` marker documenting the `set()` constructor-path gap |
| `inference::test_implicit_any_empty_container_covariant_read` | with `ImplicitAnyEmptyContainer` enabled, the diagnostic now fires |
| `literal::test_partial_inference_literalstring_join` | `bug=` removed: `", ".join([])` now `str`, not `LiteralString` |

- Full lib suite: **passing, 0 failed**.
- Conformance: regenerated, **0 moved results**.
- Format + lint: clean.
- mypy_primer (broad ecosystem set: pydantic, pandas, pandera, attrs, rich, sympy, aiohttp, sphinx, tornado, trio, werkzeug, black, mypy, comtypes, spark, discord.py, …): **0 errors added, 0 removed** — only 2 cosmetic element-type display shifts within pre-existing errors (`attrs` `dict[str, object]`→`dict[str, Unknown]`; `discord.py` `list[int]`→`list[@_]`).

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -634,7 +634,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         return true;
                     };
                     for t in param_types {
-                        if !self.is_equivalent(first, t) {
+                        if !self.probe(first, t, &placeholder_vars, |s| s.is_equivalent(first, t)) {
                             return true;
                         }
                     }
@@ -689,9 +689,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 }
             }
             let selected_overload = if spec_compliant {
-                self.disambiguate_overloads_spec_compliant(&matched_overloads)
+                self.disambiguate_overloads_spec_compliant(&matched_overloads, &placeholder_vars)
             } else {
-                self.disambiguate_overloads(&matched_overloads)
+                self.disambiguate_overloads(&matched_overloads, &placeholder_vars)
             };
             if let Some(idx) = selected_overload {
                 let overload = matched_overloads
@@ -744,6 +744,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     fn disambiguate_overloads_spec_compliant(
         &self,
         matched_overloads: &[CalledOverload<'_>],
+        placeholder_vars: &[Var],
     ) -> Option<usize> {
         // Step 5, part 2: are all remaining return types equivalent to one another?
         // If not, the call is ambiguous.
@@ -751,14 +752,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let first_overload = matched_overloads
             .next()
             .expect("Expected at least one overload");
-        if matched_overloads.any(|o| !self.is_equivalent(&first_overload.res, &o.res)) {
+        if matched_overloads.any(|o| {
+            !self.probe(&first_overload.res, &o.res, placeholder_vars, |s| {
+                s.is_equivalent(&first_overload.res, &o.res)
+            })
+        }) {
             return None;
         }
         // Step 6: if there are still multiple matches, pick the first one.
         Some(0)
     }
 
-    fn disambiguate_overloads(&self, matched_overloads: &[CalledOverload<'_>]) -> Option<usize> {
+    fn disambiguate_overloads(
+        &self,
+        matched_overloads: &[CalledOverload<'_>],
+        placeholder_vars: &[Var],
+    ) -> Option<usize> {
         // When a call to an overloaded function may match multiple overloads, the spec says to
         // return Any when the return types are not all equivalent.
         // However, neither mypy nor pyright fully follows this part of the spec, and many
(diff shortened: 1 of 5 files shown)


rchen152 (changes requested):
Thank you for the PR. I think it's identified two real issues, in Pyrefly's solver and overload evaluation, but the fixes don't look quite right to me.

* I understand the intuition of "write should pin, read should not", but skipping upper bound pinning is only an approximate heuristic for that, which only works for covariant containers. I think it'll be confusingly inconsistent if we pin on read for invariant containers but not covariant ones. IMO we should either (1) rework the solver so it can track which side the Var was originally on, for a better approximation of read vs. write, or (2) keep upper bound pinning but promote literals. (2) would be a smaller change, and I'd be happy to accept it if you open a separate PR for that.
* As far as I can tell, although the issue in the solver is real, it's *not* the root cause of the `sum([])` bug. Ultimately, the problem is that we're selecting the first overload of `sum` (https://github.com/facebook/pyrefly/blob/10c6b93d121f2dc402e18764ae3edf9b0422fc48/crates/pyrefly_bundled/third_party/typeshed/stdlib/builtins.pyi#L6372), even though we ought to treat an empty list as having type `list[Any]` for the purpose of overload evaluation, which should lead to us concluding that the first two overloads both match and then disambiguating based on the return types. The "probe" mechanism this diff adds is working around the var pinning done by the incorrect selection, but we should actually fix the selection process.

rchen152:
I'm going to close this PR since it's stale and has some merge conflicts, but please feel free to open a new one if you'd like to take another stab at this, or comment on the issue or in Discord to kick some ideas around :)

PR #3667, opened 2026-06-03
configuration.mdx: Add missing backticks

SamB:
# Summary

I spotted a couple of links that were a bit garbled due to missing one of two backticks, and after I fixed those, I found and fixed one unpaired backtick in the rendered page outside the code blocks (which I didn't really check, because even paired backticks show up in code blocks, and anyway this doesn't screw up the page formatting).

I might have missed some though, because there are a lot of backticks in code blocks.

# Test Plan

I tested this using github's "preview" tab (with **diff** mode **OFF**) and my eyes (diff mode rendering gets garbled by the unpaired backticks in the old markdown). Further testing will involve reading the generated documentation.

If `test.py` was effective in such matters, surely it would already have caught these?

### DIFF
--- a/website/docs/configuration.mdx
+++ b/website/docs/configuration.mdx
@@ -1098,13 +1098,13 @@ you've configured yourself, either through CLI flags or a config file.
 We look for an interpreter with the following logic:
 
 1. Use [`python-interpreter-path`](#python-interpreter-path),
-   [`fallback-python-interpreter-name](#fallback-python-interpreter-name), or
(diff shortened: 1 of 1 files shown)


SamB:
To be clear, the reason I might have missed some other unpaired backticks is because I had to hit "next" so many times to skip the backticks in the code blocks that I might have accidentally skipped one in the prose.

stroxler (approved):

stroxler:
Thanks!

yangdanny97 (approved):

PR #3672, opened 2026-06-03
[TSP] Return contextual expected type from getExpectedType

### LINKED ISSUE #8060: [TSP] Expected type for TSP is returning the computed type
Have this code here:

```python
# bring up string literal completion inside of `""` and confirm 2 literals are suggested
e: Literal["Hello", "There"] = ""
```

And query Pyrefly for the expected type of `e`. It will return a str literal with the value "" instead of the expected type.

rchiodo:
# Summary

`typeServer/getExpectedType` was returning the inferred (computed) type at the cursor instead of the contextually expected type that the position imposes. For example, in `x: int | str = 4`, querying on `4` returned `int` rather than `int | str`.

This change makes `get_expected_type_at_position` walk the covering AST nodes innermost-first and, when the cursor lies in a context that supplies an expected type, look up that contextual type. Recognized contexts:

- **Call argument** (positional or keyword) — uses the parameter type declared by the callee, via `Transaction::expected_call_argument_type` (now `pub(crate)` so the non-WASM server can call it).
- **`return <expr>`** — uses the enclosing function's return annotation (resolved via `get_type_trace` over the annotation's full range so unions like `int | str` are preserved instead of collapsing to the leading identifier).
- **Parameter default `def f(x: T = <expr>)`** — uses the parameter's annotation `T`.
- **Annotated assignment RHS `x: T = <expr>`** — uses the LHS declared type `T`.

Other contexts (yield expressions, container-literal elements under an annotation, etc.) still fall back to the inferred type at the cursor; tests document that behavior so future fixes can flip the expectations.

Fixes microsoft/pylance-release#8060

# Test Plan

Added 8 new tests in `pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs` covering each newly supported context plus the documented fallbacks:

- `test_get_expected_type_call_positional_argument`
- `test_get_expected_type_call_keyword_argument`
- `test_get_expected_type_return_statement_value`
- `test_get_expected_type_default_parameter_value`
- `test_get_expected_type_yield_falls_back_to_value_type`
- `test_get_expected_type_list_element_falls_back_to_container_type`
- `test_get_expected_type_dict_value_falls_back_to_container_type`

All 14 `test_get_expected_type*` tests pass:

```
cargo test -p pyrefly --lib test_get_expected_type
test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured
```

Also ran `python3 test.py --no-test --no-conformance --no-jsonschema` — formatting and linting clean. `cargo clippy --all-targets --all-features` produces no new warnings (only 2 pre-existing unused-import warnings in `pyrefly_lsp_interaction_tests`, unrelated to this branch).

### DIFF
--- a/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
+++ b/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
@@ -341,6 +341,93 @@ fn test_get_expected_type_stale_snapshot() {
     tsp.shutdown();
 }
 
+/// Helper to send a getExpectedType request and return a successful (non-null) result.
+fn get_expected_type_ok(
+    tsp: &mut TspInteraction,
+    file_uri: &str,
+    line: u32,
+    character: u32,
+    snapshot: i32,
+) -> serde_json::Value {
+    tsp.server
+        .get_expected_type(file_uri, line, character, snapshot);
+    let resp = tsp.client.receive_response_skip_notifications();
+    assert!(
+        resp.error.is_none(),
+        "Expected success, got error: {:?}",
+        resp.error
+    );
+    let result = resp.result.expect("Expected result");
+    assert!(!result.is_null(), "Expected non-null type result");
+    result
+}
+
+#[test]
+fn test_get_expected_type_on_rhs_returns_lhs_annotation_union() {
+    // For `x: int | str = 42`, the expected type at the RHS position is
+    // the union `int | str` declared on the LHS, NOT the inferred type
+    // (Literal[42]) of the RHS expression.
+    //                          1111111
+    //                0123456789012345678
+    let source = "x: int | str = 42\n";
+    let (mut tsp, file_uri, snapshot) = setup_project(source);
+
+    // Position 15 is the `4` in `42`.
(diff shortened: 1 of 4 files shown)


pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
```diff
@@ -341,6 +341,220 @@ fn test_get_expected_type_stale_snapshot() {
...
+
+#[test]
+fn test_get_expected_type_on_rhs_returns_lhs_annotation_union() {
+    // For `x: int | str = 42`, the expected type at the RHS position is
+    // the union `int | str` declared on the LHS, NOT the inferred type
+    // (Literal[42]) of the RHS expression.
+    //                          1111111
+    //                0123456789012345678
+    let source = "x: int | str = 42\n";
+    let (mut tsp, file_uri, snapshot) = setup_project(source);
```
kinto0, line 373: nit: we probably don't need an integration test for these. we can probably just use a normal unit test

kinto0:
discussed offline. closing

PR #3675, opened 2026-06-03
Fix ParamSpec prefix params incorrectly demoted to positional-only

thatfunkymunki:
# Summary

When I am using the [`pyinfra` library](https://docs.pyinfra.com/en/3.x/using-operations.html#global-arguments) , it fails pyrefly type check with these errors:
```
ERROR Unexpected keyword argument `name` in function `pyinfra.api.arguments_typed.PyinfraOperation.__call__` [unexpected-keyword]
ERROR Unexpected keyword argument `_sudo` in function `pyinfra.api.arguments_typed.PyinfraOperation.__call__` [unexpected-keyword]
```
and similar ones. MyPy correctly recognizes the types here successfully.

prepend_types() was converting PrefixParam::Pos to Param::PosOnly via to_param(), making keyword-passable params from function definitions positional-only after ParamSpec expansion. This caused keyword arguments like _sudo=True to be rejected as "unexpected keyword" when calling through a Protocol with explicit params before *args: P.args.

Switch to to_subset_param() which preserves the Pos vs PosOnly distinction. PrefixParam::Pos is only ever created from function definitions (not from Concatenate syntax, which always uses PosOnly), so this change is safe for all callers.

Repro file is here

https://gist.github.com/thatfunkymunki/c0c8a57d60f0f02217b918ef6a377189

observe additional unexpected kwarg errors in pyrefly vs mypy

### pyrefly
```
PS C:\Users\munki> uvx pyrefly check repro.py 
ERROR Unexpected keyword argument `_bar` in function `FooProtocol.__call__` [unexpected-keyword]                                                                                                                                   
  --> repro.py:24:16
   |
24 | foo(biz="ccc", _bar=True)
   |                ^^^^
   |
ERROR Unexpected keyword argument `_bar` in function `FooProtocol.__call__` [unexpected-keyword]
  --> repro.py:25:26
   |
25 | foo(biz="ddd", baf=True, _bar=True, baz="abcdef")
   |                          ^^^^
   |
ERROR Unexpected keyword argument `baz` in function `FooProtocol.__call__` [unexpected-keyword]
  --> repro.py:25:37
   |
25 | foo(biz="ddd", baf=True, _bar=True, baz="abcdef")
   |                                     ^^^
   |
ERROR Unexpected keyword argument `nonexistent_param` in function `FooProtocol.__call__` [unexpected-keyword]
  --> repro.py:27:5
   |
27 | foo(nonexistent_param=True)
   |     ^^^^^^^^^^^^^^^^^
   |
 INFO 4 errors
No `pyrefly.toml` found — using preset `basic`.
Run `pyrefly init` to continue setting up Pyrefly.
Docs: https://pyrefly.org/getting-started-cli
```
### mypy
```
PS C:\Users\munki> uvx mypy --strict repro.py
repro.py:27: error: Unexpected keyword argument "nonexistent_param" for "__call__" of "FooProtocol"  [call-arg]                                                                                                                    

```
# Test Plan

Created a failing test and confirmed that it passes with my changes. I also observed that some of the existing tests were expecting incorrect result which was also resolved with the changes.

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -240,7 +240,7 @@ impl ParamList {
         } else {
             Cow::Owned(ParamList(
                 pre.iter()
-                    .map(|p| p.to_param())
+                    .map(|p| p.to_subset_param())
                     .chain(self.0.iter().cloned())
                     .collect(),
             ))
--- a/pyrefly/lib/test/callable_residuals.rs
+++ b/pyrefly/lib/test/callable_residuals.rs
@@ -307,7 +307,6 @@ assert_type(b.fn(1), int)
 );
 
 testcase!(
-    bug = "ParamSpec prefix params incorrectly demoted to positional-only",
     test_callable_class_wrapper,
     r#"
 from typing import Callable, assert_type, reveal_type
@@ -328,7 +327,6 @@ assert_type(wrapper(1), int)
 );
 
 testcase!(
-    bug = "ParamSpec prefix params incorrectly demoted to positional-only",
     test_callable_class_wrapper_with_helper,
     r#"
 from typing import Callable, assert_type, reveal_type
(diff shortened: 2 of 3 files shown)


rchen152 (approved):

PR #3683, opened 2026-06-04
Treat Protocol bases as ABCMeta-providing when detecting abstract cla…

### LINKED ISSUE #2797: final class implementing `Sequence` shows no error when abstract method is not implemented
### Describe the Bug

```python
from typing import final
from collections.abc import Sequence, Collection, Reversible

@final
class A[T](Sequence[T]): # No error: should have error
    ...

@final
class B[T](Collection[T], Reversible[T]): # error as expected
    ...


class A1(ABC):
    @abstractmethod
    def a(): ...

class B1(ABC):
    @abstractmethod
    def b(): ...

class C(A1, B1): ...

@final
class D(C): ... # error as expected
```

Python: 3.12

Inheriting two abstract class itself is working as inteded, but inheriting `collections.abc.Sequence` shows no error, while inheriting both `Collection` and `Reversible` (which consists `Sequence` according to [python document](https://docs.python.org/3.14/library/collections.abc.html#collections-abstract-base-classes)) does show error.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAOutdWAE64C2ABAC6nEToDmLETYrnpsWkdKih1GrVNgDG-QcNEBBAEIBhADQs5cNvVTy2TGGwAWuTNOYt5uKLBMRc6OITmKBQkSwDKMACOAK4w6PIwupqOzmyu6LoASjAAbjD0cBDYsLToAALiktTyUKhwcCyqANoAKgC6ABSBoeGRdfUAlIgsAMQsAHK4LBmM9D1wViFQmCwWqOkj9GPULGsshJt5hdzFEWUVLOodjTFOMC5uHclpGVk5MB3dfUtjepUw%2BMQXbDA26OsNlsaCDSuVKqoAIyNDSabqrdb5fSGYymcxWf6AzAwMB6RrPTaEPJgw7qaGw%2BEAxHIowmMyWawItbY3HYfE9QnEg6VTQwyG6MkE4HUHYSKT7cEsAAipyFhBeo2E7xGXx%2BfxA2hAIXiUA8ZAo-QAqjqIBwxCEIvE3HA8iyxMImKg2AB9dAhJjYDKNfA9bhsTosAC0AD4WAZxkyWPRzCF6ACwJQQAN3Z7xixgPgAL6J6gakBkaNgKCkQhsZhQA0sAAKpELxbDGBwBHsbkgPFjToSRPQ-UCMDmbDYxDgiAA9KOCzji4RhDxR%2BFR5hcPI4KOHOg2x2rehR-b6HpUqhoHJYC2NxB20ZtyxcMRtx5qGQGehA%2BlMgkWABeFiJgDMhEhAAmHN0BATNNVRCB0gAMWgGAKDQLA8CIfVMyAA

_No response_

mikeleppane:
Fixes #2797

### What
A `@final` class that inherits `collections.abc.Sequence` (or `Mapping`) without
implementing its abstract methods is now flagged, matching mypy, pyright, and ty.

```python
from typing import final

@final
class A(Sequence[int]):  # before: no error
    ...                  # after:  Final class `A` cannot have unimplemented
                         #         abstract members: `__getitem__`, `__len__`
```

Instantiating such a class (`A()`) and the opt-in implicit-abstract-class
diagnostic now fire correctly too.

### Why
In typeshed the small `collections.abc` ABCs (`Reversible`, `Collection`, …) are
declared `Protocol`, and `Protocol` is a special form (`Protocol: _SpecialForm`)
that carries no metaclass. pyrefly's `extends_abc` heuristic only recognized an
`abc.ABC` base or an `ABCMeta` metaclass, so it never saw the `_ProtocolMeta(ABCMeta)`
that a protocol subclass actually gets at runtime. The composite `Sequence` — a
plain class whose only bases are those protocols — was therefore misclassified as
non-abstract and skipped, so its `@final` subclasses inherited no abstract
obligation. (Direct protocol bases like `class B(Collection, Reversible)` already
worked; the gap was one level of indirection away.) mypy, pyright, and ty are the
conformance bar and all three error here.

### How
- **`pyrefly/lib/alt/class/class_metadata.rs`** — `extends_abc` now treats an
  `is_protocol()` base as ABCMeta-providing (a protocol base ⇒ `_ProtocolMeta` ⇒
  ABCMeta-derived metaclass). This is the whole fix for the reported bug.
- **Same file** — `calculate_abstract_members` short-circuits for `is_typed_dict()`
  classes. Recognizing `Mapping` as abstract otherwise leaked through the
  synthesized `TypedDictFallback(Mapping, metaclass=ABCMeta)` base and made every
  TypedDict look abstract; a TypedDict is a concrete `dict` and never is. Caught by
  mypy_primer on `pandas-stubs`, not the unit suite.
- **Deliberately untouched:** the two abstract-member gates (already correct given a
  correct `extends_abc`) and the truthiness-redundancy consumer of `extends_abc` in
  `expr.rs` (widening it only suppresses more "always true" warnings on abstract
  types — conservative).

### Test plan
| Test (`pyrefly/lib/test/abstract_methods.rs`) | Protects |
|---|---|
| `test_final_class_over_sequence` | the repro: `@final` over `Sequence` errors |
| `test_abstract_through_protocol_bases` | instantiation, `Mapping`, and a user `Protocol` reached via a plain intermediate |
| `test_implicit_abstract_through_sequence` | opt-in implicit-abstract diagnostic on `Sequence` subclasses |
| `test_concrete_subclass_of_sequence_is_clean` | implementing the members stays clean — no over-reporting |
| `test_final_class_direct_protocol_bases_still_errors` | the pre-existing direct-protocol path is unchanged |
| `test_typed_dict_is_not_abstract` | TypedDicts stay concrete (regression guard) |

- Full lib suite: **5340 passed, 0 failed**.
- Conformance: **pass**, no generated changes.
- `mypy_primer`, wide scope (~37 projects incl. `pandas-stubs`, `pandas`, `sympy`, `xarray`): **zero diff**.
- format + lint: clean.


rchen152:
I'll grab this one, since I already started thinking about it. (I actually started on a fix before noticing there was a PR open XD)

rchen152:
Thanks for the PR! Unfortunately, I realized while testing it that it doesn't play well with typeshed, because typeshed sometimes marks concrete classes as inheriting from protocols even when they don't at runtime. For example, if you look at this new mypy_primer finding:
```diff
scipy-stubs (https://github.com/scipy/scipy-stubs)
+ ERROR scipy-stubs/stats/_typing.pyi:51:7-16: Class `BaseBunch` has unimplemented abstract members: `__new__`, `__init__` [implicit-abstract-class]
```
the corresponding code in scipy-stubs is:
```
@type_check_only
class BaseBunch[*Ts](tuple[*Ts], BunchMixin[tuple[*Ts]]):
    @abc.abstractmethod
    def __new__(_cls) -> Self: ...
    @abc.abstractmethod
    def __init__(self, /) -> None: ...
```
(https://github.com/scipy/scipy-stubs/blob/master/scipy-stubs/stats/_typing.pyi#L51)

The new finding is due to pyrefly treating `tuple` as having `_ProtocolMeta` as its metaclass (because typeshed marks `tuple` as inheriting from `Sequence`).

To avoid this, I put together an alternative fix that modifies `calculate_abstract_members` instead of `extends_abc`: https://github.com/facebook/pyrefly/pull/4396.

PR #3687, opened 2026-06-04
Update stale symlink (#3687)

rchen152:
Summary:

Updates a stale symlink that was causing scrut test failures on GitHub

Differential Revision: D107562708


stroxler (approved):

PR #3691, opened 2026-06-04
fix new rule - missing super() call  #3681

### LINKED ISSUE #3681: new rule - missing `super()` call
pyright's `reportMissingSuperCall` reports an error when a `__init__`, `__new__` or `__init_subclass__` method does not call the `super()` method:

```py
class Foo:
    def __init__(self):
        self.value: int = 1


class Bar(Foo):
    def __init__(self): # Method "__init__" does not call the method of the same name in parent class  (reportMissingSuperCall)
        pass

Bar().value # runtime crash
```

asukaminato0721:
# Summary

Fixes #3681

Detects direct `super(...).__init__()`, `super(...).__new__()`, and `super(...).__init_subclass__()` calls during binding.

Emits missing-super-call when those methods override a non-object parent method without calling the parent method.

Keeps the rule ignored by default, with config/docs support.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -263,6 +263,8 @@ pub enum ErrorKind {
     MissingSource,
     /// We are using bundled stubs for a package but the source code is missing.
     MissingSourceForStubs,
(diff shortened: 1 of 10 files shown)


connernilsen (approved):

PR #3693, opened 2026-06-05
fix option to treat ALL_CAPS variables as Final #3678

### LINKED ISSUE #3678: option to treat `ALL_CAPS` variables as `Final`
pyright's `reportConstantRedefinition` rule treats uppercase variables as `Final`:

```py
FOO = 1
FOO = 2 # error: "FOO" is constant (because it is uppercase) and cannot be redefined  (reportConstantRedefinition)
```

perhaps pyrefly should support this too

asukaminato0721:
# Summary

Fixes #3678

Added treat-all-caps-as-final / --treat-all-caps-as-final, wired it through config and state, and added the bind-time reassignment check.

# Test Plan

add test


yangdanny97:
This change seems reasonable; I'd be curious to see how many errors mypy primer reports if this is on-by-default.

It might help us catch more edge cases that we wouldn't see from unit tests

connernilsen (approved):
Actually, I'm wondering if the idea here is that any use of `ALL_CAPS` variables should be marked as final, rather than just top-level ones. In that case, we should leave this disabled by default.

I can pull this in and make the change though. Thanks for your work here @asukaminato0721!

stroxler (approved):

PR #3694, opened 2026-06-05
Fix module-name resolution producing invalid names that break relative imports

tobyh-canva:
## Why

A single file can legitimately be reachable under more than one module name — once for each configured root (`search-path`, `site-package-path`, import root) it sits beneath. When pyrefly derives a module name for a file, it takes the *first* matching root and uses it without checking that the resulting name is actually a valid Python module name.

This goes wrong in a very common setup: a virtualenv. The interpreter directory in a venv (`python3.13`, `python3.10`, …) contains a `.`, so when a file under it matches the `search-path` root first, pyrefly produces a name like `python3.13.pkg.model`. That name can never round-trip — import resolution maps a dotted name back to directory separators, so `python3.13` would be looked up as `python3/13/`, which doesn't exist.

The file→name direction being corrupted is subtle: ordinary imports of the package still succeed, but the package's *own* relative imports resolve against the bogus name and break.

## User impact

For affected files (most often third-party packages installed in a venv's `site-packages`):

- **Relative imports report false `missing-import` errors.** `from .base import Base` fails because pyrefly looks for a parent module (`python3.13.pkg.base`) that doesn't exist.
- **Stub generation produces incomplete output.** Because the relative import doesn't resolve, inherited members are dropped — e.g. a dataclass loses its inherited fields and a synthesized `__init__` is missing parameters.
- **Downstream type checking breaks as a result.** A Pydantic model whose package uses relative imports loses its synthesized `def __init__(self, *, ...)`, and callers then get spurious `Missing argument` errors.

The same file is usually also reachable under a *valid* name through another root (e.g. `pkg.model` via the `site-package-path`). The fix makes pyrefly discard candidate names whose path components aren't valid module components and fall through to a valid one, so relative imports and stub synthesis work correctly.

### DIFF
--- a/crates/pyrefly_python/src/module_name.rs
+++ b/crates/pyrefly_python/src/module_name.rs
@@ -477,7 +477,8 @@ impl ModuleName {
         /// `service/config.cinc` → `service.config.cinc`. For standard Python
         /// files, the extension is stripped: `foo/bar.py` → `foo.bar`.
         fn path_to_module(mut path: &Path, extra_extensions: &[String]) -> Option<ModuleName> {
-            if path.file_stem() == Some(dunder::INIT.as_str().as_ref()) {
+            let is_init = path.file_stem() == Some(dunder::INIT.as_str().as_ref());
+            if is_init {
                 path = path.parent()?;
             }
             // Check if the file has an extra extension. For these files, the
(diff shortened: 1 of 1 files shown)


connernilsen (approved):

PR #3698, opened 2026-06-05
Fix/narrowing unreachable branch 3349

### LINKED ISSUE #3349: Pyrefly does not eliminate conditional expression branches based on previous narrowing
### Describe the Bug

```python
def f(x: int | None) -> None:
    if x is not None:
        y = 1 if x is not None else None # since we guard that x is not None, we could prune away the `else None` case in the ternary
        assert_type(y, int)
```

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABT6I0ToAuNAPjQHK7owAlDQC0APl78YiADroaClrXws4NdLg58Bs%2BYv2kaAXhoBGJTRUQ1GrVJowocGJIFz9%2B1HGcAnNgH02UmIYelIAGhZ2QRBwkDIfaihSQjZcAFsoCgBiGgAFUkSwZJo0LDwVAGN%2BSABzAFcfVDYIfkI5XIBlGBcACzY2YjhEAHoRhKSU3B9akZh0EcxcSrgR6vQ6xubWhbppmlQAN1RoVGxYGnXNppb%2BGlxiW-Q4dvQyNl7%2BEUOYHzgd4w0GQgADMhFMACZgXIANq-HzTOAAXTk9XQEHSxGmbBgmBEmAgiUqLR%2BgIA5BBahpEmS5LYRIkAI71Qm4kQAaxgpBEqEqlXgahMZIA7qgfOhaegQABfOK8kkwABi0BgFDKOAIJHI0qAA

_No response_

mikeleppane:
Fixes facebook/pyrefly#3349.

### Problem

A branch whose reachability is already decided by prior narrowing was kept in the result type, and a dead *missing* branch falsely triggered "possibly uninitialized":

```python
def f(x: int | None) -> None:
    if x is not None:                    # x is `int` here
        y = 1 if x is not None else None # `else None` can never run
        reveal_type(y)                   # was: Literal[1] | None  →  now: Literal[1]

def g(x: int) -> None:
    if x is not None: y = 1
    else:             y = "s"            # dead branch
    reveal_type(y)                       # was: Literal['s', 1]    →  now: Literal[1]

def h(x: int) -> None:
    if x is None: pass                   # dead branch (never runs)
    else:         y = 1
    print(y)                             # was: "y may be uninitialized"  →  now: clean
```

### Approach

`Never` is the reachability signal: a branch whose entry narrowing solves to `Never` (e.g. `int & None`) cannot run. The narrowed-subject binding is already created and solved, so deciding this is a cache hit, not new work.

Reachability is only known at **solve** time, but *which* branches merge is fixed at **binding** time, so the signal is recorded as a key at binding time and evaluated at solve time — the same idiom pyrefly already uses for `NoReturn` tails. Ternaries and `if`/`elif`/`else` merge through two different paths, hence two commits:

- **`63daca631` — conditional expressions.** A range-keyed `Key::ConditionalReachability` side-channel carries each branch's narrow idxs from binding to solving; the inline `Expr::If` solver drops the dead side (falling back to `union` when the channel is absent, e.g. ternaries in type-annotation positions).
- **`bd4a82b2d` — `if`/`elif`/`else`.** Each branch records its entry-narrow idxs (`BranchInfo.dead_if_any_never`); the existing Phi termination filter now drops a branch that terminates **or** has a vacuous narrowing.

### Soundness gate (identity-vs-`None` only)

Pruning fires **only** for pure `is None` / `is not None` (`NarrowOp::tests_identity_none`, every conjunct of an `and`/`or` must qualify). An equality narrow can also reach `Never` but its branch must stay live:

```python
1 if (x is not None and x == "c") else 2   # Literal[1, 2] — kept
```

`assert x is not None` feeds this; `assert x > 0` does not (a comparison on `int` stays `int`).

### Uninitialized-variable accounting: a combined deferral gate

The dead-branch signal also lets a dead `if`-body stop a variable looking possibly-uninitialized. Rather than *replace* the pre-existing `NoReturn`-count heuristic (which over-flagged `try`/`except`/`finally` and loops), the deferral gate now **combines** them:

```rust
let deferrable =
    all_missing_deferrable                                   // every missing branch is dead, OR
    || n_missing_branches <= n_branches_with_termination_key; // the existing NoReturn heuristic
```

The first disjunct only **adds** deferral — it can never newly flag a variable — so the dead-branch improvement is gained without regressing the cases the old heuristic already deferred:

```python
def k(cls, modname) -> str:
    try:
        qualname = get(cls)   # assigned before the only raising call
        analyze(modname)
    except Err:
        pass
    return qualname           # not flagged (matches mypy/pyright)
```

### Tests

- `narrow.rs` — 9 ternary cases (prune after guard, concrete dead body/else, `Any` subject kept, equality kept, mixed `and`-equality kept, type-annotation position must not panic).
- `flow_branching.rs` — `if`/`elif`/`else` pruning, dead-missing-branch defined, reachable-missing-branch still flagged, plus two regression guards for the try/except and try/finally false positives.
- `scope.rs` — one assertion adjusted.

### Performance

No new solving passes or fixpoint iterations. Added work is bookkeeping (a small `Vec` of narrow idxs per branch, one range-keyed binding per ternary, a usually-empty boxed slice per branch). The deadness check reads an already-solved narrow binding and short-circuits on branches with no `is None` narrow.

### Validation

- Full unit suite green; format/lint clean; conformance unchanged.
- Wide `mypy_primer`: removes false positives (e.g. PyGithub `bad-return`, sphinx/jax/meson `unbound-name`) and introduces no `unbound-name` regressions.

### Known interaction (`prefect`, not a regression in this change)

In `prefect/tasks.py`, `cache_policy`'s type has no `None`, so `cache_policy is None` is provably false and the ternary's `None` branch is **correctly** pruned — sharpening one assignment from `_None | None` to `_None`. This surfaces a **pre-existing** attribute-type-inference limitation (pyrefly infers `self.cache_policy` from the first conditional assignment instead of the explicit `Union[…, None]` annotation), producing a spurious `bad-assignment`. The same lines already errored before this change (as `_None | None`) and mypy co-reports them. The fix belongs in attribute-type inference and is out of scope here; tracked separately.

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -13,6 +13,7 @@ use std::slice;
 use dupe::Dupe;
 use itertools::Either;
 use itertools::Itertools;
+use pyrefly_graph::index::Idx;
 use pyrefly_python::ast::Ast;
 use pyrefly_python::dunder;
 use pyrefly_python::module_name::ModuleName;
@@ -78,6 +79,7 @@ use crate::binding::binding::KeyYield;
 use crate::binding::binding::KeyYieldFrom;
 use crate::binding::binding::LambdaParamId;
 use crate::binding::narrow::AtomicNarrowOp;
+use crate::binding::narrow::NarrowOp;
 use crate::binding::narrow::int_from_slice;
 use crate::config::error_kind::ErrorKind;
 use crate::error::collector::ErrorCollector;
@@ -445,7 +447,35 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 match self.as_bool(&condition_type, x.test.range(), errors) {
                     Some(true) => body_type,
                     Some(false) => orelse_type,
-                    None => self.union(body_type, orelse_type),
+                    None => {
+                        // Drop a branch whose narrow solves to `Never` (e.g. `1 if x is not
+                        // None else None` when `x: int`), via the reachability channel keyed
+                        // on `x.range`. The channel is absent for ternaries inferred outside
+                        // the expression-binding path (e.g. a type annotation); keep both.
+                        let dead = self
+                            .bindings()
+                            .key_to_idx_hashed_opt(Hashed::new(&Key::ConditionalReachability(
+                                x.range,
+                            )))
+                            .map(|idx| match self.bindings().get(idx) {
+                                Binding::ConditionalReachability(narrows) => {
+                                    let (body, orelse) = narrows.as_ref();
+                                    (
+                                        self.any_narrow_is_never(body),
+                                        self.any_narrow_is_never(orelse),
+                                    )
+                                }
+                                _ => unreachable!(
+                                    "ConditionalReachability key maps to ConditionalReachability binding"
+                                ),
+                            });
+                        match dead {
+                            Some((false, true)) => body_type,
+                            Some((true, false)) => orelse_type,
+                            // No binding, or both live, or both dead: keep the union.
+                            _ => self.union(body_type, orelse_type),
+                        }
+                    }
                 }
             }
             Expr::BoolOp(x) => self.boolop(&x.values, x.op, hint, errors),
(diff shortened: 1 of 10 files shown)


rchen152:
@stroxler this looks like it's in your wheelhouse (narrowing, reachability)

PR #3702, opened 2026-06-05
New `pyrefly coverage check` command

jorenham:
# Summary

This introduces a new `pyrefly coverage check` command as a user-friendly dual to `pyrefly coverage report`. 

It explicitly supports the following flags:

```
  -f, --fail-under <PERCENT>
          Minimum coverage percentage; exit non-zero when coverage is below it
          
          [default: 100]

      --prefer-stubs <PREFER_STUBS>
          Prefer `.pyi` stubs over `.py` files when both are present
          
          [default: true]
          [possible values: true, false]

      --public-only
          Only check symbols reachable from public modules via re-export chains

      --output-format <OUTPUT_FORMAT>
          Format for the untyped-symbol findings

          Possible values:
          - min-text:    Minimal text output, one line per error
          - full-text:   Full, verbose text output
          - json:        JSON output
          - github:      Emit GitHub Actions workflow commands
          - junit-xml:   Emit JUnit XML
          - omit-errors: Only show error count, omitting individual errors
```

It's based on `typestats check`, but outputs in the same fancy style as `pyrefly check`.

image

# Test Plan

Tests added, in including tests to verify that `coverage check` and `coverage report` are in sync. 
I also manually verified that it reports what it should for numpy and scipy-stubs for the different flags.

---

To keep this PR from growing too large, I thought I'd leave the documentation for a follow-up. But if you want, I could also put that in here.

Anyway, I'm curious to see what you think. I'm not personally attached to this interface (or anything else here for that matter), so don't hesitate to speak your mind :)


jorenham:
One thing I'm not really sure about it the full `--help` output:

```
Check that type coverage from pyrefly type checking results meets a minimum threshold

Usage: pyrefly coverage check [OPTIONS] [FILES]...

Arguments:
  [FILES]...
          Files to check (glob supported). If no file is specified, switch to
          project-checking mode where the files to check are determined from the closest
          configuration file. When supplied, `project_excludes` in any config files loaded
          for these files to check are ignored, and we use the default excludes unless
          overridden with the `--project-excludes` flag

Options:
  -j, --threads <THREADS>
          Number of threads to use for parallelization. Setting the value to 1 implies
          sequential execution without any parallelism. Setting the value to 0 means to pick
          the number of threads automatically using default heuristics
          
          [env: PYREFLY_THREADS=]
          [default: 0]

      --project-excludes <PROJECT_EXCLUDES>
          Files to exclude when type checking

  -c, --config <FILE>
          Explicitly set the Pyrefly configuration to use when type checking. In
          "single-file checking mode," this config is applied to all files being checked,
          ignoring the config's `project_includes` and `project_excludes` and ignoring any
          config-finding approach that would otherwise be used. When not set, Pyrefly will
          perform an upward-filesystem-walk approach to find the nearest pyrefly.toml or
          pyproject.toml with `tool.pyrefly` section'. If no config is found, Pyrefly exits
          with error. If both a pyrefly.toml and valid pyproject.toml are found,
          pyrefly.toml takes precedence
          
          [env: PYREFLY_CONFIG=]

      --color <COLOR>
          Control whether colored output is used
          
          [env: PYREFLY_COLOR=]
          [default: auto]
          [possible values: auto, always, never]

  -p, --preset <PRESET>
          A named collection of error severities and behavior settings that serves as the
          base configuration. Any explicit settings you specify override the preset

          Possible values:
          - off:     Silences every error kind. Other settings (scalars, behavior flags) are
            left at their defaults. Useful when Pyrefly is running only for IDE features
            like hover and go-to-definition, without diagnostics
          - basic:   Minimal checking for LSP users. Raises clear/obvious type errors but
            disables stricter checks like override validation and unannotated def checking
          - legacy:  A looser, less-strict preset useful for codebases migrating from mypy.
            Pyrefly does not aim to mimic mypy's behavior precisely — this preset just
            disables a few checks that mypy does not have, so migrating users aren't hit
            with new errors for classes of issues mypy never flagged
          - default: The default Pyrefly configuration. Equivalent to having no preset at
            all
          - strict:  Enables additional error codes on top of the default for stricter
            checking
          - all:     Enables every error kind at `Error` severity. Directives like
            `reveal-type` keep their default severity

  -v, --verbose
          Enable verbose logging
          
          [env: PYREFLY_VERBOSE=]

      --disable-project-excludes-heuristics[=<DISABLE_PROJECT_EXCLUDES_HEURISTICS>]
          Configures Pyrefly to replace `project-excludes` fully rather than append whatever
          is in your configuration or passed by CLI to Pyrefly's defaults
          
          [possible values: true, false]

      --search-path <SEARCH_PATH>
          The list of directories where imports are imported from, including type checked
          files

      --disable-search-path-heuristics[=<DISABLE_SEARCH_PATH_HEURISTICS>]
          Disable Pyrefly default heuristics, specifically those around constructing a
          modified search path. Setting this flag will instruct Pyrefly to use the exact
          `search_path` you give it through your config file and CLI args
          
          [possible values: true, false]

      --enable-fallback-search-path[=<ENABLE_FALLBACK_SEARCH_PATH>]
          Enable walk-up fallback search paths, where Pyrefly walks up from each importing
          file's directory to the config root to best-effort resolve imports
          
          [possible values: true, false]

      --python-version <PYTHON_VERSION>
          The Python version any `sys.version` checks should evaluate against

      --python-platform <PYTHON_PLATFORM>
          The platform any `sys.platform` checks should evaluate against

      --site-package-path <SITE_PACKAGE_PATH>
          Directories containing third-party package imports, searched after first checking
          `search_path` and `typeshed`

      --conda-environment <CONDA_ENVIRONMENT>
          Use a specific Conda environment to query Python environment information, even if
          it isn't activated

      --python-interpreter-path <EXE_PATH>
          The path to a Python executable that will be queried for `python-version`
          `python-platform`, or `site-package-path` if any of the values are missing

      --fallback-python-interpreter-name <COMMAND>
          The Python executable name available on your PATH that will be queried for your
          `python-version`, `python-platform`, or `site-package-path` if any of the values
          are missing. We execute `which <COMMAND>` to fill in your
          `python-interpreter-path`, which is useful if you don't know where the Python
          executable will be on a given machine, but want to use one other than the default.
          When this and `python-interpreter-path` are unset, we query for `python3` and
          `python`

      --skip-interpreter-query
          Skip doing any automatic querying for `python-interpreter-path`,
          `fallback-python-interpreter-name`, or `conda-environment`

      --typeshed-path <TYPESHED_PATH>
          Override the bundled typeshed with a custom path

      --replace-imports-with-any <REPLACE_IMPORTS_WITH_ANY>
          Always replace specified imports with typing.Any, suppressing related import
          errors even if the module is found

      --ignore-missing-imports <IGNORE_MISSING_IMPORTS>
          If the specified imported module can't be found, replace it with typing.Any,
          suppressing related import errors

      --ignore-errors-in-generated-code[=<IGNORE_ERRORS_IN_GENERATED_CODE>]
          Whether to ignore type errors in generated code
          
          [possible values: true, false]

      --infer-with-first-use[=<INFER_WITH_FIRST_USE>]
          If this is true, infer type variables not determined by a call or constructor
          based on their first usage. For example, the type of an empty container would be
          determined by the first thing you put into it. If this is false, any unsolved type
          variables at the end of a call or constructor will be replaced with `Any`.
          Defaults to true
          
          [possible values: true, false]

      --use-ignore-files[=<USE_IGNORE_FILES>]
          Whether to respect ignore files (.gitignore, .ignore, .git/exclude)
          
          [possible values: true, false]

      --untyped-def-behavior <UNTYPED_DEF_BEHAVIOR>
          Deprecated: use --check-unannotated-defs and --infer-return-types instead.
          Controls how Pyrefly analyzes function definitions that lack type annotations on
          parameters and return values
          
          [possible values: check-and-infer-return-type, check-and-infer-return-any,
          skip-and-infer-return-any]

      --check-unannotated-defs[=<CHECK_UNANNOTATED_DEFS>]
          Whether to type check the bodies of unannotated function definitions
          
          [possible values: true, false]

      --infer-return-types <INFER_RETURN_TYPES>
          Controls when return types are inferred for functions without return annotations.
          Values: never, annotated, checked (default). Only applies to functions whose
          bodies are checked; unannotated functions are only eligible when
          --check-unannotated-defs is true

          Possible values:
          - never:     Never infer return types; unannotated returns are treated as `Any`
          - annotated: Infer return types only for functions with at least one parameter or
            return annotation
          - checked:   Infer return types for all checked functions, including completely
            unannotated ones

      --permissive-ignores[=<PERMISSIVE_IGNORES>]
          Whether Pyrefly will respect ignore statements for other tools, e.g. `# pyright:
          ignore`. Equivalent to passing the names of all tools to `--enabled-ignores`
          
          [possible values: true, false]

      --enabled-ignores <ENABLED_IGNORES>
          Respect ignore directives from only these tools. Can be passed multiple times or
          as a comma-separated list. Defaults to type,pyrefly. Passing the names of all
          tools is equivalent to `--permissive-ignores`

          Possible values:
          - type:    Enables `# type: ignore`
          - pyrefly: Enables `# pyrefly: ignore` and `# pyrefly: ignore-errors`
          - pyright: Enables `# pyright: ignore`
          - mypy:    Enables `# mypy: ignore-errors`
          - ty:      Enables `# ty: ignore`
          - pyre:    Enables `# pyre: ignore`, `# pyre-ignore`, `# pyre-fixme`, and `#
            pyre-ignore-all-errors`
          - zuban:   Enables `# zuban: ignore`

      --error <ERROR>
          Force this rule to emit an error. Can be passed multiple times or as a
          comma-separated list

      --warn <WARN>
          Force this rule to emit a warning. Can be passed multiple times or as a
          comma-separated list

      --ignore <IGNORE>
          Do not emit diagnostics for this rule. Can be passed multiple times or as a
          comma-separated list

      --info <INFO>
          Force this rule to emit an info-level diagnostic. Can be passed multiple times or
          as a comma-separated list

      --recursion-depth-limit <RECURSION_DEPTH_LIMIT>
          Maximum recursion depth before triggering overflow protection. Set to 0 to disable
          (default). This helps detect potential stack overflow situations

      --recursion-overflow-handler <RECURSION_OVERFLOW_HANDLER>
          How to handle when recursion depth limit is exceeded

          Possible values:
          - break-with-placeholder: Return a placeholder type and emit an internal error.
            Safe for IDE use
          - panic-with-debug-info:  Dump debug info to stderr and panic. For debugging stack
            overflow issues

      --pytorch-efficiency-lints <PYTORCH_EFFICIENCY_LINTS>
          Enable PyTorch efficiency lints that detect common GPU performance anti-patterns
          
          [possible values: true, false]

      --tensor-shapes <TENSOR_SHAPES>
          (Experimental) Enable tensor shape type inference. Supports both native (Tensor[N,
          M]) and jaxtyping (Float[Tensor, "batch channels"]) syntax
          
          [possible values: true, false]

      --strict-callable-subtyping[=<STRICT_CALLABLE_SUBTYPING>]
          Whether to strictly check callable subtyping for signatures with `*args: Any,
          **kwargs: Any`. When false (the default), callables with `*args: Any, **kwargs:
          Any` are treated as compatible with any signature (similar to `...` behavior).
          When true, parameter list compatibility is checked strictly even when `*args: Any,
          **kwargs: Any` is present
          
          [possible values: true, false]

      --spec-compliant-overloads[=<SPEC_COMPLIANT_OVERLOADS>]
          Whether to use spec-compliant overload evaluation semantics. When false (the
          default), Pyrefly attempts to resolve ambiguous calls precisely. When true,
          overload evaluation follows the typing spec exactly, falling back to `Any` more
          frequently
          
          [possible values: true, false]

      --strict
          Count `Any`-resolved annotations as untyped

  -f, --fail-under <PERCENT>
          Minimum coverage percentage; exit non-zero when coverage is below it
          
          [default: 100]

      --prefer-stubs <PREFER_STUBS>
          Prefer `.pyi` stubs over `.py` files when both are present
          
          [default: true]
          [possible values: true, false]

      --public-only
          Only check symbols reachable from public modules via re-export chains

      --output-format <OUTPUT_FORMAT>
          Format for the untyped-symbol findings

          Possible values:
          - min-text:    Minimal text output, one line per error
          - full-text:   Full, verbose text output
          - json:        JSON output
          - github:      Emit GitHub Actions workflow commands
          - junit-xml:   Emit JUnit XML
          - omit-errors: Only show error count, omitting individual errors

  -h, --help
          Print help (see a summary with '-h')
```

Because, as you can see, there are a bunch of inherited flags there that aren't relevant to `coverage check`. It's also like this for `coverage report` thought, so I'm not sure whether to go for consistency here, or if we should remove those irrelevant flags :thinking:

pyrefly/lib/test/coverage/test_files/property_basic.py
```diff
@@ -51,3 +51,13 @@ def a(self) -> int:
...
+
+
+class GetterSetterUntyped:
+    @property
+    def x(self):
+        return 0
+
+    @x.setter
+    def x(self, value):
+        pass
```
jorenham, line 63: Property getter and setters are deduped in a second pass in the individual  `coverage report` and `coverage check` commands. So these use (also) used to ensure that `report` and `check` both dedupe these as they should.

pyrefly/lib/commands/coverage/types.rs
jorenham, line 1: This is almost entirely a direct cut/paste from `coverage/report.rs`

pyrefly/lib/commands/coverage/collect.rs
jorenham, line 1: Almost everything here was directly taken from `coverage/report.rs`.

jorenham:
Awesome; thanks!

PR #3717, opened 2026-06-08
Require `/` separators in glob patterns (#3717)

samwgoldman:
Summary:

Pyrefly config files live in repositories shared across platforms, so project glob patterns should use `/` rather than OS-specific separators, so the same config has the same behavior on all platforms.

Preparation for moving to the `globset` crate, next diff in the stack.

Reviewed By: rchen152

Differential Revision: D107906011


rchen152 (approved):

PR #3718, opened 2026-06-08
Resolve real source locations by reusing the query transaction (#3716)

kinto0:
Summary:
The zero-range typeshed/synthesized declarations emitted earlier render the right names but break goto-definition and typeshed hover docs, which need a real `Node { uri, range }`. Give the converter a third resolver, `ExportLocationResolver`, that maps `(module, name)` to a source path+range (following re-exports), and use it to upgrade special-form/`typing` classes, `Quantified` TypeVars (PEP 695 → typeparam, legacy → variable), and imported/special functions whose `FuncId` lacks a `def_index` (e.g. `typing.overload`) to precise locations.

Resolving a location demands the target module's exports, which demands its `Stdlib`. Rather than spinning up a fresh cold transaction per lookup — which starts with an empty stdlib and would panic in `get_stdlib` unless warmed by an extra `Require::Exports` run — the converter now reuses the single transaction that already produced the type. Computing a type necessarily populated that transaction's stdlib, so the export lookups are warm by construction: no warm-up run, no cold-start guard, and one transaction per query instead of one per resolved symbol. Conversion moves server-side (`type_at_position`/`expected_type_at_position`) where that transaction lives, collapsing the previous five TSP interface methods into two.

Differential Revision: D107943323

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -238,6 +238,7 @@ use pyrefly_util::telemetry::TelemetryServerState;
 use pyrefly_util::thread_pool::ThreadCount;
 use pyrefly_util::thread_pool::ThreadPool;
 use pyrefly_util::watch_pattern::WatchPattern;
+use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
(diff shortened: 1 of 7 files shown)


rchen152 (approved):

PR #3734, opened 2026-06-09
Make TSP getComputedType range-aware for call expressions

rchiodo:
## Summary

Makes the `typeServer/getComputedType` TSP endpoint range-aware so that querying the full range of a **call expression** (`Foo()`) returns the call's *result* type rather than the callee declaration.

## Problem

`getComputedType` resolved the type at the requested range's **start** position. For a call expression, the start offset lands on the callee, so the declaration-preserving lookup returned the callee (e.g. the constructor function) instead of the call's result. TSP clients (e.g. Pylance) that send the whole call-expression range therefore saw the constructor instead of the constructed instance. This broke downstream features such as go-to-definition on an operator whose operand is a call expression (`ClassWithMagicMethod() < 1`), where the client needs the operand to resolve to a class **instance** to find the `__lt__` magic method.

## Change

- Add `Transaction::get_computed_type_at_range`: when the requested range exactly covers a call expression, return its recorded result type; otherwise fall back to the existing declaration-preserving position lookup.
- Add the `range_is_call_expr` helper.
- Thread the full node range (start + end) through the `getComputedType` request handler and the `TspInterface`.

The callee/identifier behavior is unchanged: an empty range, or a range that only spans the callee identifier, still preserves the callee declaration (so the existing pylance-release#8020 behavior is intact). Only a whole-call-expression range resolves to the call result.

## Tests

New tests in `get_type_queries.rs`:
- `test_get_computed_type_call_expr_range_returns_instance` — whole call range returns the constructed instance.
- `test_get_computed_type_call_expr_range_vs_callee_position` — whole call range → instance; callee position → the class itself (declaration preserved).
- `test_get_computed_type_builtin_call_expr_range_returns_instance` — `str()` range → instance.

Related: https://github.com/microsoft/pylance-release/issues/8016

### DIFF
--- a/crates/tsp_types/src/common.rs
+++ b/crates/tsp_types/src/common.rs
@@ -155,6 +155,16 @@ impl GetTypeArg {
             GetTypeArg::Node(n) => n.range.start.clone(),
         }
     }
+
+    /// Extract the end position of the range. Together with `position()` this
+    /// recovers the full requested node range, which `getComputedType` uses to
+    /// distinguish a whole call expression from its callee identifier.
+    pub fn end_position(&self) -> tsp::Position {
+        match self {
+            GetTypeArg::Declaration { node, .. } => node.range.end.clone(),
+            GetTypeArg::Node(n) => n.range.end.clone(),
+        }
+    }
 }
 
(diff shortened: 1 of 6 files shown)


ndmitchell (approved):

PR #3739, opened 2026-06-10
Fix recurser test on stable Rust

Photon101:
## Summary

- Rewrites the recurser unit-test match guard to avoid unstable `if let` guards.
- Keeps the test behavior unchanged while preserving stable Rust compatibility.

## Why

`cargo test -p pyrefly_util` currently fails on stable Rust because the test module uses an experimental pattern guard.

## Checks

- `cargo test -p pyrefly_util test_recurser_demo`
- `cargo test -p pyrefly_util`
- `cargo fmt -- --check`

Disclosure: I’m an AI coding agent operating under the Photon101 account; I reviewed this change and ran the checks above.

### DIFF
--- a/crates/pyrefly_util/src/recurser.rs
+++ b/crates/pyrefly_util/src/recurser.rs
@@ -75,11 +75,12 @@ mod tests {
                     tree(t, r, res);
                 }
             }
-            Tree::Leaf(i, n) if let Some(_guard) = r.recurse(*i) => {
-                res.push(*i);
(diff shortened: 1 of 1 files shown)


Photon101:
Closing this draft after switching the local environment to Pyrefly's project-managed stable toolchain from rust-toolchain.toml. With rustc 1.96.0, the upstream recurser test compiles as-is, so this change is not needed.

PR #3744, opened 2026-06-10
Fix performance hang and OOM on conditional tuple appends

### LINKED ISSUE #3743: Pyrefly hang / OOM during type checking of many conditional tuple appends
### Describe the Bug

During type checking of code containing a sequence of conditional tuple additions, Pyrefly hangs indefinitely (CPU pinned at 100%) and eventually runs out of memory (OOM).

### Reproduction
Checking the following Python code using `pyrefly check` hangs and times out:

```python
class D0: pass
class D1: pass
class D2: pass
class D3: pass
# ... (up to D29)

def repro(conds: list[bool]):
  z = ()
  if conds[0]: z += (D0(),)
  if conds[1]: z += (D1(),)
  # ... (up to D29)
  return z
```

### Root Cause
At each control-flow join, Pyrefly computes the union of the incoming types of `z`. Since each addition can either happen or not, this leads to an exponential combination of concrete tuple types in the union.

For $N$ conditional additions of distinct class types, the simplifier ends up processing a union containing up to $2^N$ concrete tuple variants. As the size of the union grows, the type simplifier stalls attempting to sort, de-duplicate, and check subtyping/intersections on this massive list, resulting in exponential $O(2^N)$ CPU time and memory exhaustion.

### Proposed Fix
A suggested solution is to widen/collapse wide tuple unions during type simplification.
If a union contains more than a threshold (e.g., `8`) of tuple variants, we collapse them into a single unbounded tuple variant:
`tuple[E, ...]`
where `E` is the union of all element types extracted from the constituent tuples.

For example, the union of concrete tuples:
`tuple[D0] | tuple[D0, D1] | tuple[D1, D2] | ...`
will be collapsed to:
`tuple[D0 | D1 | D2 | ..., ...]`

This is a sound over-approximation that bounds the type complexity at control-flow joins to $O(N)$ instead of $O(2^N)$. I have an implementation ready for this fix and will be sending a PR shortly. Let me know what you think of this idea or if you have any feedback on the threshold choice!

### Sandbox Link

_No response_

_No response_

QEDady:
# Summary

This PR fixes a performance hang and Out-Of-Memory (OOM) error that occurs when type checking code with many conditional tuple appends (e.g. at control-flow joins). 

### Problem
At each control-flow join, Pyrefly computes the union of incoming types. For $N$ conditional tuple appends of distinct class types, this leads to an exponential combination of up to $2^N$ concrete tuple variants in the union. Pyrefly's simplifier stalls attempting to sort, deduplicate, and intersect this list, leading to $O(2^N)$ CPU time and OOM.

### Solution
We introduce `collapse_wide_tuple_unions` during type simplification. If a union list contains more than a threshold of `8` tuple variants, they are collapsed/widened into a single unbounded tuple variant:
`tuple[E, ...]`
where `E` is the union of all element types recursively extracted from the constituent tuples (concrete, unbounded, or unpacked). This soundly bounds type complexity at control-flow joins to linear $O(N)$ complexity.

Fixes #3743

# Test Plan

1. **Rust Unit Tests**: Added `test_collapse_wide_tuple_unions_corner_cases` in `simplify.rs` to verify correct widening of concrete, unbounded, nested, unpacked, TypeVarTuple, and invalid fallback tuple combinations.
2. **Integration Regression Test**: Added a 30-branch conditional tuple append integration test to `test/basic.md` using distinct class instances to ensure the widening threshold is properly triggered.
3. **Execution**: Ran the tests
   ```bash
   python3 test.py 
   ```
4. **Performance Verification**: Checked the 30-branch reproduction script on the local build. The run, which previously hung/timed out (exceeding 60 seconds at 100% CPU), now compiles successfully in **0.18 seconds** with flat memory footprint (~360MB).


yangdanny97:
Nice, thanks

yangdanny97:
FYI I'm gonna bump the limit to 64 here

yangdanny97:
or rather, 256... pytorch has an un-annotated list of ~100 2-tuples :P

samwgoldman (approved):

PR #3770, opened 2026-06-11
Single-pass class method indexing in `pyrefly coverage`

jorenham:
# Summary

The `parse_classes` function was re-iterating every binding key in the module for each class, just to find the methods of that class. This changes the code (not the behavior) so that this now only requires a single pass.

Here's a quick little benchmark:

| Corpus | before | after | speedup |
|---|---:|---:|---:|
| synthetic, 1 module, 2,000 classes | 0.457s | 0.274s | **1.67x** |
| synthetic, 1 module, 6,000 classes | 4.182s | 0.803s | **5.2x** |
| scipy-stubs, 589 modules (and 1 maintainer :sweat_smile: ) | 1.766s | 1.783s | ~1.0x |

So it looks a bit like this could be a quadratic speedup (in the #classes/module sense)

# Test Plan

Tests still pass

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -1257,6 +1257,24 @@ fn parse_classes(
         String::new()
     };
     let deleted = bindings.module_deletes();
+
+    // group method definitions by class
+    let mut methods_by_class: HashMap<Idx<KeyClass>, Vec<Idx<KeyUndecoratedFunction>>> =
+        HashMap::new();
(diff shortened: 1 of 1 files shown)


NathanTempest:
Thanks @jorenham for your continued efforts in refactoring and improving `pyrefly coverage` !!

Replacing the per-class O(classes × bindings) rescan with a single grouping pass is the right fix, and the benchmark makes the  improvement clear. Importing and merging.

NathanTempest (approved):

yangdanny97 (approved):

PR #3747, opened 2026-06-10
Dedupe `pyrefly coverage` per-module parse pipeline

jorenham:
# Summary

This cleans up `collect.rs` by deduplicating the per-module symbol collection symbol, which was basically copy-pasted five times. This was bothering me for a while now, and I figured now would be a good time to clean this up now that there are no open `pyrefly coverage` PRs.

The only behavioral change, that did not require any changes to the tests, is that `ModuleSymbols::collect` now also parses suppressions on the `.py` side while merging. Previously this was skipped, which was probably an oversight.

There's still room for improvement in `collect.rs`. But I don't want to make the reviewers hate me more than they already do, so I'll address those in separate PRs :innocent: .

# Test Plan

Tests still pass

PR #3771, opened 2026-06-11
Dedupe symbol count fields in `pyrefly coverage`

jorenham:
# Summary

The `ModuleReport` and `ReportSummary` structs both had the same 8 `n_{symbol}` count fields, which were modified in `collect.rs` one by one in three functions. So if you wanted to add a new symbol counter, you'd have to do so  in 5 places.

This dedupes these fields using he same `#[serde(flatten)]` trick used for the "slot" (typable) counts. 

This is purely a refactor; so there are no behavior changes.

This shouldn't cause any merge conflicts with #3747.

# Test Plan

Tests still pass.

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -369,33 +369,28 @@ fn filter_module_report_to_public(report: &mut ModuleReport, public_fqns: &HashS
         .names
         .retain(|n| is_public_fqn(n, &module_prefix, public_fqns));
 
-    // recompute all aggregates in a single pass
+    // recompute all aggregates in a single pass; type ignores attach to the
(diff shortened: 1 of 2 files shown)


NathanTempest (approved):
Thanks @jorenham for your continued refactoring and improvements in `pyrefly coverage`. Pulling the eight repeated count fields into SymbolCounts with #[serde(flatten)] mirrors the existing SlotCounts pattern perfectly, and I confirmed it's output-identical (2 files changed, zero snapshot diffs). The merge() helper cleanly replaces the manual summation, and the n_type_ignores preservation in filter_module_report_to_public is handled correctly. LGTM, importing and merging.

yangdanny97 (approved):

PR #3774, opened 2026-06-11
Skip optional imports in `pyrefly coverage`

### LINKED ISSUE #3773: `pyrefly coverage` reports optional imports not used in stubs as untyped
In `numpy/f2py/crackfortran.py` ([src](https://github.com/numpy/numpy/blob/856ea449ad2d39fd41fe6c6dec5cc69e0fcab13d/numpy/f2py/crackfortran.py#L152-L155)):

```py
try:
    import charset_normalizer
except ImportError:
    charset_normalizer = None
```

In the corresponding `.pyi` there is no such import, because there's simply no need for it. 

Yet, `pyrefly coverage check` (`pyrefly==1.1.0.dev2`) reports

```
 WARN `numpy.f2py.crackfortran.charset_normalizer` is untyped [coverage-missing]
   --> numpy/f2py/crackfortran.py:155:5
    |
155 |     charset_normalizer = None
    |     ------------------
    |
```

But I don't think that annotating `charset_normalizer` as `None | ModuleType` is needed, and it might even confuse some type-checkers. So I think that `pyrefly coverage` shouldn't report optional imports like these ones. This is a very common pattern, after all.

jorenham:
# Summary

Optional imports like 

```py
try:
    import a
except ImportError:
    a = None
```

are now no longer reported as untyped by `pyrefly coverage {check, report}`.

Fixes #3773

# Test Plan

Regression tests and verified consistent behavior between `check` / `report`

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -466,6 +466,28 @@ fn parse_variables(
         }
     }
 
+    /// True if the binding resolves to an import in at least one flow branch.
+    fn involves_import(bindings: &Bindings, idx: Idx<Key>, seen: &mut SmallSet<Idx<Key>>) -> bool {
+        if !seen.insert(idx) {
+            return false; // LoopPhi can be cyclic
+        }
+        match bindings.get(idx) {
+            Binding::Module(..) | Binding::Import(..) => true,
+            Binding::Forward(i)
+            | Binding::PromoteForward(i)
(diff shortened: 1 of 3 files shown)


NathanTempest (approved):
Thank you for another clean fix @jorenham. Walking the binding graph to skip any flow merge where a branch resolves to an import is the right approach, and routing the skip through parse_variables means coverage check and coverage report stay consistent for free.

LGTM and importing this.

yangdanny97 (approved):

PR #3780, opened 2026-06-11
Treat merged stub symbols as untyped in `pyrefly coverage`

### LINKED ISSUE #3778: `pyrefly coverage` should ignore `.py` annotations if a `.pyi` exists
Static type-checkers ignore the `.py` if a `.pyi` exists (following the typing spec). So if a typable is typed in a `.py` but not in the `.pyi`, it should be considered untyped.

I encountered this in scipy-stubs:

```
 WARN `scipy.spatial.transform._rotation.Rotation.__setstate__` is not fully typed [coverage-partial]
    --> .venv/lib/python3.14/site-packages/scipy/spatial/transform/_rotation.py:2584:5
     |
2584 | /     def __setstate__(self, state: tuple[Array, bool]):
2585 | |         quat, single = state
2586 | |         xp = array_namespace(quat)
2587 | |         self._xp = xp
2588 | |         self._quat = xp.asarray(quat, copy=True)
2589 | |         self._backend = select_backend(xp, cython_compatible=self._quat.ndim < 3)
2590 | |         self._single = single
```

This should've been trated as fully untyped, not partially typed, because in the stubs there is no `__setstate__`.

jorenham:
# Summary

From the typing spec ([src](https://typing.python.org/en/latest/spec/distributing.html#partial-stub-packages)):

> Type checkers should merge the stub package and runtime package directories. This can be thought of as the functional equivalent of copying the stub package into the same directory as the corresponding runtime package and type checking the combined directory structure. **Thus type checkers MUST maintain the normal resolution order of checking `*.pyi` before `*.py` files.**

So this ensures that annotated symbols in a `.py` are treated as untyped if those symbols are not annotated in the `.pyi` by `pyrefly coverage {check, report}`.

Fixes #3778

# Test Plan

There are tests :)

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -1537,7 +1537,8 @@ impl ModuleSymbols {
 
     /// When this `.pyi` stub only covers a subset of its `.py` counterpart's public
     /// symbols, add the uncovered `py` symbols so that completeness metrics reflect
-    /// the full module interface. `transaction`/`handle` must be the stub's.
+    /// the full module interface. Merged symbols count as fully untyped, since type
+    /// checkers ignore the `.py` when a stub exists. `transaction`/`handle` must be the stub's.
     fn merge_uncovered_py_symbols(
         &mut self,
         transaction: &Transaction,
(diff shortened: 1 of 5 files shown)


yangdanny97 (approved):

jorenham:
Thanks!

PR #3784, opened 2026-06-12
Add unused-type-ignore rule

### LINKED ISSUE #3766: new rule - unused `# type: ignore` comment
https://docs.astral.sh/ty/reference/rules/#unused-type-ignore-comment

```py
# should be an error because this ignore comment is unused
a = 20 / 2  # type: ignore
```

not to be confused with [`unused-ignore-comment`](https://docs.astral.sh/ty/reference/rules/#non-callable-init-subclass) which only checks `ty: ignore` comments (pyrefly already has an equivalent rule for this - `unused-ignore`)

the benefit of having two separate rules is projects that use multiple type checkers (ie. libraries) can use  `#type: ignore` comments for mypy without it impacting other type checkers, but if a project only uses one type checker, they would want unused `# type: ignore` comments to be reported by pyrefly

aodihis:
# Summary

Adds a new `unused-type-ignore` rule that reports `# type: ignore` comments that are not suppressing any error.

Fixes #3766

This rule is off by default to avoid breaking projects using multiple type checkers (e.g. mypy + pyrefly). Enable it in `pyrefly.toml`:

```toml
[errors]
unused-type-ignore = "error"  # or "warn"
```

# Test Plan

- Unit tests for unused/used # type: ignore detection
- LSP integration tests verifying the diagnostic appears when enabled and is silent by default
- Manually verified that the diagnostic shows when the rule is set to "error" and is silent when set to "ignore

### DIFF
--- a/pyrefly/lib/state/errors.rs
+++ b/pyrefly/lib/state/errors.rs
@@ -447,7 +447,7 @@ impl Errors {
                         continue;
                     }
                     match tool {
-                        Tool::Pyrefly | Tool::Pyre => {}
+                        Tool::Pyrefly | Tool::Pyre | Tool::Type => {}
                         _ => continue,
                     }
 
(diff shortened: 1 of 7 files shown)


pyrefly/lib/state/errors.rs
```diff
@@ -478,6 +478,26 @@ impl Errors {
                         continue;
                     }
 
+                    // For Tool::Type (`# type: ignore`), codes are mypy codes (not pyrefly
```
yangdanny97: > codes are mypy codes

I'm not sure this is true, I don't think type: ignore as written in the typing spec has official codes
that said, if someone does use it with a code we proabbly can't shouldn't assume we understand it

yangdanny97:
I would have thought we could use simpler tests for this, but I checked and the other unused-ignore is tested the same way. I think this is fine for now.

This can probably be merged after some editing of the comments @NathanTempest

pyrefly/lib/state/errors.rs
aodihis: I see, sure, I will remove these unnecessary comments.

NathanTempest (approved):
Thanks for the well-scoped contribution, @aodihis  and welcome! Making this a distinct, off-by-default rule is exactly right: projects on mypy+pyrefly can keep their # type: ignore comments without pyrefly flagging them, while pyrefly-only projects can opt in. 
 @yangdanny97's review looks addressed. LGTM for importing.

aodihis:
@NathanTempest  Thank you! I appreciate the review and the feedback,

yangdanny97 (approved):

PR #3800, opened 2026-06-13
fix Support pydantic frozen field #3663

### LINKED ISSUE #3663: Support pydantic frozen field
### Describe the Bug

This is very similar to [#316](https://github.com/facebook/pyrefly/issues/316) but instead of using ConfigDict, only a certain field is declared frozen. Example:

```python


class Foo(BaseModel):
    id: int | None = Field(frozen=True)


class Bar(Foo):
    id: int = Field(frozen=True)
```

The field is frozen, so the attribute is immutable and therefore covariant. However, we get an error
Class member `Bar.id` overrides parent class `Foo` in an inconsistent manner

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3663

Pydantic Field(frozen=True) is now carried through DataclassFieldKeywords and treated as a readonly field reason, so covariant overrides like int | None to int no longer produce a bad override error.

# Test Plan

add test


yangdanny97:
Is the primer delta expected?

asukaminato0721:
> Is the primer delta expected?

The type annotation is identical (str → str), so there is no type inconsistency.

While adding frozen=True in a subclass technically violates the Liskov Substitution Principle (code expecting a mutable Serializer.type field would fail when given a subclass instance), this is a very common and intentional Pydantic pattern for discriminator fields — the subclass locks down the type tag to a specific value.

So I think this is expected? Or add specific code for this?

federicocarne:
We're relying on pyrefly across a mixed pydantic codebase and are hitting a fair number of `bad-override-mutable-attribute` false positives from exactly the pattern this PR fixes (discriminated-union subclasses covariantly narrowing a `Field(frozen=True)` field).

Any chance this lands before 1.2.0 ships?

grievejia (approved):

rchen152:
@federicocarne Thanks for the ping! This has been merged and will be in 1.2.0.

PR #3801, opened 2026-06-13
fix report baselined errors as a hint in the language server #3441

### LINKED ISSUE #3441: report baselined errors as a hint in the language server
having baselined errors still reported as such in the IDE is annoying because it obscures which errors were actually introduced by my new code vs existing code. this triggers [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue) where developers just end up ignoring all errors in their IDE and only relying on the CI.

basedpyright's baseline functionality reports all baselined diagnostics as a [hint](https://docs.basedpyright.com/latest/benefits-over-pyright/language-server-improvements/#improved-diagnostic-severity-system) instead, which makes the errors still visible but far less noisy

asukaminato0721:
# Summary

Fixes #3441

split LSP diagnostics into normal vs baseline-matched diagnostics using the existing baseline processor, downgrade baseline-matched push diagnostics to HINT, same behavior for pull diagnostics.

# Test Plan

add test and related test files.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -3000,11 +3000,17 @@ impl Server {
                 diags.insert(handle_path_buf, Vec::new());
             }
         }
-        for e in transaction
+        let (normal_errors, baseline_errors) = transaction
             .get_errors(handles)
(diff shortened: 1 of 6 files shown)


kinto0 (approved):

PR #3803, opened 2026-06-13
Fix LSP Var leak panic from lambda params in module cycles (#3789) (#3803)

### LINKED ISSUE #3789: panic: a variable has leaked from one module to another (solver.rs:451) on LSP didOpen of sympy/core/expr.py
# Bug report: panic "a variable has leaked from one module to another" (solver.rs:451) on LSP open

**Report at:** https://github.com/facebook/pyrefly/issues/new

## Summary

pyrefly 1.0.0 panics (Rust panic, exit code 101) when the **LSP server** opens
`sympy/core/expr.py` from a real sympy checkout. The panic is:

```
ERROR Thread panicked, shutting down: panicked at pyrefly\lib\solver\solver.rs:451:24:
Internal error: a variable has leaked from one module to another.
PANIC Sorry, Pyrefly crashed, this is always a bug in Pyrefly itself.
```

It only happens on the **incremental open-file / LSP (`textDocument/didOpen`)** path.
Batch `pyrefly check` of the single file does **not** crash; whole-project
`pyrefly check` surfaces a different, *recovered* internal error.

## Environment

- pyrefly **1.0.0** (latest on PyPI)
- OS: Windows
- Python: 3.11 (interpreter configured via `--python-interpreter-path`)
- Project: sympy at commit `6fcc30bbfa6fa4cbde2124640f1f84385292c6a6`

## Repro (simplest)

1. `pip install pyrefly==1.0.0`
2. Clone sympy (`git clone https://github.com/sympy/sympy`), checkout `6fcc30bb`.
3. Start the LSP server: `pyrefly lsp`
4. Drive a standard LSP `initialize` (rootUri = sympy checkout) + `initialized`,
   then `textDocument/didOpen` for `sympy/core/expr.py` with its real contents.
5. pyrefly panics within a few seconds (exit 101).

## Minimal trigger (13 lines)

Replacing the entire contents of `sympy/core/expr.py` with the following 13 lines
(while the rest of the sympy checkout stays on disk) still reproduces the panic.
This isolates the trigger to the recursive `aseries` / `as_coeff_exponent` interplay,
the self-referential `-> tuple[Expr, Expr]` return annotation, and the trailing
`from .add import Add` (which keeps module identity `sympy.core.expr` and pulls in the
real `Add`):

```python
class Expr(Basic, EvalfMixin):
    def aseries(self, x=None, n=6, bound=0, hir=False):
        s = func.series(k, 0, n)
        terms = sorted(Add.make_args(s.removeO()), key=lambda i: int(i.as_coeff_exponent(k)[1]))
        for t in terms:
            coeff, expo = t.as_coeff_exponent(k)
            if coeff.has(x):
                snew = coeff.aseries(x, n, bound=bound-1)
                s += (snew * k**expo)
        return (s + o).subs(k, exp(logw))
    def as_coeff_exponent(self, x) -> tuple[Expr, Expr]:
        return s, S.Zero
from .add import Add
```

(`Basic`, `EvalfMixin`, `func`, `k`, `o`, `exp`, `logw`, `S` are intentionally
unresolved — pyrefly tolerates them. The crash still requires the **real** sympy
`Add`/`Basic` definitions reachable on disk: substituting stub `Add`/`Basic` classes
in a tiny standalone 2-file package does **not** reproduce, so the panic depends on
sympy's actual type graph spanning `sympy.core.expr` and `sympy.core.add`.)

## Notes

- Reproduces only via the LSP / incremental open path, not single-file `pyrefly check`.
- The "variable leaked from one module to another" message and the dependence on the
  real cross-module `Add` strongly suggest a solver type-variable being shared/leaked
  across the `expr` and `add` module boundaries.

samwgoldman:
Summary:

Opening a file in the IDE could panic with "a variable has leaked from
one module to another" (solver.rs) when the file is part of a module
import cycle and contains a lambda.

Why: the lambda parameter's Unwrap `Var` is memoized in the thread-local
`lambda_param_vars` cache, which is shared across modules while a
cross-module SCC is driven iteratively. In the IDE the in-memory (open)
and on-disk versions of a module coexist as distinct module instances
with distinct `Solver`s, but the cache was keyed by `ModuleName` alone, so
they collided: a `Var` allocated in one version's solver was handed back
while solving the other, then panicked when looked up in the wrong solver.
This is why it only reproduced on the incremental LSP open path, not a
uniform batch `check`.

What: key `lambda_param_vars` by `(ModuleName, ModulePath, LambdaParamId)`,
matching the way cross-module lookups already identify a module instance
(see `get_from_module`). The two versions now get separate cache entries,
so a cached `Var` is always valid in the current solver.

Fixes https://github.com/facebook/pyrefly/issues/3789

Differential Revision: D108494662

### DIFF
--- a/pyrefly/lib/alt/answers_solver.rs
+++ b/pyrefly/lib/alt/answers_solver.rs
@@ -1588,7 +1588,10 @@ pub struct ThreadState {
     partial_answers: RefCell<FxHashMap<(Idx<Key>, usize), Arc<TypeInfo>>>,
     /// Solve-time mapping from per-module lambda parameter IDs to the
     /// thread-local Var that represents that parameter in the current solve.
-    lambda_param_vars: RefCell<FxHashMap<(ModuleName, LambdaParamId), Var>>,
+    ///
+    /// The `ModulePath` is needed to distinguish the in-memory and on-disk
+    /// versions of the same module, which can coexist in the IDE (issue #3789).
+    lambda_param_vars: RefCell<FxHashMap<(ModuleName, ModulePath, LambdaParamId), Var>>,
     /// Active trace side-effect sink for the current calculation.
     /// Set before `K::solve`, taken after. `None` when tracing is disabled
     /// or between calculations. Saved sinks form a stack to handle recursive
@@ -1870,14 +1873,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         self.thread_state
             .lambda_param_vars
             .borrow_mut()
-            .insert((self.module().name(), id), var);
+            .insert((self.module().name(), self.module().path().dupe(), id), var);
     }
 
     fn get_lambda_param_var(&self, id: LambdaParamId) -> Option<Var> {
         self.thread_state
             .lambda_param_vars
             .borrow()
-            .get(&(self.module().name(), id))
+            .get(&(self.module().name(), self.module().path().dupe(), id))
             .copied()
     }
 
--- a/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
@@ -438,6 +438,41 @@ fn test_cycle_class() {
     interaction.shutdown().unwrap();
 }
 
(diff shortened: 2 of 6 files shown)


yangdanny97 (approved):

PR #3804, opened 2026-06-13
Ban subclassing dataclasses with order=True

### LINKED ISSUE #3762: new rule - ban subtyping dataclass with `order=True`
https://docs.astral.sh/ty/reference/rules/#subclass-of-dataclass-with-order

```py

@dataclass(order=True)
class Parent:
    value: int

class Child(Parent):  # Ty emits a warning here
    pass

# At runtime, this raises TypeError:
Child(1) < Parent(2)
```

ak4-sh:
# Summary

Adds an `invalid-inheritance` error when a class subclasses a dataclass that has `order=True`. 

Fixes #3762

# Test Plan

Added test `test_parent_dataclass_with_order` to confirm behaviour

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -310,6 +310,8 @@ pub enum ErrorKind {
     NotRequiredKeyAccess,
     /// Unpacking an open TypedDict that may contain a bad key via inheritance.
     OpenUnpacking,
+    /// A class inherits from a stdlib dataclass with `order=True`.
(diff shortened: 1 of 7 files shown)


ak4-sh:
@yangdanny97 Made the changes we discussed! Please let me know if all looks okay

ak4-sh:
I think linter was failing to due to me not running cargo fmt after some previous changes, I've run the fmt and pushed changes. Should be good now hopefully. Sorry about this!

stroxler (approved):

PR #3809, opened 2026-06-14
Improve implicit any error message to show full type parameter list

### LINKED ISSUE #3711: improve the the type parameter missing error
### Describe the Bug

now the error message is

Cannot determine the type parameter `_KT` for generic class `dict`
  Either specify the type argument explicitly, or specify a default for the type variable.
Cannot determine the type parameter `_VT` for generic class `dict`
  Either specify the type argument explicitly, or specify a default for the type variable.

but user can't know the "shape"

a more guide message would be something like

```
dict[_KT, _VT]
     ^^^  ^^^-- _VT is missing
       |--- _KT is missing
```

```
def f(x: dict):
    ...
```

### Sandbox Link

_No response_

_No response_

lesbass:
## Summary

Closes #3711

When a generic class or type alias is used without explicit type arguments, the error message currently only shows the bare name (e.g. `dict`) but doesn't tell the user what type parameters are expected. This change includes the full type parameter list in the error message.

### Before
```
Cannot determine the type parameter `_KT` for generic class `dict`
```

### After
```
Cannot determine the type parameter `_KT` for generic class `dict[_KT, _VT]`
```

## Changes

- `pyrefly/lib/alt/class/targs.rs`: Modified `promote` and `promote_forall` to format the generic entity name with type parameters (e.g. `class \`dict[_KT, _VT]\``)
- `pyrefly/lib/test/generic_basic.rs`: Updated test expectations
- `pyrefly/lib/test/generic_legacy.rs`: Updated test expectations
- `pyrefly/lib/test/recursive_alias.rs`: Updated test expectations

## Verification

The implementation follows the suggestion from @yangdanny97 in the issue discussion. Test expectations in the affected test modules have been updated to match the new error format. Local build verification was not possible due to missing C toolchain in the environment; CI should validate.

### DIFF
--- a/pyrefly/lib/alt/class/targs.rs
+++ b/pyrefly/lib/alt/class/targs.rs
@@ -193,13 +193,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     /// promote(list) == list[Any]
     /// instantiate(list) == list[T]
     pub fn promote(&self, cls: &Class, range: TextRange, errors: &ErrorCollector) -> Type {
+        let tparams = self.get_class_tparams(cls);
+        let generic_entity = if tparams.is_empty() {
+            format!("class `{}`", cls.name())
+        } else {
+            let names = tparams
+                .iter()
+                .map(|t| t.name().as_str())
+                .collect::<Vec<_>>()
+                .join(", ");
+            format!("class `{}[{}]`", cls.name(), names)
+        };
         let targs = self.create_default_targs(
-            self.get_class_tparams(cls),
+            tparams,
             Some(&|tparam: &Quantified| {
                 Self::add_implicit_any_error(
                     errors,
                     range,
-                    format!("class `{}`", cls.name()),
+                    generic_entity.clone(),
                     Some(tparam.name().as_str()),
                 );
             }),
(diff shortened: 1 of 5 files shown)


pyrefly/lib/alt/class/targs.rs
```diff
@@ -193,13 +193,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     /// promote(list) == list[Any]
     /// instantiate(list) == list[T]
     pub fn promote(&self, cls: &Class, range: TextRange, errors: &ErrorCollector) -> Type {
+        let tparams = self.get_class_tparams(cls);
+        let generic_entity = if tparams.is_empty() {
+            format!("class `{}`", cls.name())
+        } else {
+            let names = tparams
+                .iter()
+                .map(|t| t.name().as_str())
```
yangdanny97: I think we should render TypeVarTuple with `*` prefix and ParamSpec with `**` prefix, so like `Class[*Ts]` or `Class[**P]`

yangdanny97: There may be a pre-existing method on TParams to do this. I know there's one that shows the bounds but that seems overkill here, maybe if no such thing exists we could add one as a helper and reuse it at the other call site

yangdanny97:
@NathanTempest I had a few comments on this one but they seem easy, so you could probably patch it after import

yangdanny97:
@lesbass seems like we're blocked from merging because of the CLA - would you mind signing it? or if you object to the CLA, just let us know

lesbass:
Hi @yangdanny97 — the Meta CLA has been signed. The bot notes it may take up to 1 hour for the status to update. Could you re-check when convenient? Thanks!

NathanTempest:
Thanks @lesbass, welcome! This is a really nice readability win where dict[_KT, _VT] actually tells you what to fill in, instead of just pointing at where it's missing. And the new display_name_with_prefix() helper is exactly what @yangdanny97 was pointing at: it mirrors the existing display_with_bounds() pattern in quantified.rs and is reusable anywhere else we want to render type params with */** prefixes.

Tests look good, mypy_primer is clean. LGTM, importing.

Tiny follow-up: the generic_entity block is basically the same in promote() and promote_forall(), just with class vs type alias swapped. I'll pull it out into a small helper myself when I import, no need for you to push another revision.

NathanTempest (approved):

yangdanny97 (approved):

PR #3816, opened 2026-06-15
Dedupe module prefix formatting in `pyrefly coverage`

jorenham:
# Summary

This extracts a ~5x~ 6x duplicated bit of helper code related to module FQN formatting into a function to reduce code duplication.

I ensured that this wouldn't lead to merge conflicts with my three other currently open PRs.

# Test Plan

Refactor only; tests still pass

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -166,6 +166,15 @@ fn class_qualified_name(
     }
 }
 
+/// The module name with a trailing `.`, or empty for the unknown module; prefixed to symbol FQNs.
+fn module_prefix(module: &Module) -> String {
+    if module.name() != ModuleName::unknown() {
+        format!("{}.", module.name())
+    } else {
+        String::new()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


NathanTempest (changes requested):
Thanks @jorenham for continuing the refactors in `pyrefly coverage`. The diff looks good to merge but main has gained a 6th copy of this exact block since you branched in parse_classes (the -> Vec<ReportClass> function):
```
let module_prefix = if module.name() != ModuleName::unknown() {
    format!("{}.", module.name())
} else {
    String::new()
};
```
Your five hunks cover parse_variables, parse_instance_attrs, parse_functions, collect_class_members, and collect_reexport_fqns, but not this one. Could you convert it to let module_prefix = module_prefix(module); too, so no inline copy lingers after the helper lands?

jorenham:
Ohh good catch! 

I rebased onto main and replaced 2 (!) new dupes.

NathanTempest (approved):
Thanks for addressing the follow-up so quickly!

yangdanny97 (approved):

PR #3825, opened 2026-06-15
Parallel per-module collection in `pyrefly coverage`

jorenham:
# Summary

This adds multi-threading support to `collect_module_reports`, so that each `ModuleReport` will now be built in parallel.
Output is unchanged (results are still sorted afterwards).

| Corpus | before | after | speedup |
|---|---:|---:|---:|
| scipy-stubs, 589 modules | 0.60s | 0.50s | ~1.2x |
| synthetic, 4,000 modules | 0.59s | 0.51s | ~1.15x |
| synthetic, 2,000 modules x 20 classes | 1.65s | 1.24s | ~1.33x |
| CPython 3.10 stdlib, 525 modules | 0.47s | 0.42s | ~1.1x |

# Tests

Tests still pass, byte-identical outputs

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -29,6 +29,7 @@ use pyrefly_types::types::Type;
 use pyrefly_util::forgetter::Forgetter;
 use pyrefly_util::includes::Includes;
 use pyrefly_util::thread_pool::ThreadCount;
+use rayon::prelude::*;
 use ruff_python_ast::Expr;
 use ruff_python_ast::Parameters;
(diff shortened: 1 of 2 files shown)


NathanTempest (approved):
Very clean parallelization.Every value shared into par_iter is genuinely Sync so Rust rules out data races at compile time. Output stays deterministic since par_iter().collect() preserves order and you sort afterward. Nice touch running it via the state's install (increased-stack pool) rather than the global rayon pool which avoids stack-overflow risk in deep solving. Thank you for the speedup!

yangdanny97 (approved):

PR #3826, opened 2026-06-15
Drop redundant 2nd flatten pass in union construction

jorenham:
# Summary

`flatten_and_dedup` flattened the member list twice. The second pass only drops the `Never`s that `simplify_intersections` adds, which can be avoid with a simple (and effectively  equivalent) `retain`.

| corpus | before | after | improvement |
|---|---:|---:|---:|
| `scipy-stubs`, (589 modules) | 1786.8ms | 1762.6ms | ~1.4% |
| `matplotlib` (inline typed) | 1199.2ms | 1190.7ms | ~0.7% |

(single-threaded, mean of 12 interleaved runs)

The speedup is modest, but it seems consistent.

# Test Plan

Tests still pass (no behavior changes).


ndmitchell (approved):

PR #3829, opened 2026-06-15
Resolve real source locations by reusing the query transaction (#3829)

kinto0:
Summary:

With the converter able to resolve export locations ([1/2]), the server must supply a resolver that actually finds them. Resolving a location demands the target module's exports, which demands its `Stdlib`. Spinning up a fresh transaction per lookup starts cold and would panic in `get_stdlib` unless warmed by an extra `Require::Exports` run.

Instead the server now reuses the single transaction that already produced the type. Computing a type necessarily populated that transaction's `Stdlib`, so the export lookups are warm by construction — no warm-up run, no cold-start guard, and one transaction per query instead of one per resolved symbol. The target handle is reached via `import_handle` so it inherits the source file's `SysInfo`; re-deriving it from the module path would pick a config-default `SysInfo` whose `Stdlib` was never computed and panic once the transaction holds more than one `SysInfo`.

To put the resolver where that transaction lives, type conversion moves server-side: `type_at_position`/`computed_type_at_range`/`expected_type_at_position` now return the TSP wire type directly, collapsing the previous five `TspInterface` methods into two and removing the per-call `convert_type` on the connection. `lookup_export_location` becomes `pub(crate)` so the converter seam can reach it.

Differential Revision: D108667579

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -238,6 +238,7 @@ use pyrefly_util::telemetry::TelemetryServerState;
 use pyrefly_util::thread_pool::ThreadCount;
 use pyrefly_util::thread_pool::ThreadPool;
 use pyrefly_util::watch_pattern::WatchPattern;
+use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
 use ruff_text_size::TextSize;
@@ -343,6 +344,7 @@ use crate::state::state::Transaction;
 use crate::state::subscriber::CompositeSubscriber;
 use crate::state::subscriber::PublishDiagnosticsSubscriber;
 use crate::state::subscriber::Subscriber;
+use crate::tsp::type_conversion::convert_type_with_resolvers;
 use crate::types::class::ClassDefIndex;
 use crate::types::class::ClassType;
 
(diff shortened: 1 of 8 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -6246,6 +6227,97 @@ impl Server {
...
+    fn convert_type_in_transaction(
+        &self,
+        transaction: &Transaction,
+        source_handle: &Handle,
+        ty: &pyrefly_types::types::Type,
+    ) -> tsp_types::Type {
+        // A `Def` function's name range, looked up in its module's binding table.
+        let resolve_func_range = |func_id: &pyrefly_types::callable::FuncId| {
+            let def_index = func_id.def_index?;
+            let handle = handle_from_module_path(&self.state, func_id.module.path().dupe());
```
rchiodo: `resolve_func_range` is the one resolver here that doesn't inherit the source file's `SysInfo`: it builds the handle via `handle_from_module_path`, which derives the **config-default** `SysInfo` (typically linux) rather than `source_handle`'s. The other two resolvers in this function (`resolve_module_path`, `resolve_export`) reach their targets through `import_handle`, so they inherit the warm `SysInfo` by construction.

Because `Handle`'s identity includes its `SysInfo`, in a multi-`SysInfo` transaction where the function's module was only analyzed under the source platform (e.g. windows), the config-default (linux) handle is a different map key, so `transaction.get_bindings(&handle)` returns `None`.

This is **not** a panic — unlike the export path, `get_bindings` is a pure read (`with_module_inner` returns `None` on a miss and never reaches `get_stdlib`). It's a silent miss: `resolve_func_range` returns `None` and conversion falls through to the export-location resolver. For a top-level exported function that still recovers a range, but **class methods and nested functions aren't module exports**, so they fall all the way through to a zero range and lose their precise declaration location.

Suggest reaching the function's module with the source file's `SysInfo`, mirroring the existing pattern in `Transaction::get_class_fields` (which builds `Handle::new(class.module_name(), class.module_path().dupe(), source_handle.sys_info().dupe())`). `func_id.module` is a `Module`, so both `.name()` and `.path()` are available — no `ModuleName` reconstruction needed, and it avoids the `config_finder` lookup entirely:

```suggestion
            let handle = Handle::new(
                func_id.module.name(),
                func_id.module.path().dupe(),
                source_handle.sys_info().dupe(),
            );
```

This makes all three resolvers consistently warm-by-construction and fixes the method/nested-function zero-range in multi-platform sessions.

samwgoldman (approved):

PR #3830, opened 2026-06-15
fix pytest goto definition support conftest #3775

### LINKED ISSUE #3775: pytest goto definition support conftest
### Describe the Bug

now pytest fixture is supported, but the definition in conftest not yet.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3775

Pytest fixture go-to-definition now falls back from same-file fixture lookup to visible conftest.py / conftest.pyi files, walking upward from the test file directory and stopping at the first matching conftest fixture.

The shared conftest discovery is now reused by the fixture annotation quick fix too.

# Test Plan

### DIFF
--- a/pyrefly/lib/state/lsp/pytest.rs
+++ b/pyrefly/lib/state/lsp/pytest.rs
@@ -6,6 +6,10 @@
  */
 
 use pyrefly_build::handle::Handle;
+use pyrefly_python::module_name::ModuleName;
+use pyrefly_python::module_name::ModuleNameWithKind;
+use pyrefly_python::module_path::ModulePath;
+use pyrefly_python::module_path::ModulePathDetails;
 use pyrefly_python::symbol_kind::SymbolKind;
 use ruff_python_ast::AnyNodeRef;
 use ruff_python_ast::Identifier;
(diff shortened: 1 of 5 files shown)


pyrefly/lib/state/lsp/pytest.rs
```diff
@@ -16,14 +20,61 @@ use vec1::Vec1;
...
+            conftest_paths.push(ModulePath::memory(conftest_pyi.clone()));
+            conftest_paths.push(ModulePath::memory(conftest_py.clone()));
+        } else {
+            if conftest_pyi.exists() {
+                conftest_paths.push(ModulePath::filesystem(conftest_pyi));
+            }
+            if conftest_py.exists() {
+                conftest_paths.push(ModulePath::filesystem(conftest_py));
+            }
+        }
```
samwgoldman: If I understand correctly, this doesn't seem right.

`is_memory` is true if the input handle is for an open file. But this logic assumes that the conftest file is open if the test file is open, which isn't necessarily true.

pyrefly/lib/state/lsp/pytest.rs
```diff
@@ -49,6 +103,38 @@ impl<'a> Transaction<'a> {
                 display_name: Some(fixture.name.as_str().to_owned()),
             })
             .collect();
+        if definitions.is_empty() {
+            for conftest_handle in pytest_conftest_handles(self, handle) {
+                let Some(conftest_ast) = self.get_ast(&conftest_handle) else {
```
samwgoldman: If the conftest file is not open, then the AST will not be available. In a real LSP, the conftest file will be checked with `Require::Indexing` which does not retain the AST. If indexing is disabled, the file might not be checked at all, or only checked up to `Step::Exports`.

pyrefly/lib/test/lsp/definition.rs
```diff
@@ -190,6 +192,37 @@ Definition Result:
...
+"#;
+    let code = r#"
+def test_thing(answer):
+#              ^
+    assert answer == 42
+"#;
+    let (handles, state) = mk_multi_file_state_assert_no_errors(
+        &[("main", code), ("conftest", conftest)],
+        Require::Exports,
+    );
```
samwgoldman: This test is misleading because `mk_multi_file_state_assert_no_errors` will analyze both `main` and `conftest` with `Require::Everything` which retains the AST for `conftest`. In a real LSP, `conftest` will not be checked with `Require::Everything`.

pyrefly/lib/state/lsp/pytest.rs
```diff
@@ -16,14 +20,61 @@ use vec1::Vec1;
...
+        return Vec::new();
+    };
+    let root = module_path
+        .root_of(handle.module())
+        .unwrap_or_else(|| dir.to_path_buf());
+    let is_memory = matches!(module_path.details(), ModulePathDetails::Memory(_));
+    let mut conftest_paths = Vec::new();
+    loop {
+        let conftest_pyi = dir.join("conftest.pyi");
+        let conftest_py = dir.join("conftest.py");
```
samwgoldman: Can you explain why we should prefer `conftest.pyi`? Usually go-to-def will prefer `.py` instead. We have a find preference enum to control this behavior, which might be helpful.

samwgoldman (changes requested):

PR #3832, opened 2026-06-16
feat(ignore): add invalid-ignore-comment and ignore-without-code rules

### LINKED ISSUE #3450: option to report an error on `# pyrefly: ignore` comments without an error code
basically a pyrefly equivalent of basedpyright's [`reportIgnoreCommentWithoutRule`](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportignorecommentwithoutrule)

### LINKED ISSUE #3752: new rule - invalid ignore comment
ty's [`invalid-ignore-comment`](https://docs.astral.sh/ty/reference/rules/#invalid-ignore-comment) rule detects `# type: ignore` and `# ty: ignore` comments that are a mistake.

```py
# pyrefly: ignoree[asdfasdf] # should be an error
# pyrefly: ignoree  # should be an error
```

mikeleppane:
Closes #3752
Closes #3450

Two new ignore-comment lint rules, mirroring ty's
[`invalid-ignore-comment`](https://docs.astral.sh/ty/reference/rules/#invalid-ignore-comment)
and basedpyright's
[`reportIgnoreCommentWithoutRule`](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportignorecommentwithoutrule).

### What & why

**`invalid-ignore-comment`** (default `warn`). A comment that *looks* like an
ignore directive but is malformed used to be **silently discarded**: it produced
neither a suppression nor any warning, so you never learned it was wrong. Three
shapes are now flagged:

```python
x: int = "oops"  # pyrefly: ignoree              # typo'd keyword
x: int = "oops"  # pyrefly: ignore[bad-assignment # unterminated bracket
x: int = "oops"  # pyrefly: ignore[]             # empty code list
```

**`ignore-without-code`** (default `ignore`, opt-in). A bare `# pyrefly: ignore`
suppresses *any* error on its line, so it can silently swallow a later, unrelated
error. Enable the rule to require an explicit code:

```python
# flagged when enabled:
x: int = "oops"  # pyrefly: ignore
# fix:
x: int = "oops"  # pyrefly: ignore[bad-assignment]
```

### Key design property: purely additive

`invalid-ignore-comment` changes **no** suppression behavior. The ignore-comment
parser now returns a single recognizer result with three disjoint outcomes
(`Valid`, `Malformed { suppression, .. }`, `NotADirective`), so exactly one fires
per comment, and a non-directive such as `# type: list[int]` (which never reaches
the `ignore` stem) can **never** be flagged. A `Malformed` result still carries
the exact suppression it produced before, so the diagnostic is layered on top:

- typo: suppresses nothing, as before
- unterminated `[`: keeps the lenient parse, still suppresses
- empty `[]`: still suppresses for tools whose codes aren't checked
  (`# type: ignore[]` remains a blanket suppression), a no-op for Pyrefly

### How

- **`crates/pyrefly_python/src/ignore.rs`**: `parse_ignore_comment` returns the
  new `ParsedIgnore` enum; `Ignore` records a `Vec<MalformedIgnore>` alongside
  its suppression map. `ignore-errors` / `ignore-all-errors` are excluded via a
  word-boundary check so genuine file-level directives aren't flagged.
- **`crates/pyrefly_config/src/error_kind.rs`**: two new `ErrorKind`s in
  lexicographic position (`InvalidIgnoreComment` = `warn`, `IgnoreWithoutCode` =
  `ignore`) plus `ErrorKind::is_unsuppressable()`, the shared predicate for the
  three ignore-comment diagnostics.
- **`pyrefly/lib/state/errors.rs`**: `collect_ignore_comment_lint_errors()`
  emits both diagnostics (gated on the tool being enabled), folded into the
  ignore-diagnostics display path so configured severity is respected.
- **Unsuppressable enforcement**: centralized in `error/collector.rs` so the
  ignore-all, per-line, and f-string paths are covered uniformly; also excluded
  from `--suppress-errors` so it can't write a comment that could never take
  effect.
- **Docs**: `website/docs/error-kinds.mdx` gains both rule sections.

Presets are unchanged: `all` promotes both automatically; `strict` is left as-is
to keep `ignore-without-code` opt-in.

### Test plan

| Test | Protects |
|------|----------|
| `ignore::tests::test_parse_ignore_comment` | recognizer: all three malformed shapes plus valid/non-directive negatives |
| `ignore::tests::test_malformed_ignores_are_additive` | malformed comments still suppress exactly as before |
| `state::errors::tests::test_invalid_ignore_comment_{emitted,variants}` | diagnostic emitted for typo / unterminated / empty; real error still fires |
| `state::errors::tests::test_empty_bracket_ignore_is_additive` | `# type: ignore[]` still suppresses (no regression for non-Pyrefly tools) |
| `state::errors::tests::test_ignore_without_code_{emitted,scoped_to_pyrefly}` | bare ignore flagged; `# type: ignore` not (Pyrefly-scoped) |
| `test/ignores.md` (scrut) | end-to-end: warn-on-by-default and additive, opt-in behavior, never-flag cases |

- Full lib suite: **5641 pass / 0 fail**; `pyrefly_config` 219; `clippy` and `fmt` clean.
- `mypy_primer` over 56 real-world projects (mypy, sphinx, scrapy, pydantic,
  cryptography, and more): **zero diagnostic diff**, confirming no existing
  diagnostic changes on real code.

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -156,6 +156,11 @@ pub enum ErrorKind {
     DivisionByZero,
     /// Explicit usage of `typing.Any` in an annotation.
     ExplicitAny,
+    /// A `# pyrefly: ignore` comment suppresses errors without naming a specific
+    /// error code. Off by default; enable it to require every Pyrefly
+    /// suppression to list the code(s) it silences, e.g.
+    /// `# pyrefly: ignore[bad-return]`.
+    IgnoreWithoutCode,
     /// Raised when a class implicitly becomes abstract by defining abstract members without
     /// inheriting from `abc.ABC` or using `abc.ABCMeta`.
     ImplicitAbstractClass,
@@ -214,6 +219,11 @@ pub enum ErrorKind {
     /// is reported under `BadClassDefinition` or `BadFunctionDefinition`
     /// instead, both of which default to `error`.
     InvalidDecorator,
+    /// A comment looks like a type checker ignore directive but is malformed, so
+    /// it silently suppresses nothing (or not what was intended). For example
+    /// `# pyrefly: ignoree` (typo), `# pyrefly: ignore[bad-code` (unterminated
+    /// bracket), or `# pyrefly: ignore[]` (empty code list).
+    InvalidIgnoreComment,
     /// An error caused by incorrect inheritance in a class or type definition.
     /// e.g. a metaclass that is not a subclass of `type`.
     InvalidInheritance,
@@ -464,6 +474,7 @@ impl ErrorKind {
             ErrorKind::Deprecated => Severity::Warn,
             ErrorKind::DivisionByZero => Severity::Warn,
             ErrorKind::ExplicitAny => Severity::Ignore,
+            ErrorKind::IgnoreWithoutCode => Severity::Ignore,
             ErrorKind::ImplicitAbstractClass => Severity::Ignore,
             ErrorKind::ImplicitAny => Severity::Ignore,
             ErrorKind::ImplicitAnyAttribute => Severity::Ignore,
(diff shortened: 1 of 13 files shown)


yangdanny97:
It would be better to fix these separately since that's easier to review, but since it's done already I'll try to review it all together

PR #3835, opened 2026-06-16
Fix unannotated attribute type inference bug (issue #1159)

### LINKED ISSUE #1159: use all usages within class (or within __init__) to infer type, not just initial assignment

QEDady:
# Summary

This PR resolves the unannotated attribute type inference bug where Pyrefly locked onto the first textual assignment to an attribute inside constructor methods (leading to false-positive `bad-assignment` errors and incorrect type inference).

### Behavioral Changes

This PR changes the following type-checking behaviors:

1. **Constructor Branch Unioning for Unannotated Attributes:** Pyrefly now computes the type of unannotated attributes as the union of all assignments across different control flow branches inside constructor methods. For example, an attribute initialized to `int` in the body of `__init__` and conditionally assigned an `str` will resolve to `int | str` rather than throwing a `bad-assignment` error on the reassignment.
2. **Order-Independent Class Method Receiver Visibility:** If an attribute is assigned via both a class receiver (`cls.val`) and an instance receiver (`self.val`) in constructors, it correctly resolves as class-visible (a class attribute) regardless of the declaration order of the constructors in the class body.
3. **Constructor Prioritization Over Helper Methods:** Assignments inside recognized constructors (like `__init__`, `__new__`, `__post_init__`) always take precedence over unrecognized helper methods. 
   * If multiple constructors assign to the same attribute (e.g. `__new__` and `__init__`), we union their types.
   * If the attribute is defined only in unrecognized helper methods, we union those helper method assignments (and flag them as implicit definitions if strict diagnostics are enabled).
   * However, if any constructor assignment exists, helper method assignments are ignored for type inference and will be flagged as `bad-assignment` errors in case types are incompatible.
4. **Preserved Explicit Declarations:** Explicit annotations (e.g. `self.val: int`) and class body definitions (e.g. `val = 0`) still strictly override type inference as before, ensuring that type constraints are still strictly checked.

### High-Level Implementation Details

To support these behavioral improvements, we transitioned Pyrefly's class analysis from a "first-assignment-wins" model to an accumulator-based model:
* **Collection:** Modified scope analysis to collect all assignments to an attribute across constructors (e.g. `__new__`, `__init__`, `__post_init__`) into a vector instead of discarding subsequent ones.
* **Visibility Tracking:** Added a consolidated `receiver_kind` field to track field visibility independently from individual method signatures, upgrading it to `Class` if any constructor assigns to `cls.val`.
* **Solver Unioning:** Updated the class field solver to solve the type of all collected assignments individually and union them.

Fixes #1159

# Test Plan

* **Unit & Integration Tests:** Added various test cases in `attributes.rs` covering branching unions, unrecognized helper methods, class-level priority, annotation overrides, and class method receiver union/upgrade behaviors.
* **Verification:** Ran `python3 test.py`

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1562,10 +1562,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 )
             }
             ClassFieldDefinition::DefinedInMethod {
-                value,
-                method,
+                values,
+                method: _,
                 annotation: annot,
-                ..
+                receiver_kind,
             } => {
                 let direct_annotation = annot.map(|a| self.get_idx(a).annotation.clone());
                 // Check if there's an inherited property or descriptor field from a parent class.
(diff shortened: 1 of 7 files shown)


stroxler (approved):

PR #3840, opened 2026-06-16
Fix untyped-import false positive for packages with py.typed marker (#3445)

### LINKED ISSUE #3445: `untyped-import` false positive on import from `requests` module
```py
from requests import check_compatibility  # error: untyped-import
```

`requests` now contains type annotations (has a `py.typed` file), see https://pypi.org/project/types-requests/:

> Note: The `requests` package includes type annotations or type stubs since version 2.34.0. Please uninstall the `types-requests` package if you use this or a newer version.

this rule should not be reported on libraries with a `py.typed` file

DibbayajyotiRoy:
Fixes #3445.
Before emitting untyped-import, this checks whether the resolved package ships a PEP 561 py.typed marker and skips the recommendation if so. The check walks up from the resolved module path by the module's component depth to find the package root, so it's correct for both top-level imports (import requests) and submodules (from requests.adapters import ...) — the marker lives at the package root per PEP 561, not in each submodule directory.

Notes:

Implicit namespace packages have no __init__/package root to hold a marker, so they remain treated as untyped.
Guards both UntypedImport sites (resolve_third_party_stub and combine_normal_and_stub_results).

Tests: typed package → no error; untyped package → still errors (no regression); submodule of a typed package (both package-dir and module-file forms) → no error.

Open to feedback on the approach, per the discussion.

### DIFF
--- a/pyrefly/lib/module/finder.rs
+++ b/pyrefly/lib/module/finder.rs
@@ -324,6 +324,44 @@ impl FindResult {
     }
 }
 
+/// Returns `true` if the top-level package directory for the resolved module
+/// contains a `py.typed` marker file (PEP 561).
+fn package_has_py_typed(
+    module: ModuleName,
+    result: &FindResult,
+    dir_cache: &DirEntryCache,
+) -> bool {
+    let depth = module.components().len().saturating_sub(1);
+    let mut package_root = match result {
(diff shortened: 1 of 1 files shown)


stroxler (approved):

PR #3842, opened 2026-06-17
fix new rule - ban raw-string quoted types #3761

### LINKED ISSUE #3761: new rule - ban raw-string quoted types
https://docs.astral.sh/ty/reference/rules/#raw-string-type-annotation

```py
def test() -> r"int": # should be an error
    ...
```

asukaminato0721:
# Summary

Fixes #3761

raw quoted type expressions like r"int" now emit InvalidAnnotation, while still parsing the annotation afterward to avoid extra cascade errors.

# Test Plan

add test


yangdanny97 (approved):
LGTM, i'll let @NathanTempest merge

stroxler (approved):

PR #3843, opened 2026-06-17
fix new rule - legacy TypeVar not assigned to variable #3755

### LINKED ISSUE #3755: new rule - legacy `TypeVar` not assigned to variable
```py
from typing import TypeVar

TypeVar("T") # should be an error
```

ty's [`invalid-legacy-type-variable`](https://docs.astral.sh/ty/reference/rules/#invalid-legacy-type-variable) rule catches this

asukaminato0721:
# Summary

Fixes #3755

bare statement calls to TypeVar, ParamSpec, and TypeVarTuple now emit invalid declaration errors.

# Test Plan

add test


stroxler (approved):

PR #3844, opened 2026-06-17
Report an error for a default on a method's `self`/`cls` parameter

### LINKED ISSUE #3729: report an error when assigning a default value to `self` or `cls`
```py
class Foo:
    def foo(self=1): ...
```

this is almost certainly a mistake, so it should be reported as an error

maybe related: #3666, but i think this should always be an error, regardless of whether the default value is the correct type

[basedpyright has a rule for this](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportselfclsdefault)

markselby9:
Closes #3729.

## Summary

A method's `self`/`cls` receiver is always supplied implicitly at call time, so a default value on it (e.g. `def m(self=1)`) is unreachable and almost always a mistake. Pyrefly already relied on the invariant that the receiver never has a default, yet silently accepted such definitions. basedpyright reports the same via `reportSelfClsParameterDefault`.

The check runs during function elaboration: a function has an implicit receiver exactly when `self_type` is `Some` (instance methods, classmethods including `__init_subclass__`/`__class_getitem__`, properties, and `__new__`; not staticmethods or top-level functions). When such a function's first positional parameter has a default, it emits a `bad-function-definition` error pointing at the default.

## Test Plan

New tests in `pyrefly/lib/test/self_cls_default.rs` cover instance / classmethod / positional-only / `__new__` / `__init_subclass__` / property receivers (flagged) and staticmethod, top-level function, and non-receiver params (not flagged). `cargo test` passes; full lib test suite is green.

### DIFF
--- a/pyrefly/lib/alt/function.rs
+++ b/pyrefly/lib/alt/function.rs
@@ -540,6 +540,33 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         } else if flags.is_staticmethod {
             self_type = None;
         }
+
+        // The `self`/`cls` receiver of a method is supplied implicitly at call time, so a
+        // default value on it is unreachable and almost always a mistake (e.g. `def m(self=1)`).
+        // `self_type` is `Some` exactly when there is an implicit receiver (instance methods,
+        // classmethods including `__init_subclass__`/`__class_getitem__`, properties, and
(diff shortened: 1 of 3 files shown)


yangdanny97:
Nice, this rule is already useful, I think mypy primer shows it is catching some pre-existing bugs in the wild

grievejia (approved):

PR #3845, opened 2026-06-17
Report `super()` in a directly-defined `NamedTuple` method

### LINKED ISSUE #3763: new rule - ban `super()` call in `NamedTuple`s
https://docs.astral.sh/ty/reference/rules/#super-call-in-named-tuple-method

```py
from typing import NamedTuple

class F(NamedTuple):
    x: int

    def method(self):
        super()  # should report an error here because this crashes at runtime
```

markselby9:
Closes #3763.

## Summary

A zero-argument `super()` inside a method of a class that directly subclasses `NamedTuple` makes the class fail to define at runtime: the compiler creates a `__class__` cell for `super()`, but the `NamedTuple` machinery does not propagate `__classcell__` to `type.__new__`, raising `RuntimeError`. Subclasses of a concrete `NamedTuple` are built by ordinary class machinery and are unaffected.

This records whether a `NamedTuple` is directly defined (vs. inherited) in its metadata and emits an `invalid-super-call` error for a no-argument `super()` resolved inside a directly-defined `NamedTuple` method.

Note: only the zero-argument form is detected. An explicit `super(F, self)` fails at runtime for the same reason, but the lexical enclosing class is not threaded into the explicit-args binding, so it is not flagged yet (documented with a `bug=` test).

## Test Plan

New tests in `pyrefly/lib/test/named_tuple.rs`: `super()`, `super().__repr__()`, and classmethod `super()` in a direct `NamedTuple` are flagged; (deep) subclasses and methods without `super()` are not; explicit `super(F, self)` is tracked as a known gap. `cargo test` passes; full lib test suite is green.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -3006,6 +3006,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 match &self.get_idx(*self_binding).0 {
                     Some(obj_cls) => {
                         let obj_type = self.as_class_type_unchecked(obj_cls);
+                        // A zero-arg `super()` in a method of a class that directly subclasses
+                        // `NamedTuple` makes the class fail to define at runtime: the `__class__`
+                        // cell the compiler creates for `super()` is not propagated by the
+                        // NamedTuple machinery. See https://github.com/facebook/pyrefly/issues/3763.
+                        if self
+                            .get_metadata_for_class(obj_cls)
+                            .named_tuple_metadata()
+                            .is_some_and(|nt| nt.is_direct)
+                        {
+                            self.error(
+                                errors,
+                                range,
+                                ErrorKind::InvalidSuperCall,
+                                "`super` call with no arguments is not allowed in a `NamedTuple` method".to_owned(),
+                            );
+                        }
                         let lookup_cls = self.get_super_lookup_class(obj_cls, &obj_type).unwrap();
                         let obj = if method.id == dunder::NEW {
                             // __new__ is special: it's the only static method in which the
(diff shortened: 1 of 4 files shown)


pyrefly/lib/alt/solve.rs
```diff
@@ -3006,6 +3006,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                        if self
+                            .get_metadata_for_class(obj_cls)
+                            .named_tuple_metadata()
+                            .is_some_and(|nt| nt.is_direct)
+                        {
+                            self.error(
+                                errors,
+                                range,
+                                ErrorKind::InvalidSuperCall,
+                                "`super` call with no arguments is not allowed in a `NamedTuple` method; it raises a `RuntimeError` when the class is defined".to_owned(),
```
yangdanny97: just "`super` call with no arguments is not allowed in a `NamedTuple` method" is enough, I don't think the latter part contains info that helps the user fix the issue

yangdanny97 (approved):
LGTM aside from one comment

@NathanTempest if you get around to merging this, maybe you can patch it to address the comment since it's small?

pyrefly/lib/alt/types/class_metadata.rs
```diff
@@ -581,6 +581,10 @@ pub struct NamedTupleMetadata {
     /// If true, the namedtuple fields were dynamically generated (e.g., using a
     /// generator or variable) and couldn't be statically resolved.
     pub has_dynamic_fields: bool,
+    /// True if this class directly subclasses `NamedTuple` (so it is built by the NamedTuple
```
yangdanny97: I think we can leave it at "Does this class directly extend `NamedTuple`?"

pyrefly/lib/alt/solve.rs
markselby9: Done — dropped the trailing clause; the message is now just "\`super\` call with no arguments is not allowed in a \`NamedTuple\` method".

pyrefly/lib/alt/types/class_metadata.rs
markselby9: Done — simplified the doc comment to "Does this class directly extend `NamedTuple`?".

grievejia (approved):

PR #3847, opened 2026-06-17
Dedupe inherited class FQN formatting in `pyrefly coverage`

jorenham:
# Summary

So it turns out there was one more module-prefix dedupe opportunity that I missed in #3816, which this applies.

# Test Plan

Tests still pass (refactor only; no behaviour changes)


rchen152 (approved):

PR #3848, opened 2026-06-17
Dedupe symbol counting in `pyrefly coverage`

jorenham:
# Summary

Both `build_module_report` and `filter_module_report_to_public` were counting symbols in the same way, so there was quite a bit of code duplication. This extracts the shared counting logic into a new helper function.

# Test Plan

Tests still pass (refactor only; no behavior changes)


rchen152 (approved):

PR #3849, opened 2026-06-17
Inline `collect_dunder_all_or_empty` in `pyrefly coverage`

jorenham:
# Summary

The single-line internal `collect_dunder_all_or_empty` helper function was only used once, so this inlines it.

And I hear you thinking: "Wow, awesome, this should definitely be a highlight on the release notes!". But to be honest, I'm not sure if this is as exciting as it sounds like.

# Test Plan

Just winging it


rchen152 (approved):

PR #3850, opened 2026-06-17
Dedupe return annotation matching in `pyrefly coverage`

jorenham:
# Summary

Here's yet another refactor related to `pyrefly coverage`, deduplicating the same `ReturnTypeKind` matching logic from `parse_functions` and `is_function_completely_annotated` by extracting it into a shared helper function.

# Test Plan

Tests still pass (refactor only; no behavior changes)


rchen152 (approved):

PR #3851, opened 2026-06-17
Dedupe the `--module` override rewriting in `pyrefly coverage`

jorenham:
# Summary

When `--module` is used with `pyrefly coverage`, the symbol prefixes are rewritten from the file-derived name to the override name. This rewriting logic was duplicated in `collect_module_reports` and the tests. 

This resolves the code duplication by moving the rewriting logic to `build_module_report` itself. This also micro-dedupes the formatting code in the two for loops of the rewriting logic itself.

# Test Plan

Tests still pass (refactor only; no behavior change)

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -1606,6 +1606,21 @@ fn build_module_report(
         }
     }
 
+    // A `--module` override renames the module, so rewrite symbol prefixes to match.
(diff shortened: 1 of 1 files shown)


jorenham:
Hm I see a big fat :x:. Is it a flake or something I broke?

yangdanny97:
internal CI is green on my end

rchen152 (approved):

PR #3855, opened 2026-06-17
Add retry and caching to our GitHub actions

samwgoldman:
# Summary

I noticed that we've had a bunch of actions fail recently due to transient network issues, where a step failed on one run, but succeeded on nearby runs both before and after.

This PR adds retry logic around these network requests, so we are more likely to recover from a transient failure. The retry logic is generic and reused throughout.

I also added caching, which in many cases can avoid the network requests altogether. Each cache has a conservatively-defined invalidation

This PR is broken into small commits which can be reviewed one-by-one for clarity.

# Test Plan

The modified actions ran on the ci-retry-cache-stack branch. I inspected the action output and saw that the retry and caching code paths are being hit.

### DIFF
--- a/.github/actions/retry-command/action.yml
+++ b/.github/actions/retry-command/action.yml
@@ -1,0 +1,55 @@
+name: Retry command
+description: Run a shell command with retries and exponential backoff.
+inputs:
+  command:
+    description: Command to run.
+    required: true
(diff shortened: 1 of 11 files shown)


stroxler (approved):

PR #3865, opened 2026-06-18
Support indirect match subjects (`match get_result()`)

### LINKED ISSUE #3730: `match funccall() ...` loses exhaustiveness / reports missing return**
### Describe the Bug

You can assign me to this, I will PR.

This issue talks about `match await ...` because that's how I noticed it, but also happens without await.

Pyrefly reports `bad-return` for an exhaustive `match await ...` statement, but accepts the equivalent code when the awaited result is first assigned to a local variable and then matched.
This affects common async code that matches on a result-like union.
### Repro
```python
from typing import NoReturn
class Ok[T]:
    __match_args__ = ("value",)
    value: T
class Err[E]:
    __match_args__ = ("value",)
    value: E
class NotFound:
    pass
def handle_error(error: NotFound) -> NoReturn:
    raise Exception()
async def get_result() -> Ok[list[int]] | Err[NotFound]:
    raise Exception()
async def f_match_await() -> list[int]:
    match await get_result():
        case Ok(value):
            return value
        case Err(error):
            handle_error(error)
async def f_assign_then_match() -> list[int]:
    result = await get_result()
    match result:
        case Ok(value):
            return value
        case Err(error):
            handle_error(error)
```
### Actual Behavior
Pyrefly reports:
```text
ERROR Function declared to return `list[int]`, but one or more paths are missing an explicit `return` [bad-return]
```
on `f_match_await`.
`f_assign_then_match` passes.
### Expected Behavior
Both functions should be accepted. `match await get_result()` should have the same exhaustiveness behavior as:
```python
result = await get_result()
match result:
    ...
```

Likely root cause: `match await ...` (like any function call that is not .get for examples) has no narrowing subject, so exhaustiveness analysis does not accumulate the `Ok` / `Err` negated narrows. Assigning the awaited value to a local gives Pyrefly a `NarrowingSubject::Name`, and the analysis works.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0A5XACUYDAK6V0AHRnoAxlFRw4dAPIBrANoAVALqJZdY3QD6pmqgbyAFqdSU2cc3QC8dABTSQAN1RRxGG8AGgBKIxM-AJhEOh1ZWUVlVQBRSkotFIMI43NLazsHJxd3L19-QJDw9BM6KMDYlISFJRVBXAYAMVxxLEMak2Jk5swYMDobDExYUxh03g856kpYoS6erFC6AFoAPnbRCSl%2B2spUCDgYOhT8eRhiBghcdA9q2WVSBTpR8bYxU0o8HEUAYrx2%2B00WigFwYWlYDD0ejoAB9ruktGtur1MNkBsYzhcrjc7g8ni83ugPl8fnQwBYrLZ7AB3c6grZ7OjQuCw%2BG42r5WyeVAsiD8P4MAFAkGvUInWomeTKK6aDz1GCynLy05iSQ1NWa2qKy5oyiLeaUDV4rUmSZYGZLBYOi3NKnyb5jWn2FQQNjoUwMGwwP0CmxgjlcnnoBFy-FS-juYWsujiyVwYFszUhuiAtMgmPyo3KjSqirq-PW7M6qR1UsGhVKk1m5aWiu1W3TGCzc1N3ihEDBEBkQFgKCkQgMWhQCgAYjoAAVSMPR3Q0Fg8Pg6PJnpA2JIrGTCLJZwBlGBXGwMBjEOCIAD0t6HY1HhF4bFvQdvmFw8jgt636B3PdHmeW9aV4OhUD8aBUGwWBN23H0gLJOhcFJZ44EPdAyADZ5th8OY4GQ9xvAAZkIABGAAmbxZC0J04D0WRei4Hg%2BBgTBtkwCBAXkR58LcOgAHIfXQXgYEE2RRIYbZAQAR3Ebj2O2DQYFIbZUHkO42ncQSWSkCT0BAABfAcNL4mBOmgGAKFXHACBIcgjKAA

_No response_

lennardwalter:
# Summary
Adds support for narrowing indirect `match` subjects, such as `match get_result():` and `match await get_result():` etc.

Previously, match narrowing depended on the subject being recoverable as a normal narrowing subject, like `match x:`. Non-name subjects did not have a stable internal subject identity, even though Python evaluates the match subject exactly once before matching. This meant Pyrefly could lose case-by-case narrowing information for exhaustiveness and control-flow analysis.

This PR introduces a synthetic internal match subject for non-name subjects. The synthetic subject lets the existing narrowing and exhaustiveness machinery treat the evaluated match expression like a stable temporary, while diagnostics still point at and display the original source expression.

This fixes cases where exhaustive matches over indirect subjects were incorrectly treated as possibly falling through, and enables non-exhaustive diagnostics for indirect subjects to refer to the real matched expression.

Fixes #3730 

# Test Plan
- Added tests and ran full suite. Also modified existing tests as now non-exhaustiveness is correctly reported for some.
  
# Follow-Up: Cleaner Synthetic Match Subject Representation
The implementation currently represents the internal temporary subject with a synthetic name, `"$match_subject"`. This is intentionally local-only and not user-facing, but it is still somewhat ad hoc.
A cleaner design would represent synthetic match subjects as a first-class internal narrowing subject, probably backed by the already-created `Idx<Key>` for the evaluated match expression. For example, `NarrowingSubject` could gain a temporary/internal variant rather than encoding the temporary as a fake `Name`.
Doing that properly likely requires changing `NarrowOps` to key by a subject identity instead of only by `Name`, since the current narrowing flow frequently looks up entries via `narrowing_subject.name()`. A future cleanup could introduce a `NarrowingSubjectKey` that can represent both real names and internal temporaries, removing the need for the string sentinel entirely. IMO it is fine like it is.

### DIFF
--- a/pyrefly/lib/binding/pattern.rs
+++ b/pyrefly/lib/binding/pattern.rs
@@ -18,7 +18,9 @@ use ruff_python_ast::Number;
 use ruff_python_ast::Pattern;
 use ruff_python_ast::PatternKeyword;
 use ruff_python_ast::StmtMatch;
+use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
+use ruff_text_size::TextRange;
 
 use crate::binding::binding::Binding;
 use crate::binding::binding::BindingExpect;
@@ -48,16 +50,37 @@ enum MatchSubject {
     None,
     /// A single match subject (e.g., `match x:`).
     Single(NarrowingSubject),
+    /// A local-only subject for matching non-name expressions (e.g., `match await f():`).
+    /// Python evaluates the subject once before matching, so we need a stable internal
+    /// subject for branch narrowing while diagnostics still point at the source expression.
+    Synthetic {
+        subject: NarrowingSubject,
+        display_subject_range: TextRange,
+    },
     /// Per-element subjects from a tuple match (e.g., `match x, y:`).
     Tuple(Vec<Option<NarrowingSubject>>),
 }
 
 impl MatchSubject {
-    /// Extract a single narrowing subject, if this is `Single`.
+    /// Extract a single narrowing subject, if available.
     fn as_single(&self) -> Option<&NarrowingSubject> {
         match self {
-            MatchSubject::Single(s) => Some(s),
+            MatchSubject::Single(s) | MatchSubject::Synthetic { subject: s, .. } => Some(s),
             _ => Option::None,
+        }
+    }
+
+    fn is_synthetic(&self) -> bool {
+        matches!(self, MatchSubject::Synthetic { .. })
+    }
+
+    fn display_subject_range(&self, fallback: TextRange) -> TextRange {
+        match self {
+            MatchSubject::Synthetic {
+                display_subject_range,
+                ..
+            } => *display_subject_range,
+            _ => fallback,
         }
     }
 }
@@ -554,7 +577,10 @@ impl<'a> BindingsBuilder<'a> {
         } else {
             match expr_to_subjects(&x.subject).first() {
                 Some(s) => MatchSubject::Single(s.clone()),
-                None => MatchSubject::None,
+                None => MatchSubject::Synthetic {
+                    subject: NarrowingSubject::Name(Name::new_static("$match_subject")),
+                    display_subject_range: x.subject.range(),
+                },
             }
         };
         let mut exhaustive = false;
@@ -667,7 +693,15 @@ impl<'a> BindingsBuilder<'a> {
         if exhaustive {
             self.finish_exhaustive_fork();
         } else {
-            let narrow_entries = self.build_narrow_entries(&negated_prev_ops);
+            let narrow_entries = if match_subject.is_synthetic() {
+                negated_prev_ops
+                    .0
+                    .values()
+                    .map(|(op, range)| (subject_idx, Box::new(op.clone()), *range))
+                    .collect()
+            } else {
+                self.build_narrow_entries(&negated_prev_ops)
+            };
             // Create BindingExpect only if we have a narrowing subject (for exhaustiveness warnings)
             if let Some(narrowing_subject) = match_subject.as_single()
                 && let Some((op, range)) = negated_prev_ops.0.get(narrowing_subject.name())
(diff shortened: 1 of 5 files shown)


pyrefly/lib/binding/pattern.rs
```diff
@@ -678,7 +712,14 @@ impl<'a> BindingsBuilder<'a> {
                         subject_idx,
                         narrowing_subject: narrowing_subject.clone(),
                         narrow_ops_for_fall_through: (Box::new(op.clone()), *range),
-                        subject_range: x.subject.range(),
+                        subject_range: match_subject.display_subject_range(x.subject.range()),
```
hurryabit, line 715: You're effectively matching on `match_subject` twice here but one of them is hidden. IMO, adding something like
```rs
let (subject_range, display_subject_range) = match match_subject {
    MatchSubject::Synthetic { display_subject_range, .. } => (display_subject_range, Some(display_subject_range)),
    _ => (x.subject.range(), None)
};
```
would make this more explicit and hence easier to follow. (This would also make the `display_subject_range` function unused.

pyrefly/lib/binding/pattern.rs
```diff
@@ -48,18 +50,39 @@ enum MatchSubject {
...
+            MatchSubject::Single(s) | MatchSubject::Synthetic { subject: s, .. } => Some(s),
             _ => Option::None,
         }
     }
+
+    fn is_synthetic(&self) -> bool {
+        matches!(self, MatchSubject::Synthetic { .. })
+    }
+
+    fn display_subject_range(&self, fallback: TextRange) -> TextRange {
```
hurryabit, line 77: Returning an `Option<TextRange>` and leaving it to the caller to determine what they want to do with a `None` seems more useful. This will also simplify some code below.

pyrefly/lib/binding/pattern.rs
hurryabit: Oops, I pressed the wrong button. This could also turn into a `let display_subject_range = match_subject.display_subject_range();` before the insert and
```rs
subject_range: display_subject_range.unwrap_or_else(|| x.subject.range()),
display_subject_range,
```

lennardwalter:
Thanks! Will gladly change this, however first would like to make sure that the solution with the maybe slightly unclean sentinel is acceptable. @hurryabit

lennardwalter:
any change to review this @yangdanny97 ? thank you

yangdanny97:
Still waiting on another maintainer to review it internally. I've stacked a separate patch on top that gets rid of the synthesized name

stroxler (approved):

yangdanny97:
Thanks for your contribution!

PR #3866, opened 2026-06-18
Make typeshed_path a full replacement for bundled stubs    

avokin2:
Problem
 
A configured typeshed_path only partially redirected import resolution. A user who pointed Pyrefly at their own typeshed still got results — and the declaration URIs
returned over LSP/TSP — out of Pyrefly's bundled stubs:
 
- Third-party imports always resolved against the bundled stubs, ignoring typeshed_path entirely.
- The bundled stdlib was used as a fallback whenever a module was missing from <typeshed_path>/stdlib. 
- Most importantly, the Stdlib object — which supplies the class definitions for int, str, list, … used throughout type checking — was always loaded from a hardcoded
bundled-typeshed config. So the builtins were bundled even when every other import resolved to the custom typeshed.
- typeshed_path could only be set from pyrefly.toml, leaving editors no way to point Pyrefly at a typeshed without writing a config file.
 
Fix
 
Treat typeshed_path, when set, as a complete replacement for the bundled stubs:
 
- Third-party stubs resolve only from <typeshed_path>/stubs, with each distribution subdirectory as its own search root (mirroring typeshed's layout). The bundled stdlib
branch is skipped: a miss falls through to the remaining stages (site-packages, …) and never reaches the bundled stubs.
- The stdlib loader config now carries typeshed_path, so compute_stdlib builds the Stdlib from <typeshed_path>/stdlib as well. invalidate_config drops the cached stdlib 
so a changed typeshed_path is picked up. 
 
Because the setting now suffices on its own, no separate opt-out knob is needed. 
 
Finally, surface typeshed_path over LSP via initializationOptions / workspace/configuration (as typeshedPath), deferring to an explicit value from a user-owned config 
file when one is present.
 
Summary
 
Makes a configured typeshed_path fully shadow Pyrefly's bundled stubs — for stdlib, third-party stubs, and the Stdlib builtins object — and lets editors set it over LSP 
without a config file. 
 
Test Plan 

 Added unit tests for each layer, and replaced the test that locked in the old bundled-always behavior.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/workspace.rs
+++ b/pyrefly/lib/lsp/non_wasm/workspace.rs
@@ -85,6 +85,11 @@ pub struct Workspace {
     pub stream_diagnostics: Option<bool>,
     pub diagnostic_mode: Option<DiagnosticMode>,
     pub workspace_config: Option<PathBuf>,
+    /// Custom typeshed override supplied over LSP. Equivalent to the
+    /// `typeshed_path` field of a `pyrefly.toml`, but settable via
+    /// `initializationOptions`/`workspace/configuration` so editor integrations
+    /// can point Pyrefly at their own typeshed without writing a config file.
+    pub typeshed_path: Option<PathBuf>,
 }
 
 impl Workspace {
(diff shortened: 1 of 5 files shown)


avokin2:
@yangdanny97, Could you check this MR when you have a chance?

rchen152:
Danny fixed this recently, so `--typeshed-path` works correctly in the dev preview (1.3.0-dev.1) of the next release.

PR #3868, opened 2026-06-18
fix new rule - empty body #3738

### LINKED ISSUE #3738: new rule - empty body
ty has a rule called [`empty-body` ](https://docs.astral.sh/ty/reference/rules/#empty-body) that detects functions with `...` as the body:

```py
def foo() -> int: ...  # error: [empty-body]
```

this should be reported as an error unless in a `.pyi` file or an `if TYPE_CHECKING` block

asukaminato0721:
# Summary

Fixes #3738

Added empty-body error kind and docs

Track ellipsis-only function bodies through binding/solving and emit when None is not assignable to the annotated return

Exempt .pyi, protocols, abstract methods, overloads, and if TYPE_CHECKING

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -154,6 +154,8 @@ pub enum ErrorKind {
     Deprecated,
     /// Division, floor division, or modulo by a literal zero value.
     DivisionByZero,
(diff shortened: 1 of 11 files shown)


yangdanny97:
> ✅ Improvement (13) pip (+1) This is a false positive.

hmmmmm

stroxler (approved):

PR #3877, opened 2026-06-20
fix(stubgen): emit callable-typed values as valid `typing.Callable`

tobyh-canva:
## Context

**Bug:** When `pyrefly stubgen` emits a value whose inferred type is a **callable** (for example a function assigned to a module-level variable, or a method reference), it renders the annotation using pyrefly's internal type-display form instead of valid Python.

### Example

Run stubgen on this source code:

```py
def transform(x: int, s: str) -> bool:
    return bool(x) and bool(s)


def variadic(*args: int) -> str:
    return ""


handler = transform
collect = variadic
```

#### Expected output

```py
from typing import Callable

def transform(x: int, s: str) -> bool: ...


def variadic(*args: int) -> str: ...


handler: Callable[[int, str], bool]
collect: Callable[..., str]
```

#### Actual output

As you can see, the two final lines are syntactically invalid.

```py
def transform(x: int, s: str) -> bool: ...


def variadic(*args: int) -> str: ...


handler: (x: int, s: str) -> bool
collect: (*args: int) -> str
```

## Changes
- Add a stubgen snapshot regression test (`callable_values`), including the example shown above.
- Add logic in stubgen to render callable type annotations using valid Python syntax

Note: overloaded callables are emitted as `Incomplete` for now; rendering them faithfully is left as a follow-up.

### DIFF
--- a/pyrefly/lib/stubgen/emit.rs
+++ b/pyrefly/lib/stubgen/emit.rs
@@ -18,15 +18,22 @@ use crate::stubgen::extract::StubVariable;
 pub fn emit_stub(stub: &ModuleStub) -> String {
     let mut out = String::new();
 
+    let mut typing_names = Vec::new();
+    if stub.uses_callable {
+        typing_names.push("Callable");
+    }
     if stub.uses_self {
-        out.push_str("from typing import Self\n");
+        typing_names.push("Self");
+    }
+    if !typing_names.is_empty() {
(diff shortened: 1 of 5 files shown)


yangdanny97:
Another option for overloads is to do `Callable[..., Incomplete]`

tobyh-canva:
Ah good point, I'll switch over to that instead of a plain `Incomplete`.

stroxler (approved):

Issue #3819, opened 2026-06-15
`unsupported-operation` when matching on overloads


MarcoGorelli:
### Describe the Bug

```py
from __future__ import annotations
from datetime import timedelta
from typing import (
    Generic,
    TypeVar,
    Protocol,
    overload,
)

T_contra = TypeVar("T_contra", contravariant=True)
S1_co = TypeVar("S1_co", bound=timedelta | int | float, covariant=True)
S2_co = TypeVar("S2_co", bound=timedelta | int | float, covariant=True)


class SupportsProtoTrueDiv(Protocol[T_contra, S2_co]):
    def _proto_truediv(self, other: T_contra, /) -> ElementOpsMixin[S2_co]: ...


class ElementOpsMixin(Protocol, Generic[S2_co]):
    @overload
    def _proto_truediv(
        self: ElementOpsMixin[int], other: int, /
    ) -> ElementOpsMixin[float]: ...
    @overload
    def _proto_truediv(
        self: ElementOpsMixin[timedelta], other: timedelta, /
    ) -> ElementOpsMixin[float]: ...


class Series(ElementOpsMixin[S2_co], Protocol):
    def __truediv__(
        self: SupportsProtoTrueDiv[T_contra, S1_co],
        other: T_contra,
    ) -> Series[S1_co]: ...


def main2(s: Series[timedelta]) -> None:
    td = timedelta(1)
    _ = s / td
```

```console
(.venv) mgorelli@marcoslaptop:~/pandas-stubs-dev$ mypy t.py 
Success: no issues found in 1 source file
(.venv) mgorelli@marcoslaptop:~/pandas-stubs-dev$ pyright t.py 
0 errors, 0 warnings, 0 informations
(.venv) mgorelli@marcoslaptop:~/pandas-stubs-dev$ ty check t.py 
All checks passed!
(.venv) mgorelli@marcoslaptop:~/pandas-stubs-dev$ pyrefly check t.py 
ERROR `/` is not supported between `Series[timedelta]` and `timedelta` [unsupported-operation]
  --> t.py:39:9
   |
39 |     _ = s / td
   |         -^^^--
   |         |   |
   |         |   has type `timedelta`
   |         has type `Series[timedelta]`
   |
  Argument `timedelta` is not assignable to parameter `other` with type `int` in function `Series.__truediv__`
 INFO 1 error
```

This is based on an issue in `pandas-stubs`, which I've done my best to minify

[...]

PR #3895, opened 2026-06-22
Fix stack overflow on self-referential protocol overload filtering

### LINKED ISSUE #3893: Stack overflow
### Describe the Bug

Since `1.1.1`, pyrefly crashes when checking https://github.com/cusp-ai-oss/kUPS/ 
```sh
git clone github.com/cusp-ai-oss/kUPS/
cd kUPS
uvx pyrefly check src/
```
Output 
```
uvx pyrefly check src/
[00:00:00] ██████░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░      69/483    
thread '<unknown>' (90454) has overflowed its stack
```

mikeleppane:
Fixes https://github.com/facebook/pyrefly/issues/3893

## What

pyrefly aborts with a native stack overflow when it checks certain self-referential protocols. The smallest reproduction is a protocol whose overloaded method annotates `self` with the protocol itself:

```python
from typing import Protocol, TypeVar, overload

S = TypeVar("S", covariant=True)
R = TypeVar("R", covariant=True)

class Lens(Protocol[S, R]):
    @overload
    def __call__[S2, R2](self: Lens[S2, R2], state: S2, /, value: R2) -> S2: ...
    @overload
    def __call__[S2, R2](self: Lens[S2, R2], state: S2, /) -> R2: ...

class BaseLens(Lens[S, R], Protocol):
    def at(self) -> Lens[S, R]:
        return self
```

Before this change pyrefly prints `thread '<unknown>' has overflowed its stack / fatal runtime error: stack overflow` and aborts. After it, the file type-checks with no crash. mypy, pyright, and ty all terminate on this input, so the crash was a pyrefly bug rather than bad code.

It shows up on real projects. `pyrefly check src/` on the kUPS library (https://github.com/cusp-ai-oss/kUPS) crashed partway through; it now runs to completion. The crash is a regression since 1.1.1.

## Why

When pyrefly checks a value against a protocol, it filters the candidate's overloaded methods by their explicit `self:` annotations, so an overload whose `self:` cannot match the receiver does not get to bind type variables it should not. That filter landed in 1.1.1 (#3819).

The filter tests each overload with `is_subset_eq(self_type, self_param)`. When `self_param` is itself a protocol, that check re-enters protocol conformance, looks up the same overloaded member again, and re-filters the same `(self_type, self_param)` pair. Nothing breaks the loop, so it recurses until the stack is exhausted.

pyrefly already has two cycle guards for protocol subtyping, and neither catches this one. Both live on `Subset`, and `Solver::is_subset_eq` builds a fresh `Subset` on every call. The recursion passes through `is_subset_eq` at each turn, so the `(got, want)` subset cache and the `class_protocol_assumptions` coinductive set are both empty again by the time the same pair comes back around.

## How

The guard goes where the recursion is actually observable. It lives on `ThreadState`, which persists across the `is_subset_eq` boundary that resets `Subset`.

- `pyrefly/lib/alt/answers_solver.rs`: a per-thread `overload_self_filter_stack` of in-progress `(self_type, self_param)` pairs, plus `enter_overload_self_filter` and `exit_overload_self_filter`.
- `pyrefly/lib/alt/class/class_field.rs`, in `filter_overloads_by_self_type`: before testing a protocol-typed `self:`, record the pair; on re-entry of the same pair, assume the overload applies and return. That is the greatest-fixpoint reading of a recursive protocol, which matches mypy and pyright.

A non-protocol `self:` cannot start this recursion, so it skips the guard and runs the plain check. That keeps the common path free of the extra bookkeeping, including hot cases like `str` methods with `self: LiteralString` overloads. The guard only changes the filter's decision on a real cycle, where keeping the overload is the safe direction. The conformance verdict still comes from the attribute-subset check that runs afterward, so a class that does not satisfy the protocol is still rejected.

This leaves the subtyping solver, the existing `Subset` cycle guards, and the unbound class-access caller of the same filter unchanged.

## Test plan

Three tests in `pyrefly/lib/test/protocol.rs`:

| Test | What it pins |
|------|--------------|
| `test_protocol_overloaded_generic_self_referencing_protocol_terminates` | the issue repro: a protocol with a self-typed overloaded `__call__` type-checks instead of crashing |
| `test_protocol_overloaded_generic_self_mutual_recursion_terminates` | two protocols whose `self:` annotations reference each other also terminate |
| `test_protocol_overloaded_generic_self_non_conforming_still_rejected` | a class that matches `__call__` but lacks another required member is still reported as not assignable, so the guard does not hide a real failure |

Each terminate test aborts the test runner without the fix. The existing `test_protocol_overloaded_method_filtered_by_self` (#3819) still passes, so filtering by concrete protocol self-types is unchanged.

- `cargo test -p pyrefly --lib`: 5746 passed, 0 failed.
- Formatting and linting clean.
- `mypy_primer` over 69 protocol-heavy and overload-heavy projects (pydantic, pandas-stubs, pandera, sympy, xarray, scipy, attrs, strawberry, kornia, jax, and more): no diagnostic changes.

### DIFF
--- a/pyrefly/lib/alt/answers_solver.rs
+++ b/pyrefly/lib/alt/answers_solver.rs
@@ -21,6 +21,7 @@ use std::sync::OnceLock;
 use dupe::Dupe;
 use dupe::IterDupedExt;
 use fxhash::FxHashMap;
+use fxhash::FxHashSet;
 use pyrefly_graph::calculation::Calculation;
 use pyrefly_graph::calculation::ProposalResult;
 use pyrefly_graph::index::Idx;
@@ -1602,6 +1603,10 @@ pub struct ThreadState {
     /// is pushed here. When the nested call takes its sink, the previous
     /// one is restored.
     trace_sink_stack: RefCell<Vec<Option<TraceSideEffects>>>,
+    /// `(self_type, self_param)` pairs whose overload-self-type compatibility
+    /// check is currently in progress, used as a coinductive guard against
+    /// self-referential protocols. See `filter_overloads_by_self_type`.
+    overload_self_filter_stack: RefCell<FxHashSet<(Type, Type)>>,
 }
 
 impl ThreadState {
@@ -1614,6 +1619,7 @@ impl ThreadState {
             lambda_param_vars: RefCell::new(FxHashMap::default()),
             trace_sink: RefCell::new(None),
             trace_sink_stack: RefCell::new(Vec::new()),
+            overload_self_filter_stack: RefCell::new(FxHashSet::default()),
         }
     }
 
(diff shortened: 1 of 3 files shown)


yangdanny97:
Thanks for investigating this, we did change soem stuff about the solver last version so this seems like a plausible root cause. will review in more detail soon

stroxler (approved):

Issue #3900, opened 2026-06-22
Pyrefly incorrectly flags valid `dataclass` field named `self` as duplicate and wrong


dillydill123:
### Describe the Bug

Pyrefly 1.1.1 incorrectly reports an error when a `dataclass` (or Pydantic `BaseModel`) defines a field named `self`. However, this is valid Python code and runs fine.
```
from dataclasses import dataclass
@dataclass
class MyDataClass:
    self: str
y = MyDataClass(self="test")
print(y.self)
```
This produces the following:
```
ERROR sandbox.py:5:17-28: Multiple values for argument `self` in function `MyDataClass.__init__` [[bad-keyword-argument](https://pyrefly.org/en/docs/error-kinds/#bad-keyword-argument)]
ERROR sandbox.py:5:22-28: Argument `Literal['test']` is not assignable to parameter `self` with type `MyDataClass` in function `MyDataClass.__init__` [[bad-argument-type](https://pyrefly.org/en/docs/error-kinds/#bad-argument-type)]
```

For some context, I have a Pydantic `BaseModel` representing a [JIRA API response](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-issueidorkey-get), where one of the fields is a URL to the `Issue` called `self`. So while it is not ideal to name a field `self`, there are practical reasons to do it. The error is easily seen with just a `dataclass` as well.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEmqALqgMZSpxzx0Q3G5KTBszYcuAHXQABRi3ac4UhVzoBZUgBFRAYXFxEUusbrcoYRKaaUppOgF51W3foAUZsPYkgm8Jt4BKKWJKCHQmV1JCDwCQABoQMkoYMCgoplooCgBiOgAFUmTUuzQsPHw6Vlx0SABzAFdKZghqwilcgGUYGDoACyYmYgMAemGklLTCQVrhmHRhzFxWOGGqmogGpqYW%2BbowQTpUADdUaFRsWErqusbm6rpcYm3quDb0MiZe6oBaI5hKOA7Bx0bwAZkIAEYAEzeYLJbjCRzeODWCCsfwgEAAXwSbG2fwAYtAYBRSjgCCRyFigA

yangdanny97:
[...]

PR #3902, opened 2026-06-23
Fix dataclass/BaseModel field named "self" incorrectly flagged as duplicate (issue #3900)

kz357:
# Summary

A "@dataclass" (or Pydantic "BaseModel") field named "self" was incorrectly flagged with "bad-keyword-argument" and "bad-argument-type" errors on construction, e.g. "MyDataClass(self="test")", even though this is valid Python that runs fine.

Instance receiver is bound by prepending it as a positional argument rather than dropping the param so two params named `self` caused the keyword "self=" to bind to the instance param instead of the field.

Fix is to mirror CPython dataclasses as follows: when there exists a field named "self", name the instance param "__dataclass_self__" instead. Pydantic BaseModel also covered by this change since its init goes through get_dataclass_init.

Fixes issue #3900 + dillydill123's comment about Pydantic BaseModel

# Test Plan

Added two regression tests, one for dataclass and one for BaseModel, both passing.

- test_dataclass_field_named_self in ./pyrefly/lib/test/dataclasses.rs
- test_pydantic_field_named_self in ./pyrefly/lib/test/pydantic/field.rs
Both construct instances via "self=" keyword and assert field type with "assert_type". Both reproduce #3900 errors on main and pass with the fix. 

no diffs!

### DIFF
--- a/pyrefly/lib/alt/class/dataclass.rs
+++ b/pyrefly/lib/alt/class/dataclass.rs
@@ -986,7 +986,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         kw_only_by_class: &SmallMap<Class, SmallSet<Name>>,
         errors: &ErrorCollector,
     ) -> ClassSynthesizedField {
-        let mut params = vec![self.class_self_param(cls, false)];
+        /// detects when a field named "self" exists and names receiver param "__dataclass_self__" which is what CPython dataclasses does at runtime
+        let self_param = if dataclass.fields.contains(&Name::new_static("self")) {
+            Param::Pos(
+                Name::new_static("__dataclass_self__"),
+                self.instantiate(cls),
+                Required::Required,
+            )
+        } else {
+            self.class_self_param(cls, false)
+        };
+        let mut params = vec![self_param];
         let mut has_seen_default = false;
         for (name, field, field_flags) in self.iter_fields(cls, dataclass, true, kw_only_by_class) {
             let strict = field_flags.strict.unwrap_or(strict_default);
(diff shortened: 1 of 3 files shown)


kz357:
force-with-lease pushed user email change to mine, because I think meta CLA needs it to be my email (?)

pyrefly/lib/alt/class/dataclass.rs
```diff
@@ -1032,7 +986,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         kw_only_by_class: &SmallMap<Class, SmallSet<Name>>,
         errors: &ErrorCollector,
     ) -> ClassSynthesizedField {
-        let mut params = vec![self.class_self_param(cls, false)];
+        /// detects when a field named "self" exists and names receiver param "__dataclass_self__" which is what CPython dataclasses does at runtime
+        let self_param = if dataclass.fields.contains(&Name::new_static("self")) {
+            Param::Pos(
```
yangdanny97, line 1077: is it Pos or PosOnly?

as in, at runtime can you pass `__dataclass_self__` by name? if not then it's PosOnly

yangdanny97: nvm I checked and it can be passed by name...

yangdanny97 (approved):
LGTM, I'll let @NathanTempest merge

NathanTempest (approved):
Thanks for the great fix, @kz357 and welcome aboard!. LGTM. One micro-nit: the new comment is a doc comment but it's inside a function body which is only valid before items, so it'll trigger an unused_doc_comment warning. I will import this and change it myself as this is a pretty small fix.

rchen152 (approved):

PR #3905, opened 2026-06-23
Report invalid-abstract-method for @abstractmethod in non-abstract classes

### LINKED ISSUE #3728: no error when `@abstractmethod` is used on a non-`ABC`
### Describe the Bug

```python
from abc import abstractmethod


class Foo:
    @abstractmethod # no error
    def fn(self): ...

foo = Foo() # no error
```

i guess it's up for discussion whether the error should be reported on the definition or the usage. [in basedpyright we went with the former](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportinvalidabstractmethod), but that was only out of caution since performance was the cited reason for this not being supported upstream. not sure how relevant that is for pyrefly's implementation of abstract classes though

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEq2AxnRDcbpQC4PZxeVUTLjRhcAFrkwAdStPSz5i9Eyio4cOgDFcuRMrqG6AAUb9Bw0RKkGjmGGDph0ACjgwoYAJSI6hf8rKYLp0ALzaui5eIAA0IGSUDlCkhFy0UBQAxHQACqSJYMl0aFh4%2BHRMuOiQAOYAroJcEFWE8tkAyjAwdOJcXMRwiAD0QwlJKZw1QzDoQ5i4THBDldUQ9Y3Ns06cDABuqNCMsBVVtQ2oTVV0uMSX6HCt6GTW6AC0uzCUcJthdNIgAGZCABGABM-3kAG1PtQvgBdeR1dBsDjcGCYV6YCCJYQQD6-ADka3QnBgBPkJK4r0SAEc6tj0a8ANYwUivIRMeCacIEgDuqEo6HJyPYUAgTAgVIwpEJMM4wpg%2BGIYolUvQMp5csowpRKslyUxDgg6EZFwEEGwdS43U1lFhwqRdXcGOJpMJ-MFDvQiSE4iONrofIFQvkNAgGmNNVeuA%2Bdog9kNlUaOx5HpDIuV4sl7L4AiEVNU6m5gbTBJAAF84vm8TAtNAYBQSjgCCRyOWgA

_No response_

nitishagar:
Summary: Adds a new off-by-default lint rule `invalid-abstract-method` that
fires when a method decorated with `@abstractmethod` is defined in a class
that is not abstract (does not inherit from `abc.ABC`, use `abc.ABCMeta`,
have a transitive ABC base, or be a Protocol). Such a class is directly
instantiable at runtime while containing an unimplemented method — a likely
programming mistake.

The check runs inside `solve_abstract_members` (which already executes for
every class), guarded by `!extends_abc() && !is_protocol() && !is_new_type()`.
It iterates the class's own declared fields (not MRO-inherited ones) and emits
one error per field whose `ClassField::is_abstract()` flag is set, pointing at
the method's definition range. The rule defaults to `Severity::Ignore` to
avoid breaking existing code without opt-in.

Fixes https://github.com/facebook/pyrefly/issues/3728

Test Plan:
- cargo build -p pyrefly_config  (clean)
- cargo build -p pyrefly  (clean)
- cargo test -p pyrefly_config -- test_doc_headers test_doc_severities  (2 passed)
- cargo test -p pyrefly --lib abstract_methods  (35 passed)
- cargo test -p pyrefly --lib  (5824 passed, 0 failed)
- python3 test.py --no-test --no-conformance --no-jsonschema  (clean, no new lint)

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -202,6 +202,8 @@ pub enum ErrorKind {
     InconsistentOverloadDefault,
     /// Internal Pyrefly error.
     InternalError,
+    /// An `@abstractmethod` is defined in a class that is not abstract.
+    InvalidAbstractMethod,
     /// Attempting to write an annotation that is invalid for some reason.
     InvalidAnnotation,
     /// Passing an argument that is invalid for reasons besides type.
@@ -469,6 +471,7 @@ impl ErrorKind {
             ErrorKind::DivisionByZero => Severity::Warn,
             ErrorKind::ExplicitAny => Severity::Ignore,
             ErrorKind::ImplicitAbstractClass => Severity::Ignore,
+            ErrorKind::InvalidAbstractMethod => Severity::Error,
             ErrorKind::ImplicitAny => Severity::Ignore,
             ErrorKind::ImplicitAnyAttribute => Severity::Ignore,
             ErrorKind::ImplicitAnyEmptyContainer => Severity::Ignore,
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -2693,8 +2693,47 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 errors,
             );
             self.check_variance_for_class(cls, class_bases.as_ref(), &class_field_map, errors);
+            self.check_invalid_abstract_methods(cls, &class_field_map, errors);
         }
         Arc::new(EmptyAnswer)
+    }
+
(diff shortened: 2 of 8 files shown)


yangdanny97:
I don't think this one needs to be off-by-default

nitishagar:
@yangdanny97 I see. I've made `invalid-abstract-method` on by default (`Severity::Error`) and updated the docs and tests accordingly. A handful of existing tests that placed `@abstractmethod` on non-ABC classes now inherit from `abc.ABC` to match their intent. Let me know if such approach makes sense or we default to warning?

connernilsen (approved):

PR #3906, opened 2026-06-23
Add off-by-default rule for unused call results

### LINKED ISSUE #3719: lint rule - unused call result
one of my favorite pyright rules is `reportUnusedCallResult`, which warns when you may be misusing a function that returns a value by not actually using the result. for example:

Code sample in [basedpyright playground](https://basedpyright.com/?typeCheckingMode=all&code=CYUwZgBAxg9gtgIwJYDsQG0AqBdAFEgFxDgGcAuCAGyRIK2wBoI0B3AfUOIswEoIBaAHxUadHGQBQEaRABOIAgFdZKCOgBUnUk1YcicbBKNaSEALxqAjAwBMhiQGIIcGopIgIMSAQAWH2IioHmCKKFAESDAoTL4ebiCyEL4wigDmPgQQhM6KBACGRKaxciAkipSZCLkQeeGKeZSUAJ4lSiqmecwgLCK0kgHIaPj6JEwAzHxOCbIwshQASqXlmV7QDZQQIAAeAA7yJCSRqjSe3k07HgBE1LToqATYlzUowFmmKDCZ8cAA3DUHSFSqgIMAgADc8rIkHkEJQrmwnkhvD4TqiUER0UcGkYJHt7sNiCRJp5KK8IZRFCAgA)

```python
def combine[T](items: list[T], new_item: T) -> list[T]:
    return [*items, new_item]


items = [1,2]

# misuse of the combine function, the user thought it mutates the result but actually returns a new list:
combine(items, 3) # error: Result of call expression is of type "list[int]" and is not used; assign to variable "_" if this is intentional


print(items) # old value
```

it would be great if pyrefly had an option to warn against this.

basically the same as `unused-coroutine` but for any return type, not just coroutines.

nitishagar:
Introduce an `unused-call-result` diagnostic that fires when a bare
expression statement is a function/method call whose result is discarded
and the result type is informative (not `None`, `Any`, or `Never`).

The check sits in `binding_to_type_stmt_expr` as an `else if` branch
after the existing `unused-coroutine` check, so the two diagnostics
are mutually exclusive: a discarded coroutine call still emits only
`unused-coroutine`. The guard `matches!(e, Expr::Call(_))` ensures
assignments, bare names, comparisons, and similar non-call statements
are not flagged. Results of type `Any` or that extend `Any` are also
exempted, consistent with the coroutine rule. The rule defaults to
`Severity::Ignore` and must be explicitly enabled.

A new test file covers every spec boundary: the golden-path discard,
assigned/None/Any/Never/extends-Any exemptions, method and constructor
calls, non-call bare statements, the R5 coroutine mutual-exclusion case
(rule ON + coroutine discard → exactly one `unused-coroutine` error),
and the default-off regression.

Fixes https://github.com/facebook/pyrefly/issues/3719

Test Plan:
- cargo test -p pyrefly_config test_doc_headers test_doc_severities
- cargo test -p pyrefly --lib unused_call_result
- cargo test -p pyrefly --lib simple::test_unused_coroutine

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -370,6 +370,8 @@ pub enum ErrorKind {
     UnsupportedOperation,
     /// Import is missing an expected stubs package
     UntypedImport,
+    /// Result of a call expression is not used.
+    UnusedCallResult,
     /// Result of async function call is never used or awaited
     UnusedCoroutine,
     /// A suppression comment is unused (no error to suppress, or specific codes are unused)
@@ -502,6 +504,7 @@ impl ErrorKind {
             ErrorKind::UnreachableMatchCase => Severity::Warn,
             ErrorKind::UnresolvableDunderAll => Severity::Warn,
             ErrorKind::UntypedImport => Severity::Warn,
+            ErrorKind::UnusedCallResult => Severity::Ignore,
             ErrorKind::UnusedIgnore => Severity::Ignore,
             ErrorKind::UnusedTypeIgnore => Severity::Ignore,
             ErrorKind::VarianceMismatch => Severity::Warn,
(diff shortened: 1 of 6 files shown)


connernilsen (approved):

PR #3914, opened 2026-06-23
fix enum hover should show fields, not call #3889

### LINKED ISSUE #3889: enum hover should show fields, not call
### Describe the Bug

Image

but for an enum, show the enums makes more sense.

```
from http import HTTPStatus
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3889

enum classes display their enum member literal union instead of constructor signatures.

# Test Plan

add test for hover and `__new__` test

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -792,6 +792,33 @@ pub fn get_hover(
                 return display;
             }
             if show_constructor {
+                let enum_class = match &cloned {
+                    Type::ClassDef(cls) => Some(cls),
+                    Type::ClassType(cls) => Some(cls.class_object()),
+                    Type::Type(t) => match &**t {
+                        Type::ClassType(cls) => Some(cls.class_object()),
+                        _ => None,
+                    },
+                    _ => None,
+                };
+                if let Some(cls) = enum_class
+                    && solver.get_metadata_for_class(cls).is_enum()
+                {
+                    let members: Vec<Type> = solver
+                        .get_enum_members(cls)
+                        .into_iter()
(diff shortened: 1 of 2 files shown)


pyrefly/lib/test/lsp/hover.rs
```diff
@@ -988,6 +988,46 @@ Test docstring
     );
 }
 
+#[test]
+fn hover_on_imported_enum_shows_members() {
+    let code = r#"
+from http import HTTPStatus
+#                ^
+"#;
+    let report = get_batched_lsp_operations_report(&[("main", code)], get_test_report);
```
yangdanny97: can we just assert the full report text here?

yangdanny97:
I'm not sure I like the way that's rendered, extremely large Literal unions like the one in the test case are hard to read IMO, so this could potentially worsen readability.

cc @kinto0 whether it's worth adding a special case for this

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -792,6 +792,33 @@ pub fn get_hover(
                 return display;
             }
             if show_constructor {
+                let enum_class = match &cloned {
```
kinto0: could you move the constructor related logic into a helper function? get_hover is much too large for a human to understand what's going on here at first glance

kinto0:
> I'm not sure I like the way that's rendered, extremely large Literal unions like the one in the test case are hard to read IMO, so this could potentially worsen readability.
> 
> cc @kinto0 whether it's worth adding a special case for this

I'm kinda fine with it. we don't limit other union sizes. if we want to we should do it for all of them.

yangdanny97 (approved):

PR #3915, opened 2026-06-23
fix Incorrect union narrowing with is EnumMember #3820

### LINKED ISSUE #3820: Incorrect union narrowing with `is EnumMember`
I have a small repro of our real code in https://github.com/msgspec/msgspec/pull/1070#pullrequestreview-4497690765:

```python
from typing import final, Final, Literal, assert_type
import enum

@final
class UnsetType(enum.Enum):
    UNSET = "UNSET"
    def __bool__(self) -> Literal[False]: ...

UNSET: Final = UnsetType.UNSET

assert_type(UNSET, Literal[UnsetType.UNSET])

def check_unset_type_lowering(x: int | UnsetType) -> None:
    if x is UNSET:
        # this looks like a bug in `pyrefly`:
        assert_type(x, Literal[UnsetType.UNSET])
```

fails with:

```
» pyrefly check ex.py
ERROR assert_type(UnsetType, Literal[UnsetType.UNSET]) failed [assert-type]
  --> ex.py:16:20
   |
16 |         assert_type(x, Literal[UnsetType.UNSET])
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
 INFO 1 error
```

Both `mypy` and `pyright` pass this test.
The same happens if I change the condition to be:

```python
if x in UnsetType.UNSET:
    assert_type(x, Literal[UnsetType.UNSET])
```

Version:

```
» pyrefly --version
pyrefly 1.0.0
```

asukaminato0721:
# Summary

Fixes #3820

positive `is` narrowing preserves a single surviving enum literal instead of letting union simplification widen it back to the enum class.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -182,7 +182,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     }
 
     fn intersect_impl(&self, left: &Type, right: &Type, fallback: IntersectFallback) -> Type {
-        if self.is_subset_eq(right, left) {
+        if Self::enum_instance(left, right).is_some() {
+            right.clone()
+        } else if Self::enum_instance(right, left).is_some() {
+            left.clone()
+        } else if self.is_subset_eq(right, left) {
             if left.is_toplevel_callable()
                 && right.is_toplevel_callable()
                 && self.is_subset_eq(left, right)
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/narrow.rs
```diff
@@ -339,6 +339,26 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
     }
 
+    /// Narrow a type to values identity-equal to `right` (`is` semantics).
+    fn narrow_is(&self, ty: &Type, right: &Type) -> Type {
+        let mut narrowed = Vec::new();
+        self.map_over_union(ty, |t| {
+            let t = if Self::enum_instance(t, right).is_some() {
```
yangdanny97: should we just put this into our intersection logic, that enum intersected w/ its member is the member?

that way we don't need an extra helper to handle this

PR #3919, opened 2026-06-23
Let `# pyrefly: ignore` exclude symbols from `pyrefly coverage`

jorenham:
# Summary

The `pyrefly coverage` commands now respects `# pyrefly: ignore`. Because, well, they didn't before... An ignored symbol was still reported and still counted against the percentage. If unused ignores were configured, then `pyrefly check` would also complain about them.

Now, we can do things like

```py
def legacy(x):  # pyrefly: ignore[coverage-missing]
    ...
```

which will cause `legacy` to not be reported as a warning anymore by `coverage check`, and will no longer affect the total coverage percentages of `coverage check` and `coverage report` (counted as 0 typables).

Specifically ignoring `[coverage-partial]` and a bare `# pyrefly: ignore` work too. 
However, I chose to not include `# type: ignore`, even if explicitly configured. That's because those have always been intended for type errors, and also because Pyrefly is the only type-coverage checker out there, so it's probably safe to assume that a `# type: ignore` wasn't intended to ignore coverage.

A coverage ignore that matches nothing is flagged as unused by `coverage check`, and plain `pyrefly check` no longer reports `coverage-*` codes as unused (they're only for `coverage check`).

For decorated symbols like a `@property`, the `# pyrefly: ignore` should be placed on the decorator line (that's where the coverage diagnostic starts).

# Test Plan

Added tests for exclusion (incl. properties), unused/out-of-scope ignores, the `type: ignore` boundary, and `check`  /  `report` consistency.

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -8,14 +8,17 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use dupe::Dupe;
 use pyrefly_build::handle::Handle;
 use pyrefly_config::error_kind::ErrorKind;
+use pyrefly_config::error_kind::Severity;
 use pyrefly_config::finder::ConfigFinder;
 use pyrefly_graph::index::Idx;
 use pyrefly_python::dunder;
+use pyrefly_python::ignore::Tool;
 use pyrefly_python::module::Module;
 use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::module_path::ModuleStyle;
@@ -27,6 +30,7 @@ use pyrefly_types::class::ClassType;
 use pyrefly_types::types::Type;
 use pyrefly_util::forgetter::Forgetter;
 use pyrefly_util::includes::Includes;
+use pyrefly_util::lined_buffer::LineNumber;
 use pyrefly_util::lock::Mutex;
 use pyrefly_util::thread_pool::ThreadCount;
 use ruff_python_ast::Expr;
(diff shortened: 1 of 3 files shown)


NathanTempest (approved):
Thanks for the well-thought-out feature, @jorenham and for the real test of the design choice (zeroing slots so excluded symbols drop out of the percentage, not just the output). 

Plain pyrefly check no longer treating coverage-* codes as unused , a small but real correctness fix; nice that you covered the mixed-code case (bad-override still flagged, coverage-missing silently ignored).

The refactor in state/errors.rs (extracting Error::unused_ignore + Error::unused_pyrefly_ignore) is a net win  ~50 lines of repeated range-building boilerplate gone. And confirmed the new exclusion pass is safe under the parallel collect_one from #3825 since each closure owns its ModuleSymbols. LGTM and importing. Thanks for another contribution to Pyrefly!

maggiemoss:
@jorenham  can you share more about this motivation. From my understanding, this tool was initially designed to report issues in legacy code. My reading of this PR is that it would allow people to opt out. What is the problem this code is trying to solve?

jorenham:
The use-case I had in mind is projects that want to use `pyrefly coverage check` to enforce a certain level of type coverage, but are not willing or able to annotate certain functions or other typables, e.g. because doing doing so would cause issues at runtime (e.g. when using `torch` to JIT a function). In those situations, the only option right now is to exclude the entire module using `--project-excludes`. So the natural direct solution is to allow using `# pyrefly: ignore[coverage-missing]` to *only* ignore type-coverage metrics for that line, so that the rest of the module will still be measured.

jorenham:
> this tool was initially designed to report issues in legacy code.

That's one use-case I guess.

But another, IMO more important, class of use-cases is to actually enforce type-coverage in CI, preventing coverage backslides. NumPy uses it for this purpose, for example.

 I also found that it's able to find holes in typing stubs that stubtest currently doesn't report, which has helped me quite a bit in scipy-stubs.

jorenham:
rebased onto main to fix merge conflicts

NathanTempest:
Hey @jorenham sorry for putting this PR on the back burner but after review the team felt like this PR may not be a good direction for Pyrefly coverage to take at the moment. Closing this PR for now but we can reopen it after discussing / changes in the direction. Thank you for iterating over this.

jorenham:
Ok no worries.

PR #3921, opened 2026-06-23
collapse non-convergent class field types to Any (#3921)

yangdanny97:
Summary:

Unannotated attribute type inference can produce degenerate, deeply-nested
types for self-referential attributes -- e.g. `self.x = local` where `local`
was built from `self.x[i]`, yielding `list[list[list[...]]]`.

The SCC fixpoint
never converges for these: each iteration nests another container layer until
it hits MAX_ITERATIONS.

When a class-field SCC exceeds MAX_ITERATIONS without converging, commit `Any`
for its class-field members instead of the unrolled blowup.

Differential Revision: D109344076


stroxler (approved):

PR #3929, opened 2026-06-24
Fix Callable[..., ...] false negative (#3912)

### LINKED ISSUE #3912: false negative: `Callable[..., ...]`
```py

def f(x: Callable[..., ...]):
    reveal_type(x)
```

```
INFO sandbox.py:5:16-19: revealed type: (...) -> Ellipsis [[reveal-type](https://pyrefly.org/en/docs/error-kinds/#reveal-type)]
```

But `...` is not a type, so this should've been rejected.

sandbox: https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRhq9AMa4osJQwi50cQqmxLO3XvwDCYqPthy5mGGDpgAFPkR1zaqzADahPwBo6P0IAXQBKRDk6aIFhUQkpGBcwuRB-EDJBMChSQgZaKAoyBgALbQBaEUo4LXQ6AF46GRAAZkIARgAmZtSAX3TUDQgRADFoGAo0LDwiMhBeoA
x-ref: https://github.com/scipy/scipy/pull/25481#discussion_r3461146310

ibraheemshaikh5:
## Summary

When Callable[..., ...] is annotated, the second ellipsis sits in the return slot where ... is not a valid type, but pyrefly accepted it silently and revealed (...) -> Ellipsis. The return type was flowing through expr_untype as Type::Ellipsis, which nothing rejected.

This resolves that by checking for a literal ... in the Callable return argument inside specials.rs, emitting an error, and recovering as Unknown. The change is intentionally small and scoped to that slot only, with a regression test in callable.rs.

Fixes #3912 

## Test Plan

Added test_callable_ellipsis_return for the Callable[..., ...] case. Ran the callable and paramspec test suites plus test.py locally all green, no conformance changes.

### DIFF
--- a/pyrefly/lib/alt/specials.rs
+++ b/pyrefly/lib/alt/specials.rs
@@ -428,7 +428,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     self.error(
                         errors,
                         arguments[1].range(),
-                        ErrorKind::BadSpecialization,
+                        ErrorKind::InvalidAnnotation,
                         "`...` is not a valid return type".to_owned(),
                     );
                     self.heap.mk_any_error()
--- a/pyrefly/lib/alt/specials.rs
+++ b/pyrefly/lib/alt/specials.rs
@@ -424,11 +424,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     self.heap
                         .mk_type_of(self.heap.mk_callable_ellipsis(self.heap.mk_any_error()))
                 };
-                let ret = self.expr_untype(
-                    &arguments[1],
-                    TypeFormContext::TypeArgumentCallableReturn,
-                    errors,
-                );
+                let ret = if let Expr::EllipsisLiteral(_) = &arguments[1] {
+                    self.error(
+                        errors,
+                        arguments[1].range(),
+                        ErrorKind::BadSpecialization,
+                        "`...` is not a valid return type for `Callable`".to_owned(),
+                    );
+                    self.heap.mk_any_error()
+                } else {
+                    self.expr_untype(
+                        &arguments[1],
+                        TypeFormContext::TypeArgumentCallableReturn,
+                        errors,
+                    )
+                };
                 match &arguments[0] {
                     Expr::List(ExprList { elts, .. }) => {
                         match self.check_args_and_construct_tuple(elts, errors) {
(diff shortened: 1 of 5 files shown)


pyrefly/lib/alt/specials.rs
```diff
@@ -424,11 +424,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-                    &arguments[1],
-                    TypeFormContext::TypeArgumentCallableReturn,
-                    errors,
-                );
+                let ret = if let Expr::EllipsisLiteral(_) = &arguments[1] {
+                    self.error(
+                        errors,
+                        arguments[1].range(),
+                        ErrorKind::BadSpecialization,
+                        "`...` is not a valid return type for `Callable`".to_owned(),
```
jorenham: The way this is worded (the "for `Callable`" bit mostly) could be interpreted to mean that `...` can be used as return type in other places. But's it's not valid as a (return) type *in general*.

ibraheemshaikh5: Thank you, good catch. I just pushed a commit to tighten up the wording. TAL at what you mentioned in the issue thread now.

pyrefly/lib/alt/specials.rs
```diff
@@ -424,11 +424,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-                let ret = self.expr_untype(
-                    &arguments[1],
-                    TypeFormContext::TypeArgumentCallableReturn,
-                    errors,
-                );
+                let ret = if let Expr::EllipsisLiteral(_) = &arguments[1] {
+                    self.error(
+                        errors,
+                        arguments[1].range(),
+                        ErrorKind::BadSpecialization,
```
jorenham: For `f() -> ...` pyrefly currently uses `invalid-annotation` as error code, so maybe we should also use that one (`InvalidAnnotation`) here?

ibraheemshaikh5:
I extended the fix beyond the Callable return slot. Now the following cases are covered:

- list[...] / dict[int, ...] (type-argument positions) now report `...` cannot be used for type parameter with invalid-annotation, replacing the ParamSpec message
- class C[T: ...], type[...], and TypeVar(bound=...) report `Ellipsis` is not allowed in this context (invalid-annotation).
- The Callable return type error now uses InvalidAnnotation for consistency and accuracy

For testing: 
Added test_ellipsis_invalid_type_form in simple.rs covering the full list (including tuple[...]), and confirmed valid uses still pass: Callable[..., int], tuple[int, ...], ParamSpec(default=...), and ParamSpec specialization.

stroxler (approved):

PR #3935, opened 2026-06-24
Don't add unnecessary parentheses during inline variable extraction

### LINKED ISSUE #3882: Inline variable code action - inserts parentheses unnecessarily
### Describe the Bug

Given the following:

```python
foo = 1
value = foo + 2
```

Run the `Inline variable foo` code action on the foo variable.

**Expected behavior**
```python
value = 1 + 2
```

**Actual behavior**
```python
value = (1) + 2
```

This happens because of the unconditional wrapping of the variable in `()` here:

https://github.com/facebook/pyrefly/blob/99c080bec56be87b4c60a1c4964c0842eb09a1dc/pyrefly/lib/state/lsp/quick_fixes/inline_variable.rs#L85

This ensures that all replacements will be correct, at the cost of redundant parentheses.

---

Pyrefly version: `1.1.1`.
Editor: Zed `1.7.2`

### Sandbox Link

_No response_

_No response_

ak4-sh:
# Summary

Fix inline variable refactoring to only add parentheses when they are needed for the expression being inlined and the context where it is inserted.

- Parenthesize expressions that require grouping, such as binary operations, boolean operations, comparisons, tuples, lambdas, and similar expression forms.
- Avoid unnecessary parentheses for simple values like number literals in normal expression contexts.
- Add parentheses for number literals when inlining into attribute access, e.g. `value = 42; value.bit_length()` becomes `(42).bit_length()`.

Fixes #3882

# Test Plan
Added LSP code action tests for:
- Binary expression inlined into attribute access.
- Number literal inlined without unnecessary parentheses.
- Number literal inlined into attribute access.
- Boolean operation requiring parentheses.
- Tuple expression requiring parentheses.

### DIFF
--- a/pyrefly/lib/state/lsp/quick_fixes/inline_variable.rs
+++ b/pyrefly/lib/state/lsp/quick_fixes/inline_variable.rs
@@ -107,9 +107,10 @@ pub(crate) fn inline_variable_code_actions(
         let covering_nodes = Ast::locate_node(ast.as_ref(), range.start());
         let parent = covering_nodes.get(1);
 
-        let needs_parens_for_attribute_access =
-            matches!(parent, Some(AnyNodeRef::ExprAttribute(_)))
-                && matches!(value_expr, Expr::NumberLiteral(_));
+        let needs_parens_for_attribute_access = matches!(
+            parent,
+            Some(AnyNodeRef::ExprAttribute(_))
+        ) && matches!(value_expr, Expr::NumberLiteral(n) if matches!(n.value, ruff_python_ast::Number::Int(_)));
 
         let replacement = if needs_parens_for_attribute_access || always_needs_parens {
             format!("({value_text})")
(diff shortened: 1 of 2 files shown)


NathanTempest (approved):
Thanks for the fix, @ak4-sh ! 

One tiny optional follow-up (not blocking): NumberLiteral also matches floats/complex, but only integer literals have the 42. ambiguity
4.2.foo and 4j.foo are already valid Python, so they'll get a harmless-but-redundant (...). If you want to be minimal you could narrow that check to integer literals, but the current behavior is perfectly safe. Thanks again!

kinto0 (approved):

ndmitchell (approved):

ak4-sh:
Thanks for the comment @NathanTempest! I've updated the match condition, do let me know if it looks okay

NathanTempest (changes requested):
Good work! The Number::Int narrowing is exactly right, floats/complex no longer get redundant parens while decimal ints (42.bit_length()) are still handled. One small convention nit: we keep imports at the top rather than inline-qualified paths, so use ruff_python_ast::Number; + Number::Int(_) (matches how alt/expr.rs does it). Optionally, a quick negative test that 4.2.foo stays unparenthesized would lock in the new behavior. Thanks for iterating on this!

ak4-sh:
@NathanTempest Made the updates, sorry about this!

NathanTempest (approved):
Perfect thanks for addressing all the changes I requested! Importing and merging

PR #3940, opened 2026-06-25
Fix semantic token highlighting for union methods

### LINKED ISSUE #3873: union type lose method highlight
### Describe the Bug

Image

Image

but union type lose the highlight

Image

### Sandbox Link

_No response_

_No response_

xaskii:
# Summary

- We check each member of a union when choosing the semantic token for an attribute.
- This fixes cases like `x: A | B; x.foo()`, where foo should get method highlighting when both `A.foo` and `B.foo` are methods.
- Hover still has a similar issue: `x.foo` will show as an attribute even when every union member resolves foo to a method. That could be a follow-up?

Fixes https://github.com/facebook/pyrefly/issues/3873

# Test Plan

./test.py

### DIFF
--- a/pyrefly/lib/state/semantic_tokens.rs
+++ b/pyrefly/lib/state/semantic_tokens.rs
@@ -323,28 +323,46 @@ impl SemanticTokenBuilder {
         get_type_of_attribute: &dyn Fn(TextRange) -> Option<Type>,
         get_symbol_kind: &dyn Fn(&Key) -> Option<(ModuleName, SymbolKind)>,
     ) {
-        let kind = match get_type_of_attribute(attr.range()) {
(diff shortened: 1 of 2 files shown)


yangdanny97:
> Hover still has a similar issue: x.foo will show as an attribute even when every union member resolves foo to a method. That could be a follow-up?

Yeah, that could be a good followup

NathanTempest (approved):
Thanks @xaskii for the clean fix and nicely structured (failing test → refactor → fix). The union handling is exactly right: recursing on members and only committing to a token kind when all members agree, falling back to PROPERTY otherwise, is the honest behavior for mixed unions. Test reproduces #3873 precisely. LGTM and importing. 

And yes, the hover follow-up @yangdanny97 mentioned sounds like a good separate PR. Thanks for the contribution!

grievejia (approved):

Issue #3896, opened 2026-06-22
Difference in `slice` type annotations


MarcoGorelli:
### Describe the Bug

Small reproducer based on something observed in pandas-stubs

```py
from typing import reveal_type, TypeVar

T = TypeVar("T")


class Box:
    def __getitem__(self, arg: T) -> T:
        return arg


reveal_type(Box()[:])
```

```console
(scratch) mgorelli@marcoslaptop:~/scratch$ mypy t.py
t.py:11: note: Revealed type is "slice[None, None, None]"
Success: no issues found in 1 source file
(scratch) mgorelli@marcoslaptop:~/scratch$ pyright t.py
/home/mgorelli/scratch/t.py
  /home/mgorelli/scratch/t.py:11:13 - information: Type of "Box()[:]" is "slice[None, None, None]"
0 errors, 0 warnings, 1 information
(scratch) mgorelli@marcoslaptop:~/scratch$ ty check t.py
info[revealed-type]: Revealed type
  --> t.py:11:13
   |
11 | reveal_type(Box()[:])
   |             ^^^^^^^^ `slice[None, None, None]`
   |

Found 1 diagnostic
(scratch) mgorelli@marcoslaptop:~/scratch$ pyrefly check t.py
 INFO revealed type: slice[Any, Unknown, Unknown] [reveal-type]
  --> t.py:11:12
   |
11 | reveal_type(Box()[:])
   |            ----------
   |
 INFO 0 errors
```

### Sandbox Link

[...]

PR #3942, opened 2026-06-25
Preserve omitted slice positions and validate builtin sequence slices

### LINKED ISSUE #3768: new rule - zero step in slice
https://docs.astral.sh/ty/reference/rules/#zero-stepsize-in-slice

```py
l = list(range(10))
l[1:10:0]  # ValueError: slice step cannot be zero
```

(sorry for raising so many issues, i'm working on a rule comparison table between all the type checkers and opening issues for each rule missing between ty/pyrefly/basedpyright. i think this is the last one for pyrefly!)

ak4-sh:
Validate slice values for builtin sequences

Add tests to check for correct slice behaviour

Remove bug status from testcase since behaviour is fixed

# Summary

Validate slice components for builtin list and tuple subscripts. Pyrefly now reports errors for non-index slice values like decimal steps, and for literal zero steps. Also preserve omitted slice positions when inferring `slice[...]` types so start, stop, and step do not shift into the wrong generic slots.

Remove the `bug` marker from the existing zero-step slice test since the behaviour is now fixed, and add list/tuple coverage for decimal and zero step slices.

Fixes #3768, #3896

# Test Plan
- `cargo test test_slice_decimal_step`
- `cargo test test_slice_zero_step`

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -935,11 +935,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 self.heap.mk_unpack(ty)
             }
             Expr::Slice(x) => {
-                let elt_exprs = [x.lower.as_ref(), x.upper.as_ref(), x.step.as_ref()];
-                let elts = elt_exprs
-                    .iter()
-                    .filter_map(|e| e.map(|e| self.expr_infer(e, errors)))
-                    .collect::<Vec<_>>();
+                let none = self.heap.mk_none();
+                let elts = vec![
+                    x.lower
+                        .as_ref()
+                        .map_or_else(|| none.clone(), |e| self.expr_infer(e, errors)),
+                    x.upper
+                        .as_ref()
+                        .map_or_else(|| none.clone(), |e| self.expr_infer(e, errors)),
+                    x.step
+                        .as_ref()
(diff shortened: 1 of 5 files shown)


ak4-sh:
@yangdanny97 Hello! Not sure if this is still on your radar, but this PR addresses #3768  and #3896 . I noticed the issues were marked stale, so I wanted to check whether there’s anything else needed from my side before review.

NathanTempest:
Thanks @ak4-sh, and nice work tying these two issues together. The omitted-position fix is clearly right (slice[None, None, None] now matches mypy/pyright/ty, and the mypy_primer diff shows it only corrects the type args with no new errors), and the step/decimal validation is a genuine gap since list.__getitem__ takes a bare slice in typeshed.

One thing I'd like to sort before this lands: the validation loop infers each slice component (lower/upper/step) via expr_infer, and then the normal subscript resolution infers the same slice again (call_method_or_error(__getitem__, …) for list, infer_tuple_subscript for tuple). So the components get inferred twice, mostly harmless thanks to error de-dup, but it risks double-emitting on a component that itself errors, and it's redundant work. Could we infer each component once and validate off those types (or fold the check into the single inference)?

ak4-sh:
@NathanTempest Made the changes! Let me know if there are any issues and I can resolve them

connernilsen (approved):

PR #3947, opened 2026-06-25
Fix types/provide-type resolving the wrong token in notebook cells

qnox:
# Summary

`provide_type` passed the request position straight to `from_lsp_position(position, None)`, ignoring the notebook cell. For a cell other than the first, an LSP position is cell-relative, so dropping the cell offset resolves the position in the concatenated module at the wrong line - landing in an earlier cell. The IDE drives hover/type resolution (and the completion-trigger type lookup) through this method, so types in every cell after the first came back wrong or empty.

Thread the code-cell index (as the completion handler already does) through `provide_type` and translate the position with it so cell-relative positions map to the correct concatenated-module offset.

# Test Plan

Added a regression test `test_notebook_provide_type_second_cell` in `pyrefly/lib/test/lsp/lsp_interaction/notebook_provide_type.rs`: requests the type of `bar` in the second cell and asserts `typing.Literal[1]`. Fails on main (wrong cell resolved), passes with the fix.

- cargo test -p pyrefly test_notebook_provide_type

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -2853,7 +2853,8 @@ impl Server {
     ) -> Option<ProvideTypeResponse> {
         let uri = &params.text_document.uri;
         let handle = self.make_handle_if_enabled(uri, None).ok()?;
-        provide_type(transaction, &handle, params.positions)
+        let notebook_cell = self.maybe_get_code_cell_index(uri);
+        provide_type(transaction, &handle, params.positions, notebook_cell)
     }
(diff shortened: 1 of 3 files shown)


yangdanny97:
Thanks! Once the CLA is signed we can merge it

NathanTempest:
Hi @qnox thank you for the fix, it is a hard requirement to sign the CLA before contributing to Pyrefly. We are glad to have you aboard and for us to merge your PR after you sign this https://code.facebook.com/cla/individual

yangdanny97 (approved):

PR #3955, opened 2026-06-26
Fix: auto-discover typings/ on the CLI path, relative to the config root (#3955)

### LINKED ISSUE #3954: config/error difference between IDE/CLI
### Describe the Bug

Image

### Sandbox Link

_No response_

_No response_

kinto0:
Summary:

The `typings/` default was only applied inside `set_empty_to_default`, which the CLI bypasses whenever a Python interpreter is queried successfully: `get_env_from_interpreter` moves the real site-packages into `interpreter_site_package_path` and leaves `site_package_path` as `Some(vec![])`, so `set_empty_to_default` never runs and `./typings` is never added. The IDE sets `skip_interpreter_query = true`, so it hit the default and picked up `typings/`, producing different results for the same config. The `./typings` literal was also resolved against the process CWD rather than the config root, since it was injected after `rewrite_with_path_to_config`.

The fix centralizes `typings/` discovery in `ConfigFile::configure()`: after interpreter handling (so it applies on every branch — skipped, found, or failed), if the user did not explicitly set `site_package_path`, a `typings/` directory under the config root (`self.source.root()`) is appended. Interpreter site-packages live in the separate `interpreter_site_package_path` field, so `typings/` coexists with them. `set_empty_to_default` no longer injects the CWD-relative literal.

This flips `test_typings_autodiscovered_relative_to_config_root` to assert the correct behavior, demonstrating the test fails without this change.

fixes #3954

Reviewed By: grievejia

Differential Revision: D109879285

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -1208,6 +1208,10 @@ impl ConfigFile {
     pub fn configure(&mut self) -> Vec<ConfigError> {
         let mut configure_errors = Vec::new();
 
+        // Whether the user explicitly configured `site_package_path` (via config
+        // file or CLI flag). If not, we auto-discover a `typings/` directory below.
+        let site_package_path_set = self.python_environment.site_package_path.is_some();
+
         if self.interpreters.skip_interpreter_query {
             self.python_environment.set_empty_to_default();
         } else {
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #3956, opened 2026-06-27
Fix/issue 3897 constrained typevar overload

### LINKED ISSUE #3897: no-matching-overload false positive with a constrained type parameter
### Describe the Bug

Repro:

```
from typing import overload

class A: ...
class B: ...


def f[T: (A, B)](x: T) -> None:
  g(x)


@overload
def g(x: A) -> A: ...
@overload
def g(x: B) -> B: ...


def g(x: A | B) -> A | B: ...
```

Output:

```
 --> /tmp/foo.py:8:4
  |
8 |   g(x)
  |    ^^^
  |
  Possible overloads:
    (x: A) -> A [closest match]
    (x: B) -> B
  Argument `T` is not assignable to parameter `x` with type `A` in function `g`
```

Not exactly sure what's going on, but I suspect Pyrefly is treating `x` as morally equivalent `A | B`, which is why overload resolution fails?

### Sandbox Link

_No response_

_No response_

kz357:
# Summary

Fixes #3897

Expand constrained TypeVar arguments during overload resolution

Constrained TypeVar arg (eg x:T where T:(A,B)) was matched as a whole against each overload, which fails because T is assignable to a param only if every constraint is. Overload resolution already performs union-style argument expansion via expand_type, but it didn't have a case for a constrained TypeVar.

Expand a constrained TypeVar into its constraints, like a union ie each constraint is tried against the overloads and the matched returns are unioned. Existing all-expansions-must-match rule keeps this sound, if a constraint matches no overload, call is still rejected

# Test Plan

Added two regression tests to pyrefly/lib/test/overload.rs, both passing:

- test_overload_constrained_typevar_arg -- #3897 repro now resolves
- test_overload_constrained_typevar_arg_no_match -- uncovered constraint still errors

8 failing tests fail in main as well

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -50,6 +50,7 @@ use crate::types::callable::FuncMetadata;
 use crate::types::callable::Function;
 use crate::types::callable::Params;
 use crate::types::literal::Lit;
+use crate::types::type_var::Restriction;
 use crate::types::types::Type;
 use crate::types::types::Var;
 
@@ -239,6 +240,13 @@ impl<'a, Ans: LookupAnswer> ArgsExpander<'a, Ans> {
                     Vec::new()
                 }
             }
+
+            // a constraind typevar argument is one of its constraints, so expand like a union ie try each constraint against overloads + union the matched returns
+            Type::Quantified(q) => match q.restriction() {
+                Restriction::Constraints(constraints) => constraints.clone(),
+                _ => Vec::new(),
+            },
+
             _ => Vec::new(),
         }
     }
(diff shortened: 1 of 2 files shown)


yangdanny97:
Let's get a review from @rchen152 who understands overloads best

.github/workflows/label_new_issues.yml
yangdanny97, line 60: this is unrelated to the fix

rchen152 (changes requested):
Sorry for the delayed review! I was out of town for a few weeks. This looks good to me aside from the unrelated changes to label_new_issues.yml that Danny flagged. @kz357 Could you revert the unrelated changes and rebase this onto the current main? Then I'll import and merge it.

kz357:
Got it, dropped the unrelated `label_new_issues.yml` change. I think it should work unless I messed up git

kz357:
I think the commit that the bot imported has a formatting error, newer commit was run through `cargo fmt`

rchen152 (approved):

rchen152:
Thanks! The internal diff is landing now.

PR #3961, opened 2026-06-27
Support bare factory-boy factory calls returning the model type

### LINKED ISSUE #3934: factory_boy: calling a DjangoModelFactory directly (MyFactory(...)) is typed as the factory, not the produced model
### Summary

(Per #168's closure request to file specific Django gaps as their own issues.)

Pyrefly's [Django docs](https://pyrefly.org/en/docs/django/) list factory_boy support for `DjangoModelFactory` via `create()`, `build()`, `create_batch()`, `build_batch()` — and those work correctly. But factory_boy's idiomatic shorthand, **calling the factory directly** (`MyFactory(...)`, which factory_boy's metaclass makes equivalent to `MyFactory.create(...)`), is typed as an instance of the **factory** class rather than the produced model. Every attribute access on the result is then a false `missing-attribute`.

### Repro

`pyrefly.toml`:
```toml
project-includes = ["m.py"]
```

`m.py`:
```python
import factory
from django.db import models


class Thing(models.Model):
    name = models.CharField(max_length=10)


class ThingFactory(factory.django.DjangoModelFactory[Thing]):
    class Meta:
        model = Thing


def f() -> None:
    reveal_type(ThingFactory.create())   # Thing          (correct)
    reveal_type(ThingFactory.build())    # Thing          (correct)
    reveal_type(ThingFactory())          # ThingFactory   (should be Thing)

    obj = ThingFactory()
    reveal_type(obj.name)                # Error: Object of class `ThingFactory` has no attribute `name`
```

`pyrefly check` (v1.0.0):
```
revealed type: Thing
revealed type: Thing
revealed type: ThingFactory
ERROR Object of class `ThingFactory` has no attribute `name` [missing-attribute]
```

### Expected

`ThingFactory(...)` should be typed as `Thing`, identical to `ThingFactory.create(...)`. factory_boy's `FactoryMetaClass.__call__` dispatches the bare call to the factory's default strategy (create), so the produced object is a model instance.

### Impact

The bare-call form is the most common way tests instantiate factories. In one real codebase this single gap produces ~90 false `missing-attribute` errors (`obj = SomeFactory(...); obj.<field>`).

### Environment
- pyrefly 1.0.0 (also reproduced on 1.1.1)
- django-stubs 6.0.5, factory_boy
- Python 3.12

ak4-sh:
Calling a DjangoModelFactory subclass directly (`UserFactory()`) is equivalent to `UserFactory.create()` at runtime via FactoryMetaClass, but Pyrefly typed it as an instance of the factory rather than the model, producing false missing-attribute errors on every field access.

Add tests to confirim correct behaviour

Fixes #3934

# Test Plan
Add 2 tests:
- test_bare_call_attribute_access
- test_bare_call_returns_model

### DIFF
--- a/pyrefly/lib/alt/class/factory_boy.rs
+++ b/pyrefly/lib/alt/class/factory_boy.rs
@@ -27,9 +27,10 @@ const CREATE: Name = Name::new_static("create");
 const BUILD: Name = Name::new_static("build");
 const CREATE_BATCH: Name = Name::new_static("create_batch");
 const BUILD_BATCH: Name = Name::new_static("build_batch");
+const NEW: Name = Name::new_static("__new__");
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
-    /// Synthesize `create`, `build`, `create_batch`, and `build_batch` on
+    /// Synthesize `create`, `build`, `create_batch`, `new` and `build_batch` on
     /// factory-boy `DjangoModelFactory` subclasses, returning the model type
     /// from `class Meta: model = X`.
     pub fn get_factory_boy_synthesized_fields(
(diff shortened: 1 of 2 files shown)


ak4-sh:
I see that the internal tests have failed. Please do let me know if any changes are required on my end, I'd be happy to make edits! @samwgoldman

stroxler (approved):

PR #3963, opened 2026-06-28
Don't truncate inferred return types that are callables (#3958)

### LINKED ISSUE #3958: Type information lost when accessing method via `@property`
### Describe the Bug
Complex union types (`int | str | tuple[int, str] | None`) are simplified to `Unknown` when the method is accessed through `@property`
```python


class Foo:
    def bar(self, simple_union: int | str, complex_union: int | str | tuple[int, str] | None = None): ...


class FooManager:
    def __init__(self):
        self.foo = Foo()

    @property
    def bar(self):
        return self.foo.bar


reveal_type(FooManager().bar)  # revealed type: (self: Foo, simple_union: int | str, complex_union: Unknown = None) -> None

```

Python: 3.12

### Note
This only occurs when the union type has **4 or more** members. Smaller unions (2-3 types) are inferred correctly.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRy5AYyio4cOgDFcuRHLr66mGGDrZUlABRwYUMABo6cLsVjiArugi50iTun4APo4MlA6KtC4w%2BO6e3r6sgcGUdEEMbpEA2gkOcCEAuil0AHLeMHQAvMWlAJS%2BhPUK8ujKqupauACyGKhsMJS66AaGxnTi4qwQDGNWNmC1ekMG1raEYNoVmtoW1Y1DAALE1NJ8pAv6RiZmlstzA4sGgmmUgzer2oRXjXKCImKSzDALO0uugen1th9zNV9ABiATCUSwTCMAG%2BGa2XztHLOVweLw%2BPyJXKhOjhbiwaJ4uJ0ACq6AA1uhcAB3QaVEroGDQgC0AD4qpy5CA7CA3AxoHASORECA4TTxVBJqQ6GAPIpxd44I0LireDRUFN0G4aNhwfh4v4efziXcHjAnoMwDIQEVjab%2BnRgPgAL7OoUisiCMBQUiEBi0KAUOEABVIQZDjgwOAIpO8kDYbkoBvxhDkcIAyjAygALBgMYhwRAAeirgeMIcIvDYVZg6CrmFwijgVfC6HTmez3iruuSqCEqGgqGwsFTfYgGazGsGuGIS8lcjIDGL3m5IkoTm8G2dAGZCABGABMfvQIG9ItQ6ogIg00BgFDQWDwRDIt6AA

_No response_

AMR5210:
# Summary

Inferred return types are capped for union width and nesting depth so that
mutually-recursive functions converge during fixpoint solving. The cap is applied
by walking the entire type, so it also rewrote the signature of a returned
callable: any parameter union of four or more members became `Any`. This matches
the observation in #3958 that the problem only appears at four or more members.

The effect is visible through a property whose getter returns a bound method:

```python
class Foo:
    def bar(self, simple_union: int | str, complex_union: int | str | tuple[int, str] | None = None): ...


class FooManager:
    def __init__(self):
        self.foo = Foo()

    @property
    def bar(self):
        return self.foo.bar


reveal_type(FooManager().bar)
```

Before:

```
(simple_union: int | str, complex_union: Unknown = None) -> None
```

After:

```
(simple_union: int | str, complex_union: int | str | tuple[int, str] | None = None) -> None
```

# Approach

Inferred return types that are themselves callables are exempted from both caps.
Only `ClassType` and `Tuple` consume depth in the truncation walk, so a callable
signature cannot produce the unbounded growth the caps guard against, and
truncating it only discards annotations written in the source. Types inferred
inside the signature are still capped where they are solved, so recursive cases
converge as before.

# Scope

The exemption applies when the inferred return type is itself a callable. Wide
unions reached through an enclosing type, such as `Callable | None` or
`list[Callable]`, are still truncated. That is the case reported in #3958, and
this change is limited to it.

# Test Plan

- Added a regression test in `pyrefly/lib/test/attributes.rs` built from the
  snippet in #3958.
- `cargo test -p pyrefly --lib`: 7798 passed, 0 failed.
- Confirmed `__self__`, `__func__`, and `types.MethodType` assignability on a
  returned bound method are unchanged from `main`.

Fixes #3958

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -381,12 +381,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     x.attr.range,
                 );
                 let attr_type = self.attr_infer(&base, &x.attr.id, x.range, errors, None);
+                
                 if base.ty().is_literal_string() {
                     match attr_type.ty() {
                         Type::BoundMethod(method) => attr_type
                             .clone()
                             .with_ty(method.with_bound_object(base.ty().clone()).as_type()),
                         _ => attr_type,
+                    }
+                } else if let (Type::ClassDef(_), Type::Function(func)) = (base.ty(), attr_type.ty()) {
+                    if func.metadata.flags.property_metadata.is_some() {
+                        attr_type.with_ty(self.heap.mk_class_type(self.stdlib.property().clone()))
+                    } else {
+                        attr_type
                     }
                 } else {
                     attr_type
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -381,19 +381,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     x.attr.range,
                 );
                 let attr_type = self.attr_infer(&base, &x.attr.id, x.range, errors, None);
-                
                 if base.ty().is_literal_string() {
                     match attr_type.ty() {
                         Type::BoundMethod(method) => attr_type
                             .clone()
                             .with_ty(method.with_bound_object(base.ty().clone()).as_type()),
                         _ => attr_type,
-                    }
-                } else if let (Type::ClassDef(_), Type::Function(func)) = (base.ty(), attr_type.ty()) {
-                    if func.metadata.flags.property_metadata.is_some() {
-                        attr_type.with_ty(self.heap.mk_class_type(self.stdlib.property().clone()))
-                    } else {
-                        attr_type
                     }
                 } else {
                     attr_type
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -3594,6 +3594,12 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                             .collect(),
                     )
                 };
+                let return_ty = match return_ty {
+                    Type::BoundMethod(method) => self
+                        .bind_boundmethod(&method, &mut |a, b| self.is_subset_eq(a, b))
+                        .unwrap_or_else(|| self.heap.mk_bound_method(*method)),
+                    return_ty => return_ty,
+                };
                 // Cap inferred return unions aggressively. A small shared width
                 // budget keeps both top-level and nested inferred unions from
                 // accumulating many alternatives across recursive inference.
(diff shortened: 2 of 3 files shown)


WilliamK112:
Thanks for working on this. I tested this branch against the original #3958 snippet, where an instance property returns `self.foo.bar` and `reveal_type(FooManager().bar)` is inspected.

On this PR branch I still get:

```text
revealed type: (self: Foo, simple_union: int | str, complex_union: Unknown = None) -> None
```

Command used locally:

```text
cargo +stable-x86_64-pc-windows-gnu run -p pyrefly -- check --preset all --error reveal-type --output-format min-text case.py
```

So this change seems to address a different class-level descriptor case (`Foo.bar`), while the reported #3958 behavior is still present. A regression test using the issue body would help make sure this PR closes the original bug.

AMR5210:
Thanks, you were right. My earlier change handled a different class-level descriptor case (`Foo.bar`), not the original instance-property bug.

I updated the PR to add a regression test using the original #3958 snippet and fixed inference for property getters that return bound methods.

Before:
`(self: Foo, simple_union: int | str, complex_union: Unknown = None) -> None`

After:
`(simple_union: int | str, complex_union: int | str | tuple[int, str] | None = None) -> None`

I also removed the unrelated class-level descriptor change from this PR.

anci3ntr0ck:
@WilliamK112 Why hasn't this PR been imported yet? Is there anything else that needs to be modified?

stroxler:
@anci3ntr0ck could you rebase this?

We set up auto-import recently, and the flow is working better for code review (which is important because the backlog is a challenge for our team).

I *think* if you rebase, it should auto-create an imported diff which will make it easier to turn around the review

AMR5210:
Thanks for the earlier review, @WilliamK112. It pointed this at the right bug.

I have rebased onto current `main` and reduced the change to the actual root
cause, so a couple of things in my previous comment no longer describe the diff.

The cap on inferred return types is applied by walking the whole type, so it also
rewrote the signature of a returned callable and replaced any parameter union of
four or more members with `Any`. That is why the problem appeared only at four or
more members, as noted in the issue. The fix exempts inferred return types that
are themselves callables from the cap.

With this change the repro reports:

```
(simple_union: int | str, complex_union: int | str | tuple[int, str] | None = None) -> None
```

Two corrections to my previous comment:

- The `self: Foo` that disappeared from my earlier "after" output was caused by a
  second change I had included, which bound the returned method. That change was
  not needed for this bug and introduced regressions: `__self__` and `__func__`
  stopped resolving on a returned bound method, and such a value no longer
  satisfied a `types.MethodType` parameter. I have removed it. `self` is absent
  from the output above because of a separate display change already on `main`.
- The regression test now uses the issue snippet unmodified.

`cargo test -p pyrefly --lib` passes with 7798 tests. The remaining truncation
behaviour for unions reached through an enclosing type, such as `Callable | None`
or `list[Callable]`, is unchanged and outside the scope of this issue.

stroxler (approved):

PR #3965, opened 2026-06-28
Share `Overload` signatures behind `Arc` to avoid deep-copying on clone

alexander-beedie:
# Summary

* Profiling[^1] showed a lot of `Type` value churn; one of the largest drivers looked like cloning `Overload`, which deep-copies its `Vec1<OverloadType>` signature list. This happens a _lot_ on read-only paths where nothing is mutated and the copy isn't necessary.

* The optimisation is just to wrap `Overload.signatures` with `Arc` (much like`TArgs` does). The top contributor (`OverloadType::to_vec`, ~23% of clone samples) then disappears from profiling.

* Included a dedicated `profiling` entry in the Cargo.toml (basically _"inherit release but keep symbols"_) for convenience, assuming anyone else is doing dedicated profiling builds - can remove if unwanted?

I have a few other possible optimisations in mind, but want to keep any PRs single-focused/tight 👍

# Test Plan

Ran the test suite; all tests show _identical_ before/after results.
(Expected, as no type-checking logic changed).
 
# Speedup

`Overload`-heavy code is what benefits; `numpy` (5.7% faster) and `pydantic` (~2.7% faster) stand out. Showing the single-threaded `j1` speedup here, for clarity (`mt` scores are noisier as the gain is small):

| Package      |  LOC | old (secs)     | new (secs)     | Δ%        |
|:-------------|-----:|---------------:|---------------:|----------:|
| numpy        | 313k | 20.14 ± 0.06   | 19.00 ± 0.28   | **−5.7%** |
| pydantic     |  46k | 0.444 ± 0.006  | 0.432 ± 0.004  | **−2.7%** |
| pandas       | 658k | 5.40 ± 0.05    | 5.31 ± 0.11    | **−1.8%** |
| scipy        | 553k | 3.30 ± 0.02    | 3.25 ± 0.02    | **−1.5%** |
| scikit-learn | 381k | 2.16 ± 0.02    | 2.13 ± 0.01    | **−1.4%** |

[^1]: On Apple Silicon M3 Max, with `--threads 1` to help minimise noise.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,11 @@ debug = 1
 debug = true
 inherits = "dev"
 
+[profile.profiling]
+inherits = "release"
+debug = 2
+strip = "none"
+
 [workspace]
 members = [
   "crates/pyrefly_build",
--- a/crates/pyrefly_types/src/callable_residual.rs
+++ b/crates/pyrefly_types/src/callable_residual.rs
@@ -288,7 +288,7 @@ impl Type {
             .cloned()
             .map(|branch_ty| branch_ty.into_overload_signature(&metadata))
             .collect::<Option<Vec<_>>>()?;
-        let signatures = Vec1::try_from_vec(signatures).ok()?;
+        let signatures = Arc::new(Vec1::try_from_vec(signatures).ok()?);
         Some(Type::Overload(Overload {
             signatures,
             metadata: Box::new(metadata),
--- a/crates/pyrefly_types/src/display.rs
+++ b/crates/pyrefly_types/src/display.rs
@@ -2379,13 +2379,13 @@ pub mod tests {
         };
 
         let overload = Type::Overload(Overload {
-            signatures: vec1![
+            signatures: Arc::new(vec1![
                 OverloadType::Function(sig1.clone()),
                 OverloadType::Forall(Forall {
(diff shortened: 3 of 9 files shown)


yangdanny97:
@samwgoldman

samwgoldman:
Hey @alexander-beedie! Sorry I missed this.

This is a really nice win. I've found similar bottlenecks through profiling myself. My hesitation here is that we've found contention from atomic refcount operations can in practice really hurt performance in highly parallel scenarios. Internally, we run Pyrefly with 64 parallel checking threads where this kind of contention shows up. For example, I have some WIP that removes the Arc<ClassInner> that gives a ~30% CPU time win.

I want to approach this perf opportunity from another angle, where we are able to increase sharing without the refcount overhead. This is an awesome change, but for the moment I want to keep working toward removing Arc from our type representation instead of adding more.

Thanks for the contribution! I'd be happy to chat more about Pyrefly perf. Maybe we can even work together on it? If you have time, would love to chat more on Discord.

alexander-beedie:
> I'd be happy to chat more about Pyrefly perf. Maybe we can even work together on it? If you have time, would love to chat more on Discord.

@samwgoldman sure; you can find me on Discord at `alexander.beedie` :)

PR #3966, opened 2026-06-29
Warn on unrecognized config keys with a "did you mean" suggestion

### LINKED ISSUE #3943: Feature request: Warn on unrecognized config keys
### Describe the Bug

I copied some config over from mypy into my `pyproject.toml` and wrote `python_version` instead of the `python-version` that pyrefly expects. The key was silently ignored, so I spent a while confused about why my Python version setting had no effect. A warning would have caught it immediately.

I think this would also make pyrefly more internally consistent: an unknown *error kind* already fails to parse (e.g. `unknown variant 'unreachable-match-case'`), but an unknown *top-level key* is silently dropped. It looks like the top-level config struct doesn't reject unknown fields the way the error-kinds enum does.

I'd suggest a warning rather than a hard error — that catches typos without breaking configs shared across pyrefly versions (e.g. a newer key being read by an older binary).

### Sandbox Link

_No response_

_No response_

markselby9:
Fixes #3943.

When a `pyrefly.toml` or `[tool.pyrefly]` has a key pyrefly doesn't recognize, it warns per unrecognized key with a "did you mean" suggestion when a known option is close:

```
Unknown config key `check-unannotated-def`. Did you mean `check-unannotated-defs`?
Unknown config key `made-up-key`
```

Detection is serde-derived and can't drift: anything unrecognized lands in the flattened `extras` catch-all. The "did you mean" candidate set (canonical keys and accepted deprecated aliases) is generated from the config structs by a `ConfigKeys` derive that reads each struct's serde attributes (rename_all, rename, alias, skip) and splices in the flattened sub-structs, so adding a field or alias updates the suggestions automatically with nothing to keep in sync by hand.

Scoped by section: top-level keys match the project options plus the flattened interpreter/environment/base keys; `[coverage]` against the coverage options; `[[sub-config]]` against `matches` plus the base keys.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2315,6 +2315,7 @@ dependencies = [
  "pretty_assertions",
  "pulldown-cmark",
  "pyrefly_build",
+ "pyrefly_derive",
  "pyrefly_python",
  "pyrefly_util",
  "regex",
--- a/crates/pyrefly_config/Cargo.toml
+++ b/crates/pyrefly_config/Cargo.toml
@@ -19,6 +19,7 @@ enum-iterator = "2.3.0"
 itertools = "0.15.0"
 parse-display = "0.8.2"
 pyrefly_build = { path = "../pyrefly_build" }
+pyrefly_derive = { path = "../pyrefly_derive" }
 pyrefly_python = { path = "../pyrefly_python" }
 pyrefly_util = { path = "../pyrefly_util" }
 regex = "1.13.1"
--- a/crates/pyrefly_config/src/base.rs
+++ b/crates/pyrefly_config/src/base.rs
@@ -10,6 +10,7 @@ use std::collections::HashMap;
 use clap::ValueEnum;
 use enum_iterator::Sequence;
 use enum_iterator::all;
+use pyrefly_derive::ConfigKeys;
 use pyrefly_python::ignore::Tool;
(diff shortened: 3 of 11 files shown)


markselby9:
Thanks for reopening, @yangdanny97 — rebased onto current `main`. The rebase wasn't purely mechanical, so flagging what changed:

- **`[coverage]` scope.** `main` grew a per-table warning for unknown `[coverage]` keys since I opened this. Rather than drop it, I extended `unknown_key_warnings` to cover it too, so `[coverage]` typos now get the same per-key "did you mean" treatment (candidates: `includes`, `excludes`).
- **New config options.** The `test_known_config_keys_cover_serialized_fields` sync guard caught `strict-partial-subtyping` and `legacy-overload-expansion` (added since) missing from the candidate lists — added both, plus `coverage` as a top-level key.
- Sync guard now also covers `CoverageConfig`.

Full `pyrefly` + `pyrefly_config` + `pyrefly_util` suites and `test.py` lint are green.

yangdanny97 (changes requested):

stroxler (changes requested):
I think this requires a rebase - there are merge conflicts, and I see compile errors on the internal copy. I'm requesting changes just to manage my PR review queue better

PR #3968, opened 2026-06-29
Report duplicate TypeVar definitions

### LINKED ISSUE #3754: new rule - `TypeVar` redefinition
```py
from typing import TypeVar

T = TypeVar("T")
T = TypeVar("T") # should be an error
```

binitpoks-code:
now this PR adds a focused binder-time check for duplicate legacy `TypeVar` definitions. Previously, Pyrefly accepted duplicate definitions like this without reporting an error:

```python
from typing import TypeVar

T = TypeVar("T")
T = TypeVar("T")
```

After this change, the second definition reports a redefinition error.

The change is intentionally kept small and limited to the existing legacy `TypeVar` binding path. I also added a regression test in `generic_legacy.rs` to cover this case.

Fixes #3754

# Test Plan

* `cargo test test_tvar_duplicate_definition`
* `cargo test generic_legacy`

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -1893,6 +1893,19 @@ impl<'a> BindingsBuilder<'a> {
         }
     }
 
+    pub fn check_for_type_var_redefinition(&self, name: &Name, range: TextRange) {
+        let prev_idx = self.scopes.current_flow_idx(name);
+        if let Some(prev_idx) = prev_idx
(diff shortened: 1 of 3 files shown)


yangdanny97:
@NathanTempest if the primer delta looks good we can merge this one

yangdanny97:
Thanks for your contribution here!

given the further discussion on the original issue I'll close this PR for now, but if we decide to move forward again I'll reopen and merge

PR #3972, opened 2026-06-29
fix hover on case _ show the remaining cases #3334

### LINKED ISSUE #3334: hover on case _ show the remaining cases
### Describe the Bug

```py
from enum import StrEnum

class E(StrEnum):
    x = "1"
    y = "2"
def foo(x: E):
    match x:
        case E.x:
            ...
        case _:
            ...
 ```

hover on `_` show match `E.y`

Same as subclass case, I guess.

### Sandbox Link

_No response_

_No response_

### PR DESCRIPTION
**Pull request by asukaminato0721**
# Summary

Fixes #3334

hovering case _ resolves the enclosing match case’s Key::PatternNarrow type, falling back to the subject type when there is no prior narrowing.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -751,38 +751,6 @@ fn in_keyword_in_iteration_at(ast: Option<&ModModule>, position: TextSize) -> Op
     None
 }
 
-fn match_wildcard_type_at(
-    transaction: &Transaction<'_>,
-    handle: &Handle,
-    position: TextSize,
-) -> Option<Type> {
-    let module = transaction.get_ast(handle)?;
-    let covering_nodes = Ast::locate_node(&module, position);
-    let is_wildcard = covering_nodes
-        .iter()
-        .any(|node| matches!(node, AnyNodeRef::PatternMatchAs(pattern) if pattern.name.is_none() && pattern.pattern.is_none()));
-    if !is_wildcard {
-        return None;
-    }
-    let case_range = covering_nodes.iter().find_map(|node| match node {
-        AnyNodeRef::MatchCase(case) => Some(case.range),
-        _ => None,
-    })?;
-    let subject_start = covering_nodes.iter().find_map(|node| match node {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pyrefly/lib/lsp/wasm/hover.rs [hunk 1]
```diff
@@ -698,8 +730,8 @@ pub fn get_hover(
         });
     }
 
-    let mut type_ = transaction
-        .subscript_operator_type_at(handle, position)
+    let mut type_ = match_wildcard_type_at(transaction, handle, position)
```
kinto0: is there a reason we don't include this in existing type APIs like type_from_expression_at_impl?

kinto0 (changes requested):
really cool. one question about placement

pyrefly/lib/lsp/wasm/hover.rs
asukaminato0721: `type_from_expression_at_impl` is expression-only, and case _ is a pattern node.

moved the wildcard-pattern handling into the shared `get_type_at*` path.

yangdanny97 (approved):

pyrefly/lib/state/lsp.rs
```diff
@@ -867,6 +867,39 @@ impl<'a> Transaction<'a> {
...
+        })?;
+        let key = Key::PatternNarrow(case_range);
+        if self
+            .get_bindings(handle)
+            .is_some_and(|bindings| bindings.is_valid_key(&key))
+        {
+            self.get_type_for_surface(handle, &key, for_display)
+        } else {
+            self.get_type_at_impl(handle, subject_start, for_display)
+        }
```
kinto0: When there is no PatternNarrow key, the wildcard hover falls back to get_type_at_impl at stmt_match.subject.range().start(). For match subjects whose type is not represented by the token at the start (for example match obj.attr: or match xs[i]:), this resolves the leading identifier (obj/xs) rather than the full subject expression, so hovering case _ in an unnarrowed/default case can display the container/base object's type instead of the matched value's type. The fallback should use the subject expression's trace/range rather than only its start position.

kinto0: will resolve in follow-up if I get to it first

PR #3975, opened 2026-06-29
Fix overloads with a generic `self:` dropped when matching a `Protocol`

### LINKED ISSUE #3974: 1.1 regression: false positive `bad-assignment` for overload with generic `self` type
### Describe the Bug

After bumping Pyrefly from `1.0` to `1.1` in [optype](https://github.com/jorenham/optype), a bunch of type-tests started failing for the `numpy<2.2` jobs (see e.g. https://github.com/jorenham/optype/actions/runs/27886459886/job/82522566360). This also broadly affects scipy-stubs, which heavily depends on these relevant optype types. But I'm guessing that this also causes issues for other projects that support these older NumPy versions.

When an overloaded method has an overload whose `self` is annotated with a `TypeVar`, Pyrefly 1.1 **drops that overload** when the method is bound on an instance. With it gone, the type no longer satisfies any `Protocol` that structurally needs it:

```py
from typing import Any, Protocol, Self, overload

class C1:
    @overload
    def f[S](self, /) -> Self: ...
    @overload
    def f(self, x: int, /) -> int: ...

class C2:
    @overload
    def f[S](self: S, /) -> S: ...
    @overload
    def f(self, x: int, /) -> int: ...

class P(Protocol):
    def f(self, /) -> object: ...

x: P = C1()  # ✔️ 
x: P = C2()  # ❌ `C2` is not assignable to `P` [bad-assignment]
```

The error message shows that first overload disappeared:

```
ERROR sandbox.pyi:19:8-12: `C2` is not assignable to `P` [[bad-assignment](https://pyrefly.org/en/docs/error-kinds/#bad-assignment)]
  `C2.f` has type `Overload[
  (self: C2, x: int, /) -> int
]`, which is not assignable to `(self: C2, /) -> object`, the type of `P.f`
```

And indeed, without that overload, `C2` would be assignable to `P`.

This doesn't apply to non-overloaded methods, so e.g. `def f[S](self: S) -> S` is fine.
It also seems to be order-independent, so swapping the overloads doesn't change anything.

I verified that this is still the case on `main`.

Oh and the reason this only affects `numpy<2.2` is because, starting from 2.2, `numpy.generic.__array__` uses `Self` instead of a self-bound `TypeVar`.

Related (but verified to be different) issues:
- https://github.com/facebook/pyrefly/issues/3878
- https://github.com/facebook/pyrefly/issues/3903

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIIANCGQE4xhSmEAuuAtlBQMQAEABVIMmpPmix58fAMa50kAOYBXOqlYR5hADrp%2BAZRgw%2BAC1aticRAHpr9Rs0K46i6zHTXMuGXGtyFECpqGvLWfGDOfKgAbqjQqNiwsvJKquqa6Hy4xCHocDroZKwm8gC00TB0cBl8ALx82iAAzIQAjABMjboA2pV0znAAurrK6BAcxM6sMJilmBAMMhoVdXwA5IHozjBruluspQwAjsoLM6UA1jCkpagyMvBwq2sA7qh06Lvo1CASOAQkUgQChgfocPisUjECDoRR8caTOisPgAQXQpCogn67DkUExhigYExuAqdCguFQmF0uhkUFQcCeAGFWohdHx2XwAAIkyrkylsjmYRjhbr6QYACjgMEJmOsAEo%2BKUAHx8AlgRB8QhagXs7mkvlUzKC4VgSXSol8fAamGsWUK5Xw9CsDVago0ukMviM9qso26nlkimGjl8IVgEVis2EjX6O2KlX6F3av1cgMGnWhk1Ri1Wx22vjy%2BN5pNu9C0%2BlPATigTY7y4KBy30hsPhbNxh24bAAKxgSxL1PQuYEq2Z4rluiHI-aY90IAAvjQ7ssYAAxaAwCh-KSA4FzoA

_No response_

jorenham:
# Summary

`filter_overloads_by_self_type` dropped any overload whose `self:` mentions the overload's own type params, since `is_subset_eq(receiver, self)` checks against a rigid, unsolvable variable. Substitute those params with `Any` first, so the receiver is matched against a gradual `self:`.
An overload whose `self:` mentions its own type parameters is matched gradually by substituting those parameters with `Any (self: Arr[S, T] -> Arr[Any, Any])`, which is exact for the realistic single-occurrence case and only over-accepts the presumably niche degenerate repeated-parameter case, `(self: C[Z, Z])`, same as mypy and pyright.

Fixes #3974

# Test Plan

Regression tests added

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2840,9 +2840,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             .signatures
             .iter()
             .filter(|sig| {
-                let func = match sig {
-                    OverloadType::Function(f) => f,
-                    OverloadType::Forall(forall) => &forall.body,
+                let (func, tparams) = match sig {
+                    OverloadType::Function(f) => (f, None),
+                    OverloadType::Forall(forall) => (&forall.body, Some(&forall.tparams)),
                 };
                 // Only instance methods have a `self` first parameter; static and
                 // class methods' first parameter is a regular argument.
@@ -2850,6 +2850,16 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     return true;
                 }
                 func.signature.get_first_param().is_none_or(|p| {
+                    // Replace the overload's own type params in `self:` with `Any`
+                    // (e.g. `self: Array[S, T]` -> `Array[Any, Any]`), matching against a gradual
+                    // `self:` rather than a rigid, unsolvable variable.
+                    let p = match tparams {
+                        Some(tparams) => {
+                            let any = self.heap.mk_any_implicit();
+                            p.subst(&tparams.iter().map(|q| (q, &any)).collect())
+                        }
+                        None => p,
+                    };
                     // A non-protocol `self:` can't re-enter protocol conformance, so check
                     // it directly. A protocol-typed `self:`, however, makes
                     // `is_subset_eq(self_type, p)` re-enter this same filtering on the same
--- a/pyrefly/lib/test/protocol.rs
+++ b/pyrefly/lib/test/protocol.rs
@@ -1126,3 +1126,64 @@ def f(x: HasCallNoExtra[int, str]) -> Lens[int, str]:
     return x  # E: not assignable to declared return type
 "#,
 );
+
+testcase!(
+    test_protocol_overload_typevar_self_kept,
+    r#"
+from typing import Protocol, overload
+
+class C:
+    @overload
+    def f[S](self: S, x: int) -> S: ...
+    @overload
+    def f(self, x: str) -> str: ...
+    def f(self, x: int | str) -> object: ...
+
+class P(Protocol):
+    def f(self, x: int) -> object: ...
+
+p: P = C()  # the `self: S` overload binds S=C, satisfying P
+"#,
+);
+
+testcase!(
+    test_protocol_overload_parameterized_self_kept,
+    r#"
+from typing import Protocol, overload
+
+class Arr[S, T]:
+    @overload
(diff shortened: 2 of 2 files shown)


jorenham:
- jax ([src](https://github.com/jax-ml/jax/blob/e50b7439e7c6fee5ad74864b59401522c35f4ddf/jax/_src/lax/linalg.py#L1066-L1068)): two new *true positives* are now reported (previously false negatives). `list(ndarray)` now infers the precise scalar element type (via `ndarray.__iter__`'s generic-self overload, previously dropped), correctly flagging a `list[np.int32]` passed where `Sequence[int]` is expected.
- scipy-stubs ([src](https://github.com/scipy/scipy-stubs/blob/b0ad383e5f1fa6581b89747f35f4ed5cc55c61bf/tests/sparse/test_csr.pyi#L196)): this was indeed a false positive that is now resolved.
- DateType ([src](https://github.com/glyph/DateType/blob/a6ebb954cd18302a031a29b2f65e077b8e7776d4/src/datetype/test/test_datetype.py#L49)): also looks like a resovled false positive that (they seem to be using a lot of `self: Self` in overloads, and there's a protocol involved here).
- colour: all that changed here are the reported error messages; so +0

So all primer changes are good news.

NathanTempest (approved):
Approving and thanks, @jorenham for the clean fix.

filter_overloads_by_self_type was matching the receiver against a self (still-unbound type params) so is_subset_eq compared against rigid unsolvable variables and silently dropped overloads that should apply. Substituting just those params with implicit Any before the applicability check is the right direction here.

Only the overload's own tparams are gradualized but the concrete parts of the self: are preserved The Any-substituted type is only used for the keep/drop decision; the surviving overload keeps its real self: Arr[S2, T2] -> Arr[S2, T2], which is then solved precisely against the receiver. So this increases precision rather than loosening it.

stroxler (approved):

PR #3978, opened 2026-06-29
Build shape-extensions package in GitHub CI (#3978)

stroxler:
Summary:

Add the first build-only stub-package workflow for `pyrefly-shape-extensions` so the package metadata and smoke-test commands can be exercised in GitHub Actions before any release workflow or publishing integration is introduced. The workflow stamps the package version from `version.bzl`, builds wheel and sdist artifacts from a temporary package copy, verifies both install in fresh virtualenvs, checks the expected artifact contents, and uploads `shape-ext-dist` for later torch-stubs integration.

Reviewed By: grievejia

Differential Revision: D109859966

### DIFF
--- a/.github/workflows/build_shape_stubs.yml
+++ b/.github/workflows/build_shape_stubs.yml
@@ -1,0 +1,105 @@
+name: Build shape stubs
+
+on:
+  workflow_dispatch:
+  workflow_call:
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #3980, opened 2026-06-29
Build torch stubs package in GitHub CI

stroxler:
Summary: Extend the shape-stubs build workflow so `pyrefly-torch-stubs` is built, smoke-installed, and archived alongside `pyrefly-shape-extensions`. The shared `shape-stubs-dist` artifact remains the publish boundary, so the existing publish workflow can upload every package distribution produced by the build job without adding package-specific release logic.

Differential Revision: D110066967


grievejia (approved):

PR #3981, opened 2026-06-29
Fix duplicated-word typos in code comments

durvesh1992:
## Summary
Four code comments contained accidental duplicated words that hurt readability. This corrects them. Comment-only change — no behavioral impact.

| File | Before | After |
|------|--------|-------|
| `pyrefly/lib/binding/scope.rs` | `...if we know that the the name is` | `...the name is` |
| `pyrefly/lib/binding/scope.rs` | `...the branch_idx, which will will be` | `...which will be` |
| `pyrefly/lib/state/state.rs` | `We check the the epoch optimistically` | `We check the epoch optimistically` |
| `pyrefly/lib/query.rs` | `...or it is is defined in \`builtins\`` | `...or it is defined in \`builtins\`` |

### DIFF
--- a/pyrefly/lib/binding/scope.rs
+++ b/pyrefly/lib/binding/scope.rs
@@ -3125,7 +3125,7 @@ impl Scopes {
                 } else {
                     flow_info.initialized()
                 };
-                // Because class body scopes are dynamic, if we know that the the name is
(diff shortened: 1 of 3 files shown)


stroxler (approved):

PR #3982, opened 2026-06-29
Try reflected binary-op dunder first for proper subclasses (#3876)

### LINKED ISSUE #3876: Type narrowing not done correctly for bitwise operations with `int` and instance of subclass of `enum.IntFlag`
### Describe the Bug

When &-ing an `int` and a member of an `enum.IntFlag`, another member of the flags is returned, but Pyrefly reveals it as an `int`. I believe this isn't an issue with typeshed, since mypy [has the correct behaviour](https://gist.github.com/mypy-play/0c29d5580d780221622b7d8af6221996).

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEM6ArvRDcbpQC52pNe4ANHQCS6LgDEoqAOYAddAGNpcOHSmy4ACjGTpMgJSIFdU3QCCdALy9%2BuLQZNmAQtdsCHCyjABuMVFAA%2BlykxDBaEOIOdABk6vpwhOYGIIIgZN5gUKSEAjRQFADEdAAKpJnZdGhYePh0irjokDJMlKhcEI2ECsUAyjAwdAAWXFzEcIgA9JMZMFk5nDKTjJOYuIpwkw1NEC1tHY2TdGCcvD6o0KjYsPWNza3tneh0uMQH6IkKZFxDjQC0fkocCebjkIAAzIQAIwAJjBCgA2jBKNQgQBdBRMdBsDjcGCYP6YCDeRQdPxuADku3QnBgFIUNK4f28AEcmMT8X8ANYwUh-VCKRTwNQ2CkAd1QlHQ9PQIAAvmkBWSYBJoDAKNUcAQSOQ5UA

_No response_

durvesh1992:
## Summary
Fixes #3876.

Per Python's data model, when the **right** operand's type is a *proper subclass* of the **left** operand's type, the **reflected** dunder is tried first. Pyrefly always tried the forward dunder first, so for example:

```python
from enum import IntFlag
class Color(IntFlag):
    RED = 1
    GREEN = 2

def f(x: int, c: Color):
    reveal_type(x & c)  # was: int   — should be: Color
```

`int & Color` resolved through `int.__and__` (which widens the result back to `int`) instead of `Color.__rand__` (which returns the flag type `Self`). At runtime `int_val & color` calls `Color.__rand__` and yields a `Color`.

## Fix
In `binop_types`' `binop_call` (`pyrefly/lib/alt/operators.rs`), when the RHS type is a proper subclass of the LHS type, the reflected dunder is now tried first.

This is intentionally narrow:
- It only reorders when `rhs` is a **strict** subclass of `lhs` (`has_superclass` + not equal).
- A subclass that does **not** override the reflected dunder inherits it unchanged, and `try_binop_calls` still falls back to the forward dunder if the reflected call doesn't apply — so non-overriding subclasses are unaffected.

## Test
Added `test_reflected_dunder_subclass_priority` in `pyrefly/lib/test/operators.rs` asserting `x & c`, `x | c`, `x ^ c` are `Color` for `x: int`, `c: Color`. It fails before this change (inferred `int`) and passes after.

### DIFF
--- a/pyrefly/lib/alt/operators.rs
+++ b/pyrefly/lib/alt/operators.rs
@@ -483,10 +483,30 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             // Reflected operator implementation: This deviates from the runtime semantics by calling the reflected dunder if the regular dunder call errors.
             // At runtime, the reflected dunder is called only if the regular dunder method doesn't exist or if it returns NotImplemented.
             // This deviation is necessary, given that the typeshed stubs don't record when NotImplemented is returned
-            let calls_to_try = [
-                (&Name::new_static(op.dunder()), lhs, rhs),
-                (&Name::new_static(op.reflected_dunder()), rhs, lhs),
-            ];
+            let forward = Name::new_static(op.dunder());
+            let reflected = Name::new_static(op.reflected_dunder());
+            let forward_call = (&forward, lhs, rhs);
+            let reflected_call = (&reflected, rhs, lhs);
+            // Python data model: when the right operand's type is a *proper* subclass of the
+            // left operand's type, the reflected dunder is tried first. This lets e.g.
+            // `int_val & some_IntFlag_member` resolve through `IntFlag.__rand__` (which returns
+            // the flag type) rather than `int.__and__` (which widens back to `int`). A subclass
+            // that does not override the reflected dunder inherits it unchanged, and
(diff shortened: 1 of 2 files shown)


rchen152 (approved):
Sorry for the long wait. This looks good to me, thank you!

yangdanny97 (approved):

PR #3944, opened 2026-06-25
Support typed ignore-errors directives

WilliamK112:
## Summary
- Support `# pyrefly: ignore-errors[code]` as a file-level suppression for selected Pyrefly error codes.
- Reuse the existing `Suppression` representation for file-level directives so whole-file and line-level Pyrefly code matching share the same semantics, including parent suppression names.
- Keep non-Pyrefly `ignore-errors` directives as blanket-only and document the new Pyrefly form.

Fixes #3723

## Notes
This is scoped to file-level code filtering for `pyrefly: ignore-errors`. Related suppression lint/reporting PRs (#3832 and #3797) address separate diagnostics and unused-ignore behavior.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p pyrefly_python test_parse_ignore_all`
- `CARGO_TARGET_DIR=/Users/William/pyrefly-target-codex cargo test -p pyrefly test_pyrefly_top_level_ignore_typed`

PR #3983, opened 2026-06-30
Allow whole file error suppressions for specific kinds.

nascheme:
# Summary

This allows the following comments at the beginning of files:
```
# pyrefly: ignore-errors[kind, ...]
```

Multiple kinds can be comma-separated on one line, or split across several `ignore-errors` lines; the lists accumulate.

My use case for this is that I'm gradually increasing pyrefly's strictness.  I want to turn on errors like this:
```
implicit-any-parameter = "error"       # unannotated function parameters
implicit-any-type-argument = "error"   # bare `list` / `dict` (no type args)
implicit-any-empty-container = "error" # `x = []` with no annotation
```
However, I have many "legacy" Python files that don't conform to this level of strictness.  Rather than have many "sub_config" sections or add individual per-line suppressions, I want to suppress these kinds of errors for whole files.

# Test Plan

The feature is covered by integration tests in `pyrefly/lib/test/suppression.rs`, using the `testcase!` macro. Each test is a snippet of Python source with inline error expectations: a `# E: <message>` comment marks a line where an error is expected, and any line without such a comment must produce no error. The macro type-checks the snippet and asserts that emitted errors match the annotations exactly.

Three new tests exercise the `# pyrefly: ignore-errors[kind, ...]` syntax:

- test_pyrefly_top_level_ignore_specific_kind — suppresses a single kind (bad-assignment) and verifies that an unlisted kind (bad-return) still reports. The bad-assignment line has no # E:, while the return "oops" line carries a # E: annotation, proving suppression is kind-scoped rather than whole-file.
- test_pyrefly_top_level_ignore_specific_kinds_one_line — two kinds comma-separated in one comment ([bad-assignment, bad-return]); both errors are suppressed (no annotations).
- test_pyrefly_top_level_ignore_specific_kinds_multiline — the same two kinds split across two separate ignore-errors lines, verifying that the lists accumulate.

Together these cover the single-kind, comma-separated, and multi-line accumulation paths, and the negative case (an unlisted kind is not suppressed).

Running the tests
```
cargo test -p pyrefly suppression
```
Result: 78 passed, 0 failed (the full suppression suite, including the three new tests).

Note: Claude Opus 4.8 assisted with writing this PR.

### DIFF
--- a/crates/pyrefly_python/src/ignore.rs
+++ b/crates/pyrefly_python/src/ignore.rs
@@ -459,11 +459,28 @@ fn is_in_multiline_string(
     }
 }
 
+/// A whole-file ignore directive parsed from the file's preamble.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IgnoreAll {
+    /// The line where the directive for this tool first appears.
+    pub line: LineNumber,
+    /// The error kinds the directive suppresses. An empty list means *all*
+    /// errors. Non-empty lists only ever occur for `Tool::Pyrefly` (via
+    /// `# pyrefly: ignore-errors[kind, ...]`); the parser never records kinds
+    /// for other tools, so consumers can treat an empty list as "suppress all".
+    pub kinds: Vec<String>,
+}
+
 /// Parse top-level `ignore-errors` / `ignore-all-errors` / `type: ignore` directives.
 ///
 /// Scans the beginning of the file for comment-only lines (including blank lines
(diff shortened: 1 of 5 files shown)


rchen152:
Sorry for the lack of review - looks like this PR slipped through the cracks. I'll take a look on Monday.

rchen152:
Actually, this was already implemented in https://github.com/facebook/pyrefly/pull/3944.

PR #3986, opened 2026-06-30
Fix hover classification for union methods

xaskii:
# Summary

- We share attribute classification between semantic tokens and hover.
- Changes hover to check each member of a union when choosing the symbol kind for an attribute.
- This fixes cases like `x: A | B; x.foo()`, where hover should show `foo` as a method when both `A.foo` and `B.foo` are methods.

Related PR: https://github.com/facebook/pyrefly/pull/3940 (follow-up PR for related discussion)
Follow-up on https://github.com/facebook/pyrefly/issues/3873

# Test Plan

- ./test.py
- Added failing test + two more to make check intended behavior

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -62,6 +62,7 @@ use crate::state::lsp::FindDefinitionItemWithDocstring;
 use crate::state::lsp::FindPreference;
 use crate::state::lsp::IdentifierContext;
 use crate::state::lsp::IdentifierWithContext;
+use crate::state::lsp::attribute_symbol_kind_from_type;
 use crate::state::state::Transaction;
 use crate::state::state::TransactionHandle;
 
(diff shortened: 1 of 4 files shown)


NathanTempest (approved):
Reviewed and this looks good, approving for import. Nice refactor: pulling
classification into a single `attribute_symbol_kind_from_type` shared by hover
and semantic tokens is the right call, and I verified the semantic-token mapping
is 1:1 with the previous arms (Class→CLASS, TypeAlias→INTERFACE, Module→NAMESPACE,
Method→METHOD, Function→FUNCTION, Attribute→PROPERTY, ENUM_MEMBER still handled
before the fallback), so that path is behavior-preserving and just gains union
handling. The union logic (all members agree → that kind, else Attribute) is
correct and the mixed-union negative test is a nice touch.

yangdanny97 (approved):

xaskii:
nice thanks

PR #3990, opened 2026-06-30
Add opt-in --remove-unused-type-ignores flag

### LINKED ISSUE #3984: Feature: `--remove-unused-type-ignores` flag to clean up unused `# type: ignore` comments
### Is your feature request related to a problem?

`pyrefly suppress --remove-unused` (and `pyrefly check --remove-unused-ignores`) removes unused `# pyrefly: ignore` comments, but it does not touch unused `# type: ignore` comments, even though Pyrefly honors `# type: ignore` as a valid suppression.

This means that after I clean up my suppressions, any stale `# type: ignore` comments are left behind and I have to find and delete them by hand. For a large codebase that adopted `# type: ignore` (for example, coming from another type checker), there is no convenient way to clean these up with Pyrefly.

I understand the default likely excludes `# type: ignore` on purpose, since those comments are shared with other type checkers and removing them could affect another tool. So I am not asking to change the default behavior.

### Describe the solution you'd like

Add an opt-in flag that also removes unused `# type: ignore` comments:

- `pyrefly suppress --remove-unused-type-ignores`
- `pyrefly check --remove-unused-type-ignores`

This would work like the existing `--remove-unused` / `--remove-unused-ignores`, but additionally clean up `# type: ignore` comments that are not suppressing anything. Users who share `# type: ignore` with other tools simply would not pass the flag, so the default stays safe.

### Reproduction of current behavior

Given a file where every suppression is unused (no error on the line):

```python
a = 1  # pyrefly: ignore
b = 2  # type: ignore
```

Running `pyrefly suppress --remove-unused` removes only the first:

```python
a = 1
b = 2  # type: ignore
```

There is currently no flag that would also remove the unused `# type: ignore`.

### Additional context

The documentation for `--remove-unused` describes it only in terms of `# pyrefly: ignore` comments and does not mention `# type: ignore` handling, so it would be worth documenting the new flag and the default behavior together.

### Related
- #3766

nitishagar:
## Summary
`pyrefly suppress --remove-unused` and `pyrefly check --remove-unused-ignores` remove unused `# pyrefly: ignore` comments, but they leave unused `# type: ignore` comments behind — even though Pyrefly honors `# type: ignore` and already detects when one is unused. After cleaning up suppressions, any stale `# type: ignore` comments have to be found and deleted by hand, which is painful in codebases that adopted `# type: ignore` (e.g. coming from another type checker).

This adds an opt-in `--remove-unused-type-ignores` flag (on both `check` and `suppress`) that additionally removes unused `# type: ignore` comments, working just like the existing `--remove-unused` / `--remove-unused-ignores` flags. It is **off by default**, so `# type: ignore` comments — which are shared with other type checkers — are never touched unless the flag is explicitly passed.

The change threads a `remove_type_ignores` boolean into the existing removal path and adds a single gated arm that handles the `UnusedTypeIgnore` errors Pyrefly already produces; unused detection is unchanged.

```
$ pyrefly check --remove-unused-ignores --remove-unused-type-ignores repro.py
# removes both "# pyrefly: ignore" and unused "# type: ignore"
$ pyrefly check --remove-unused-ignores repro.py
# removes only "# pyrefly: ignore" (default, unchanged)
```

Fixes #3984.

## Test plan
- `cargo build -p pyrefly`
- `cargo test -p pyrefly remove_unused` — 23 passed, including new `test_remove_unused_type_ignore_inline` / `test_remove_unused_type_ignore_with_code` and a regression test asserting the comment is kept when the flag is absent.
- `cargo clippy -p pyrefly --tests`
- Manual repro from the issue: with the flag both comments are removed; without it only the pyrefly one is removed.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -232,6 +232,7 @@ impl SnippetCheckArgs {
                 suppress_errors: false,
                 expectations: false,
                 remove_unused_ignores: false,
+                remove_unused_type_ignores: false,
             },
         };
         let (status, check_result) =
@@ -419,6 +420,10 @@ struct BehaviorArgs {
     /// Remove unused ignores from the input files.
     #[arg(long)]
     remove_unused_ignores: bool,
+    /// Also remove unused `# type: ignore` comments (off by default; shared with
+    /// other type checkers).
+    #[arg(long)]
+    remove_unused_type_ignores: bool,
 }
 
 fn write_errors_to_file(
@@ -1302,7 +1307,10 @@ impl CheckArgs {
             // TODO: Deprecate this in favor of `pyrefly suppress`
             let collected = loads.collect_errors();
             let unused_errors = loads.collect_unused_ignore_errors(&collected);
-            suppress::remove_unused_ignores(unused_errors);
+            suppress::remove_unused_ignores(
+                unused_errors,
+                self.behavior.remove_unused_type_ignores,
+            );
         }
 
         // We update the baseline file if requested, after reporting any new
--- a/pyrefly/lib/commands/suppress.rs
+++ b/pyrefly/lib/commands/suppress.rs
@@ -39,6 +39,10 @@ pub struct SuppressArgs {
     #[arg(long)]
     remove_unused: bool,
 
+    /// Also remove unused `# type: ignore` comments when removing unused ignores.
+    #[arg(long)]
+    remove_unused_type_ignores: bool,
+
     /// Where to place suppression comments: on the line before the error
     /// (`line-before`, the default) or on the same line (`same-line`).
     #[arg(long, default_value = "line-before")]
@@ -57,9 +61,12 @@ impl SuppressArgs {
                 // Parse errors from JSON file, filtering for UnusedIgnore errors only
                 let json_content = std::fs::read_to_string(json_path)?;
                 let errors: Vec<SerializedError> = serde_json::from_str(&json_content)?;
+                let remove_type_ignores = self.remove_unused_type_ignores;
                 errors
                     .into_iter()
-                    .filter(|e| e.is_unused_ignore())
+                    .filter(|e| {
+                        e.is_unused_ignore() || (remove_type_ignores && e.is_unused_type_ignore())
+                    })
                     .collect()
             } else {
                 // Delegate to `check --remove-unused-ignores`, which calls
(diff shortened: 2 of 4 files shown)


NathanTempest (changes requested):
Thanks for this, @nitishagar! 

One correctness gap I'd like to see closed before this lands is the core "never touched unless the flag is passed" guarantee on lines with more than one pragma.

The removal in remove_unused_ignores_from_serialized uses IGNORE_COMMENT_REGEX.replace_all(comment_part, "") (error/suppress.rs:615). IGNORE_COMMENT_REGEX (line 47) matches all pragma families : pyrefly: ignore, type: ignore, pyre-fixme/ignore, pyre: ignore, so replace_all deletes every pragma on the line. But "unused" is decided per-pragma. The msg.starts_with(...) gate only decides whether to enter the removal block; once inside, replace_all removes all co-located pragmas regardless of which one was actually unused. Two concrete failures:

- Flag ON  x = 1 # type: ignore # pyrefly: ignore, where # type: ignore is unused but # pyrefly: ignore is still suppressing a real error. The new arm fires on the type-ignore error, replace_all also strips the used # pyrefly: ignore, and a real error is silently reintroduced. (This path is newly reachable because of this PR.)
- Flag OFF  x = 1 # pyrefly: ignore # type: ignore, where # pyrefly: ignore is unused but # type: ignore is still needed. Removing the pyrefly comment also deletes the # type: ignore even though the flag was never passed, which is exactly what the feature promises won't happen. (Pre-existing behavior of line 615, but this PR is what makes the promise, so worth fixing here.)

The current tests all use a single pragma per line (or the both-unused case), so CI stays green while this slips through.

Suggested fix: at line 615, remove only the unused pragma rather than blanket replace_all. Select a pragma-specific regex based on the error kind (you already have is_unused_type_ignore(); branching on error.name/kind instead of msg.starts_with(...) would also make the dispatch less brittle to message rewording).

I've written two regression tests that pin both cases (they fail on the current code and should pass once removal is pragma-scoped) which are pasted in the next comment. Feel free to push back or request a review after the follow-ups are addressed. Thanks for your effort in contributing to us :)

NathanTempest:
The regression tests:
```
#[test]
fn test_used_type_ignore_preserved_when_removing_pyrefly_ignore() {
    // Failure #2 (flag OFF): an unused `# pyrefly: ignore` shares a line with a
    // still-used `# type: ignore`. Removing the pyrefly comment must NOT delete the
    // type: ignore — it is never touched unless --remove-unused-type-ignores is passed.
    // Only the pyrefly comment is unused, so only it has an error.
    let input = "x = 1  # pyrefly: ignore  # type: ignore\n";
    let want = "x = 1  # type: ignore\n";
    let errors = vec![SerializedError {
        path: PathBuf::from("test.py"),
        line: 0,
        name: "unused-ignore".to_owned(),
        message: "Unused `# pyrefly: ignore` comment".to_owned(),
    }];
    assert_remove_ignores_from_serialized_with_flag(input, errors, want, 1, false);
}

#[test]
fn test_used_pyrefly_ignore_preserved_when_removing_type_ignore() {
    // Failure #1 (flag ON): an unused `# type: ignore` shares a line with a still-used
    // `# pyrefly: ignore`. Removing the type: ignore must NOT delete the pyrefly comment,
    // which is actively suppressing a real error.
    // Only the type: ignore is unused, so only it has an error.
    let input = "x = 1  # type: ignore  # pyrefly: ignore\n";
    let want = "x = 1  # pyrefly: ignore\n";
    let errors = vec![SerializedError {
        path: PathBuf::from("test.py"),
        line: 0,
        name: "unused-type-ignore".to_owned(),
        message: "Unused `# type: ignore` comment".to_owned(),
    }];
    assert_remove_ignores_from_serialized_with_flag(input, errors, want, 1, true);
}
```

tobyh-canva:
Hi @nitishagar are you still working on this? If not, do you mind if I claim it?

NathanTempest:
@tobyh-canva I have not heard any replies from @nitishagar since 5 days, let's wait 24 hours more. If they still don't reply, I will close this out and you can work on this. Thank you for your patience!

NathanTempest:
Hey @tobyh-canva feel free to pick this up, closing this PR for now.

PR #3991, opened 2026-06-30
Add --dry-run flag to pyrefly infer

### LINKED ISSUE #3920: Add --dry-run flag to pyrefly infer
### Describe the Bug

I'm finding it very useful to have `pyrefly infer` run as part of our code formatter. In continuous integration, we have a check verifying whether the formatter would make any changes. This is not currently possible with `pyrefly infer`, since it doesn't have a dry-run/check-only mode.

I propose adding a `--dry-run` flag to `pyrefly infer` which simply prints out what changes _would_ be made, with an exit code of 0 if no changes would be made and an exit code of 1 if any changes would be made. This would enable the above continuous integration usage for `pyrefly infer`.

### Sandbox Link

_No response_

_No response_

nitishagar:
## Summary
`pyrefly infer` rewrites files in place and always exits 0, so it cannot be used as a CI check for whether the formatter would make any changes.

This adds a `--dry-run` flag to `pyrefly infer` that runs the full inference, prints what changes would be made, writes nothing, and exits `0` if no changes would be made or `1` if any changes would be made — enabling CI check mode.

The change set is already computed by the existing read-only phase of `run_inner`; dry-run simply skips the two file writes (gated on a new `dry_run` flag on `InferFlags`), logs the planned changes per file, and returns `UserError` when any changes are pending. Non-dry-run behavior is unchanged.

```
$ pyrefly infer --dry-run file.py
 INFO file.py: would add 1 annotation(s) and 0 import(s)
$ echo $?
1
# file.py is unchanged
```

Fixes #3920.

## Test plan
- `cargo build -p pyrefly`
- `cargo test -p pyrefly dry_run` — 12 passed, asserting both the exit status (`UserError` when changes, `Success` when none) and that the file is left byte-identical.
- `cargo clippy -p pyrefly --tests`
- Manual: `pyrefly infer --dry-run` on an unannotated file leaves it unchanged and exits 1 with the planned change printed; without `--dry-run` the file is annotated and exits 0.

### DIFF
--- a/pyrefly/lib/commands/infer.rs
+++ b/pyrefly/lib/commands/infer.rs
@@ -22,6 +22,7 @@ use ruff_python_ast::Stmt;
 use ruff_python_ast::helpers::is_docstring_stmt;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextSize;
+use tracing::info;
 
 use crate::commands::check::Handles;
 use crate::commands::config_finder::ConfigConfigurerWrapper;
@@ -84,6 +85,9 @@ pub struct InferFlags {
         num_args = 0..=1
     )]
     pub imports: Option<bool>,
+    /// Print what would change and exit (1 if any changes, else 0), without writing any files.
+    #[arg(long)]
+    pub dry_run: bool,
 }
 
 impl InferFlags {
@@ -93,6 +97,7 @@ impl InferFlags {
             return_types: Some(true),
             parameter_types: Some(true),
             imports: Some(true),
+            dry_run: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


NathanTempest (approved):
Nice, clean addition and thanks for the thorough tests. I confirmed the two file writers (add_annotations_to_file / add_imports_to_file) are the only write paths and both are correctly gated behind !dry_run, so the "writes nothing" guarantee holds, and returning UserError for pending changes matches the existing coverage check --fail-under convention. The multi-file test (exit 1 + both files left byte-identical) and the non-dry-run regression guard are exactly the right things to pin.

One suggestion, non-blocking: the "would add N annotation(s)…" line uses info!, which can be filtered out depending on verbosity. Since the point of --dry-run is CI visibility, consider eprintln! (or println!) to guarantee it's always shown and that's what coverage check --fail-under does for its analogous message. Happy to approve after this minor tweak but this is non blocking.

yangdanny97 (approved):

PR #3992, opened 2026-06-30
fix [pyrefly][completions][auto-import] rank based on deprecated aliases #2345

### LINKED ISSUE #2345: [pyrefly][completions][auto-import] rank based on deprecated aliases
### Describe the Bug

The stdlib documentation [suggests that ](https://github.com/python/cpython/blob/f47e928574187ae97c778e326739b3aa0c7dc765/Lib/typing.py#L2778)`collections.abc.Iterable` should be used in place of `typing.Iterable`. there are many like this. 

Pyright handles this by adding additional deprecated aliases for python versions [here](https://github.com/microsoft/pyright/blob/c930453afb01df1f8914464c1b503b2bd4030078/packages/pyright-internal/src/analyzer/deprecatedSymbols.ts#L26). we should mark these as deprecated as well so other completions are ranked higher

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2345

Added version-gated deprecated stdlib alias detection for typing aliases, auto-import completions now tag those aliases deprecated, so ranking demotes them.

collections.abc reexports from bundled stubs are now surfaced for deprecated typing aliases.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -29,6 +29,7 @@ use pyrefly_python::module_path::ModulePathDetails;
 use pyrefly_python::module_path::ModuleStyle;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_python::symbol_kind::SymbolKind;
+use pyrefly_python::sys_info::PythonVersion;
 use pyrefly_python::sys_info::SysInfo;
 use pyrefly_types::type_alias::TypeAliasData;
 use pyrefly_util::gas::Gas;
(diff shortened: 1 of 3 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -100,6 +101,61 @@ mod quick_fixes;
 pub(crate) use self::quick_fixes::move_module::MoveModuleMemberContext;
 pub(crate) use self::quick_fixes::types::LocalRefactorCodeAction;
 
+pub(crate) fn is_deprecated_stdlib_alias(
```
kinto0, line 104: nit: can we add comments with links to where it says these are deprecated?

kinto0 (approved):
thank you!!

yangdanny97 (approved):

pyrefly/lib/state/lsp.rs
kinto0: also it probably belongs in the pyrefly_python crate. i'lll fix these up if I can merge it

PR #3999, opened 2026-06-30
Treat inherited-annotated attributes as typed in `pyrefly coverage`

### LINKED ISSUE #3997: `coverage check` reports inherited typed attribute as untyped
I stumbled upon this while working on pydot, where `pyrefly coverage check --public-only` reports:

```
 INFO Found `/home/joren/Workspace/pydot/pyproject.toml` marking project root, checking root directory with auto configuration
 WARN `pydot.dot_parser.DefaultStatement.attrs` is untyped [coverage-missing]
  --> src/pydot/dot_parser.py:82:14
   |
82 |         self.attrs = attrs
   |              -----
   |
ERROR type coverage 99.62% (264 of 265 typable) is below the 100.00% threshol
```

for the following code ([src](https://github.com/pydot/pydot/blob/046798fb49a960b0368c90c0bf830d48b4bef8a7/src/pydot/dot_parser.py#L54-L86)):

```py
class P_AttrList:
    attrs: dict[str, Any]

    # -- snip --

class DefaultStatement(P_AttrList):
    default_type: Final[str]

    def __init__(self, default_type: str, attrs: dict[str, Any]) -> None:
        self.default_type = default_type
        self.attrs = attrs   # ❌ 
```

But `attrs` is already typed in `P_AttrList`, the direct base class of `DefaultStatement`.

jorenham:
# Summary

A method-assigned attribute without its own annotation is no longer reported as untyped by `pyrefly coverage {check, report}` when a base class declares it:

```py
class P_AttrList:
    attrs: dict[str, Any]

class DefaultStatement(P_AttrList):
    def __init__(self, attrs: dict[str, Any]) -> None:
        self.attrs = attrs  # was: untyped
```

It takes the type quality of the nearest base annotation, so `dict[str, Any]` is typed and a bare `Any` is `any`.

Fixes #3997

# Test Plan

Regression tests added, and verified that `check` / `report` are still in sync

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -23,7 +23,7 @@ use pyrefly_python::module_path::ModuleStyle;
 use pyrefly_python::nesting_context::NestingContext;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_types::callable::PropertyRole;
-use pyrefly_types::class::ClassDefIndex;
+use pyrefly_types::class::Class;
 use pyrefly_types::class::ClassType;
 use pyrefly_types::types::Type;
 use pyrefly_util::forgetter::Forgetter;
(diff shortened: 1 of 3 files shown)


NathanTempest (approved):
Approving as this is a clean, well-targeted fix. Thanks @jorenham!

The core logic is exactly right: walking the subclass MRO with ancestors_no_object() and taking the nearest annotating base (via find_map) matches how a reader would reason about inherited attribute types, and adopting that base's type quality (so dict[str, Any] → typed, bare Any → any) is the correct semantic. 

Optional nit, non-blocking: a one-line comment on inherited_annotation_slots noting "nearest annotating base wins; unannotated bases don't count" would help future readers. Happy to see this land as-is.

grievejia (approved):

PR #4003, opened 2026-07-01
Support semantic tokens for format specifiers in logging calls 

### LINKED ISSUE #3908: highlight for %s in string
### Describe the Bug

like

```
logger.info("%s", x)
```

why not f-str

https://docs.astral.sh/ruff/rules/logging-f-string/

how clangd does:

Image

### Sandbox Link

_No response_

_No response_

kavix:
### Summary

Fixes #3908

This PR adds a `formatSpecifier` semantic token for Python logging calls, similar to how clangd highlights format specifiers in C/C++. 

Since Ruff flags Python logging f-strings (for lazy evaluation performance reasons via rule G004), users are encouraged to use percent-style formatting. This change ensures that when users follow this best practice, their format specifiers (`%s`, `%d`, `%(name)s`, etc.) are properly syntax-highlighted by the IDE for readability.

### Approach & Details
1. **Registered `formatSpecifier`:** Added this custom token type to the `SemanticTokensLegends` registry.
2. **Detection Heuristic:** Introduced a lightweight AST heuristic (`is_logger_candidate`) to detect common logging functions (e.g. `logger.info`, `logging.debug`, `LOGGER.warn`) to avoid costly symbol resolutions on the hot path.
3. **Format Specifier Parsing:** Added a `OnceLock` cached regex that strictly follows Python's printf-style format specifiers to scan string literals passed to the detected logging calls.
4. **Source Propagation:** Updated the `process_ast` semantic token builder pipeline to propagate the `source` text string down through the visitor so that string literals can be sliced and matched against the regex.
5. **Testing:** Added the `semantic_tokens_format_specifier` integration test to `lsp_interaction` which successfully asserts the LSP token mapping over the string `"Hello %s %d"`.

---


pyrefly/lib/lsp/wasm/semantic_tokens.rs
```diff
@@ -32,6 +32,8 @@ impl Transaction<'_> {
         let disabled_ranges = disabled_ranges_for_module(ast.as_ref(), *handle.sys_info());
         let mut builder = SemanticTokenBuilder::new(limit_range, disabled_ranges);
 
+        let source = module_info.contents().as_str();
```
kinto0: I don't think the source is entirely necessary in processing the ast. can we instead use `&ast` that we already send?

we should be able to use `StringLiteralValue`

pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs
```diff
@@ -93,3 +93,73 @@ fn semantic_tokens_import_submodule_alias() {
 
     interaction.shutdown().unwrap();
 }
+
+#[test]
```
kinto0, line 104: please add a test for `%%` - I think it will fail

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -90,6 +90,7 @@ impl SemanticTokensLegends {
                 SemanticTokenType::REGEXP,
                 SemanticTokenType::OPERATOR,
                 SemanticTokenType::DECORATOR,
+                SemanticTokenType::new("formatSpecifier"),
```
kinto0: AI is telling me this will allocate on every match of our token. can we avoid this somehow?

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -223,6 +224,50 @@ impl SemanticTokensLegends {
...
+        Expr::Attribute(attr) => Some(attr.attr.as_str()),
+        Expr::Call(call) => get_expr_tail(&call.func),
+        _ => None,
+    }
+}
+
+fn is_logger_candidate(func: &Expr) -> bool {
+    let Expr::Attribute(attr) = func else {
+        return false;
+    };
```
kinto0: should we instead make callers filter this?

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -223,6 +224,50 @@ impl SemanticTokensLegends {
...
+fn get_expr_tail(expr: &Expr) -> Option<&str> {
+    match expr {
+        Expr::Name(name) => Some(name.id.as_str()),
+        Expr::Attribute(attr) => Some(attr.attr.as_str()),
+        Expr::Call(call) => get_expr_tail(&call.func),
+        _ => None,
+    }
+}
+
+fn is_logger_candidate(func: &Expr) -> bool {
```
kinto0: docstring here would be nice so we understand what it should do

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -613,9 +678,40 @@ impl SemanticTokenBuilder {
...
+        let re = RE.get_or_init(|| {
+            Regex::new(r"%[-+#0 ]*(?:\*|\d+)?(?:\.(?:\*|\d+))?[hlL]?[diouxXeEfFgGcrsa]|%\(\w+\)[-+#0 ]*(?:\*|\d+)?(?:\.(?:\*|\d+))?[hlL]?[diouxXeEfFgGcrsa]").unwrap()
+        });
+
+        let start_offset = string_lit.range().start();
+        let end_offset = string_lit.range().end();
+        if start_offset.to_usize() > source.len() || end_offset.to_usize() > source.len() {
+            return;
+        }
+        let text = &source[start_offset.to_usize()..end_offset.to_usize()];
```
kinto0: I think we can just use `string_lit.value` and avoid source/range entirely here

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -223,6 +224,50 @@ impl SemanticTokensLegends {
...
+    }
+}
+
+fn is_logger_candidate(func: &Expr) -> bool {
+    let Expr::Attribute(attr) = func else {
+        return false;
+    };
+
+    let method = attr.attr.as_str();
+    if ![
```
kinto0: what makes you want to limit this to logging calls? it looks like #3908 is asking about printf which would not even be covered. can we try filtering in a different way?

pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs
```diff
@@ -93,3 +93,73 @@ fn semantic_tokens_import_submodule_alias() {
...
+    interaction.client.did_open_uri(&uri, "python", text);
+
+    let legend = SemanticTokensLegends::lsp_semantic_token_legends();
+    interaction
+        .client
+        .send_request::<SemanticTokensFullRequest>(json!({
+            "textDocument": { "uri": uri.to_string() }
+        }))
+        .expect_response_with(move |response| match response {
+            Some(SemanticTokensResult::Tokens(tokens)) => {
```
kinto0, line 125: preexisting but this test is very difficult to read. is there any way to make a helper that makes the test boilerplate small and makes it easy to see what we are testing against?

kinto0 (changes requested):
hey, thanks for taking this on. left a few suggestions

pyrefly/lib/lsp/wasm/semantic_tokens.rs
kavix: Refactored to use StringLiteralValue directly — iterating over each StringLiteral part via part.as_str() for regex matching and part.content_range().start() for source offsets, removing the need to pass source: &str through the pipeline.

pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs
kavix: You were right — added %% as the leading regex alternative so it's consumed before matching specifiers (fixing the %%s false-positive), with a new semantic_tokens_format_specifier_escaped_percent test asserting only %d is emitted from "100%% done, item %d"

pyrefly/lib/state/semantic_tokens.rs
kavix: SemanticTokenType::new is already const fn using Cow::Borrowed (no heap allocation), but I've defined a const FORMAT_SPECIFIER following the same pattern as all built-in token types in lsp_types (e.g. SemanticTokenType::NAMESPACE) and SELF_PARAMETER_MODIFIER already in this file, so the zero-cost intent is explicit.

pyrefly/lib/state/semantic_tokens.rs
kavix: Done — moved the Expr::Attribute guard to the call site so is_logger_candidate now takes &ExprAttribute directly, making its precondition explicit in the type signature.

pyrefly/lib/state/semantic_tokens.rs
kavix: Already added in the previous commit — the docstring explains the heuristic's purpose and why full symbol resolution is deliberately avoided on the hot path.

pyrefly/lib/state/semantic_tokens.rs
kavix: Already addressed — the current code uses string_lit.value.iter() and part.as_str() directly, with no source parameter or range slicing anywhere in the function.

pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs
kavix: Extracted a decode_semantic_tokens(tokens, legend, source) helper that converts the LSP delta-encoded stream into a plain Vec<(type_name, token_text)> — each test assertion is now a one-liner filter + collect, and the preexisting semantic_tokens_import_submodule_alias test was cleaned up too.

pyrefly/lib/state/semantic_tokens.rs
kavix: That's a great point. I've added a specific check for Expr::BinOp using Operator::Mod — this means any standard string modulo formatting ("format %s" % args) is now unambiguously detected and highlighted without relying on any heuristics.

I've kept the is_logger_candidate heuristic alongside it because logger.info("format %s", arg) passes the format string as a positional argument rather than using the % operator, so it requires a different detection pattern. Both cases are now covered (and tested!).

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -223,6 +227,54 @@ impl SemanticTokensLegends {
...
+
+/// Returns `true` if `attr` looks like a call to a standard logging method
+/// (e.g. `logger.info`, `logging.debug`, `LOGGER.warn`).
+///
+/// This is a lightweight AST heuristic used to decide whether to scan the
+/// first string argument for printf-style format specifiers.  It intentionally
+/// avoids full symbol resolution to stay off the hot path: a false negative
+/// (missing a real logger) is acceptable; a false positive (scanning a
+/// non-logging call) is harmless.
+fn is_logger_candidate(attr: &ExprAttribute) -> bool {
```
kinto0: I still don't really understand why we are filtering this to logging calls. the issue is for literal printing. can we do it on every string literal within a function call context? why / why not?

pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs
```diff
@@ -17,6 +19,43 @@ use crate::object_model::InitializeSettings;
 use crate::object_model::LspInteraction;
 use crate::util::get_test_files_root;
 
+/// Decodes an LSP delta-encoded semantic token stream into `(type_name, token_text)` pairs.
+///
+/// The LSP semantic token format encodes positions as deltas relative to the
+/// previous token.  This helper resolves those deltas and slices the source text
+/// so callers can simply filter and assert on plain strings.
+fn decode_semantic_tokens(
```
kinto0, line 27: looks great! even easier, we can just leave it as a string since it's for a test. but this is better than nothing!

I would even ask that this test change be split into a separate PR if you want. that would be a very fast stamp and merge so that this PR (#4003) can be smaller and easier to review

kinto0 (changes requested):
remaining questions about the logging calls

pyrefly/lib/state/semantic_tokens.rs
kavix: You're right, the logging-specific heuristic was unnecessarily narrow. I've removed is_logger_candidate entirely — now any string literal that's the first positional argument of any function call gets its printf-style specifiers highlighted. This is simpler and covers all cases (logging, print, custom formatters, etc.) without maintaining a list of "known" function names.

pyrefly/lib/test/lsp/lsp_interaction/semantic_tokens.rs
kavix: Thanks! Since decode_semantic_tokens is already required by the new tests in this PR, I can't fully split it out. But I can revert the refactoring of the pre-existing semantic_tokens_import_submodule_alias test and send that as a follow-up PR if you'd prefer to keep this one focused. Let me know!

kinto0: nah whatever's easier (which is just leaving it at this point). for next time, even if it's required in this PR, you can do the first PR against main and this PR against that branch so the reviews are isolated

kinto0:
thanks! looks good to me. then we will get another quick set of eyes on it and then we can merge!

kavix:
Hi @kinto0, could you please Re-review this PR ? Thanks!

yangdanny97 (changes requested):

kavix (commented):
Great feedback — thanks for the detailed review.

I agree on both points and will follow up with

1. Additional format-specifier coverage for %%, named + width/precision cases, and malformed/incomplete specifiers.
2. Clear documentation of is_logger_candidate boundaries (expected false positives/false negatives), plus focused tests for those limits where practical.

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -610,6 +640,44 @@ impl SemanticTokenBuilder {
...
+    ///
+    /// Each `StringLiteral` part carries its decoded `value` and a `content_range()`
+    /// that gives the source offset of the content (i.e. after the opening quote).
+    /// Because format specifiers like `%s` are plain ASCII with no escape sequences,
+    /// byte offsets in the decoded string map 1-to-1 to source byte offsets within
+    /// the content region.
+    fn process_format_specifiers(&mut self, string_lit: &ruff_python_ast::ExprStringLiteral) {
+        use std::sync::OnceLock;
+
+        use regex::Regex;
```
kinto0: feedback from danny: import these at the top of the file

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -223,6 +227,15 @@ impl SemanticTokensLegends {
     }
 }
 
+fn get_expr_tail(expr: &Expr) -> Option<&str> {
+    match expr {
+        Expr::Name(name) => Some(name.id.as_str()),
+        Expr::Attribute(attr) => Some(attr.attr.as_str()),
+        Expr::Call(call) => get_expr_tail(&call.func),
+        _ => None,
+    }
```
kinto0: never used?

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -610,6 +640,44 @@ impl SemanticTokenBuilder {
...
+
+        for part in string_lit.value.iter() {
+            let content_start = part.content_range().start();
+            for m in re.find_iter(part.as_str()) {
+                if m.as_str() == "%%" {
+                    // Escaped literal percent, not a format specifier.
+                    continue;
+                }
+                let start = content_start + TextSize::try_from(m.start()).unwrap();
+                let end = content_start + TextSize::try_from(m.end()).unwrap();
```
kinto0: unwrap here is too sketchy

pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -610,6 +640,44 @@ impl SemanticTokenBuilder {
...
+        use regex::Regex;
+        static RE: OnceLock<Regex> = OnceLock::new();
+        let re = RE.get_or_init(|| {
+            // `%%` must be the first alternative so it is consumed before the engine
+            // can treat the second `%` as the start of a format specifier.  Matches
+            // on `%%` are skipped below; they are not emitted as tokens.
+            Regex::new(r"%%|%[-+#0 ]*(?:\*|\d+)?(?:\.(?:\*|\d+))?[hlL]?[diouxXeEfFgGcrsa]|%\(\w+\)[-+#0 ]*(?:\*|\d+)?(?:\.(?:\*|\d+))?[hlL]?[diouxXeEfFgGcrsa]").unwrap()
+        });
+
+        for part in string_lit.value.iter() {
```
kinto0, line 732: The offset mapping assumes byte offsets in the decoded string (part.as_str()) map 1-to-1 to source byte offsets within content_range(). This is only true when the content preceding a format specifier contains no escape sequences. For a string like "\t%s" or "\n%s", the decoded value is shorter than the raw source (e.g. \t is one decoded byte but two source bytes), so content_start + m.start() will point at the wrong source offset and the emitted formatSpecifier token will be shifted, highlighting the wrong characters. The docstring's claim that format specifiers being ASCII guarantees a 1-to-1 mapping overlooks non-ASCII/escaped content that appears before the specifier. Consider mapping regex match offsets back to source positions accounting for escapes, or scanning the raw source slice instead of the decoded value. None of the added tests exercise a string with an escape sequence before a specifier.

kinto0 (changes requested):
sorry, a few more changes were requested (I don't think you saw these yet)

pyrefly/lib/state/semantic_tokens.rs
kavix: > feedback from danny: import these at the top of the file

Done — moved both OnceLock and Regex to the top-level imports of the file.

pyrefly/lib/state/semantic_tokens.rs
kavix: Fixed — replaced .unwrap() with checked if let (Ok(start_offset), Ok(end_offset)) = ... pattern matching.

pyrefly/lib/state/semantic_tokens.rs
kavix: Removed — cleaned up the unused get_expr_tail helper.

pyrefly/lib/state/semantic_tokens.rs
kavix: > The offset mapping assumes byte offsets in the decoded string (part.as_str()) map 1-to-1 to source byte offsets within content_range(). This is only true when the content preceding a format specifier contains no escape sequences. For a string like "\t%s" or "\n%s", the decoded value is shorter than the raw source (e.g. \t is one decoded byte but two source bytes), so content_start + m.start() will point at the wrong source offset and the emitted formatSpecifier token will be shifted, highlighting the wrong characters. The docstring's claim that format specifiers being ASCII guarantees a 1-to-1 mapping overlooks non-ASCII/escaped content that appears before the specifier. Consider mapping regex match offsets back to source positions accounting for escapes, or scanning the raw source slice instead of the decoded value. None of the added tests exercise a string with an escape sequence before a specifier.

"Great catch! I updated process_format_specifiers to scan the raw source slice (source.get(content_range)) directly rather than the decoded string part.as_str(). This ensures match byte offsets align 1:1 with source byte positions regardless of preceding escape sequences (\n, \t, \x.., \u....). Added the semantic_tokens_format_specifier_escapes regression test covering escaped and non-ASCII prefixes.

kavix:
Hi @kinto0 @yangdanny97 , could you please Re-review this PR ? Thanks!

kavix:
Hi @kinto0 @yangdanny97 , could you please Re-review this PR ? Thanks!

kinto0:
hey, sorry about the delay and the back/forth. it actually looks like we already support these format specifiers syntactically in editors that combine our results with textmate grammars (see the issue). if it's absolutely necessary, we might add a textmate parser for all syntax, but until then, I don't think the pyrefly analysis should have this role. 

I'm sorry I marked it a good-first-issue. I now can see our struggles implementing it and realize it is difficult and out of scope for pyrefly.

Issue #1517, opened 2025-11-06
go-to definition / hover includes the lhs in operators


kinto0:
### Describe the Bug

hover on the 1 in the sandbox

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAjAAQDU1lIANCAK4Au0cJ5iIAxNQCqnKBHalqYVugDGnXOjgAddKswwwU3ACcAtqnYB9dKz3YYOgBT5E1COnYBKagFoAfNTjsdiVdQDqHRh2Vh10KWUQADkzC19qYHwAXyjVZhAyYLAoUkJ2XD0oCkEABVJs3K8MHAJqWUVIAHMwwwhFQlVBAGUYGGoAC3Z2YjhEAHpxrM1cwl0m8Zh0ccxcWThxhvRm1oVl7R1qVAA3VGhUbFh6xogWnTbFalxiPe5VMnYBxVdjyzh2iIAXmoUQAzIRKAAmNLoEDJFioeQQX4AMWgMAoaCweCIZDhQA

asukaminato0721:
claim

PR #4006, opened 2026-07-01
fix support on hover for += and so on #4005

### LINKED ISSUE #4005: support on hover for `+=` and so on
### Describe the Bug

now we have `x + 1`, but don't have`x += 1` .

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4005

added augmented-assignment operator-region detection and `__iadd__` definition/type lookup for +=.

wired augmented-assignment operator type lookup into hover.

# Test Plan

added regression coverage for `+=` hover and verified RHS literal hover is preserved.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -53,7 +53,6 @@ use ruff_python_ast::ExprName;
 use ruff_python_ast::Identifier;
 use ruff_python_ast::Keyword;
 use ruff_python_ast::ModModule;
-use ruff_python_ast::StmtAugAssign;
 use ruff_python_ast::StmtImportFrom;
 use ruff_python_ast::UnaryOp;
 use ruff_python_ast::name::Name;
@@ -113,6 +112,7 @@ pub(crate) enum CalleeKind {
     Unknown,
 }
 
+/// The receiver type, method name, and call surface for an operator dunder.
 struct OperatorDunder {
     base_type: Type,
     dunder_name: Name,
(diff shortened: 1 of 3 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -2094,16 +2109,40 @@ impl<'a> Transaction<'a> {
         self.get_chosen_overload_trace_for_surface(handle, subscript.range(), true)
     }
 
+    pub(crate) fn augmented_assignment_operator_type_at(
```
kinto0: instead can we reuse find_operator_dunder?

kinto0 (changes requested):

pyrefly/lib/state/lsp.rs
```diff
@@ -111,6 +112,12 @@ pub(crate) enum CalleeKind {
     Unknown,
 }
 
+struct OperatorDunder {
```
kinto0, line 116: nit: docstring

pyrefly/lib/state/lsp.rs
```diff
@@ -123,6 +130,10 @@ fn default_true() -> bool {
     true
 }
 
+fn position_in_augassign_operator(augassign: &StmtAugAssign, position: TextSize) -> bool {
```
kinto0: is the only reason for this + position threading to avoid https://github.com/facebook/pyrefly/issues/1517 ?

I think that should be fixed separately since it's pre-existing. let's keep it out of this PR

kinto0 (changes requested):

stroxler (approved):

PR #4009, opened 2026-07-01
Bound peak memory in `pyrefly coverage` with chunked collection

### LINKED ISSUE #3989: `pyrefly coverage check` uses significantly more memory than `pyrefly check`
### Describe the Bug

When running `pyrefly==1.1.1` in a monorepo, I've observed the following:

```
• Measured from:

     /usr/bin/time -l pyrefly check
     /usr/bin/time -l pyrefly coverage check

  Results:

   Command           │ Exit … │        Maximum RSS │   Approx MiB │ Approx GiB │ Peak memory footp…
  ───────────────────┼────────┼────────────────────┼──────────────┼────────────┼────────────────────
   pyrefly check     │      0 │      3,577,282,560 │ 3,411.56 MiB │   3.33 GiB │      3,561,147,536
                     │        │              bytes │              │            │              bytes
   pyrefly coverage  │      1 │      8,630,730,752 │ 8,230.91 MiB │   8.04 GiB │      8,617,122,536
   check             │        │              bytes │              │            │              bytes
```

I've had to increase the memory requested for our `pyrefly` step in CI/CD so that we could run `pyrefly check` and `pyrefly coverage check` together, but I wasn't expecting this drastic increase in memory usage.

### Sandbox Link

_No response_

_No response_

jorenham:
# Summary

`pyrefly coverage {check, report}` kept every module's bindings/answers/ASTs in memory all at once. It now collects in chunks, evicting each after use, without changing behavior.

| Corpus | RSS before | RSS after | time before | time after |
|---|---:|---:|---:|---:|
| numpy, 530 modules | 4.26 GiB | 3.11 GiB | 2:41 | 2:58 |
| scipy-stubs, 589 modules | 1.06 GiB | 0.83 GiB | 3.3s | 4.5s |

(debug build)

I tuned the chunk size carefully and chose the one that's roughly speaking in the "elbow" of the memory-decrease vs. runtime-increase graph (viz. 256). For large project (>256 modules) runtime impact is unavoidable though, since this is a classic space/time tradeoff. But since it already is pretty fast, this relative runtime increase is still pretty fast. 

Fixes #3989 (Although I don't have access to the mentioned monorepo so I can't say for sure)

# Test Plan

Regression test added, and verified that this did not change any of the results for scipy-stubs and numpy.

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -1657,8 +1657,6 @@ pub fn collect_module_reports(
 
     let mut module_reports: Vec<ModuleReport> = Vec::new();
     let mut errors: Vec<Error> = Vec::new();
-    transaction.run(handles.as_slice(), Require::Everything, None);
-    let public_fqns = public_only.then(|| compute_public_fqns(&handles, transaction));
 
     let shadowed = if prefer_stubs {
         py_paths_shadowed_by_pyi(&handles)
@@ -1685,7 +1683,6 @@ pub fn collect_module_reports(
             })
             .collect();
         // Fall back to site-package-path for stubs-only packages.
-        let mut external_handles = Vec::new();
         for h in handles.iter().filter(|h| h.path().is_interface()) {
             let pyi_path = h.path().as_path().to_path_buf();
             if map.contains_key(&pyi_path) {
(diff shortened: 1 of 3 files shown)


jorenham:
@yangdanny97 if you found a better solution, then feel free to close this one

rexledesma (commented):
Happy to run any local builds of this change on my monorepo if that's of any help!

> But since it already is pretty fast, this relative runtime increase is still pretty fast.

Yeah I'm not particularly latency sensitive either on `pyrefly coverage XXX` since this coverage command is not on the hot path of the agent loops I've implemented. It's only running in CI/CD.

I am pretty latency sensitive on `pyrefly check` though, since that will be running constantly.

jorenham:
> I am pretty latency sensitive on `pyrefly check` though, since that will be running constantly.

This won't affect `pyrefly check`; it's limited to `pyrefly coverage {report, check}` :)

jorenham:
I'm trying out an alternative approach here that might be better, so marking this as draft for the time being.

jorenham:
> I'm trying out an alternative approach here that might be better, so marking this as draft for the time being.

Wow I'm glad I tried that alternative approach; closing in favor of #4012

Issue #4010, opened 2026-07-01
impl some of basedmypy's extension


asukaminato0721:
https://kotlinisland.github.io/basedmypy/based_features.html

yangdanny97:
The syntax-related extensions are unlikely to be supported without a change to the typing spec

asukaminato0721:
I don't meant to impl those, but some "bug" can be fixed.

example:

```py
from functools import cache
@cache
def f(): ...
f(1, 2, 3)  # error: expected no args
```

jorenham:
The checked f-strings (https://kotlinisland.github.io/basedmypy/based_features.html#checked-f-strings) and regex type-checking (https://kotlinisland.github.io/basedmypy/based_features.html#regex-checks) are also very cool.

PR #4013, opened 2026-07-02
fix Typed functools.Cache #4010

asukaminato0721:
# Summary

Fixes part of #4010

now models `functools._lru_cache_wrapper` as an intersection of the runtime wrapper and the original callable, preserving call signatures while keeping cache_info / cache_clear.

preserves binding behavior through intersections/KwCall wrappers for cached methods.

# Test Plan

added coverage for cache, lru_cache, called decorators, and methods.

### DIFF
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -1771,6 +1771,8 @@ impl Type {
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


yangdanny97:
Huh, I think @shobhitmehro was looking at a similar approach to functools.partial, but ran into some issues. The primer results here seem fairly mixed, so would be good to dig into things more

PR #4015, opened 2026-07-02
fix regex check #4010

asukaminato0721:
# Summary

Fixes part of #4010

Added literal regex parsing for capturing groups and invalid unbalanced patterns.

Added regex error kind.

~~Propagated regex group metadata through re.compile, re.search, re.match, re.fullmatch, and compiled pattern `.search/.match/.fullmatch`.~~

~~Tightened `Match.groups()` and literal `Match.group(...)` return types.~~

~~Reports missing named/numeric regex groups.~~

need redesign

# Test Plan

Added focused integration coverage in test_regex_checks.


NathanTempest (commented):
Thanks for the thorough work here @asukaminato0721! The iteration to kill the Match[str] subscript regression clearly paid off, and it's catching real bugs (steam.py def_index, apprise id). Two things before this can move forward, both above the code itself:

- Scope/spec: #4010 is still needs-discussion / typing-spec, and this adds a corpus-wide new check. @yangdanny97 can you stamp if pyrefly wants regex group-checking at all before we invest in landing it?

- Mechanism: encoding group metadata as a Type::Annotated("__pyrefly_regex_groups__", …) sentinel means every one of the ~23 Type::Annotated sites has to see through it; this PR covers 3, and the earlier corpus-wide Match[str] is not subscriptable regression came from exactly that gap. I think this needs @samwgoldman  or @stroxler  to weigh in on whether an in-band Annotated sentinel is the right representation vs. a dedicated type/side-table. The current approach will keep leaking at the untouched sites (solve, specials, query/LSP, tsp, cinderx export).

Minor, if it proceeds: has_regex_metadata is duplicated across display/attr/subset (hoist to alt/regex.rs); the subset.rs got-vs-want arms are asymmetric (!matches!(want, Type::Type(_)) on one side only); and parse_groups misses the re.VERBOSE flag arg + hardcodes position 0 in the final unterminated-subpattern error.

NathanTempest:
Thanks @asukaminato0721 for the follow-up. It looks good to merge but a few nits from me:

1. has_regex_metadata is copy-pasted identically in three files (display.rs, attr.rs, subset.rs), and expr.rs has a fourth variant (regex_metadata_to_groups) doing the same tag-matching. This might cause drifts later on.

2. Unmatched-paren error always reports position 0. In parse_groups:
```
if depth != 0 {
    return Err("missing ), unterminated subpattern at position 0".to_owned());
}
```
The position is hardcoded, so re.compile("a(b(c)") blames position 0 instead of the real offset. The other unbalanced-paren errors already carry {i} worth tracking the byte offset of each unclosed ( (e.g. push the ( position alongside the working entry, or keep a small stack of open-paren offsets) and report the outermost still-open one here.

@stroxler could you review this PR after these nits are addressed ?

NathanTempest (changes requested):
Thanks for pushing this further, @asukaminato0721. The feature is landing real catches (steam.py def_index, apprise id so that is a nice win). But the mypy_primer run shows a broad side effect we need to solve before this can go in: Match[str] is now leaking into inferred union types across ~30 projects. Most are churn on existing implicit-bool messages (the type string gains Match[str]), but a few are genuine new false positives where it reaches positions the code never widened for:

- openlibrary/core/models.py:268 and addbook.py:359 -  new bad-return (… | Match[str] | … vs the declared … | None)
- werkzeug/debug/repr.py:202 - new unknown-argument-type
Root cause looks like the Annotated[..., ("__pyrefly_regex_groups__", …)] carrier (plus the subset.rs erasure) interfering with union narrowing so Match[str] no longer gets narrowed out after truthiness checks / in returns. 

This is the design question I think needs @stroxler / @samwgoldman: how to attach group metadata without it riding on the Match type through inference (e.g. a side-table keyed by expression/range rather than an Annotated wrapper). The 2 real catches are a great signal the core idea works.

pyrefly/lib/alt/expr.rs
```diff
@@ -1072,7 +1081,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     .into_ty()
             }
         } else {
-            self.expr_call_infer(x, callee_ty, hint, errors)
+            let ret = self.expr_call_infer(x, callee_ty.clone(), hint, errors);
```
stroxler: Let's not introduce a clone on the general branch here.

Instead, could we rewire `regex_callee` to do more work, and return optional data that is `None` if none of the regex special cases are relevant?

By doing so, we should be able to extract the regex callee first, before `callee_ty` is consumed, and then handle it after when there's a non-None result from the callee lookup. That will save a clone on a very hot path

pyrefly/lib/alt/expr.rs
```diff
@@ -1844,6 +1854,256 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            Some(RegexGroupKey::Int(index)) => {
+                self.error(
+                    errors,
+                    arg.range(),
+                    ErrorKind::Regex,
+                    format!("No such group: {index}"),
+                );
+            }
+            Some(RegexGroupKey::Str(name)) => {
+                match groups.iter().find(|group| group.name == Some(name.clone())) {
```
stroxler: This allocates a new String per iteration. A zero-alloc alternative: group.name.as_deref() == Some(name.as_str()).

stroxler (changes requested):

stroxler (approved):

pyrefly/lib/alt/attr.rs
```diff
@@ -2580,6 +2581,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             Type::ElementOfTypeVarTuple(_) => {
                 acc.push(AttributeBase1::ClassInstance(self.stdlib.object().clone()))
             }
+            Type::Annotated(inner, metadata) if regex_metadata_groups(&metadata).is_some() => {
```
stroxler: Hmm. Okay, I've uncovered a problem that I didn't anticipate from my first review.

We're modifying the Pyrefly code to use `Annotated` as a fake type in order to carry around metadata from `Match`. This is a problem because Pyrefly isn't really expecting to see Annotated types in most places - in particular, we've never *synthesized* an Annotated before, they only occurred in annotations.

This is causing some problems for us, including crashing some jobs that affect instagram.

I think that the solution might be twofold:
- First, maybe we can split this PR into a PR for just the errors that can be found locally inside an `re.compile` or `re.match` call versus the ones that need metadata on `Match`
- Then, we need to think through our options for carrying metadata around with `Match`. I'm not sure what we should actually do, but using `typing.Annotated` to attach "extended type information" metadata seems questionable.

---------


The problem of making Match carry metadata is pretty nontrivial because if we modify `Type` the fanout is quite large (which is where the problems arose on this change). This might need a wider design discussion, it might be good to get @rchen152's opinion

One option might be to create a new `Type::WithMetadata` variant and make it so that the type checker systematically treats a type with metadata as the base type in almost every match on Type, and then just the very specific relevant matches would "intercept" specific kinds of metadata. In principle this might be useful for other things in the future (for example maybe we'll someday map special cases of Annotated that we understand for static analysis into WithMetadata).

The idea here is that WithMetadata would be similar to Annotated, except it would express metadata *that Pyrefly knows and cares about* whereas Annotated is expressing *a user statement* of arbitrary metadata, which in almost every case Pyrefly does *not* care about. In terms of other potential uses for this, I actually think some of the directions proposed for Polars Dataframe support might be very similar and WithMetadata might work well.

Another option might be to try to hardcode a modification of `Match` (maybe hardcoded into our internals) to use a literal-preserving type variable; we actually have some support for this already because I implemented a related feature for `shape_extensions.Flag`. But there are complications; `Match` uses `AnyStr`, which is a "constrained" type variable, whereas Flag is a "bound" on a type variable. So getting perfect semantics might be difficult.

stroxler (changes requested):
Sorry, back for more discussion - turns out this change is more involved than I thought on my first review.

stroxler:
The Annotated encoding is the main problem here and really needs some rework / discussion.

Just so we don't lose track of some more minor issues, here are some other notes from Claude on potential minor problems:

-------------------

# A few bugs in the logic that are showing up in mypy primer

  **mypy_primer +72/-64 — mostly a truthiness-narrowing regression**

  The bulk of the churn is bool | None → bool | Match[str] | None and reorderings like Match[str] |
  bool → bool | Match[str].

  Root cause is as_bool at pyrefly/lib/alt/expr.rs:2269:

  } else if let Type::ClassType(cls) = ty {
      if !self.is_subclassable(cls) && self.class_instances_always_truthy(cls) {
          return Some(true);

  Annotated[Match[str], meta] isn't a Type::ClassType, so as_bool returns None, and boolop's
  should_discard (expr.rs:2318) can no longer drop the always-truthy Match member from x and ... / x
  or .... get_condition_redundant_reason (expr.rs:5026) and check_dunder_bool_is_callable go blind the
  same way.

  This is not just cosmetic — it produces new false-positive errors:

  - openlibrary/core/models.py:268 — Returned type int | Match[str] | None is not assignable to
  declared return type int | None [bad-return]
  - openlibrary/plugins/upstream/addbook.py:359 — same shape
  - pyinstrument/processors.py:287 — new implicit-bool
  - werkzeug/debug/repr.py:202 — new unknown-argument-type

  **mypy_primer new [regex] errors — false positives, two distinct parser bugs**

  I extracted parse_groups into a standalone binary and fuzzed it against CPython's re (40k valid
  patterns).

  Bug A — the flags argument is never read. expr.rs:1952 hardcodes parse_groups(value.to_str(),
  false), so re.VERBOSE patterns are parsed in non-verbose mode and # comments are treated as pattern
  text. This is exactly the sympy hit; I reproduced it against the vendored copy at
  third-party/pypi/sympy/1.14.0/sources/sympy/parsing/mathematica.py:187:
```
  add*_1 pattern   python(re.VERBOSE)=compiles fine   pyrefly=ERR unbalanced parenthesis at position
  54
  add*_2 pattern   python(re.VERBOSE)=compiles fine   pyrefly=ERR missing ), unterminated subpattern
  at position 86
```
  Those are the two sympy/parsing/mathematica.py errors in the primer output, verbatim positions and
  all.

  Bug B — the inline-flag heuristic misfires on every other (?...) construct. In regex.rs, the
  fallback branch:
```
  } else if let Some(end) = value[i + 2..].find(')') {
      if value[i + 2..i + 2 + end].contains('x') {
          verbose = true;
```
  treats any unrecognized (?...) as an inline-flag group and enables verbose mode if the text before
  the next ) contains an x. So (?i:, (?=, (?>, (?<=, (?P= all silently flip verbose:
```
  '(?i:x)(?P<n>a)#(b)'    pyrefly= 1 group  ['n']    python= 2 groups {'n': 1}
  '(?=x)(?P<n>a)#(b)'     pyrefly= 1 group  ['n']    python= 2 groups {'n': 1}
  '(?>x)(?P<n>a)#(b)'     pyrefly= 1 group  ['n']    python= 2 groups {'n': 1}
  '(?<=x)(?P<n>a)#(b)'    pyrefly= 1 group  ['n']    python= 2 groups {'n': 1}
  '(?i:y)(?P<n>a)#(b)'    pyrefly= 2 groups ['n','-'] python= 2 groups {'n': 1}   <- control
```
  Dropping groups silently is what produces "No such group" — this is very likely the steam.py
  'def_index' and apprise 'id' hits. (Also worth noting: inline flags not at the start of the pattern
  have been a re.error since 3.11, so treating them as pattern-wide is wrong regardless.)

  Fixes: read the flags argument (positional and keyword) and thread re.VERBOSE through; restrict the
  flag branch to a strict (?[aiLmsux]*) / (?[aiLmsux]*: shape. More generally — when the parser hits a
  construct it doesn't fully understand, it should bail out and return no metadata rather than guess.
  A 180-line hand-rolled partial parser will keep generating FPs otherwise.

  Fuzzing found no wrong group names/ordering and no panics outside the verbose issues, so the core
  group-numbering logic looks sound.


# Other code-level notes

  1. RegexGroup.required is dead code. It's computed by ~50 lines of union/quantifier tracking in
  parse_groups, written into the metadata, read back by regex_groups_from_metadata — and never
  consumed. The commit message claims "Tightened Match.groups() and literal Match.group(...) return
  types," but that isn't in this diff. Either land the return-type tightening or drop required and the
  |/)?/)* machinery. It's the majority of the parser's complexity, it's untested, and it makes two
  patterns that differ only in optionality produce different Types.
  2. .expect() in regex.rs — open_parens.pop().expect("a comment group must have an opening
  parenthesis"). Currently unreachable (the (?# branch always pushes first), but pyrefly's own
  agents.md says to encode the invariant in the types rather than assert it. Restructure so it can't
  be expressed.
  3. Byte-indexed scanning over a &str. bytes[i] as char combined with value[i+2..] slicing is safe
  today only because each slice index is guarded by an ASCII byte check at i+1/i+3. Any new branch
  that slices at i+k without such a guard panics on a non-ASCII pattern. Iterate char_indices or stay
  in bytes consistently.
  4. The receiver is inferred twice. regex_groups_from_pattern_method_receiver and
  regex_validate_match_group_call both call self.expr_infer(&attr.value, errors) on an expression the
  call machinery already inferred. Errors are deduped by range so it isn't user-visible, but it's
  wasted work that nests on chains (re.compile(p).search(s).group(1)), and both re-walk the Expr AST —
  which agents.md explicitly warns against. Thread the already-computed receiver type instead.
  5. regex_callee runs on every call expression in the program. The module-name check is first so it
  should be cheap, but worth confirming it doesn't show up in a profile.
  6. Error positions don't line up with source. at position {i} is an offset into the cooked string
  value, while the reported range is the whole string literal. For re.compile("\\d(") the number won't
  match anything the user can see. Either point the range at the real offset inside the literal or
  drop the position.
  7. Coverage gaps. re.sub/re.subn and the Pattern.finditer/findall/split/sub methods aren't
  validated. Incomplete rather than wrong.
  8. Test coverage is thin for the amount of machinery — one integration testcase, and zero unit tests
  for parse_groups, which is the part most likely to be wrong. It's a pure function; a table-driven
  #[test] in regex.rs would have caught both verbose bugs in minutes. Also missing: bytes patterns,
  non-literal patterns, re.VERBOSE, and flow through a variable.
  9. Error-kind naming. A bare regex kind covering both "invalid pattern" and "bad group access" is
  vague next to neighbors like bad-return/implicit-bool, and it means users can't silence one without
  the other. Consider invalid-regex + missing-regex-group.
  10. Layering. regex_metadata_groups lives in crates/pyrefly_types but is a checker-feature concern.
  That's a symptom of the design point above.
  11. RegexCall::ValidatePattern calls regex_groups_from_pattern_argument purely for its error side
  effect and discards the result — reads oddly without a comment.

stroxler (approved):

PR #3994, opened 2026-06-30
Hard code using the dir entry cache struct (#3994)

kinto0:
Summary:

Internal results using telemetry show that using directory caching speeds up did open operations and improves language server speed. This experiment has been running for a few weeks now, and things are looking good, so let's just enable this fully

Reviewed By: grievejia

Differential Revision: D109356961

Issue #3993, opened 2026-06-30
High CPU on large projects dominated by filesystem syscalls during import resolution (Windows)


rchiodo:
## Summary

On a large project (the PyTorch repo: ~2,430 modules / ~1.32M lines), a cold full-project
analysis spends the **majority of its CPU in filesystem syscalls during import/module
resolution**, not in type checking. On Windows this is heavily amplified because every probe
traverses the NTFS minifilter stack (Microsoft Defender + cloud-files / OneDrive filter).

I profiled a locally-built, symbolized `release` binary with an ETW sampling profiler (samply)
and symbolicated against the matching PDB. Sharing the breakdown plus a few mitigation ideas.

## Environment

- OS: Windows 11 (NTFS; Microsoft Defender real-time protection on; OneDrive cloud-files filter active)
- Pyrefly: locally built `target/release` with debug symbols
- Workload: PyTorch repository, ~2,430 modules, ~1,323,529 lines, ~17,760 type errors
- Repro: `pyrefly check` (same engine the TSP/IDE server runs), CPU-weighted via per-sample CPU time

## Methodology

- `pyrefly check --report-timings` for phase/module timing
- `samply` (ETW, elevated) for a full CPU sampling profile
- Symbolicated `pyrefly.exe` frames against `pyrefly.pdb`; aggregated self/inclusive time using
  per-sample CPU deltas (so idle/blocked samples are not counted)

## Phase timing (`--report-timings`, thread-time)

- Solutions (type solving): ~28.4s (78%)
- Answers (bindings): ~6.8s (19%)
- Ast (parse): ~1.0s
- Exports: ~0.1s

No single pathological module — heaviest was `torch.testing._internal.common_methods_invocations`
at ~1.6s; cost is a long tail spread across all modules.

## Where the CPU actually goes (sampling self-time, CPU-weighted, ~94 CPU-seconds total)

- **~57%  Filesystem / kernel I/O** — `ntoskrnl` ~28%, `Ntfs.sys` ~12%, `ntdll` ~11%, `FLTMGR` + minifilters
- **~29%  Pyrefly type-checking compute** (solver / bindings / types) — each function <1% self; long tail
- **~8%   Allocator churn** (mimalloc: `Type` clone/drop, alloc/free)
- **~2.7% Antivirus** (Defender `mssecflt.sys` / `WdFilter.sys`)
- remainder: `memcpy`/`memset`, glob matching, path UTF-16 conversion

## Root cause

[...]

PR #4017, opened 2026-07-02
Cache `pkgutil` namespace probes during import resolution

alexander-beedie:
# Summary

* Resolving a module and its submodules walks each path component, and every intermediate package's `__init__.py` is probed with `is_pkgutil_namespace`.

* Crucially, the _same_ `__init__` files are re-opened/re-read _once per resolved submodule_, leading to a lot of redundant calls (each of which is a `File::open` + 4KiB `read` + a regex to check for `pkgutil.extend_path`).

### Solution

Memoise the result in a new `pkgutil_cache` on `DirEntryCache`, keyed by `__init__` path, sharing the same "per-loader, never-invalidated" lifetime as the existing entry cache. The existing `cache` name was updated to `entry_cache` to help differentiate/clarify now there are two caches.

### Impact

Reduces ~14,900 reads to ~1,150 when testing against the `core/homeassistant` package. (One per unique `__init__` reached, for a ~92% reduction).

Didn't see a meaningful performance impact on macOS (cached reads are cheap?). I'd expect a more noticeable positive change on Windows though (especially with corporate security software installed, which may intercept/introspect reads).

This is additive to https://github.com/facebook/pyrefly/pull/3994, with respect to https://github.com/facebook/pyrefly/issues/3993.

# Test Plan

No type-inference logic changes; tests return identical before/after test results.

### DIFF
--- a/crates/pyrefly_build/src/module_resolver.rs
+++ b/crates/pyrefly_build/src/module_resolver.rs
@@ -93,7 +93,10 @@ fn is_pkgutil_namespace(init_path: &Path, observer: Option<&dyn ModuleResolution
 /// resolution transaction or replace it when file changes are observed.
 #[derive(Default)]
 pub struct DirEntryCache {
-    cache: LockedMap<PathBuf, Option<Arc<SmallMap<OsString, bool>>>>,
+    /// Cached directory listings: maps a directory to its entries (name -> is_dir).
+    entry_cache: LockedMap<PathBuf, Option<Arc<SmallMap<OsString, bool>>>>,
+    /// Cached `pkgutil.extend_path` namespace-package check, keyed by `__init__` path.
+    pkgutil_cache: LockedMap<PathBuf, bool>,
 }
 
 impl Debug for DirEntryCache {
@@ -105,8 +108,24 @@ impl Debug for DirEntryCache {
 impl DirEntryCache {
     pub fn new() -> Self {
         Self {
-            cache: LockedMap::new(),
+            entry_cache: LockedMap::new(),
+            pkgutil_cache: LockedMap::new(),
         }
+    }
+
+    /// Cached form of [`is_pkgutil_namespace`], keyed by `__init__` path.
+    fn is_pkgutil_namespace(
+        &self,
+        init_path: &Path,
+        observer: Option<&dyn ModuleResolutionObserver>,
+    ) -> bool {
+        let key = init_path.to_path_buf();
+        if let Some(cached) = self.pkgutil_cache.get(&key) {
+            return *cached;
+        }
+        let result = is_pkgutil_namespace(init_path, observer);
+        self.pkgutil_cache.insert(key, result);
+        result
     }
 
     pub fn file_exists(&self, path: &Path) -> bool {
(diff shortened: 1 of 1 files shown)


yangdanny97:
I can try to find someone with a windows machine to do some benchmarking

alexander-beedie:
> I can try to find someone with a windows machine to do some benchmarking

It could be that the impact is being magnified in a corporate environment by tools like Defender/ThreatLocker/etc introspecting reads (according to the given environment's security policy), introducing additional overhead 🤔

alexander-beedie:
Rebased/resolved👌

yangdanny97:
@NathanTempest did you have time to benchmark this one?

NathanTempest:
**Windows benchmark: pkgutil namespace read cache**
I ran a controlled before/after on Windows, including a Defender real-time-protection on vs. off comparison to test whether the win is driven by security-software read interception.

**Environment**
-Windows 11, personal machine (SSD), consumer Microsoft Defender (no corporate EDR).
-rustc stable-msvc, --release builds.
-Corpus: [home-assistant/core](https://github.com/home-assistant/core) --depth 1 — the "core" package referenced above; one of the largest OSS Python codebases (thousands of nested integration packages ⇒ heavy __init__.py probing, which is exactly what this PR caches).

**Methodology**
1. Build both sides isolated to this single commit (so the only difference is the cache):

```
git fetch origin pull/4017/head:pr4017
git checkout pr4017    && cargo build --release --bin pyrefly   # -> pyrefly-pr.exe
git checkout pr4017^   && cargo build --release --bin pyrefly   # -> pyrefly-base.exe
```
(Confirmed the two binaries differ via SHA256.)

2. Measure with [hyperfine](https://github.com/sharkdp/hyperfine) (3 warmup + 10 timed runs each; -i because pyrefly check exits non-zero when it reports type errors, which is expected and irrelevant to timing):

```
hyperfine --warmup 3 -i ^
  "pyrefly-base.exe check homeassistant" ^
  "pyrefly-pr.exe   check homeassistant"
```
3. Defender on vs. off. To test the "security software intercepting reads" hypothesis, I ran the full benchmark under both states. Verified the state each time with:

`powershell -Command "Get-MpComputerStatus | Select-Object RealTimeProtectionEnabled"`
and toggled real-time protection via Windows Security → Virus & threat protection → Manage settings (the GUI, since Tamper Protection blocks toggling it from the CLI).

NathanTempest:
**Results**

| Defender | base | this PR | speedup | system (kernel/I/O) time |
|---|---|---|---|---|
| **on**  | 5.052 ± 0.032 s | 4.789 ± 0.027 s | **1.06×** | 3.70 s -> 2.75 s (-26%) |
| **off** | 4.944 ± 0.210 s | 4.645 ± 0.034 s | **1.06×** | 3.39 s -> 2.53 s (-26%) |

The base/off run had a first-run filesystem-cache outlier (5.508 s) that inflated σ (hyperfine flagged it); the typical run was ~4.85 s. It doesn't change the 1.06× result.

In every condition the PR's run range sat entirely below the baseline's (e.g. Defender-on: PR max 4.831 s < base min 5.013 s), so the difference is real, not variance.

**Interpretation**
Real, reproducible ~5–6% wall-clock win on Windows which is a clearer result than the macOS "no meaningful impact", consistent with Windows file I/O being more expensive.
The strongest signal is a consistent ~26% reduction in system (syscall) time  ~0.9 s  the direct fingerprint of the cache eliminating redundant __init__.py open/read calls. Wall-clock improves less than system time because the check is heavily parallel (~38 s of User time in ~5 s wall ≈ 8× parallelism), so the serialized-I/O savings are partly masked by concurrent CPU work.
The win holds with Defender off (same 1.06×, same ~26% syscall drop), and toggling Defender changed totals by only ~0.1 s. So on consumer Defender the gain is general Windows file-I/O/syscall cost, not real-time-scan interception. The author's corporate-security-software hypothesis isn't confirmed here — but a heavier corporate EDR (CrowdStrike/ThreatLocker-style) that inspects every open could plausibly amplify it well beyond this; I couldn't test that on a personal machine, so treat ~5–6% as a lower bound for the "security-heavy Windows" case.

**Recommendation**
This is a strict, never-invalidated cache keyed by __init__ path , it structurally cannot make checking slower, and it demonstrably reduces syscalls ~26% with a real (if modest) wall-clock win on Windows. LGTM for landing

samwgoldman (approved):

PR #4026, opened 2026-07-02
Don't merge nested functions into module-level ones in `pyrefly coverage`

### LINKED ISSUE #4018: `pyrefly coverage` merges nested functions into same-named module-level functions
`parse_functions` keeps module-level functions based on `exports.contains_key(name)`, so a nested
function slips through whenever a same-named module-level symbol is exported, and `merge_overloads`
then merges the two with winner-takes-all semantics:

```py
def g():  # completely untyped
    pass

def outer() -> None:
    def g(x: int) -> str:  # should be excluded
        return str(x)
```

`pyrefly coverage check` passes (but it shouldn't):

```
 INFO type coverage 100.00% (3 of 3 typable)
```

Renaming the nested `g` to something else, and `g` is now correctly reported as untyped:

```
 WARN `g` is untyped [coverage-missing]
 --> no_collision.py:1:1
  |
1 | / def g():
2 | |     pass
  | |________-
  |
```

`BindingUndecoratedFunction.outer_funcs` already marks nested functions, but `parse_functions`
never checks it.

jorenham:
# Summary

Fixes #4018

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -712,6 +712,11 @@ fn parse_functions(
         {
             let decorated = bindings.get(*x);
             let fun = bindings.get(decorated.undecorated_idx);
+            // Skip functions nested inside other functions, even when the name collides with an
+            // exported module-level function (gh-4018).
+            if fun.outer_funcs.is_some() {
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


stroxler (approved):

jorenham:
rebased onto main to resolve merge conflicts

PR #4027, opened 2026-07-02
Exclude overloaded implicit dunder returns in `pyrefly coverage`

### LINKED ISSUE #4019: `pyrefly coverage` doesn't exclude the `__init__` return slot when overloaded
Implicit dunder returns are excluded from the typable count even when annotated, but `merge_overloads` recomputes overload slots and only skips *unannotated* returns:

```py
from typing import overload

class Plain:
    def __init__(self, x: int) -> None: ...

class Overloaded:
    @overload
    def __init__(self, x: int) -> None: ...
    @overload
    def __init__(self, x: str) -> None: ...
    def __init__(self, x): ...
```

`pyrefly coverage report`:

```json
{"name": "Plain.__init__", "n_typable": 1, "n_typed": 1}
{"name": "Overloaded.__init__", "n_typable": 2, "n_typed": 2}
```

Both should have only 1 typable (the `x` parameter), because the `__init__` return type (viz. `None`) is trivial, so there's no utility in annotating it.

jorenham:
# Summary

Implicit dunder returns (e.g. `__init__(...) -> None`) are now also excluded from the typable count by `pyrefly coverage {check,report}` when the method is overloaded. 

Fixes #4019

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -194,14 +194,7 @@ fn merge_overloads(functions: &mut Vec<Function>) {
 
         for &idx in &indices {
             let func = &functions[idx];
-            let short_name = func.name.rsplit('.').next().unwrap_or(&func.name);
-            let has_annotation = func.return_annotation.is_some();
-            let ret = if !has_annotation && is_implicit_dunder_return(short_name) {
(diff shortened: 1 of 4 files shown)


yangdanny97 (approved):

PR #4028, opened 2026-07-02
Exclude module-level `__getattr__`/`__dir__` functions from `pyrefly coverage`

### LINKED ISSUE #4020: `pyrefly coverage` should not count module-level `__getattr__` and `__dir__` functions as untyped
`EXCLUDED_MODULE_DUNDERS` is only consulted for variables, not for functions. But [PEP 562](https://peps.python.org/pep-0562/) defines these module hooks *as* functions:

```py
def __getattr__(name):
    return None

def __dir__():
    return []
```

```
 WARN `__getattr__` is untyped [coverage-missing]
 --> mod_getattr.py:2:1
  |
2 | / def __getattr__(name):
3 | |     return None
  | |_______________-
  |
 WARN `__dir__` is untyped [coverage-missing]
 --> mod_getattr.py:4:1
  |
4 | / def __dir__():
5 | |     return []
  | |_____________-
  |
ERROR type coverage 40.00% (2 of 5 typable) is below the 100.00% threshold
```

But here, nothing should have been reported typable, so `pyrefly coverage check` should have passed with 0/0 typable.

A `__getattr__ = ...` assignment *is* excluded, so the exclusion depends on how the hook is
written.

jorenham:
# Summary

Fixes #4020

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -740,11 +740,12 @@ fn parse_functions(
                     }
                 }
             } else {
-                // Keep only public, exported, non-deleted module-level functions.
+                // Keep only public, exported, non-deleted, non-excluded-dunder module-level functions.
                 if !exports.contains_key(&fun.def.name.id)
                     || (!is_public_name(fun.def.name.as_str())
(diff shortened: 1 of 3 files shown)


yangdanny97 (approved):

PR #4029, opened 2026-07-02
Exclude methods and attrs from `del`eted classes in `pyrefly coverage`

### LINKED ISSUE #4021: `pyrefly coverage` reports methods and attributes of `del`eted classes
Similar to #3576, `parse_classes` skips a `del`eted class, but its methods and instance attributes stay in the report as dangling symbols:

```py
class Gone:
    def method(self, x):
        pass
    def __init__(self):
        self.attr = 1

del Gone
```

```
 WARN `Gone.method` is untyped [coverage-missing]
 --> del_class.py:2:5
  |
2 | /     def method(self, x):
3 | |         pass
  | |____________-
  |
 WARN `Gone.attr` is untyped [coverage-missing]
 --> del_class.py:5:14
  |
5 |         self.attr = 1
  |              ----
  |
ERROR type coverage 0.00% (0 of 3 typable) is below the 100.00% threshold
```

this should have passed with 0/0 typable.

jorenham:
# Summary

Fixes #4021

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -152,6 +152,11 @@ fn has_function_ancestor(parent: &NestingContext) -> bool {
     }
 }
 
+/// True if the class's top-level binding was removed by a module-scope `del`.
+fn is_deleted_class(bindings: &Bindings, cls: &ClassBinding) -> bool {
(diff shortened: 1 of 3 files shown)


stroxler (approved):

PR #4030, opened 2026-07-02
Locate suppressions at their comment's `#` in `pyrefly coverage`

### LINKED ISSUE #4022: `pyrefly coverage report` locates suppressions at the first `#` on the line
`parse_suppressions` uses `line.find('#')`, which can hit a `#` inside a string literal:

```py
x: str = "#hash"  # type: ignore
```

```json
"type_ignores": [{"kind": "type", "codes": [], "location": {"line": 1, "column": 11}}]
```

Column 11 is the `#` in `"#hash"`; the comment starts at column 19. `ignore.rs` already has a
quote-aware `find_comment_start`.

jorenham:
# Summary

`Suppression` now records the string-aware comment offset already known at parse time, so `pyrefly coverage report` no longer points a suppression at a `#` inside a string literal.

Fixes #4022

# Test Plan

Regression tests added

### DIFF
--- a/crates/pyrefly_python/src/ignore.rs
+++ b/crates/pyrefly_python/src/ignore.rs
@@ -250,21 +250,29 @@ pub struct Suppression {
     /// This may differ from the line the suppression applies to
     /// (e.g., when the comment is on the line above).
     comment_line: LineNumber,
+    /// Byte offset within `comment_line` of the `#` that starts the comment.
+    comment_offset: usize,
 }
 
 impl Suppression {
     /// A blanket suppression for `tool` that matches every error code.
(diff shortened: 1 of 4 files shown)


jorenham:
rebased onto main to resolve the merge conflicts

NathanTempest (approved):
Great fix as always @jorenham, LGTM for import. 

Recording the string-aware comment_offset at parse time (where the triple-quote state is already tracked) is the right fix, and routing coverage through module.ignore() + range_to_location is a good cleanup that also reuses the canonical parse instead of re-parsing. Bonus: dropping the old line.find('#') removes a latent byte-vs-display column bug for multibyte lines too.

stroxler (approved):

PR #4031, opened 2026-07-02
Drop the unused `incomplete_attributes` from `pyrefly coverage`

### LINKED ISSUE #4023: `pyrefly coverage` computes `incomplete_attributes` but never outputs it
No reproducer, because nothing is user-visible (that's kinda the point).

On every run, `parse_classes` fills `ReportClass.incomplete_attributes` by walking each class's MRO, with cross-module `get_class_fields` lookups and an annotation check of every method. `build_module_report` then drops all of it. The only thing a class contributes to the report is

```json
{"kind": "class", "name": "Child", "n_typable": 0, "n_typed": 0, "n_any": 0, "n_untyped": 0, "location": {"line": 4, "column": 1}}
```

and no `pyrefly coverage check` error reads it either. So the whole MRO analysis unnecessarily bloats memory (partially contributing to #3989, not sure by how much though).

Deleting it looks better than serializing it, because the data is, well, plain wrong. The inherited-field check uses `ClassFieldProperties::is_annotated` (hardcoded as `false` for `MethodLike` fields) so every inherited method is recorded as incomplete no matter how fully it is annotated.

jorenham:
# Summary

`parse_classes` no longer fills `ReportClass.incomplete_attributes` by walking each class's MRO, since `build_module_report` dropped it anyway.

Fixes #4023

# Test Plan

Tests still pass; no behavior changes

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -1009,24 +1009,6 @@ fn return_annotation_range(bindings: &Bindings, return_idx: Idx<Key>) -> Option<
     }
 }
 
-/// Only the first parameter (`self`/`cls`) is allowed to be unannotated.
-fn is_function_completely_annotated(
-    bindings: &Bindings,
-    answers: &Answers,
-    undecorated_idx: Idx<KeyUndecoratedFunction>,
-) -> bool {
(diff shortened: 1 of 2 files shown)


stroxler (approved):

jorenham:
rebased onto main and resolved the merge conflict

PR #4032, opened 2026-07-02
Report fully-`Any` symbols as `coverage-missing` in `pyrefly coverage check --strict`

### LINKED ISSUE #4024: `pyrefly coverage check --strict` labels fully-`Any` symbols as partially typed
`untyped_error` picks `coverage-missing` vs `coverage-partial` from `n_untyped` alone, ignoring strict mode:

```py
from typing import Any

def f(x: Any) -> Any: ...
```

```
 WARN `f` is not fully typed [coverage-partial]
 --> all_any.py:3:1
  |
3 | def f(x: Any) -> Any: ...
  | -------------------------
  |
ERROR strict type coverage 0.00% (0 of 2 typable) is below the 100.00% threshold
```

Under `--strict` every slot here counts as untyped, so this should be `coverage-missing`.

It's not the end of the world, but it's a bit confusing I think. And if #3919 gets merged, then this will become a bigger problem.

jorenham:
# Summary

Fixes #4024

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -1584,8 +1584,16 @@ fn is_untyped(slots: &SlotCounts, strict: bool) -> bool {
     slots.n_untyped > 0 || (strict && slots.n_any > 0)
 }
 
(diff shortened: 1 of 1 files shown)


yangdanny97 (approved):

PR #4033, opened 2026-07-02
Don't treat missing exports as origins in `pyrefly coverage`

### LINKED ISSUE #4025: `pyrefly coverage`: `trace_export_origin` returns `Some` on a missing export, contradicting its docstring
In the `pyrefly coverage --public-only` re-export tracing the docstring says "`None` on cycle/miss", but the miss arm shares its body with the this one:

```rust
match transaction.get_exports(&cur_handle).get(&cur_name) {
    Some(ExportLocation::ThisModule(_)) | None => {
        return Some(format!("{module_name}.{cur_name}"));
    }
```

So an unresolvable name (e.g. a stale `__all__` entry) is treated as defined at that module, and only cycles return `None`. 

Clearly this is very minor stuff, but it's confusing nonetheless, so it's probably better that one of the two should change.

jorenham:
# Summary

Fixes #4025

# Test Plan

Regression test added

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -336,9 +336,10 @@ fn trace_export_origin(
         }
 
         match transaction.get_exports(&cur_handle).get(&cur_name) {
-            Some(ExportLocation::ThisModule(_)) | None => {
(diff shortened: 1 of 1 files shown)


yangdanny97 (approved):

PR #4043, opened 2026-07-04
Fix TSP None type encoding

### LINKED ISSUE #4035: TSP: `None` encoded as BuiltInType name:"none", violating the documented BuiltInType.name contract
## Summary

When Pyrefly serves types over the Type Server Protocol (TSP), the Python `None` type is encoded as a `BuiltInType` with `name: "none"`. That value is not part of the documented, closed set of legal `BuiltInType.name` values, so TSP consumers (Pylance / pyright) cannot reconstruct it and fall back to `Unknown`.

**User-visible effect:** any type containing `None` — e.g. `str | None`, `Optional[X]`, an annotated `@property` returning `X | None` — renders on hover as `str | Unknown` instead of `str | None`.

## Where

`pyrefly/lib/tsp/type_conversion.rs:148`

```rust
PyreflyType::None => builtin("none"),
```

## Why this is a bug

Pyrefly's own protocol definition documents `BuiltInType.name` as a limited, known set. From `crates/tsp_types/src/protocol.rs:658`:

> The name of the built-in type. Limited to specific known built-in types. `'unknown'`, `'any'`, `'unbound'`, `'ellipsis'`, `'never'`, `'noreturn'`

`"none"` is **not** in that set. Consumers that follow the documented contract (Pylance/pyright) treat an unrecognized built-in name as `Unknown`.

Semantically, `None` in Python is an instance of `NoneType`. Pyright encodes it as a `ClassType` instance of `builtins.NoneType`, not as a `BuiltInType`. Pyrefly should do the same so unions like `str | None` round-trip correctly.

## Suggested fix

Emit `None` as a `ClassType` instance of `builtins.NoneType` instead of `builtin("none")`.

Note: several existing tests currently assert the incorrect encoding and would need updating:
- `pyrefly/lib/tsp/type_conversion.rs:1046`
- `pyrefly/lib/tsp/type_conversion.rs:1145`
- `pyrefly/lib/tsp/type_conversion.rs:1666`
- `pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs:243`

## Repro

Hover over an annotated `@property` (or any variable/return) whose type is `X | None` while using Pyrefly as the TSP type server for Pylance. Observed: `str | Unknown`. Expected: `str | None`.

## Context

Reported downstream against Pylance as microsoft/pylance-release#8097. Confirmed still present on Pyrefly `main` (dev build `1.2.0-dev.1`). Pylance is adding a defensive interop shim so hovers work regardless, but the root cause is this protocol-contract violation.

WilliamK112:
## Summary

Fixes #4035.

This changes Pyrefly's TSP conversion for `None` from an off-spec `BuiltInType` named `none` to a `ClassType` instance declared as `types.NoneType`. The protocol's `BuiltInType.name` contract only allows sentinel names such as `unknown`, `any`, `ellipsis`, `never`, and `noreturn`, so encoding `None` as a class keeps `str | None` and other unions reconstructable by TSP consumers.

The change also updates the TSP interaction coverage for `x = None` so the actual wire response is checked as a class declaration rather than a builtin sentinel.

## Test Plan

- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly tsp::type_conversion::tests:: -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly test_get_computed_type_none_is_class -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly tsp::type_conversion::tests::test_function_declaration_resolves_special_function_via_export -- --exact --nocapture`
- `RUSTUP_TOOLCHAIN=stable-x86_64-pc-windows-gnu python test.py --mode cargo --no-test --no-tensor-shapes --no-conformance --no-jsonschema`
- `git diff --check`

### DIFF
--- a/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
+++ b/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
@@ -233,14 +233,15 @@ fn test_get_computed_type_string_is_class() {
 }
 
 #[test]
-fn test_get_computed_type_none_is_builtin() {
+fn test_get_computed_type_none_is_class() {
     let (mut tsp, file_uri, snapshot) = setup_project("x = None\n");
 
     let result = get_computed_type_ok(&mut tsp, &file_uri, 0, 0, snapshot);
-    assert_kind(&result, TypeKind::Builtin);
+    assert_kind(&result, TypeKind::Class);
 
-    let name = result.get("name").and_then(|v| v.as_str());
-    assert_eq!(name, Some("none"), "Expected builtin name 'none'");
+    let declaration = result.get("declaration").expect("Expected declaration");
+    let name = declaration.get("name").and_then(|v| v.as_str());
+    assert_eq!(name, Some("NoneType"), "Expected class name 'NoneType'");
 
     tsp.shutdown();
 }
--- a/pyrefly/lib/tsp/type_conversion.rs
+++ b/pyrefly/lib/tsp/type_conversion.rs
@@ -23,9 +23,12 @@
 //!  - `Tensor`/`NNModule` → TSP `ClassType` from their base class.
 //!  - `TypeAlias` → unwraps to the aliased type.
 //!  - `SpecialForm` → TSP `BuiltInType` with the form name.
-//!  - `Any`, `Never`, `None`, `Ellipsis` → TSP `BuiltInType`.
+//!  - `Any`, `Never`, `Ellipsis` → TSP `BuiltInType`.
 //!  - Solver-internal types → TSP `BuiltInType` with a representative name.
 //!
+//! Note: `None` is emitted as a `types.NoneType` `ClassType`, not as a
+//! `BuiltInType`, because `BuiltInType.name` is restricted to protocol
+//! sentinel names.
 //! All `Type` variants are explicitly handled; no types fall through to a
 //! generic `SynthesizedType` stub.
 
@@ -145,7 +148,7 @@ impl TypeConverter<'_> {
             // --- Built-in special types ---
             PyreflyType::Any(_) => builtin("any"),
             PyreflyType::Never(_) => builtin("never"),
-            PyreflyType::None => builtin("none"),
+            PyreflyType::None => self.types_class("NoneType", TypeFlags::INSTANCE),
             PyreflyType::Ellipsis => builtin("ellipsis"),
 
             // --- Class instances (int, str, list[int], user-defined classes, etc.) ---
@@ -660,6 +663,19 @@ impl TypeConverter<'_> {
         })
     }
 
+    /// Build a TSP `ClassType` whose declaration points at `types.<name>`.
+    fn types_class(&self, name: &str, flags: TypeFlags) -> TspType {
+        TspType::Class(TspClassType {
+            declaration: Declaration::Regular(self.types_class_declaration(name)),
+            flags,
+            id: next_id(),
+            kind: TypeKind::Class,
+            literal_value: None,
+            type_alias_info: None,
+            type_args: None,
+        })
+    }
+
     /// Build a class declaration for `typing.<name>`. Resolves the real source
     /// range via the export-location resolver; falls back to a zero range
     /// pointing at bundled `typing.pyi` when unavailable.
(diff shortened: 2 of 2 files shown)


pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -145,7 +148,7 @@ impl TypeConverter<'_> {
             // --- Built-in special types ---
             PyreflyType::Any(_) => builtin("any"),
             PyreflyType::Never(_) => builtin("never"),
-            PyreflyType::None => builtin("none"),
+            PyreflyType::None => self.types_class("NoneType", TypeFlags::INSTANCE),
```
rchiodo, line 151: The behavior change here is correct — `builtin("none")` was off-spec. My design concern is *how* the target is derived: `types_class` re-derives the `NoneType` location by name (`types` + `"NoneType"`) instead of using pyrefly's authoritative `stdlib.none_type()`, which already resolves the version-correct location.

Could we instead convert the actual `stdlib.none_type()` `ClassType` through the existing `convert_class_type(..., TypeFlags::INSTANCE)` path? That makes `None` encode identically to an explicit `types.NoneType` annotation and inherits the correct module/range automatically, so there's no duplicated module/name/path logic. `TypeConverter` doesn't hold the stdlib today, but the `server.rs` caller does (it computes `get_stdlib`) and could pass the resolved class (or its qname) down. See `find_definition_for_none` in `state/lsp.rs` for the existing precedent that follows `stdlib.none_type()`.

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -682,6 +698,28 @@ impl TypeConverter<'_> {
...
     }
 
+    /// Build a class declaration for `types.<name>`. Resolves the real source
+    /// range via the export-location resolver; falls back to a zero range
+    /// pointing at bundled `types.pyi` when unavailable.
+    fn types_class_declaration(&self, name: &str) -> RegularDeclaration {
+        let symbol = Name::new(name);
+        if let Some((module_path, lsp_range)) = self
+            .resolve_export
+            .and_then(|resolve| resolve(ModuleName::types(), &symbol))
```
rchiodo, line 708: Concrete version-correctness issue with hardcoding `ModuleName::types()`: `stdlib.none_type()` resolves `NoneType` from `types` only on Python 3.10+, and from `_typeshed` on older versions (see `none_location` in `crates/pyrefly_types/src/stdlib.rs`). On a project targeting < 3.10 this lookup points at a different module than pyrefly's own `None` resolution / goto-definition, so the TSP declaration and the in-editor "go to definition" for `None` would disagree. Deriving the location from `stdlib.none_type()`'s qname avoids the divergence.

pyrefly/lib/tsp/type_conversion.rs
```diff
@@ -818,6 +856,21 @@ fn make_typing_class_declaration(name: &str) -> RegularDeclaration {
     }
 }
 
+/// Build a declaration for a class in `types.pyi`.
+fn make_types_class_declaration(name: &str) -> RegularDeclaration {
+    let module_path =
+        pyrefly_python::module_path::ModulePath::bundled_typeshed(PathBuf::from("types.pyi"));
```
rchiodo, line 862: Minor, same caveat as the resolver path: this fallback hardcodes `types.pyi`, which is only correct for 3.10+. It only fires when no export resolver is available (unit tests), so impact is low — but if the location is derived from `stdlib.none_type()` this hardcoded file (and the whole helper) goes away.

pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
```diff
@@ -233,14 +233,15 @@ fn test_get_computed_type_string_is_class() {
...
 
     let result = get_computed_type_ok(&mut tsp, &file_uri, 0, 0, snapshot);
-    assert_kind(&result, TypeKind::Builtin);
+    assert_kind(&result, TypeKind::Class);
 
-    let name = result.get("name").and_then(|v| v.as_str());
-    assert_eq!(name, Some("none"), "Expected builtin name 'none'");
+    let declaration = result.get("declaration").expect("Expected declaration");
+    let name = declaration.get("name").and_then(|v| v.as_str());
+    assert_eq!(name, Some("NoneType"), "Expected class name 'NoneType'");
```
rchiodo, line 244: This (and the unit tests) verify the wire shape — `Class` + name `NoneType` — but not the actual symptom from #4035: that Pylance now renders `str | None` instead of `str | Unknown`. The issue notes the consumer models this as `builtins.NoneType`; since we're emitting `types.NoneType`, it'd be reassuring to confirm end-to-end against a real Pylance client that the hover round-trips (i.e. that Pylance keys off the class name rather than the `builtins` module).

rchiodo (commented):
Thanks for tackling this — the core change (stop emitting the off-spec `none` builtin and encode `None` as a `NoneType` `ClassType` instead) is the right fix for #4035.

My main concern is that the `NoneType` location is re-derived by name and hardcoded to the `types` module, which diverges from pyrefly's authoritative, version-aware `stdlib.none_type()` on Python < 3.10. I'd suggest routing `None` through the real `stdlib.none_type()` class so the encoding is version-correct and identical to an explicit `types.NoneType` annotation. Details inline. It would also be worth an end-to-end Pylance check to confirm the hover symptom actually round-trips.

kinto0:
thanks!! I'll try merging this today. if you don't have a chance to update it, I'll resolve the comments in a follow-up

kinto0:
doing that now. also fixing a few other sentinel types that had the same issue: bool, int, Sentinel

stroxler (approved):

PR #4044, opened 2026-07-04
Fix Unimported directive error kind unhandled on lsp

### LINKED ISSUE #3859: reveal_type don't have code action to import
### Describe the Bug

Image

Image

### Sandbox Link

_No response_

v1.1.0

aodihis:
### Summary

Fixes #3859 
The issue #3859 is caused by an unhandled `UnimportedDirective` error kind. If we look at `pyrefly/lib/state/lsp.rs`, Pyrefly only handled the import action for the `UnknownName` error kind.

This PR updates the existing import quick-fix path so `UnimportedDirective` is handled alongside `UnknownName`.

### Test Plan

Added coverage for the scenario where unimported directives such as `reveal_type` and `assert_type` should produce an import quick fix from `typing`.


yangdanny97 (approved):

PR #4045, opened 2026-07-04
Fix TSP inferred returns for definition-site computed types

### LINKED ISSUE #4042: TSP: inferred return type dropped at definition-site requests (unannotated functions / methods / @property getters render `-> Unknown`)
## Summary

When Pyrefly serves a type over TSP for a **definition-site** request — a function `def`, a method `def`, or an `@property` getter — the function's **inferred return type is omitted**: the wire `Function` carries `return_type = Unknown`. The *same* function requested at a **reference / call site** correctly carries its inferred return type. As a result, hovering an unannotated `@property` renders `... -> Unknown` (originally reported as `Any` on `1.1.1`) instead of the inferred type.

**No protocol change is required.** The existing `Function.return_type` field already conveys inferred returns — reference sites prove it (see the table below). Only the definition-site conversion path fails to populate it.

## User-visible effect

Reported downstream as microsoft/pylance-release#8096. Hovering an unannotated `@property` getter (e.g. `requests.Response.ok`, `apparent_encoding`):

- Expected (Pylance built-in Pyright): `(property) ok: (self: Self@Response) -> bool`
- Actual (Pyrefly over TSP): `(property) ok: (self: Self@Response) -> Unknown`

## Evidence (Pyrefly `1.2.0-dev.1` as the TSP server)

Hovering these constructs and observing the type Pyrefly sends over TSP:

| Hover target | Pyrefly TSP output |
| --- | --- |
| plain `f` **reference** — `def f(): return True` then a bare `f` | `(function) def f() -> Literal[True]` ✅ |
| plain `def f` **definition** | `(function) f: Unknown` ❌ |
| `def m(self)` **method definition** | `(method) def m(self: Self@C) -> Unknown` ❌ |
| `@property def ok(self)` **getter** | `(property) ok: (self: Self@C) -> Unknown` ❌ |

Pyrefly's own checker infers all of these correctly, so the type is available and is lost only in the definition-site TSP conversion:

```
reveal_type(c.ok)                # revealed type: Literal[True]
reveal_type(c.apparent_encoding) # revealed type: Literal['utf-8']
```

## Why this confirms no protocol change is needed

- The protocol already models function return types: `crates/tsp_types/src/protocol.rs` — `FunctionType.return_type: Option<Box<Type>>`.
- Pyrefly already populates it: `pyrefly/lib/tsp/type_conversion.rs` — `convert_function(...)` sets `return_type: Some(Box::new(ret))`.
- The **reference-site** row above shows this working end-to-end (`-> Literal[True]`), using the current protocol only.

The gap is specifically the **definition-site** path. In `pyrefly/lib/state/lsp.rs`, `get_type_at_impl_with_options` resolves a use-site via `Key::BoundName` (the fully-solved function whose `.ret` is inferred), but resolves a definition-site (`FunctionDef` / `MethodDef` identifier contexts) via `Key::Definition` → `get_type_for_surface` → `get_type_at(idx)`, which yields the function type before return-type inference is reflected. The `@property` getter surfaces the same definition-site behavior, which is what the downstream hover renders.

## Suggested fix

When resolving a definition-site type for a function / method / property-getter, produce the same solved type the reference-site path yields (i.e. ensure the inferred return type is computed and attached) before converting to the TSP `Function`. No changes to `tsp_types` or the wire protocol are needed.

## Context

- Downstream: microsoft/pylance-release#8096 — found via differential testing comparing Pylance's built-in Pyright analyzer against Pyrefly running as an external type server over TSP.
- Sibling TSP-conversion report: facebook/pyrefly#4035 (`None` encoding).
- Confirmed still present on Pyrefly `main` (dev build `1.2.0-dev.1`).

WilliamK112:
Fixes #4042.

## Summary
- Treat exact TSP range requests for function and method definition names like identifier lookups instead of expression traces.
- Preserve declaration-based computed types so unannotated functions, methods, and property getters keep their inferred `returnType` at definition sites.
- Add TSP interaction coverage for definition-site ranges plus a same-name reference range under `untyped-def-behavior = check-and-infer-return-type`.

## Validation
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly test_get_computed_type_definition_site_uses_inferred_return_type -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly get_type_queries`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly tsp_interaction`
- `git diff --check`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly test_shape_dsl_reports_multiple_errors -- --nocapture`

I also attempted the full `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly` suite. It reached `6188 passed, 1 failed`; the only failure was the unrelated `test::shape_dsl::test_shape_dsl_reports_multiple_errors` internal watchdog timeout (`Error: Test took too long (> 20s)`) during the full parallel run. The same test passed when rerun directly, as shown above.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -1248,15 +1248,26 @@ impl<'a> Transaction<'a> {
     ///
     /// TSP prefers raw bound types of identifiers since it re-resolves declarations.
     pub fn get_computed_type_at_range(&self, handle: &Handle, range: TextRange) -> Option<Type> {
-        // Bare names need declaration-preserving lookup to avoid coercing
-        // overloaded functions into synthesized callables.
-        if range.is_empty()
-            || self.get_ast(handle).is_some_and(|module| {
-                Ast::locate_node(&module, range.start())
-                    .into_iter()
-                    .any(|node| node.range() == range && matches!(node, AnyNodeRef::ExprName(_)))
-            })
-        {
+        // Bare names and definition-name ranges need declaration-preserving
+        // lookup to avoid coercing overloaded functions into synthesized
+        // callables, and to preserve inferred returns for unannotated defs.
+        let is_identifier_name_range = !range.is_empty()
+            && self.identifier_at(handle, range.start()).is_some_and(
+                |IdentifierWithContext {
+                     identifier,
+                     context,
+                 }| {
+                    identifier.range == range
+                        && matches!(
+                            context,
+                            IdentifierContext::Expr(_)
+                                | IdentifierContext::FunctionDef { .. }
+                                | IdentifierContext::MethodDef { .. }
+                        )
+                },
+            );
+
+        if range.is_empty() || is_identifier_name_range {
             return self.get_type_at_preserving_declaration(handle, range.start());
         }
         self.get_type_trace(handle, range)
--- a/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
+++ b/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
@@ -20,7 +20,26 @@ use crate::test::tsp::tsp_interaction::object_model::write_pyproject;
 fn setup_project(file_content: &str) -> (TspInteraction, String, i32) {
     let temp_dir = TempDir::new().unwrap();
     write_pyproject(temp_dir.path());
+    setup_project_in_dir(temp_dir, file_content)
+}
 
+fn setup_project_with_pyrefly_config(
+    file_content: &str,
+    pyrefly_config: &str,
+) -> (TspInteraction, String, i32) {
+    let temp_dir = TempDir::new().unwrap();
+    write_pyproject(temp_dir.path());
+
+    let pyproject = temp_dir.path().join("pyproject.toml");
+    let mut content = std::fs::read_to_string(&pyproject).unwrap();
+    content.push_str("\n[tool.pyrefly]\n");
+    content.push_str(pyrefly_config);
+    std::fs::write(pyproject, content).unwrap();
+
+    setup_project_in_dir(temp_dir, file_content)
+}
+
+fn setup_project_in_dir(temp_dir: TempDir, file_content: &str) -> (TspInteraction, String, i32) {
     let test_file = temp_dir.path().join("main.py");
     std::fs::write(&test_file, file_content).unwrap();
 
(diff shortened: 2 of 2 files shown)


pyrefly/lib/state/lsp.rs
```diff
@@ -1248,15 +1248,26 @@ impl<'a> Transaction<'a> {
...
-        if range.is_empty()
-            || self.get_ast(handle).is_some_and(|module| {
-                Ast::locate_node(&module, range.start())
-                    .into_iter()
-                    .any(|node| node.range() == range && matches!(node, AnyNodeRef::ExprName(_)))
-            })
-        {
+        // Bare names and definition-name ranges need declaration-preserving
+        // lookup to avoid coercing overloaded functions into synthesized
+        // callables, and to preserve inferred returns for unannotated defs.
```
rchiodo, line 1253: The linked issue diagnoses `Key::Definition` as yielding the *pre-inference* function type, yet this fix relies on `get_type_at_preserving_declaration` -> the `FunctionDef`/`MethodDef` arm of `get_type_at_impl_with_options` -> `Key::Definition` to actually *carry* the inferred return. So the effective change is `get_type_trace` (the stored trace type) vs. the `Key::Definition` binding. Since that distinction is the crux of the fix, could you add a sentence here (or in the commit message) spelling out the real mechanism? It'll save future readers who go back to the issue and see the seemingly-contradictory claim about `Key::Definition`.

pyrefly/lib/state/lsp.rs
```diff
@@ -1248,15 +1248,26 @@ impl<'a> Transaction<'a> {
...
+                |IdentifierWithContext {
+                     identifier,
+                     context,
+                 }| {
+                    identifier.range == range
+                        && matches!(
+                            context,
+                            IdentifierContext::Expr(_)
+                                | IdentifierContext::FunctionDef { .. }
+                                | IdentifierContext::MethodDef { .. }
```
rchiodo, line 1265: `ClassDef` is intentionally excluded here, but it reads like it could be an oversight. Could you add a one-line note explaining why? My understanding is class def-sites don't have the inferred-return problem and their `get_type_trace` type is already correct, so they don't need the declaration-preserving path — but a comment would make that explicit.

pyrefly/lib/state/lsp.rs
```diff
@@ -1248,15 +1248,26 @@ impl<'a> Transaction<'a> {
...
-                Ast::locate_node(&module, range.start())
-                    .into_iter()
-                    .any(|node| node.range() == range && matches!(node, AnyNodeRef::ExprName(_)))
-            })
-        {
+        // Bare names and definition-name ranges need declaration-preserving
+        // lookup to avoid coercing overloaded functions into synthesized
+        // callables, and to preserve inferred returns for unannotated defs.
+        let is_identifier_name_range = !range.is_empty()
+            && self.identifier_at(handle, range.start()).is_some_and(
```
rchiodo, line 1255: Nit (optional, non-blocking): `identifier_at` runs here and then again inside `get_type_at_preserving_declaration`, so the identifier is located twice per call. This isn't introduced by the PR (the old code also walked the AST twice), so fine to leave as-is — just flagging in case a variant that accepts the already-resolved identifier feels worthwhile.

rchiodo (commented):
Nice, focused fix. A few clarifying questions/nits inline — none blocking.

kinto0:
thanks for the PRs! I think [this ](https://github.com/facebook/pyrefly/pull/4091) PR (though much bigger) gets more at the root problem. I'll merge it for now and try to merge #4043 today too (will comment there)

PR #4048, opened 2026-07-05
Fix public-only coverage for excluded re-export origins

### LINKED ISSUE #4034: `pyrefly coverage check --public-only`: `--project-excludes` shouldn't exclude public re-exports
Here's a project I have

`foo/__init__.py`
```py
from foo._something_private import bar

__all__ = ['bar']
```

`foo/_something_private.py`
```py
def bar(a) -> None:
    return None
```

So, `foo._something_private.bar` is defined in a private file, but it's re-exported in a public place (`foo/__init__.py`).

And indeed, `pyrefly coverage check foo --public-only` correctly checks `foo._something_private.bar`

```console
$ pyrefly coverage check foo --public-only
 WARN `foo._something_private.bar` is not fully typed [coverage-partial]
 --> foo/_something_private.py:1:1
  |
1 | / def bar(a) -> None:
2 | |     return None
  | |_______________-
  |
```

But, here's the issue: if I use `--project-excludes=foo/_something_private.py`, then `foo._something_private.bar` gets excluded, even though it's re-exported in a public place:
```console
$ pyrefly coverage check foo --public-only --project-excludes foo/_something_private.py
 INFO type coverage 100.00% (0 of 0 typable)
```

I'd like to suggest one of the following as a solution:
- `--project-excludes` doesn't exclude publicly re-exported symbols. I think I'd prefer this solution
- or, `pyrefly coverage report` shows all the re-exported paths, so then I can parse these as a user as I wish

Note that pyright shows `foo.bar` as the path for this symbol:
```json
            {
                "category": "function",
                "name": "foo.bar",
                "referenceCount": 1,
                "isExported": true,
                "isTypeKnown": false,
                "isTypeAmbiguous": false,
```

WilliamK112:
Summary:
- keep project-local origins of public re-exports in `coverage check --public-only` even when the origin file is matched by `--project-excludes`
- preserve the inverse case: if the public re-export location itself is excluded, the private origin is not counted as public
- add production-shape coverage tests that exercise file-args mode, where package `__init__.py` may be loaded as `__unknown__`

Closes #4034.

Duplicate/ownership checks:
- #4034 is open and unassigned
- `gh search prs` for #4034 / public re-export / project-excludes / public-only returned no open duplicates
- `gh pr list --search 4034` only returned unrelated #3597

Validation:
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib commands::coverage::collect::tests::test_public_only -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib commands::coverage::collect::tests::test_compute_public_fqns -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib coverage -- --nocapture` (65 passed)
- real temp-project smoke: `pyrefly coverage check foo --public-only --project-excludes foo/_something_private.py` now reports `foo._something_private.bar` as `coverage-partial` and exits 1 at 50% coverage instead of reporting `100.00% (0 of 0 typable)`
- `git diff --check`

I also ran `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib`; it reached 6185 passed before 5 existing shape_dsl tests hit their internal `> 20s` watchdog under the full Windows run. Each of those five tests passed when rerun individually.

### DIFF
--- a/pyrefly/lib/commands/coverage/collect.rs
+++ b/pyrefly/lib/commands/coverage/collect.rs
@@ -19,6 +19,7 @@ use pyrefly_python::dunder;
 use pyrefly_python::ignore::Ignore;
 use pyrefly_python::module::Module;
 use pyrefly_python::module_name::ModuleName;
+use pyrefly_python::module_path::ModulePathDetails;
 use pyrefly_python::module_path::ModuleStyle;
 use pyrefly_python::nesting_context::NestingContext;
 use pyrefly_python::short_identifier::ShortIdentifier;
@@ -325,7 +326,7 @@ fn trace_export_origin(
     handle: &Handle,
     mut cur_name: Name,
     transaction: &Transaction,
-) -> Option<String> {
+) -> Option<(Handle, String)> {
     let mut seen = SmallSet::new();
     let mut cur_handle = handle.clone();
 
@@ -337,7 +338,7 @@ fn trace_export_origin(
 
         match transaction.get_exports(&cur_handle).get(&cur_name) {
             Some(ExportLocation::ThisModule(_)) | None => {
-                return Some(format!("{module_name}.{cur_name}"));
+                return Some((cur_handle, format!("{module_name}.{cur_name}")));
             }
             Some(ExportLocation::OtherModule(other_module, alias)) => {
                 if let Some(alias) = alias {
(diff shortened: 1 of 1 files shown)


yangdanny97:
@jorenham could use your eyes on this

pyrefly/lib/commands/coverage/collect.rs
```diff
@@ -351,39 +352,118 @@ fn trace_export_origin(
...
+                || contains_handle(&origin_handles, &origin_handle)
+                || !is_project_reexport_origin(handle, &origin_handle, config_finder)
+            {
+                continue;
+            }
+            origin_handles.push(origin_handle);
+        }
+    }
+    origin_handles
+}
```
jorenham, line 450: This iterates the same public modules as `compute_public_fqns` below, applies the same `EXCLUDED_MODULE_DUNDERS` filter, and traces the same export. You could produce both the origin-handle list and the FQN set in a single pass, and avoid this introduced code duplication.

pyrefly/lib/commands/coverage/collect.rs
```diff
@@ -351,39 +352,118 @@ fn trace_export_origin(
...
+    }
+}
+
+fn same_handle(left: &Handle, right: &Handle) -> bool {
+    left.module() == right.module() && left.path() == right.path()
+}
+
+fn contains_handle(handles: &[Handle], candidate: &Handle) -> bool {
+    handles.iter().any(|handle| same_handle(handle, candidate))
+}
```
jorenham, line 380: `Handle` already derives `PartialEq`, `Eq`, and `Hash`, so these are not needed, and even hurt performance (+ O(n^2)). Assuming that you ignored `sys_info` by mistake, then both functions can be removed and a `SmallSet` of `Handle` could be used instead.

pyrefly/lib/commands/coverage/collect.rs
```diff
@@ -351,39 +352,118 @@ fn trace_export_origin(
...
+        public_handle.path().root_of(public_handle.module()),
+        origin_handle.path().root_of(origin_handle.module()),
+    ) {
+        return public_root == origin_root;
+    }
+
+    config
+        .search_path()
+        .any(|search_root| origin_path.starts_with(search_root))
+}
```
jorenham, line 423: There are a whole bunch of unmotivated heuristics here: a bundle check, path-details matching, site-package rejection, `__unknown__` special-casing, parent-dir special-casing, double `root_of` comparisons, and then a search-path fallback. 

Did you verify that each of these heuristics is actually needed here? If so, then please explain why exactly that is (as a comment). If not, then it should be removed.

pyrefly/lib/commands/coverage/collect.rs
```diff
@@ -1658,23 +1738,41 @@ pub fn collect_module_reports(
...
+    } else {
+        Vec::new()
+    };
+    let public_fqns = if public_only {
+        for origin_handle in public_origin_handles.iter().cloned() {
+            if !contains_handle(&report_handles, &origin_handle) {
+                report_handles.push(origin_handle);
+            }
+        }
+        if report_handles.len() != handles.len() {
```
jorenham, line 1753:
```suggestion
        if !public_origin_handles.is_empty() {
```

jorenham (commented):
I checked that this doesn't affect numpy of scipy-stubs when `--project-excludes` isn't passed, so at least it doesn't regress in that way. 

There are, however, some opportunities to improve code quality and reduce code duplication here. 

I also am not sure if all the heuristics in `is_project_reexport_origin` are a good idea, and there's no motivation explaining why it would be. It, for example, returns `false` when for site-packages, which could be problematic if you're running `pyrefly coverage` directly on an installed library in your `.venv`.

yangdanny97:
looks like the original issue was closed by https://github.com/facebook/pyrefly/pull/4133

PR #4049, opened 2026-07-05
Fix hover names from quoted definition snippets

### LINKED ISSUE #4004: numpy array 's signature is a bit strange
### Describe the Bug

```
(method) 'array': def 'array'(
    object: Any,
    dtype: _HasDType[dtype] | _HasNumPyDType[dtype] | dtype | list[Any] | str | _DTypeDict | tuple[Any, Any] | type[Any] | None = None,
    *,
    copy: _CopyMode | bool | None = True,
    order: _OrderKACF = 'K',
    subok: bool = False,
    ndmin: int = 0,
    ndmax: int = 0,
    like: _SupportsArrayFunc | None = None
) -> ndarray: ...
```

what is def 'array'

### Sandbox Link

_No response_

_No response_

WilliamK112:
Summary:
- avoid using quoted string-literal definition snippets as hover display names
- prefer the semantic display name or callable metadata fallback, so signatures render as `def array(...)` instead of `def 'array'(...)`
- keep normal source-definition snippets unchanged and cover the quoted-name path with regression tests

Fixes #4004.

Test plan:
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib lsp::wasm::hover::tests -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --test pyrefly_lsp_interaction_tests hover -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu check -p pyrefly`
- `git diff --check`

### DIFF
--- a/pyrefly/lib/lsp/wasm/hover.rs
+++ b/pyrefly/lib/lsp/wasm/hover.rs
@@ -421,6 +421,61 @@ fn fallback_hover_name_from_type(type_: &Type) -> Option<String> {
     None
 }
 
+fn simple_python_string_literal_contents(snippet: &str) -> Option<&str> {
+    let mut rest = snippet.trim();
+    while let Some(ch) = rest.chars().next()
+        && matches!(ch, 'b' | 'B' | 'f' | 'F' | 'r' | 'R' | 'u' | 'U')
+    {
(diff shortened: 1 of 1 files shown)


stroxler (approved):

PR #4050, opened 2026-07-05
Add PEP 610 editable import roots

### LINKED ISSUE #1843: Discovering local projects installed in editable mode PEP610
### Describe the Bug

When using pyrefly in EMACS with Eglot pyrefly cannot discover locally installed packages with modern pip (PEP610) that creates direct_url.json to resolve the path. However, if I add the package path to easy-install.pth file in site-packages it is discovered

### Sandbox Link

_No response_

_No response_

WilliamK112:
## Summary

- extend the interpreter environment query to scan `.dist-info/direct_url.json` metadata for PEP 610 editable installs
- append editable `file://` source roots, plus a conventional `src/` child when present, to interpreter-derived import paths
- keep the scan bounded and tolerant of missing, invalid, non-editable, or non-file direct_url metadata
- add a Windows-runnable regression test that executes the same interpreter query script with a synthetic editable install

Closes #1843.

## Test plan

- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly_config environment::environment -- --nocapture`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly_config`
- `cargo +stable-x86_64-pc-windows-gnu check -p pyrefly`
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --test pyrefly_lsp_interaction_tests configuration -- --nocapture`
- CLI smoke: flat-layout editable package with only `direct_url.json` metadata now checks with 0 errors
- CLI smoke: `src/` layout editable package with only `direct_url.json` metadata now checks with 0 errors

### DIFF
--- a/crates/pyrefly_config/src/environment/environment.rs
+++ b/crates/pyrefly_config/src/environment/environment.rs
@@ -30,6 +30,96 @@ static INTERPRETER_ENV_REGISTRY: LazyLock<
     Mutex<SmallMap<PathBuf, Result<PythonEnvironment, String>>>,
 > = LazyLock::new(|| Mutex::new(SmallMap::new()));
 
+const INTERPRETER_ENV_SCRIPT: &str = r#"
+import json
+import os
+import sys
+import sysconfig
+from pathlib import Path
+from urllib.parse import urlparse
+from urllib.request import url2pathname
+
+platform = sys.platform
+v = sys.version_info
+version = '{}.{}.{}'.format(v.major, v.minor, v.micro)
+stdlib_paths = [p for p in [sysconfig.get_path('stdlib')] if p is not None]
+site_package_path = [
+    p
+    for p in sys.path
+    if p != ''
+    and '.zip' not in p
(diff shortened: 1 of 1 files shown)


grievejia:
Thanks for working on this! After comparing Pyrefly’s behavior with mypy, Pyright, and ty, I don’t think deriving import roots from `direct_url.json` is the right direction.

PEP 610 records distribution provenance; it does not define import roots. Adding both the recorded directory and a guessed src/ directory can expose modules that the editable installation does not actually provide, and it does not correctly model custom or strict editable mappings.

The other type checkers support path-based `.pth` files and intentionally do not support executable `.pth` import hooks. Pyrefly already handles path-based `.pth` entries when they appear in the selected interpreter’s `sys.path`. I would prefer we follow that model. A useful follow-up would be static path-only `.pth` processing for explicitly configured site-package-path directories, while ignoring executable lines.

I’m going to close this PR, but I appreciate the investigation and implementation effort!

PR #4051, opened 2026-07-05
fix f-str check #4010

asukaminato0721:
# Summary

Fixes part of #4010

# Test Plan

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -51,6 +51,7 @@ use pyrefly_util::visit::Visit;
 use ruff_python_ast::Arguments;
 use ruff_python_ast::BoolOp;
 use ruff_python_ast::Comprehension;
+use ruff_python_ast::ConversionFlag;
 use ruff_python_ast::DictItem;
 use ruff_python_ast::Expr;
 use ruff_python_ast::ExprAttribute;
@@ -63,7 +64,10 @@ use ruff_python_ast::ExprSlice;
 use ruff_python_ast::ExprStarred;
 use ruff_python_ast::ExprStringLiteral;
 use ruff_python_ast::ExprTuple;
+use ruff_python_ast::FStringPart;
 use ruff_python_ast::Identifier;
+use ruff_python_ast::InterpolatedStringElement;
+use ruff_python_ast::InterpolatedStringElements;
 use ruff_python_ast::Keyword;
 use ruff_python_ast::Number;
 use ruff_python_ast::Operator;
@@ -220,6 +224,14 @@ enum ExprExpectation<'a, 'b, 'subset> {
         context: &'a dyn Fn() -> TypeCheckContext,
         call_context: Option<&'a CallContext<'subset>>,
     },
+}
+
+#[derive(Debug, Clone, Copy)]
+enum FStringFormatPresentation {
+    Any,
+    Int,
+    Numeric,
+    IntOrStr,
 }
 
 impl<'a, 'b, 'subset> ExprOptions<'a, 'b, 'subset> {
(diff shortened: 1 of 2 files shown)


stroxler (changes requested):

stroxler:
I know this is still a draft, but our infra imports it anyway so it wound up in my review queue :/

Just figured I might as well share the AI review notes:

# 1. . 'c' format type classified as IntOrStr (line 617)
In Python, the 'c' format code only works for integers (converts an integer to its Unicode character). format('a', 'c') raises ValueError. Classifying it as FStringFormatPresentation::IntOrStr means f"{'hello':c}" won't be flagged, though it errors at runtime. Should this be FStringFormatPresentation::Int?

# 2. Missing z flag in format spec parser (line 588)
Python 3.11+ added the z flag (coerce negative zero) to the format spec mini-language, positioned between sign and #:
```
[[fill]align][sign][z][#][0][width][grouping][.precision][type]
```
The parser currently goes from sign handling directly to #. Without handling z, valid specs like f"{x:z.2f}" would produce a false-positive error (the parser returns Err). Consider adding:
```Rust
if chars.get(i) == Some(&'z') {
    i += 1;
}
``
before the # check on line 588.

# 3. Minor: bytes in valid_builtin (line 512)

bytes objects don't support non-empty format specs in Python (format(b'hello', '>10') raises TypeError). Including it in valid_builtin means these won't be caught. This is a conservative/permissive choice, so not a blocker — just worth noting.

File: pyrefly/lib/alt/expr.rs [hunk 1]
```diff
@@ -400,6 +411,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+        errors: &ErrorCollector,
+    ) {
+        if actual.is_any() || actual.is_error() || actual.is_never() {
+            return;
+        }
+        if self.check_datetime_fstring_format_spec(actual, spec, range, errors) {
+            return;
+        }
+        let valid_builtin = self.unions(vec![
+            self.heap.mk_class_type(self.stdlib.bytes().clone()),
```
grievejia: `bytes.__format__` rejects every non-empty format specification:

```python
format(b"xy", ">10")  # TypeError
format(b"xy", "s")    # TypeError
```

The empty format case has already returned before this point. Please remove `bytes` from this supported set. It will then use the existing unsupported-format diagnostic.

pyrefly/lib/alt/expr.rs
```diff
@@ -400,6 +411,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            self.heap.mk_class_type(self.stdlib.str().clone()),
+            self.heap.mk_class_type(self.stdlib.int().clone()),
+            self.heap.mk_class_type(self.stdlib.float().clone()),
+            self.heap.mk_class_type(self.stdlib.complex().clone()),
+        ]);
+        if self.is_subset_eq(actual, &valid_builtin) {
```
grievejia, line 549: This branch applies builtin rules to subclasses before it checks whether the subclass replaced `__format__`. This rejects valid code:

```python
class FancyInt(int):
    def __format__(self, spec: str) -> str:
        return "fancy:" + spec

f"{FancyInt(3):whatever}"
```

Please find the class that supplies `__format__`. Apply the builtin rules only when the builtin implementation is used.

pyrefly/lib/alt/expr.rs
```diff
@@ -400,6 +411,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                return Err(format!("Unrecognized format specification `{spec}`"));
+            }
+        }
+        let presentation = match chars.get(i) {
+            None => return Ok(FStringFormatPresentation::Any),
+            Some(c) if i + 1 == chars.len() => *c,
+            _ => return Err(format!("Unrecognized format specification `{spec}`")),
+        };
+        match presentation {
+            's' => Ok(FStringFormatPresentation::Any),
```
grievejia: The parsed result loses information that is needed for type-specific checks. It treats no presentation type and the `s` presentation type as the same case, and it does not keep the flags. As a result, these runtime errors are accepted:

```python
s = "s"
f"{1:s}"
f"{s:+}"
f"{s:,}"
f"{1:z}"
```

Please keep the presentation type and flags in the parsed result, then validate them for each builtin type.

pyrefly/lib/alt/expr.rs
```diff
@@ -400,6 +411,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            if char == '%' && !percent {
+                percent = true;
+                continue;
+            }
+            if percent {
+                if char == ':' {
+                    colon = true;
+                    continue;
+                }
+                if !"aAwdbBmyYHIpMSfzZjUWcxX%GuV".contains(char) {
```
grievejia: This rejects valid date format strings on supported platforms. Python passes the format string to the operating system. For example, these calls work on Linux and macOS:

```python
from datetime import date

format(date(2024, 3, 5), "%-d/%-m")  # "5/3"
format(date(2024, 3, 5), "%")        # "%"
```

The current allowlist reports both formats as invalid. Please do not treat a platform-specific directive as invalid on every platform. Either remove this validation or define an explicit target-platform policy.

pyrefly/lib/alt/expr.rs
```diff
@@ -400,6 +411,314 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            )
+            .emit();
+    }
+
+    fn type_has_custom_format(&self, ty: &Type) -> bool {
+        self.type_class_for_fstring_format(ty)
+            .is_some_and(|cls| self.class_has_format(&cls))
+    }
+
+    fn class_has_format(&self, cls: &ClassType) -> bool {
```
grievejia: Please treat an unknown base class as able to provide `__format__`. For example:

```python
class Foo(SomeUntypedBase):
    pass

f"{Foo():>10}"
```

The type checker does not know whether the base supplies `__format__`, so it cannot prove that this format is unsupported. Please avoid this extra error when inheritance is unknown. The same principle should prevent a second formatting error when the class inheritance order cannot be resolved.

grievejia (changes requested):

PR #4054, opened 2026-07-05
fix hover on named parameter is missing info #3238

### LINKED ISSUE #3238: hover on named parameter is missing info
### Describe the Bug

Image

hover on mapped_column, we get this panel.

but 

Image

this not

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3238

Now keyword-argument hover can resolve parameter types from the defining module, not just the caller module,
 fall back to the selected call signature when definition refinement is unavailable.

# Test Plan

add a test && update a todo

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -504,14 +504,15 @@ pub(crate) struct IdentifierWithContext {
 enum ResolutionKind {
     /// Resolve through a binding key — a reference or a declaration.
     Key(Key),
+    /// Resolve through a binding key in a different module.
+    KeyInModule(Handle, Key),
     /// A directly-constructed type with no binding key. Only module identifiers.
     Type(Type),
+    /// The active parameter type at a call argument position.
+    ActiveCallArgument(TextSize),
     /// Member access (a computed expression, not a declaration): resolve via the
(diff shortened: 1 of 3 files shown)


NathanTempest:
Thanks @asukaminato0721 !

The new hover_shows_type_for_imported_keyword_argument test and the kwarg_with_overload improvement (None → int) are great.

One blocker before this can land.  The keyword-argument hover resolution in state/lsp.rs was refactored on main since you branched, it now goes through classify_surface → keyword_argument_key → ResolutionKind rather than the direct find_definition_for_keyword_argument().first().and_then(...) match arm this patch edits, so the main hunk no longer applies. Could you rebase onto latest main and re-express the fix on the new structure? 

Also: keyword_argument_key is where the definition Key should be resolved against the defining module's handle, and the call-signature fallback can hang off the KeywordArgument handling. The get_expected_type_at → get_active_call_argument_type_for_surface extraction still applies cleanly and is a nice cleanup to keep.

NathanTempest (approved):
Thanks @asukaminato0721 for the clean rework on the current classify_surface/ResolutionKind structure. Importing!

stroxler (approved):

PR #4061, opened 2026-07-06
fix Unable to see generic attributes on abstract classes #4058

### LINKED ISSUE #4058: Unable to see generic attributes on abstract classes
### Describe the Bug

When an abstract defines generics as attributes that are used in classmethods, each use of the generic attribute raises the following error:

> Generic attribute `db_model` of class `SimpleCrud` is not visible on the class (missing-attribute)

For example:

```
class SomeClass[SomeGenericType]:
    some_attribute: type[SomeGenericType]

    @classmethod
    async def some_method(cls, arg: str) -> SomeGenericType:
        return cls.some_attribute(arg) # <-- pyrefly(missing-attribute) raised


class ConcreteType:
    def __init__(self, value: str):
        self.value = value

class ConcreteClass(SomeClass[ConcreteType]):
    some_attribute = ConcreteType

    @classmethod
    async def some_other_method(cls, arg: str) -> ConcreteType:
        return cls.some_attribute(arg) # <-- no pyrefly error raised

```

We have recently migrated from basedpyright where this was not an issue. We do not think this should be an issue as it is pretty standard to specify generic types for an abstract parent that get defined on concrete versions of the class.

### Sandbox Link

_No response_

VSCode Version: 1.127.0 (Universal)
Pyrefly version: 1.1.1
OS: MacOS Tahoe v26.5.1

asukaminato0721:
# Summary

Fixes #4058

generic class attributes remain visible when accessed through `type[Self@...]`, which is the cls receiver in classmethods.

Bare class access like C.x still keeps the existing missing-attribute behavior.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -3440,7 +3440,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 ..
             } => {
                 ty = self.normalize_attr_ty(ty);
-                if ambiguous {
+                if ambiguous && !matches!(cls, ClassBase::SelfType(_)) {
                     ClassAttribute::no_access(NoAccessReason::ClassAttributeIsGeneric(
                         cls.class_object().dupe(),
                     ))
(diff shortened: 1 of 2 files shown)


yangdanny97:
seems reasonable to me, I'll leave the merge to @NathanTempest

yangdanny97 (approved):

PR #4075, opened 2026-07-07
Fix notebook inlay hints lost in code cells after a markdown cell

qnox:
# Summary

Adding a markdown cell above a code cell made that code cell lose its inlay hints - and every code cell after it. The same bug silently hit semantic tokens, document symbols, references, and diagnostic grouping, since they all share one per-cell filter.

The filter compares two cell indices that are supposed to mean the same thing. `LinedBuffer::to_cell_for_lsp` returned ruff's all-cells index (position among every cell, markdown included), but callers compare it against the code-cell index from `LspNotebook::get_code_cell_index` (position among code cells only, matching `cell_offsets()`). The two agree only when a notebook has no markdown cells, so a leading markdown cell shifts them apart and the filter matches nothing.

Translate the all-cells index to the code-cell index inside `to_cell_for_lsp` by counting the preceding code cells, mirroring `get_code_cell_index`, so the indices agree and every feature sharing this filter is fixed at once.

# Test Plan

Added a unit test `test_to_cell_for_lsp_returns_code_cell_index` in `crates/pyrefly_util/src/lined_buffer.rs`: for a `[code, markdown, code]` notebook, asserts a position in the second code cell resolves to code-cell index 1 (not all-cells index 2). Added a regression test `test_inlay_hints_with_preceding_markdown_cell` in `pyrefly/lib/test/lsp/lsp_interaction/notebook_inlay_hint.rs`: a code cell after a markdown cell still gets its variable-type hint. Both fail on main, pass with the fix.

- cargo test -p pyrefly_util lined_buffer
- cargo test -p pyrefly --test pyrefly_lsp_interaction_tests -- inlay sync

### DIFF
--- a/crates/pyrefly_util/src/lined_buffer.rs
+++ b/crates/pyrefly_util/src/lined_buffer.rs
@@ -261,7 +261,15 @@ impl LinedBuffer {
                 LineNumber::from_one_indexed(loc.line),
             )
         {
-            Some(cell.to_zero_indexed())
+            // ruff's `cell()` is the raw all-cells index (markdown included), but callers
+            // compare against the code-cell index from `get_code_cell_index`. Translate.
+            let abs = cell.to_zero_indexed();
+            Some(
+                notebook.cells()[..abs]
+                    .iter()
+                    .filter(|c| c.is_code_cell())
+                    .count(),
+            )
         } else {
             None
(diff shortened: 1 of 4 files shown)


stroxler (approved):

PR #4085, opened 2026-07-08
synthesize __required_keys__ and __optional_keys__ for TypedDicts

### LINKED ISSUE #3745: TypedDict-to-dict erasure in subtyping runs before the protocol check, so TypedDicts fail Protocols requiring __required_keys__ / __optional_keys__ (breaks langgraph)
## Summary

Pyrefly rejects TypedDict subclasses as members of structural Protocols that require `__required_keys__: ClassVar[frozenset[str]]` and `__optional_keys__: ClassVar[frozenset[str]]`, even though every TypedDict has these attributes at runtime, pyrefly itself models them (attribute access on both the class and its instances correctly resolves to `frozenset[str]`), and mypy and pyright both accept this pattern.

The root cause appears to be dispatch order in `is_subset_eq_impl`: the TypedDict catch-all arm erases the TypedDict to `dict[str, V]` / `Mapping[str, V]` *before* the generic protocol arm is reached, so the structural check runs against `dict` — which has no `__required_keys__` — instead of against the TypedDict and its synthesized `TypedDictFallback` base. Details and code pointers below.

This breaks every user of [langgraph](https://github.com/langchain-ai/langgraph) under pyrefly, since langgraph's core `StateT` type variable is bounded by a union of exactly these Protocols.

## Reproduction

```python
# repro.py
from typing import ClassVar, Protocol, TypedDict


class TypedDictLike(Protocol):
    __required_keys__: ClassVar[frozenset[str]]
    __optional_keys__: ClassVar[frozenset[str]]


class Foo(TypedDict):
    x: int


def takes[T: TypedDictLike](x: type[T]) -> type[T]:
    return x


takes(Foo)  # pyrefly: bad-specialization; mypy/pyright: OK

f: Foo = {"x": 1}
ok: TypedDictLike = f  # pyrefly: bad-assignment; mypy/pyright: OK
```

### Pyrefly 1.0.0 (also reproduces on 0.53.0)

```
ERROR `Foo` is not assignable to upper bound `TypedDictLike` of type variable `T` [bad-specialization]
  --> repro.py:17:6
   |
17 | takes(Foo)  # pyrefly: bad-specialization; mypy/pyright: OK
   |      ^^^^^
   |
ERROR `Foo` is not assignable to `TypedDictLike` [bad-assignment]
  --> repro.py:20:21
   |
20 | ok: TypedDictLike = f  # pyrefly: bad-assignment; mypy/pyright: OK
   |     -------------   ^
   |
  Protocol `TypedDictLike` requires attribute `__required_keys__`
```

### Mypy 2.1.0 and pyright 1.1.410

Both accept the snippet with no errors.

### Runtime

```pycon
>>> Foo.__required_keys__
frozenset({'x'})
>>> Foo.__optional_keys__
frozenset()
```

Python synthesizes these on every TypedDict at class-creation time ([docs](https://docs.python.org/3/library/typing.html#typing.TypedDict.__required_keys__)).

## Root cause

Notably, pyrefly *does* model these attributes correctly — the failure is specific to the subtyping path, not attribute lookup:

```python
reveal_type(Foo.__required_keys__)  # revealed type: frozenset[str]  ✅
def f(x: Foo) -> None:
    reveal_type(x.__required_keys__)  # revealed type: frozenset[str]  ✅
```

That works because pyrefly injects `_typeshed._type_checker_internals.TypedDictFallback` — which declares `__required_keys__: ClassVar[frozenset[str]]` and `__optional_keys__: ClassVar[frozenset[str]]` — as a base of every TypedDict:

- fallback injection: [`pyrefly/lib/alt/types/class_bases.rs#L394-L403`](https://github.com/facebook/pyrefly/blob/4c5d43a45da1d57f0ee0d7971c91938aec36e740/pyrefly/lib/alt/types/class_bases.rs#L394-L403)
- fallback stub: [`crates/pyrefly_bundled/.../_type_checker_internals.pyi#L20-L23`](https://github.com/facebook/pyrefly/blob/4c5d43a45da1d57f0ee0d7971c91938aec36e740/crates/pyrefly_bundled/third_party/typeshed/stdlib/_typeshed/_type_checker_internals.pyi#L20-L23)

The protocol check never sees any of that, because of match-arm ordering in `is_subset_eq_impl` (`pyrefly/lib/solver/subset.rs`):

1. The TypedDict catch-all arm at [`subset.rs#L1916-L1937`](https://github.com/facebook/pyrefly/blob/4c5d43a45da1d57f0ee0d7971c91938aec36e740/pyrefly/lib/solver/subset.rs#L1916-L1937) matches `(Type::TypedDict(..), _)` and rewrites the check as `dict[str, V] <: want` (or `Mapping[str, V] <: want`).
2. This arm matches **before** the generic structural-protocol arm at [`subset.rs#L2043`](https://github.com/facebook/pyrefly/blob/4c5d43a45da1d57f0ee0d7971c91938aec36e740/pyrefly/lib/solver/subset.rs#L2043), so `Foo <: TypedDictLike` becomes `dict[str, int] <: TypedDictLike`.
3. `dict` has no `__required_keys__`, so the structural check fails with `MissingAttribute` — producing the errors above.

In other words: the TypedDict-to-dict erasure (useful for protocols like `SupportsKeysAndGetItem`) is applied unconditionally, discarding the TypedDict's own attribute surface that pyrefly's attribute machinery already handles correctly. Mypy and pyright route this check through the TypedDict's `typing._TypedDict` fallback instead, which is why they accept.

## Suggested fix

In the TypedDict catch-all arm: when `want` is a protocol, additionally try the structural protocol check against the TypedDict itself, falling back between the two paths — e.g. keep the existing dict/`Mapping` erasure attempt first (so protocols that rely on precise erased value types, like `_typeshed.SupportsKeysAndGetItem[str, V]`, keep working), and on failure retry with `is_subset_protocol(got, want)` directly. All the machinery already exists: `is_protocol_subset_at_attr` handles TypedDict attribute bases, and the lookup resolves `__required_keys__` / `__optional_keys__` via `TypedDictFallback` as shown above. The ClassVar-to-ClassVar case already passes `is_class_attribute_subset`.

I'm happy to submit a PR for this if the approach sounds right.

## Real-world impact: langgraph

langgraph's public generics are bounded by exactly this pattern ([`langgraph/_internal/_typing.py`](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/_internal/_typing.py), [`langgraph/typing.py`](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/typing.py)):

```python
class TypedDictLikeV1(Protocol):
    __required_keys__: ClassVar[frozenset[str]]
    __optional_keys__: ClassVar[frozenset[str]]

class TypedDictLikeV2(Protocol):
    __required_keys__: frozenset[str]
    __optional_keys__: frozenset[str]

class DataclassLike(Protocol):
    __dataclass_fields__: ClassVar[dict[str, Field[Any]]]

StateLike: TypeAlias = TypedDictLikeV1 | TypedDictLikeV2 | DataclassLike | BaseModel

StateT = TypeVar("StateT", bound=StateLike)
```

`StateT` parameterizes `StateGraph` / `CompiledStateGraph`, and the idiomatic langgraph state type is a TypedDict, so under pyrefly essentially every typed langgraph codebase hits `bad-specialization` at graph factories and call sites. The current workaround is `# type: ignore[bad-specialization]` at each one (15+ ignores in our codebase).

Two notes that constrain the fix:

- Matching `TypedDictLikeV1` (the `ClassVar` variant) alone is sufficient to fix this, since the bound is a union. The non-`ClassVar` `TypedDictLikeV2` is rejected by mypy as well (`Protocol member TypedDictLikeV2.__required_keys__ expected instance variable, got class variable`) — langgraph defines both variants precisely because checkers disagree on that point, so pyrefly matching only V1 would be consistent with mypy.
- This "any-TypedDict" Protocol bound is the community-recommended pattern, since `bound=TypedDict` isn't expressible; other libraries use it too, so the fix isn't langgraph-specific.

## Environment

- pyrefly 1.0.0 and 0.53.0 (both reproduce; installed via pip, defaults aside from `project-includes`)
- mypy 2.1.0 and pyright 1.1.410 used for comparison (both accept)
- Code pointers above are pinned to main @ `4c5d43a`

### Sandbox Link

_No response_

_No response_

yangdanny97:
Summary:
During attribute lookup we fall back to `dict` instead of `TypedDictFallback` for TypedDicts, which is a usability tradeoff since dict more closely models the runtime value but allows some operations (mutation) that may not be safe.

Unfortunately it means that some things only present on TypedDicts are missing, so this diff synthesizes them.

fixes https://github.com/facebook/pyrefly/issues/3745

Differential Revision: D111085718

### DIFF
--- a/pyrefly/lib/alt/class/typed_dict.rs
+++ b/pyrefly/lib/alt/class/typed_dict.rs
@@ -71,6 +71,8 @@ const DEFAULT_PARAM: Name = Name::new_static("default");
 const UPDATE_METHOD: Name = Name::new_static("update");
 const ITEMS_METHOD: Name = Name::new_static("items");
 const VALUES_METHOD: Name = Name::new_static("values");
+const REQUIRED_KEYS: Name = Name::new_static("__required_keys__");
+const OPTIONAL_KEYS: Name = Name::new_static("__optional_keys__");
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     pub fn check_dict_items_against_typed_dict(
@@ -923,12 +925,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     pub fn get_typed_dict_synthesized_fields(&self, cls: &Class) -> Option<ClassSynthesizedFields> {
         let metadata = self.get_metadata_for_class(cls);
         let td = metadata.typed_dict_metadata()?;
+        let keys_type = self.heap.mk_class_type(
+            self.stdlib
+                .frozenset(self.heap.mk_class_type(self.stdlib.str().clone())),
+        );
         let mut fields = smallmap! {
             dunder::INIT => self.get_typed_dict_init(cls, &td.fields),
             ITEMS_METHOD => self.get_typed_dict_items(cls, &td.fields),
             GET_METHOD => self.get_typed_dict_get(cls, &td.fields),
             UPDATE_METHOD => self.get_typed_dict_update(cls, &td.fields),
             VALUES_METHOD => self.get_typed_dict_values(cls, &td.fields),
+            REQUIRED_KEYS => ClassSynthesizedField::new_classvar(keys_type.clone()),
+            OPTIONAL_KEYS => ClassSynthesizedField::new_classvar(keys_type),
         };
         if let Some(m) = self.get_typed_dict_clear(cls, &td.fields) {
             fields.insert(CLEAR_METHOD, m);
--- a/pyrefly/lib/test/typed_dict.rs
+++ b/pyrefly/lib/test/typed_dict.rs
@@ -1023,6 +1023,8 @@ testcase!(
 from typing import TypedDict, assert_type
 class C(TypedDict): ...
 assert_type(C.__total__, bool)
+assert_type(C.__required_keys__, frozenset[str])
+assert_type(C.__optional_keys__, frozenset[str])
     "#,
 );
 


grievejia (changes requested):

grievejia (approved):

Issue #438, opened 2025-06-06
Infer input type of lambda based on 2nd argument to map()


bigfootjon:
### Describe the Bug

It's really easy to write `map(lambda foo: ..., ...)` but this isn't typechecked

### Sandbox Link

https://pyrefly.org/sandbox/?code=MQAgKgFglgziMQPYFcA2ATEBTATjxOAhAFAC2AhgA4AUq5pARuuSAB4BcbAdOoljAH0AdogAuArK1ijqASgA0IANoBGALqzixdFgBmIUQE9K-YWIFQhu3DizpqAN3KpODRIlSyQAWgB8IADlEISx2YhAIkEpyGBgtChojE0ERcUtrPDtFVQ0gA

yangdanny97:
This is a bit tricky, we'll need to use the type of the second argument to map's constructor as part of the contextual type for the first argument. Right now I suspect we're just evaluating arguments left to right.

samwgoldman:
Making contextual typing work for calls like this is part of the scope of #105. Going to leave this open because it's a good example, and we _could_ solve #105 without fixing this in theory.

PR #4086, opened 2026-07-08
Fix bidirectional type checking not working for lambdas

### LINKED ISSUE #3950: Bidirectional type checking not working for lambdas
### Describe the Bug

The function, like list.sort in this example, already has the list as the self argument and thus this should be solvable with regular bidirectional type checking only. This shouldn't need the more complex bidirectional type checking that isn't yet implemented like #438 and #1485

The main problem seems to be that `CallWithTypes` is pre-inferring the type of the lambda before it gets to do any bidirectional type checking. I'm working on a potential fix that is relatively simple, will submit soon, just doing some final cleanup #assign

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRyocBojpQISgNqsGAXToBeOhp0KlhOLwYAKANYxS%2BqKhrZMqOjBWCRYycxhWMACUQXR0AMR0AJIA5Jge%2BNIAxvwMABbqjLh02DCc6GAwlILxqHD5DAA0dOi4-ACq6Da1AO7y6CCVIGSCYFCkhAy0UBSRAAqkvf10aFh4%2BHRJuAUQbACulKgMEMuEcpEAyjB5aQwMxHCIAPRXPTB9A7xsVzDoV5i4SXBXSyvrm9tllc6GBeHRUEJUNBUNhYItlpB-lsdug6LhiID0HA9ugyOllgBaESUOAogx0GQgADMhAAjAAmSlyDRFagkkzoNboLg8PgwTAEzAQQQpCAickxVa1QQxOS1BgEwQARzWwv5BLspAJqCSSXg5UMMRaqEo6FlHQAvl0ddsRAAxaAwCizHAEEjkC1AA

_No response_

hrolfurgylfa:
# Summary

I addressed the missing bidirectional type-checking for lambdas with a fairly easy one line fix, but it revealed a separate problem with overloads leaking errors between each other in my sort unit test so I ended up fixing that as the first commit with a separate regression test.

There is one other unit test that got modified by this change, as it was relying on the previous behvaiour of eagerly inferring lambdas in function calls,  but it should be fine this way as well. Pyrefly should never have a problem with casting a literal back down to a string right?

Fixes #3950 

# Test Plan

Only tested through unit tests, this only touches on the solving step, so it shouldn't be close to anything else.

No files generated by `test.py`

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -65,11 +65,24 @@ use crate::types::types::Var;
 /// This is used to avoid re-inferring types for arguments multiple types.
 ///
 /// Implemented by keeping an `Owner` to hand out references to `Type`.
-pub struct CallWithTypes(Owner<Type>);
+pub struct CallWithTypes {
+    owner: Owner<Type>,
+    bidirectional_lambda_inferrence: bool,
+}
 
 impl CallWithTypes {
     pub fn new() -> Self {
-        Self(Owner::new())
+        Self {
+            owner: Owner::new(),
+            bidirectional_lambda_inferrence: false,
+        }
+    }
+
+    pub fn new_with_lambda(bidirectional_lambda_inferrence: bool) -> Self {
+        Self {
+            owner: Owner::new(),
+            bidirectional_lambda_inferrence,
+        }
     }
 
     pub fn type_or_expr<'a, 'b: 'a, Ans: LookupAnswer>(
@@ -85,9 +98,17 @@ impl CallWithTypes {
                 // the function's parameter types.
                 TypeOrExpr::Expr(e)
             }
+            TypeOrExpr::Expr(e @ Expr::Lambda(_)) if self.bidirectional_lambda_inferrence => {
+                // Even more of a hack: When the generic arguments going into the lambda are
+                // already solved, we can use bidirectional typechecking for the lambda, this only
+                // works if the type arguments are already solved in order, as it is not yet
+                // supported to solve type arguments in the order they are simplest in, they are
+                // always solved in order.
+                TypeOrExpr::Expr(e)
+            }
             TypeOrExpr::Expr(e) => {
                 let t = solver.expr_infer(e, errors);
-                TypeOrExpr::Type(self.0.push(t), e.range())
+                TypeOrExpr::Type(self.owner.push(t), e.range())
             }
             TypeOrExpr::Type(t, r) => TypeOrExpr::Type(t, r),
         }
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -56,6 +56,7 @@ struct CalledOverload<'f> {
     func: &'f TargetWithTParams<Function>,
     res: Type,
     ctor_targs: Option<TArgs>,
+    arg_errors: ErrorCollector,
     call_errors: ErrorCollector,
     specialization_errors: Vec<TypeVarSpecializationError>,
     /// Maps each argument's source range to the parameter it was matched against.
@@ -258,11 +259,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // If we're constructing a class, its type arguments. A successful call will fill these in.
         ctor_targs: Option<&mut TArgs>,
     ) -> (Type, Callable) {
+        let allow_bidirectional_lambda_inferrence = match &metadata.kind {
+            pyrefly_types::callable::FunctionKind::Def(def) => def.cls.is_some(),
+            _ => false,
+        };
         // There may be Expr values in args and keywords.
         // If we infer them for each overload, we may end up inferring them multiple times.
         // If those overloads contain nested overloads, then we can easily end up with O(2^n) perf.
         // Therefore, flatten all TypeOrExpr's into Type before we start
-        let call = CallWithTypes::new();
+        let call = CallWithTypes::new_with_lambda(allow_bidirectional_lambda_inferrence);
         let args = call.vec_call_arg(args, self, errors);
         let keywords = call.vec_call_keyword(keywords, self, errors);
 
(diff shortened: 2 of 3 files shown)


hrolfurgylfa:
This is a lot of changes through mypy_primer, I might split the two commits up into separate PRs to make it easier to go through each one individually.

hrolfurgylfa:
Nevermind, this is all the same bug, this overload error propagation doesn't happen anywhere else. I can't get it to happen with mutable builtin containers either, [something like this](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BhVFCipssADR0AMhAYxKwqQBVmMTABEIAYwZTcANwVRcqTFMowjwgPpNiMADrpn20XDh0AggApVDzR0GAEpEZzoIum0AC1RKRBk5BWEAbUcQVHSAXWdXd08AIT81QN1Q8MiYuITZeUUoNJBsbNz0AAFDY1NMZ0wYMDpUYmIoUh8wBK9gugBaAD46ADlcdBgEwg3nDqNKEzNe-sHh0fGEgun5pZW1ug3CA4GhkbGJ7zoAHzoCqXk4BgS8LgoHQALx0ABiwjgMAuC2Wq3Wmxc6CeJ2A6SqlHSCXSzRAAF9gs4QBIQGRLGBRoQGLQoBQAMR0AAKpApozoaCweHwURWkDYAFdFAwICt7uhGQBlGAwOjRBgMYhwRAAehV5P6VN4bBVMHQKswuG0cBV2j5EEFwtF%2BroYF4gwMqGgYlgvPQ-KFqBFKzouGI3vQcHFZAY0RWMx2cGtoLo6QAzIQAIwAJnSzhSCmolDgOXQAvQXB4fHUM0wEEsuggRhjAHILeheDAa84GwwZpYAI4C8slgDWMFIM1Q2m08E8YJrAHc4uhm%2BgCaThyKjODoDAKJycAQSOR8UA) works just fine even though I'd expect it to exercise the exact same path. I guess I might be fixing this in the wrong place and these containers address the same problem elsewhere.

But for now I'll just go through mypy_primer results one by one and see if they are all expected. Will re-open the PR once I have reviewed all of those.

hrolfurgylfa:
My review of the first few mypy_primer finds, one probably a bug I introduced, others are all expected so far, will continue this weekend:

First 3 are just fixes because of bidirectional type checking

### steam.py (https://github.com/Gobot1234/steam.py)

Not really sure, I must have lost types somewhere in the overload changes.

### jax (https://github.com/google/jax)

These are all showing up because the lambda return in the following statement is no longer inferred to be Unknown, so it went further in the type checking process and somehow wended up with object. To be fully solved this would require the more complex bidirectional type checking in #438

```
jax2tf_limits = tuple(filter(lambda l: l.filter(), limitations))
```

### spark (https://github.com/apache/spark)

These are now showing up in lambdas as previously the argument inside them was Unknown, since they were eagerly evaluated.

hrolfurgylfa:
I've gone through about half of the mypy_primer list, here are my thoughts so far:

First 3 are just fixes because of bidirectional type checking

### steam.py (https://github.com/Gobot1234/steam.py)

This is using a function similar to map/filter, the first argument is the lambda and the second argument is the iterable that has the type variable information. The reason this regressed seems to be that previously the lambda would be solved to Unknown and the function would use the second argument to infer the type, but now that the lambda isn't eagerly evaluated, it now tries to use the lambda to find the type, which fails into it being Unknown. Not completely sure on this one though, but it would be solved by #438

Here is a minimal example of this regression:

```py
from collections.abc import AsyncIterable, Awaitable, Callable, Iterable
from typing import overload, reveal_type


@overload
def find[T](predicate: Callable[[T], bool], iterable: AsyncIterable[T], /) -> Awaitable[T | None]: ...


@overload
def find[T](predicate: Callable[[T], bool], iterable: Iterable[T], /) -> T | None: ...


def find[T](
    predicate: Callable[[T], bool], iterable: Iterable[T] | AsyncIterable[T], /
) -> T | None | Awaitable[T | None]: ...


@dataclass
class WithName:
    name: str


@dataclass
class PartialReaction:
    emoticon: WithName | None
    sticker: WithName


emoticon: WithName = WithName("abc")
partial_reactions: list[PartialReaction] = []

reaction = find(lambda r: (r.emoticon or r.sticker).name == emoticon.name, partial_reactions)
reveal_type(reaction)  # E: revealed type: PartialReaction | None
```

### jax (https://github.com/google/jax)

These are all showing up because the lambda return in the following statement is no longer inferred to be Unknown, so it went further in the type checking process and somehow ended up with object. To be fully solved this would require the more complex bidirectional type checking in #438

```
jax2tf_limits = tuple(filter(lambda l: l.filter(), limitations))
```

### spark (https://github.com/apache/spark)

These errors are now showing up inside lambdas as previously the argument inside them was Unknown, since they were eagerly evaluated.

### prefect (https://github.com/PrefectHQ/prefect)

TODO, very confusing why it fails to find the overload of `sorted`

### freqtrade (https://github.com/freqtrade/freqtrade)

trade_parallelism.py:

Error in the lambda of a sorted call now showing, just a fault of actually type checking the lambda now.

PercentChangePairList.py:

Not exactly sure why it isn't being inferred into the second overload of `sorted`, I thought Unknown would be treated as Any and it would be able to match into it just fine.

### pwndbg (https://github.com/pwndbg/pwndbg)

For the first ones:

Lambdas that accept *args seem to be treated as zero argument functions instead of taking any number of arguments, hence getting these errors calling for too few arguments. Reproduction example:

```py
def add(a: int, b: int) -> int:
    return a + b


def mul(a: int, b: int) -> int:
    return a * b


FUNCS = {"add": add, "mul": mul}

key: str = "any"
a = FUNCS.get(key, lambda *c: 3)(1, 2)
```

For the last two:

A `filter(lambda x: x is not None)` is run on a list with a really large union `list[str | Unknown] | list[str | Unknown | None] | list[str] | list[Unknown]` and it collapses into a list[object], not really sure why that is happening.

hrolfurgylfa:
At this point, I'm not too sure if this is this right direction I am heading in, if it is, I could continue, but seeing if these regressions can be fixed and going through the entire list is going to take me some time, probably weeks.

Maybe it is not worth-it enabling this until #438 is solved, but there are lots of cases this does improve, including quite a few cases in the closed source repos I work on in my day job, which I can't migrate to pyrefly until this is resolved.

PR #4091, opened 2026-07-09
getComputedType: route all declaration-name ranges through the binding path (#4091)

### LINKED ISSUE #4042: TSP: inferred return type dropped at definition-site requests (unannotated functions / methods / @property getters render `-> Unknown`)

kinto0:
Summary:

fix https://github.com/facebook/pyrefly/issues/4042

by using the classifications in classify_surface

Reviewed By: stroxler

Differential Revision: D111155589

### DIFF
--- a/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
+++ b/pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
@@ -787,6 +787,195 @@ fn test_get_computed_type_function_has_return_type() {
     tsp.shutdown();
 }
 
+// =======================================================================
+// getComputedType range routing: current behavior of identifier-name ranges,
+// which (apart from bare-name references) are NOT routed through the
+// declaration-preserving path.
+//
+// These document the stable baseline so a later change to the routing
+// predicate has something to diff against. The cases marked BUG return `null`
+// today even though a point query at the same position resolves correctly; the
+// following diff routes every identifier-name range through the binding path.
+// =======================================================================
+
+/// Like `get_computed_type_range_ok` but returns the raw result value, which
+/// may be JSON `null`. Used to document ranges that currently resolve to no
+/// type.
+fn get_computed_type_range_raw(
+    tsp: &mut TspInteraction,
+    file_uri: &str,
+    start_line: u32,
+    start_character: u32,
+    end_line: u32,
+    end_character: u32,
+    snapshot: i32,
+) -> serde_json::Value {
+    tsp.server.get_computed_type_range(
+        file_uri,
+        start_line,
+        start_character,
+        end_line,
+        end_character,
+        snapshot,
+    );
+    let resp = tsp.client.receive_response_skip_notifications();
+    assert!(
+        resp.error.is_none(),
+        "Expected success, got error: {:?}",
+        resp.error
+    );
+    resp.result.expect("Expected result field")
+}
+
+#[test]
+fn test_get_computed_type_class_def_name_range() {
+    // BUG: a class-def name range is routed to the trace path, which has no
+    // recorded type for a bare declaration name, so this returns null. A point
+    // query at the same position returns the class.
+    let (mut tsp, file_uri, snapshot) = setup_project("class Foo:\n    x: int = 0\n");
+
+    let result = get_computed_type_range_raw(&mut tsp, &file_uri, 0, 6, 0, 9, snapshot);
+    assert!(
+        result.is_null(),
+        "class-def name range currently returns null, got: {result}"
+    );
+
+    tsp.shutdown();
+}
+
+#[test]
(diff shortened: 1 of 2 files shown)


stroxler (approved):

pyrefly/lib/test/tsp/tsp_interaction/get_type_queries.rs
```diff
@@ -787,6 +806,242 @@ fn test_get_computed_type_function_has_return_type() {
...
+            .get("returnType")
+            .unwrap_or_else(|| panic!("Expected returnType for {label}: {result}"));
+        assert_kind(return_type, TypeKind::Class);
+    }
+
+    tsp.shutdown();
+}
+
+// =======================================================================
+// getComputedType range routing: an identifier-name range resolves through
```
rchiodo, line 885: This range-routing section covers def/class/param/typevar/import/except/global/match, but not a keyword-argument label range — which is the one case whose routing actually changed in this PR. Previously the `ExprName`-only guard in `get_computed_type_at_range` sent the `x` range in `f(x=1)` to the trace path; now `classify_surface` maps `KeywordArgument` to a `Key` (the matched parameter's declaration), so it routes through the declaration-preserving path instead (or `NoType` → trace when no parameter matches).

Since that behavior change isn't pinned by any test, could we add one here? Something like `def f(x: int): ...` then `f(x=1)`, asserting the `x=` label range resolves to the parameter's type — plus a no-match case (e.g. `f(y=1)`) that should still fall back to the trace path.

PR #4095, opened 2026-07-10
Route Size/Dim through the stdlib int class (#4095)

kinto0:
Summary:

`Size`/`Dim` (integer tensor dimensions) were emitted as `int` `BuiltInType`. The protocol restricts `BuiltInType.name` to a fixed sentinel set that does not include `int`, so consumers render the off-spec name as `Unknown` — the same class of bug fixed for `None` and `TypeGuard`/`TypeIs`.

This encodes `Size`/`Dim` as the stdlib's `int` `ClassType` via the existing `convert_class_type(..., INSTANCE)` path, threaded through the `StdlibClasses` bundle. The declaration inherits `int`'s real module and range, so it is navigable and identical to an explicit `int` annotation.

Reviewed By: yangdanny97

Differential Revision: D111366731


yangdanny97 (approved):

PR #4097, opened 2026-07-10
fix support show range when hover on and or #4065

### LINKED ISSUE #4065: support show range when hover on `and` `or`
Now it only highlight the background of `and`.

I want it to highlight `1 and 2` , not just 1 `and` 2

asukaminato0721:
# Summary

Fixes #4065

hover results can carry an explicit range, and and/or operator hovers now return the enclosing boolean expression range.

# Test Plan

add test


NathanTempest (approved):
Thanks @asukaminato0721 for the nice UX fix and importing

kinto0 (changes requested):

stroxler (approved):

PR #4098, opened 2026-07-10
Fix TOML parse error locations

### LINKED ISSUE #4060: TOML parse error shows incorrect location
### Describe the Bug

I was trying out `pytorch-efficiency-lints` and created the following `pyrefly.toml`:
```
preset = "strict"
pytorch-efficiency-lints = "true"
```

Here I incorrectly used `"true"` instead of `true`. However, `pyrefly` incorrectly reports where the error in toml parsing occured:
```
$ pyrefly check example.py 
ERROR /home/dillon/sbx/demos/pyrefly/pyrefly.toml: TOML parse error at line 1, column 1
  |
1 | preset = "strict"
  | ^
invalid type: string "true", expected a boolean

Fatal configuration error
```

Reproducible in the sandbox, which just displays the first line as well.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIIANCGQE4xhSmEAuuAtlBQMQAEABVIMmpPmix58fAMa50kAOYBXOqlYR5hADrp%2BAZRgw%2BAC1aticRAHpr9Rs0K46i6zHTXMuGXGtyFECpqGvLWfGDOfKgAbqjQqNiwsvJKquqa6Hy4xCHocDroZKwm8gC00TB0cBl8ALx82iAAzIQAjABMjbrEDHAwrHUNIHCsdBAyrF2FpOx0MialjJAyEO4ypKVQEOiscIONo8owUyAAvjSoExAVAGLQMBQSOAQk5KdAA

_No response_

goutamadwant:
# Summary

Fixes #4060

This fixes TOML parse diagnostics so invalid config values point at the offending value instead of the start of the file.

The config parser now keeps usable TOML span information and adjusts broad serde validation spans to the matching TOML value span when possible. The playground now uses the same config parser and reports the derived range instead of hardcoding line 1, column 1.

# Test Plan

- `cargo test -p pyrefly_config`
- `cargo test -p pyrefly test_config_toml_parse_error_range -- --nocapture`

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -53,6 +53,7 @@ use pyrefly_util::telemetry::TelemetrySourceDbRebuildStats;
 use pyrefly_util::watch_pattern::WatchPattern;
 use serde::Deserialize;
 use serde::Serialize;
+use serde::de::DeserializeOwned;
 use serde_with::skip_serializing_none;
 use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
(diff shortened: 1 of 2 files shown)


NathanTempest (changes requested):
Thanks @goutamadwant for this clean fix. Nice that it fails safe (falls back to the original span when a better one can't be derived) and that both the config-crate and playground tests pin the corrected location.

One follow-up worth considering (non-blocking): the span adjustment recovers the key/value by parsing toml_edit's error message text (in `key.path`, invalid type: string "…"), which is coupled to toml_edit's wording and could silently stop adjusting on a version bump. It degrades safely, so it's fine to land, but longer-term a toml::Spanned/serde_spanned-based approach (or reconstructing a toml_edit error with the adjusted span rather than the hand-rolled renderer) would be more robust.

goutamadwant:
Thanks @NathanTempest. I pushed a follow-up to address your comments. let me know if you have more suggestions. thanks!

NathanTempest (approved):
This is a big improvement and thanks for reworking it. Tests for both pyrefly.toml and [tool.pyrefly] plus the playground case are great. Looks import ready to me and thank you for this!

grievejia (changes requested):

grievejia (approved):

PR #4099, opened 2026-07-10
fix make function wrap easier to read on hover #3618

### LINKED ISSUE #3618: make function wrap easier to read on hover
### Describe the Bug

this is current panel.

Image

it's hard to get the meaning is

```
(T, str, ParamSpec(P)) -> R
to
(T, ParamSpec(P)) -> R
```

Even the bracket is hard to match, no rainbow color here.

Especially in python there is no auto curry.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3618

Nested callable types now wrap in hover display, including Concatenate[..., ParamSpec] shapes.

Hover display now recursively uses the hover-aware formatter for nested callables.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -308,33 +308,39 @@ impl ParamList {
     pub fn fmt_with_type_with_newlines<O: TypeOutput>(
         &self,
         output: &mut O,
-        write_type: &impl Fn(&Type, &mut O) -> fmt::Result,
+        write_type: &impl Fn(&Type, &mut O, usize) -> fmt::Result,
+        indent: usize,
     ) -> fmt::Result {
         let mut named_posonly = false;
         let mut kwonly = false;
 
         for (i, param) in self.0.iter().enumerate() {
             if i > 0 {
-                output.write_str(",\n    ")?;
+                output.write_str(",\n")?;
+                write_indent(output, indent)?;
             }
 
             if matches!(param, Param::PosOnly(Some(_), _, _)) {
                 named_posonly = true;
             } else if named_posonly {
                 named_posonly = false;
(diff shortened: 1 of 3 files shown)


samwgoldman:
This seems reasonable to me, but I think that we can do better over time pretty-printing types, where the types _might_ be long and _might_ be confusing, but won't always be so. If I understand correctly, after this change, we will now break lines even on simple callbacks like `def map[T,U](f: (T) -> U, xs: list[T]) -> list[U]` which is easy to read on a single line.

In the past, I've used OCaml's [Format](https://ocaml.org/api/Format.html) module which provides useful primitives like break hints. The [Format Unraveled](https://rbonichon.github.io/papers/format-unraveled.pdf) paper is a nice introduction to this and the background.

I wonder if a similar library exists for Rust?

stroxler (approved):

asukaminato0721:
> This seems reasonable to me, but I think that we can do better over time pretty-printing types, where the types _might_ be long and _might_ be confusing, but won't always be so. If I understand correctly, after this change, we will now break lines even on simple callbacks like `def map[T,U](f: (T) -> U, xs: list[T]) -> list[U]` which is easy to read on a single line.
> 
> In the past, I've used OCaml's [Format](https://ocaml.org/api/Format.html) module which provides useful primitives like break hints. The [Format Unraveled](https://rbonichon.github.io/papers/format-unraveled.pdf) paper is a nice introduction to this and the background.
> 
> I wonder if a similar library exists for Rust?

did some research https://crates.io/crates/pretty, https://doc.rust-lang.org/stable/nightly-rustc/rustc_ast_pretty/pp/

and maybe https://docs.rs/elegance/latest/elegance/ is the closest?

PR #4100, opened 2026-07-10
fix Django: ForeignKey(mymodule.MyModel) not recognized #4092

### LINKED ISSUE #4092: Django: ForeignKey(mymodule.MyModel) not recognized
### Describe the Bug

Pyrefly's Django support infers a `ForeignKey` target correctly when the model is imported directly, but not when the same model is accessed through an imported module alias.

This works:

```python
from app1.models import MyItem

class FKDirect(models.Model):
    item = models.ForeignKey(MyItem, on_delete=models.CASCADE)
```

It doesn't work, if I import the module as `from app1 import models as models1` and use `models1.MyModel`

`app1/models.py`:

```python
from django.db import models

class MyItem(models.Model):
    name = models.CharField(max_length=100)
```

`app2/models.py`:

```python
from django.db import models
import app1.models as models1

class FKModuleAlias(models.Model):
    item = models.ForeignKey(models1.MyItem, on_delete=models.CASCADE)

assert_type(FKModuleAlias().item, models1.MyItem)
```

Pyrefly reports:

```
ERROR assert_type(Any, MyItem) failed [assert-type]
  --> app2/models.py:14:12
   |
14 | assert_type(alias_item, MyItem)
   |            ^^^^^^^^^^^^^^^^^^^^
```

I have attached a ZIP file with this set up, as it requires multiple files to reproduce: [pyrelfy-django-repro.zip](https://github.com/user-attachments/files/29857074/pyrelfy-django-repro.zip)

I'm aware of 2 related Django ForeignKey issues #3657, #3932. However, those seem to require new analysis to correctly resolve the model from a string. This issue seems easier to fix, as the model type is right there in the argument. Pyrefly is able to resolve the `models1.MyItem` type in other contexts.

Tested on Pyrefly 1.2.0-dev.1, older versions behaved the same way. Django-stubs version 6.0.6

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4092

Django relationship targets now resolve attribute expressions.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/django.rs
+++ b/pyrefly/lib/alt/class/django.rs
@@ -253,8 +253,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
 
     fn resolve_target(&self, to_expr: &Expr, class: &Class) -> Option<Type> {
         match to_expr {
-            // Use expr_infer to resolve the name in the current scope
-            Expr::Name(_) => {
+            // Use expr_infer to resolve the model in the current scope.
+            Expr::Name(_) | Expr::Attribute(_) => {
                 let model_type = self.expr_infer(to_expr, &self.error_swallower());
                 Some(self.class_def_to_instance_type(&model_type))
             }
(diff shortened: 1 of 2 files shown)


yangdanny97 (approved):

PR #4102, opened 2026-07-10
Add combined full text and GitHub output

### LINKED ISSUE #4087: support simultaneous full-text and github output
The `github` output format emits github "workflow commands", which the github UI
interprets and uses to annotate source files with the detected errors. These
commands are elided from the full workflow logs, which means that users looking
at these logs cannot see file and line information for pyrefly errors. It would
be ideal if pyrefly could emit both `github` and `full-text` output
simultaneously.

I'm happy to submit a PR for this (subject to my company approving the CLA), and
would appreciate some feedback on how to best proceed.

Some history I've found:
 - When `github`-formatted output was introduced, workflow commands were
   automatically emitted in addition to the selected output format when the
   environment variable `GITHUB_ACTIONS` was true.
   - https://github.com/facebook/pyrefly/issues/1599
   - https://github.com/facebook/pyrefly/pull/1785
 - That behavior caused a number of problems and was reverted.
   - https://github.com/facebook/pyrefly/issues/2392
   - https://github.com/asukaminato0721/pyrefly/commit/5c27c1b6

Proposals:
1. Add a new output format (maybe `full-github`?) that emits both `github` and
   `full-text` outputs.
2. Allow arbitrary combinations of output formats.
3. make `--[no-]-github-workflow-commands` a separate option, orthogonal to `--output-format`

Of these, I think 1 would be the better choice? 2 seems like it creates a lot of
new "test surface" and its not clear what other use cases might be relevant.

paranoa233:
# Summary

- add a `full-text-with-github` output format that emits readable diagnostics followed by GitHub workflow commands
- use the combined format in the Pyrefly composite action so source locations remain visible in raw workflow logs
- cover config parsing, combined rendering, and CLI output behavior

Fixes #4087.

# Test Plan

- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly_config output_format` (3 passed)
- `cargo +stable-x86_64-pc-windows-gnu test -p pyrefly commands::check::tests` (16 passed)
- `py -3.11 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` (formatting and linting passed)
- ran the compiled CLI with `--output-format=full-text-with-github` and confirmed it prints both the full diagnostic and the corresponding `::error` workflow command

# AI disclosure

I am an AI agent working under the `paranoa233` account. I inspected the implementation path, added focused tests, and validated the change with the checks above.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -51,4 +51,4 @@ runs:
       working-directory: ${{ inputs.working-directory }}
       env:
         EXTRA_ARGS: ${{ inputs.args }}
-      run: pyrefly check --output-format=github $EXTRA_ARGS
+      run: pyrefly check --output-format=full-text-with-github $EXTRA_ARGS
--- a/action.yml
+++ b/action.yml
@@ -51,4 +51,9 @@ runs:
       working-directory: ${{ inputs.working-directory }}
       env:
         EXTRA_ARGS: ${{ inputs.args }}
-      run: pyrefly check --output-format=full-text-with-github $EXTRA_ARGS
+      run: |
+        OUTPUT_FORMAT=github
+        if pyrefly check --help | grep -q "full-text-with-github"; then
+          OUTPUT_FORMAT=full-text-with-github
+        fi
+        pyrefly check --output-format="$OUTPUT_FORMAT" $EXTRA_ARGS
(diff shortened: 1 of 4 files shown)


stroxler (approved):

PR #4105, opened 2026-07-10
fix unknown even when has definition #4064

### LINKED ISSUE #4064: unknown even when has definition
### Describe the Bug

Image

vs

Image

```py
    def __contains__(self, key: object, /) -> bool:
        """Return bool(key in self)."""
        ...
```

So I think even when key don't have type, the operation is sure?

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4064

 Membership with an Any/unknown item now invokes the known container's `__contains__` and returns bool.

Hover now displays the resolved `list[str].__contains__` signature instead of Unknown.

# Test Plan

add test


stroxler (approved):

PR #4106, opened 2026-07-10
fix Insert import with alias creates invalid import #2495

### LINKED ISSUE #2495: Insert import with alias creates invalid import
### Describe the Bug

Pyrefly shows import suggestions also based on alias imports in other files. When using such an import it creates a `from foo import NameAlias` statement instead of  `from foo import Name as NameAlias`. Here is a minimal example to reproduce it:

```python
# model.py
class MyModel: pass
```

```python
# alias_user.py
from pyrefly_repro.model import MyModel as MyModelAlias
```

```python
# normal_user.py
x: MyM # and continue typing
```

Image

Pressing enter results in an invalid import:

```python
# normal_user.py
from pyrefly_repro.model import MyModelAlias # Could not import `MyModelAlias` from `pyrefly_repro.model`
x: MyModelAlias
```

### Sandbox Link

_No response_

VS Code with Pyrefly v0.53.0

asukaminato0721:
# Summary

Fixes #2495

Auto-import resolution now preserves the original exported name and generates `from model import MyModel as MyModelAlias`.

Unknown-name quick fixes receive the same alias-aware behavior.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/test/lsp/completion.rs
+++ b/pyrefly/lib/test/lsp/completion.rs
@@ -2542,7 +2542,7 @@ Completion Results:
 }
 
 #[test]
-fn autoimport_aliased_import_uses_original_name() {
+fn autoimport_aliased_import_uses_aliasing_module() {
     let code = r#"
 x: MyM
 #     ^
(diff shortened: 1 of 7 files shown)


pyrefly/lib/test/lsp/completion.rs
```diff
@@ -2482,6 +2482,37 @@ Completion Results:
...
+"#;
+    let files = [
+        ("main", code),
+        ("model", "class MyModel: pass\n"),
+        ("alias_user", "from model import MyModel as MyModelAlias\n"),
+    ];
+    let (handles, state) = mk_multi_file_state(&files, Require::Exports, false);
+    let handle = handles.get("main").unwrap();
+    let position = extract_cursors_for_test(code)[0];
+    let autoimport = state
```
kinto0: curious if MyModel is also in the result? if so, can we make sure MyModel is ranked beneath MyModelAlias?

asukaminato0721: Yes, MyModel is also returned. Previously it ranked above MyModelAlias.

I updated ranking so aliases are preferred within the same public/private auto-import tier, and added an assertion that: MyModelAlias < MyModel

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -2482,6 +2482,48 @@ Completion Results:
...
+        .position(|item| item.label == "MyModel")
+        .expect("expected MyModel to be in completions");
+    assert!(
+        alias_index < original_index,
+        "expected the matching alias to rank above its original name"
+    );
+    let autoimport = &completions[alias_index];
+
+    assert_eq!(
+        autoimport.detail.as_deref(),
```
kinto0: unless i'm misinterpreting this test, I think this behavior is incorrect. 

- we want MyModel (from model) and MyModelAlias (from alias_user)
- the bug was that we got MyModel (from model) and MyModel (from alias_user)

here, it looks like we are asserting that we have MyModel (from model) and MyModel as MyModelAlias (from alias_user)

kinto0 (changes requested):

yangdanny97 (approved):

PR #4111, opened 2026-07-11
Fix completions for import aliases

### LINKED ISSUE #4107: `import pandas as pd` completes to `import pandas as pdb`
### Describe the Bug

this is an annoying since pressing enter to add a newline adds the d. we shouldn't suggest module completions in the `as` alias

https://github.com/user-attachments/assets/7a8ee8f2-87fa-4024-8cb5-52c98d0bb7a8

### Sandbox Link

_No response_

_No response_

roian6:
# Summary

Fixes #4107

Import alias names are local binding targets, but completion treated them as import paths. This produced module suggestions for `import pandas as pd` and package-export suggestions for `from package import name as alias`.

Classify `Alias.asname` identifiers as `IdentifierContext::AliasDefinition` inside the shared `identifier_from_covering_nodes` path for both import forms. Completion now suppresses suggestions from that semantic context instead of performing a second AST match, while normal module and imported-name completions remain unchanged. Alias type and go-to-definition behavior continue through the existing definition-key paths.

**AI agent disclosure:** This PR was prepared by Hermes Agent, an AI coding agent, under the contributor's direction. The agent traced the relevant source paths, added and exercised regression tests, ran the repository validation commands listed below, and performed an independent automated code review. The contributor is aware of the AI-agent involvement.

# Test Plan

- `cargo test -p pyrefly import_alias_has_no_completions -- --nocapture` — 2 passed, 0 failed
- `cargo test -p pyrefly 'test::lsp::completion::' -- --nocapture` — 100 passed, 0 failed
- `cargo test -p pyrefly 'test::lsp::'` — 651 passed, 0 failed
- `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` — formatting and linting passed
- `git diff --check`

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -440,6 +440,9 @@ pub(crate) enum IdentifierContext {
         /// ex: For `from ... import x`, the name is `x`. For `from ... import x as y`, the name is `y`.
         name_after_import: Identifier,
     },
+    /// An identifier introduced as a local import alias.
+    /// ex: `y` in `import x as y` or `from x import z as y`.
+    AliasDefinition,
     /// An identifier appeared as the name of a function.
     /// ex: `x` in `def x(...): ...`
     FunctionDef { docstring_range: Option<TextRange> },
@@ -479,6 +482,7 @@ impl IdentifierContext {
                 }
                 | IdentifierContext::ImportedModule { .. }
                 | IdentifierContext::ImportedName { .. }
+                | IdentifierContext::AliasDefinition
                 | IdentifierContext::FunctionDef { .. }
                 | IdentifierContext::MethodDef { .. }
                 | IdentifierContext::ClassDef { .. }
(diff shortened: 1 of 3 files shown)


kinto0 (changes requested):
thank you! great fix. it looks like you found exactly what to change. 

I have one architectural request:
we already do an AST match in lsp.rs identifier_from_covering_nodes. instead of doing another extra match in the completion logic, could we use that function?

we likely need a new `AliasDefinition` variant and we would match within this same function

roian6:
Thanks @kinto0 — updated in `9cae46582` as requested.

`AliasDefinition` is now produced by the shared `identifier_from_covering_nodes` classifier for `asname` identifiers in both import forms, and completion consumes that context instead of doing its own AST match. Type and go-to-definition continue through the existing definition-key paths.

Validation:
- alias regressions: 2 passed
- completion suite: 100 passed
- lightweight LSP suite: 651 passed
- repository formatting/lint gate: passed

pyrefly/lib/state/lsp.rs
```diff
@@ -900,7 +911,17 @@ impl<'a> Transaction<'a> {
                 _,
             ) => {
```
kinto0, line 912: I think it's cleaner to put the if statement in the match. then we don't need to do it in the expression below

kinto0 (approved):
thanks! looks great. very minor feedback. can fix this up if I get to it first

yangdanny97 (approved):

kinto0:
fixing match + merging

PR #4112, opened 2026-07-11
Fix duplicate module auto-import completions

### LINKED ISSUE #4109: completions auto-import after import exists
### Describe the Bug

https://github.com/user-attachments/assets/1fc8d094-cdb5-4380-b4e6-2b297c7de644

### Sandbox Link

_No response_

_No response_

aodihis:
# Summary

Ensure modules imported with regular imopr statements, such as import json, are treated as local completion candidates to avoid double import.

Fixes #4109 

# Test Plan

Added regression test for duplicate module auto-import completions.

### DIFF
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -39,6 +39,7 @@ use ruff_text_size::TextSize;
 use starlark_map::small_set::SmallSet;
 
 use crate::alt::attr::AttrInfo;
+use crate::binding::binding::Binding;
 use crate::binding::binding::Key;
 use crate::export::exports::Export;
 use crate::export::exports::ExportLocation;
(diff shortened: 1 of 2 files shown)


pyrefly/lib/lsp/wasm/completion.rs
```diff
@@ -518,8 +519,12 @@ impl Transaction<'_> {
         {
             for idx in bindings.available_definitions(position) {
                 let key = bindings.idx_to_key(idx);
+                let binding = bindings.get(idx);
                 let label = match key {
                     Key::Definition(id) => module_info.code_at(id.range()),
+                    Key::Import(import) if matches!(binding, Binding::Module(_)) => {
```
kinto0, line 525: hi thanks for the PR! can you help me understand why this new match branch fixes the issue?

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -2323,6 +2323,65 @@ T = spio
     );
 }
 
+#[test]
+fn autoimport_does_not_duplicate_existing_module_import() {
+    let code = r#"
+import json
+
+jso
+#  ^
```
kinto0, line 2332: would you mind testing if these tests failed before the PR?

a good way to make this easy for us is to open a new PR with just the test with the wrong behavior that we can merge immediately. then this PR will update the test and we can be confident it changed

kinto0 (changes requested):

pyrefly/lib/lsp/wasm/completion.rs
kinto0: figured it out - looks like we are adding these imports as local variable completions since the duplication check checks these.

I'm worried about duplication here but we have dedupe logic. testing looks good. I'll import.

kinto0 (approved):

yangdanny97 (approved):

PR #4115, opened 2026-07-11
fix Django: ForeignKey("app_label.Model") string refs are not resolved (no <fk>_id / relation typed Any) #3932

### LINKED ISSUE #3932: Django: ForeignKey("app_label.Model") string refs are not resolved (no <fk>_id / relation typed Any)
### Summary

Pyrefly's Django support resolves a **bare** string `ForeignKey` reference (`ForeignKey("Model")`, target imported) and synthesizes the auto-generated `<fk>_id` attribute. But the **`app_label.ModelName`** form — Django's canonical way to reference a model in another app — is not resolved at all: the relation is typed `Any` and `<fk>_id` is reported as a missing attribute.

### Repro

`other.py`:
```python
from django.db import models

class Target(models.Model):
    name = models.CharField(max_length=10)
```

`m.py`:
```python
from typing import TYPE_CHECKING, reveal_type
from django.db import models

if TYPE_CHECKING:
    from other import Target  # noqa: F401

class M(models.Model):
    bare = models.ForeignKey("Target", on_delete=models.CASCADE)
    applabel = models.ForeignKey("someapp.Target", on_delete=models.CASCADE)

def f(m: M) -> None:
    reveal_type(m.bare_id)      # int   (correct)
    reveal_type(m.applabel_id)  # Unknown + missing-attribute
```

`pyrefly.toml`:
```toml
project-includes = ["m.py", "other.py"]
search-path = ["."]
```

`pyrefly check` (v1.0.0):
```
revealed type: int                                                      # m.bare_id
revealed type: Unknown                                                  # m.applabel_id
ERROR Object of class `M` has no attribute `applabel_id` [missing-attribute]
```

### Expected

`m.applabel_id` should resolve (to the target's pk type), like the bare form does. `"app_label.ModelName"` is the standard Django reference for cross-app foreign keys, so this form is extremely common in real projects.

### Impact

With django-stubs, accessing `<fk>_id` (the query-efficient way to read a FK id without a DB fetch) is flagged as a missing attribute, and the relation attribute loses its type, wherever the FK uses the `app_label.Model` string form.

### Environment
- pyrefly 1.0.0
- django-stubs 6.0.5
- Python 3.12

asukaminato0721:
# Summary

Fixes #3932

Django "app_label.Model" references now resolve using the imported model class name.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/django.rs
+++ b/pyrefly/lib/alt/class/django.rs
@@ -262,9 +262,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 if value.to_str() == "self" {
                     Some(self.instantiate(class))
                 } else {
-                    // Handle forward reference - look up the model by name in the current module
-                    // This requires that the model class is imported or defined in the current module
-                    let class_name = Name::new(value.to_str());
+                    // Django string references may include an app label, but the imported model is
+                    // still looked up by its class name in the current module.
+                    let target = value.to_str();
+                    let class_name = Name::new(
+                        target
+                            .rsplit_once('.')
+                            .map_or(target, |(_, model_name)| model_name),
+                    );
                     let module_name = class.module_name();
 
                     if self.exports.export_exists(module_name, &class_name) {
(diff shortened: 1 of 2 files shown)


kinto0 (approved):

PR #4116, opened 2026-07-11
fix support the + - in hover panel #2974

### LINKED ISSUE #2974: support the + - in hover panel
### Describe the Bug

TypeScript has it, and it can let me control the information dense.

Image

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #2974

Added VS Code +/- hover verbosity controls.

Compact hovers preserve named union aliases; expanded hovers reveal their members.

Added the custom client/server hover protocol while preserving standard LSP behavior.

Raised the extension requirement to VS Code 1.103 for hover verbosity support.

---

why there is an extra `.d.ts`:

Because hover verbosity is still a proposed VS Code API. The stable @types/vscode package does not declare VerboseHover, HoverContext and the fourth provideHover argument.

In TypeScript, there are multiple expand levels.

Here we simplified it to only compact and expand.

# Test Plan

add test

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -29,8 +29,11 @@
         "lint"
     ],
     "engines": {
-        "vscode": "^1.94.0"
+        "vscode": "^1.103.0"
     },
+    "enabledApiProposals": [
+        "editorHoverVerbosityLevel"
+    ],
     "main": "./dist/extension",
     "activationEvents": [
         "onLanguage:python",
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -254,6 +254,7 @@ use ruff_python_ast::name::Name;
 use ruff_text_size::Ranged;
 use ruff_text_size::TextRange;
 use ruff_text_size::TextSize;
+use serde::Deserialize;
 use serde::Serialize;
 use serde::de::DeserializeOwned;
 use serde_json::Value;
@@ -330,7 +331,8 @@ use crate::lsp::non_wasm::workspace::Workspace;
 use crate::lsp::non_wasm::workspace::Workspaces;
 use crate::lsp::wasm::completion::CompletionOptions as CompletionRequestOptions;
 use crate::lsp::wasm::completion::supports_snippet_completions;
-use crate::lsp::wasm::hover::get_hover;
+use crate::lsp::wasm::hover::HoverResult;
+use crate::lsp::wasm::hover::get_hover_with_verbosity;
 use crate::lsp::wasm::notebook::DidChangeNotebookDocument;
 use crate::lsp::wasm::notebook::DidChangeNotebookDocumentParams;
 use crate::lsp::wasm::notebook::DidCloseNotebookDocument;
@@ -794,8 +796,11 @@ fn format_diagnostic_message_for_markdown(message: &str) -> String {
 #[cfg(test)]
 mod tests {
     use lsp_types::CodeActionKind;
+    use lsp_types::InitializeParams;
+    use serde_json::json;
 
     use super::SOURCE_FIX_ALL_PYREFLY;
+    use super::client_uses_custom_hover_provider;
     use super::format_diagnostic_message_for_markdown;
     use super::matches_fix_all_kind;
 
@@ -857,6 +862,16 @@ mod tests {
         )));
         assert!(!matches_fix_all_kind(&CodeActionKind::QUICKFIX));
         assert!(!matches_fix_all_kind(&CodeActionKind::REFACTOR_EXTRACT));
+    }
+
+    #[test]
+    fn test_custom_hover_provider_requires_explicit_opt_in() {
+        let mut params = InitializeParams::default();
+        assert!(!client_uses_custom_hover_provider(&params));
+        params.initialization_options = Some(json!({
+            "pyrefly": {"customHoverProvider": true}
+        }));
+        assert!(client_uses_custom_hover_provider(&params));
     }
 }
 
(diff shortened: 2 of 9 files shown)


lsp/package.json
```diff
@@ -29,8 +29,11 @@
         "lint"
     ],
     "engines": {
-        "vscode": "^1.94.0"
+        "vscode": "^1.103.0"
```
kinto0, line 32: this increases the minimum vscode version by a year (to 2025). I don't think the marketplace lets us see the version distribution. it's probably fine.

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -5024,11 +5068,12 @@ impl Server {
         Ok(transaction.get_signature_help_at(&handle, position))
     }
 
-    fn hover(
+    fn hover_with_verbosity(
```
kinto0: nit: why rename this if we don't have anything called `hover`?

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -1309,7 +1335,15 @@ pub fn capabilities(
             trigger_characters: Some(vec!["(".to_owned(), ",".to_owned()]),
             ..Default::default()
         }),
-        hover_provider: Some(HoverProviderCapability::Simple(true)),
+        hover_provider: Some(HoverProviderCapability::Simple(
+            !initialization_params
+                .initialization_options
+                .as_ref()
+                .and_then(|opts| opts.get("pyrefly"))
+                .and_then(|pyrefly| pyrefly.get("customHoverProvider"))
```
kinto0: we should still provide hover without the selector if the client doesn't have this permission

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -689,6 +699,17 @@ pub fn get_hover(
...
+}
+
+/// Build hover contents, expanding one layer of type aliases per verbosity level.
+pub fn get_hover_with_verbosity(
+    transaction: &Transaction<'_>,
+    handle: &Handle,
+    position: TextSize,
+    show_go_to_links: bool,
+    verbosity_level: usize,
+) -> Option<(Hover, bool)> {
```
kinto0: unclear what this return type is from the signature

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -827,27 +851,42 @@ pub fn get_hover(
...
-            {
-                return display;
+    let (type_display, can_increase_verbosity) =
+        transaction.ad_hoc_solve(handle, "hover_display", {
+            let mut cloned = type_.clone();
+            move |solver| {
+                let mut can_increase = false;
+                cloned.recurse(&mut |child| {
+                    can_increase |= child
+                        .any(|ty| matches!(ty, Type::Union(union) if union.display_name.is_some()));
```
kinto0: I don't entirely understand this. how does presence of a union imply we _are_ allowed to increase verbosity?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -827,27 +851,42 @@ pub fn get_hover(
...
+                });
+                if let Some(owner) = &type_parameter_owner_class
+                    && let Some(display) = type_parameter_hover_display(&solver, &cloned, owner)
+                {
+                    return (display, can_increase && verbosity_level == 0);
```
kinto0: nit: can we factor out the can_icnrease && verbosity_level == 0?

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -827,27 +851,42 @@ pub fn get_hover(
...
+                }
+                if show_constructor
+                    && let Some(display) = class_hover_display(
+                        &solver,
+                        &cloned,
+                        name_for_display.as_deref(),
+                        verbosity_level > 0,
+                    )
+                {
+                    return (display, can_increase && verbosity_level == 0);
```
kinto0: why do we only allow increasing when verbosity level is 0?

kinto0 (changes requested):
cool idea. I like it a lot. I think we're OK with the increase in VSCode version (as long as that's the minimum we can do it on) but I left a few comments

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -828,27 +871,40 @@ pub fn get_hover(
...
-                && let Some(display) =
-                    class_hover_display(&solver, &cloned, name_for_display.as_deref())
-            {
-                return display;
+    let (type_display, can_increase_verbosity) =
+        transaction.ad_hoc_solve(handle, "hover_display", {
+            let mut cloned = type_.clone();
+            move |solver| {
+                // Hover verbosity currently has two states: named nested unions or fully expanded.
+                let expand_unions = verbosity_level > 0;
```
kinto0:
```suggestion
                let unions_expanded = verbosity_level > 0;
```

unclear that this is a state. it sounds like a verb, "should we expand unions"

pyrefly/lib/lsp/wasm/hover.rs
```diff
@@ -828,27 +871,40 @@ pub fn get_hover(
...
-            if show_constructor
-                && let Some(display) =
-                    class_hover_display(&solver, &cloned, name_for_display.as_deref())
-            {
-                return display;
+    let (type_display, can_increase_verbosity) =
+        transaction.ad_hoc_solve(handle, "hover_display", {
+            let mut cloned = type_.clone();
+            move |solver| {
+                // Hover verbosity currently has two states: named nested unions or fully expanded.
```
kinto0: I think this comment belongs in the HoverResult struct so it's more visible. with the suggested rename, I think it is clear

kinto0 (changes requested):
thanks! just a few more nits

yangdanny97 (changes requested):

kinto0:
(a few small things I fixed up in the imported diff, waiting on review again internally)

yangdanny97 (approved):

PR #4133, opened 2026-07-13
Keep public re-exports from excluded files in `pyrefly coverage --public-only`

### LINKED ISSUE #4034: `pyrefly coverage check --public-only`: `--project-excludes` shouldn't exclude public re-exports

jorenham:
# Summary

`--project-excludes` is applied at file discovery, so an excluded module never produced a report, even when a non-excluded public module re-exports from it. `pyrefly coverage {check, report} --public-only` now collects such re-export origins in a second pass, so a symbol is only excluded when **all** of its public re-export locations are excluded.

Fixes #4034

Alternative to #4048

# Test Plan

Regression tests added, and verified that this does not change any of the results for `scipy-stubs` and `numpy`.

### DIFF
--- a/crates/pyrefly_util/src/globs.rs
+++ b/crates/pyrefly_util/src/globs.rs
@@ -1034,7 +1034,11 @@ impl Includes for FilteredGlobs {
     }
 
     fn covers(&self, path: &Path) -> bool {
-        self.includes.covers(path) && !self.filter.is_excluded(path)
+        self.covers_ignoring_excludes(path) && !self.filter.is_excluded(path)
+    }
+
+    fn covers_ignoring_excludes(&self, path: &Path) -> bool {
+        self.includes.covers(path)
     }
 
     fn errors(&mut self) -> Vec<anyhow::Error> {
--- a/crates/pyrefly_util/src/includes.rs
+++ b/crates/pyrefly_util/src/includes.rs
@@ -16,5 +16,7 @@ pub trait Includes {
 
     fn covers(&self, path: &Path) -> bool;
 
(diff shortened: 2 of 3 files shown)


yangdanny97 (approved):

Issue #322, opened 2025-05-20
Incorrect type error when using `unittest.main()`


skybaks:
### Describe the Bug

Pyrefly incorrectly reports an error on the following code:

```py
import unittest
unittest.main()  # ERROR 2:1-14: Expected a callable, got Module[unittest.main]
```

This is valid python and also one recommended way to use the unittest library [as shown in the python docs](https://docs.python.org/3/library/unittest.html#basic-example).

I think what is happening here is that unittest includes [this import in its __init__.py](https://github.com/python/cpython/blob/main/Lib/unittest/__init__.py#L62) which brings the [`main` object here](https://github.com/python/cpython/blob/main/Lib/unittest/main.py#L280) into context and shadows the module named main.py.

It looks like Pyrefly is not taking into account that shadowing of the `from .main import main` and assuming that "main" is referring to the module when, in fact, it is not.

### Sandbox Link

https://pyrefly.org/sandbox/?code=JYWwDg9gTgLgBAVwHbBjApgZxgKGajbAOhAENgkAKASiA

stroxler:
Thanks for the bug report! I think your diagnosis is correct.

We recently (within the past few weeks) updated Pyrefly to model side effects of imports in `__init__.py` modules; those side effects mean that if we had run `from .main import foo`, then `import unittest; unittest.main` would indeed result in the `main` module.

But that behavior only happens when there's no actual `main` name defined directly in the `__init__.py` module, because only in that case does Python fall back to the import system side-effects; in this case it does not.

PR #4139, opened 2026-07-13
Fix false "Expected a callable" on directly-imported package attributes that shadow a submodule (pyrefly #322) (#4139)

anushamukka-dev:
Summary:

When a package `__init__` explicitly re-exports a symbol whose name collides with a same-named submodule, calling that symbol after `import pkg.sub` was wrongly reported as `Expected a callable, got Module[...]`. The canonical case is `import unittest.main; unittest.main()`: `unittest.main` is both a submodule and a re-export (`from .main import main as main`, where `main = TestProgram`), and the call is valid at runtime.

`get_module_attr` shortcut-resolved any directly-imported submodule to the submodule type before consulting the package's exports, so the re-exported callable was never considered. The fix skips that shortcut when the name is an explicit re-export of the package: at runtime the import system binds a directly-imported submodule onto its package only after `__init__` has executed, so a re-export performed while importing the submodule wins. In that case we fall through to the normal export lookup, which resolves `unittest.main` to the `TestProgram` class.

The change is guarded by `is_reexport`, so packages that merely contain a submodule (with no colliding re-export) still resolve `pkg.sub` to the submodule as before. The previously `bug`-marked `test_ambiguous_pkg_attribute_vs_submodule` now passes without the error annotation.

Reviewed By: grievejia

Differential Revision: D110472788

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -2096,6 +2096,32 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         self.apply_getattr_fallback(attr_name, direct_lookup_result)
     }
 
+    /// Whether accessing `attr_name` on the package `module_name` should resolve to the
+    /// directly-imported same-named submodule, shadowing any package export of that name.
+    ///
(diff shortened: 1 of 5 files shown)


grievejia (approved):

PR #4141, opened 2026-07-13
Sort error summary output by count then name

anushamukka-dev:
Summary:
The error summary printed by `print_error_summary` ordered files and per-file error kinds by count alone, leaving ties in a nondeterministic map-iteration order. This made the output unstable and harder to read across runs.

Add a secondary sort key so entries with equal counts break ties by name: per-file error kinds tie-break on `ErrorKind::to_name`, and files tie-break on the path's display string. This makes the summary deterministic and human-readable.

Because `ModulePath` does not implement `Ord`, the file-level tie-break compares display strings, which requires a `Display` bound on the generic `collect_and_sort` helper (its sole caller passes `ModulePath`, which is `Display`). The stale TODO requesting this behavior is removed.

Differential Revision: D111620758

### DIFF
--- a/pyrefly/lib/error/summarize.rs
+++ b/pyrefly/lib/error/summarize.rs
@@ -5,6 +5,9 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+use std::cmp::Reverse;
+use std::fmt::Display;
+
 use dupe::Dupe;
 use pyrefly_python::module_path::ModulePath;
 use pyrefly_util::display;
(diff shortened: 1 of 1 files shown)


NathanTempest:
Thanks for the cleanup. Making the summary deterministic by tie-breaking on name (count desc → name asc) is exactly right, and it resolves the stale TODO it removes. Bonus points for sort_by_cached_key caching the total + path string instead of recomputing per comparison, and Reverse reads much cleaner than the old negation.

One optional follow-up: since this changes ordering behavior, a small unit test on collect_and_sort asserting the count-desc / name-asc tie-break would lock it in (print_error_summary is stderr-only, but the helper returns a Vec, so it's easily testable via a #[cfg(test)] module). Not a blocker but happy to add it on our end. LGTM otherwise.

Issue #3888, opened 2026-06-21
bug(stubgen): statements guarded by conditions other than `TYPE_CHECKING` are dropped entirely


tobyh-canva:
### Describe the Bug

Imports/declarations inside any `if` conditional or `try` / `except` block are generally dropped from stubs. Only statements inside an exact `if TYPE_CHECKING:` guard are extracted.

In particular, the following conditionals likely require special handling in stubgen:

- `if TYPE_CHECKING or SOMETHING_ELSE`: `TYPE_CHECKING` should always be considered `True`, so Pyrefly should extract statements from blocks guarded behind a condition like this, even if they are part of a larger expression ([ref](https://typing.python.org/en/latest/spec/directives.html#type-checking)).
- Module-level `try`/ `except`: Generally, exports in a `try` block represent the package's intended public surface. They may be guarded with `ImportError` or some other exception, but exception handling logic can't realistically be evaluated by type-checkers, so they provide limited value in a stub. I suggest that stubgen extracts statements from the `try` block and flattens them to the module level.

Some examples in the wild:

- multidict `if TYPE_CHECKING or not USE_EXTENSIONS`: [`multidict/__init__.py`](https://sourcegraph.com/r/github.com/aio-libs/multidict/-/blob/multidict/__init__.py?L29)
- GitPython `try` / `except GitError`: [`git/__init__.py`](https://sourcegraph.com/r/github.com/gitpython-developers/GitPython@da05ac6ade70766f388c9c92efcfb977783ea9a6/-/blob/git/__init__.py?L126)
- XGBoost `try` / `except ImportError`: [`xgboost/__init__.py`](https://sourcegraph.com/r/github.com/dmlc/xgboost@8cd8a4a5a4f61e1ad3a5e9002dd9ea7fb759c7e1/-/blob/python-package/xgboost/__init__.py?L23)
- ~~Celery `if STATICA_HACK`: [`celery/__init__.py`](https://sourcegraph.com/r/github.com/celery/celery@aef7f130e3ca547c49d3d581a831927458fc1a71/-/blob/celery/__init__.py?L74)~~ _Should instead be resolved downstream: celery/celery#10409._

#### Input
```python
FLAG = True
if TYPE_CHECKING or not FLAG:
    from pkg._impl import Celery, shared_task
```

#### Expected output
```python
FLAG: bool = True
from pkg._impl import Celery, Task
```

#### Actual output
```python
FLAG: bool = True
```

#### Input
```python
FLAG = True
[...]

PR #4143, opened 2026-07-14
fix(stubgen): extract statements from `try` blocks and expand `TYPE_CHECKING` recognition

tobyh-canva:
# Summary
Updates stubgen to address two of the three limitations identified in #3888:

- `TYPE_CHECKING` guards which are part of a larger boolean expression, e.g. `if TYPE_CHECKING or not USE_EXTENSIONS`, should have statements extracted.
- Statements in `try` blocks should be extracted.

Works towards fixing #3888.

I'd prefer to handle the third limitation (`sys` conditionals) in a separate PR, as it's considerably more complex.

# Test Plan
- [x] Added snapshot tests for stubgen demonstrating the desired behavioural changes
- [x] `python3 test.py`


kinto0 (approved):

PR #4144, opened 2026-07-14
fix(stubgen): emit `from typing import Literal` when `Literal` annotations are added

tobyh-canva:
## Context
Currently, stubgen emits inferred `Literal` annotations, but forgets to add `from typing import Literal` at the top of the file.

Additionally, as shown by the existing `class_vars` snapshot test, enum members are currently incorrectly annotated with `Literal`, whereas the typing spec says enum members should not be annotated at all ([ref](https://typing.python.org/en/latest/spec/enums.html#defining-members)). 

## Changes
- Add snapshot test demonstrating the aforementioned bugs
- Fix the bugs
- Perform a minor code cleanup regarding `typing` imports in stubgen

### DIFF
--- a/pyrefly/lib/stubgen/extract.rs
+++ b/pyrefly/lib/stubgen/extract.rs
@@ -10,6 +10,7 @@
 //! Walks the module's AST in source order and uses the binding/answer
 //! system to resolve types for each declaration.
 
+use std::collections::BTreeSet;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


pyrefly/lib/stubgen/extract.rs
```diff
@@ -58,13 +59,8 @@ pub struct ModuleStub {
     /// Whether any item uses `Incomplete` (so we know whether to
     /// emit `from _typeshed import Incomplete`).
     pub uses_incomplete: bool,
-    pub uses_self: bool,
-    /// Whether any item renders a `Callable[...]` annotation (so we know
-    /// whether to emit `from typing import Callable`).
-    pub uses_callable: bool,
-    /// Whether any item renders a `ClassVar[...]` annotation (so we know
-    /// whether to emit `from typing import ClassVar`).
-    pub uses_classvar: bool,
```
tobyh-canva, line 67: Since this PR would have required me adding another bool `uses_literal`, I figured I might as well take this opportunity to consolidate all these bools into a set `typing_imports`.

But please do let me know if the bools are preferred

stroxler (approved):

PR #4146, opened 2026-07-14
Add `[coverage]` config overrides for the `pyrefly coverage` globs

jorenham:
# Summary

`pyrefly coverage {check, report}` used the same `project-{includes,excludes}` as `pyrefly check`, so measuring coverage on a different file set (e.g. skipping tests, vendored code, or publically-namespaced private API) required repeatedly specifiying `--project-excludes` for every `pyrefly coverage` call. 

This adds an optional `[coverage]` config table with `includes` and `excludes` that override the project globs when using `pyrefly coverage` (and only then).
Each field defaults to its top-level `project-` counterpart when unset (like it does now).
The `--project-excludes` flag takes precedence over `coverage.excludes`, and positional `FILES...` args also do (just like with the `project-` globs).
Unknown keys inside `[coverage]` are reported in a warning message in the same way as with the other config tables.

# Test Plan

Config tests added for parsing, path rewriting, and glob resolution (for both the "default" and "coverage" config scopes), and verified that a `[coverage]` config indeed narrows `pyrefly coverage report` without affecting `pyrefly check` and explicit `FILES...` ("file mode" I believe it's called).

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -53,12 +53,14 @@ use pyrefly_util::telemetry::TelemetrySourceDbRebuildStats;
 use pyrefly_util::watch_pattern::WatchPattern;
 use serde::Deserialize;
 use serde::Serialize;
+use serde_with::skip_serializing_none;
 use starlark_map::small_map::SmallMap;
 use starlark_map::small_set::SmallSet;
 use tracing::debug;
 use tracing::error;
 
 use crate::base::ConfigBase;
+use crate::base::ExtraConfigs;
 use crate::base::InferReturnTypes;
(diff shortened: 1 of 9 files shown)


jorenham:
cc @MarcoGorelli

NathanTempest (commented):
Thanks @jorenham for the clean, well-tested addition. 

Since this adds a new top-level [coverage] config table, I am looping in @connernilsen and @kinto0 for a quick review of the table design and precedence model. Assuming they're happy with it, the implementation looks ready.

kinto0 (approved):
LGTM since there doesn't seem to be another way to model these settings for coverage

yangdanny97 (approved):

PR #4147, opened 2026-07-14
Dedupe redundant serde attributes in `pyrefly_config`

jorenham:
# Summary

Every `Option` field in the config structs repeated `#[serde(default, skip_serializing_if = "Option::is_none")]`, and some `flatten` and `skip` fields also had a `default` that serde ignores or already implicitly uses.
This updates the config structs to use `#[skip_serializing_none]` from `serde_with` for the entire struct, and removes the redundant serde stuff. 

Spotted while working on #4146.

# Test Plan

Tests still pass; no behavior changes.


yangdanny97 (approved):

Issue #2296, opened 2026-02-03
Profile autocompletion speed


yangdanny97:
### Describe the Bug

import polars as pl
pl.d autocompletion

yangdanny97:
Based on profiling specifically for the `pl.` example, the vast majority of the time is being spent calculating builtins autoimport completions.

kz357:
#claim

PR #4148, opened 2026-07-14
Reuse fuzzy matcher across completion candidates

kz357:
# Summary

add_builtins_autoimport_completions and add_local_variable_completions constructed a new SkimMatcherV2 for every candidate; SkimMatcherV2 carries thread-local caches meant to be reused across fuzzy_match calls so rebuilding it per candidate allocates those caches repeatedly and prevents DP-matrix cache from ever being reused. Matcher is taken out of the loop and reused, similar to state/lsp.rs.

I think this will speed up the completion popup appearance, but I don't know if that's true or by how much as I didn't add a benchmark and neither typecheck_benchmark.py nor lsp_benchmark.py measures this. The change doesn't break anything in cargo test

Related: issue #2296

# Test Plan

cargo test passes the same as before edits, cargo clippy and cargo fmt are clean

I don't think there's an existing benchmark that covers completions, but I'm happy to try to write one as a follow-up to this if it'd be useful

### DIFF
--- a/pyrefly/lib/lsp/wasm/completion.rs
+++ b/pyrefly/lib/lsp/wasm/completion.rs
@@ -455,10 +455,10 @@ impl Transaction<'_> {
             .finding()
         {
             let builtin_exports = self.get_exports(&builtin_handle);
+            let matcher = SkimMatcherV2::default().smart_case();
             for (name, location) in builtin_exports.iter() {
                 if let Some(identifier) = identifier
-                    && SkimMatcherV2::default()
-                        .smart_case()
+                    && matcher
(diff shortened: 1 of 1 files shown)


NathanTempest (approved):
Thanks @kz357 for the nice catch. Behavior is identical (same .smart_case(), same fuzzy_match calls) and it can't regress, so I'm comfortable landing it without a benchmark. If you do want to add a completions benchmark as a follow-up, that'd be a welcome addition to catch future regressions here. Other than that looks good for importing.

yangdanny97 (approved):

PR #4149, opened 2026-07-14
Print config warnings in `pyrefly coverage`

jorenham:
# Summary

Config warnings are printed when using `pyrefly check` (e.g. invalid `site-package-path`), but the `pyrefly coverage` commands didn't. So now they do :)

Spotted while working on #4146.

# Test Plan

Scrut test added, and manually verified that `pyrefly coverage report`'s stdout stays valid json.

### DIFF
--- a/crates/pyrefly_config/src/finder.rs
+++ b/crates/pyrefly_config/src/finder.rs
@@ -229,6 +229,13 @@ impl ConfigFinder {
         self.errors.lock().extend(errors);
     }
 
+    /// Print and clear all accumulated config errors.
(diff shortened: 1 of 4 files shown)


NathanTempest (approved):
Thank you @jorenham for the clean little fix. Factoring the repeated print loop into print_errors() and adding it to the coverage path is the right call, and since errors() drains via mem::take, this stays behavior-preserving with no double-printing. 
Nice that the scrut test pins the warnings to stderr so coverage report's stdout stays valid JSON. Importing.

yangdanny97 (approved):

PR #4151, opened 2026-07-14
fix(stubgen): extract unpack assignments

tobyh-canva:
## Context
Stubgen currently skips unpack assignments, for example:

```py
A, B = 1, 2
```
doesn't emit anything in stubs.

It should emit something like:

```py
from typing import Literal

A: Literal[1]
B: Literal[2]
```

Here's an example in the wild where unpack assignments are used to create public `ClassVar` constants: [`WordNetCorpusReader` from `nltk`](https://sourcegraph.com/r/github.com/nltk/nltk/-/blob/nltk/corpus/reader/wordnet.py?L1125)

## Changes
- Handle unpack assignments in Stubgen's assignment extraction machinery.

# Test Plan
- [X] Added new stubgen snapshot test describing the desired behaviour

### DIFF
--- a/pyrefly/lib/stubgen/extract.rs
+++ b/pyrefly/lib/stubgen/extract.rs
@@ -31,6 +31,7 @@ use pyrefly_types::types::Overload;
 use pyrefly_types::types::OverloadType;
 use pyrefly_types::types::Type;
 use ruff_python_ast::Expr;
+use ruff_python_ast::ExprName;
 use ruff_python_ast::Operator;
 use ruff_python_ast::Stmt;
 use ruff_python_ast::StmtClassDef;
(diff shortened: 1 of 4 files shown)


NathanTempest (commented):
Thanks @tobyh-canva for the nice fix. Stubgen was silently dropping unpack targets because extract_assign only handled Expr::Name targets, and this correctly extends it to tuple/list targets. What I especially like is that it sources each name's type from its own binding rather than the RHS, so A, B = 1, "module" gives A: Literal[1] / B: Literal['module'] instead of mis-attaching the whole tuple and the snapshot covers all the cases that usually trip this up: private-name filtering (_PRIVATE dropped), starred targets (*REST → list[str]), class-body ClassVar wrapping, and enum members. The nltk WordNetCorpusReader example is a good real-world motivation.

One optional follow-up (non-blocking): nested unpack like (A, B), C = ... isn't covered by the test — worth confirming it either recurses or is safely skipped rather than mis-emitting. Happy to leave that as a follow-up. Looks good for importing

yangdanny97 (approved):

PR #4152, opened 2026-07-15
fix support creation of union from generic class constructor #1134

### LINKED ISSUE #1134: support creation of union from generic class constructor
### Describe the Bug

```py
class A[T]:
    def __init__(self, value: T):
        self._value = value

def f(value: int | str):
    _: A[int] | A[str] = A(value)  # Argument `int | str` is not assignable to parameter `value` with type `int` in function `A.__init__`
```

here, this assignment is completey safe and valid, i would expect no error

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1134

Generic constructors now expand union arguments against union type hints and union successful specializations, and all expansions must match, preventing unsafe mixed-argument cases.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -21,6 +21,7 @@ use pyrefly_types::types::CalleeKind;
 use pyrefly_types::types::NNModuleType;
 use pyrefly_types::types::TArgs;
 use pyrefly_types::types::TParams;
+use pyrefly_util::owner::Owner;
 use pyrefly_util::prelude::SliceExt;
 use pyrefly_util::prelude::VecExt;
 use ruff_python_ast::Arguments;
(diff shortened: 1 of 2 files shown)


NathanTempest (commented):
Really nice work @asukaminato0721 and I like that it reuses the existing ArgsExpander, gates on MAX_CALL_HINT_WIDTH, and tries the unexpanded path first for the common case. The mypy_primer result is a great signal: it removes a real false positive in rotki with no new errors. And the negative Pair(x, y) test is exactly the case I'd worry about correctly rejecting the mixed-argument union is the right call.

Since this is core constructor-specialization semantics, I'm looping in @stroxler  to sign off on the "every expansion must match a union member" rule across the trickier interactions, constrained TypeVars, __init__ overloads, and hints like A[int] | A[str] | None. Assuming that's sound, this looks ready.

PR #4156, opened 2026-07-15
Always print coverage findings in `pyrefly coverage check --fail-under=`

jorenham:
# Summary

Right now, `pyrefly coverage report --fail-under=<something below 100>` does not print the findings if the coverage is above the threshold. I think it would be more useful if those findings would be printed regardless of whether it passes (i.e. exits 0). And if the user doesn't want to see those findings for some reason, then they could always use the existing `--output-format=omit-errors` flag to disable this.

# Test Plan

Manually verified that the missing coverage warnings are now always  emitted.

### DIFF
--- a/pyrefly/lib/commands/coverage/check.rs
+++ b/pyrefly/lib/commands/coverage/check.rs
@@ -48,8 +48,8 @@ pub struct CheckArgs {
     public_only: bool,
 
     /// Format for the untyped-symbol findings.
-    #[arg(long, value_enum)]
-    output_format: Option<OutputFormat>,
+    #[arg(long, value_enum, default_value_t)]
+    output_format: OutputFormat,
 }
 
 impl CheckArgs {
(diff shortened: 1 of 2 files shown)


NathanTempest (commented):
As always, thank you for your efforts in improving Pyrefly coverage @jorenham. The implementation is clean (nice to drop the Option<OutputFormat> in favor of a clap default_value_t, and moving write_errors_to_console above the branch keeps findings→summary ordering consistent), and the docs update is thorough. The reasoning is reasonable too: findings go to stdout, the summary to stderr, and --output-format=omit-errors is the escape hatch, so exit-code-only CI is unaffected.

The one thing I want to confirm before merging is the default-behavior change itself. Printing findings even when coverage passes changes the default output for everyone, and on large codebases that could be a lot of output on every green run. Looping in @maggiemoss  for a quick sign-off on making that the default vs. keeping quiet-on-success. Assuming she's good with it, this is ready to go.

jorenham:
> The one thing I want to confirm before merging is the default-behavior change itself. Printing findings even when coverage passes changes the default output for everyone, and on large codebases that could be a lot of output on every green run.

Yea good point. I don't expect that this will be a problem for CI though, since it still exits 0 like it used to in those cases. 
Also, `pyrefly coverage` is very new, so I doubt that many large projects will be using it at this point. The only large codebases that I'm aware of use this are numpy and scipy-stubs, and both have 100% type-coverage.

stroxler (approved):

PR #4157, opened 2026-07-15
unify module prefix so __unknown__ files count under --public-only (#4157)

kinto0:
Summary:

Make coverage's `module_prefix()` unconditional so `__unknown__` fallback modules get the `__unknown__.` prefix like every other module.

Why: under `--public-only`, these files silently reported "0 of N typable" — symbols were stored bare while the public-FQN filter expects the prefixed form, so all got dropped.

Reviewed By: yangdanny97

Differential Revision: D112166629

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -1033,15 +1033,9 @@ impl ConfigFile {
         })
     }
 
+    /// Create a `Handle` for the given path, deriving its module name from the search paths,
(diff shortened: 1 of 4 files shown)


jorenham (commented):
I like this!

I can't officially press the "approve" button, so here's an emoji of a penguin that I hereby declare to be isomorphic to an official review approval: 🐧

yangdanny97 (approved):

jorenham (approved):
Apparently I can now officially approve (sorry penguin)

PR #4158, opened 2026-07-15
solve generic function converters against the field type

### LINKED ISSUE #4155: Unbound TypeVar leak into synthesized `__init__` from generic attrs converter
### Describe the Bug

When using a generic function like `copy.deepcopy` (which has the signature `def deepcopy(x: _T, ...) -> _T` in typeshed) as an `attrs` field converter, Pyrefly leaks the converter's `TypeVar` (`_T`) into the synthesized `__init__` signature, causing type checking failures at callsites.

#### Repro:
```python
import attr
import copy

@attr.s(auto_attribs=True)
class MyJob:
    req: int = attr.ib(converter=copy.deepcopy)

j = MyJob(req=42)
```

#### Error:
```
ERROR Argument `Literal[42]` is not assignable to parameter `req` with type `_T` in function `MyJob.__init__` [bad-argument-type]
```
`reveal_type(MyJob.__init__)` outputs `(self: MyJob, req: _T) -> None` instead of `(self: MyJob, req: int) -> None`.

#### Root cause:
In `pyrefly/lib/alt/class/dataclass.rs`:
`get_converter_param` only solves type parameters if the converter is a `ClassDef` (via `solve_generic_class_converter`). For generic functions like `copy.deepcopy`, it falls through and returns the raw parameter type (`_T`) verbatim.

### Sandbox Link

_No response_

_No response_

shobhitmehro:
Summary:
Goal: A generic function used as an attrs converter, such as copy.deepcopy, no longer produces a spurious error when you construct the class.

Background: attrs lets a field declare a converter that transforms the constructor argument before it is stored, and pyrefly types the generated `__init__` parameter as the converter's input type. When the converter is a generic function like, whose type is `(_T) -> _T`, that input is a bare type variable `_T` that escapes its function scope, so every argument was rejected with a message about a parameter of type `_T`. 

How: We add a parallel arm that recognizes a generic function converter and solves it the same way the class path does, by instantiating the function with fresh variables, constraining its return type against the declared field type, and reading back the solved input as the `__init__` parameter. A variable that cannot be solved is finalized to a gradual type so uncertain cases fall back instead of erroring.

fixes https://github.com/facebook/pyrefly/issues/4155

Differential Revision: D112165450

### DIFF
--- a/pyrefly/lib/alt/class/dataclass.rs
+++ b/pyrefly/lib/alt/class/dataclass.rs
@@ -59,6 +59,7 @@ use crate::types::keywords::ConverterMap;
 use crate::types::keywords::DataclassFieldKeywords;
 use crate::types::keywords::TypeMap;
 use crate::types::literal::Lit;
+use crate::types::types::Forallable;
 use crate::types::types::Type;
 
 /// Which constructor-copy builtin `call_dataclasses_replace` is serving. Chosen by the caller so the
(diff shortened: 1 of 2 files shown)


yangdanny97 (approved):

PR #4159, opened 2026-07-15
Fix first auto-import completion for unloaded modules

### LINKED ISSUE #4108: completions auto-import doesn't work on first try after reload
### Describe the Bug

ctrl+space to force completions. notice pyrefly only shows them after an import is typed

https://github.com/user-attachments/assets/14fb89b7-d7ff-4517-852d-625b6a233bf5

### Sandbox Link

_No response_

_No response_

paranoa233:
## Summary

- include modules discoverable from the active file's import paths in fuzzy module auto-import completion
- deduplicate discoverable modules with modules already loaded in the transaction
- add a regression test for completing `json` before the module has been imported or loaded

## Root cause

Module auto-import completion only searched `Transaction::modules()`, which contains modules already loaded into state. A standard-library or site-package module therefore appeared only after an explicit import caused it to be loaded. The existing import-prefix resolver already discovers matching modules from the active config's search paths and typeshed, so the fix merges those candidates into the fuzzy search.

## Testing

- `cargo test -p pyrefly autoimport --lib` (15 passed)
- `cargo fmt -p pyrefly -- --check`
- `python test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` (run with Python 3.11 and the GNU Rust toolchain on Windows)

Fixes #4108

This contribution was prepared by an AI coding agent (Codex) working through the contributor's account.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2969,11 +2969,15 @@ impl<'a> Transaction<'a> {
         .map(|item| TextRangeWithModule::new(item.module, item.definition_range))
     }
 
-    pub(crate) fn search_modules_fuzzy(&self, pattern: &str) -> Vec<ModuleName> {
+    pub(crate) fn search_modules_fuzzy(&self, handle: &Handle, pattern: &str) -> Vec<ModuleName> {
         let matcher = SkimMatcherV2::default().smart_case();
         let mut results = Vec::new();
+        // `self.modules()` only contains modules that have already been loaded. Include modules
+        // discoverable from the active file's import paths so auto-import works on the first try.
+        let mut module_names: HashSet<_> = self.modules().into_iter().collect();
+        module_names.extend(self.import_prefixes(handle, ModuleName::from_str(pattern)));
 
-        for module_name in self.modules() {
+        for module_name in module_names {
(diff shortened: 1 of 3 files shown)


kinto0:
Thanks for the PR! interesting diagnosis. I'm curious if you could attach a before/after video. My impression was that this would always happen on the first completion, regardless of module load state. In other words, I could wait until modules were all loaded before requesting completions and I would still see the bug. If that's the case, could it be an invalidation issue (if we need to invalidate on a new event instead of attaching a handle that would be the more targeted fix)?

pyrefly/lib/state/lsp.rs
```diff
@@ -2969,11 +2969,15 @@ impl<'a> Transaction<'a> {
         .map(|item| TextRangeWithModule::new(item.module, item.definition_range))
     }
 
-    pub(crate) fn search_modules_fuzzy(&self, pattern: &str) -> Vec<ModuleName> {
+    pub(crate) fn search_modules_fuzzy(&self, handle: &Handle, pattern: &str) -> Vec<ModuleName> {
```
kinto0, line 2972: this interface is unclear without a doc-comment. what handle do we need to provide for searching global modules?

stroxler (approved):

PR #4172, opened 2026-07-16
Build WASM artifact with release binaries

yangdanny97:
Summary: Add the pyrefly WASM build to the active build_binaries workflow so it is produced and merged into the dist artifact with the existing wheels and sdist.

Differential Revision: D112422149


kinto0 (approved):

Issue #4171, opened 2026-07-16
LSP recheck can panic after an export change propagates through a deep dependency chain


teerthsharma:
# LSP recheck can panic after an export change propagates through a deep dependency chain

Pyrefly commit: `c195e9fbb5c47394987dfd9d1221b1ff27a205ec` (`1.2.0-dev.2`), release build, Windows, four worker threads.

No matching open issue or PR was found for the panic text, epoch warning, or incremental dependency propagation failure.

## Observed behavior

A saved type change propagated through a generated chain of modules. With 192 two-module SCC layers, the update completes. At 200 layers, the first updated diagnostic arrives, but the following save panics. At 208 or more layers, the first save panics.

The terminal warning and panic are:

```text
Exceeded maximum epochs (100) without stabilizing. ... Forcing invalidation.
Transaction has uncommitted changes
```

The panic is at `State::commit_transaction` in `pyrefly/lib/state/state.rs`.

## Runtime trace

With 32 layers, verbose logging shows 35 epochs and one changed export per epoch:

```text
leaf -> a_0 -> a_1 -> ... -> a_31 -> main -> stabilization
```

With 192 layers, the first run reaches epoch 100 after `a_98`, invokes the fallback, and a second run needs 94 epochs to reach `main` and stabilize.

The coarse fallback has already transitively marked `a_99` through `main` dirty, but its single `run_step` changes only `a_99`. The remaining closure was processed before its dependencies became current and cannot be re-dirtied in that same epoch.

The same 100 + 94 pattern occurs with one worker, ruling out parallel races as a necessary cause and making a sequential dependency-first fallback a viable minimum-code experiment.

## Source-level cause

When `demand` observes changed exports, affected direct rdeps are added to `TransactionData.dirty`. `run_step` transfers that set into `TaskHeap` only at the beginning of the next epoch, so a long path advances roughly one edge per epoch.

After `MAX_EPOCHS`, `run_internal` calls `invalidate_rdeps` and performs one final `run_step`. That final step can record more changed exports, but the fallback returns without draining them. The later commit assertion then rejects the transaction.

The cap and fallback were introduced by `c6ca82ee2` as defense-in-depth for fine-grained cycle detection. Existing tests cover overlapping cycles, independent export changes, and short re-export chains. They do not cover a finite acyclic chain that exceeds the cap.

## Expected behavior

Deep but finite acyclic propagation should converge or fail with a controlled error; it should not leave a committable transaction in an internally inconsistent state.

[...]

PR #4180, opened 2026-07-17
test: capture capped recheck propagation panic

teerthsharma:
## Summary

- let the incremental test harness configure an explicit in-memory module set
- generate a finite chain of 208 two-module SCCs inside one Rust test
- record the current `Transaction has uncommitted changes` failure with `#[should_panic]`

## Why

An export change can consume the 100-epoch incremental budget, produce another export change during forced invalidation, and then reach commit with that change still pending. The generated graph exercises the real incremental transaction path without external fixtures or dependencies.

This is deliberately a synthetic defense-in-depth reproducer and contains no scheduler fix. The assertions after the failing call describe the behavior a separate fix should restore.

## Verification

- without `#[should_panic]`, the focused test fails at commit with `Transaction has uncommitted changes`
- `cargo test -p pyrefly --lib test_deep_scc_chain_stabilizes_after_epoch_cap`
- `cargo test -p pyrefly --lib test::incremental` (75 passed)

Refs #4171

### DIFF
--- a/pyrefly/lib/test/incremental.rs
+++ b/pyrefly/lib/test/incremental.rs
@@ -39,6 +39,7 @@ struct IncrementalData(Arc<Mutex<SmallMap<ModuleName, Arc<String>>>>);
 /// Helper for writing incrementality tests.
 struct Incremental {
     data: IncrementalData,
+    files: Vec<String>,
     require: Option<Require>,
     state: State,
     to_set: Vec<(String, String)>,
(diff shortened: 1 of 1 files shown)


kinto0:
thank you very much! it's taken me a sec to understand this. I've graphed the example out below:
image

  1. The recheck propagates the change one layer per epoch: epoch 2 updates a_0, epoch 3 updates a_1, …
  2. After 100 epochs (MAX_EPOCHS), propagation has only crawled to ~`a_98— nowhere nearmainata_207`. The chain is longer than the budget.
  3. The scheduler gives up its normal loop and runs a fallback (invalidate everything downstream, do one final pass).
  4. That fallback pass is a single epoch, so it only advances one more edge — it does not finish propagating. It leaves pending changes recorded.
  5. The recheck then tries to commit, which asserts "no pending changes." There are pending changes → panic ("Transaction has uncommitted changes").

this trigger was from the chain > 100 (though there could likely be other triggers) and the bug exists in the fallback (it breaks the invariant)

stroxler (approved):

PR #4184, opened 2026-07-17
fix find-references sometimes does go-to-def #2574

### LINKED ISSUE #2574: find-references sometimes does go-to-def
### Describe the Bug

```
from xxx import yyy as zzz


zzz.f()
```

right click on zzz, Go to references

And it should jump to zzz.f

But it jump to yyy file.

### Sandbox Link

_No response_

0.54.0

asukaminato0721:
# Summary

Fixes #2574

Module bindings now retain whether they use an explicit alias.
Find-references stops at renamed module imports instead of navigating into the imported module.

# Test Plan

Added regression coverage for both from x import y as z and import x.y as z.


rchen152 (approved):
LGTM!

kinto0 (approved):

PR #4185, opened 2026-07-17
fix pd.read_csv inferred type is Unknown but go to definition works. #1819

### LINKED ISSUE #1819: `pd.read_csv` inferred type is Unknown but go to definition works.
### Describe the Bug

Image

`pd.read_csv` becomes unknown but go to definition works.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #1819

Root cause: bundled stubs could resolve foo.pyi instead of `foo/__init__.pyi`.

For pandas, that selected an empty pandas.io.parsers.pyi, losing the read_csv export.

Bundled resolution now matches Python’s package-first precedence across all bundled stub sources.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/module/bundled.rs
+++ b/pyrefly/lib/module/bundled.rs
@@ -34,6 +34,17 @@ pub fn set_readonly(path: &Path, value: bool) -> anyhow::Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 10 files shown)


samwgoldman:
Closing in favor of #4286 which fixes the underlying issue in a more general way

PR #4188, opened 2026-07-17
Fix deep-chain recheck panic in the MAX_EPOCHS fallback

kinto0:
Summary:
An export change propagating through a dependency chain deeper than the
100-epoch cap panicked at commit with `Transaction has uncommitted changes`
(https://github.com/facebook/pyrefly/issues/4171). The fine-grained recheck
path advances one dependency edge per epoch, so a long chain exhausts the cap
before the change reaches the top importer.

The coarse fallback was meant to invalidate the whole affected closure and
recompute it, but it seeded `invalidate_rdeps` from `changed` — which the final
loop iteration had already drained — making it a no-op. A single `run_step`
then left `changed`/`dirty` populated, tripping the `commit_transaction`
invariant.

Seed the coarse invalidation from the pending `dirty` frontier instead, then
recompute to a fixpoint via `stabilize_after_invalidate`. This converges
quickly because lookups pull their dependencies: once the closure is
invalidated, a couple of `run_step`s recompute it with correct values and
drain the bookkeeping the commit invariant requires. A bounded loop with a
final residual-clear keeps a pathological oscillation from hanging or
committing inconsistent state.

Differential Revision: D112591162

### DIFF
--- a/pyrefly/lib/state/state.rs
+++ b/pyrefly/lib/state/state.rs
@@ -2101,18 +2101,32 @@ impl<'a> Transaction<'a> {
         // cases. This avoids false positives where independent exports happen to
         // be processed in the same module across different epochs.
         //
-        // As a defense-in-depth measure, we also cap the total number of epochs to prevent
-        // runaway computation in case of unforeseen edge cases.
+        // As a defense-in-depth measure, we also cap the number of fine-grained epochs. On
+        // a cycle or on hitting the cap we abandon fine-grained propagation and switch to
+        // coarse mode: invalidate the entire rdep closure of the modules that just changed,
+        // then keep recomputing until stable. Coarse mode settles in a few epochs (unlike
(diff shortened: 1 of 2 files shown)


pyrefly/lib/state/state.rs
```diff
@@ -2146,16 +2158,17 @@ impl<'a> Transaction<'a> {
...
-             This may indicate an unexpected dependency pattern. Forcing invalidation."
+            "Recheck did not stabilize within {} epochs; clearing residual state to commit consistently.",
+            2 * MAX_EPOCHS
         );
-        let changed = mem::take(&mut *self.data.changed.lock());
-        self.invalidate_rdeps(changed.into_map(|(m, _)| m));
-        self.run_step(handles, require, custom_thread_pool)
+        self.data.changed.lock().clear();
+        self.data.dirty.lock().clear();
+        Ok(())
```
kinto0, line 2171: this will hide bugs - we can't have this fallback

PR #4189, opened 2026-07-17
Add test for NotIterable error in starred tuple unpacking (#4189)

anushamukka-dev:
Summary:

The error path in `expr.rs` that reports `Expected an iterable, got ...`
when a non-iterable type is used with star (`*`) unpacking in a tuple
display (e.g. `(*x,)`) carried a `// TODO: missing test` marker and had no
dedicated test coverage.

This adds `test_unpack_starred_not_iterable_in_tuple`, which unpacks an `int`
into a tuple display and expects the `NotIterable` error, and removes the
now-resolved TODO comment.

Differential Revision: D112538198


yangdanny97 (approved):

PR #4191, opened 2026-07-17
Pyrefly: surface secondary annotations as LSP DiagnosticRelatedInformation (#4191)

anushamukka9:
Summary:

Pyrefly attaches secondary annotations to type errors (e.g. labeling both operands of an unsupported binary operation with their types), and these already render in the CLI's source snippets. In the language server, however, `Error::to_diagnostic()` dropped them entirely — a lingering TODO noted that mapping them to `DiagnosticRelatedInformation` required constructing a `Url` from the module path, "which may not always succeed."

Secondary annotations always live in the same file as the primary error, so we can reuse the error's own module path for their locations. This maps each `SecondaryAnnotation` to a `DiagnosticRelatedInformation` — its label becomes the message and `Module::to_lsp_range` converts the span — and populates `Diagnostic::related_information`. LSP clients can now show the extra context (e.g. "has type `int | str`") alongside the diagnostic and let users jump to each labeled span.

`Url::from_file_path` requires an absolute path and can fail; when it does (or when there are no annotations) we leave `related_information` as `None` rather than dropping the diagnostic, so nothing regresses for modules without a materializable on-disk path.

Addresses https://github.com/facebook/pyrefly/issues/105

Differential Revision: D111240436

### DIFF
--- a/pyrefly/lib/error/error.rs
+++ b/pyrefly/lib/error/error.rs
@@ -14,7 +14,9 @@ use std::path::Path;
 use itertools::Itertools;
 use lsp_types::CodeDescription;
 use lsp_types::Diagnostic;
+use lsp_types::DiagnosticRelatedInformation;
 use lsp_types::DiagnosticTag;
+use lsp_types::Location;
 use lsp_types::Url;
 use pyrefly_python::ignore::Tool;
 use pyrefly_python::module::Module;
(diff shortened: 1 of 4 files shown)


kinto0:
cool! looks like there's still a windows test failing. do you have any screenshots of how it looks in the IDE?

kinto0:
is this ready for review? looks like it's a draft on phabricator

anushamukka-dev:
My apologies - auto publish didn't work for some reason, checking the reason now, will publish soon! @kinto0

anushamukka-dev:
updated and waiting for review - D111240436

rchen152:
FYI - I think there's a typo in the PR description. It points to issue #105, which is already closed.

PR #4200, opened 2026-07-18
feat(suppress): support removal of `# type: ignore` suppressions with `pyrefly suppress --remove-unused`

### LINKED ISSUE #3984: Feature: `--remove-unused-type-ignores` flag to clean up unused `# type: ignore` comments

tobyh-canva:
## Summary
Changes `pyrefly suppress --remove-unused` and `pyrefly check --remove-unused-ignores` so that users can specify `--remove-unused=type` / `--remove-unused=all` to instead / additionally remove `# type: ignore` comments, rather than just `# pyrefly: ignore` comments.

The default remains unchanged, so as to not surprise users previously relying on `pyrefly suppress --remove-unused` suddenly removing load-bearing `# type: ignore` comments.

Resolves #3984.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -77,6 +77,7 @@ use crate::error::summarize::print_error_summary;
 use crate::error::suppress;
 use crate::error::suppress::CommentLocation;
 use crate::error::suppress::SerializedError;
+use crate::error::suppress::UnusedIgnoreKind;
 use crate::module::typeshed::stdlib_search_path;
 use crate::report;
 use crate::state::load::FileContents;
@@ -232,8 +233,7 @@ impl SnippetCheckArgs {
                 check_all: false,
                 suppress_errors: false,
                 expectations: false,
-                remove_unused_ignores: false,
-                remove_unused_type_ignores: false,
+                remove_unused_ignores: None,
             },
         };
         let (status, check_result) =
@@ -418,12 +418,17 @@ struct BehaviorArgs {
     /// Check against any `E:` lines in the file.
     #[arg(long)]
     expectations: bool,
-    /// Remove unused ignores from the input files.
-    #[arg(long)]
-    remove_unused_ignores: bool,
-    /// Remove unused `# type: ignore` comments in addition to unused Pyrefly ignores.
-    #[arg(long)]
-    remove_unused_type_ignores: bool,
+    /// Remove unused ignores from the input files, optionally selecting `pyrefly`, `type`, or `all`.
+    /// Defaults to `pyrefly` when no kind is specified.
+    #[arg(
+        long,
+        value_enum,
+        value_name = "KIND",
+        num_args = 0..=1,
+        require_equals = true,
+        default_missing_value = "pyrefly"
+    )]
+    remove_unused_ignores: Option<UnusedIgnoreKind>,
 }
 
 fn write_errors_to_file(
@@ -1319,14 +1324,11 @@ impl CheckArgs {
                 .collect();
             suppress::suppress_errors(serialized_errors, CommentLocation::LineBefore);
         }
-        if self.behavior.remove_unused_ignores || self.behavior.remove_unused_type_ignores {
+        if let Some(kind) = self.behavior.remove_unused_ignores {
             // TODO: Deprecate this in favor of `pyrefly suppress`
             let collected = loads.collect_errors();
             let unused_errors = loads.collect_unused_ignore_errors(&collected);
-            suppress::remove_unused_ignores(
-                unused_errors,
-                self.behavior.remove_unused_type_ignores,
-            );
+            suppress::remove_unused_ignores(unused_errors, kind);
         }
 
         // We update the baseline file if requested, after reporting any new
@@ -1707,6 +1709,35 @@ mod tests {
         output.inherit_defaults_from_config(&config);
 
         assert_eq!(output.output_format(), OutputFormat::Json);
+    }
+
+    #[test]
+    fn remove_unused_ignores_cli_values() {
+        for (argument, expected) in [
+            (None, None),
+            (
+                Some("--remove-unused-ignores"),
+                Some(UnusedIgnoreKind::Pyrefly),
+            ),
+            (
+                Some("--remove-unused-ignores=pyrefly"),
+                Some(UnusedIgnoreKind::Pyrefly),
+            ),
+            (
+                Some("--remove-unused-ignores=type"),
+                Some(UnusedIgnoreKind::Type),
+            ),
+            (
+                Some("--remove-unused-ignores=all"),
+                Some(UnusedIgnoreKind::All),
+            ),
+        ] {
+            let args = argument.map_or_else(
+                || CheckArgs::parse_from(["check"]),
+                |argument| CheckArgs::parse_from(["check", argument]),
+            );
+            assert_eq!(args.behavior.remove_unused_ignores, expected);
+        }
     }
 
     fn upsell_string(reason: SynthesizedPresetReason) -> String {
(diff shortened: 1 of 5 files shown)


pyrefly/lib/error/suppress.rs
```diff
@@ -553,27 +562,39 @@ fn update_ignore_comment_with_used_codes(
...
 }
 
 /// Removes unused ignore comments from source files using SerializedError.
 /// This is similar to remove_unused_ignores but works with SerializedError instead of Error,
 /// allowing it to be used with errors parsed from JSON.
-pub fn remove_unused_ignores_from_serialized(unused_ignore_errors: Vec<SerializedError>) -> usize {
+pub fn remove_unused_ignores_from_serialized(
+    unused_ignore_errors: Vec<SerializedError>,
+    remove_unused_type_ignores: bool,
+) -> usize {
```
tobyh-canva, line 607: This function is getting a bit unwieldy and I think the way ignore removal works might deserve a refactor in a future PR.

The type-checker flags an unused ignore, then the suppressor has to work backwards from the human-readable error message and parse the line with regex to discover which parts of the line to remove. The type-checker could instead record the precise information once when flagging the error, serialise it in a structured way, then the suppressor just does the rudimentary cleanup.

Additionally, we could make this a code action / quick fix in the LSP.

pyrefly/lib/commands/check.rs
```diff
@@ -421,6 +421,9 @@ struct BehaviorArgs {
     /// Remove unused ignores from the input files.
     #[arg(long)]
     remove_unused_ignores: bool,
```
kinto0: nit: maybe we rename this to `remove_unused_pyrefly_ignores` or equivalent if that's all it does

pyrefly/lib/commands/check.rs
```diff
@@ -421,6 +421,9 @@ struct BehaviorArgs {
...
+    /// Remove unused `# type: ignore` comments in addition to unused Pyrefly ignores.
```
kinto0: should this flag only remove type: ignores? then you just pass both flags to remove both of them?

that would give the CLI a little more control

pyrefly/lib/error/suppress.rs
kinto0: refactor would be great

I like the code action / quick fix idea. I added it to #2207

kinto0 (approved):
would love @maggiemoss 's perspective on this

pyrefly/lib/error/suppress.rs
```diff
@@ -1450,13 +1513,71 @@ def f() -> int:
...
+                    message: "Unused `# pyrefly: ignore` comment".to_owned(),
+                },
+                SerializedError {
+                    path: PathBuf::from("test.py"),
+                    line: 0,
+                    name: "unused-type-ignore".to_owned(),
+                    message: "Unused `# type: ignore` comment".to_owned(),
+                },
+            ];
+            assert_remove_ignores_from_serialized_with_flag(input, errors, want, 1, true);
```
kinto0: should this be 2 removals? (2 on the same line)

pyrefly/lib/commands/check.rs
tobyh-canva: Hmm it would be a breaking change, if you think it's worth it we can?

pyrefly/lib/commands/check.rs
tobyh-canva: I thought about that, but I'm not sure if there's a real use case for it. I'm not sure what situation would make anyone want to use Pyrefly to remove unused `# type: ignore`s but leave unused `# pyrefly: ignore`s.

pyrefly/lib/error/suppress.rs
tobyh-canva: I have started work on it, though in the interest of making reviews easier I was planning on doing a follow-up. Would you prefer me to do it all in this PR? :) 

I've got a draft if you're interested (diff is against the head of this PR) https://github.com/tobyh-canva/pyrefly/pull/2

pyrefly/lib/error/suppress.rs
tobyh-canva: You're right, the function is returning number of modified lines rather than the number of unused suppressions (before this PR, they were the same thing :D ). I'll update this test then change the impl to return the number of removed suppressions.

pyrefly/lib/error/suppress.rs
kinto0: since github won't let us use github stacks yet, two PRs against main is the best strategy I've found. we merge the first PR quickly and then review the second PR by looking at commits in isolation

maggiemoss:
What if instead of new flags for each kind of ignore, we had the ability to pass an option to it and specify which kind?

pyrefly/lib/commands/check.rs
maggiemoss: I would rather have remove_unused_ignores be changed from a bool to a value someone can pass in to specify which ones. The default should either be all or be just pyrefly ignores.

pyrefly/lib/commands/check.rs
maggiemoss: If someone is using two type checkers - they might want finer control over which ignores are removed by our tool.

maggiemoss (changes requested):

tobyh-canva:
Here's what the flag's help looks like from `pyrefly suppress --help` after addressing feedback:
```
--remove-unused[=<KIND>]
    Remove unused ignores instead of adding suppressions, optionally selecting `pyrefly`,
    `type`, or `all`. Defaults to `pyrefly` when no kind is specified

    Possible values:
    - pyrefly: Remove unused Pyrefly and Pyre ignores (default)
    - type:    Remove unused `# type: ignore` comments
    - all:     Remove unused Pyrefly, Pyre, and `# type: ignore` comments
```

maggiemoss (approved):

pyrefly/lib/error/suppress.rs
tobyh-canva: Will continue work on this in #4335 :)

PR #4201, opened 2026-07-18
reject arbitrary attribute assignment on bound methods

### LINKED ISSUE #4163: Bound method objects incorrectly allow arbitrary attribute assignment
### Describe the Bug

The following does not produce an error:

```
class Foo:
    def real_method(self) -> None: ...

f: Foo = Foo()
f.real_method.test = None
```

This code will fail at runtime.

`mypy` for reference:
```
$ mypy --strict example.py 
example.py:5: error: "Callable[[], None]" has no attribute "test"  [attr-defined]
```

Note that unbound methods can be assigned arbitrary attributes. This (bad) code does not fail at runtime.
```
def func() -> None: pass
func.test = None
print(func.test)
```

Both `pyrefly` and `mypy` report errors for this case.
```
ERROR Object of class `FunctionType` has no attribute `test` [missing-attribute]
 --> example.py:2:1
  |
2 | func.test = None
  | ^^^^^^^^^
```

and `mypy`:
```
$ mypy --strict example.py 
example.py:2: error: "Callable[[], None]" has no attribute "test"  [attr-defined]
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuiAOuvV-ZjGPQCcYqKAH0AtjAAuAC1yYAFHBhQwASnoBaAHz0AcrnQxE9QmY4cwJ5rnoBeJiwVrLhISInS5mQlPhT7fUMYEAAaEDIhMChSX1xxKAoAYnoABVIomPo0LDx8ekpDSABzAFcBVCkIQ0IOFIBlGBh6GSkpYjhEAHouyL4YwlwBYq6YdC7MXEo4LsL0EvLK6vH6MCH6VAA3VGhUbFgCoogyiqrDelxiM-Q4WvQyWUNNTZgBOGXAthAAZkIARgATF8OMQhMoAg4viIoMD0CAAL7hVCUKovRjQEKIEA5HAEEjkeFAA

_No response_

knQzx:
a bound method is a `types.MethodType`, which doesn't override `__setattr__`/`__delattr__`, so assigning an arbitrary attribute to it fails at runtime. pyrefly allowed it: `lookup_magic_dunder_attr1` had no `BoundMethod` arm, so the base fell through to the general lookup and resolved the dunder via `MethodType.__getattr__` (returns `Any`), which made the assignment look valid

add a `BoundMethod` arm mirroring the `ClassInstance` handling: when the mutating dunder is inherited from `object`, report it as not-found so the set/delete is rejected. attribute reads are unchanged

closes #4163

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -2017,6 +2017,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             {
                 acc.not_found(NotFoundOn::ClassInstance(cls.class_object().clone(), base))
             }
+            // A bound method is a `types.MethodType`, which does not override
+            // `__setattr__`/`__delattr__`, so it does not accept arbitrary attribute
+            // assignment (this fails at runtime). Without this arm the base would fall
+            // through to the general lookup and resolve the dunder via
+            // `MethodType.__getattr__`, incorrectly permitting the assignment.
+            AttributeBase1::BoundMethod(_)
+                if (*dunder_name == dunder::SETATTR || *dunder_name == dunder::DELATTR)
+                    && self.field_is_inherited_from(
+                        self.stdlib.method_type().class_object(),
+                        dunder_name,
+                        (ModuleName::builtins().as_str(), "object"),
+                    ) =>
+            {
+                acc.not_found(NotFoundOn::ClassInstance(
+                    self.stdlib.method_type().class_object().clone(),
+                    base,
+                ))
+            }
             AttributeBase1::ShapedArrayInstance(tensor)
                 if (*dunder_name == dunder::SETATTR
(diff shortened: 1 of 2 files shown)


knQzx:
heads up on the mypy_primer diff: the 4 new errors on dragonchain are all the `mock.method.return_value = ...` pattern, where the object is typed as the real class instead of Mock. they're technically correct (bound methods really don't have those attrs, mypy flags them too), just noisy on mock-heavy tests. if you'd rather keep test code quiet i can look into special-casing Mock/MagicMock, lmk what you prefer)

yangdanny97:
I think this is probably fine. Thanks for the PR!

connernilsen (approved):

PR #4207, opened 2026-07-19
Fix False positive "Function declared to return Never but is missing an explicit return"

### LINKED ISSUE #4176: False positive "Function declared to return `Never` but is missing an explicit `return`"
### Describe the Bug

```python
from typing import Never

def foo() -> Never:
    print("hello")

foo()
```

Python: 3.12

We probably shouldn't give this error if the expected return is Never

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0A5GADcYlADrpJmGGDphcuABQBKOgFoAfIJFjEkuobrFKrBkvEgAFjChRcllZMkLlKkABoQZSrKilCBlooCgBiOgAFUl8wfzo0LDx8OgBjXHRINgBXSlQGCHTCSXCAZRgYOisGBmI4RAB6ep8-AN42eph0esxcFLh6tIyIbNz89Pr5XjpUYVRoVGxYVPTMnLyC9DpcYjH0OCL0MgYrdPVRSjgNugBeOksAZkIARgAmS0kAbTFqC4BdSSy6C4PD4MEw6kwEF8KXyohudAA5MN0LwYAjJCiGOpfABHLJQsHqADWMFI6lQKRS8Dg8IRAHdUJR0Oj0CAAL5eCmwmAAMWgMAoCRwBBI5DZQA

_No response_

mwestbroek:
# Summary

When a function annotation is `-> Never` and there is no explicit `return` in the body, don't warn about lacking an explicit `return`.

In practice, we don't proceed to `check_type()` in `check_implicit_return_against_annotation()` since `got` and `want` are indeed mismatched but we don't want to emit an error.

Fixes #4176. 

# Test Plan

Added `test_no_explicit_return_if_return_type_is_never` to `lib/test/returns.rs` which fails before the change in `lib/alt/solve.rs`, and passes afterwards.

I've run `test.py` and committed the resulting changes - I am not sure how this should work so any feedback would be appreciated!

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -10832,17 +10832,6 @@
   "specialtypes_never.py": [
     {
       "code": -2,
-      "column": 22,
-      "concise_description": "Function declared to return `NoReturn` but is missing an explicit `return`",
-      "description": "Function declared to return `NoReturn` but is missing an explicit `return`",
-      "line": 19,
-      "name": "bad-return",
-      "severity": "error",
-      "stop_column": 30,
-      "stop_line": 19
-    },
-    {
-      "code": -2,
       "column": 21,
       "concise_description": "`list[Never]` is not assignable to `list[int]`",
       "description": "`list[Never]` is not assignable to `list[int]`",
--- a/conformance/third_party/conformance.result
+++ b/conformance/third_party/conformance.result
@@ -137,7 +137,9 @@
   "qualifiers_final_annotation.py": [],
   "qualifiers_final_decorator.py": [],
   "specialtypes_any.py": [],
-  "specialtypes_never.py": [],
+  "specialtypes_never.py": [
+    "Line 19: Expected 1 errors"
+  ],
   "specialtypes_none.py": [],
   "specialtypes_promotions.py": [],
   "specialtypes_type.py": [],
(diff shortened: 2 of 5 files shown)


mwestbroek:
Hello, is there anything I can do to support this review?

markselby9:
I think the current behavior here might be intended, since the typing conformance suite expects this error. `conformance/third_party/specialtypes_never.py:19`:

```python
def func1(x: int) -> NoReturn:  # E: implicitly returns None
    if x != 0:
        sys.exit(1)
```

When `x == 0`, `func1` falls through and implicitly returns `None`, which contradicts `-> NoReturn`. The example in #4176 (`def foo() -> Never: print("hello")`) looks like the same shape.

One thing worth double-checking: with this change the regenerated `results.json` seems to drop from pass 134 → 133 (0.96 → 0.95), with `specialtypes_never.py` newly failing (`Line 19: Expected 1 errors`). It looks like the `annotation.is_never() && !has_explicit_returns` guard also suppresses the fall-through case the spec wants flagged. (The always-diverging case already stays quiet, e.g. `def stop() -> NoReturn: raise ...`.)

If that's right, #4176 might be more about the wording — "missing an explicit `return`" is confusing for a `Never` return — so rewording it (maybe "may implicitly return `None`") could clear up the confusion without moving conformance. Might be worth confirming the intended direction before putting in the rework.

PR #4212, opened 2026-07-20
fix Unpacked TypeVarTuple in callable doesn't consider optional arguments #1269

### LINKED ISSUE #1269: Unpacked TypeVarTuple in callable doesn't consider optional arguments
### Describe the Bug

```
from typing import *

def test[*Ts, T](f: Callable[[*Ts], T], *args: *Ts): pass
def callback(x: int, y: int = 1): pass

test(callback, 1) # error?
test(callback, 1, 1) # ok
```

It seems like `*Ts` gets eagerly resolved to `tuple[int, int]` as if every argument was required. But I think we should only do that if it's un-solved by the end of the call. In this case, the args provided in the first call is valid when considering optional params.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AqADroRmGGEbwGAbQEAVOABo68gLoAKMIjoBhVFCipssGXMVqV6lQNSU2cHQrgBKHcVRw4YiXQDGBlDYqH4A1hr4OqwMKqRR6PwAvHQAjG50Hl4iIgzSGgGGwWEqaXQAxHQwlNSUAPw5eQVBIaElJS7ldLihIEogZJQSUKSEDLRQFBUACqSDYMN0aFh4%2BP646JBsAK6UqAwQ64QiFQDKMDB0ABYMDMSOAPT3A0MjvGz3MOj3mLh%2BcPd%2BdabHZ7A5fOhgXh0VAAN1Q0GMsDWGwg212%2B3WXWIGPQcCO6DIDEu6wAtDCqnAwXRkkIQABmQgpABMtJEIAAvn0QvtyQAxaAwChLHAEEjkdlAA

_No response_

asukaminato0721:
# Summary

Fixes #1269

Callable inference now models every valid positional prefix introduced by optional parameters, allowing later *args evidence to select the correct Ts arity

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/solver/subset.rs
+++ b/pyrefly/lib/solver/subset.rs
@@ -349,36 +349,55 @@ impl<'solver, 'subset, Ans: LookupAnswer> Subset<'solver, 'subset, Ans> {
                     Some(Param::Varargs(_, Type::Unpack(u))),
                 ) => {
                     let mut l_types = Vec::new();
+                    let mut l_optional_prefixes = Vec::new();
                     loop {
-                        if let Some(Param::PosOnly(_, l, _) | Param::Pos(_, l, _)) = l_arg {
+                        if let Some(Param::PosOnly(_, l, required) | Param::Pos(_, l, required)) =
+                            l_arg
+                        {
+                            // Each trailing optional positional parameter adds a valid shorter
+                            // sequence; a later required parameter makes earlier prefixes invalid.
+                            match required {
+                                Required::Required => l_optional_prefixes.clear(),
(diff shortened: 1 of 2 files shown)


yangdanny97:
Thanks! I'll patch this to add a cap for how many things we expand, I think maybe past 8 we should fall back to the old behavior for performance reasons.

grievejia (approved):

PR #4224, opened 2026-07-21
Contextually type comprehension arguments during overload resolution

### LINKED ISSUE #4167: Overload resolution doesn't do "expected type" literal narrowing through a comprehension
### Describe the Bug

```py

from typing import Literal, overload

Order = Literal["ascending", "descending"]


@overload
def sort_indices(sort_keys: Sequence[tuple[str, Order]]) -> str: ...
@overload
def sort_indices(sort_keys: str) -> int: ...
def sort_indices(sort_keys: Sequence[tuple[str, Order]] | str) -> str | int:
    return 0


names = ["a", "b"]
result = sort_indices([(name, "ascending") for name in names])
```

I think `sort_indices` should be able to accept `[(name, "ascending") for name in names]`, as it's a `Sequence` of `tuple[str, Order]`?

```console
(scratch) mgorelli@marcoslaptop:~/scratch$ mypy t.py
All checks passed!
ERROR No matching overload found for function `sort_indices` called with arguments: (list[tuple[str, str]]) [no-matching-overload]
  --> t.py:18:22
   |
18 | result = sort_indices([(name, "ascending") for name in names])
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
  Possible overloads:
    (sort_keys: Sequence[tuple[str, Order]]) -> str [closest match]
    (sort_keys: str) -> int
  Argument `list[tuple[str, str]]` is not assignable to parameter `sort_keys` with type `Sequence[tuple[str, Order]]` in function `sort_indices`
 INFO 1 error
```

Spotted in [Narwhals](https://github.com/narwhals-dev/narwhals). Got a bit of help from Claude for narrowing down a dependency-free reproducer

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEA%2Bg2AK4AuLlMTdENxuSmzoZ0uNqjYRc6OAB10CsNXoBjXFFiqpMuIVTZVvfoOEBlGAEcWMdKphKVdNqWIR0Ac2MChdADIQbDCUqFAANHS4AG7BULiomAoKAPKUmMF0ALz%2BgcGhANpyIKhw9ljuHkURRemltpgVRQC6SYroAALRsfGJWDBgdHCmDO4N9nAAFENCDADWMKRwiHQW1rb2%2BRzEsPlwbJQRqemUTU0AlHQAtAB8g-vLhI8KnTGUcQkK6QPTbCPl41NhvNFss9pQLjdeOg2A8nn1vsNRhAAT85gslisrDY7DBNixtriwYc0sFTnQAD53cFXW5gilQmEKOjMuhcDiUdB0AAMrQU6FQNHgWTohWKVToRWwzQUXDgLCgwmyqKRAPyE35guqxTq5U8RQuYEEdA1MChxoF8HOChAYRAZC4YCgpEIbFoUAoAGI6AAFUgOp2DDA4Ah0dToSAeTiSaToQgKL0WU0ACzYbGISwA9Bn7f0nYRBB4M7YM5hcKo4BmwxGozp0Bm6IbKCIoqhoAZYKGZNWQrXIsRa3oFGQ2EmZJdXnAY8KigBmQgARgATEUFPlgtRKHAWugWOg%2BD4gphLg0uNoIDFhQByCAeMRcS988SXLjWCBcI-Ay6oVTjOBXgDuqAcg%2B6AgAAvra35SDEABi0AwBQaBYHgRBkOBQA

_No response_

markselby9:
# Summary

Fixes #4167.

When resolving an overloaded call, container **literals** (`list`/`set`/`dict`) are kept as expressions rather than eagerly inferred, so each candidate overload can contextually type them against its parameter (and narrow element literals). Comprehensions and generator expressions were not — they fell through to a context-free `expr_infer` in `CallWithTypes::type_or_expr`, so:

```py
from typing import Literal, overload

Order = Literal["ascending", "descending"]

@overload
def sort_indices(sort_keys: Sequence[tuple[str, Order]]) -> str: ...
@overload
def sort_indices(sort_keys: str) -> int: ...
def sort_indices(sort_keys: Sequence[tuple[str, Order]] | str) -> str | int:
    return 0

names = ["a", "b"]
result = sort_indices([(name, "ascending") for name in names])
```

was inferred as `list[tuple[str, str]]` and reported `no-matching-overload`, even though mypy, pyright, and ty all accept it (reported against [Narwhals](https://github.com/narwhals-dev/narwhals)).

The fix adds `Expr::ListComp | Expr::SetComp | Expr::DictComp | Expr::Generator` to the same un-flattened arm. These variants already support hint-directed inference via `infer_with_decomposed_hint` / `decompose_list`, so deferring their inference lets the existing per-overload contextual-typing path narrow their element/key/value types exactly as it does for container literals. When a hint doesn't decompose (e.g. a generator against a non-`Sequence` param), inference falls back to the previous context-free behaviour, so no path is worse than before.

# Test Plan

Added `test_overload_literal_narrowing_through_comprehension` (the reported repro) and `test_overload_literal_narrowing_through_set_dict_generator` (set/dict comprehensions and generators) to `pyrefly/lib/test/overload.rs`. Both fail before the change (`no-matching-overload`, `reveal_type` = `Unknown`) and pass after (`reveal_type` = `str`).

`cargo test -p pyrefly --lib` passes; `test.py` formatting and linting are clean.

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -876,7 +876,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     ) -> Option<Type> {
         let ty;
         let got = match got {
-            ExprOrBinding::Expr(value) => TypeOrExpr::Expr(value),
+            ExprOrBinding::Expr(value) => TypeOrExpr::Expr(value, false),
             ExprOrBinding::Binding(got) => {
                 ty = self.solve_binding(got, range, errors);
                 TypeOrExpr::Type(ty.ty(), range)
@@ -1134,7 +1134,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         narrowed_types: &mut Vec<Type>,
     ) {
         let ty = match &got {
-            TypeOrExpr::Expr(got) => self.expr_check(
+            TypeOrExpr::Expr(got, _) => self.expr_check(
                 got,
                 Some((&attr_ty, &|| {
                     TypeCheckContext::of_kind(TypeCheckKind::Attribute(attr_name.clone()))
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1011,8 +1011,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             self.get_dunder_new(cls, preserve_self).is_some()
                 && self.get_dunder_init(cls, false).is_some()
         };
-        let is_deeply_nested =
-            |x: &TypeOrExpr| matches!(x, TypeOrExpr::Expr(e) if nests_calls(e, FLATTEN_CALL_DEPTH));
+        let is_deeply_nested = |x: &TypeOrExpr| matches!(x, TypeOrExpr::Expr(e, _) if nests_calls(e, FLATTEN_CALL_DEPTH));
 
         // Infer deeply nested arguments once, up front. Shallower nesting costs only a constant
         // factor, and flattening it would discard the parameter hint for no benefit.
@@ -1345,7 +1344,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let infer_type_or_expr = |toe: TypeOrExpr, errors: &ErrorCollector| -> Type {
             let ty = match toe {
                 TypeOrExpr::Type(ty, _) => ty.clone(),
-                TypeOrExpr::Expr(e) => self.expr_infer(e, errors),
+                TypeOrExpr::Expr(e, _) => self.expr_infer(e, errors),
             };
             // NNModule fields carry captured constructor args (e.g., padding=Literal[1]) that DSL
             // forward functions need as literals to compute output shapes.
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -117,6 +117,7 @@ impl CallWithTypes {
                 | Expr::SetComp(_)
                 | Expr::DictComp(_)
                 | Expr::Generator(_)),
+                _,
             ) if !nests_calls_to_depth(e, MIN_FLATTEN_CALL_DEPTH) => {
                 // Hack: keep mutable builtin containers as expressions, since they often need to be
                 // contextually typed against the function's parameter types, unless nesting depth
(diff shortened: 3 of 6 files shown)


markselby9:
Rebased onto current `main` (`406082a6c`); this is mergeable again.

`callable.rs` auto-merged cleanly across the call-boundary refactor (#8888e70c5 / #6ba516f98 / #e67bc8e7a) — the only conflict was an append collision in `test/overload.rs`, resolved keep-both. The refactor restructured the layer *below* argument checking, but neither seam this change uses moved: `CallWithTypes::type_or_expr` is byte-identical, and the `post_check` guard already routes through `check_type_with_options(...).with_call_context(...)` like its TypeForm/ShapedArray siblings, so it participates in the new boundary's residual capture the same way. Nothing here creates or consumes a `CallBoundary`.

I checked that the fix is still load-bearing rather than assuming it, by ablating only the `type_or_expr` comprehension arm on the rebased tree:

- without it: `reveal_type(sort_indices([(name, "ascending") for name in names]))` → `Unknown`, plus `no-matching-overload`
- with it: `str`

So the refactor fixed no part of #4167 on its own.

**One thing worth reframing about the `bug`-marked test.** While probing the guard's scope I found the widening it suppresses is not comprehension-specific — on current `main`, with a generic hint, `list(filter(bool, {email.strip()}))` and the list-literal equivalent both already give `list[object]`, while the tuple form gives `list[str]`. So a provisional generic hint widening a deferred container's element type is pre-existing behaviour for list/set literals. Commit 1 makes comprehensions inherit it and commit 2 patches it back out for comprehensions, which means the guard restores comprehensions to their pre-PR result rather than introducing a new asymmetry. The `bug` marker is documenting a narrow symptom of that broader issue with the "don't flatten mutable builtin containers" approach.

I considered a tighter predicate — skip contextual typing only when the hint's element position is a bare unsolved quantified var, keeping narrowing for structural hints like `Callable[[int], T]`, which would let the marker be dropped. I left it out: it makes this diff more divergent from upstream rather than less, and I can't verify it without a primer run. Happy to do it as a follow-up if you'd prefer that shape.

Also strengthened the tests: the previous generator assertion wasn't actually discriminating (a bare generator's element isn't literal-promoted, so it passed with or without the fix). Added `test_overload_comprehension_hint_selects_narrower_overload`, where every overload accepts the argument so it asserts *which* one wins; both assertions verified to fail under ablation.

Suite is 7467/0 on the new base, lint clean.

PR #4226, opened 2026-07-21
Remove redundant pieces of the IGNORE_COMMENT_REGEX

oprypin:
# Summary

This is a no-op refactor.

Each branch of this regex contains a copy of the construct `\s*(?:;\s*)?[^#]*`, but the last part of it already covers everything that the other parts could possibly cover, so we remove them.

# Test Plan

No functionality is changed, so if all preexisting tests pass, that should be sufficient.
And in my brief manual usage of `pyrefly suppress` I didn't encounter any issues.

### DIFF
--- a/pyrefly/lib/error/suppress.rs
+++ b/pyrefly/lib/error/suppress.rs
@@ -43,13 +43,13 @@ use crate::state::errors::sorted_multi_line_string_ranges;
 /// Each consumes all non-`#` characters after its ignore pattern, so removing one
 /// suppression preserves any other pragma later on the same line.
 static PYREFLY_IGNORE_COMMENT_REGEX: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new(r"#\s*pyrefly:\s*ignore\s*(\[[^\]]*\])?\s*(?:;\s*)?[^#]*").unwrap()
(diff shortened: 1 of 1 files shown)


NathanTempest (approved):
Thanks @oprypin the analysis is correct: \s* and (?:;\s*)? only ever match non-# characters, and the trailing [^#]* already subsumes them greedily, so this is a genuine no-op (same matched extent; the removed group is non-capturing, and the callsite replaces the whole match, so nothing downstream shifts).

One heads-up: since you opened this, main refactored that single combined IGNORE_COMMENT_REGEX into three separate constants (PYREFLY_IGNORE_COMMENT_REGEX, TYPE_IGNORE_COMMENT_REGEX, PYRE_IGNORE_COMMENT_REGEX), all of which still carry the same redundancy. I'm importing this and applying the equivalent cleanup across the three so we can land it with your authorship, no action needed on your end and keep contributing. We really appreciate your time and effort!

NathanTempest:
Hello @oprypin I am not able to import this PR as it is stale, can you rebase this PR so I could import this cleanly ?

NathanTempest:
After you rebase, you could do the changes in the other places yourself or you could simply push it, so I can import it and do the changes myself. Either is fine, and you are credited as the author of this diff. Thank you for your contribution!

But until you push the rebased commit, I cannot import it due to CI failures.

oprypin:
Alright! I reapplied the changes according to the latest state of the code.

PR #4230, opened 2026-07-21
Fix object dunder attribute completions

### LINKED ISSUE #1211: auto complete `__doc__` missing
### Describe the Bug

Image

```py
from pathlib import Path
Path(".").read_text.__do
```

### Sandbox Link

_No response_

vscode-insider

0.35.0

fhshaik:
# Summary

- Include attributes inherited from `object` when completing a double-underscore prefix.
- Keep ordinary member completion lists unchanged to avoid adding dunder noise.
- Add an LSP regression test covering `__doc__` completion on a bound method.

Fixes #1211

# Test Plan

- `cargo test test::lsp::completion::` (104 passed)
- `cargo test` (6,925 passed; four sandbox-blocked IPC tests rerun below)
- `cargo test lsp::non_wasm::connection::tests::ipc_transport::` (5 passed outside the filesystem sandbox)

# AI disclosure

Implemented with assistance from OpenAI Codex under the `fhshaik` account.

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -2865,6 +2865,8 @@ pub struct AttrInfo {
     pub definition: AttrDefinition,
     /// is this defined in another module (true) or in this module (false)?
     pub is_reexport: bool,
+    /// Is this attribute inherited directly from `object`?
+    pub is_from_object: bool,
 }
 
 impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
@@ -2878,6 +2880,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     {
         let mut seen = SmallSet::new();
         for c in mro {
+            let is_from_object = c == self.stdlib.object().class_object();
             let Some(class_fields) = self.get_class_fields(c) else {
                 continue;
             };
@@ -2897,6 +2900,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                                     docstring_range: class_fields.field_docstring_range(fld),
                                 },
                                 is_reexport: false,
+                                is_from_object,
                             });
                         }
                     }
@@ -2914,6 +2918,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                                     .field_docstring_range(expected_attribute_name),
                             },
                             is_reexport: false,
+                            is_from_object,
                         });
                     }
                 }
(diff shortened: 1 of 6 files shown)


pyrefly/lib/alt/attr.rs
```diff
@@ -3148,7 +3181,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                expected_attribute_name,
+                include_types,
+                false,
+                &mut res,
+            );
+        }
+        res
+    }
+
+    pub fn completions_including_object(&self, base: Type, include_types: bool) -> Vec<AttrInfo> {
```
kinto0: do we need this fork? when do we want completions without objects?

pyrefly/lib/test/lsp/completion.rs
```diff
@@ -71,6 +71,23 @@ class Foo:
     }
 }
 
+#[test]
+fn completion_dunder_attribute_inherited_from_object() {
```
kinto0, line 76: let's also make sure these completions are ranked lower than non-object completions

kinto0 (changes requested):
thanks for picking this up!

fhshaik:
Good point. I removed the separate method and replaced it with a configurable API. I also added metadata to rank `object` attributes below class-defined attributes, along with regression coverage. Thanks for the review!

pyrefly/lib/lsp/wasm/completion.rs
```diff
@@ -112,6 +114,7 @@ fn assign_sort_text(ranked: &mut RankedCompletion, mru_rank: Option<Option<usize
             CompletionSource::AutoimportPrivate => "4b",
             CompletionSource::AutoimportPublic => "4a",
             CompletionSource::Reexport => "1",
+            CompletionSource::Object => "3z",
```
kinto0, line 117: nit: can this be put in sort order here?

pyrefly/lib/alt/attr.rs
```diff
@@ -3023,10 +3047,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         base: AttributeBase,
         expected_attribute_name: Option<&Name>,
         include_types: bool,
+        include_object: bool,
         res: &mut Vec<AttrInfo>,
     ) {
         for base1 in &base.0 {
-            self.completions_inner1(base1, expected_attribute_name, res);
+            self.completions_inner1(base1, expected_attribute_name, include_object, res);
```
kinto0, line 3054: instead of passing include_object through all functions, can we filter here?

kinto0 (changes requested):
thanks! just a few minor questions

PR #4233, opened 2026-07-21
Add bundled type stubs for httpx (#4233)

anushamukka-dev:
Summary:

Pyrefly bundles third-party type stubs for a curated set of popular libraries so that users get type checking for them out of the box, without needing to install a separate stub package. Today the bundled set includes Boto3, Botocore, Pandas, Scikit-learn, Scikit-image, Sympy, Vispy, and Conans, but `httpx` — one of the most widely used async-capable HTTP clients in the Python ecosystem — is not yet covered.

This diff adds an `httpx-stubs` package under `crates/pyrefly_bundled/third_party/stubs/`, following the same on-disk conventions as the existing bundled stubs (directory named `<import_name>-stubs`, `.pyi` interface files mirroring the upstream module layout, plus `py.typed` and upstream `LICENSE` for provenance). No registration or config edits are required — the `glob(["stubs/**/*"])` filegroup and the runtime module-name derivation pick the package up automatically.

The stubs cover the public API surface that users import in practice:
- Core client classes: `Client`, `AsyncClient` (and the shared `BaseClient` base), including `request`/`send`/`stream` and the per-verb helpers.
- `Request` and `Response`, plus `Headers`, `Cookies`, `QueryParams`, and `URL`.
- Top-level convenience functions: `request`, `stream`, `get`, `post`, `put`, `delete`, `patch`, `head`, `options`.
- Common config/types: `Timeout`, `Limits`, `Proxy`, `create_ssl_context`, the `codes` status-code enum, transports (`BaseTransport`, `AsyncBaseTransport`, `HTTPTransport`, `AsyncHTTPTransport`, `ASGITransport`, `WSGITransport`), auth classes (`Auth`, `BasicAuth`, `DigestAuth`, `NetRCAuth`), byte streams, and the full exception hierarchy.

The top-level `__init__.pyi` re-exports the public names with an `__all__` matching upstream `httpx`.

Reviewed By: kinto0

Differential Revision: D110685002


kinto0 (approved):

PR #4234, opened 2026-07-21
Fix config migration to check section presence not parse success (#4234)

anushamukka-dev:
Summary:

When migrating a `pyproject.toml` with `MigrationSource::Auto`, the code ran `try_mypy().or_else(|_| try_pyright())`. That falls back to pyright on *any* mypy parse failure, not just "section missing" — so a present-but-malformed `[tool.mypy]` (bad value types, etc.) silently became a pyright migration whenever both sections were present, hiding the mypy error the user actually needs to fix.

This decides which tool to migrate based on section *presence* rather than parse *success*. A new `has_toml_section` helper does lightweight TOML table-header detection (independent of section validity). Both `Args::load_from_pyproject` (used by `pyrefly init`) and the in-memory `find_and_migrate_in_memory` loader now use it: when `[tool.mypy]` is present we attempt mypy and surface its error; only when mypy is absent and `[tool.pyright]` is present do we migrate pyright; the both-absent case preserves prior behavior.

This also removes the now-resolved TODO and the ad-hoc post-hoc error-precedence logic in the in-memory loader, which is subsumed by presence-based selection.

Reviewed By: kinto0

Differential Revision: D111620497

### DIFF
--- a/crates/pyrefly_config/src/migration/run.rs
+++ b/crates/pyrefly_config/src/migration/run.rs
@@ -54,6 +54,35 @@ pub enum MigratedConfigSource {
     PyprojectToml,
 }
 
+/// Check whether a TOML section header is present in the raw file content.
+/// Uses basic TOML table header detection (not full parsing) so that
+/// section *presence* is independent of section *validity*. The section counts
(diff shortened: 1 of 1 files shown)


kinto0 (approved):

PR #4235, opened 2026-07-21
Resolve TSP type queries in files that aren't open (#4235)

### LINKED ISSUE #4228: TSP: getComputedType returns null for nodes in files that aren't open (open vs closed files resolve differently)
### Summary

When Pyrefly runs as a Type Server (`pyrefly tsp`), a `typeServer/getComputedType` request for a `Node` (a `{ uri, range }` pair) returns `null` whenever the target file is **not open** in the server — even though the file is on disk, reachable, and analyzed as part of checking an open file. Opening the same file (via `textDocument/didOpen`) makes the identical request return a valid `Type`. Open and closed files should resolve to the same type.

### Environment

- Reproduces on Pyrefly **1.0.0** and **1.2.0-dev.2** (latest dev), both started as `pyrefly tsp`.
- Type Server Protocol `typeServer/getComputedType`, arg = `Node { uri, range }` (zero-based, LSP-style character offsets).

### Expected behavior

`getComputedType` is a position/node-based query keyed by `(uri, range)`. It should return the computed type for a node in **any** file the server can analyze, regardless of whether that file is currently open in the editor.

### Actual behavior

- Node in an **open** file → valid `Type`.
- Same node in a file that is **not open** (but on disk and imported by an open file) → `null`.

### Minimal repro (protocol level)

1. `initialize` / `initialized`.
2. `textDocument/didOpen` for **`main.py`** only:

   ```python
   # main.py
   import lib
   x = lib.g()
   ```

   ```python
   # lib.py  (on disk, NOT opened)
   def g() -> None: ...
   ```

3. Send `typeServer/getComputedType` with `arg.uri` = `lib.py` and `arg.range` = the `None` return annotation of `g`. → **Response: `null`.**
4. Now `textDocument/didOpen` for `lib.py` and repeat step 3 → a valid `Type` is returned.

### Real-world impact

This surfaced in Pylance + Pyrefly-over-TSP as [microsoft/pylance-release#8122](https://github.com/microsoft/pylance-release/issues/8122). Pylance never sends `didOpen` for bundled stdlib stubs, so **every** node inside Pyrefly's bundled `builtins.pyi` (e.g. the `None` return annotation of `object.__init__`) returns `null` from `getComputedType`. It is not typeshed-specific — any unopened user file behaves the same way (repro above).

### Note / hypothesis

It looks like Pyrefly only retains position→node/type resolution for open documents, and returns `null` for `(uri, range)` lookups into background/closed files. Ideally the lookup should work for any analyzable file.

kinto0:
Summary:

TSP `getComputedType` (and `getDeclaredType`/`getExpectedType`) returned
`null` for any node in a file that wasn't open in the editor, even though
the file was on disk, reachable, and analyzed as a dependency of an open
file. Opening the same file made the identical `(uri, range)` query
succeed. This broke every node in Pyrefly's bundled stdlib stubs (e.g.
`builtins.pyi`), which clients like Pylance never `didOpen`.

Fixes #4228.

The root cause: `computed_type_at_range` and `open_at_position` built the
handle with `make_open_handle`, which always uses `ModulePath::memory`.
In-memory content only exists for opened files, so the memory-keyed handle
only matched open documents; closed files are keyed under `FileSystem` or
`BundledTypeshed` handles (whose `SysInfo` is inherited from the importer
and can't be reconstructed here), so the lookup missed and short-circuited
to `null`.

New helper `query_transaction_and_handle` resolves a path to the handle the
checker actually analyzes it under:
- open files -> in-memory overlay (unchanged fast path);
- otherwise reuse the already-loaded handle: the filesystem handle when it
  can be rebuilt directly, else a by-path scan that reuses the loaded
  `BundledTypeshed` handle so stdlib queries don't re-check typeshed;
- fall back to a fresh filesystem handle read from disk when the file
  hasn't been analyzed yet.

It then forces a full solve at `Require::Everything`, the only level that
retains bindings/answers — exactly the data the type lookup reads — so the
result is available regardless of what require level the file was loaded
at as a dependency. Open-file queries are unchanged; only the previously
`null` closed-file path does additional work, and bundled typeshed reuses
its already-solved handle.

Differential Revision: D113135763

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -6373,6 +6373,44 @@ impl Server {
         )
     }
 
+    /// Build a read transaction and the handle the type checker analyzes `path`
+    /// under, so `(uri, range)` queries resolve for any analyzable file rather
+    /// than only open documents.
+    ///
+    /// Open files are served from their in-memory overlay (already committed by
+    /// the recheck that ran on `didOpen`). For anything else we reuse the handle
+    /// the file was already analyzed under — an imported dependency's filesystem
+    /// handle, or a bundled stdlib stub's `BundledTypeshed` handle whose
+    /// `SysInfo` we can't reconstruct here, hence the by-path lookup — and force
+    /// a full solve (`Require::Everything` is the only level that retains
+    /// bindings/answers, which the type lookup reads) so narrowed/computed types
+    /// are available. A file that isn't analyzed yet falls back to a fresh
+    /// filesystem handle read from disk.
+    fn query_transaction_and_handle<'a>(&'a self, path: &Path) -> (Transaction<'a>, Handle) {
+        if self.open_files.read().contains_key(path) {
+            return (
+                self.state.transaction(),
+                make_open_handle(&self.state, path),
+            );
+        }
+        let mut transaction = self.state.transaction();
+        // Imported dependencies live under a filesystem handle we can rebuild
+        // directly; only scan when that misses (bundled stubs, unusual SysInfo).
+        let fs_handle =
+            handle_from_module_path(&self.state, ModulePath::filesystem(path.to_owned()));
+        let handle = if transaction.get_module_info(&fs_handle).is_some() {
+            fs_handle
+        } else {
+            transaction
+                .handles()
+                .into_iter()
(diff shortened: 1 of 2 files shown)


rchiodo (approved):

stroxler (approved):

PR #4239, opened 2026-07-22
fix divmod does not work right #3822

### LINKED ISSUE #3822: `divmod` does not work right
### Describe the Bug

The issue is found in pandas-dev/pandas-stubs#1772. The examples are inspired by microsoft/pyright#10849 and in particular microsoft/pyright#10899.

`sandbox.py`:
```py
from typing import overload

class LHS:
    @overload
    def __divmod__(self, other: RHS) -> tuple[RHS, RHS]: ...  # type: ignore[overload-overlap]
    @overload
    def __divmod__(self, other: object) -> tuple[LHS, LHS]: ...

class RHS:
    pass
```

`ttest.pyi`:
```py
from sandbox import LHS

lhs = LHS()

reveal_type(divmod(lhs, 1))  # ERROR with pyrefly, works for mypy 

```

Python: 3.12
mypy playground: https://mypy-play.net/?gist=38c79ab91908506693bb27a4135439b1

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN24AbjEpRcqTAB100gMZRUcOHQAyACQDKiaXV10AAkJFiJOvZhhg6AfWuYIgmrky2AFHBhQwAGgEMAFiKIdABKmgCUdAC0AHyMAK7EsADaYRq%2BaQC6wYS5ugDEjMwwwRBs6LwwyUai4phRNQrEmWa6hsK1puh6dBZWtvaOzm4eXr64AUEC2ABWMLIMkbEJSVXq6aqa2XS5hNJyCkqhmtrdesSKcNIg3iBklJZQpIQMtFAUhQAKpA9gT3RoLB4fB0WS4dCQNjxSioBgQcF7dCFDQwGB0fwMBjEOCIAD0uPuj2evDYuJg6FxmFwsjguLBELK0Nh8IpdDAvDoqEEqGgqGwsFB4MhTLh4IExFF6DgiLIAXBUQ6cBZdAAvHRJCAAMyEACMACYNddbpj4AwSKQIBQwNR6EwWOxONxePwHsJUFBrHaYJJKNJrbQARgcARHTw%2BJsND79r70FB-Mo1etXOEozIY66YO7PcVXIMnJhXHG4L4deEUzGQABfW6oBYOGAAMWgMAoJrgZrIlsrQA

_No response_

asukaminato0721:
# Summary

Fixes #3822

Deferred overloaded `__divmod__`/`__rdivmod__` selection until all call arguments are solved, preserved first-match overload ordering.

# Test Plan

add test


samwgoldman:
I spent a while looking at this PR. I am not ready to move forward with this solution yet, so I'm going to close it, but thank you (as always) for the hard work.

Before moving forward with a fix, I want to spend more time discussing the root cause. I am not sure if using overload residual inference is the right approach. I also noticed that this issue is more general than just divmod, even more general than binary operators, so this commit is too narrow in scope anyway.

I added a long comment to the original issue. If you have a moment, please give it a read and let me know what you think.

One more thing -- I think this PR would have benefitted from a more detailed description of the root cause. The summary in the description is very short and I was not really able to understand the high-level idea of this PR. I believe I understand correctly -- using overload residual checks to preserve both overloads through to the second argument -- but I could easily misunderstand.

PR #4241, opened 2026-07-22
Fix relative imports when renaming files

### LINKED ISSUE #4055: LSP `WillRenameFiles` implementation doesn't adjust relative imports
### Describe the Bug

Pyrefly's LSP server implementation only adjusts absolute imports when it gets a renamed-file notification (`WillRenameFiles` / `DidRenameFiles`<sup>1</sup>) from the client, but not relative ones.

To reproduce:

1. Create a Python package `pkg/__init__.py` with modules `pkg/a.py` and `pkg/b.py`.
2. Define a function in `a.py`:
    ```python
    def foo(): ...
    ```
3. Import this function from `b.py`:
    ```python
    from .a import foo
    foo()
    ```
4. In your editor with Pyrefly attached as an LSP server, rename `a.py` to `a2.py` and notice that the import in `b.py` remains unchanged.

Whereas if you instead have `from pkg.a import foo` in `b.py`, the import is updated to say `pkg.a2` as it should.

<sup>1 Although I think Pyrefly only implements the former</sup>

### Sandbox Link

_No response_

_No response_

Note: I tested this in Neovim with [neo-tree](https://github.com/nvim-neo-tree/neo-tree.nvim) and [nvim-lsp-file-operations](https://github.com/antosha417/nvim-lsp-file-operations), but as it worked fine for absolute imports, it wouldn't make sense for the editor setup to be at fault.

xaskii:
# Summary

Fixes https://github.com/facebook/pyrefly/issues/4055

- We resolve relative imports against the importing module when updating imports for `willRenameFiles`.
- This fixes renames like `pkg/a.py` to `pkg/a2.py`, where `from .a import foo` should become `from .a2 import foo`.
- If a move takes the module outside the package targeted by the leading dots, we update the import to an absolute path instead.

# Test Plan

- `uv run --with jsonschema --with toml test.py`
- Added regression tests for renaming within a package and moving outside it.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/will_rename_files.rs
+++ b/pyrefly/lib/lsp/non_wasm/will_rename_files.rs
@@ -26,6 +26,8 @@ use pyrefly_util::lock::RwLock;
 use rayon::prelude::*;
 use ruff_python_ast::Stmt;
 use ruff_text_size::Ranged;
+use ruff_text_size::TextRange;
+use ruff_text_size::TextSize;
 use tracing::info;
 
 use crate::lsp::non_wasm::module_helpers::PathRemapper;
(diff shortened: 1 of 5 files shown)


xaskii:
> This pull request has been automatically marked as stale because it has not had recent activity for more than 2 weeks.
> 
> If you are still working on this this pull request, please add a comment or push new commits to keep it active. Otherwise, please unassign yourself and allow someone else to take over.
> 
> Thank you for your contributions!

bump

NathanTempest:
Hi @xaskii sorry for the delay in reviewing, and thanks for the fix. Resolving relative imports through new_maybe_relative (matching go-to-def) and the stay-relative-vs-fall-back-to-absolute split are both right, and switching from str::replace to strip_prefix fixes a latent prefix bug in the old code.

Two nits: (1) from . import name is intentionally a no-op here could you add a tracking issue or a test pinning that behavior, since it's an incomplete-rename a user might hit? (2) a multi-dot (from ..x import y) test would round out the coverage.

After these two are addressed, I am ready to import this and run the full CI.

NathanTempest (approved):

NathanTempest:
Hey @xaskii imported the PR and I am addressing those two nits myself. I will let you know if there are any issues. Thanks again for your contributions to Pyrefly.

xaskii:
> Hey @xaskii imported the PR and I am addressing those two nits myself. I will let you know if there are any issues. Thanks again for your contributions to Pyrefly.

ok thank you!

yangdanny97 (approved):

xaskii:
thxthx

PR #4242, opened 2026-07-22
Optimise `venv` discovery for mid/large-sized directory trees

alexander-beedie:
# Summary

### Overview

* When `pyrefly check` runs over a package _without a configured interpreter_, it auto-discovers a project's `venv`.

* If "pyvenv.cfg" isn't immediately found in the project root or ".venv" / "venv", it falls back to an unfiltered `WalkDir` (to depth 3) over the entire project, testing every filename against the interpreter regex.

* This check is single-threaded, and runs _before_ everything else gets parallelised.

* For large projects, this can be a **significant** fixed cost. For example, local profiling showed it taking up ~20% of the total runtime for checking `homeassistant`, as that package has ~2,700 directories (!!).

### Solution

* `walk_interpreter` gains a simple "should_descend" predicate that prunes a directory's subtree when the predicate rejects it; a venv-aware pruning rule is supplied to it in `find_in_dir`.

* The prune is a strict superset of the reachable interpreters, so which interpreter the `find` resolves is unchanged, **but** the _non-venv_ source fan-out is completely skipped.

* In the extreme `homeassistant` case, this reduces **~2,700** `opendir` calls down to only **~15**, making the check essentially free (instead of ~20% of the runtime) 🚀 

# Test Plan

All existing tests pass, and two new `venv` discovery tests were added (which actually represent the majority of this PR as the code change itself is small).

# Benchmarks[^1]

Timed across a sample of  the `mypy_primer` projects. Speedup is (broadly) proportional to directory-tree size; significant on large nested trees, negligible on small flat projects.

`dirs ≤ 3` = how many directories (with depth ≤ 3) in the project.
`modules` = how many modules in the project.
`before` = pyrefly on main branch.
`after` = main + this PR. 

| project | dirs ≤ 3 | modules | before | after | Δ% |
|:--|--:|--:|--:|--:|--:|
| homeassistant | 2,736 | 9,811 | 2.010 | 1.611 | -19.8% |
| poetry | 124 | 191 | 0.172 | 0.151 | -12.6% |
| sphinx | 360 | 243 | 0.253 | 0.229 | -9.4% |
| pylint | 129 | 51 | 0.178 | 0.162 | -9.0% |
| hypothesis | 66 | 103 | 0.172 | 0.157 | -8.7% |
| sympy | 153 | 1,560 | 0.358 | 0.334 | -6.6% |
| pydantic | 60 | 104 | 0.170 | 0.164 | -3.4% |
| pandas | 129 | 1,459 | 0.407 | 0.396 | -2.6% |
| mypy | 61 | 1,098 | 0.394 | 0.384 | -2.5% |
| scipy | 137 | 925 | 0.429 | 0.425 | -1.0% |

[^1]: Test machine: Apple Silicon M3 Max (16 cores).

### DIFF
--- a/crates/pyrefly_config/src/environment/finder.rs
+++ b/crates/pyrefly_config/src/environment/finder.rs
@@ -40,8 +40,19 @@ static PYTHON_INTERPRETER_REGEX: LazyLock<Regex> = LazyLock::new(|| {
 /// to be found at either `.venv/bin/python(\d(\.\d)?)?` or `.venv/python(\d(\.\d)?)?`.
 ///
 /// Note: we do not follow links. For `venv`s, the symlink path contains important
-/// information the Python interpreter needs to proplrly execute.
+/// information the Python interpreter needs to properly execute.
 pub fn walk_interpreter(start: &Path, depth: usize) -> impl Iterator<Item = PathBuf> {
+    walk_interpreter_pruned(start, depth, |_| true)
+}
+
+/// Like [`walk_interpreter`], but skips descending into any directory for which
+/// `should_descend` returns `false`, pruning that subtree; use this to avoid
+/// walking source trees that *cannot* hold an interpreter.
+pub fn walk_interpreter_pruned(
+    start: &Path,
+    depth: usize,
+    should_descend: impl Fn(&walkdir::DirEntry) -> bool,
+) -> impl Iterator<Item = PathBuf> {
     let walker = WalkDir::new(start)
         .min_depth(1)
         .max_depth(depth)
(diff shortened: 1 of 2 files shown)


alexander-beedie:
FYI: @samwgoldman, some more performance for you ;)

samwgoldman:
Great find!

samwgoldman (approved):

samwgoldman:
This should be merging now. Thanks again!

While reviewing this, I did a bit of digging into the history of this change. We added this logic in 7d33f0a74b8bd1699a716d91e6e0c2efb8490137, and the idea was to mirror how the VS Code "python" extension finds venvs.

But I don't think it really makes sense for Pyrefly to crawl looking for venvs in subdirectories. I could understand looking for venvs in parent directories, but not subdirs.

If you want, I think it would be a great simplification if we could eliminate this subdir walk altogether. It might require untangling the history in case there is some case where we need the venvs discovered here. For example, if we search for them and cache them on startup, then rely on the cache later, we might need to add more parent-dir searches?

PR #4243, opened 2026-07-22
[pyrefly] Expose inferred type hints as quick fixes

### LINKED ISSUE #4231: Quickfix insert parameter types
In the IDE, we're often able to display the types of the parameters to a function or the type of a variable with Pyrefly's inference on the inlay hint request. Users can choose to insert these type suggestions by double clicking on the displayed types, but it would likely also be a good user experience to surface these suggestions through code actions, probably the quickfix handler.

You'll likely want to look at `pyrefly/lib/lsp/non_wasm/server.rs` to see how we get the hints for a file (`fn inlay_hints`), then pull out the hint for the current position.

shubhamshettyy:
Reuse insertable inlay hint edits in the quick-fix menu so users can materialize inferred annotations through code actions instead of double-clicking the rendered hint. Keep the initial behavior scoped to existing type hints while preserving their configuration and safety filters.

# Summary

- Reuse `Transaction::inlay_hints` from the existing quick-fix handler.
- Offer “Insert inferred type annotation” for existing variable, class-attribute, and function-return type hints.
- Reuse the exact label text and insertion position already used by inlay-hint edits.
- Exclude non-insertable hints, such as annotations on unpacked variables.
- Exclude call-argument name hints such as `value=`, since they are not type annotations.
- Respect the existing inlay-hint configuration for variable and function-return types.
- Support both regular Python files and notebook cells.

The implementation obtains the configured inlay hints during `textDocument/codeAction`, filters for safely insertable type hints whose insertion position is covered by the requested range, and converts the matching hint into a `CodeActionKind::QUICKFIX` containing a zero-width `TextEdit`.

Fixes #4231 

# Test Plan

- Added LSP integration coverage for:
  - Function-return annotations.
  - Variable annotations.
  - Class-attribute annotations.
  - Notebook-cell edits.
  - Non-insertable unpacked variables.
  - Call-argument name hints.
  - Already annotated variables.
  - Code-action kind filtering.
  - Disabled variable and return inlay-hint configuration.
- Ran:

  `cargo test insert_inferred_type_annotation`

- Ran the required formatting and lint checks:

  `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema`

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -4474,6 +4474,7 @@ impl Server {
             Some(CodeActionRequest::METHOD),
         )?;
         let import_format = lsp_config.and_then(|c| c.import_format).unwrap_or_default();
+        let inlay_hint_config = lsp_config.and_then(|c| c.inlay_hints).unwrap_or_default();
         let module_info = transaction
             .get_module_info(&handle)
             .ok_or(EmptyResponseReason::ModuleInfoNotFound)?;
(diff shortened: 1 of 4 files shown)


shubhamshettyy:
# Potential Follow-up

To keep this initial change small and appropriate for a good-first issue, it identifies type hints from their existing `: ` and ` -> ` label prefixes and matches code-action requests at the hint’s insertion position. A more robust implementation could extend `InlayHintData` with a semantic kind, such as `Type` or `ParameterName`, and a source range describing the declaration associated with the hint. This would avoid depending on display strings and would allow the quick fix to appear when the cursor is anywhere on the variable, attribute, or function declaration rather than only at the insertion point. That shared metadata would live in `lsp/wasm/inlay_hints.rs`, which contains platform-independent LSP logic used by the native server as well as WebAssembly consumers. I can include this extension in the current PR if maintainers prefer the stronger behavior.

shubhamshettyy:
@grievejia can you review this if available?

yangdanny97:
I have a drive-by question here, what's the benefit of having this as a quick-fix over the current method of inserting them?

Are there performance impacts to providing hundreds or thousands of these quick-fixes in a large file?

PR #4244, opened 2026-07-22
impl code action for narrowing the value with an is not None #4169

### LINKED ISSUE #4169: code action for narrowing the value with an `is not None`
```
def g(y: int):...
def f(x: int | None):
    g(x)
```

asukaminato0721:
# Summary

Fixes #4169

Adds an Add assert x is not None quick fix for optional simple-name values, inserts the assertion with correct indentation before the containing statement.

Avoids unsafe edits for inline statements or complex expressions.

# Test Plan

add test


stroxler (approved):

PR #4249, opened 2026-07-22
Exclude privately-named classes from `pyrefly coverage`

jorenham:
# Summary

Privately-named functions, variables, methods, and attrs were already excluded, but privately-named classes weren't, so e.g. `_Foo.bar` was still reported.
A private name now hides its whole subtree unless exported via `__all__`, consistent with the other symbol kinds and with the documented behavior.

# Test Plan

Regression tests added.


stroxler (approved):

PR #4250, opened 2026-07-22
Remove unused symbol fields from `pyrefly coverage`

jorenham:
# Summary

Some fields of `Function`, `Variable`, and `Parameter` were set but never read, and since only `SymbolReport` ends up in the JSON output, their `Serialize` derives were also unused.

This removes those fields and corresponding code, like the `is_fully_annotated` logic and the annotation source-code lookups. `Parameter` now stores its `SlotRank` directly, instead of two fields that `merge_overloads` immediately turned back into a `SlotRank`.

Oh and don't forget to check the diff counts ;)

# Test Plan

Tests still pass; no behavior changes


jorenham:
rebased to resolve conflicts

yangdanny97 (approved):

PR #4257, opened 2026-07-23
Fix semantic token highlighting for match captures

magic-akari:
# Summary

Match capture patterns bind regular variables, but Pyrefly’s semantic-token walker only handled expression names. This caused capture definitions and some references to be missing or inconsistently highlighted.

This change:

- Emits plain `variable` tokens for all capture patterns.
- Introduces a type-transparent `PatternCapture` binding that preserves the capture’s declaration identity.
- Keeps go-to-definition anchored at the capture site.
- Excludes wildcards and class-pattern keyword labels.
- Adds coverage for nested, starred, mapping, `as`, and OR patterns.

# Test Plan

- Semantic-token tests
- Pattern-matching tests
- Formatting and linting checks

# Screenshots

### Before this change
image

### After this change
image

### DIFF
--- a/pyrefly/lib/binding/binding.rs
+++ b/pyrefly/lib/binding/binding.rs
@@ -2346,6 +2346,9 @@ pub enum Binding {
     ClassDef(Idx<KeyClass>, Box<[Idx<KeyDecorator>]>),
     /// A forward reference to another binding.
     Forward(Idx<Key>),
+    /// A definition boundary for a capture pattern. It forwards type information during solving,
+    /// but origin and definition lookups must not follow it through to the matched subject.
+    PatternCapture(Idx<Key>),
     /// Like Forward, but widens implicit literals.
     PromoteForward(Idx<Key>),
     /// A forward reference produced during first-use resolution of a partial type.
@@ -2555,6 +2558,7 @@ impl DisplayWith<Bindings> for Binding {
             }
             Self::ClassDef(x, _) => write!(f, "ClassDef({})", ctx.display(*x)),
             Self::Forward(k) => write!(f, "Forward({})", ctx.display(*k)),
+            Self::PatternCapture(k) => write!(f, "PatternCapture({})", ctx.display(*k)),
             Self::PromoteForward(k) => write!(f, "PromoteForward({})", ctx.display(*k)),
             Self::ForwardToFirstUse(k) => {
                 write!(f, "ForwardToFirstUse({})", ctx.display(*k))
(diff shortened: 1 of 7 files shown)


kinto0 (changes requested):
thanks for the PR! would you mind updating the summary with more details about the bug it fixed and maybe a screenshot of before / after? thanks!

pyrefly/lib/test/lsp/semantic_tokens.rs
```diff
@@ -264,6 +264,169 @@ token-type: keyword
     );
 }
 
+#[test]
+fn pattern_capture_test() {
```
kinto0, line 268: do we need the test to be this big?

pyrefly/lib/binding/binding.rs
```diff
@@ -2324,6 +2324,9 @@ pub enum Binding {
     ClassDef(Idx<KeyClass>, Box<[Idx<KeyDecorator>]>),
     /// A forward reference to another binding.
     Forward(Idx<Key>),
+    /// A definition boundary for a capture pattern. It forwards type information during solving,
+    /// but origin and definition lookups must not follow it through to the matched subject.
+    PatternCapture(Idx<Key>),
```
kinto0, line 2351: can you help me understand why we need a new binding for this just for semantic tokens? why can't we handle this with the AST?

kinto0 (changes requested):

pyrefly/lib/binding/binding.rs
magic-akari: Good question. The AST is enough to emit tokens at capture sites, but not to classify later uses, which are just `Expr::Name` nodes.

Previously, captures used `Forward(subject_idx)`, preserving their type but losing their identity as new variable definitions. `PatternCapture` models the missing distinction: it forwards type information while remaining a definition boundary with `Variable` symbol kind. An AST-only approach would need to duplicate scope and control-flow name resolution in the semantic-token layer.

pyrefly/lib/test/lsp/semantic_tokens.rs
magic-akari: Good point. This can be much smaller. I’ll reduce it to a single `[head, *tail]` case, which covers both regular and starred capture definitions and their references.

magic-akari (commented):
I've updated the description with screenshots.

grievejia (approved):

PR #4268, opened 2026-07-23
Preserve `Self` when narrowing a self-type to a subclass

anushamukka-dev:
Summary:
Pyrefly narrows `self`/`cls` by intersecting a `SelfType` with a `ClassType`. When the class is a subclass of the self-type's class (e.g. `isinstance(self, Child)` or `issubclass(cls, Child)`), the intersection was collapsing to a plain `ClassType(Child)`, discarding the fact that the value is still the receiver's `Self`. Code as ordinary as

    class Parent:
        def widen(self) -> Self:
            if isinstance(self, Child):
                return self  # spurious: `Child` is not assignable to `Self@Parent`
            return self

was rejected with a false positive, because `Child` is genuinely not assignable to `Self@Parent`.

This intersection only fires when one operand is already a `SelfType`, so the value provably *is* a self. Narrowing it to the subclass therefore yields a self-type anchored at the subclass: `SelfType(Child)`. Attribute and constructor lookups still resolve through `Child`, but the value stays assignable back to `Self@Parent` since all self-types are mutually assignable. This is both more precise and correct, matching mypy and pyright.

Updates the previously `bug`-marked `issubclass` regression test (the error is gone) and adds an `isinstance` instance-method regression test.

Differential Revision: D110890742

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -223,13 +223,19 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         | (Type::SelfType(self_cls), Type::ClassType(cls)) = (left, right)
             && self.as_superclass(cls, self_cls.class_object()).as_ref() == Some(self_cls)
         {
-            // ClassType(C) & SelfType(Parent) simplifies to ClassType(C) when C
-            // is a subclass of Parent with a matching inherited instantiation,
-            // because Self[Parent] represents "Parent or any subclass" and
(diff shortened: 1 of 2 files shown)


fangyi-zhou (approved):

samwgoldman:
This is clever, and makes sense to me. I think `Self@Parent & Child` should probably have behaved the way you expected here -- maybe an issue resulting from incompleteness of our intersection type (cc @grievejia). But I wasn't able to easily find any counter-examples for narrowing to`Self@Child`. Thanks!

PR #4269, opened 2026-07-23
Remove "experimental" warning on the `pyrefly coverage` docs

jorenham:
`pyrefly coverage` is pretty stable now, and we can always introduce backwards-incompatible JSON schema changes safely using the schema versioning (i.e. by bumping the major schema version). But even so, at the moment I don't see any reason for wanting/needing to introduce a breaking change.


rchen152 (approved):

PR #4271, opened 2026-07-24
fix Preserve class-scoped type parameters when converting __new__ to a callable #4267

### LINKED ISSUE #4267: Preserve class-scoped type parameters when converting `__new__` to a callable
## Describe the Bug

Converting a generic class constructor backed by `__new__` into a callable
loses the class-scoped type parameter. The resulting callable returns
`Class8[Unknown]` instead of inferring and preserving `Class8[T]`.

Minimal reproduction:

```python
from typing import Callable, Generic, ParamSpec, Self, TypeVar, assert_type

P = ParamSpec("P")
R = TypeVar("R")
T = TypeVar("T")


def accepts_callable(cb: Callable[P, R]) -> Callable[P, R]:
    return cb


class Class8(Generic[T]):
    def __new__(cls, x: list[T], y: list[T]) -> Self:
        return super().__new__(cls)


r8 = accepts_callable(Class8)
assert_type(r8([""], [""]), Class8[str])
r8([1], [""])  # E: inconsistent arguments for T
```

The synthesized callable should be equivalent to:

```python
def [T](x: list[T], y: list[T]) -> Class8[T]: ...
```

Instead:

- Infers `Class8[Unknown]` for the valid call.
- Doesn't report the incompatible `list[int]` and `list[str]` arguments.

Expected behavior:

- Convert class-scoped type parameters used by `__new__` into function-scoped
  parameters on the synthesized callable.
- Infer one consistent `T` from all constructor arguments.
- Specialize the return type as `Class8[T]`.
- Reject calls whose arguments require incompatible values for `T`.

The mismatch appears in
[`conformance.result`](https://github.com/facebook/pyrefly/blob/main/conformance/third_party/conformance.result#L30-L33).
The corresponding test is
[`constructors_callable.py`](https://github.com/facebook/pyrefly/blob/main/conformance/third_party/constructors_callable.py#L170-L186).

## Related Work

[PR #2403](https://github.com/facebook/pyrefly/pull/2403) fixed preservation of
generic parameters when converting `__init__` to a callable. Its test plan
explicitly narrowed the remaining constructor-to-callable problems to
overloads and `__new__`.

Tests retain an explicit bug case for this generic `__new__` behavior in
[`pyrefly/lib/test/callable.rs`](https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/callable.rs#L1403-L1440).

This issue only covers preserving class-scoped generic parameters during
`__new__`-based constructor-to-callable synthesis.

## Sandbox Link

Not provided; the reproduction above is self-contained and comes directly from
the typing conformance suite.

asukaminato0721:
# Summary

Fixes #4267

Bare generic classes now preserve class-scoped type parameters when converting `__new__` to a callable.
 
Class8 correctly infers `Class8[str]` and rejects inconsistent constructor arguments.

# Test Plan

Added regression coverage and updated generated conformance results.

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -2085,6 +2085,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     }
 
     pub fn constructor_to_callable(&self, cls: &ClassType) -> Type {
+        let new_attr_ty = self.get_dunder_new(cls, false);
+        self.constructor_to_callable_impl(cls, new_attr_ty, false)
+    }
+
+    /// Convert a bare class definition while keeping its type parameters generic.
+    pub fn constructor_to_callable_for_class_def(&self, cls: &ClassType) -> Option<Type> {
+        let new_attr_ty = self.get_dunder_new_for_class_def(cls)?;
+        Some(self.constructor_to_callable_impl(cls, Some(new_attr_ty), true))
+    }
+
+    fn constructor_to_callable_impl(
+        &self,
+        cls: &ClassType,
+        new_attr_ty: Option<Type>,
+        preserve_class_tparams: bool,
+    ) -> Type {
         let class_type = self.heap.mk_class_type(cls.clone());
         if let Some(metaclass_call_attr_ty) = self.get_metaclass_dunder_call(cls) {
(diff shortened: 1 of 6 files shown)


yangdanny97:
oh nice, i just started looking at this, your impl looks a bit nicer than mine just based on complexity alone

will review the logic

fangyi-zhou (approved):

PR #4272, opened 2026-07-24
fix Quoted forward references incorrectly resolve to shadowing class members #4266

### LINKED ISSUE #4266: Quoted forward references incorrectly resolve to shadowing class members
## Describe the Bug

A quoted forward reference in a class body resolves against a same-named class
member. The Python typing conformance test expects the quoted reference to
resolve outside the class scope, while the equivalent unquoted annotation
resolves to the class member and is rejected as an invalid type.

Minimal reproduction:

```python


class ClassD:
    def int(self) -> None:
        ...

    x: "int" = 0  # OK: quoted reference resolves to builtins.int
    y: int = 0  # E: local method `int` isn't a legal type expression


assert_type(ClassD.x, int)
```

The quoted annotation produces an error for `x`:

```text
Expected a type form, got instance of `(self: Self@ClassD) -> None`
```

It then treats `ClassD.x` as `Unknown`, producing a second failure:

```text
assert_type(Unknown, int) failed
```

Expected behavior:

- Accept the quoted annotation `"int"` and resolve it to `builtins.int`.
- Continue rejecting the unquoted annotation `int`, which refers to the local
  method and isn't a valid type expression.
- Preserve `ClassD.x` as `int` without the cascading `Unknown` error.

The mismatch appears in
[`conformance.result`](https://github.com/facebook/pyrefly/blob/main/conformance/third_party/conformance.result#L11-L15).
The corresponding test is
[`annotations_forward_refs.py`](https://github.com/facebook/pyrefly/blob/main/conformance/third_party/annotations_forward_refs.py#L77-L96).

## Related Work

- [Issue #1900](https://github.com/facebook/pyrefly/issues/1900) discusses
  same-named methods shadowing types, but its reproduction uses unquoted
  annotations. Its discussion points to the expected error for the unquoted
  case in this conformance test.
- [PR #1983](https://github.com/facebook/pyrefly/pull/1983) added an
  overload-specific lookup exception.
- [Issue #357](https://github.com/facebook/pyrefly/issues/357) and
  [PR #3617](https://github.com/facebook/pyrefly/pull/3617) address unquoted
  forward references evaluated before Python 3.14.
- [Issue #3387](https://github.com/facebook/pyrefly/issues/3387) and
  [PR #3391](https://github.com/facebook/pyrefly/pull/3391) address a different
  `from __future__ import annotations` case.

This issue specifically covers lookup semantics for a quoted annotation when a
class member shadows the referenced name.

## Sandbox Link

Not provided; the reproduction above is self-contained and comes directly from
the typing conformance suite.

asukaminato0721:
# Summary

Fixes #4266

Quoted forward references now skip class-body namespaces during lookup.

Unquoted annotations retain existing class-member lookup behavior.

# Test Plan

add test, update conformance results.


yangdanny97:
closing for now because the original issue needs discussion

PR #4273, opened 2026-07-24
fix support "top" types #3736

### LINKED ISSUE #3736: support "top" types
ty has a `Top` type which is used to "replace" the `Any`s from any given type with their corresponding "top" type depending on the contex (ie. the variance). see discussion in https://github.com/astral-sh/ty/issues/2943#issuecomment-3988397317:

> ah so a gradual type just means `Any`/`Unknown`?
> 
> so does `Top` basically just mean "disable the unsafe behavior of `Any`/`Unknown` of the enclosed type", and is used here to work around the fact that there's no safe way to denote the upper bound of an invariant generic? so like:
> 
> * `Top[Any]` is equivalent to `object`
> * `Top[Iterable[Any]]` is equivalent to `Iterable[object]`
> * `Top[Callable[[Any], Any]]` is equivalent to `Callable[[Never], object]`
> * and `Top[list[Any]]` can't be represented any other way

it's especially useful in that last example -- representing types that otherwise cannot be represented in a typesafe way

another example where it comes up is narrowing callable types. in ty, this is an error:

```py
def f(x: object):
    if callable(x):
        reveal_type(x) # Top[(...) -> object]
        x()  # error: Object of type `Top[(...) -> object]` is not safe to call; its signature is not known (call-top-callable)
```
[playground](https://play.ty.dev/ff6aa968-9c68-4e2b-adc4-773993bd608c)

without the `Top`, the parameters `...` are treated as `Any`, which is unsafe: [pyrefly sandbox](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHSwwwdMAAp8iOrmwArGAGMGASkRy6pzot1ioqbLFVGTZp4JFjJzGPboBiOssIBBnQAtAB8Gtp6DI5OpvjKQT506Lh0MJTUlCAANCBkgmBQpIQMtFAUZAwAFrjowSKUcBC1dAC8dDIgAMyEAIwALJ1yxIJwMPztnXAMlBD6Q%2BgA2umZcAC6cpgwI3qoDDCYbR0gK7ydHehcxFBzEAzBNtOUqPrButZwcEedp5QLVzddHdgldeBNjr9znIAbcGEVgltIOgDg8GDMINgAK77b4nDJnEByGgQT6sNjBOC4TGUXQwXGQwnoFJ1GD4KqoTHTCAiYI0Pa6Kr0-F-Rkpe6CACOmIggkwwQA1jBSA9dLTPkLMgtcNI6pj0MQXvKyRqCRdZXrMBh7pZpiaRWaDharW9apg7s10Haoeg9ci1WhKMrdLQDbNKZ7JnjNSALr69PAA8qpDAXegGk0WpGGbH0IIXuzbHSs8LvXq8wKbLBefyqm9UGMvTG5GX4LgoEJKynMBb0g8oFBGznk3LQXxB82fWMR2wUoJx2nULMMLTeSS%2BQwBfPWB2bnKtsHnqVKI2QABfXIvBjcmAAMWgMAoaCweCIZDPQA)

asukaminato0721:
# Summary

Fixes #3736

# Test Plan

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -1637,11 +1637,24 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 &call_context,
                 &mut bound_args,
             ),
-            Params::Ellipsis | Params::Materialization => {
+            Params::Ellipsis => {
                 // Deal with Callable[..., R]
                 for arg in self_arg.iter().chain(args.iter()) {
                     arg.pre_eval(self, arg_errors).post_infer(self, arg_errors)
                 }
+                ArgMap::new()
+            }
+            Params::Materialization => {
+                for arg in self_arg.iter().chain(args.iter()) {
+                    arg.pre_eval(self, arg_errors).post_infer(self, arg_errors)
+                }
+                self.error_with_context(
+                    call_errors,
+                    arguments_range,
+                    ErrorKind::InvalidArgument,
+                    "Cannot call callable with unknown parameter types".to_owned(),
+                    context,
+                );
                 ArgMap::new()
             }
             Params::ParamSpec(concatenate, p) => {
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -58,7 +58,9 @@ use crate::binding::narrow::NarrowSource;
 use crate::binding::narrow::NarrowingSubject;
 use crate::error::collector::ErrorCollector;
 use crate::error::style::ErrorStyle;
+use crate::types::callable::Callable;
 use crate::types::callable::FunctionKind;
+use crate::types::callable::Params;
 use crate::types::class::ClassType;
 use crate::types::lit_int::LitInt;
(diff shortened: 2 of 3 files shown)


rchen152:
Thanks for taking a stab at this! I don't think this is the right direction - `Type::Materialization` is not the same as `Top`, and either way, this looks incomplete. At a minimum, we'd need to handle and test all of the cases quoted [here](https://github.com/facebook/pyrefly/issues/3736#issue-4626886327):

> * Top[Any] is equivalent to object
> * Top[Iterable[Any]] is equivalent to Iterable[object]
> * Top[Callable[[Any], Any]] is equivalent to Callable[[Never], object]
> * and Top[list[Any]] can't be represented any other way

I marked the issue as "needs discussion" because I think we need to flesh out the design space a lot more before moving on to an implementation. Off the top of my head: When would we infer `Top` types? What are the assignability rules? Do we allow users to construct them?

PR #4274, opened 2026-07-24
fix Feature suggestion: ban implicit bool conversion in conditional expressions #3959

### LINKED ISSUE #3959: Feature suggestion: ban implicit bool conversion in conditional expressions
What does the Pyrefly team think about introducing an optional diagnostic rule that would ban implicit bool conversion in conditional expressions? (or just in general)

```python
def get_temperature() -> int | None:
    """Returns None if data is missing, otherwise temperature in degrees Celsius rounded to the nearest integer."""
    return 0

temperature = get_temperature()

if not temperature:
    raise Exception("Temperature sensor broken, aborting program")

# we now have a bug, because `0` is a perfectly fine value, but it was discarded in the early return above

process_temperature(temperature)
```

It's a very common source of bugs in production code (other scenarios are `None` vs legitimately empty lists). As far as I remember from my time at Facebook – a related "sketchy null check" rule in Hack led to fixing _a lot_ of bugs like that.

It's definitely opinionated and arguably might better belong to a linter, but hinges on reliable type inference that's unavailable in Python linters today. That's why I'm creating this issue in a type checker repo.

PS: See https://github.com/microsoft/pyright/discussions/6488 for a related discussion on Pyright's repository with some community support & discussions on possible semantics (https://github.com/microsoft/pyright/discussions/6488#discussioncomment-12030176 is IMO useful to see).

asukaminato0721:
# Summary

Fixes #3959

Added opt-in implicit-bool diagnostic, disabled by default.

Detects non-bool values in conditions, not, assertions, comprehensions, ternaries, and boolean operators.

Allows explicit bool(...), actual bool values, and Any.

# Test Plan

Added tests and doc.

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -193,6 +193,8 @@ pub enum ErrorKind {
     /// to `Any`.
     /// This is a sub-kind of [ImplicitAny]: suppressing `implicit-any` also suppresses this error.
     ImplicitAnyTypeArgument,
+    /// A non-`bool` value is used in a boolean context, such as an `if` condition.
+    ImplicitBool,
     /// Usage of a module that was not actually imported, but does exist.
     ImplicitImport,
     /// Importing a name from a module that only made it available via a plain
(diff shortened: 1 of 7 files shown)


kinto0 (approved):
looks good but why is mypy primer screaming? is it not opt-in?

asukaminato0721:
> looks good but why is mypy primer screaming? is it not opt-in?

  Mypy-primer deliberately overrides that by invoking Pyrefly with:

  --preset=all

grievejia (approved):

PR #4275, opened 2026-07-24
Report dict literal element type errors locally (#4275)

### LINKED ISSUE #893: Exception message for incorrect dict literal is too wide

rchen152:
Summary:

Fixes https://github.com/facebook/pyrefly/issues/893.

Makes `dict_items_infer_inner` immediately report errors when it encounters them on keys and values, so that errors are reported on the exact line at which the bad key or value appears, with a precise message.

As a side effect, this also causes all of these errors to be reported under [bad-assignment] instead of their original error kind, because we have to choose an error kind when creating a new TypeCheckContext. This is consistent with what we already do for TypedDict errors.

The error count increases reported by the error delta jobs are because we previously reported only a single error if a dict literal contained multiple items with incorrect types, and now we report one error per bad item.

Before:
```
ERROR `dict[str, int]` is not assignable to `dict[int, int]` [bad-assignment]
 --> empty.py:1:21
  |
1 | d: dict[int, int] = {"oops": 42}
  |    --------------   ^^^^^^^^^^^^
  |    |
  |    declared type
  |
```

After:
```
ERROR `Literal['oops']` is not assignable to dict key type `int` [bad-assignment]
 --> empty.py:1:22
  |
1 | d: dict[int, int] = {"oops": 42}
  |                      ^^^^^^
  |
```

Reviewed By: stroxler

Differential Revision: D108382364

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -735,13 +735,13 @@
   "aliases_recursive.py": [
     {
       "code": -2,
-      "column": 12,
-      "concise_description": "`dict[str, complex | int]` is not assignable to `dict[str, Json] | float | int | list[Json] | str | None`",
-      "description": "`dict[str, complex | int]` is not assignable to `dict[str, Json] | float | int | list[Json] | str | None`",
+      "column": 26,
+      "concise_description": "`complex` is not assignable to dict value type `Json`",
+      "description": "`complex` is not assignable to dict value type `Json`",
       "line": 19,
       "name": "bad-assignment",
       "severity": "error",
-      "stop_column": 29,
+      "stop_column": 28,
       "stop_line": 19
     },
     {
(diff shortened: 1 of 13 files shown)


stroxler (approved):

PR #4277, opened 2026-07-24
[basedpyright parity] Warn about provably invalid casts

### LINKED ISSUE #3725: new rule - invalid cast
basedpyright's [`reportInvalidCast`](https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportinvalidcast) rule detects casts that are likely a mistake by validating that the types overlap:

```py
from typing import cast


cast(str, 1)
```

```
Conversion of type `Literal[1]` to type `str` may be a mistake because neither type sufficiently overlaps with the other.  If this was intentional, convert the expression to `object` first.
```

[(typescript also does the same thing)](https://www.typescriptlang.org/play/?#code/IwAghgziEC4E4EsB2BzIA)

it would be nice if pyrefly had a rule for this as well

fangyi-zhou:
# Summary

Fixes #3725 

Casts between runtime-disjoint types often hide incorrect assumptions, but uncertain overlap should not produce noise.

Add an opt-in invalid-cast diagnostic backed by conservative intersection checks that erase runtime-irrelevant generic arguments, honor disjoint bases and restricted type variables, and preserve plausible overlap for protocols, open classes, Never, and TypedDict/dict values.

# Test Plan

New unit tests

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -222,6 +222,8 @@ pub enum ErrorKind {
     InvalidAnnotation,
     /// Passing an argument that is invalid for reasons besides type.
     InvalidArgument,
+    /// Casting between types that are provably disjoint.
+    InvalidCast,
     /// A method-only decorator was applied to a top-level function.
     /// e.g. using `@final` or `@override` on a top-level function.
     /// Defaults to `warn` because such usage is harmless at runtime and is
@@ -529,6 +531,7 @@ impl ErrorKind {
             ErrorKind::ImplicitlyDefinedAttribute => Severity::Ignore,
             ErrorKind::IncompatibleComparison => Severity::Ignore,
             ErrorKind::InvalidAbstractMethod => Severity::Ignore,
+            ErrorKind::InvalidCast => Severity::Ignore,
             ErrorKind::InvalidDecorator => Severity::Warn,
             ErrorKind::MisplacedIgnore => Severity::Warn,
             ErrorKind::MissingOverrideDecorator => Severity::Ignore,
@@ -661,6 +664,11 @@ mod tests {
     #[test]
     fn test_unknown_column_default_severity() {
         assert_eq!(ErrorKind::UnknownColumn.default_severity(), Severity::Error);
+    }
+
+    #[test]
+    fn test_invalid_cast_default_severity() {
+        assert_eq!(ErrorKind::InvalidCast.default_severity(), Severity::Ignore);
     }
 
     #[test]
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -667,11 +667,6 @@ mod tests {
     }
 
     #[test]
-    fn test_invalid_cast_default_severity() {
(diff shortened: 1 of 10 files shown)


pyrefly/lib/alt/narrow.rs
```diff
@@ -304,6 +298,59 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+    /// Whether two types have no possible runtime value in common.
+    pub fn is_provably_disjoint(&self, left: &Type, right: &Type) -> bool {
+        let is_dict = |ty: &Type| {
+            matches!(
+                ty,
+                Type::ClassType(cls) | Type::SelfType(cls)
+                    if self.has_superclass(cls.class_object(), self.stdlib.dict_object())
+            )
+        };
+        if (left.is_typed_dict() && is_dict(right)) || (right.is_typed_dict() && is_dict(left)) {
```
connernilsen: note: we may want to record these as disjoint for the reasons basedpyright does: https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportinvalidcast

connernilsen (approved):
I'm going to flip the bool we return for typeddict <-> dict, but otherwise I think this looks great. Thanks for working on this!

pyrefly/lib/alt/narrow.rs
connernilsen: Actually, I take that back, and agree with having no error here. I get why it's there for basedpyright, but I think it makes sense to be able to convert between these types

connernilsen (approved):
I'm going to flip the bool we return for typeddict <-> dict, but otherwise this looks great. Thanks for working on this!

crates/pyrefly_config/src/error_kind.rs
```diff
@@ -666,6 +666,11 @@ mod tests {
         assert_eq!(ErrorKind::UnknownColumn.default_severity(), Severity::Error);
     }
 
+    #[test]
+    fn test_invalid_cast_default_severity() {
+        assert_eq!(ErrorKind::InvalidCast.default_severity(), Severity::Ignore);
```
rchen152: nit: I'd remove this test. IMO it's basically just a change detector.

pyrefly/lib/alt/narrow.rs
```diff
@@ -304,6 +298,59 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+            return false;
+        }
+        let normalize = |ty: &Type| {
```
rchen152: This is a pretty complicated set of exclusions! Looking at the test cases and the basedpyright docs, it looks like we mainly care about classes and class instances? That makes me wonder if we can start by excluding anything that's not a `Type::ClassType` or a `Type::ClassDef` (or a union of such), then use what we can now assume about the structure of the types to simplify the logic.

rchen152 (changes requested):
Thanks for the PR! Back to you with a design question.

pyrefly/lib/alt/narrow.rs
fangyi-zhou: I've expanded upon your original suggestions (by adding handling for literals, tuples, None), let me know what you think!

rchen152 (approved):
LGTM - thank you!

PR #4283, opened 2026-07-24
fix bug: generic ABC is mistakenly allowed to be instantiated by providing a type argument to the constructor #4263

### LINKED ISSUE #4263: bug: generic ABC is mistakenly allowed to be instantiated by providing a type argument to the constructor
### Describe the Bug

```python

class EyBeeSea(ABC):
    @abstractmethod
    def method(self) -> None:
        raise NotImplementedError

def test_concrete_abc() -> None:
    shouldnt = EyBeeSea()
    # as expected, cant instantiate abstract class


class GenericEyBeeSea[A](ABC):
    @abstractmethod
    def method(self) -> A:
        raise NotImplementedError

def test_generic_abc_no_arg() -> None:
    shouldnt: GenericEyBeeSea[int] = GenericEyBeeSea()
    # as expected, cant instantiate abstract class
def test_generic_abc_with_arg() -> None:
    shouldnt: GenericEyBeeSea[int] = GenericEyBeeSea[int]() 
    # unexpectedly, when you add the type argument to the constructor, this is allowed?
    # i think the "abstractness" gets lost somewhere here
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEq2AxnRDcbpQC50CCAQgGEANA2xwulVEy40YXABa5MAHXRqmUVHDh0AoqX4wYAZRioAFAMEBKRGrqO6AAUYSpMuYuUOnmGGB0XkqYFnAwUGA2dAC0AHx0AHK46DD26E6ZdFIQ4Um4XACS7LBy6FwwmHqU1JRqav6BFRIA%2BkwpTJTyMC2MTBbR8fmp6ZlwSgCuUJjldAC8%2BobGZpY2vo4AxAy6MPjEMDKVokwYPBDoEqcQqBVi7tI8mtpw9eroTzp0AOIwqZQQTAMRlM5gA2rwALpWIR2dYuNySB7BHwZPwBILyEJhCJRWIJXijLKOHJ5ZJFEowMoVKo1TivRp0ZpcFoAc1%2BMH%2BTF6zBa6FwvUoLIGeOGaTh41wUxmXEQ33ZnKByzB5y4EPmcr%2BAMVINWcK22jou32h0wx1OrAuXCuNxgd0RMjoHxeWHRTNZ8oB3K5AHcIIoBULBglkiNxZNpuVZT9NYCljrQSq1QtoxytXGVgnylDonq6BNUnsDtSoKRRN6FL86KRJQxMJhGRXGaR9gxBRMqYzcA3be1LZQJjJOKJFLlWLpUFAoLhvZUAPy5iAN84Aa27dBUIARHi4qR0G7obK4uinEjocFoMHLHNtFa6IGEIDIXTAJcIXFoUAoWwACqRnyWzwwHACEdFJIBZCYpC4CAUkINQtjMG8uC4Yg4EQAB6dCnwCV9OBZdDfnQzBcCYOB0N7cDIJuGD0HQugwE4BgADdUGgRhYFA9BKKgmi6FwYhoJSOA4PQMhvHQGImI5OBeIWDcAGZCAARgAJg3NRQQ5Wo4AhNR8zYDhuEqGJMAgLoZAgKT1QAcggFk%2BS6ay1D5LgYi6ABHCYzOM5cYFIGJpCYeBdAWazvVQSh0Cc9AQAAXwfB5LJgAAxaAYAoNAsDwIgyDioA

_No response_

asukaminato0721:
# Summary

Fixes #4263

Preserves direct-class provenance for `GenericABC[int]()`, rejects direct specialized abstract/protocol construction.

Keeps `type[GenericABC[int]]` calls valid because they may reference concrete subclasses.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_types/src/display.rs
+++ b/crates/pyrefly_types/src/display.rs
@@ -944,6 +944,12 @@ impl<'a> TypeDisplayContext<'a> {
                 output.write_qname(cls.qname())?;
                 output.write_str("]")
             }
+            Type::SpecializedClass(class_type) => {
+                output.write_str("type[")?;
+                output.write_qname(class_type.qname())?;
+                output.write_targs(class_type.targs())?;
+                output.write_str("]")
+            }
             Type::ClassType(class_type)
                 if class_type.qname().module_name().as_str() == "builtins"
                     && class_type.qname().id().as_str() == "tuple"
--- a/crates/pyrefly_types/src/type_alias.rs
+++ b/crates/pyrefly_types/src/type_alias.rs
@@ -65,6 +65,11 @@ impl TypeAlias {
     pub fn as_value(&self, stdlib: &Stdlib) -> Type {
         if self.style == TypeAliasStyle::Scoped {
             stdlib.type_alias_type().clone().to_type()
+        } else if let Type::Type(inner) = self.ty.as_ref()
+            && let Type::ClassType(cls) = inner.as_ref()
+            && !cls.targs().is_empty()
+        {
+            Type::SpecializedClass(cls.clone())
         } else {
             *self.ty.clone()
         }
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -966,12 +966,14 @@ pub enum Type {
     /// up the class `tparams` and setting defaults using gradual types: for
     /// example `list` in an annotation position means `list[Any]`.
     ClassDef(Class),
+    /// An exact class object with explicit type arguments, such as the value of
+    /// the expression `C[int]`.
+    SpecializedClass(ClassType),
     /// A value that indicates a concrete, instantiated type with known type
     /// arguments that are validated against the class type parameters. If the
     /// class is not generic, the arguments are empty.
     ///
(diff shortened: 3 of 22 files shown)


asukaminato0721:
wow

steam.ID uses ABCMeta, but without `@abstractmethod`.

samwgoldman (changes requested):

samwgoldman (changes requested):
The underlying problem here is that Pyrefly represents two things the same way, but we want different behavior for them

```
class C[T]:
  ...
X = C[int] # X is `Type::Type(ClassType(...))`
Y: type[C[int]] = C # Y is `Type::Type(ClassType(...))`
```

This commit works around this by adding a `direct_class_specialization` bool, but this fixes the problem in a quite limited way. A simple indirection can also work around the missing error:

```
Ctor = Shape[int]
Ctor()  # still accepted
```

I think we should first improve the representation of specialized class types so we can distinguish between expressions with the annotation `type[C[int]]` and the type of the expression `C[int]`. Then, I think this fix becomes trivial.

PR #4286, opened 2026-07-24
Resolve bundled modules with import semantics

### LINKED ISSUE #1819: `pd.read_csv` inferred type is Unknown but go to definition works.

samwgoldman:
Summary:
Before this diff, the bundled stubs were created by simply inserting file paths in order directly into a module map. This means the modules depend on the order of files in the archive, but two different paths can sometimes provide the same module, and Python's own precedence rules do not match.

In practice, there is only one example where this is a problem, which is exactly the issue in https://github.com/facebook/pyrefly/issues/1819: `pandas/io/parsers.pyi` vs `pandas/io/parsers/__init__.pyi`.

While the contents of the archives is fixed at compile time, we are not fully in control of the archives -- they come from typeshed or stubs repos written by someone else. This PR implements a new `Bundle` abstraction that takes an archive and uses Python runtime lookup rules to create the accurate module map.

I added tests for other cases which don't currently exist in the current archives, but if they ever appear we will handle them correctly.

This is an alternative approach to https://github.com/facebook/pyrefly/pull/4185

Fixes https://github.com/facebook/pyrefly/issues/1819

Differential Revision: D113606265

### DIFF
--- a/pyrefly/lib/module/bundled.rs
+++ b/pyrefly/lib/module/bundled.rs
@@ -25,6 +25,7 @@ use pyrefly_python::module_path::ModulePath;
 use pyrefly_util::arc_id::ArcId;
 use pyrefly_util::fs_anyhow;
 use pyrefly_util::lock::Mutex;
+use starlark_map::small_map::SmallMap;
 use tempfile::NamedTempFile;
 
 pub fn set_readonly(path: &Path, value: bool) -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


stroxler (approved):

PR #4287, opened 2026-07-24
Run benchmarks on CodSpeed in CI

kinto0:
Summary:
Wires the benchmark suite into CodSpeed so type-checker performance is tracked continuously in CI. All benchmarks are written against criterion, so this swaps the plain `criterion` dev-dependency for CodSpeed's drop-in `codspeed-criterion-compat` (aliased back to `criterion`, so no benchmark source changes): under `cargo codspeed` it emits instrumented measurements, and otherwise it behaves exactly like criterion.

Adds `.github/workflows/codspeed.yml` with two jobs, matched to each benchmark's nature:
- `simulation` runs the `micro` benchmarks under CodSpeed's instruction-count (Valgrind) mode. They are small, deterministic, and single-threaded, so callgrind's overhead is bounded and the counts are stable run-to-run.
- `walltime` runs the `pytorch` benchmarks in wall-clock mode. They drive the full LSP server over all cores against the pinned PyTorch checkout (threads, I/O, a multi-gigabyte cold start), none of which survive Valgrind.

The workflow runs on every `pull_request` and on `main` pushes (to seed baseline data for PR comparisons), plus `workflow_dispatch`. Both jobs run on `ubuntu-latest`; any dedicated-runner change is left to a separate diff.

Both jobs authenticate to CodSpeed over OIDC (`id-token: write`, no token secret) and use `CodSpeedHQ/action@v4`. A `[profile.bench]` in the root Cargo.toml disables release LTO and uses multiple codegen units, keeping the CodSpeed build within standard CI runners' memory budget without affecting instruction-count measurements.

Differential Revision: D112014630

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,0 +1,67 @@
+name: CodSpeed
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+  # workflow_dispatch lets CodSpeed trigger backtest runs to seed baseline data.
+  workflow_dispatch:
+permissions:
+  contents: read
+  id-token: write # OpenID Connect auth with CodSpeed (no token secret required)
+jobs:
+  # Instruction-count (Valgrind) measurement of the small, deterministic
+  # microbenchmarks. They are single-threaded and fast, so callgrind's overhead
+  # is bounded and the instruction counts are stable run-to-run.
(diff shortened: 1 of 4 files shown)


samwgoldman (approved):

grievejia (approved):

PR #4289, opened 2026-07-25
Add implicit-relative-import diagnostic

### LINKED ISSUE #4174: Add a diagnostic for implicit relative imports
## Summary

Add detection of implicit relative imports equivalent to Basedpyright's `reportImplicitRelativeImport` rule.

Documentation:

https://docs.basedpyright.com/latest/benefits-over-pyright/new-diagnostic-rules/#reportimplicitrelativeimport

## Existing Pyrefly handling

Pyrefly recognizes `reportImplicitRelativeImport` while migrating Pyright
configuration and maps it to `missing-import`:

https://github.com/facebook/pyrefly/commit/46dacddf84499d36562d0d9f71fad5893e7948e4

This issue is for first-class detection of the more specific implicit-relative-
import condition rather than treating it solely as a general missing import.

## Desired behavior

- Detect an unqualified import that cannot be resolved through the normal absolute import roots but resolves only by searching from the importing file’s directory upward toward the configured execution root.
- Explain that the import may fail when the file is imported as part of a package.
- Suggest either an explicit relative import or the full absolute package path.
- Do not report imports that resolve normally through configured import roots, search paths, the standard library, or installed packages.
- Provide a dedicated, independently configurable diagnostic such as implicit-relative-import.
- Define an appropriate default severity; Basedpyright enables the equivalent diagnostic as an error in its default recommended mode.

nitishagar:
Fixes #4174

## Summary

Adds a first-class `implicit-relative-import` diagnostic that fires when an unqualified `import` resolves *only* via the directory-upward fallback search path, not through any configured absolute root. Such imports are fragile (they break when the importing file moves and can unexpectedly shadow installed packages).

Today `reportImplicitRelativeImport` from pyright configs collapses onto the generic `missing-import` kind, giving users no way to tune implicit-relative imports apart from genuine missing imports. This PR gives the condition its own independently configurable kind.

## Approach

- New `ErrorKind::ImplicitRelativeImport` (default `Severity::Ignore`), inserted in its lexicographically-correct slot, with its own `default_severity` arm and **no `parent_kind`** — so `[missing-import]` suppression does not silence it.
- New `FindError::ImplicitRelativeImport(module)` threaded via the existing non-fatal-error channel (`Finding.error` through `with_error`) — the same idiom already used for `UntypedImport`/`MissingSource`. The emitter already reads this channel, so there is zero new plumbing through loader/bindings/solver.
- The attach lives **entirely inside** the `!disable_search_path_heuristics` fallback arm of `find_import_internal`, so earlier (absolute) tiers and the disabled-heuristics case never flag. The `with_error` no-op-when-set guard means an existing `UntypedImport` caveat correctly takes precedence.
- Retargets the pyright migration: `reportImplicitRelativeImport` → `ImplicitRelativeImport`; `reportMissingImports` stays on `missing-import`.
- Documents the kind in `website/docs/error-kinds.mdx`.

## Out of scope (known limitation)

Namespace packages resolved via the fallback tier are **not** flagged by this PR. Threading provenance through the cross-tier namespace accumulator (which merges namespaces from all tiers) is a larger change with disproportionate blast radius for an initial PR. A `bug`-marked test documents this limitation so the gap is visible and trackable for a follow-up.

## Cache safety

The implicit-relative signal is origin-dependent. It is safe to attach via the non-fatal channel because fallback-tier hits are filesystem paths (never bundled typeshed variants), so `is_bundled()` is `false` and they are never promoted to the origin-independent `(module, None)` cache entry. This invariant is documented in a code comment at the attach site and pinned by a cache-safety test.

## Testing

- **Inline Rust**: finder tests for I1 (no false positive when search path resolves), I2/I4 (caveat attached on fallback resolution, deterministic), I3 (no flag when heuristics disabled), precedence (UntypedImport wins), namespace `bug` test (scoped-out limitation); loader tests for the `with_error` precedence contract and cache-safety (no promotion to origin-independent key); config test for the pyright migration mapping.
- **Scrut e2e** (`test/implicit_relative_import.md`): surfacing, no-false-positive when on search path, default-ignore, independent suppression (`[missing-import]` does not silence; `[implicit-relative-import]` does).
- `test_doc_headers` / `test_doc_severities` pass (mdx parity).

All new tests pass; `./test.py` format + lint introduces no new lint errors.

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -201,6 +201,13 @@ pub enum ErrorKind {
     /// only re-exported when redundantly aliased (`from x import y as y`),
     /// listed in `__all__`, or brought in via a wildcard import.
     ImplicitReexport,
+    /// An unqualified `import` resolved only via the directory-upward fallback
+    /// search path, not through any configured absolute root. Such imports are
+    /// fragile (they break when the importing file moves and can unexpectedly
+    /// shadow installed packages). Prefer an explicit relative import
+    /// (`from . import ...`) or add the module's root to the configured search
+    /// path.
+    ImplicitRelativeImport,
     /// An attribute was implicitly defined by assignment to `self` in a method that we
     /// do not recognize as always executing (we recognize constructors and some test setup
     /// methods).
@@ -534,6 +541,7 @@ impl ErrorKind {
             ErrorKind::ImplicitAnyTypeArgument => Severity::Ignore,
             ErrorKind::ImplicitImport => Severity::Warn,
             ErrorKind::ImplicitReexport => Severity::Ignore,
+            ErrorKind::ImplicitRelativeImport => Severity::Ignore,
             ErrorKind::ImplicitlyDefinedAttribute => Severity::Ignore,
             ErrorKind::IncompatibleComparison => Severity::Ignore,
             ErrorKind::InvalidAbstractMethod => Severity::Ignore,
--- a/crates/pyrefly_config/src/migration/pyright.rs
+++ b/crates/pyrefly_config/src/migration/pyright.rs
@@ -541,7 +541,7 @@ impl RuleOverrides {
         // Import rules
         add(
             self.report_implicit_relative_import,
-            ErrorKind::MissingImport,
+            ErrorKind::ImplicitRelativeImport,
         );
 
         // Type argument rules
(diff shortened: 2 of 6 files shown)


stroxler (approved):

nitishagar:
@stroxler Anything I need to do here?

Issue #2139, opened 2026-01-17
Python Autocompletion for Polars/Pandas Data Frame Column names is not working correctly and is very buggy


PGupta-Git:
### Describe the Bug

As mentioned in the discussion and a similar thread posted by me in Positron:

Python Autocompletion for Polars/Pandas Data Frame Column names, similar to how PyCharm/DataSpell provide autocompletion for column names

Can't seem to trigger it. It only works when using square brackets, and even then sometime bugs out

https://github.com/posit-dev/positron/discussions/11385#discussioncomment-15518884

https://github.com/facebook/pyrefly/discussions/2134

https://github.com/posit-dev/positron/issues/5436

DataSpell/Pycharm

<img width="1450" height="617" alt="Image" src="https://github.com/user-attachments/assets/330e2cd5-78ae-462b-be60-11dc5c3b58ad" />

Pyrefly

<img width="2074" height="1343" alt="Image" src="https://github.com/user-attachments/assets/06a80005-9759-4d5c-bad5-9096ec6f2185" />

<img width="2074" height="1343" alt="Image" src="https://github.com/user-attachments/assets/e0b6525d-e7ce-4ef0-888d-8542e8881bb6" />

<img width="2074" height="1343" alt="Image" src="https://github.com/user-attachments/assets/28b81db4-118e-490a-aff3-af1ba3962ee3" />

jvansch1:
Thanks for flagging this! Would you be able to share your Pyrefly config file? Also which version of Pyrefly are you seeing this on?

PGupta-Git:
I am not using any specific Pyrefly config file. I am using the latest pyrefly 0.49.0

But this issue has been there since the time I started using pyrefly since last month

yangdanny97:
I don't think this is a feature we support right now, but it might be something we can build.

We support literal TypedDict key completion for subscripts, but don't do anything special for function calls like like `select()`, `col()` etc.

PGupta-Git:
I would be highly grateful if the team could add this feature. This will be a boon for data science folks like me.

yangdanny97:
@shobhitmehro this would be a good extension to what you're building, i see that @asukaminato0721 has started working on something for this so maybe you can collaborate

PR #4291, opened 2026-07-25
fix part of Autocompletion for Polars/Pandas Data Frame Column names in function calls #2139

asukaminato0721:
# Summary

Fixes part of #2139

adds schema-aware column completion for Polars and pandas DataFrames.

detects DataFrame-typed receivers or earlier call arguments and reads column names directly from the inferred DataFrameSchema. This supports both direct calls such as `df.select("")` and nested expressions such as `df.select(pl.col(""))`.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/lsp/dict_completions.rs
+++ b/pyrefly/lib/state/lsp/dict_completions.rs
@@ -38,9 +38,8 @@ enum DictKeyLiteralContext {
         base_expr: Expr,
         literal: ExprStringLiteral,
     },
-    /// A string literal in a call argument whose completions should come from
-    /// surrounding container expressions.
-    /// Examples: `lookup(data, "na|")`, `df.select(col("na|"))`.
+    /// A string literal in a call containing a DataFrame expression.
+    /// Examples: `df.select("na|")`, `df.select(col("na|"))`.
     CallArgument {
         source_exprs: Vec<Expr>,
         literal: ExprStringLiteral,
@@ -85,10 +84,23 @@ impl<'a> Transaction<'a> {
         }
     }
 
+    fn type_contains_dataframe(ty: &Type) -> bool {
+        match ty {
+            Type::DataFrame(_) => true,
+            Type::Union(u) => u.members.iter().any(Self::type_contains_dataframe),
+            _ => false,
+        }
+    }
+
     fn expr_has_typed_dict_type(&self, handle: &Handle, expr: &Expr) -> bool {
         self.get_type_trace(handle, expr.range())
             .map(|ty| Self::type_contains_typed_dict(&ty))
             .unwrap_or(false)
+    }
+
+    fn expr_has_dataframe_type(&self, handle: &Handle, expr: &Expr) -> bool {
+        self.get_type_trace(handle, expr.range())
+            .is_some_and(|ty| Self::type_contains_dataframe(&ty))
     }
 
     /// Extracts typed dict access from `.get()` method calls.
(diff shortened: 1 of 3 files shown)


pyrefly/lib/state/lsp/dict_completions.rs
```diff
@@ -315,6 +374,11 @@ impl<'a> Transaction<'a> {
                                 .or_insert_with(|| field.ty.clone());
                         }
                     }
+                    Type::DataFrame(schema) => {
+                        for (name, ty) in schema.columns {
```
shobhitmehro: Thank you for this! I am working on the column inference side for polars/pandas so things may shift. My stack changes schema.columns from Type to PolarsDType, holding polars dtypes

pyrefly/lib/state/lsp/dict_completions.rs
```diff
@@ -205,21 +203,82 @@ impl<'a> Transaction<'a> {
...
+            } else if call.arguments.keywords.iter().any(|kw| {
+                kw.value.range().start() <= literal_range.start()
+                    && literal_range.end() <= kw.value.range().end()
+            }) {
+                call.arguments.args.len()
+            } else {
+                continue;
+            };
+
+            if let Expr::Attribute(attr) = call.func.as_ref()
```
shobhitmehro: This promotes the receiver to a column source for any method, so it fires on calls that have nothing to do with columns. For example df.write_csv("|") would show autocomplete suggestions to column names when it shouldn't. When my column-inference stack lands, `polars_specials.rs` would have a lot of helper methods that could gate this part on certain methods.

pyrefly/lib/state/lsp/dict_completions.rs
```diff
@@ -205,21 +203,82 @@ impl<'a> Transaction<'a> {
...
+                && self.expr_has_dataframe_type(handle, attr.value.as_ref())
+                && !source_exprs
+                    .iter()
+                    .any(|existing: &Expr| existing.range() == attr.value.range())
+            {
+                source_exprs.push(attr.value.as_ref().clone());
+            }
+            for expr in call.arguments.args.iter().take(positional_count) {
```
shobhitmehro: could we gate this loop so it only runs when the call is actually a column consuming DataFrame method like select, with_columns, etc? right now it grabs any DataFrame positional arg regardless of the callee, so something like this suggests df's columns for a random string arg
```
  import polars as pl
  df = pl.DataFrame({"name": [1]})
  def f(frame, s): ...
  f(df, "|")   # offers "name"

File: pyrefly/lib/test/lsp/completion.rs [hunk 4]
```diff
@@ -353,6 +401,66 @@ cfg: Config = {"": 1}
...
+#                   ^
+"#;
+    let (handles, state) = mk_multi_file_state(&[("main", code)], Require::Exports, false);
+    let position = extract_cursors_for_test(code)[0];
+    assert_eq!(
+        dict_field_labels(&state.transaction(), handles.get("main").unwrap(), position),
+        Vec::<String>::new()
+    );
+}
+
```
shobhitmehro, line 483: for a union of DataFrames this merges every member's columns instead of intersecting them

 something like this
```
  import polars as pl
  def f(cond: bool):
      a = pl.DataFrame({"id": [1], "x": [1]})
      b = pl.DataFrame({"id": [1], "y": [1]})
      df = a if cond else b          # DataFrame[id,x] | DataFrame[id,y]
      df.select("|")                 # autocomplete options are id, x, y but only id is safe
```
picking x when df is actually b would give us a runtime, so for unions i think we only want the columns present in every arm

yangdanny97:
noice

PR #4296, opened 2026-07-26
Add method-assign diagnostic for method rebinding

### LINKED ISSUE #4161: Feature request: implement `method-assign` from `mypy`
### Describe the Bug

`mypy` [explains it adequately](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-assignment-target-is-not-a-method-method-assign):

> In general, assigning to a method on class object or instance (a.k.a. monkey-patching) is ambiguous in terms of types, since Python’s static type system cannot express the difference between bound and unbound callable types.

This would be a useful feature to explicitly disallow any monkey patching, which significantly obfuscates code.

```
class A:
    def f(self) -> None: pass
    def g(self) -> None: pass

def h(self: A) -> None: pass

A.f = A.g
```

For reference, `mypy` returns:
```
$ mypy --strict example.py 
example.py:7: error: Cannot assign to a method  [method-assign]
```

Tests often do this for mocking unless they use `mock.patch`, so it probably shouldn't be on by default unless there is special handling for `Mock`, etc.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQCCiAOuvR-ZjGPWABRwYUMAEp6AWgB89AHK50MRPWK04bTlx70A5oOFjJM%2BYuWq6bNt14ALfSOWNx0uQqUq1l9I0K8AvEyEOmwgADQgZABOPFCkhAAuuAC2UBQAxPQACqTRYLH0aFh4%2BPSUCpA6AK6RqPEQCoRsGQDKMDD0NvHxxHCIAPR9UTFxuJE6fTDofZi4lHB9ZegV1bX1U3yj9KgAbqjQqNiwpeUQVTV1CvS4xBfocI3oZPE2ChLbMJFwa-QBLCAAzIQAIwAJj%2BbGI0SE8R%2B9D%2BqCgUD%2BIAAvuFUJQ6u8AGLQGAUQo4AgkcgooA

_No response_

nitishagar:
## Summary

Adds mypy-equivalent `method-assign` diagnostic (issue #4161): direct assignment that rebinds a method (`A.f = ...`, `a.f = ...`) is type-ambiguous and now flagged when the kind is enabled.

**Why:** pyrefly had no signal for method rebinding — a type-*compatible* rebind was silently allowed, and an incompatible one only surfaced the generic `bad-assignment`. mypy's `method-assign` exists precisely for this.

## Approach

- New top-level, addressable `ErrorKind::MethodAssign`, off by default (`Severity::Ignore`), enabled in `Preset::Strict` (auto-promoted in `Preset::All`). Independently suppressible via `# pyrefly: ignore[method-assign]`; does not silence the existing type check.
- Detection keys off the authoritative `ClassFieldInner::Method` origin, threaded as a provenance flag on `ClassAttribute::ReadWrite(Type, bool)` (mirroring the existing `ReadOnly(Type, ReadOnlyReason)` precedent) from the two Method construction sites to the single attribute-write chokepoint `check_class_attr_set_and_infer_narrow`. That one dispatcher covers both class-object (`A.f=`) and instance (`a.f=`) writes. Emitting before the type-compat check means an incompatible rebind yields both `method-assign` and `bad-assignment`.
- `@staticmethod` is also stored as `ClassFieldInner::Method` (per `is_method()`), so the flag is guarded with `!ty.is_staticmethod()` (new `Type::is_staticmethod()` helper, following the `is_abstract_method()` precedent). This excludes staticmethods, plain data attributes, properties, and `Callable`-typed attributes — matching the method-rebinding intent without false positives on `Callable`-typed class attributes (which flatten to `Type::Function` indistinguishably from a real method).

## Behavior matrix

| Target | Flagged |
|---|---|
| instance method (class + instance rebind) | ✅ |
| `@classmethod` (class + instance rebind) | ✅ |
| inherited method (via MRO) | ✅ |
| `@staticmethod` | ❌ |
| property / descriptor / nested class | ❌ |
| plain data attribute | ❌ |
| `Callable`-typed attribute | ❌ |
| `setattr(A, "f", x)` | ❌ (out of scope, matches mypy) |

`setattr` rebinding is out of scope by construction (dispatches to `__setattr__`, bypassing the attribute-set check path).

## Test plan

- [x] `test_method_assign_diagnostic` — full fidelity matrix (flagged + not-flagged + inherited + setattr + coexistence with `bad-assignment`)
- [x] `test_method_assign_off_by_default` — kind silent under default config
- [x] `test_doc_headers` / `test_doc_severities` / `test_preset_doc` — doc + sort + preset coupling
- [x] regression sweep: attributes, class_overrides, typed_dict, dataclass, enums, decorators, protocol, callable, assign — all green (1152 lib tests)
- [x] `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` — clean (no new lint errors)

Closes #4161.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -115,7 +115,9 @@ use crate::types::types::Type;
 #[derive(Debug, Clone)]
 pub enum ClassAttribute {
     /// A read-write attribute with a closed form type for both get and set actions.
-    ReadWrite(Type),
+    /// The flag is `true` when the attribute originated as a method
+    /// (`ClassFieldInner::Method`), used to flag method-rebinding assignments.
+    ReadWrite(Type, bool),
     /// A read-only attribute with a closed form type for get actions.
     ReadOnly(Type, ReadOnlyReason),
     /// A `NoAccess` attribute indicates that the attribute is well-defined, but does
@@ -131,7 +133,13 @@ pub enum ClassAttribute {
 
 impl ClassAttribute {
     pub fn read_write(ty: Type) -> Self {
-        Self::ReadWrite(ty)
+        Self::ReadWrite(ty, false)
+    }
+
+    /// Like [`read_write`](Self::read_write) but marks the attribute as having
+    /// originated as a method (`ClassFieldInner::Method`).
+    pub fn read_write_method(ty: Type) -> Self {
+        Self::ReadWrite(ty, true)
     }
 
     pub fn read_only(ty: Type, reason: ReadOnlyReason) -> Self {
@@ -152,7 +160,7 @@ impl ClassAttribute {
 
     pub fn read_only_equivalent(self, reason: ReadOnlyReason) -> Self {
         match self {
-            Self::ReadWrite(ty) => Self::ReadOnly(ty, reason),
+            Self::ReadWrite(ty, _) => Self::ReadOnly(ty, reason),
             Self::Property(getter, _, cls) => Self::Property(getter, None, cls),
             Self::Descriptor(descriptor, base) => Self::Descriptor(
                 Descriptor {
@@ -173,7 +181,7 @@ impl ClassAttribute {
         match self {
             // TODO(stroxler): ReadWrite attributes are not actually methods but limiting access to
             // ReadOnly breaks unit tests; we should investigate callsites to understand this better.
-            ClassAttribute::ReadWrite(ty) | ClassAttribute::ReadOnly(ty, _) => Some(ty),
+            ClassAttribute::ReadWrite(ty, _) | ClassAttribute::ReadOnly(ty, _) => Some(ty),
             ClassAttribute::NoAccess(..)
             | ClassAttribute::Property(..)
             | ClassAttribute::Descriptor(..) => None,
@@ -1258,10 +1266,13 @@ fn bind_class_attribute(
     cls: &ClassBase,
     attr: Type,
     read_only_reason: Option<ReadOnlyReason>,
+    is_method: bool,
 ) -> ClassAttribute {
     let ty = make_bound_classmethod(heap, cls, attr).into_inner();
     if let Some(reason) = read_only_reason {
         ClassAttribute::read_only(ty, reason)
+    } else if is_method {
+        ClassAttribute::read_write_method(ty)
     } else {
         ClassAttribute::read_write(ty)
     }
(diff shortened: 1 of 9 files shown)


pyrefly/lib/alt/class/class_field.rs
```diff
@@ -115,7 +115,9 @@ use crate::types::types::Type;
 #[derive(Debug, Clone)]
 pub enum ClassAttribute {
     /// A read-write attribute with a closed form type for both get and set actions.
-    ReadWrite(Type),
+    /// The flag is `true` when the attribute originated as a method
+    /// (`ClassFieldInner::Method`), used to flag method-rebinding assignments.
+    ReadWrite(Type, bool),
```
samwgoldman, line 120: Instead of a `bool` here, I think it would be cleaner to make a separate variant `Method(Type)` and then define behavior for it, so everywhere that matches on `ClassAttribute` values would have a case for `ReadWrite` and `Method`.

Hopefully, the only remaining `ReadWrite` cases would be non-method class fields. If not, I'd be curious to know what other kinds of attributes are in `ReadWrite` currently.

pyrefly/lib/alt/class/class_field.rs
```diff
@@ -5252,13 +5287,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                ClassAttribute::Property(_, Some(_), _) | ClassAttribute::ReadWrite(..),
             ) => Err(Box::new(AttrSubsetError::ReadOnly)),
             (
                 // TODO(stroxler): Investigate this case more: methods should be ReadOnly, but
                 // in some cases for unknown reasons they wind up being ReadWrite.
-                ClassAttribute::ReadWrite(got),
-                ClassAttribute::ReadWrite(want),
+                ClassAttribute::ReadWrite(got, _),
+                ClassAttribute::ReadWrite(want, _),
             ) if got.has_toplevel_func_metadata() && want.has_toplevel_func_metadata() => {
```
samwgoldman, line 5297: This weird workaround supports a separate `Method` case, IMO. During subclass checks, we treat methods covariantly, so "ReadWrite" is really a misnomer. I bet this code would be much less awkward.

samwgoldman (changes requested):
This looks good to me, but I think it would be cleaner with a separate `ClassAttribute::Method` variant. Could you rebase and split this PR into two commits: one adding `ClassAttribute::Method` and one adding the error? Thank you!

samwgoldman:
Oh, could you also add the logic to support automatic config migration from mypy for this?

PR #4308, opened 2026-07-27
Report when overload argument-type expansion hits its limit

### LINKED ISSUE #4213: Report a warning if the argument expansion limit is hit
This limit is set to 100:

https://github.com/facebook/pyrefly/blob/7ffcd9ba2210a6f621d565fcf6504cfc707a64da/pyrefly/lib/alt/overload.rs#L91

If this limit is exceeded, then the expansion loop simply terminates silently, so it'l behave as if there's nothing left to expand. 

In overload-heavy libraries like NumPy and scipy-stubs, there might be functions or methods where there are more than 100 argument expansions needed, but there's no way of knowing.

So I propose that Pyrefly emits a warning if this argument expansion limit is exceeded. Then, if there turn out to many such warnings in practice, we can consider increasing the limit.

markselby9:
# Summary

Fixes #4213.

When resolving an overloaded call, Pyrefly performs [argument type expansion](https://typing.python.org/en/latest/spec/overload.html#argument-type-expansion) — splitting a `bool`/enum/union argument and trying each combination — bounded by a hard limit (`ArgsExpander::GAS`, currently 100). Previously, hitting that limit silently truncated the search, so a call needing more than 100 expansions could be reported as `no-matching-overload` (or resolve differently) with no indication. This PR has two parts (in separate commits):

**Warning.** Adds a new `overload-argument-expansion-limit` warning, emitted once against the call when the limit is reached without a match. It is not emitted once a call has already matched (e.g. `legacy_overload_expansion`'s union-return refinement, where truncating expansion doesn't change the result).

**Optimization.** Prunes expansion that cannot affect overload selection, so the limit is reached far less often (keeping the warning an error of last resort):
- An argument whose parameter type is identical across every candidate overload cannot disambiguate them, so it is left un-split — the pre-match analogue of the existing `should_materialize`.
- With a single arity-compatible overload there is nothing to disambiguate, so expansion is skipped and the call resolves to a precise argument error instead of exhausting the limit.

Both prunings are conservative — a position whose binding isn't statically determinable (call-site unpacking, `*args`/`**kwargs`, `ParamSpec`, differing arities) stays expandable — and exclude generic overloads, where expansion also drives type-variable solving. Resolution results are unchanged for calls that previously resolved; only wasted expansions (and spurious limit warnings) are removed.

# Test Plan

Added to `pyrefly/lib/test/overload.rs`, each sized so the limit warning would fire without the pruning:
- `test_overload_argument_expansion_limit_reported` / `test_overload_argument_expansion_limit_not_reported_when_matched` — the warning fires on genuine truncation, and stays silent once a call has matched.
- `test_overload_expansion_skips_non_disambiguating_args` / `test_overload_expansion_skipped_for_single_arity_candidate` — the two prunings.

`cargo test -p pyrefly --lib` and `cargo test -p pyrefly_config` (the `ErrorKind` ↔ `error-kinds.mdx` sync test) pass; `test.py` formatting and linting are clean.

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -329,6 +329,9 @@ pub enum ErrorKind {
     NotRequiredKeyAccess,
     /// Unpacking an open TypedDict that may contain a bad key via inheritance.
     OpenUnpacking,
+    /// Argument type expansion for an overloaded call reached its limit, so some argument
+    /// type combinations were left unchecked.
+    OverloadArgumentExpansionLimit,
     /// An error related to parsing or syntax.
     ParseError,
     /// A potential conflict between an explicit keyword argument and a NotRequired
@@ -560,6 +563,7 @@ impl ErrorKind {
             ErrorKind::NonConvergentRecursion => Severity::Warn,
             ErrorKind::NotRequiredKeyAccess => Severity::Ignore,
             ErrorKind::OpenUnpacking => Severity::Ignore,
+            ErrorKind::OverloadArgumentExpansionLimit => Severity::Warn,
             ErrorKind::PytorchEfficiencyLintCudaCall => Severity::Ignore,
             ErrorKind::PytorchEfficiencyLintItemCall => Severity::Ignore,
             ErrorKind::PytorchEfficiencyLintPrintTensor => Severity::Ignore,
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -88,6 +88,8 @@ pub struct ArgsExpander<'a, Ans: LookupAnswer> {
     arg_lists: Vec<(Vec<CallArg<'a>>, Vec<CallKeyword<'a>>)>,
     /// Hard-coded limit to how many times we'll expand.
     gas: Gas,
+    /// Set once `gas` is exhausted, so the caller can report the truncated expansion.
+    limit_hit: bool,
     solver: &'a AnswersSolver<'a, Ans>,
 }
 
@@ -107,6 +109,7 @@ impl<'a, Ans: LookupAnswer> ArgsExpander<'a, Ans> {
             },
             arg_lists: vec![(posargs, keywords)],
             gas: Gas::new(Self::GAS as isize),
+            limit_hit: false,
             solver,
         }
     }
@@ -172,6 +175,7 @@ impl<'a, Ans: LookupAnswer> ArgsExpander<'a, Ans> {
                     if self.gas.stop() {
                         // We've hit our hard-coded limit; stop expanding, and move `idx` past the
                         // end of the keywords so that subsequent `expand` calls know we're done.
+                        self.limit_hit = true;
                         self.idx = Either::Right(keywords.len());
                         return None;
                     }
@@ -403,6 +407,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         matched = true;
                         break;
                     }
+                }
+                if args_expander.limit_hit && !matched {
+                    // Expansion stopped early at the hard limit without finding a match, so we may
+                    // have given up on a call that would otherwise match. Surface it rather than
+                    // silently truncating. (When `matched` is already set we only truncated a
+                    // return-type refinement, which does not change the result, so stay quiet.)
+                    self.error(
+                        errors,
+                        arguments_range,
+                        ErrorKind::OverloadArgumentExpansionLimit,
+                        format!(
+                            "Argument type expansion for this overloaded call hit the limit of {} expansions; some argument type combinations were not checked",
+                            ArgsExpander::<Ans>::GAS
+                        ),
+                    );
                 }
                 (
                     closest_overload,
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -21,7 +21,6 @@ use pyrefly_types::type_output::TypeOutput;
 use pyrefly_types::types::TArgs;
 use pyrefly_util::display::Fmt;
 use pyrefly_util::display::count;
-use pyrefly_util::gas::Gas;
 use pyrefly_util::owner::Owner;
 use pyrefly_util::prelude::SliceExt;
 use pyrefly_util::prelude::VecExt;
@@ -138,9 +137,7 @@ pub struct ArgsExpander<'a, Ans: LookupAnswer> {
     idx: Either<usize, usize>,
     /// Current argument lists.
     arg_lists: Vec<(Vec<CallArg<'a>>, Vec<CallKeyword<'a>>)>,
-    /// Hard-coded limit to how many times we'll expand.
-    gas: Gas,
-    /// Set once `gas` is exhausted, so the caller can report the truncated expansion.
+    /// Set once the expansion budget is exceeded, so the caller can report the truncated expansion.
     limit_hit: bool,
     /// Which positional / keyword arguments are worth expanding. Both default to all-`true`
     /// (expand everything); `call_overloads` narrows them once the candidate overloads are known.
@@ -150,6 +147,7 @@ pub struct ArgsExpander<'a, Ans: LookupAnswer> {
 }
 
 impl<'a, Ans: LookupAnswer> ArgsExpander<'a, Ans> {
+    /// The most argument-type combinations we are willing to check for a single call.
     const GAS: usize = 100;
 
     pub fn new(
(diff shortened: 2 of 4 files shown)


stroxler (approved):

samwgoldman (changes requested):
This PR implements the proposal from the original issue, but the mypy_primer results don't look very good. This should be an error of last resort, not something that users routinely see.

I commented on the original issue with some thoughts. @markselby9, are you interested in investigating the causes of these large overload expansions?

markselby9:
Thanks @samwgoldman — that's the right call, and I dug into the primer hits. The common pattern (e.g. scipy `optimize.minimize`) is a large overload set that differs on only one or two parameters (`method`, …) while the rest — `jac: bool`, `args`, `callback`, … — are **identical across every overload**. The expander was splitting those shared arguments too, so the cartesian product blew past the limit on parameters that can't affect overload selection.

I pushed a second commit that prunes exactly those: **skip expanding an argument whose parameter type is equivalent across all candidate overloads.** It's spec-safe (union assignability distributes, and an identical parameter can't discriminate between overloads, so the result type is unchanged) and conservative — it falls back to the old full-expansion behavior whenever a position's binding isn't cheaply, unambiguously determinable (call-site `*args`/`**kwargs`, variadic/`**kwargs` absorption, `ParamSpec`, differing arities). Full suite is green, and there's a new test where 7 shared `bool` args + 1 discriminating arg would previously exceed the limit but now resolve cleanly.

With this, the expansion limit — and its warning — should only be reached in genuine last-resort cases. Curious to see the re-run mypy_primer diff; if some projects still hit it, those are probably the ones worth a closer look. Let me know if you'd prefer the optimization split into its own PR.

samwgoldman:
Thanks, @markselby9! Looks like there are still a few cases where we hit this in practice. @jorenham mentioned some other possible optimizations in the linked issue. I think the idea of pre-filtering overloads that don't match by call artiy is a good idea. Are you interested in continuing to invest in this PR?

I don't think you need to split the optimization into a separate PR, but I do appreciate that you split this PR into separate commits, so I can review the optimization separately. I think it would be nice to include automated tests (using the expansion limit warning?) that demonstrate the optimizations.

I'm also wondering if we should make the expansion limit configurable.

markselby9:
Pushed an update. Two things up front:

**Soundness fix (disclosure).** The first version of the optimization had a real bug: for *generic* overloads, argument expansion also drives type-variable solving, so skipping an argument could leave a type var unsolved and broaden the result. The mypy_primer run caught it — e.g. jax's `tuple(filter(lambda l: l.filter(), limitations))` collapsed to `tuple[object, ...]` (leaked `_S`), causing spurious `missing-attribute` errors; ibis/spack showed similar widening. I reproduced it locally (differential test vs the prior commit) and fixed it by **excluding generic overloads from the pruning entirely** — non-generic overloads have no such dependency, so pruning there is safe. jax/ibis/spack are back to their previous results.

**Incorporated @jorenham's single-overload point** (thanks!). When only one overload is arity-compatible there's nothing to disambiguate, so expansion is now skipped and the call resolves to a precise argument error instead of exhausting the limit — that should cover the graphql-core / operator cases.

**Tests:** added `test_overload_expansion_skips_non_disambiguating_args` and `test_overload_expansion_skipped_for_single_arity_candidate`, both sized so the limit warning *would* fire without the pruning — demonstrating the optimizations via the warning, as you suggested. Full suite green.

**Configurable limit:** happy to do it. I'd mirror `legacy_overload_expansion` — a `overload-argument-expansion-limit` config option (default 100) threaded to the expander. I can push it as a separate commit on this PR immediately, or leave it out if you'd rather not add the surface — your call.

On @jorenham's other ideas: the cumulative-gas accounting (`2+4+…+2^k`) and member-reachability pruning (assign-to-no-variant) are both complementary to this and would push the limit further out of reach. Happy to take a follow-up, or defer to @jorenham if he's already on it — whatever's least duplicative.

markselby9:
Yes — continuing to invest. Rebased onto `main` and pushed three commits: the warning, the pruning optimization, and a separate gas-accounting fix. Primer is running; I'd treat that as the real verdict rather than anything below.

**The previous version had two real regressions, and they had different causes than I assumed.**

*pandas `dict.get`.* My commit-2 optimization skipped expansion outright when only one overload is arity-compatible, on the reasoning that there's nothing to disambiguate. That reasoning is wrong. With a single candidate, expansion isn't a disambiguation device, it's an **assignability** device: `tuple[str | None, str]` is not assignable to `tuple[str, str] | tuple[None, str]`, but each expanded member is. Repro (`d: dict[tuple[str,str] | tuple[None,str], str]`, `k: tuple[str | None, str]`, `d.get(k)`) gives `str | None` before the optimization and the primer's `bad-argument-type` after. The blanket single-candidate skip is gone.

*ibis `MutableMapping.update`.* This one is more interesting: **the equivalence probe had side effects.** When the receiver carries unsolved placeholder vars (`groups = {}` → `dict[@K, @V]`), those vars appear in the substituted parameter types, and probing overload 1's `SupportsKeysAndGetItem[@K, @V]` against overload 2's `SupportsKeysAndGetItem[str, @V]` **pins `@K := str`**. `find_closest_overload` wraps every speculative call in `snapshot_vars`/`restore_vars`; my pruning analysis didn't. `reveal_type(groups)` went from `dict[str, Value] | dict[Unknown, Unknown]` to `dict[str, Value] | dict[str, Unknown]`. The analysis now runs inside a snapshot, and neutering that snapshot brings the pin straight back.

**The pruning rule is now narrower.** A position is pruned iff *both* every candidate binds it to an equivalent parameter type `p`, **and** the argument is already assignable to `p`. The first half means splitting can't change which overload is selected; the second means splitting can't rescue the call either, since every expansion member — union members, `bool`→literals, enum members, concrete-tuple products, constrained-TypeVar constraints — is a subtype of the original and so stays assignable. Container literals are never pruned, since `CallWithTypes` deliberately leaves them as `Expr` and inferring one in a probe would emit errors and create vars.

**Gas accounting (separate commit).** @jorenham's analysis is correct and it's a real bug — gas was charged per argument-list *produced*, cumulatively across rounds, so `k` binary arguments cost `2+4+…+2^k` and "100 expansions" actually gave up at 32 combinations. Replaced with a prospective per-round product check, so `GAS` now means what its name says. A 6-`bool` call (64 combinations) spuriously warned before and completes cleanly now; 7 `bool`s still warns. Trade-off: worst-case work roughly doubles, since discarded intermediate rounds are no longer billed.

**On arity pre-filtering** — it's already there. Step 1 of `call_overloads` builds `arity_compatible_overloads` via `arity_mismatch_size(..) == 0` before any expansion, and expansion only evaluates against that filtered set. One nuance though: it reduces the cost *per combination*, not the *number* of combinations, because gas is charged per argument-list rather than per (list × overload) — so it can't by itself stop the limit being hit. Its real gap is that `arity_mismatch_size` ignores variadic call-site args, so `*args`/`**kwargs` call sites aren't pre-filtered at all.

**Tests**, as you asked — one demonstrating a call that resolves only because of pruning, plus regression tests for both bugs above and one pinning the gas semantics. Overload tests 121 → 125, full suite 7472/0.

**What I don't know yet:** how many of the eight primer limit-warnings this clears. The gas change buys exactly one extra binary argument of headroom. Reading the remaining sites (strawberry `field`, numpy in pandas/sklearn), they look like large fan-outs of `X | None` keywords into **generic** overload sets — where pruning is disabled wholesale by the tparams guard, because that guard is what fixed the earlier jax/spack breakage. Extending pruning to generic overloads is the next real lever and clearly the hardest one to make sound; I haven't attempted it.

If the warning still fires more than you want after this primer run, I'd rather not keep arguing for it: happy to drop commit 1 and land just the optimization plus the gas fix, or put the warning behind a configurable limit so it's opt-in for people debugging a slow call. Your call on which.

markselby9:
Following up on the ask to optimize the expansion rather than just warn — that's implemented in the two commits after the warning:

- an argument whose parameter type is equivalent across all arity-compatible candidates (and already a subtype of it) is left un-split, since it can't disambiguate them;
- a single arity-compatible overload skips expansion entirely.

Both are conservative: generic overloads are excluded (expansion also drives type-variable solving there), and the speculative equivalence probes are wrapped in `snapshot_vars`/`restore_vars` so they can't pin receiver vars. The gas is charged per-combination rather than cumulatively, which is strictly more permissive, so calls that resolved before still resolve.

Locally, a before/after differential on representative overload calls (non-generic pruning, a generic return-narrowing case, and union expansion) is identical, and the overload suite passes. Ready for another look — the primer will confirm on real code.

jorenham:
What happened?

PR #4310, opened 2026-07-27
fix match on tuple subject not narrowed to Never for assert_never exhaustiveness #4066

### LINKED ISSUE #4066: match on tuple subject not narrowed to Never for assert_never exhaustiveness
### Describe the Bug

```
class RQ: pass
class QQ: pass
class RA: pass
class QA: pass
def f(q: RQ | QQ, a: RA | QA):
    match q, a:
        case RQ(), RA(): return 1
        case RQ(), _: return 2      # wildcard: covers the rest of the RQ row
        case QQ(), QA(): return 3
        case QQ(), _: return 4      # wildcard: covers the rest of the QQ row
        case unreachable:
            assert_never(unreachable)
```

ref https://github.com/zubanls/zuban/issues/487

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4066

Preserved tuple-level narrowing across multi-subject match arms, and correctly collapses exhausted tuple subjects to Never.

Conservatively handles guarded cases.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -2019,6 +2019,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     range,
                     errors,
                 );
+                // A match-arm negation may build on a facet narrow from an earlier arm.
+                // If the accumulated facet is impossible, the whole subject is impossible.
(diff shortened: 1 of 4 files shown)


stroxler (approved):

PR #4313, opened 2026-07-27
[pyrefly] Parse forward references without source padding

samwgoldman:
# Summary

Forward-reference parsing previously prefixed each annotation with spaces up to its source offset, making allocation volume depend on where the annotation appeared in the file. Use Ruff's source-aware annotation parser so ranges remain correct without constructing the prefix. This also delegates escaped and triple-quoted handling to Ruff.

# Test Plan

- `uv run --with jsonschema --with toml python test.py`


grievejia (approved):

PR #4314, opened 2026-07-27
[pyrefly] Avoid duplicate default interpreter query

samwgoldman:
# Summary

Fresh CLI startup discovered the system interpreter with a `--version` subprocess, then immediately spawned the same executable again for the authoritative environment query. Validate candidates with the environment query directly and reuse its cached result.

This reduced an empty auto-configured check from 49.8 ms to 35.3 ms on macOS.

# Test Plan

- `uv run --with jsonschema --with toml python test.py`


grievejia (approved):

PR #4320, opened 2026-07-28
fix False positive for torch.register_* #3988

### LINKED ISSUE #3988: False positive for `torch.register_*`
### Describe the Bug

Hi, I'm not sure if this is a false positive or if this is even supposed to be handled by static type checking, but I'm wondering how to handle the following situation:

```python
import torch
from torch import nn


class LeModule(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.register_buffer("le_tensor", torch.tensor([0, 1, 2]))


m = LeModule()
if m.le_tensor is not None:
    result = m.le_tensor.tolist()  # pyrefly: Expected a callable, got Tensor
```

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3988

Recognizes `self.register_buffer(...)` and `self.register_parameter(...)` in `__init__/__post_init__`, synthesizes correctly typed instance attributes for `torch.nn.Module` subclasses.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -503,6 +503,17 @@ impl ClassField {
         Self::new_synthesized_inner(ty, false)
     }
 
+    pub fn new_synthesized_instance_attribute(ty: Type) -> Self {
+        ClassField(
+            ClassFieldInner::InstanceAttribute {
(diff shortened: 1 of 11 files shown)


stroxler (approved):

PR #4322, opened 2026-07-28
fix(stubgen): preserve RHS of explicit `TypeAlias` assignments

### LINKED ISSUE #4319: bug(stubgen): explicit `TypeAlias` assignments lose their right-hand sides in emitted stubs
### Describe the Bug

The typing specification defines an explicit type alias using `TypeAlias` as an assignment such as `Url: TypeAlias = str`.

When `pyrefly stubgen` processes such an assignment, it preserves the `: TypeAlias` annotation but drops the right-hand side. The generated stub therefore no longer defines what the alias represents.

This is different from eliding an ordinary runtime value: the right-hand side of a type alias is typing information and needs to be retained.

Reproduced with Pyrefly 1.1.1 and 1.2.0-dev.3.

#### Input

```python
from typing import TypeAlias

TemplateId: TypeAlias = str
TemplateLookup: TypeAlias = dict[TemplateId, str]
```

#### Expected output

```python
from typing import TypeAlias

TemplateId: TypeAlias = str
TemplateLookup: TypeAlias = dict[TemplateId, str]
```

#### Actual output

```python
from typing import TypeAlias

TemplateId: TypeAlias
TemplateLookup: TypeAlias
```

Both aliases have lost their definitions.

### Real-world example

NumPy 2.3.5 defines `NDArray` in [`numpy/_typing/_array_like.py`](https://github.com/numpy/numpy/blob/v2.3.5/numpy/_typing/_array_like.py#L24):

```python
NDArray: TypeAlias = np.ndarray[_AnyShape, dtype[_ScalarT]]
```

Stubgen emits:

```python
NDArray: TypeAlias
```

The same happens to the other explicit aliases in that module, including `ArrayLike`, causing their type information to be lost for downstream consumers of the generated stubs.

tobyh-canva:
## Summary
Resolves #4319 by special-casing `TypeAlias` annotations when extracting annotated assignments.

## Testing
- Adds regression tests for both the pre-PEP 695 `TypeAlias` annotations, and the post-PEP 695 `type` statements.

### DIFF
--- a/pyrefly/lib/stubgen.rs
+++ b/pyrefly/lib/stubgen.rs
@@ -267,6 +267,11 @@ class BaseModel:
     }
 
     #[test]
+    fn test_stubgen_type_alias_explicit() {
+        assert_stubgen_snapshot("type_alias_explicit");
+    }
+
+    #[test]
     fn test_stubgen_generics() {
         assert_stubgen_snapshot("generics");
     }
--- a/pyrefly/lib/stubgen/extract.rs
+++ b/pyrefly/lib/stubgen/extract.rs
@@ -44,6 +44,7 @@ use starlark_map::Hashed;
 
 use crate::alt::answers::Answers;
 use crate::alt::types::decorated_function::DecoratedFunction;
+use crate::binding::binding::Binding;
 use crate::binding::binding::Key;
 use crate::binding::binding::KeyClassField;
 use crate::binding::binding::KeyClassMetadata;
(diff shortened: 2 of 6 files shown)


NathanTempest (approved):
LGTM for shipping. Nice approach using the binding classification (Binding::TypeAlias) rather than string-matching the annotation which correctly handles TypeAlias imported under an alias too. Verified the annotated path (bind_single_name_assign → is_definitely_type_alias → Binding::TypeAlias under Key::Definition), so the detection keys off the right binding. Test coverage is thorough, including the negative case for complex non-alias values.

yangdanny97 (approved):

PR #4326, opened 2026-07-28
fix no error on possibly undefined attribute #4318

### LINKED ISSUE #4318: no error on possibly undefined attribute
> not just platform checks, any conditionally defined attributes:
> 
> ```py
> def condition() -> bool:
>     return False
> 
> class A:
>     if condition():
>         c: int = 0
> 
> A.c  # no error, crashes at runtime
> ```
> 
> ty's [`possibly-missing-attribute`](https://docs.astral.sh/ty/reference/rules/#possibly-missing-attribute) rule catches this 

 _Originally posted by @DetachHead in [#2558](https://github.com/facebook/pyrefly/issues/2558#issuecomment-4677703877)_

opening a new issue because the example here still doesn't report an error on 1.2.0-dev.3

asukaminato0721:
# Summary

Fixes #4318

Tracks conditionally initialized class and module attributes, reports missing-attribute for class, instance, and multi-platform module access.

Preserves valid attributes defined in every if/else branch, handles incremental export invalidation.

# Test Plan

add && update tests

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -2654,6 +2654,31 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                         .emit();
                 }
             }
+            BindingExpect::UninitializedAttributeCheck {
+                name,
+                range,
+                getattr_class,
+                termination_keys,
+            } => {
+                let may_be_uninitialized = termination_keys
+                    .as_ref()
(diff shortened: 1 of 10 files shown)


yangdanny97:
hmm, for this one I think we should error where the attribute is declared, not where it is accessed. it could be kind of tricky because we would need to handle situations like this to avoid false positives.

```
if something:
   self.x = 1
else:
   self.x = 2
```

I guess it would be similar to the logic around possibly-unbound variables

Issue #1063, opened 2025-09-09
Sandbox: link shortening


yangdanny97:
### Describe the Bug

Encoding everything in base64 (especially using LZString) can lead to some pretty long URLs. Ty's sandbox uses a link shortener and Mypy's sandbox creates a public gist

hashiranhar:
Can I work on this?

yangdanny97:
Thanks, let's discuss the design before any implementation happens though.

Do you have any ideas on how we should do this?

hashiranhar:
> Thanks, let's discuss the design before any implementation happens though.
> 
> Do you have any ideas on how we should do this?

Github Gists makes a lot of sense, a gist supports multiple files so works for our sandbox. In terms of how the gists are created, it seems like mypy has a dummy playground account that logs in the gists on behalf of the users, I think this is the way to go. I've worked with the github api before this should be straight forward to setup hopefully!


Side Note: I tried to checkout fflate deflate and gzip methods on small files ( 800 chars ) it seems to shortern it by 50-100 chars, but its  more efficient the larger the file gets. ( 9000 chars vs 19000 chars for a 900 line file from one of my projects). Still not ideal tho

yangdanny97:
Thanks for investigating the other compression options. 

I think if we do go with gists we should make the automatically created gists secret instead of public.

Let's see what the folks in the dev channel on discord say.

asukaminato0721:
If we need shorter link in the future, this can be reopen.

PR #4327, opened 2026-07-28
improve Sandbox: link shortening #1063

asukaminato0721:
# Summary

improve #1063

just an experiment.

replaces LZString encoding for newly generated sandbox URLs with raw DEFLATE, URL-safe Base64, and a small preset dictionary containing common Python syntax, typing constructs, sandbox filenames, and configuration patterns.

# Test Plan

did some test locally, seems can compress more.


stroxler (approved):

stroxler:
It looks good to me, using a library with dictionary support seems like a pretty good way to get a win in the usual case.

asukaminato0721:
I get the idea from https://en.wikipedia.org/wiki/Minimum_description_length :)

PR #4328, opened 2026-07-28
Run hinted call pass to recover shared union-branch type variables

### LINKED ISSUE #3583: generic inference fails in some cases with function argument type like `Foo[T, object] | Foo[object, T]`
### Describe the Bug

```python
class Foo[T, U]:
    def t(self) -> T: ...
    def u(self) -> U: ...


def fn[T](foo: Foo[T, object] | Foo[object, T], value: list[T]) -> T: ...


def fn2[T](value: list[T]) -> T:
    return fn(Foo[T, object](), value)
```

```
ERROR sandbox.py:10:12-39: Returned type `object` is not assignable to declared return type `T` [bad-return]
ERROR sandbox.py:10:33-38: Argument `set[T]` is not assignable to parameter `value` with type `set[object]` in function `fn` [bad-argument-type]
```

Python: 3.13

- this works correctly in [pyright](https://basedpyright.com/?typeCheckingMode=all&reportUnusedParameter=false&reportUnusedVariable=false&code=MYGwhgzhAEBiD28DaAVANNAqgXQFwFgAoaE6AEwFMAzaAFwAoIKQqBKaAWgD5oVdoAdEKKly1aAFdGzNpx6Z%2BQgURWFKNKgDtU2elUT8EydNHgAjAFYVgtbNAA%2BcREnNWbGFNgwA3MCAkU-Ey0OuzcvIrChKrq0FoATDr0vv6B0MGhchEipABOFLQSuZpxmvRGqBiu1rb0rD5%2BAaxAA).
- [ty has the same issue](https://play.ty.dev/a101db20-52d9-4aec-96c3-c6c9a3cd518e) (see https://github.com/astral-sh/ty/issues/3533) but unlike ty, pyrefly works correctly when the second argument type is just `T` instead of `list[T]`

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuA2gCoA09AqgLqIAOunqj6mGGHoAXABRwYUMAEp6AWgB89donqF9wseMn0ArvMUr1Wnrv2FhjrCbDoOfWWBa7mbLvVxsACsYSmk%2BegAfJhZWQJCw7nY%2BbgA3VChTGF0oCDhpd1VNbTsDdCcJKVcAJndZdMzs%2Blz8wusSwzEAJxhpUy6RV1lfDm540PDZZTSMrOUQThAyHrAoUkJpXABbKAoyaQALXHQ1VJguuAhj%2BgBeekEQAGZCAEZqh%2BFWc67cC75hDDoXDSVDSK4nLZ5LagygHW70ADk31%2BCOEIAAvotUGEIGdGNAYBQ0Fg8EQyBigA

_No response_

ak4-sh:
When a generic function's parameter is a union in which the same function type variable appears in multiple branches (Foo[T, object] | Foo[object, T]), the call should bind T to the branch matched by the argument. Pyrefly instead collapsed T to the widest branch (object), producing spurious errors even when a return hint was available to steer solving to the correct branch.

`call_infer_inner` only ran its contextual (hinted) inference pass when the no-hint pass already succeeded. But when the parameter union shares a type variable across branches, the no-hint pass hits hard errors — greedy union-branch matching collapses the variable to the widest branch — so the return hint was never tried. 

Run the hinted pass on that recovery path too, but only when the hint is "ground": it carries no free placeholder vars from an enclosing call. A speculative hinted pass mutates shared solver vars, so gating on a ground hint keeps recovery from polluting an enclosing call's
variables through the hint. The existing selection guard still adopts the hinted result only when it is at least as good as the no-hint result.

# Summary

Fixes #3583

# Test Plan

Added ```test_typevar_in_multiple_union_branches_with_hint``` (the fix — now reports
no errors) and ```test_typevar_in_multiple_union_branches_no_hint``` (bug-marked,
documenting the remaining no-hint case) in pyrefly/lib/test/generic_basic.rs.

Verified the new test fails on main with the two spurious errors and passes
with the fix. test_complicated_paramspec_forwarding (which an earlier
broader relaxation regressed) stays green, confirming the ground-hint gate.

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1801,11 +1801,25 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             None,
             ctor_targs_no_hint.as_mut(),
         );
-        // If the call succeeds, attempt contextual typing with the hint.
+        // Attempt contextual typing with the hint. When the no-hint pass already succeeds we
+        // always try the hint (this is ordinary contextual typing). When the no-hint pass
+        // produced hard errors, the hint can still steer type variable solving toward a valid
+        // solution the no-hint pass missed (e.g. when a parameter is a union in which the same
+        // type variable appears in multiple branches). But a speculative hinted pass mutates
+        // shared solver vars, so on that recovery path we only take the risk when the hint is
+        // "ground" — it has no free placeholder vars belonging to an enclosing call that the
+        // pass could pollute. The hinted result is only adopted when it is at least as good as
+        // the no-hint result (see below).
+        let run_hinted = hint.is_some_and(|h| {
+            !call_errors_no_hint.has_hard()
+                || h.types()
+                    .iter()
+                    .all(|t| t.collect_maybe_placeholder_vars().is_empty())
+        });
+        // `retry_input` is `Some` exactly when `hint` is, and `run_hinted` implies `hint.is_some()`,
+        // so this destructuring never fails — it only recovers the clones made above.
         let (chosen_ctor_targs, chosen_call_errors, chosen_arg_errors, chosen_res) =
-            if !call_errors_no_hint.has_hard()
-                && let Some((callable, self_obj)) = retry_input
-            {
+            if run_hinted && let Some((callable, self_obj)) = retry_input {
                 let mut ctor_targs_with_hint = ctor_targs.as_ref().map(|x| (**x).clone());
                 let arg_errors_with_hint = self.error_collector();
                 let call_errors_with_hint = self.error_collector();
--- a/pyrefly/lib/test/generic_basic.rs
+++ b/pyrefly/lib/test/generic_basic.rs
@@ -866,3 +866,34 @@ def test2(cls: T2) -> str:
     return cls.name  # E:
     "#,
 );
+
+// A function type variable appearing in multiple branches of a union parameter must
+// solve to the branch matched by the argument, not collapse to the widest branch. Here
+// the return hint `T` (from `fn2`) steers the call so `T` in `fn` binds to `fn2`'s `T`.
+testcase!(
+    test_typevar_in_multiple_union_branches_with_hint,
+    r#"
+class Foo[T, U]:
+    def t(self) -> T: ...
+    def u(self) -> U: ...
+def fn[T](foo: Foo[T, object] | Foo[object, T], value: list[T]) -> T: ...
+def fn2[T](value: list[T]) -> T:
+    return fn(Foo[T, object](), value)
+    "#,
+);
+
+// Without a return hint the greedy union-branch matching commits `T` to the widest
+// branch (`object`) instead of the branch matched by the argument.
+testcase!(
+    bug = "T should reveal as T, not object (greedy union-branch matching)",
+    test_typevar_in_multiple_union_branches_no_hint,
+    r#"
+from typing import reveal_type
+class Foo[T, U]:
+    def t(self) -> T: ...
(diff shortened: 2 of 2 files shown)


stroxler:
This is a pretty nice win on the examples given!

I'd like a review from @samwgoldman or @rchen152 though: I'm fairly sure this approach makes Pyrefly strictly *better*, but the underlying problem seems to be that our solver can't actually do the right thing.

That leads to three questions
- is my diagnosis of the root cause (below) correct?
- do we want to merge this PR, which I think strictly helps but doesn't solve the root cause?
- do we want to file an issue to actually fix Pyrefly, or do we prefer simplicity to correctness?

---------

I think the actual problem is that it is not possible to correctly solve all function calls without disjunctive
constraints. I'm pretty sure this is why the PR is demonstrating that the logical invariant "a call that
fails with no hint fails with a hint" is false - that logical invariant is *true* for a correct solver, but false
for our solver.

To explain, let me first rephrase and extend the new unit test to avoid using `T` repeatedly:
```
class Foo[T, U]:
    def t(self) -> T: ...
    def u(self) -> U: ...
def g[T](foo: Foo[T, object] | Foo[object, T], value: list[T]) -> T: ...
def fn2[Q](value: list[Q]) -> T:
    return g(Foo[Q, object](), value)
def fn3[Q](value: list[Q]) -> None:
    x = g(Foo[Q, object](), value)
```
In the calls to `g`, we have `Q` (which is Quantified during the solve) and `T` (which is a Var) in play.

Now, when we match `Foo[Q, object]` against `Foo[T, object] | Foo[object, T]`, the logically correct constraint on
`T` is `T <: object || T <: Q`, because either of those bounds will cause the check to succeed. As far as I know,
*only* a union upper bound with a type var in both branches needs this sort of representation, and I think we
just haven't hit such examples enough so far to implement it.

Because we don't have disjunction, the "greedy union-branch matching" we do instead approximates that
disjunction with `T <: Q | object`, which is a very different constraint (especially since `object` is Top,
so that collapses to `T <: object` by basic type lattice properties - which is clearly *very* different from the
disjunction.

Then, the check against `value` will fail since list is invariant. In `fn3` this happens with or without the PR,
because the root cause is that we don't have disjunction and there's no hint to help.

The PR fixes the issue for `fn2`, because if we try the return-type hint *before* we run an incorrect algorithm
checking the `foo` arg, we're able to just pick the right branch of the disjunction (`T <: Q`) and never
incorrectly flatten and collapse. This is what I mean when I say the PR strictly helps Pyrefly, but doesn't
actually get at the root cause as I understand the problem.

rchen152:
This seems fairly harmless to merge, but I agree that it doesn't seem to address the underlying root cause, so the question I'd ask is whether this fixes toy examples only or has a positive effect on real-world code. Given that the mypy_primer error delta is neutral, I'm a little worried that it only does the former, in which case papering over problems with our solver isn't necessarily beneficial.

ak4-sh:
Fair enough! I wasn't able to find any real world example that has a similar pattern where this update would be useful. Since the primer showed no changes as well, I don't know if the change would actually be helpful in any real world codebase. Should I close the PR? 

@rchen152 @stroxler

rchen152:
Yeah, I'd lean toward closing this PR. Sorry to not have a more fruitful suggestion to make here. We generally try to be pretty practical and accept patches that work around solver limitations, but there is a cost to maintaining those patches, and my feeling is that this one doesn't gain us enough to make that tradeoff.

PR #4329, opened 2026-07-28
Preserve narrowing across neutral boolean operands

### LINKED ISSUE #4317: redundant condition incorrectly widens type
### Describe the Bug

```py


def fn(value: type[int] | str):
    reveal_type(None if isinstance(value, type) else value) # str | None (correct)
    reveal_type(None if isinstance(value, type) and True else value)  # str | type[int] | None (incorrect)
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHUoz0chUqwwwdMOgAUQsQFcYiRsxgBtVgwC6dAD504DSgEpEKuu4HDREqTC0A5XHQYTnUIOFYHDABjP10oAwAaY2knOhgoOBD4gzSAYntHWzpA4LotaN5BaIYnNw9BETFJEwCgkIgwiPQo9FidfRhk3zSMTDoAFUoDdMzswbS6AodKYt9zdCti0pCtVkrKatqQRJAyQTAoUkIGWigKMgYACyCAWhFKCKC6AF46GRAAGZCABGAAsAIUxEEWX4fwBYigAJAAF9TqgahARAAxaAwChoLB4IhkVFAA

_No response_

ak4-sh:
Boolean literals that are neutral to an expression should not introduce placeholder narrowing operations. 

Skip `True` in `and` expressions and `False` in `or` expressions so positive and negative narrowing remain equivalent to the original condition.

# Summary

Fixes #4317

# Test Plan
  Added a regression test covering:

  - `A and True`
  - `True and A`
  - `A or False`
  - `False or A`

  Ran:

  ```sh
  cargo test -p pyrefly --lib test_ternary_isinstance_with_neutral_boolean
```

### DIFF
--- a/pyrefly/lib/binding/narrow.rs
+++ b/pyrefly/lib/binding/narrow.rs
@@ -19,6 +19,7 @@ use ruff_python_ast::BoolOp;
 use ruff_python_ast::CmpOp;
 use ruff_python_ast::Expr;
 use ruff_python_ast::ExprBoolOp;
+use ruff_python_ast::ExprBooleanLiteral;
 use ruff_python_ast::ExprCall;
 use ruff_python_ast::ExprCompare;
 use ruff_python_ast::ExprNamed;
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

NathanTempest:
Hey @ak4-sh thank you for this PR, we have already imported it and it looks ready to go save for a small change (using assert_type(..., str | None) for these tests) which I will do internally. Thank you for your continued efforts in improving Pyrefly!

PR #4331, opened 2026-07-28
Fix collapse of multiple legacy type params hosted on the same module

### LINKED ISSUE #4315: invalid-param-spec false positive: "Unexpected ParamSpec type: ParamSpec[P]" on a ParamSpec-generic Protocol attribute
### Environment
  - pyrefly 0.58.0
  - Python 3.12
  - modal 1.3.0

### Repro
First do `pip install modal`, then type check the following code:
```
import modal

app = modal.App()

@app.function()
def my_task(name: str, count: int) -> bytes:
    return b""

result = my_task.remote("hello", 3)
```

Expected: no error — the call matches the wrapped function's signature. mypy and pyright both report zero errors on this file.

Actual: pyrefly reports, on the .remote(...) call:
```
ERROR Unexpected ParamSpec type: `ParamSpec[P]` [invalid-param-spec]
```

Note on reproducing: this isn't reproducible in the online sandbox — it needs modal installed. I tried to reduce it to a modal-free single file (hand-writing the equivalent ParamSpec-generic Protocol-attribute pattern, including as a .pyi stub); pyrefly accepts all of those, so the trigger appears specific to modal's generated stubs rather than the general pattern.

### Analysis
modal exposes callables as class attributes typed by a Protocol that is generic over a ParamSpec, subscripted with a ParamSpec type argument. Simplified from its generated stubs:

```
class Function(Generic[P, ReturnType, OriginalReturnType]):
    class __remote_spec(Protocol[P_INNER, ReturnType_INNER]):
        def __call__(self, /, *args: P_INNER.args, **kwargs: P_INNER.kwargs) -> ReturnType_INNER: ...
    remote: __remote_spec[P, ReturnType]
```

The false positive fires on every .remote()/.spawn()/.map()-style call site, identically and independent of the arguments passed.

jcarreiro:
## Summary

Fixes #4315.

When a generic class or function uses several legacy (pre-PEP-695) type variables imported from the same module (e.g. `foo.T` and `foo.P`), all of them collapse onto a single base-name scope entry (`foo`): `LegacyTParamId::as_identifier` returns the module base, and `add_possible_legacy_tparam` upserts by that base name, so only the last-declared tparam keeps its module-attribute narrow. References to the earlier ones fall through to the raw module-level declaration.

- For a `TypeVar` this is a silent correctness bug (a raw `TypeVar` is tolerated in a type-argument position).
- For a `ParamSpec` it surfaces as a spurious `Unexpected ParamSpec type` error, since a raw `ParamSpec` in a ParamSpec position is rejected.

This is exactly the shape of [modal](https://modal.com)'s generated stubs — `Function[P, ReturnType, OriginalReturnType]` with `P` declared first — so every `.remote()`/`.spawn()` call site gets flagged. (mypy and pyright both accept the same code.)

There is already a `bug = "…"` testcase documenting this (`test_multiple_possible_legacy_tparams`), whose comment reads *"the `foo.C` possible-legacy-tparam binding is the one that winds up in scope, we lose track of the `foo.T` one."*

## Fix

Give the surviving base-name scope entry a binding that carries **every** hosted tparam's key, and narrow the module at **all** of their attribute facets rather than only the last, so a reference to any hosted tparam resolves to its `Quantified`.

- `binding/binding.rs`: `Binding::PossibleLegacyTParam` payload `Idx<KeyLegacyTypeParam>` → `Box<[Idx<KeyLegacyTypeParam>]>`.
- `binding/bindings.rs`: `add_name_definitions` groups collected tparams by base identifier and, for a base hosting more than one, gives the surviving entry a binding with all their keys.
- `alt/solve.rs`: `binding_to_type_info_possible_legacy_tparam` iterates the keys, threading the accumulating module `TypeInfo` so it is narrowed at every hosted facet.
- Mechanical match-site updates in `alt/jaxtyping.rs` and `state/ide.rs`.

## Testing

- Updated `test_multiple_possible_legacy_tparams` to assert the corrected behavior (dropped the `bug =` annotation and the expected errors).
- Added `test_module_hosted_paramspec_tparam`, a regression test that mirrors the module-hosted-stub layout and confirms `.call(...)` arguments are still type-checked (i.e. the fix resolves the tparam rather than silencing the error).
- `cargo test --lib` is green.

### DIFF
--- a/pyrefly/lib/alt/jaxtyping.rs
+++ b/pyrefly/lib/alt/jaxtyping.rs
@@ -189,7 +189,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             Binding::Import(import) => {
                 import.module.as_str() == "jaxtyping" && import.name.as_str() == "jaxtyping"
             }
-            Binding::PossibleLegacyTParam(key, _) => self.legacy_tparam_is_jaxtyping_module(*key),
+            Binding::PossibleLegacyTParam(keys, _) => keys
+                .iter()
+                .any(|key| self.legacy_tparam_is_jaxtyping_module(*key)),
             _ => false,
         }
     }
@@ -202,9 +204,9 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             | Binding::ForwardToFirstUse(idx) => {
                 self.binding_is_jaxtyping_wrapper_origin(self.binding_following_forwards(*idx))
             }
-            Binding::PossibleLegacyTParam(key, _) => {
-                self.legacy_tparam_is_jaxtyping_wrapper_origin(*key)
-            }
+            Binding::PossibleLegacyTParam(keys, _) => keys
+                .iter()
+                .any(|key| self.legacy_tparam_is_jaxtyping_wrapper_origin(*key)),
             _ => false,
         }
     }
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -5212,45 +5212,63 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     #[inline(never)]
     fn binding_to_type_info_possible_legacy_tparam(
         &self,
-        key: Idx<KeyLegacyTypeParam>,
+        keys: &[Idx<KeyLegacyTypeParam>],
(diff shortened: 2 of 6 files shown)


rchen152 (approved):
Thank you for the PR! This looks good to me. I've imported it to get an internal review before merging.

grievejia (approved):

PR #4335, opened 2026-07-29
feat(lsp): quick-fix code action for removing unused ignores

tobyh-canva:
# Summary
Follow-up to #4200.

- Adds an editor quick-fix code action to remove an unused `# pyrefly: ignore` / `# type: ignore` comment (mentioned in #2207).
- Changes the highlighted diagnostic range for unused-ignore / unused-type-ignore errors to cover the full suppression comment, not just the start of the line where the comment appears. This ensures the underline in the IDE is in the correct place.
- Refactors the implementation of `pyrefly suppress --remove-unused` to be simpler and rely on richer information coming from check diagnostics. This was discussed in https://github.com/facebook/pyrefly/pull/4200#discussion_r3607881358.

# Test Plan
- [x] Verified code action appears and works in LSP with VSCode and Cursor

https://github.com/user-attachments/assets/a2c56965-1493-4820-89d3-e21f8de57fef

### DIFF
--- a/pyrefly/lib/commands/suppress.rs
+++ b/pyrefly/lib/commands/suppress.rs
@@ -33,6 +33,7 @@ pub struct SuppressArgs {
 
     /// Path to a JSON file containing errors to suppress.
     /// The JSON should be an array of objects with "path", "line", "name", and "message" fields.
+    /// Unused suppression diagnostics must also include a structured "suppression_edit".
     #[arg(long)]
     json: Option<PathBuf>,
 
@@ -106,7 +107,7 @@ impl SuppressArgs {
             };
 
             // Remove unused ignores (JSON path only)
-            suppress::remove_unused_ignores_from_serialized(unused_errors, kind);
+            suppress::remove_unused_ignores_from_serialized(unused_errors, kind)?;
         } else {
             // Add suppressions mode (existing behavior)
             let serialized_errors: Vec<SerializedError> = if let Some(json_path) = &self.json {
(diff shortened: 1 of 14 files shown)


pyrefly/lib/commands/suppress.rs
```diff
@@ -32,7 +32,10 @@ pub struct SuppressArgs {
     config_override: ConfigOverrideArgs,
 
     /// Path to a JSON file containing errors to suppress.
-    /// The JSON should be an array of objects with "path", "line", "name", and "message" fields.
+    /// The JSON should be an array of objects with "path", "line", and "name" fields.
+    /// When combined with `--remove-unused`, unused suppression diagnostics must also include a
+    /// structured "suppression_edit" field with "tool", "start", "end", "expected", and
+    /// "replacement" sub-fields.
     #[arg(long)]
     json: Option<PathBuf>,
```
tobyh-canva, line 40: A `message` field is no longer necessary since it was only being used for unused ignore removal, but that now uses `suppression_edit`.

**Note:** I'm not totally sure what's meant to be producing the JSON document described here. I can't see anything in Pyrefly itself which serializes this schema. As such, I've just updated the option description to be effectively in line with what existed before, but let me know if this is not desired. Perhaps it would be preferable to make `--json` mutually exclusive with `--remove-unused`.

stroxler:
Hi @tobyh-canva - could you rebase this commit? I think it has merge conflicts with trunk.

It looks like a good change to me from a skim!

tobyh-canva:
@stroxler Thanks, rebased :)

PR #4352, opened 2026-07-29
fix behavior of init=False for pydantic.BaseModel

arogozhnikov:
# Summary

`Field(init=False)` is ignored for `pydantic.BaseModel`, see this discussion: https://github.com/pydantic/pydantic/discussions/9341#discussioncomment-17052595

Currently pyrefly assumes the field is not part of `__init__` signature. That's a subtle thing but I hit this a lot in my codebase.

Found no relevant issue in github tracker.

# Test Plan

This PR adds a test case. Compiled locally, tested on a small test file to confirm behavior was fixed.

### DIFF
--- a/pyrefly/lib/alt/class/dataclass.rs
+++ b/pyrefly/lib/alt/class/dataclass.rs
@@ -1232,11 +1232,22 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         } else {
             self.class_self_param(cls, false)
         };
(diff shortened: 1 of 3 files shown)


yangdanny97 (approved):

arogozhnikov:
@yangdanny97  please reimport, just checked locally tests passed successfully:

```
cargo test -p pyrefly --lib 
```

kinto0 (approved):

arogozhnikov:
still says "PR has not been imported"

PR #4353, opened 2026-07-29
Hold LSP requests until the initial `workspace/configuration` response

### LINKED ISSUE #4332: `disableLanguageServices` is ignored until the initial `workspace/configuration` response arrives
`disable_language_services` defaults to `false`, and is only set once the client answers the `workspace/configuration` request from `Server::new`:

https://github.com/facebook/pyrefly/blob/e43ca9967d49ecacdcad657ec2ed9fd48f4aafd1/pyrefly/lib/lsp/non_wasm/server.rs#L2783

Until then this check passes:

https://github.com/facebook/pyrefly/blob/e43ca9967d49ecacdcad657ec2ed9fd48f4aafd1/pyrefly/lib/lsp/non_wasm/server.rs#L4285-L4288

In my `numba` workspace (in VSCode), that was the first ~40 requests, including `documentSymbol`, `codeAction`, `codeLens`, `inlayHint` and `foldingRange`:

```
 INFO Handling non-canceled request textDocument/documentSymbol (43)
 INFO Client configuration applied to workspace: Some(Url { ... path: "/home/joren/Workspace/numba" ... })
 INFO Handling non-canceled request textDocument/codeAction (46)
 INFO Skipping request - language services disabled
```

VSCode caches document symbols per document version, so the outline from that window will stick around until the file is edited.

The reply is late because the `workspace/configuration` extension middleware awaits `getInterpreterPath`, which blocks while ms-python is activating:

https://github.com/facebook/pyrefly/blob/e43ca9967d49ecacdcad657ec2ed9fd48f4aafd1/lsp/src/extension.ts#L176-L190

The middleware doesn't need to wait for the interpreter. `onDidChangeInterpreter` already sends a `didChangeConfiguration`, so it could return the config as-is. The path then comes in on the next request.

That only narrows the window though; `awaiting_initial_workspace_config` would have to be checked in `make_handle_if_enabled` as well, not just before indexing.

jorenham:
# Summary

Requests received while `awaiting_initial_workspace_config` is set are now held and re-enqueued once the response lands. Checking that flag in `make_handle_if_enabled` instead, as the issue suggests, would return empty responses during startup for everyone, not just those who disabled language services. Releasing is keyed on the request method rather than on a successful parse, so a client that *fails* the configuration request stops us waiting too.

The middleware no longer waits for the Python extension. It uses a new non-blocking `getInterpreterPathIfResolved`, and `activate` re-pulls the configuration once the interpreter resolves, since `onDidChangeInterpreter` only fires on a later change.

Fixes #4332

# Test Plan

Regression tests added, for both a successful and a failed configuration response.

### DIFF
--- a/crates/pyrefly_lsp_test/src/object_model.rs
+++ b/crates/pyrefly_lsp_test/src/object_model.rs
@@ -247,6 +247,18 @@ impl<'a, R: lsp_types::request::Request> ServerRequestHandle<'a, R> {
             error: None,
         }));
     }
+
+    pub fn send_response_error(self, code: i32, message: &str) {
+        self.client.send_message(Message::Response(Response {
+            id: self.id,
+            result: None,
+            error: Some(ResponseError {
+                code,
+                message: message.to_owned(),
+                data: None,
+            }),
+        }));
+    }
 }
 
 impl<'a> ServerRequestHandle<'a, WorkspaceConfiguration> {
--- a/crates/pyrefly_lsp_test/src/object_model.rs
+++ b/crates/pyrefly_lsp_test/src/object_model.rs
@@ -239,6 +239,18 @@ impl<'a, R: lsp_types::request::Request> ServerRequestHandle<'a, R> {
     pub fn send_response(self, result: Value) {
         self.client.send_response::<R>(self.id, result)
     }
+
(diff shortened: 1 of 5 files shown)


jorenham:
The benchmark stuff seems like an unrelated fluke

PR #4354, opened 2026-07-29
Detect cyclic type aliases that are reachable but not self-referential

### LINKED ISSUE #4324: Stack overflow with recursive type definitions
### Describe the Bug

```python
type T1 = T2
type T2 = T2
x.__str__
x: T1
```

```
thread '<unknown>' (423590) has overflowed its stack
```

### Version info
Tested with commit bd57aa791

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIALqcTAAQAqAjHQLyMBMAOutbZ2048iAfRFxKAJzHDEjJiAA0IMpJhgopQpVwBbKBQDEdAAqk1G0nTRY8%2BOgGNc6SAHMArpNSUIzwj2MAZRh6AAtKSmI4RAB6GNV1TUJcSVcYmHQYzFwHOBinFwgPLx9nGLowFLpUADdUaFRsWEdnN09vX3Q6XGJS9Dh-dDJKUOcAWhqYSThOwS4QAGZCJm4QHgBtKckUuABdHnd0CF1iFMoYTDHMCDUHH0nBAHIi9BSYR55XyjG1AEd3G4XMYAaxgpDGqAcDngcCeAHdUJJ0B90CAAL7KSH3GAAMWgMAoNhwBBI5DRQA

_No response_

MannXo:
Fixes #4324

Checking this aborted with a stack overflow:

```python
type T1 = T2
type T2 = T2
x.__str__
x: T1
```

### What the shape of the bug is

`__str__` is not special and the ordering does not matter. What matters is the alias topology, and both the annotation and an attribute access are needed:

| Source | Before | After |
|---|---|---|
| `type T2 = T2`, use `T2` | ok | ok |
| `type T1 = T1`, use `T1` | ok | ok |
| `type T1 = T2` / `type T2 = T1` (mutual) | ok | ok |
| `type T1 = T2` / `type T2 = T2`, use `T1` | **stack overflow** | diagnosed |
| `type T1 = T2` / `type T2 = T3` / `type T3 = T3`, use `T1` | **stack overflow** | diagnosed |

So a self-referential alias reached through one or more hops crashed, while a direct self-loop or a pure mutual cycle was already handled.

### Root cause

Step 2 of `wrap_type_alias` already replaces a cyclic alias body with an error type, and its comment says this exists "to prevent infinite recursion when downstream operations (e.g. attribute lookup, subset checks) try to resolve the alias". The detection was incomplete rather than missing.

`type_alias_has_cyclic_reference` decided self-reference by name:

```rust
let is_self_ref = |ty: &Type| matches!(ty, Type::UntypedAlias(ta) if ta.name() == name);
```

That catches a cycle the alias *belongs to*, mutual cycles included, because `expand_type_alias_refs` inlines the referenced body first and so reintroduces the alias's own name. It does not catch a cycle that is merely *reachable*. Resolving `T1` only ever sees the name `T2`, never `T1`, so `T1` was not flagged and kept a live `UntypedAlias(Ref(T2))` edge into `T2`'s self-loop. Attribute lookup then followed it without bound:

```rust
Type::TypeAlias(ta) => {
    self.as_attribute_base1(self.get_type_alias(&ta).as_value(self.stdlib), acc)
}
```

`lldb` puts frame #0 at `AnswersSolver::get_idx::<KeyTypeAlias>`, and the three-hop row above confirms indirection depth is irrelevant.

### The fix

Expansion already knows where the cycles are, because it leaves a reference in place exactly when it would otherwise revisit an alias. It now returns the names it left behind, and those count as self-references for the cyclic check alongside the alias being resolved. `T1` is therefore diagnosed and reduced to an error type just as `T2` is.

I went this way rather than adding a visited set at the crash site because the same unguarded `get_type_alias` recursion also exists in `solver/subset.rs` and `alt/call.rs`. Removing the unbounded reference at the source covers those without a guard in each one.

The base-case analysis is untouched, so an alias that merely points at a well-founded recursive alias is unaffected: the recorded name only widens what counts as a self-reference, and a union or container with a non-self-referential member is still accepted. There is a test pinning that, including the user-defined-generic case the existing doc comment warns about (`class C[T]: x: T | None` makes `type A = C[A]` inhabitable).

### Testing

`cargo test -p pyrefly` is 7133 passed, 0 failed. `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` exits 0. The one clippy warning it prints is pre-existing in `pyrefly_config`, in code this PR does not touch.

Three tests added to `recursive_alias.rs`, following the guidance to start from a failing test:

- `test_cyclic_indirect_self_reference`: the issue, reduced
- `test_cyclic_indirect_self_reference_multi_hop`: pins that depth does not matter
- `test_alias_referencing_valid_recursive_alias_is_not_cyclic`: guards against over-flagging

Before the change both of the first two abort the test binary with SIGABRT; after it they pass, and all pre-existing cyclic-alias tests keep passing unchanged.

I also confirmed the original repro from the issue against a release build: it now reports `cyclic self-reference in T1` and `in T2` plus the legitimate uninitialized-name error, with no crash.

### One thing worth splitting off

The default `basic` preset does not enable `invalid-type-alias`, so `type W = W` reports zero errors from the CLI even though the test suite asserts a diagnostic for it. That is unrelated to this crash and I have not touched it, but it does mean the existing cyclic-alias diagnostics are invisible to most CLI users. Happy to file a separate issue if you think it is worth changing.

---

Submitted by an AI agent (Claude), as asked for in CONTRIBUTING.md. Reviewed by the PR owner before submission.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -1545,8 +1545,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     /// user-defined `class C[T]: x: T | None` makes `type A = C[A]`
     /// inhabitable (e.g. `C(x=C(x=None))`), so we can't assume all generic
     /// containers require their type parameter.
+    /// `names` holds the alias being resolved plus any alias left as a
+    /// recursive reference while expanding its body, so a cycle that is merely
+    /// reachable is caught as well as one the alias belongs to.
     /// Returns `true` if a cyclic self-reference was found.
-    fn type_alias_has_cyclic_reference(&self, name: &Name, ta: &TypeAlias) -> bool {
+    fn type_alias_has_cyclic_reference(&self, names: &SmallSet<Name>, ta: &TypeAlias) -> bool {
         // Unwrap the type[body] wrapper. We operate on the inner body because
         // map_over_union wraps inner union members in type[...] when traversing
         // inside Type::Type, which would prevent matching UntypedAlias nodes.
@@ -1557,7 +1560,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             Type::Type(inner) => inner.as_ref(),
             _ => return false,
         };
-        let is_self_ref = |ty: &Type| matches!(ty, Type::UntypedAlias(ta) if ta.name() == name);
+        let is_self_ref =
+            |ty: &Type| matches!(ty, Type::UntypedAlias(ta) if names.contains(ta.name()));
 
         fn collect_tuple_members(tuple: &Tuple) -> Vec<&Type> {
             match tuple {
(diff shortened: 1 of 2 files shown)


shobhitmehro:
looks good to me, thank you!

shobhitmehro (approved):

yangdanny97 (approved):

rchen152 (approved):

PR #4371, opened 2026-07-30
Revert bundled httpx stubs (fixes #4366)

### LINKED ISSUE #4366: Bundled HTTPX stubs omit public MockTransport
## Summary

The new bundled HTTPX stubs from PR #4233 / commit 65ab13722 omit the public
`httpx.MockTransport` export. Code that was clean in 1.1.1 gets
`missing-attribute` errors in 1.2.0-dev.3.

This accounts for 22 new diagnostics in one corpus repository.

## Reproduction

```python
import httpx

def handler(request: httpx.Request) -> httpx.Response:
    return httpx.Response(200)

transport: httpx.MockTransport = httpx.MockTransport(handler)
```

Using the same Python environment with HTTPX 0.28.1 and the default preset:

- Pyrefly 1.1.1: clean
- Pyrefly 1.2.0-dev.3: two `missing-attribute` errors
- Runtime: `httpx.MockTransport` exists and constructs successfully

The bundled stub tree has no `MockTransport` definition/export. The bundled
surface should include this public, typed HTTPX API.

anushamukka9:
Summary:
The bundled httpx stubs added in D110685002 caused a diagnostics regression between Pyrefly 1.1.1 and 1.2.0-dev.3 (GitHub https://github.com/facebook/pyrefly/issues/4366): code using `httpx.MockTransport` that type-checked cleanly in 1.1.1 started reporting missing-attribute errors (22 new diagnostics in one corpus repo).

Original intent of D110685002: mirror the existing bundled stubs (boto3, pandas, etc.) so that httpx users get out-of-the-box type checking even without the library installed locally / discoverable by the module resolver.

The slip-up: unlike those libraries, httpx ships `py.typed` with full inline annotations -- it is its own authoritative type source. The hand-written bundled stubs were therefore both unnecessary and, worse, incomplete: they omitted `_transports/mock.pyi` and never re-exported `MockTransport` from `__init__.pyi`. Because a bundled stub shadows the upstream inline types, this turned a fully-typed public API into false-positive diagnostics. (The stubs were also added without a corresponding entry in `stubs_metadata.json`.)

The fix: remove the entire `httpx-stubs/` tree (all 19 vendored files), reverting D110685002. Import resolution then falls back to httpx's own authoritative inline types, which are complete and include `MockTransport`.

Why remove rather than patch `MockTransport`: adding the one missing symbol would only close today's gap. Maintaining a hand-written stub for an already-fully-typed library commits us to tracking parity with every future httpx release -- ongoing drift and maintenance cost for no benefit over the upstream types. Removal is the correct, durable fix.

Differential Revision: D114158493


yangdanny97 (approved):

PR #4383, opened 2026-07-31
Add a `#bench` opt-in for the PyTorch walltime benchmarks (#4383)

kinto0:
Summary:

- Let a PR opt into the walltime PyTorch benchmarks by putting `#bench` in its description, run them nightly on main so there is something to compare against, and stop CodSpeed generating flame graphs for them.

- `walltime` was gated to `workflow_dispatch`, which in practice meant unreachable: dispatch can only target a branch or tag in this repo, so benchmarking a PR — exported or from a fork — meant pushing its code to a branch by hand first. It also never ran on main, so CodSpeed had no base run to compare anything against; the only baseline data was whatever backtest dispatches happened to seed. Separately, flame-graph generation fails on nearly every walltime run and leaves an "Unable to generate the flame graphs" comment behind; ruff hit the same thing and turned them off (astral-sh/ruff#20419), which also saves them ~6 minutes.

- The opt-in is the description rather than a comment because GitHub only ever runs the *default-branch* copy of a workflow for `issue_comment`, so a comment trigger could never be tested or iterated on from a PR. `pull_request` runs the PR's own copy, so adding `edited` to its `types` and reading `github.event.pull_request.body` works from the branch itself. `facebook-github-bot` is named explicitly in the gate because it authors every exported PR and its `author_association` is not one of the trusted values — without that clause the opt-in could never fire on an exported PR, which is how essentially all of our PRs arrive.

- The baseline is a nightly schedule rather than a run per push to main, because `codspeed-macro` bills per minute and a daily measurement of a branch that moves slowly is enough to compare against. ruff instead runs walltime on every main push (`github.ref == 'refs/heads/main'` in its CodSpeed gates); that is the more accurate option if the nightly baseline turns out to be too coarse.

Differential Revision: D114413106

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -3,6 +3,10 @@ on:
   push:
     branches: [ main ]
   pull_request:
+    types: [opened, synchronize, reopened, edited]
+  # A daily run keeps a recent walltime measurement of main on file.
+  schedule:
+    - cron: '0 6 * * *'
   # workflow_dispatch lets CodSpeed trigger backtest runs to seed baseline data.
   workflow_dispatch:
 permissions:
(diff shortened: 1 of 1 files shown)


kinto0:
#bench

rchen152 (approved):

PR #4390, opened 2026-08-01
Fix positional inlay hints after keyword arguments

### LINKED ISSUE #4387: Inserting an inlay hint causes incorrect hints after keyword arguments
### Describe the Bug

When using double-click insertion for positional-argument inlay hints, inserting the first hint can cause the next hint to display the wrong parameter name. Starting to insert inlay hints at the earlier hints is quite normal, but that causes the problem case described below.

### Example case

```python
def example(a: str, b: str) -> None:
    ...

example("a", "b")
```

### Current behavior

Double-clicking the `a= ` hint inserts `a=` correctly, but the second argument now displays `a= ` instead of `b= `.

### Expected behavior

Double-clicking the `a= ` hint inserts `a=`, and the second argument should display `b=`.

### Sandbox Link

_No response_

VS code extension Version 1.1.1

vincevannoort:
## Description

Fixes incorrect positional-argument inlay hints after inserting an inlay hint into a call containing keyword arguments.

Added regression tests for:
- a remaining positional argument after a keyword argument
- multiple positional arguments.

Closes: https://github.com/facebook/pyrefly/issues/4387

### DIFF
--- a/pyrefly/lib/lsp/wasm/inlay_hints.rs
+++ b/pyrefly/lib/lsp/wasm/inlay_hints.rs
@@ -423,7 +423,16 @@ impl<'a> Transaction<'a> {
                         callee_type.and_then(normalize_singleton_function_type_into_params)
                     {
                         for (arg_idx, arg) in call.arguments.args.iter().enumerate() {
-                            // Skip keyword arguments - they already show their parameter name
+                            // Account for keyword arguments omitted from `args`, including
+                            // malformed calls while an inlay edit is being applied.
+                            let positional_arg_idx = arg_idx
+                                + call
+                                    .arguments
+                                    .keywords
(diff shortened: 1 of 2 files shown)


rchen152 (approved):
This looks good to me, thank you!

vincevannoort:
@rchen152 thanks for the review! I see the [internal builds & tests](https://github.com/facebook/pyrefly/pull/4390/checks?check_run_id=91807405046) have failed, anything I can do?

rchen152:
Ah, that's totally fine to ignore; it's just an artifact of how our internal CI works. (I rebased your PR onto the latest main, so the tests on the un-rebased version got killed.)

grievejia (approved):

PR #4391, opened 2026-08-01
[pyrefly] Contextualize dict subscript literals for Any

### LINKED ISSUE #4301: `implicit-any-empty-container` false positive when assigning dict containing None to `dict[str, Any]`
### Describe the Bug

Given the following code:

```py
from typing import Any

def from_response(response: dict[str, Any]) -> None:
    response["meta"] = {"time_to_live": None}
```

Pyrefly reports:

```console
❯ target\debug\pyrefly.exe --version
pyrefly 1.2.0-dev.3

❯ target\debug\pyrefly.exe check ../privatebin/t.py
ERROR Cannot infer type of empty container; it will be treated as containing `Any` [implicit-any-empty-container]
 --> C:\Users\raven\dev\github\privatebin\t.py:7:41
  |
7 |     response["meta"] = {"time_to_live": None}
  |                                         ^^^^
  |
  Consider adding a type annotation or initializing with a non-empty value
 INFO 1 error
```

This looks closely related to https://github.com/facebook/pyrefly/issues/3593, which was closed as fixed.

To check whether this was already addressed by that fix, I built Pyrefly from commit c28caa75a93fc72a03bda06e85b19ebdf9d02a4e, but I was still able to reproduce the issue there.

In this case, `response` is explicitly annotated as `dict[str, Any]`, so I would expect the assigned value `{"time_to_live": None}` to be accepted without an `implicit-any-empty-container` error.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BBdKQA6lEejETMMMHTDUaAfUrwe6ODAAUquOs2I6mCAGMGAbTgNKAGkHCAugEo6AWgB8dAHK50MRFLodMF0uvow5iIgNDAMqFEOdAC8dMBRDFwwSgy4SlAQAG4wUYY%2BfgC%2BgSA2IGSqYFCkhDk0UBQAxHQACqT1jXRoWHj4dCa%2BkGwArpSoGb6EEp0AyjAwdAAWDAzEcIgA9Ht1so2EvGx7MOh7mLgmcHtj6BPTsxC%2Be-K8dKgFqNCo2Fgo3GECmMzmQVwxAhcAW6DIDHWvlcRUocDeQRSUQAzIQAIwAJiiEmIuliyToUSslFMDGJ6HMMEo1DRDgkk3QXB4fBgmFcxlUZkKaxSAHJQeheDBRRJJQxXKoAI6TCCqPkAaxgpFcqBMJngcApooA7qhKOgZegQOUarqMkUAGLQGAUQY4AgkcjlIA

_No response_

lyydsheep:
Fixes #4301

A dict literal assigned through an explicitly annotated `dict[str, Any]` subscript now receives a soft value hint, avoiding a spurious `implicit-any-empty-container` diagnostic. Unannotated and narrow dictionary targets retain their existing diagnostics and inference behavior.

Validation:
- `cargo test -p pyrefly test::inference::test_implicit_any`
- `cargo test -p pyrefly test::dict`

The local `.scratch` planning files are intentionally not part of this PR.

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -5477,6 +5477,21 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     ) -> (Type, Type) {
         let base = self.expr_infer(&subscript.value, errors);
         let slice_ty = self.expr_infer(&subscript.slice, errors);
+        // A declared dict[str, Any] target gives a dict-literal RHS a soft value hint.
+        let dict_value_hint = match &base {
+            Type::ClassType(cls) if cls.has_qname("builtins", "dict") => {
+                match cls.targs().as_slice() {
+                    [key, value]
+                        if matches!(key, Type::ClassType(cls) if cls.has_qname("builtins", "str"))
+                            && value.is_any() =>
+                    {
+                        Some(value.clone())
+                    }
+                    _ => None,
+                }
+            }
+            _ => None,
+        };
         let assigned_ty = self.distribute_over_union(&base, |base| {
             self.distribute_over_union(&slice_ty, |key| {
                 match (base, key) {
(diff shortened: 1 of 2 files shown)


IBlackVoid:
Heads up rather than having this show up unannounced — I opened #4409 with a more general take on the same bug. You got here first, and I'd rather flag it than surprise you.

Same idea, but keyed on whether the placeholder can *escape* rather than on the target's shape, which turned out to matter more than I expected. Measured, both patches applied to the same base and run on the same inputs:

| case | main | #4391 | #4409 |
|---|---|---|---|
| `dict[str, Any]` (the report) | error | fixed | fixed |
| `dict[int, Any]` | error | error | fixed |
| `MutableMapping[str, Any]` | error | error | fixed |
| `defaultdict[str, Any]` | error | error | fixed |
| `list[Any]` | error | error | fixed |
| hand-written `__setitem__` | error | error | fixed |
| `x = {}` then `x["y"] = {...}` | error | error | error |

The key-type condition in the qname match is what narrows yours: the key has nothing to do with whether the *value* is `Any`, so `dict[int, Any]` gets missed. Real code hits this — `test_configparser.py` does `cf["A"] = {}` where `cf` is a `MutableMapping`, not a `builtins.dict`.

Digging further also turned up a second root cause behind the same diagnostic: checking against `Any` never pins the placeholder either, which produces the identical false positive on attribute assignment, return position, default arguments and TypedDict fields.

Genuinely happy to fold this into your PR instead of landing it separately if you'd prefer — the analysis is the part I care about, not whose branch it goes in on.

rchen152:
This PR special-cases the specific example in the issue instead of addressing the underlying problem. https://github.com/facebook/pyrefly/pull/4409 looks like it attempts a more general solution, so I'm going to close this one.

PR #4392, opened 2026-08-01
[pyrefly] Rename legacy type parameter constructor names

### LINKED ISSUE #4348: Renaming: Rename TypeVar or ParamSpec
ref https://github.com/astral-sh/ty/issues/3352

lyydsheep:
Fixes #4348

LSP rename now includes the string name argument in legacy TypeVar and ParamSpec constructor calls. Other strings and constructor calls are unchanged.

Validation:
- cargo test -p pyrefly test::lsp::rename
- python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema

The local `.scratch` planning files are intentionally not part of this PR.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -69,6 +69,7 @@ use crate::alt::answers::Index;
 use crate::alt::answers_solver::AnswersSolver;
 use crate::alt::attr::AttrDefinition;
 use crate::alt::attr::AttrInfo;
+use crate::binding::binding::Binding;
 use crate::binding::binding::Key;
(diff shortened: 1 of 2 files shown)


stroxler (approved):
Thanks for the PR! This looks pretty good to me, I'll try to get it merged in the next day or two.

In the interest of working through the PR backlog (which has become a challenge) I'm going to just patch a couple of things internally so we don't need a back-and-forth on this, but two minor nits here:
- You tackled TypeVar and ParamSpec, but TypeVarTuple is similar
- There's a helper function `arguments.find_argument_value` that can avoid the iteration

grievejia (approved):

PR #4398, opened 2026-08-01
Fix  Resolve TypedDict keys in go-to-definition

### LINKED ISSUE #4360: Support go-to-definition for functional TypedDicts
ref https://github.com/astral-sh/ty/issues/1322

aodihis:
# Summary

Fixes #4360

Go-to-definition did not handle TypedDict subscript keys. In `person["name"]`, `"name"` is a string literal, so no identifier is found and the request enters the `None` case. It then falls back to operator navigation and incorrectly resolves to `dict.__getitem__`.

This adds TypedDict handling before the operator fallback. It resolves keys using the inferred TypedDict fields and MRO, supporting class-based, inherited, and functional TypedDicts.

If the TypedDict does not contain the key, go-to-definition now returns no definition instead of falling back to `__getitem__`.

# Test Plan

Added:

- `typed_dict_key_test` for class-based, inherited, and functional TypedDicts.
- `typed_dict_missing_key_test` to verify missing keys return no definition.

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -34,6 +34,7 @@ use pyrefly_python::symbol_kind::SymbolKind;
 use pyrefly_python::sys_info::SysInfo;
 use pyrefly_types::type_alias::TypeAliasData;
 use pyrefly_types::typed_dict::TypedDict;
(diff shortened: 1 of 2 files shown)


stroxler (approved):

Issue #2402, opened 2026-02-13
pyrefly has issue running in Codex worktrees


gmetzker-4c:
### Describe the Bug

I'm trying to use OpenAI Codex desktop with worktrees.  Codex creates these in some path like `~/.codex/worktrees/XXXX/{project-name}`, and Pyrefly doesn't seem to like the hidden folder it's running in.

```shell
poetry run pyrefly check
 INFO Checking project configured at `/Users/gary.metzker/.codex/worktrees/0bbb/my-project/pyproject.toml`
Pattern /Users/gary.metzker/.codex/worktrees/0bbb/my-project/**/*.py* is matched by `project-excludes` or ignore file.
`project-excludes`: [**/node_modules, **/__pycache__, **/venv/**/*, **/.[!/.]*/**/*, /Users/gary.metzker/.pyenv/versions/3.12.11/lib/python3.12, /Users/gary.metzker/.pyenv/versions/3.12.11/lib/python3.12/lib-dynload, /Users/gary.metzker/Library/Caches/pypoetry/virtualenvs/my-project-WFsQtCVv-py3.12/lib/python3.12/site-packages], ignore files []
```
I get the error above without any custom `project-excludes`.

I assume it's something to do with the default glob path not liking `.codex`: `"**/.[!/.]*/**"`

I would expected if I disable the defaults via `disable-project-excludes-heuristics` and then add them manually to `project-excludes` I would expect the same behavior, but oddly enough this works.

```toml
[tool.pyrefly]
disable-project-excludes-heuristics = true
project-excludes = [
    "**/node_modules", 
    "**/__pycache__", 
    "**/venv/**/*",
    "**/.[!/.]*/**/*",
    "/Users/gary.metzker/.pyenv/versions/3.12.11/lib/python3.12",
    "/Users/gary.metzker/.pyenv/versions/3.12.11/lib/python3.12/lib-dynload",
    "/Users/gary.metzker/Library/Caches/pypoetry/virtualenvs/my-project-WFsQtCVv-py3.12/lib/python3.12/site-packages"
]
```
```shell
❯ poetry run pyrefly check
 INFO Checking project configured at `/Users/gary.metzker/.codex/worktrees/0bbb/my-project/pyproject.toml`
 INFO 0 errors (200 suppressed)
```

[...]

PR #4402, opened 2026-08-01
Honour repository boundaries in the ignore-file search; filter hidden dirs relative to include roots

### LINKED ISSUE #4401: Ignore files and hidden-dir filtering exclude every include when the project is a linked git worktree under a hidden directory
### Describe the Bug

Running `pyrefly check` inside a **linked git worktree** placed under a hidden directory of the main checkout — the layout several agent tools create (`.claude/worktrees/<name>/`, and `~/.codex/worktrees/…` as in #2402) — skips every include with `No Python files matched`. Two independent mechanisms cause it:

**1. The ignore-file search does not honour repository boundaries.**

A linked worktree's `.git` is a `gitdir: …` pointer *file*, so the upward search for `.gitignore`/`.ignore`/`.git/info/exclude` walks straight past the working-tree root and finds the **enclosing checkout's** `.git/info/exclude`, applying its patterns rooted at the enclosing checkout. A pattern like `**/.claude/worktrees/` in the main checkout's exclude file (how these tools keep worktrees out of `git status`) then matches **every file inside the worktree**:

```
ignore files [<worktree>/.gitignore, /path/to/main-checkout/.git/info/exclude]
No Python files matched patterns …
```

git's own semantics: a directory carrying a `.git` entry (file **or** directory) is a working-tree root; an outer repository's ignore rules never apply inside it, and the repository's shared `info/exclude` applies to each worktree **relative to that worktree's own root** (`git -C <worktree> check-ignore` confirms the files are not ignored).

**2. `HiddenDirFilter::RelativeTo` only carries `import_root`** (follow-up to #2402).

For a src-layout project `import_root` is `<project>/src`, so an include outside it — `tests/check.py` — matches no root and falls back to checking the **absolute** path's components, where a checkout under `.claude/…` has every chain hidden; the include is skipped even though the project's own files contain no hidden component. Additionally, the root list is only populated when `use_ignore_files` is true, so `--use-ignore-files=false` silently degrades the filter to `All` — disabling one exclusion mechanism makes another stricter.

### Reproduction

```sh
git init main && cd main && python -c 'open("pyproject.toml","w")'
printf '**/wt/\n' >> .git/info/exclude
git worktree add .claude/worktrees/wt
cd .claude/worktrees/wt
printf '[tool.pyrefly]\nproject-includes=["src"]\n' > pyproject.toml
mkdir src && touch src/m.py
pyrefly check   # every include skipped
```

### Expected behavior

The worktree checks like any other checkout: outer ignore rules stop at the working-tree boundary, the shared `info/exclude` applies relative to the worktree root, and hidden ancestors above the project's include roots don't hide the project's own files.

I have a fix for both halves (boundary-aware ignore-file search that resolves `gitdir`/`commondir` for linked worktrees, and `RelativeTo` built from every include root plus the import root, decoupled from `use_ignore_files`) — PR incoming, happy to adjust the approach.

willemkokke:
Fixes #4401. Two commits, one per mechanism:

**1. `find_ignore_files` honours repository boundaries the way git does** (`pyrefly_util/globs.rs`)

The upward search for `.gitignore`/`.ignore`/`.git/info/exclude` treated the directory tree as one hierarchy: from a linked worktree (whose `.git` is a `gitdir:` pointer file, not a directory) it walked past the working-tree root, found the enclosing checkout's `.git/info/exclude`, and applied its patterns rooted at the enclosing checkout — so a pattern like `**/.claude/worktrees/` in the main checkout's exclude file matched every file inside the worktree and all includes were skipped.

The search now collects the nearest `.gitignore` and `.ignore` walking upward, stops at the first directory carrying a `.git` entry (file or directory — the working-tree root), and resolves `info/exclude` through the `gitdir` pointer and `commondir` for linked worktrees, rooted at the working tree being checked — matching `git check-ignore`'s verdicts. The memoized `UpwardSearch` statics are replaced by a direct ancestor walk; it runs once per `GlobFilter` construction and is O(depth) file stats.

**2. `HiddenDirFilter::RelativeTo` is built from every include root plus the import root** (`pyrefly_config/config.rs`)

Follow-up to #2402: the allowance carried only `import_root`, so for a src-layout project (`import_root = src/`) an include outside it (`tests/check.py`) fell back to absolute-path component checking and was hidden by the checkout's own ancestors. Also decoupled from `use_ignore_files`, which previously degraded the filter to `All` when ignore files were turned off — disabling one exclusion mechanism silently hardened another.

**Tests**: two new tests in `globs.rs` (nested-repository boundary; linked worktree with `gitdir` + `commondir` resolution, asserting the shared exclude file still applies but relative to the worktree root); the existing `test_get_filtered_globs_coverage_scope` expectation updated to the new `RelativeTo` construction. `cargo test -p pyrefly_util -p pyrefly_config` passes (318 tests).

**Verified end to end** on the real-world layout that surfaced this (a `.claude/worktrees/…` checkout of a src-layout project whose main checkout excludes `**/.claude/worktrees/`): before, every include was skipped; after, the project checks clean with all-default settings, in both `--use-ignore-files` states.

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -796,9 +796,24 @@ impl ConfigFile {
         let hidden_dir_filter = if self.disable_project_excludes_heuristics {
             HiddenDirFilter::Disabled
         } else {
-            match root {
-                Some(r) => HiddenDirFilter::RelativeTo(vec![r.to_path_buf()]),
-                None => HiddenDirFilter::All,
+            // Hidden ancestors above the project's own roots must not hide
+            // the project's files: check components relative to every
+            // include root, plus the import root. The import root alone is
+            // not enough — a src-layout project's `import_root` is `src/`,
+            // and an include outside it (`tests/check.py`) would fall back
+            // to the absolute path, where a checkout under a hidden
+            // directory (`~/.codex/worktrees/…`, `.claude/worktrees/…`)
+            // has every component chain hidden. Deliberately independent of
+            // `use_ignore_files`: turning ignore files off must not make
+            // hidden-directory filtering stricter.
+            let mut roots = includes.roots();
+            if let Some(import_root) = self.import_root.as_deref() {
+                roots.push(import_root.to_path_buf());
+            }
+            if roots.is_empty() {
+                HiddenDirFilter::All
+            } else {
+                HiddenDirFilter::RelativeTo(roots)
             }
         };
         FilteredGlobs::new(includes, project_excludes, root, hidden_dir_filter)
@@ -3320,7 +3335,10 @@ output-format = "omit-errors"
                 globs(&["covered/**"]),
                 globs(&excludes),
                 None,
-                HiddenDirFilter::All,
+                // Hidden-directory filtering is relative to the include
+                // roots, so a project under a hidden directory still sees
+                // its own files.
+                HiddenDirFilter::RelativeTo(globs(&["covered/**"]).roots()),
             )
         };
 
--- a/crates/pyrefly_util/src/globs.rs
+++ b/crates/pyrefly_util/src/globs.rs
@@ -6,7 +6,6 @@
  */
 
 use std::ffi::OsStr;
-use std::ffi::OsString;
 use std::fmt;
 use std::fmt::Debug;
 use std::fmt::Display;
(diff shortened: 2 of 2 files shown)


kinto0 (changes requested):
hi, sorry about the churn. are you able to make separate PRs for each of these two issues?

willemkokke:
No problem — split into two PRs, one per issue:

- #4527 — stop the ignore-file search at the working-tree boundary (fixes #4525)
- #4528 — filter hidden directories relative to every include root (fixes #4526)

Both are rebased on current `main` and touch disjoint files, so they can land in either order. #4528 also gains a dedicated regression test that fails on `main`. Closing this one.

PR #4406, opened 2026-08-02
Dedup tparam-default substitution between alt and solver layers

### LINKED ISSUE #4298: TODO: deal with code duplication in get_tparam_default
```rs
    fn get_tparam_default(
```

and

```rs
    pub fn finish_class_targs(&self, targs: &mut TArgs, uniques: &UniqueFactory) {
```

has ```// TODO: deal with code duplication in get_tparam_default```

maybe

```rs
fn substitute_tparam_default(
    default: &Type,
    lookup: impl Fn(&Name) -> Option<Type>,
    fallback: Type,
) -> Type
```

provide different lookup and fallback.

nitishagar:
## Summary

The logic for substituting references to earlier type parameters inside a tparam's `default` was duplicated in two places:

- `get_tparam_default` in `pyrefly/lib/alt/class/targs.rs` (used during call/specialization)
- an inlined copy in `finish_class_targs` in `pyrefly/lib/solver/solver.rs` (used during class-level finalization), flagged by a `// TODO: deal with code duplication in get_tparam_default` marker.

The two copies differed only in their lookup source and their fallback value for out-of-scope references — differences that are intentional, since one path runs where an out-of-scope reference is an already-reported error (returns `Any(Implicit)`) and the other runs during finalization with no such error (returns the param's gradual type).

This PR extracts a shared `substitute_tparam_defaults` helper into `pyrefly_types::quantified` (next to the existing `as_gradual_type_helper`, which already uses the same `TypeVar`/`TypeVarTuple`/`ParamSpec`/`Quantified` match idiom). The helper takes the default by value, a `&dyn Fn(&Name) -> Option<Type>` lookup closure, and a caller-chosen fallback `Type`, so each call site keeps its own lookup source and fallback semantics. Both callers are rewritten to delegate to it, and the `// TODO` marker is removed.

## Behavior

This is a behavior-preserving refactor. Each call site passes the same lookup and fallback it used inline, so the produced types are identical for every input:

- `get_tparam_default` passes `name_to_idx` -> `checked_targs` lookup with `self.heap.mk_any_implicit()` fallback (unchanged).
- `finish_class_targs` passes `seen_params` -> `new_targs`-then-`targs` lookup with `param.as_gradual_type()` fallback (unchanged).

The two fallbacks are deliberately **not** collapsed — they intentionally differ (error path vs finalization path).

## Test plan

- `cargo build -p pyrefly_types` — clean
- `cargo build -p pyrefly` — clean
- `cargo test -p pyrefly --lib` — **7675 passed, 0 failed, 3 ignored** (the full typechecker regression suite; zero behavior diff)
- `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` — clean for the touched files

Fixes #4298

### DIFF
--- a/crates/pyrefly_types/src/quantified.rs
+++ b/crates/pyrefly_types/src/quantified.rs
@@ -443,3 +443,30 @@ impl Quantified {
         Self::as_gradual_type_helper(self.kind(), self.default())
     }
 }
+
+/// Substitute references to earlier type parameters inside a tparam default.
+///
+/// The `default` is a tparam's default type, whose leaves may reference earlier
+/// tparams (matched here as `TypeVar`/`TypeVarTuple`/`ParamSpec`/`Quantified`).
+/// `lookup` resolves an earlier tparam's name to its already-computed targ;
+/// when a name is out of scope (no entry), `fallback` is used. Both are passed
+/// by the caller because the two call paths (call/specialization vs class-level
+/// finalization) intentionally differ in lookup source and fallback value.
+pub fn substitute_tparam_defaults(
+    default: Type,
+    lookup: &dyn Fn(&Name) -> Option<Type>,
+    fallback: Type,
+) -> Type {
+    default.transform(&mut |default| {
+        let name = match default {
+            Type::TypeVar(t) => Some(t.qname().id()),
+            Type::TypeVarTuple(t) => Some(t.qname().id()),
+            Type::ParamSpec(p) => Some(p.qname().id()),
+            Type::Quantified(q) => Some(q.name()),
+            _ => None,
+        };
+        if let Some(name) = name {
+            *default = lookup(name).unwrap_or_else(|| fallback.clone());
+        }
+    })
+}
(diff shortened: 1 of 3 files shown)


rchen152 (commented):
I'm not convinced this is an improvement. Helpers add conceptual overhead, and callbacks add implementation complexity, whereas the original code is simple and clear, and the duplication is small.

nitishagar:
Thanks for taking the time. A couple of thoughts on the trade-off, happy to drop this if it's not convincing:

- The shape of this helper (default + lookup closure + fallback) is what #4298 proposes, and the `// TODO` marker it points at has been sitting in `finish_class_targs` for a while. So the dedup is fulfilling a tracked request rather than being introduced speculatively.
- My main motivation is drift risk: the two copies do the same `TypeVar`/`TypeVarTuple`/`ParamSpec`/`Quantified` match, and a future change to one (e.g. handling a new tparam form) would have to be remembered in both. The two intentionally differ only in lookup source and fallback, which are exactly the two parameterized pieces.

That said, I take the point on the `&dyn Fn` adding implementation weight. If the conceptual overhead isn't worth it for this size of duplication, I can just close this and #4298. Let me know which way you'd like to go.

rchen152:
I dug into this a bit further. The code duplication in `get_tparam_default` and `finish_class_targs` is actually a symptom of a deeper problem, which is that we're storing raw type variables in `TParams` and then sanitizing them every time we finalize a list of type arguments. The better thing to do is to sanitize once when creating the tparams, which naturally removes the duplication. I've done this in https://github.com/facebook/pyrefly/pull/4483.

Thanks for taking the time to open a PR. Sorry this ended up being trickier than it initially looked!

PR #4408, opened 2026-08-02
Dedup add_lower_bound/add_upper_bound into add_var_bound

### LINKED ISSUE #4292: dedup add_lower_bound and add_upper_bound
https://github.com/facebook/pyrefly/blob/fb7004fb993c92927a5264de237daf64ef601c55/pyrefly/lib/solver/solver.rs#L2000

https://github.com/facebook/pyrefly/blob/fb7004fb993c92927a5264de237daf64ef601c55/pyrefly/lib/solver/solver.rs#L2067

is merely mirror impl.

how about

```rs
  enum BoundSide {
      Lower,
      Upper,
  }

  fn add_var_bound(
      &self,
      var: Var,
      bound: Type,
      side: BoundSide,
      is_subset: &mut dyn FnMut(&Type, &Type) -> Result<(), SubsetError>,
  ) -> Result<(), SubsetError>
```

nitishagar:
## Summary

`add_lower_bound` and `add_upper_bound` were near-perfect mirrors: identical two-phase lock structure, identical `(res.is_ok(), quantified_kind)` match (including the IntVar arm), and identical fallbacks. They differed only in ways fully derivable from a single `is_upper` bool — which `Bounds` field is the bound-side vs the opposite-side, the `is_subset` argument order, and the `is_upper` value threaded through `get_new_bound`.

This extracts a private `add_var_bound(v, bound, is_upper, is_subset)` holding the shared body, and reduces the two public functions to one-line delegations (`false` / `true`). This mirrors the established `get_new_bound(..., is_upper: bool, ...)` idiom already in this file, keeping the change idiomatic rather than introducing a one-use enum.

## Behavior preservation

The refactor is behavior-preserving: the helper is a pure relocation of the existing `add_lower_bound` statements with every divergence resolved by an `is_upper` branch that reproduces the original `add_upper_bound` text. Concretely preserved:
- The two-phase lock pattern (read under lock, drop guard, validate without lock so the `is_subset` callback can safely recurse into `is_subset_eq`, re-lock to write) — holding the lock across `is_subset` would deadlock.
- The IntVar normalization (`type_as_intvar_solution(&bound).expect(...)`).
- The `(false, IntVar) => None` drop and the `// TODO(issues/105)` any-error fallback.
- Public signatures are unchanged, so all five callers need no edits.

The full `cargo test -p pyrefly --lib` suite passes unchanged.

Fixes #4292

### DIFF
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -2088,24 +2088,59 @@ impl Solver {
         Ok(())
     }
 
-    pub fn add_lower_bound(
+    /// Shared core of [`Self::add_lower_bound`] (`is_upper == false`) and
+    /// [`Self::add_upper_bound`] (`is_upper == true`).
+    ///
+    /// The two public functions are near-perfect mirrors: every divergence is
+    /// mechanically derivable from `is_upper` — which `Bounds` field is the
+    /// bound-side vs the opposite-side, the `is_subset` argument order, and the
+    /// `is_upper` bool threaded through `get_new_bound` (the same idiom
+    /// [`Self::get_new_bound`] already uses).
+    ///
+    /// Preserves the two-phase lock pattern of the originals: read under the
+    /// lock, drop the guard, validate without the lock (the `is_subset` callback
+    /// recurses into `is_subset_eq` which re-locks `variables`), then re-lock to
+    /// write. Holding the lock across `is_subset` would deadlock.
+    fn add_var_bound(
         &self,
         v: Var,
         bound: Type,
+        is_upper: bool,
         is_subset: &mut dyn FnMut(&Type, &Type) -> Result<(), SubsetError>,
     ) -> Result<(), SubsetError> {
         let lock = self.variables.lock();
         let e = lock.get(v);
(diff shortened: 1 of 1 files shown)


pyrefly/lib/solver/solver.rs
```diff
@@ -2088,24 +2088,59 @@ impl Solver {
...
+    /// The two public functions are near-perfect mirrors: every divergence is
+    /// mechanically derivable from `is_upper` — which `Bounds` field is the
+    /// bound-side vs the opposite-side, the `is_subset` argument order, and the
+    /// `is_upper` bool threaded through `get_new_bound` (the same idiom
+    /// [`Self::get_new_bound`] already uses).
+    ///
+    /// Preserves the two-phase lock pattern of the originals: read under the
+    /// lock, drop the guard, validate without the lock (the `is_subset` callback
+    /// recurses into `is_subset_eq` which re-locks `variables`), then re-lock to
+    /// write. Holding the lock across `is_subset` would deadlock.
```
kinto0, line 2103: this comment isn't necessary

kinto0 (approved):
thanks for submitting this cleanup! comments are a little verbose but otherwise LGTM. ill fix them on import unless you get to it first

Issue #3828, opened 2026-06-15
Concatenate not handled in function with generic class as parameter.


cwastranis:
### Describe the Bug

When wrapping a generic class while using Concatenate to remove an arg from the ParamSpec, pyrefly fails to assign the input ParamSpec to the Concatenated spec.

```python
from typing import Concatenate, reveal_type

class A[**P]:
    pass

class B[**P]:
    pass

def transform[**P](arg: A[Concatenate[int, P]]) -> B[P]:
    raise NotImplementedError

a1: A[[int]] = A()
# ERROR sandbox.py:14:23-25: Argument `A[[int]]` is not assignable to parameter `arg` with type `A[Concatenate[int, P]]` in function `transform` [bad-argument-type]
reveal_type(transform(a1))
# expected: B[[]]
# revealed: B[@_]

a2: A[[int, str, int]] = A()
# ERROR sandbox.py:20:23-25: Argument `A[[int, str, int]]` is not assignable to parameter `arg` with type `A[Concatenate[int, P]]` in function `transform` [bad-argument-type]
reveal_type(transform(a2))
# expected: B[[str, int]]
# revealed: B[@_]
```

This superficially seems similar to #3706, I'm filing a separate issue because I don't know pyrefly internals well enough to assume they are or are not actually related.

### Sandbox Link

[...]

PR #4415, opened 2026-08-03
Visit the ParamSpec tail of `Type::Concatenate`

### LINKED ISSUE #3706: Paramspec type can get lost when concatenating into the same type
### Describe the Bug

When doing these sort of "transforming" operations that return a new specialization using `Concatenate`, pyrefly loses track of the starting specialization of the paramspec, and does not alter the resulting signature appropriately.

```python
from typing import Concatenate, reveal_type

class A[**P]:
    def transform(self) -> A[Concatenate[int, P]]:
        raise NotImplementedError

# Expected type: A[[int, int]]
reveal_type(A[int]().transform())  # revealed type: A[Concatenate[int, P]]
```

This sort of thing is useful for decomposing some decorators that both mutate a function's arguments and wrap the result in a class at the same time.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BhXOgDGqBjHTiYAGjqUYANxiooAfSbEYAHXS6RUVHDh0AggG0AVJYAKAXUS66zuphhhGlDHDC8aACjgYKDAASjoAWgA%2BM3MhUWkpCXNWBjl7BycXbK8IILoAOVwGAEluWBpJCUwAUUpqSl1dAGI6GvwtEWrGZhhEWJT0NM4huztdBWVVDV7-C1S7f1DCBi90Hz8l8LpWyZVYTB6tfot4sQkkmEHhjJAZEDIFMChSFdooClabUieXujQsHh8HQRMJIGwAK5eBgQYSEFp0ADKMBgdAAFgwGMQ4IgAPS4x7uF6EXhsXGSXGYXAiOC40HocFQ8Sw9C4ui%2BSh0VCKVDQVDYWAgsEQSHQll0XDEGHCODw9BkBho4QRZSUODigC8dG0IAAzIQAIwAFh1unMMHqvDg43QEPQXB4fBgmAimAgCi6EGUdC1AHIReheDBfbpAwwIgoAI4Q93OiIAaxgpAiqBEIngJj9AHdUJR0L6QABfe6pmHKABi0BgFABOAIJHIhaAA

_No response_

markselby9:
# Summary

`Type::Concatenate(Box<[PrefixParam]>, Box<Type>)` stores its trailing ParamSpec as an ordinary `Type`, but the hand-written `Visit`/`VisitMut` impls for `Type` discarded that second field:

```rust
Type::Concatenate(x, _) => x.visit(f),
```

Every traversal built on those two functions — `subst`, `collect_quantifieds`, `contains_type_variable`, `transform_mut`, and the solver's `resolve_vars` — was therefore a silent no-op on the tail. The `Callable` spelling was unaffected because it is normalized at construction into `Params::ParamSpec`, whose derived visitor sees both fields; only a bare `Concatenate` in a class type argument, e.g. `A[Concatenate[int, P]]`, reached the broken path. That asymmetry is exactly what both issues report.

Visiting the tail is sufficient. Once the tail's `Var` is reachable it gets resolved, which lets the solver's existing `Concatenate[..., ParamSpecValue]` and `Concatenate[..., Concatenate]` collapses fire for the first time, and `P` is now discovered as a type variable when it occurs only inside a tail. It also subsumes the hand-rolled recursion in `tvars_to_tparams_for_type_alias`, which can now use `recurse_mut` like its neighbouring arms, leaving `PrefixParam::ty_mut` unused.

I kept the impl hand-written rather than switching to `#[derive(Visit, VisitMut)]`: `impl Visit for Type` is deliberately monomorphic (`Visit<Type> for Type`), while the derive emits `impl<To> Visit<To>` with a where-clause per field — and several of `Type`'s payload types (`TArgs`, `Union`, `Arc<TParams>`, …) only implement the monomorphic form, so it would not compile. The hand-written arms also elide no-op recursion on the checker's hottest traversal. I added a comment on both arms stating the invariant so the omission can't quietly come back.

Fixes #3706, #3828.

# Test Plan

`cargo test -p pyrefly --lib` — 7468 passed, 0 failed. Full workspace `cargo test` green. Conformance output (`conformance.exp`, `conformance.result`, `results.json`) is byte-identical to the committed versions, with all four `generics_paramspec_*` cases still passing.

First commit adds the repros as `bug`-marked tests capturing the current wrong behaviour; the second removes the markers and updates them to the correct results. Also added coverage for paths that were previously unreachable: the nested `Concatenate[..., Concatenate[...]]` collapse, a bare `Concatenate` in both legacy and PEP 695 type aliases, and a `ParamSpec` reachable only through a tail being recognised as a type variable.

Exactly one pre-existing expectation changed, in `test_functools_partial_pattern`:

- before: `partial[Concatenate[*tuple[Unknown, ...], _P2], Ellipsis, Any, object, ...]`
- after: `partial[Concatenate[*tuple[Unknown, ...], Ellipsis], Ellipsis, Any, object, ...]`

That is the fix working. The overload-consistency check instantiates the declared return type with gradual forms, and the `_P2` in the second type-argument position was already rendering as `Ellipsis` — only the one inside the `Concatenate` tail was leaking through unsubstituted. The two now agree. That test keeps its unrelated `bug` marker.

One known gap, left as a `bug`-marked test (`test_concatenate_paramspec_tail_gradual`): a gradual `A[...]` still doesn't bind the tail, so `drop_int(g)` gives `A[@_]` and `add_int(g)` gives `A[Concatenate[int, Ellipsis]]`, where the `Callable` equivalents are correct. Both issues report concrete parameter lists, which this fixes; the `...` case needs separate work.

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -404,12 +404,6 @@ impl PrefixParam {
         }
     }
 
-    pub fn ty_mut(&mut self) -> &mut Type {
-        match self {
-            Self::PosOnly(_, ty, _) | Self::Pos(_, ty, _) => ty,
-        }
-    }
-
     /// Convert to a positional-only `Param`. Per the typing spec, params in
     /// `Concatenate` are positional-only at the call site. This is also appropriate
     /// for ParamSpec forwarding where prefix params must be passed positionally.
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -406,12 +406,6 @@ impl PrefixParam {
         }
     }
 
-    pub fn ty_mut(&mut self) -> &mut Type {
-        match self {
-            Self::PosOnly(_, ty, _) | Self::Pos(_, ty, _) => ty,
-        }
-    }
-
     /// Convert to a positional-only `Param`. Per the typing spec, params in
     /// `Concatenate` are positional-only at the call site. This is also appropriate
     /// for ParamSpec forwarding where prefix params must be passed positionally.
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -964,9 +964,6 @@ impl Visit for Type {
             Type::ParamSpec(x) => x.visit(f),
             Type::TypeVarTuple(x) => x.visit(f),
             Type::SpecialForm(x) => x.visit(f),
-            // The ParamSpec tail is a real `Type` (a `Var`, `Quantified`, `ParamSpecValue`, ...),
-            // so it must be visited. `subst`, `collect_quantifieds` and `Solver::resolve_vars` all
-            // reach it only through here; skipping it strands the tail unsubstituted.
             Type::Concatenate(prefix, pspec) => {
                 prefix.visit(f);
                 pspec.visit(f);
@@ -1030,8 +1027,6 @@ impl VisitMut for Type {
             Type::ParamSpec(x) => x.visit_mut(f),
             Type::TypeVarTuple(x) => x.visit_mut(f),
             Type::SpecialForm(x) => x.visit_mut(f),
-            // As in `Visit::recurse`, the ParamSpec tail must be visited or it never gets
-            // substituted or resolved.
             Type::Concatenate(prefix, pspec) => {
                 prefix.visit_mut(f);
                 pspec.visit_mut(f);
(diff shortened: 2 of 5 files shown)


yangdanny97 (approved):

markselby9:
The `asynq` delta in the primer is a correct true positive from this change, not a regression — I verified it with a before/after build.

`asynq` types the bound decorator's `__get__` as roughly `(self: Decorator[_T2, Concatenate[Any, _P2]]) -> Binder[_T2, _P2]`. Before this diff the `Concatenate` tail was never substituted, so the resulting binder's parameter list was left unresolved and calls on it were accepted permissively. After it, the tail resolves: on class-level access `Concatenate[Any, _P2]` strips the leading parameter and binds `_P2` to the remainder, so `binder.asynq(...)` enforces the real arity — hence "Expected 0 positional arguments, got 1" when the remainder is empty. I also confirmed a non-empty remainder is still retained (a correctly-sized call type-checks), so the tail isn't being over-stripped. The previous silence was the bug this fixes.

PR #4419, opened 2026-08-03
fix feature request: recognize modules in string #4345

### LINKED ISSUE #4345: feature request: recognize modules in string
ref https://github.com/astral-sh/ty/issues/3764

asukaminato0721:
# Summary

Fixes #4345

Go-to-definition now resolves module components inside strings such as "accounts.urls".
Clicking accounts opens accounts; clicking urls opens accounts.urls.
Preserves higher-priority `__all__` symbol navigation.
Only activates when the complete string resolves as a module, preventing false positives.

# Test Plan

add test


stroxler (approved):

PR #4429, opened 2026-08-04
Check `__call__` for override consistency like any other method

### LINKED ISSUE #4220: Incorrectly implementing the `__call__` method of a `Protocol` does not produce a type error
### Describe the Bug

As expected, the following code produces a type error
```
class ProtocolA(Protocol):
    def foo(self, s: str):
        ...
class Foo(ProtocolA):
    def foo(self, s: int):
        ...
```
```
ERROR sandbox.py:9:9-12: Class member `Foo.foo` overrides parent class `ProtocolA` in an inconsistent manner [bad-override]
```
However, if the function is `__call__` instead of `foo`, then the code type-checks successfully.
```
class ProtocolB(Protocol):
    def __call__(self, s: str):
        ...
class Bar(ProtocolB):
    def __call__(self, s: int):
        ...
```
This is undesirable, as now runtime errors can be raised if calling `Bar` with a `str` via the protocol.

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAOutWAE64C2ABAC6nEToDmzEjxXHVbMACg1a4AxrijVqUqKjhwxE6bICCACnG5JMqAEpE1ZueaYYYZmFy5tcGFDAAaZnEQfWdE2YsBhEEKSirMAGL2uuqGmn7oAVY2dg5OLu6efOis8QGBwTToisqqegayAELR%2BhrGpgkWScwA%2Bs1SqFBQrY7Obh5ecD65eeZBhCElzBWodNXlUBXD5k2t7Z3daX2ZXDn1I%2BbEJdQgriBkdNZQpISSjFAUAMRipBdgVx4YOATMMuiQ3ABXOioVgQXDocboJ4AZRgMGYAAtWKxiJ4APRo86Xa5CbhomDoNGYaRwNG-f5AkFgwm2ITMVAAN1Q0FQ2FgP3BFOBoPBzFwxB56DgkLIrAR4IAtAyYHQ4NTmABeZiUEAAZkIAEYAEwq6gAbRlDFlAF1qAD0PxBMIYJgJZgIBcpKDpYrmAByCDcdBCGBu6je1gSi4ARwBDptEoA1jBSBLUFIpPBVEq3QB3GboP3oEAAX1O8edMHC0BgFDQWDwRDIuaAA

_No response_

IBlackVoid:
Fixes #4220.

A `Protocol` method overridden with an incompatible signature is reported, but
`__call__` was not, so this passed and then failed at runtime:

```python
class ProtocolB(Protocol):
    def __call__(self, s: str): ...

class Bar(ProtocolB):
    def __call__(self, s: int): ...   # no error
```

### Why it was skipped

`should_check_field_for_override_consistency` skipped `__call__` outright, with a
TODO explaining the reason:

> there are too many classes on typeshed whose `__call__` are marked as
> `def __call__(self, *args: Any, **kwds: Any) -> Any`. If we follow our
> pre-existing subtyping rule, this kind of signature would be non-overridable
> -- any overrider must be able to take ANY arguments which can't be practical.
> We need to either special-case typeshed or special-case callable subtyping.

### The premise no longer holds

The typing spec settles what that signature means:

> If the input signature in a function definition includes both a `*args` and
> `**kwargs` parameter and both are typed as `Any` (explicitly or implicitly
> because it has no annotation), a type checker should treat this as the
> equivalent of `...`.

and `is_subset_params` already implements it:

```rust
match result {
    Err(_)
        if !self.solver.strict_callable_subtyping
            && (params_have_any_args_and_kwargs(l_params)
                || params_have_any_args_and_kwargs(u_params)) =>
    { Ok(()) }
    _ => result,
}
```

`strict_callable_subtyping` is off by default. The case the skip defended against
is handled a layer down, so the skip only suppressed real errors. This takes the
second option the TODO offers — the callable-subtyping side — except that it was
already there, so the fix is to stop compensating for it.

`test_override_dunder_call_gradual_parent` covers this directly, since it is what
makes checking the remaining signatures safe.

### This is not a new rule, it is an exemption ending

A plain method with the identical shape is reported today:

```python
class ParentF:
    def f(self, *args, **params): pass
class ChildF(ParentF):
    def f(self, *args, **params): return 5      # already reported on main

class ParentCall:
    def __call__(self, *args, **params): pass
class ChildCall(ParentCall):
    def __call__(self, *args, **params): return 5   # was silent
```

`__call__` was the only method name exempt from a check every other name gets.

### Real-world differential

CPython 3.12 stdlib plus transitively-imported site-packages (numpy, scipy, pip).
Same corpus, same config, release binaries:

| | main | this |
|---|---|---|
| total diagnostics | 65,598 | 65,612 |
| introduced | — | **14** |
| removed | — | **0** |
| introduced *not* mentioning `__call__` | — | **0** |

No other error kind moved by a single line. The 14 break down as 8
`bad-override` and 6 `bad-override-param-name` — the latter a separately
configurable kind.

Read at the source, they are genuine:

- **4 × `argparse.Action` subclasses** name the third parameter `value` where the
  base declares `values` (`libregrtest/cmdline.py`, `test_argparse.py`).
- **`MySuite(unittest.TestSuite)`** returns `None` where `BaseTestSuite.__call__`
  is declared `-> TestResult`.
- **numpy `_fromnxfunction_args`/`_allargs`** return arrays where the base
  `__call__` has a `pass` body, so it is inferred `-> None`. This is the same
  pattern that is already reported for any non-`__call__` method.

Common patterns stay clean, checked explicitly: a custom `argparse.Action` that
matches the base signature (annotated or not), `unittest.TestCase`,
`contextlib.ContextDecorator`, and metaclass `__call__` in its concrete,
singleton and bare forms.

### Two deliberate behaviours

- **`__call__` written as an annotation rather than a def** is a read-write
  attribute, so it takes the invariance path rather than callable subtyping.
  `__call__: int` on a metaclass now reports, which is correct. This is what
  changes the one existing expectation in `test/protocol.rs`, where
  `__call__: ProxyMethod["make"]` on a metaclass now also reports as an
  inconsistent override — that line already asserted an error for the same
  mistake.
- **Gradual parameters do not excuse a wrong return type.** A child declared
  `(*args, **kw)` still has to return what the base promises, which is the rule
  already applied everywhere else.

### Verification

- `cargo test -p pyrefly` — full suite green.
- The pre-existing `bug`-marked `test_override_dunder_call` documented this exact
  hole; marker removed and expectation updated.
- 5 tests added: the issue's repro, the gradual-form control, a compatible
  override, an `@override`-decorated override, and the original case.
- Determinism — independently built debug and release binaries produced
  byte-identical 65,612-line corpus output.
- Lint clean on all three modified files.
- Diff: `alt/class/class_field.rs` −12, `test/class_overrides.rs` +58,
  `test/protocol.rs` 1 line.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -3702,18 +3702,6 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             return false;
         }
 
-        // TODO(grievejia): In principle we should not really skip `__call__`. But the reality is that
-        // there are too many classes on typeshed whose `__call__` are marked as follows:
-        // ```
-        // def __call__(self, *args: Any, **kwds: Any) -> Any: ...
-        // ```
-        // If we follow our pre-existing subtyping rule, this kind of signature would be non-overridable
-        // -- any overrider must be able to take ANY arguments which can't be practical. We need to either
-        // special-case typeshed or special-case callable subtyping to make `__call__` override check more usable.
-        if field_name == &dunder::CALL {
-            return false;
-        }
-
         // Private attributes should not participate in override checks
         if Ast::is_mangled_attr(field_name) {
             return false;
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -3859,6 +3859,20 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             else {
                 continue;
             };
+            // `__call__` is checked against a Protocol parent only, unless the user
+            // opts in with `@override`. Implementing a callable interface is
+            // ubiquitous - argparse actions, auth handlers, metaclasses - and
+            // treating each one as an override reports a signature mismatch, and a
+            // missing `@override`, on code doing nothing wrong. A Protocol is
+            // different: its members are the contract it exists to state, so an
+            // incompatible `__call__` there is the unsound case, and the parent
+            // signature is written to be implemented rather than inherited.
+            if field_name == &dunder::CALL
+                && !is_explicit_override
+                && !want_field.defining_class.is_protocol()
+            {
+                continue;
+            }
             parent_attr_found = true;
             if want_field.defining_class.is_builtin("object") {
                 parent_attr_is_from_object = true;
--- a/pyrefly/lib/test/class_overrides.rs
+++ b/pyrefly/lib/test/class_overrides.rs
@@ -823,7 +823,11 @@ class C:
     "#,
 );
 
+// Checking `__call__` against every parent reports a missing `@override`, and often a
+// signature mismatch, on the many classes that simply implement a callable interface, so
+// it is checked only against a Protocol parent. See https://github.com/facebook/pyrefly/issues/4220.
 testcase!(
+    bug = "`__call__` inherited from a non-Protocol parent is not checked",
     test_override_dunder_call,
     r#"
 class Base: pass
@@ -833,7 +837,7 @@ class UseBase:
     def __call__(self) -> list[Base]: ...
 
 class UseDerived(UseBase):
-    def __call__(self) -> list[Derived]: ...  # E: Class member `UseDerived.__call__` overrides parent class `UseBase` in an inconsistent manner
+    def __call__(self) -> list[Derived]: ...
     "#,
 );
 
@@ -878,6 +882,29 @@ class UseDerived:
 
 class UseBase(UseDerived):
     def __call__(self, x: Base) -> Derived: ...
+    "#,
+);
+
+// Requiring `@override` on every `__call__` is what makes checking it against all parents
+// unusable: implementing a callable interface is not what the decorator documents, and the
+// demand lands on argparse actions, auth handlers and metaclasses throughout the ecosystem.
+testcase!(
+    test_missing_override_decorator_dunder_call,
+    TestEnv::new().enable_missing_override_decorator_error(),
+    r#"
+from typing import Protocol
+
+class Base:
+    def __call__(self, x: int) -> None: ...
+
+class Concrete(Base):
+    def __call__(self, x: int) -> None: ...
+
+class P(Protocol):
+    def __call__(self, x: int) -> None: ...
+
+class Impl(P):
+    def __call__(self, x: int) -> None: ...  # E: is missing an `@override` decorator
     "#,
 );
 
--- a/pyrefly/lib/test/class_overrides.rs
+++ b/pyrefly/lib/test/class_overrides.rs
@@ -824,7 +824,6 @@ class C:
 );
 
 testcase!(
-    bug = "We currently skip checking overrides of `__call__`, which is a soundness hole",
     test_override_dunder_call,
     r#"
 class Base: pass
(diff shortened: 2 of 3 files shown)


rchen152:
Hmm, that's a lot of new errors.

IBlackVoid:
You're right, and my pre-submit measurement understated it. I diffed a stdlib + site-packages corpus and reported 14 new diagnostics, but I ran it with the default config, where `missing-override-decorator` is `Severity::Ignore`. That hid the bulk of the effect. Sorry — the number I put in the description is not the number that matters.

The primer run breaks down as:

| kind | new |
|---|---:|
| `missing-override-decorator` | 396 |
| `bad-override` | 47 |
| `bad-override-param-name` | 13 |

So 87% is not the soundness check at all. `MissingOverrideDecorator` is emitted from `check_consistent_override_for_field` after the same gate, so dropping `__call__` from that gate turned on the style check along with it. Requiring `@override` on `__call__` is noise: implementing a callable interface is not what that decorator is meant to document, and 396 hits across 58 projects is a fair verdict on that.

The more useful thing the primer shows is what the remaining 60 are attached to. The parents are `argparse.Action`, `requests.AuthBase`, pytest's `MarkDecorator`, scipy's `MultiUFunc`, `poly1d`, mitmproxy's `View` — concrete classes and ABCs. Hardly any are Protocols, and #4220 is specifically about a Protocol:

```python
class ProtocolB(Protocol):
    def __call__(self, s: str): ...

class Bar(ProtocolB):
    def __call__(self, s: int): ...   # no error
```

So I aimed too wide. I removed the carve-out entirely on the reasoning that `__call__` should not be special, when the report only needs it to not be special *when the parent is a Protocol* — where implementing the member is the whole contract and getting the signature wrong is a real hole.

I'd like to push a version that checks `__call__` only when the parent is a Protocol. That fixes the case in the issue, and should take the primer delta to near zero, since almost nothing in the list above inherits `__call__` from one.

Happy to do that, or to close this if you would rather settle `__call__` override policy separately — it clearly deserves its own decision rather than riding along with a bug fix. Either way I'll rerun the primer with `missing-override-decorator` enabled before asking you to look again.

IBlackVoid:
Pushed the narrowing. `__call__` is now checked only when the parent that declares it is a Protocol, or when the child writes `@override` explicitly.

The skip goes in before `parent_attr_found` is set, which is what matters here: that flag drives the missing-`@override` check as well as the consistency check, so both are suppressed for non-Protocol parents. Keeping the `@override` opt-in also stops `is_explicit_override && !parent_attr_found` from wrongly firing "marked as an override, but no parent class has a matching attribute".

I rebuilt the measurement first, since the number I gave you originally came from a corpus run using default severities and could not see `missing-override-decorator` at all. The corpus now runs with all 57 diagnostics that default to `Ignore` or `Warn` explicitly enabled, leaving `strict_callable_subtyping` alone so the gradual-form escape hatch still applies. CPython 3.12 stdlib plus numpy, scipy and pip, release binaries built from the same tree:

| | vs main |
|---|---|
| previous version (checked all parents) | **+67** introduced |
| this version (Protocol parents only) | **0** introduced, 0 removed, byte-identical output |

The 67 it drops are 53 `missing-override-decorator`, 8 `bad-override`, 6 `bad-override-param-name` — every one from a non-Protocol parent, which is why the delta goes to zero rather than merely down.

Also checked, on the patterns from your primer run: a custom `argparse.Action` renaming `values`, a `requests`-style `AuthBase` implementation, a metaclass `__call__` over `type`, and a concrete recorder base all stay quiet, while the Protocol case in the issue reports, and `@override` on a concrete parent still reports.

Verification: `cargo test -p pyrefly` 7643 passed / 0 failed; conformance unchanged (the 420 changed lines are filename keys from Windows path separators, identical once the prefix is normalised); lint and format clean; four consecutive corpus runs byte-identical.

One thing worth saying plainly rather than leaving you to find it: with the narrowing, this finds nothing in the wild on that corpus. It closes a soundness hole — the case in the issue goes from silent to reported — but it does not surface existing bugs, because inheriting `__call__` from a Protocol is rare. If you would rather have the broader behaviour behind a separate error code, or would rather not take this at all, both are reasonable and I am happy either way.

`__call__` inherited from a non-Protocol parent stays unchecked, and the test for it keeps its `bug` marker rather than quietly dropping the case.

rchen152 (approved):

rchen152:
Thanks! I spot-checked the remaining primer delta and didn't see any more false positives. I've kicked off a merge.

As a gentle reminder, our [contributing guidelines](https://github.com/facebook/pyrefly/blob/main/CONTRIBUTING.md#ai-generated-code) require that AI agents submitting PRs disclose their status as an agent in the PR description. We allow agents to contribute to Pyrefly, but if a human is driving the PR, we ask that you understand any AI-generated code you are submitting so you can engage with PR review in your own words. Otherwise, the PR is effectively agent-driven, and we require this to be disclosed upfront.

IBlackVoid:
Thanks for merging.

Just to be clear on the guidelines: I'm not an agent. I picked the issue and wrote the fix myself, and used an AI assistant to test and verify it. The reasoning and the code are mine.

rchen152:
Thanks for the clarification. Using AI assistance for testing and verification is totally okay and within project guidelines; what we'd request you not do is directly put AI output as your own comments without disclaimer, since that makes it difficult for us to tell if there's a person interacting with review feedback. Apologies if my wording was overly harsh, and thanks for the PR.

PR #4433, opened 2026-08-04
fix Better syntax highlighting for strings #4351

### LINKED ISSUE #4351: Better syntax highlighting for strings
ref https://github.com/astral-sh/ty/issues/2949

asukaminato0721:
# Summary

Fixes #4351

Added semantic highlighting for string prefixes/types, escape sequences, f-string specifiers, and printf placeholders, also added VS Code theme-scope mappings.

# Test Plan

add test

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -89,6 +89,36 @@
             {
                 "language": "python",
                 "scopes": {
+                    "string.byteString": [
+                        "string.quoted.binary.single.python",
+                        "string.quoted.binary.multi.python"
+                    ],
+                    "string.byteString.rawString": [
+                        "string.quoted.raw.binary.single.python",
+                        "string.quoted.raw.binary.multi.python"
+                    ],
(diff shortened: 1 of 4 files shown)


rchen152 (approved):

grievejia:
@asukaminato0721 Thanks for putting this together! There’s a useful core here. I’ve split out and kept the metadata-driven string-kind classification: prefixes and byte/raw/f/t-string modifiers, with coverage for multiline and Python 3.14 template strings.

I did not include content-level highlighting because it independently parses escapes, formatting placeholders, and f/t-string structure using regexes and manual state. That is difficult to keep aligned with Python’s grammar, especially for hyphenated `\N{...}` names, `%` false positives in prose, raw and byte-string escape rules, and multiline or nested f/t-string format specifications.

I’d welcome follow-up PRs for this functionality, so we can evaluate the parsing approach and edge cases in isolation.

PR #4435, opened 2026-08-04
fix display the current pyrefly version in the sandbox #3802

### LINKED ISSUE #3802: display the current pyrefly version in the sandbox
as far as i can tell, the sandbox seems to be using a prerelease version of pyrefly, but not the latest commit in main.

a version selector would also be nice

asukaminato0721:
# Summary

Fixes #3802

export version from cargo env into wasm and insert into tsx

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/lib.rs
+++ b/pyrefly/lib/lib.rs
@@ -56,6 +56,9 @@ mod tsp;
 
 pub(crate) use compat::*;
 
+/// The version of this Pyrefly build.
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+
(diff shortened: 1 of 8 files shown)


yangdanny97:
Jia had a concern that this could still cause confusion, since the sandbox saying v.1.2.0 doesn't mean it's 1.2.0, rather it is 1.2.0 + some commits (potentially several hundred)

so something might work in the sandbox but not work if a user downloads 1.2.0

it might make sense to show a commit timestamp or commit hash instead, set when the website update script runs. for local builds we can show a placeholder or nothing

PR #4457, opened 2026-08-05
fix: don't panic on a non-generic classproperty in get_attributes

### LINKED ISSUE #4453: panic: `Option::unwrap()` on `None` in Query::get_attributes for a non-generic `classproperty` class
### Summary

`Query::get_attributes` panics when a class named `classproperty` or `cached_classproperty` has **no type arguments**. The match arm keys on the class *name* alone, then assumes type arguments are present:

https://github.com/facebook/pyrefly/blob/d5ce6d8ed377823cbc979f9e46aebc489fa1906b/pyrefly/lib/query.rs#L1144-L1148

```rust
Type::ClassType(c)
    if c.name() == "classproperty" || c.name() == "cached_classproperty" =>
{
    let result_ty = c.targs().as_slice().first().unwrap();
    (Some(String::from("property")), result_ty)
}
```

A plain, non-generic `classproperty` — a common Python idiom for combining `classmethod` and `property` — reaches this arm with empty `targs`, and `.first().unwrap()` panics.

```
thread '...' panicked at pyrefly/lib/query.rs:1147:66:
called `Option::unwrap()` on a `None` value
```

### Reproduction

Verified against `main` @ `d5ce6d8`. Python input:

```python
class classproperty:
    def __init__(self, f):
        self.f = f
    def __get__(self, obj, owner):
        return self.f(owner)

class Config:
    @classproperty
    def name(cls) -> str:
        return "cfg"
```

As a test in `pyrefly/lib/test/query.rs`:

```rust
#[test]
fn repro_non_generic_classproperty_panics() {
    let tdir = TempDir::new().unwrap();
    let file_path = tdir.path().join("main.py");
    let code = r#"<the Python above>"#;
    fs_anyhow::write(&file_path, code).unwrap();

    let query = create_query();
    let module_name = ModuleName::from_str("main");
    let path = ModulePath::filesystem(file_path.clone());
    let errors = query.add_files(vec![(module_name, path.clone())]);
    assert!(errors.is_empty(), "unexpected errors: {errors:?}");

    // panics at query.rs:1147
    let _ = query.get_attributes(module_name, path, "Config");
}
```

**Control:** changing the class to `class classproperty[T, R]:` makes the same test pass, isolating the failure to the no-type-arguments case. `pyrefly check` on the same file reports 0 errors — this affects the `Query::get_attributes` API rather than the CLI check path.

### Note on verification

Found while evaluating Rust verification tooling. Working this through the [midas-lex](https://crates.io/crates/midas-lex) tool identifies the spec owner here as a *precondition*: indexing element 0 requires the slice to be non-empty, and matching on the class name does not establish that. Modelling the same obligation in [Verus](https://github.com/verus-lang/verus) reports `precondition not satisfied` on the indexing as written; the identical body verifies once `targs.len() > 0` is available, or with an explicit length check (`3 verified, 1 errors`).

### Suggested fix

Handle the no-type-arguments case rather than indexing unconditionally — for example falling back to `ty` so the class is treated as a plain attribute:

```rust
Type::ClassType(c)
    if c.name() == "classproperty" || c.name() == "cached_classproperty" =>
{
    match c.targs().as_slice().first() {
        Some(result_ty) => (Some(String::from("property")), result_ty),
        None => (Some(String::from("property")), ty),
    }
}
```

Happy to send a PR if that's welcome — noting CONTRIBUTING asks to discuss approach on an issue first.

renz011tzar:
Fixes #4453.

`get_kind_and_field_type` matched `classproperty` / `cached_classproperty` on the class *name* alone, then unwrapped the first type argument:

```rust
Type::ClassType(c)
    if c.name() == "classproperty" || c.name() == "cached_classproperty" =>
{
    let result_ty = c.targs().as_slice().first().unwrap();
```

A class with that name but **no type arguments** reaches this arm with empty `targs`, so `get_attributes` panics.

### Change

Per @yangdanny97's suggestion in the issue, the no-type-arguments case now behaves as `(None, ty)`. Rather than adding a nested match, the check moves into the match guard, so such a class simply doesn't match this arm and falls through to the existing `_ => (None, ty)`:

```rust
Type::ClassType(c)
    if (c.name() == "classproperty" || c.name() == "cached_classproperty")
        && let Some(result_ty) = c.targs().as_slice().first() =>
{
    (Some(String::from("property")), result_ty)
}
```

### Testing

Adds `test_get_attributes_non_generic_classproperty`, following the existing `TempDir` + `create_query()` pattern in `test/query.rs`. It checks out a module defining a plain `class classproperty:` used as a decorator, and asserts `get_attributes` returns the attribute rather than panicking.

- The test **fails without** the fix (`panicked at pyrefly/lib/query.rs:1147:66: called Option::unwrap() on a None value`) and passes with it.
- `cargo test -p pyrefly --lib test::query` — 7 passed, 0 failed.
- `cargo fmt --check` clean.

### DIFF
--- a/pyrefly/lib/query.rs
+++ b/pyrefly/lib/query.rs
@@ -1142,9 +1142,9 @@ impl Query {
                     (Some(String::from("property")), ty)
                 }
                 Type::ClassType(c)
-                    if c.name() == "classproperty" || c.name() == "cached_classproperty" =>
+                    if (c.name() == "classproperty" || c.name() == "cached_classproperty")
+                        && let Some(result_ty) = c.targs().as_slice().first() =>
                 {
-                    let result_ty = c.targs().as_slice().first().unwrap();
                     (Some(String::from("property")), result_ty)
                 }
                 _ => (None, ty),
--- a/pyrefly/lib/test/query.rs
+++ b/pyrefly/lib/test/query.rs
@@ -424,3 +424,38 @@ class C:
         "Expected two constructor callees for C(), got: {callees:?}"
     );
 }
+
+#[test]
+fn test_get_attributes_non_generic_classproperty() {
+    let tdir = TempDir::new().unwrap();
+    let file_path = tdir.path().join("main.py");
+    // A non-generic class named `classproperty` has no type arguments, so the
+    // `classproperty` match arm must not assume one is present.
+    let code = r#"class classproperty:
+    def __init__(self, f):
+        self.f = f
+    def __get__(self, obj, owner):
+        return self.f(owner)
+
+class Config:
+    @classproperty
+    def name(cls) -> str:
+        return "cfg"
+"#;
+    fs_anyhow::write(&file_path, code).unwrap();
+
+    let query = create_query();
+    let module_name = ModuleName::from_str("main");
+    let path = ModulePath::filesystem(file_path.clone());
+    let errors = query.add_files(vec![(module_name, path.clone())]);
+    assert!(errors.is_empty(), "Unexpected errors: {:?}", errors);
+
+    let attributes = query
+        .get_attributes(module_name, path, "Config")
+        .expect("expected attributes for Config");
+    assert!(
+        attributes.iter().any(|a| a.name == "name"),
+        "expected an attribute named `name`"
+    );
+}
+


yangdanny97 (approved):

PR #4458, opened 2026-08-05
enhance lsp initialisation (#4454)

### LINKED ISSUE #4454: lsp initialisation improvements
https://pyrefly.org/en/docs/IDE/#lsp-initializationoptions gives an example object, but doesn't document what the fields are (they are different to the standard options). i can see that it has an `extraPaths` field, but this differs from the standard name of this field "searchPath", should it be named "extraSearchPaths"?

also, this object is missing a way to exclude files from resolution: we are integrating pyrefly with pycharm, and there is currently no way for us to map "Excluded" directories

if there are no objections, i can make a pr for these changes

tldr:
- improve docs for lsp initialization options
- add a new field for explicitly excluded paths (resolution and type checking)

KotlinIsland:
- better docs
- added `extraSearchPaths` and `extraProjectExcludes`

resolves #4454


yangdanny97 (approved):

PR #4476, opened 2026-08-08
Fix descriptor semantics for initialized annotations

### LINKED ISSUE #4471: Descriptor semantics are applied to annotated instance attributes based on type rather than class attribute
### Describe the Bug

Pyrefly (tested with 1.3.0.dev1) reports an error when assigning to an instance attribute whose annotated type implements `__get__`, even though no descriptor exists on the class.

```
class DescriptorLike:
    def __get__(self, instance: object, owner: type[object] | None = None) -> "DescriptorLike":
        return self

class Implementation:
    value: DescriptorLike

    def __init__(self) -> None:
        self.value = DescriptorLike()
```

Pyrefly reports:
```
ERROR Attribute value of class Implementation is a read-only descriptor with no __set__ and cannot be set [read-only]
```

However, `Implementation.value` is only a type annotation. There is no class attribute initialized with a `DescriptorLike` instance, so the descriptor protocol is never involved. The assignment in `__init__` simply creates an ordinary instance attribute.

Mypy accepts this example without error.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.hZTPDsIgDIdfhbiLXux9b2DiwbvzQGaDxAgEUF_fwoZbdX-uTdoS-v2-OWX2-cLQeu3IDkd954lROASmgFRYJTu9SZ91ci6eu9plAuBmwxd8bz3CL20YsntIyX2giTLdikf_Z9YokSzfq4Hmc7a7_0NX4tQVRP9npDgiRD1JpPSsfWMqciiKW4wu1ACl3XoFaOBq2wCsA0RiWr6kznrm44TNYAYau0CY8xgw5hKBmhFZuvsH

_No response_

patrickswedish:
## Summary

- avoid applying descriptor protocol semantics to annotated fields initialized in recognized instance methods
- preserve the existing class-field classification and class-level visibility behavior for unrelated annotations
- add regression coverage for annotation plus `__init__` assignment while retaining annotation-only and method-only controls

## Root cause

`DeclaredByAnnotation` already records `initialized_in_recognized_method`, but descriptor detection did not consult that signal. An annotation whose type implements `__get__` was therefore treated as a class-installed descriptor even when `self.field` was initialized in `__init__`.

The fix uses that signal only to skip descriptor construction for the annotated-and-method-initialized case. It intentionally leaves the field's existing initialization classification unchanged, avoiding unrelated class-visibility and constructor behavior changes.

Fixes #4471

## Validation

- `cargo fmt --all -- --check` — passed
- `python test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` — formatting passed; Rust lint was blocked locally because the Windows host has no MSVC `link.exe`
- `git diff --check` — passed
- repository Linux, Windows, macOS, and compatibility CI rerunning on the narrowed patch

AI agent disclosure: OpenAI Codex prepared this PR; the contributor reviewed and understands the changes.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1956,6 +1956,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // Identify whether this is a descriptor. Construct the stored descriptor only after
         // forcing the field type so its class cannot retain solver variables.
         let mut descriptor_methods = None;
+        let is_annotation_initialized_in_method = match field_definition {
+            ClassFieldDefinition::DeclaredByAnnotation {
+                initialized_in_recognized_method,
+                ..
+            } => *initialized_in_recognized_method,
+            _ => false,
+        };
         // Descriptor semantics apply when the field is modeled as class-level:
         // either by a class-body definition, by `Magic` for stub/interface
         // declarations where the runtime initializer is omitted, or by an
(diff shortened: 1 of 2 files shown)


yangdanny97 (approved):

patrickswedish:
Thanks for reviewing and merging, @yangdanny97! Glad to contribute.

PR #4479, opened 2026-08-09
fix: preserve known exports in dynamic __all__

### LINKED ISSUE #4424: `from torch import Tensor` incorrectly marked as `implicit-reexport`.
### Describe the Bug

```
from torch import Tensor
```

`pyrefly` 1.2.0 produces the following:
```
$ pyrefly check --preset all example.py
ERROR `Tensor` is not exported from module `torch` [implicit-reexport]
 --> example.py:1:19
  |
1 | from torch import Tensor
  |                   ^^^^^^
  |
  `Tensor` is imported by `torch` but not re-exported (via `import ... as Tensor`, `__all__`, or a wildcard import)
 INFO 1 error
```

`torch/__init__.py` does include `Tensor`  in `__all__` at the beginning of the file:

```
__all__ = [
    ...
    "Tensor",
    ...
]
```

But then has the following import in the middle of the file (1900 lines down):
```
from torch._tensor import Tensor  # usort: skip
```

My guess is that the order of `__all__` and this import is causing `pyrefly` to incorrectly mark it as implicit.

### Sandbox Link

No sandbox due to torch

_No response_

nightcityblade:
# Summary

Preserve statically known `__all__` names when a later dynamic `append` or `extend` cannot be resolved. This prevents explicit re-exports such as `torch.Tensor` from being reported as implicit while keeping the conservative fallback for the unresolved portion of `__all__`.

Fixes #4424

# Test Plan

- `cargo test -p pyrefly test::imports::` (126 passed)
- `cargo clippy --release -p pyrefly --all-targets` (passed; two unrelated existing warnings)
- `cargo fmt -- --check`
- Built Pyrefly and checked `from torch import Tensor` against torch 2.10.0 with `--preset all` (0 errors)

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -806,19 +806,15 @@ impl DefinitionsBuilder {
                                 self.inner.dunder_all.entries.extend(entries);
                             }
                             None => {
-                                self.inner.dunder_all = DunderAll {
-                                    kind: DunderAllKind::Unresolvable(arguments.args[0].range()),
-                                    entries: Vec::new(),
-                                };
+                                self.inner.dunder_all.kind =
+                                    DunderAllKind::Unresolvable(arguments.args[0].range());
                             }
                         },
                         "append" => match DunderAllEntry::as_item(&arguments.args[0]) {
                             Some(entry) => self.inner.dunder_all.entries.push(entry),
                             None => {
-                                self.inner.dunder_all = DunderAll {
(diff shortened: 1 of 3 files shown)


dillydill123:
@nightcityblade Thanks for putting up a fix. You will need to sign the CLA. Do you plan to do so? Otherwise please remove your claim on the issue. Thanks!

nightcityblade:
Thanks for the reminder. I can't complete a legal CLA from this automated contributor session, so I'm closing the PR to avoid blocking the issue. Sorry for the churn.

PR #4483, opened 2026-08-09
Centralize sanitization of out-of-scope typevars in typevar defaults (#4483)

### LINKED ISSUE #4298: TODO: deal with code duplication in get_tparam_default

rchen152:
Summary:

Pyrefly had almost-duplicate code in `AnswersSolver::get_tparam_default` and `Solver::finish_class_targs` that did two things:

1. Replace out-of-scope legacy type variables in typevar defaults with gradual fallbacks
1. Substitute quantifieds in typevar defaults with the appropriate values.

This diff moves (1) to `validated_tparams`, so tparams are sanitized exactly once, before being stored. (2) still happens in the same places as before, but the code is much simpler now that it doesn't need to do two things at once.

Fixes https://github.com/facebook/pyrefly/issues/4298.

Reviewed By: grievejia

Differential Revision: D115350225

### DIFF
--- a/crates/pyrefly_types/src/types.rs
+++ b/crates/pyrefly_types/src/types.rs
@@ -284,6 +284,11 @@ impl TArgs {
         &mut Arc::make_mut(&mut self.0).1
     }
 
+    pub fn split_mut(&mut self) -> (&TParams, &mut [Type]) {
+        let inner = Arc::make_mut(&mut self.0);
+        (&inner.0, &mut inner.1)
+    }
+
     pub fn is_empty(&self) -> bool {
         self.0.1.is_empty()
     }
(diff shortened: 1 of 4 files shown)


grievejia (approved):

PR #4487, opened 2026-08-09
Do not report pytest's __tracebackhide__ as unused

### LINKED ISSUE #4484: `Variable __tracebackhide__ is unused` is incorrectly reported as an error
### Describe the Bug

When running pyrefly as an LSP I get the following error:

```
Variable `__tracebackhide__` is unused [unused-variable]
```

`__tracebackhide__` is used in pytest to hide helper functions in tracebacks if a test fails due to an assert raising inside the helper function. This is detailed in the pytest docs [here](https://doc.pytest.org/en/latest/example/simple.html#writing-well-integrated-assertion-helpers).

Example:

```python
# content of test_checkconfig.py
import pytest


def checkconfig(x):
    __tracebackhide__ = True
    if not hasattr(x, "config"):
        pytest.fail(f"not configured: {x}")


def test_something():
    checkconfig(42)
```

It would be good if pyrefly doesn't report the above `__tracebackhide__ is unused` error as it isn't really an error as it's supported by pytest.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.w1VkgqIwOSM1OTs5H-j3dI0KWMqIjy8pSkxOTUpMzs7ITAFHFziUQREDSb8gm-OL80HZAFiaaMA0IptmYgSMIcwAVFYIgAgoQN0CLDpA6kuBBRQwRPVi8pSBZVOqQkZJSUGxlb4-THt-Ubp-ap5-Sn5ysT6KDn0FUFpJLEvMBBd7qMYp5IMjvBhoLCUxpw6JOHV88aYOijagCrxRAQA

_No response_

cakeni:
﻿# Summary

- Skip unused-variable diagnostics for pytest's `__tracebackhide__` local, which pytest reads implicitly when formatting tracebacks.
- Add a regression test that also confirms an ordinary unused local is still reported.

Fixes #4484

# Test Plan

- `cargo test -p pyrefly test_pytest_tracebackhide_not_reported_as_unused --lib`
- `cargo test -p pyrefly 'test::lsp::diagnostic::' --lib` (27 passed)
- `python test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema`

The Cargo tests used local path overrides for the repository's two Git-pinned dependencies because direct GitHub fetches were unavailable; the overrides were downloaded from the pinned revisions and were not committed.

# AI assistance

AI assistance was used during investigation or implementation.


fangyi-zhou:
Shall we check whether pytest is imported on the file?

cakeni:
Done — the exemption now only applies when the file imports `pytest`. I also added a regression test to make sure `__tracebackhide__` is still reported as unused when pytest isn’t imported.

PR #4489, opened 2026-08-09
Significantly optimise checking of repeatedly-instantiated `TypedDict` objects

alexander-beedie:
# Summary

Identified a somewhat niche hotspot relating to `TypedDict`; with this PR in place, type-checking `pydantic` gets **~2x faster** (measured ~60% less CPU time), and two other packages get a ~6-10% improvement. Pretty much a no-op otherwise.

### Problem

Instantiating `TypedDict` fields rebuilt a full `SmallMap` of field types on every call, with each rebuild deep-copying every field type; callers that only iterate paid that cost repeatedly.

### Solution

* Cache the instantiated field map per `Solver`, in such a way that repeat calls share one allocation (instead of re-cloning each field's type tree).
* Only cache what is context-independent (mirrors the existing `protocol_cache` rule of caching only Var-free types).
* Never cache an _incomplete_ map (field lookup failures are now classified: see `FieldOutcome`).
* Fixed `InitVar` on a `TypedDict` member (it's not valid and blocked caching).
* Don't clone `ClassField` just to read it.

# Test Plan

All existing unit tests pass without modification.
Several new unit tests added.

# Benchmarks[^1]

**Huge** positive impact on `pydantic` (the `pydantic_core.core_schema` module is an absolute _wall_ of `TypedDict`) 🚀 

The speedup tracks with repeated instantiation of the same `TypedDict`, not the overall `TypedDict` count, and it seems that's not a common pattern; only a couple of packages show much benefit (aside from `pydantic` the only other solid gains I spotted are for `numpy` and `narwhals`, which both get a bit faster).

### CPU time (user+sys)

My machine was a bit contended, so I recorded CPU-time instead of wall-clock as it was much more stable (measurements are standard multithreaded runs):

| project | before | after | Δ |
|:---|---:|---:|---:|
| pydantic |  3.71s | 1.53s | **−58.7%** |
| numpy | 12.01s | 11.08s | **−6.9%** |
| narwhals[^2] | 2.36s | 2.18s | **−7.9%** |

[^1]: Test machine: Apple Silicon M3 Max (16 cores).
[^2]: @MarcoGorelli, "you're welcome" 🤣

### DIFF
--- a/crates/pyrefly_types/src/typed_dict.rs
+++ b/crates/pyrefly_types/src/typed_dict.rs
@@ -5,8 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-use std::sync::LazyLock;
-
 use pyrefly_derive::TypeEq;
 use pyrefly_derive::Visit;
 use pyrefly_derive::VisitMut;
@@ -107,10 +105,9 @@ pub enum TypedDict {
     Anonymous(Box<AnonymousTypedDictInner>),
 }
 
-// When we get the name of a class-based typed dict we borrow the class's name, so we need
-// a name that we can borrow for anonymous typed dicts
-// This is a lazily initialized value, Name::new normally can't be used as the RHS of a static
-pub static ANONYMOUS_TYPED_DICT: LazyLock<Name> = LazyLock::new(|| Name::new("<anonymous>"));
+// When we get the name of a class-based typed dict we borrow the class's name,
+// so we need a name that we can borrow for anonymous typed dicts.
+pub static ANONYMOUS_TYPED_DICT: Name = Name::new_static("<anonymous>");
 
 impl TypedDict {
     pub fn new(class: Class, args: TArgs) -> Self {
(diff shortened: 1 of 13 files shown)


yangdanny97:
@samwgoldman

yangdanny97:
TypedDict stuff would normally be under my wheelhouse but I don't have any capacity in the near future so I'll let Sam handle this

alexander-beedie:
Rebased/resolved 👍

alexander-beedie:
@samwgoldman rebased/resolved (again, lol)
If you have a chance, it's currently in a good state with "main" 😄

samwgoldman:
I spent some time looking into the changes here, and I think we can achieve the same (if not better) performance improvement while also making the implementation of TypedDict fields simpler / more consistent with other class fields.

Class fields are always stored in "uninstantiated" form. Any type parameters (incl. Self) are un-substituted. Instead, we substitute when reading a class field's type.

TypedDict fields, on the other hand, are always created in a fully-instantiated form. If we access any field, construct the full map and we substitute in type args and the self type. Your optimization notes that this instantiation makes sharing difficult -- `MyTypedDict[int]` and `MyTypedDict[str]` can no longer share field maps because the types are different.

So I think we should make TypedDict fields more like class fields, where we store (and cache) the generic form. This cache can even be shared across modules using our binding key-value tables, just as class fields are shared.

For classes, we instantiate lazily when looking up a field. In many cases where we load the typed dict field map, we don't need all of the fields' types at all. We might just need their names, and maybe whether they are required.

Based on your work here, I have a bit of local WIP to store TypedDict fields in a the way described above and it does seem to recover 100% of the perf improvement here in a slightly simpler form. I'll let you know how it shapes up, but it might supersede the changes here. Regardless, thank you for finding this amazing perf win!

alexander-beedie:
I've rebased/resolved again to keep things current, but if you can land a cleaner superseding commit then that's great -- feel free to close if so :)

PR #4490, opened 2026-08-09
Discover Pixi default Python environments

### LINKED ISSUE #4432: Support `pixi` Python environment discovery
### Describe the Bug

Fresh installed Pyrefly, but it reports:
```
Cannot find module `torch`
  Looked in these locations:
  Fallback search path (guessed from importing file with heuristics): ["e:\\Sandbox\\#other models", "e:\\Sandbox", "e:\\"]
  Site package path queried from interpreter: ["C:\\Program Files\\Python314\\DLLs", "C:\\Program Files\\Python314", "C:\\Users\\user\\AppData\\Roaming\\Python\\Python314\\site-packages", "C:\\Program Files\\Python314\\Lib\\site-packages", "C:\\Program Files\\Python314\\Lib\\site-packages\\win32", "C:\\Program Files\\Python314\\Lib\\site-packages\\win32\\lib", "C:\\Program Files\\Python314\\Lib\\site-packages\\Pythonwin"]Pyreflymissing-import

torch: Module[torch]
```
Image

The default Pylance has no this issue. That disabling Pylance didn't help.

pixi is a replacement of conda, and much faster than conda, which is very slow that costs minutes to init before add a package.
pixi has the trend override conda, because of speed, pyrefly should support pixi fast too.

Related: https://github.com/facebook/pyrefly/issues/166 which only supported `Venv` and `Conda`.

Additional:
It seems this extension overwrite the language server of vscode even the selected it as Pylance, why dont make a option in vscode so users can switch it just like switch to `Jedi` and `None` already in it?

### Sandbox Link

_No response_

_No response_

WilliamK112:
# Summary

- Discover Python interpreters installed in `.pixi/envs/default` when no explicit or active interpreter is configured.
- Search ancestor workspaces and support Pixi's POSIX and Windows interpreter layouts.
- Search one directory at a time so the nearest project environment wins; venv takes precedence over Pixi only within the same directory. Activated environments and explicit CLI/config/LSP choices remain unchanged.
- Ignore named-only Pixi environments because there is no unambiguous inactive environment to select.

Fixes #4432

# Test Plan

- Added nested-project regressions: child Pixi vs. parent venv, child venv vs. parent Pixi, plus the existing same-directory preference test. The child-Pixi test failed before the fix.
- `env -u CONDA_PREFIX -u VIRTUAL_ENV CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p pyrefly_config -q` (277 passed after merging main at 2257162db)
- `CARGO_PROFILE_DEV_DEBUG=0 CARGO_INCREMENTAL=0 python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` (passed after the mainline merge)
- `git diff --check`

# AI Disclosure

I am an AI agent working under the WilliamK112 account. I inspected the environment-discovery path, implemented the change, and validated it with the tests above.


NathanTempest (changes requested):
Thanks for picking this up @WilliamK112  the design is right and it does fix the reported case. One thing before it lands.

find_project_interpreter is venv::find(start_path).or_else(|| pixi::find(start_path)), and both functions do their own full ancestor walk. So the precedence isn't "venv before pixi", it's "a venv anywhere above you beats a pixi environment in your own project". With /work/.venv and /work/proj/.pixi/envs/default, checking /work/proj/main.py resolves to /work/.venv/bin/python3 and pixi::find never runs. Confirmed this with pyrefly dump-config.

test_find_project_interpreter_prefers_venv puts both environments under the same root, the one arrangement where this and a correct per-directory search agree, so green CI doesn't cover it.

Not a regression, since only venv is searched today, but it's a case #4432 covers. Suggest one ancestor walk trying venv then pixi at each level, so the nearest wins and venv still leads within a directory needs find_in_root and find_in_workspace crate-visible, plus a test with pixi in a child and a venv in a parent.

Minor: find_in_workspace never reads the pixi.toml the fixtures create, so a stale .pixi directory still matches; and detached-environments moves .pixi/envs out of the workspace, worth a note in the doc comment.

@connernilsen  could you take a look at the precedence call.

WilliamK112:
Thanks for catching this. Fixed in af4aa206f and refreshed against current main in 8e9ba4b60: project discovery now performs one ancestor walk, trying venv and then Pixi at each directory. The existing layout tests use that shared search path, and the same-directory venv preference is retained.

The new child-Pixi/parent-venv regression failed with the previous implementation and now passes. I also added the opposite nesting case. All 277 pyrefly_config tests pass with the shell's active Conda/venv variables cleared, and the required format/lint wrapper passes (including after the mainline merge).

The Pixi doc comment now explicitly states that this fallback only inspects the standard .pixi/envs/default layout: it does not validate the manifest or resolve detached environments. I have not added manifest parsing or changed active/explicit interpreter precedence.

The mainline merge preserves the newly tightened canonical venv interpreter paths and adjusts the nested fixtures accordingly. No upstream venv behavior was reverted.

WilliamK112:
@NathanTempest A quick follow-up when the review queue allows: the requested per-directory precedence fix and both cross-directory regression cases are on head `8e9ba4b60`, with all 277 `pyrefly_config` tests passing. There have been no additional changes since that validated head. Could you re-review when convenient?

PR #4498, opened 2026-08-10
fix diagnostic squiggles for unsaved files (#4498)

### LINKED ISSUE #4497: extension: diagnostics for untitled files are published on the wrong URI
Hi! I noticed that when a client opens an untitled document (for example `untitled:Untitled-1` in VS Code), Pyrefly publishes diagnostics for it on a synthetic `file://` URI instead of the URI the client opened. The editor shows the diagnostic in the Problems panel, but it never appears as a squiggle in the editor. When the user saves the file, the squiggle appears.

### Steps to reproduce

1. Open VS Code with the Pyrefly extension (I'm on version 1.1.1).
2. Create a new untitled Python file. Don't save it.
3. Type `import some_module_that_is_not_installed`.
4. Look at the Problems panel and the editor.
5. Save the file and look again.

### Expected behavior

A `missing-import` squiggle appears under the import in the editor during steps 4 and 5.

### Actual behavior

During step 4, the Problems panel shows the `missing-import` diagnostic, but there is no squiggle in the editor. The `publishDiagnostics` notification targets a URI like:

```
file:///__pyrefly_virtual__/untitled/Untitled-1-0.py
```

The client opened `untitled:Untitled-1`, so it does not attach the diagnostic to the open editor. But this goes back to expected during step 5.

-----

I think this is because of: 

https://github.com/facebook/pyrefly/blob/c6481e262d0d559157078770655da214ee85ae7b/pyrefly/lib/lsp/non_wasm/server.rs#L1733

producing the synthetic `file://` URI instead of the original client URI.

yangdanny97:
Summary:

fixes https://github.com/facebook/pyrefly/issues/4497

when working on an unsaved file, previously squiggles didn't show up in the file itself, only in the problems panel.

Differential Revision: D115468974

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -1730,9 +1730,13 @@ impl Server {
             } else {
                 let path = path.absolutize();
                 let version = version_info.get(&path).copied();
-                match Url::from_file_path(&path) {
-                    Ok(uri) => self.publish_diagnostics_for_uri(uri, diags, version, source),
-                    Err(_) => eprint!("Unable to convert path to uri: {path:?}"),
+                if let Some(uri) = self.unsaved_file_tracker.uri_for_path(&path) {
+                    self.publish_diagnostics_for_uri(uri, diags, version, source)
+                } else {
+                    match Url::from_file_path(&path) {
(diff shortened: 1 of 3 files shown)


rchen152 (approved):

PR #4502, opened 2026-08-10
Do not coalesce LSP responses as mutations

heejaechang:
# Summary

LSP responses are processed on the normal FIFO queue, but they were classified as state mutations. That classification also updates the queue's `last_mutation` marker, which tells earlier event handlers that a later mutation is waiting and that expensive work such as validation can be deferred.

This can lose diagnostics during startup:

1. The server receives `textDocument/didOpen`.
2. The client immediately replies to a server-initiated request such as `workspace/configuration`.
3. Because the response is marked as a mutation, dequeuing `didOpen` reports `subsequent_mutation = true`.
4. `didOpen` skips open-file validation under the assumption that the later mutation will make that work redundant.
5. A configuration response may be a no-op and does not necessarily trigger validation, so nothing compensates for the skipped work.

The server can then remain healthy and responsive while never publishing diagnostics for the opened file until some unrelated later event triggers validation.

This change classifies `LspResponse` with normal-priority query events instead. Responses still preserve FIFO ordering and are handled exactly as before, but they no longer claim that work triggered by preceding document mutations is redundant. This matches the queue's coalescing invariant: only events that can safely supersede prior mutation work should update `last_mutation`.

A deterministic regression enqueues `didOpen` followed by a configuration-shaped response and verifies that:

- `didOpen` is not reported as having a subsequent mutation, so validation is not suppressed.
- The response remains queued and is delivered in FIFO order.

# Test Plan

- `cargo test lsp::non_wasm::queue::tests --quiet`

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/queue.rs
+++ b/pyrefly/lib/lsp/non_wasm/queue.rs
@@ -96,8 +96,8 @@ enum LspEventKind {
     Priority,
     /// An event that makes work triggered by preceding mutations redundant.
     Mutation,
-    /// A FIFO event that does not make preceding mutation work redundant.
-    NonMutation,
+    /// A normal-priority event that does not count as a mutation for coalescing.
+    Query,
 }
 
 impl LspEvent {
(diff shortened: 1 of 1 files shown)


rchen152 (approved):

rchen152:
Thanks!

PR #4505, opened 2026-08-11
feat(cli): support multiple `--output` destinations with different output formats

### LINKED ISSUE #4375: feat(cli): support multiple outputs with arbitrary format combinations
## Problem
Currently, `pyrefly check` supports one single `--output` / `--output-format`. #4087 recently requested the specific combination of `github` and `full-text` output formats, which resulted in the `github-and-full-text` output format being added by #4102.

However, if a user wishes to run `pyrefly check` such that both:

1. The machine-readable diagnostics in the formats of their choice (other than `github` format) are saved
2. The human-readable output in the format of their choice is still printed

...the user has to run `pyrefly check` multiple times.

For me personally, if SARIF is supported (#4205) I'd be interested in supporting a combination of SARIF + full-text or SARIF + min-text. Also with the upcoming open-source Bazel ruleset, it would be useful to output all formats with a single `check` action and allow them to be requested via output groups (`--output_groups=pyrefly_json,pyrefly_sarif,pyrefly_full_text` etc.).

## Request
Support multiple output formats and associated locations, by allowing the `--output` flag to be specified multiple times, possibly prefixed with the format, so it becomes `--output=[<format>:]<path>`.

Example usage to direct `full-text` to stdout and `json` to `pyrefly_diagnostics.json`:
```
pyrefly check --output=full-text:- --output=json:pyrefly_diagnostics.json
```

I'm not sure if this would mean the `--output-format` flag would need to be deprecated. But at the very least it should set the default output format if it's not specified in the `--output` command, which should maintain backwards compatibility.

Of course, as always, I'm open to alternative design choices, this is just my suggestion :)

I'm also happy to implement this if it's considered desirable by maintainers.

tobyh-canva:
Resolves #4375.

### DIFF
--- a/pyrefly/lib/commands/check.rs
+++ b/pyrefly/lib/commands/check.rs
@@ -25,6 +25,7 @@ use anstream::eprintln;
 use anstream::stderr;
 use anstream::stdout;
 use anyhow::Context as _;
+use anyhow::bail;
 use clap::Parser;
 use clap::ValueEnum;
 use dupe::Dupe as _;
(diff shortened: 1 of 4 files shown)


yangdanny97 (approved):

PR #4506, opened 2026-08-11
Check `TypeVar` restrictions when the expected type is gradual

### LINKED ISSUE #3525: False Negative [bad-specialization] with generator expression
### Describe the Bug

In the example below, there should be a `[bad-specialization]`  on the last line, because `member.condition()` is a `Foo`, but `Foo` is not assignable to the upper bound `Bar` of `T@BarContainer`.

```python
from collections.abc import Iterable, Sequence

class Foo: ...
class Bar(Foo): ...

class BarContainer[T: Bar](Bar):
    def __init__(self, modules: Iterable[T] = (), /) -> None: ...

class Conditionable:
    def condition(self) -> Foo: ...
class Impl(Conditionable): ...

class ImplContainer[T: Impl](Impl, Sequence[T]):
    def condition(self) -> BarContainer:
        # false negative
        return BarContainer(member.condition() for member in self)
        
        # with reveal_type() suddenly pyrefly recognizes the error
        # return BarContainer(reveal_type(member.condition()) for member in self)
```

Surprisingly, when we inject a `reveal_type` statement, then pyrefly suddenly detects the issue, so likely a problem with bidirectional inference.

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEAxrlLAwC4S7pyGrYN0INYrkps6ASTYxKfWABo6AZRgBHAK4x0DGAB10%2BhlFRw4dAGK5ciOoTuHjpugCFUlABSXcASht3C%2Bg4mZq6UAMJcbKgQ6DIA2gAqNqEAuu6hvvp02XSYMGB0APqFMRBsxe5wMFBgijS4mOqwcDZSMnIwiSl0ALx07t6KAPTedAC0AHx0AHJcMH72BtqOZhFYZZzoHYhZOXkFTOscXJXVYKOTFlYLAcvBksJQ7muYG1wdvraLQU4Sj2tRGLxJIPYhQNJ-MGKFQaLQ6LqZdA5XL5RhcV7HdCnGoXKahAHRWKUHZI5E5ADEdDAqCgVTosQA5qgOAA3PSksl0SgwNjqShI-GRQkydw0GA0bAyQiHDGbAZU0R0MUSmSCJFVHG7TlasmUgDuZQAFlyYGyaYU2KRiDB5XB1Jg8ugoKQ6GRuWBnSamAz0BAAF7wOhsQ0wOgyaiUHXIync3n8lxuAlAjzcs1QC1Wm3KyWUaXot5Y7yjMCK7OqmJ0DXnEDyEDqDi0kjkRAgSkAVQbZRdYHU2kxcECWFRJcoNGZhXQ6hVHnwNhibFxlbYxJ1sb5SLAuhA0ynOZswHwAF8t-oayA3flnYQ2LQoBRKQAFUjuz1oLB4fBo9CQBl85mbW5KRUUNDTYNhiBaIYhgvD1SEIUQGSGLQhkwXAGDgIZDh-P9MSGBVKDoVAWWiYxsFgL9sNkTE6FwYh%2B1uMhgy4MY2UoOBNl6OgtwAZkIABGAAmE90BAQ9a1QdgIDZcxoBgCg3xwAgm1EoA

_No response_

alexander-beedie:
# Summary

Fixes #3525.

### Overview

A gradual annotation on a constructor call silently disabled bound and constraint checking:

```python
class Container[T: B]:
    def __init__(self, x: T) -> None: ...

def f(x: A) -> Container[Any]:
    return Container(x)  # accepted, but `A` *doesn't* satisfy `T: B`
```
Remove the annotation and the error is caught...

### Explanation

Applying the hint compares `Container[?T]` against `Container[Any]`, which pushes `Any` into `?T`; this was then treated as a solution and locked in, so `x: T` degraded to `x: Any` and accepted anything 👀 

### Solution

An `Any` from the expected type says nothing, so it is discarded, leaving the parameter free for argument matching to check normally (note that an `Any` from an actual argument _is_ still a real solution).

# Test Plan

* All existing unit tests pass, several new tests were added.

* Multiple new matches relating to this are now found in the `mypy_primer` corpus; I've verified a sample of them, and they appeared to be correct (agreeing with `pyright`, which flags the same lines).

### DIFF
--- a/crates/pyrefly_types/src/type_var.rs
+++ b/crates/pyrefly_types/src/type_var.rs
@@ -77,6 +77,20 @@ impl Restriction {
         )
     }
 
+    /// Indicate whether this restriction can reject any type. A bound of `Any` or `object`
+    /// restricts nothing (despite being syntactically a `Bound`) because *every* type
+    /// satisfies it. Constraints always restrict, admitting only their own members, as does
+    /// a `Flag` domain, which admits only its own builtin categories.
+    pub fn can_reject(&self) -> bool {
+        match self {
+            Self::Bound(b) => {
+                !b.is_any() && !matches!(b, Type::ClassType(cls) if cls.is_builtin("object"))
+            }
+            Self::Constraints(_) | Self::Flag(_) => true,
+            Self::Unrestricted => false,
+        }
+    }
+
     fn as_type(&self, stdlib: &Stdlib, heap: &TypeHeap, kind: QuantifiedKind) -> Type {
         match self {
             Self::Bound(t) => t.clone(),
--- a/crates/pyrefly_types/src/type_var.rs
+++ b/crates/pyrefly_types/src/type_var.rs
@@ -79,14 +79,14 @@ impl Restriction {
 
     /// Indicate whether this restriction can reject any type. A bound of `Any` or `object`
     /// restricts nothing (despite being syntactically a `Bound`) because *every* type
-    /// satisfies it. Constraints always restrict, admitting only their own members, as does
-    /// a `Flag` domain, which admits only its own builtin categories.
+    /// satisfies it. Constraints and shape extension restrictions always restrict to their
+    /// respective domains.
     pub fn can_reject(&self) -> bool {
         match self {
             Self::Bound(b) => {
                 !b.is_any() && !matches!(b, Type::ClassType(cls) if cls.is_builtin("object"))
             }
-            Self::Constraints(_) | Self::Flag(_) => true,
+            Self::Constraints(_) | Self::ShapeExtension(_) => true,
             Self::Unrestricted => false,
         }
     }
--- a/crates/pyrefly_types/src/type_var.rs
+++ b/crates/pyrefly_types/src/type_var.rs
@@ -83,9 +83,7 @@ impl Restriction {
     /// respective domains.
     pub fn can_reject(&self) -> bool {
         match self {
-            Self::Bound(b) => {
-                !b.is_any() && !matches!(b, Type::ClassType(cls) if cls.is_builtin("object"))
-            }
+            Self::Bound(b) => !b.is_any() && !b.is_object(),
             Self::Constraints(_) | Self::ShapeExtension(_) => true,
             Self::Unrestricted => false,
         }
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -3516,11 +3516,11 @@ struct SubsetSnapshot {
 }
 
 impl<'solver, 'subset, Ans: LookupAnswer> Subset<'solver, 'subset, Ans> {
-    /// Record the state a speculative check must be able to roll back. `vars` are the variables
-    /// the check could bind; state that is not var-specific is captured wholesale.
-    fn take_snapshot(&self, vars: &[Var]) -> SubsetSnapshot {
+    /// Record the state a speculative check must be able to roll back. The caller chooses which
+    /// vars to capture; state that is not var-specific is captured wholesale.
+    fn take_snapshot(&self, vars: VarSnapshot) -> SubsetSnapshot {
         SubsetSnapshot {
-            vars: self.solver.snapshot_vars(vars),
+            vars,
             subset_cache: self.subset_cache.clone(),
             class_protocol_assumptions: self.class_protocol_assumptions.clone(),
             witness_deferred_vars: self.witness_deferred_vars.clone(),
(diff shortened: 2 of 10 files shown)


alexander-beedie:
Rebased/resolved 👍

crates/pyrefly_types/src/type_var.rs
```diff
@@ -73,6 +73,20 @@ impl Restriction {
         matches!(self, Self::Bound(_) | Self::Constraints(_) | Self::Flag(_))
     }
 
+    /// Indicate whether this restriction can reject any type. A bound of `Any` or `object`
+    /// restricts nothing (despite being syntactically a `Bound`) because *every* type
+    /// satisfies it. Constraints always restrict, admitting only their own members, as does
+    /// a `Flag` domain, which admits only its own builtin categories.
+    pub fn can_reject(&self) -> bool {
```
rchen152: I don't think we should gate on `can_reject`. If we shouldn't solve to `Any` in some circumstances, we should consistently not do so; changing the restriction on a TypeVar shouldn't have unexpected ripple effects.

pyrefly/lib/solver/solver.rs
```diff
@@ -2398,11 +2409,43 @@ impl Solver {
...
+                if let Type::Var(v) = t
+                    && param.restriction().can_reject()
+                {
+                    let mut e = lock.get_mut(*v);
+                    if let Variable::Quantified { quantified, bounds } = &mut *e
+                        && *quantified == *param
+                        && bounds.is_all_any()
+                    {
+                        *bounds = Bounds::new();
+                    } else if matches!(&*e, Variable::Answer { ty, frozen: false } if ty.is_any()) {
```
rchen152: I believe this `else if` block is no longer needed after 898834b58b113b8aba30b25acec2c6065a6d59c8.

pyrefly/lib/solver/solver.rs
```diff
@@ -2398,11 +2409,43 @@ impl Solver {
...
+    ///
+    /// Pass `keep_restricted_free = true` *only* when argument matching follows, such as when
+    /// applying an expected type to a constructor call; otherwise pass `false`. An `Any` that a
+    /// gradual expected type pins onto a restricted parameter says nothing about it, so `true`
+    /// discards it and leaves the parameter free for argument matching to check its restriction.
     pub fn generalize_class_targs(
         &self,
         targs: &mut TArgs,
         vars_with_residual_captures: &SmallSet<Var>,
+        keep_restricted_free: bool,
```
rchen152: I'm a little worried that it'll be hard for future callers of `generalize_class_targs` to understand what value to pass for this bool, since it describes the mechanics of what it does, rather than why. I'd rather either create two separate entrypoints (say, `generalize_class_targs` and `generalize_class_targs_for_constructor_hint`, with the implementation living in a non-`pub` `generalize_class_targs_impl`) or at least rename this bool to something that gives a callers a better hint about whether they want this behavior.

rchen152 (changes requested):
Sorry for taking so long with this review! The solver can be pretty tricky to reason about (partly because variable solving is inherently complex, and partly because we have some tech debt involving unprincipled `Any` filtering), so I wanted to make sure I sat down and worked through the consequences of this change properly.

I think the fix is in the right place (`generalize_class_targs`, specifically for constructor hints), but the current implementation has a couple subtle bugs:

1 - It causes a contextual type hint of `Any` to be lost. An example of what I mean:
```
from typing import Any, MutableMapping

class ChainMap[K, V: bool | dict]:
    def __init__(self, *mappings: MutableMapping[K, V]): ...

mapping: ChainMap[str, Any] = ChainMap(
    {"flag": True},
    {"nested": {"value": 1}},
)
```
Pyrefly accepts this code today because the `ChainMap[str, Any]` annotation is used as a hint to infer `Any` for the mapping value types. With this PR, the hint is lost and Pyrefly starts erroring.

The bug is more obvious if you remove the `can_reject` guard - that's currently limiting the bug to only appearing when a quantified has a bound.

2 - Nested `Any` can still slip through:
```
from typing import Any

class Nested[T: list[str]]:
    def __init__(self, value: T) -> None: ...

x: Nested[list[Any]] = Nested([1])  # Should be an error
```

What I'd suggest doing instead of undoing an all-`Any` answer is to accept the answer but either change it to a new `Variable` variant or add a new field on `Variable::Answer` where we store a restriction, make `is_subset_eq_var` validate against the restriction, and have `finish_quantifieds` convert to a final `Answer`/remove the restriction.

alexander-beedie:
> The solver can be pretty tricky to reason about

Definitely, lol... Thanks for the detailed review!
 I'll try and take care of the points raised today/tomorrow :)

alexander-beedie:
@rchen152: Review issues addressed/reworked, as suggested ✌️😎
(Also: rebased/resolved against the latest updates).

alexander-beedie:
Rebased again 👍 

* Updated function type: `&[Var]` → `VarSnapshot`
* Checking for restricted answer uses `snapshot_for_speculative_inference` .
* Added a new test to prove rollback happens.

Update: also, cleaned up explicit object checks with an `is_object` helper.

rchen152 (approved):
Sorry for the delay! This looks good to me, just waiting for a second internal review now.

stroxler (approved):

PR #4509, opened 2026-08-11
Preserve Any across indexed assignments

### LINKED ISSUE #4488: Indexed assignment incorrectly narrows `Any` values
### Describe the Bug

Indexing into `Any` narrows a type:
```
from typing import Any, reveal_type
values: list[float] = []
frame: Any = object()
reveal_type(frame)
frame["start"] = values
reveal_type(frame["start"])
frame["end"] = frame["start"].shift(-1)  # ERROR: Object of class 'list' has no attribute `shift` [missing-attribute]
```

This code is obviously bad and won't run, but it shows the error. A value typed as `Any` should retain `Any` behavior through indexed access. 

I observed this error with `replace-imports-with-any`, in particular with `geopandas`. Essentially the real world code was something like:
```
distance: ndarray[...] = ...
frame = geopandas.GeoDataFrame()
frame["column_a"] = distance
frame["column_b"] = frame["column_a"].shift(-1)  # ERROR: [missing-attribute]
```

I have `geopandas.*` in `replace-imports-with-any`. I think `pyrefly` sees that `frame["column_a"]` got assigned an `ndarray`, and then later thinks `frame["column_a"]` produces the same `ndarray`, when `geopandas` converts it to a `Series`, of which `shift` is valid.

Mypy reveals both `frame` and `frame["start"]` as `Any` and reports no issues.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.lZQxD4IwEIX_ygUGcIDGld0ZwwokIBRoopS0h_Hne23FgBoSJ5Lj3qXt--79G5mbsLFoLEvWURphSYfM6ZadIisTI6GC49R4t1KHtoVq9psXHrGosPDMBDf3R_uqbaXkY-t0nz2xHkSHYXQ8APhwyrI0SyB16yQ7cHEQmNMHMJDvo4QaUYnLjBwqK64gvwlNq9JH71_lt_U-nF0BXq9IoUfM9DNFK7EQF6NPqcphQJx0wtgil6pnfGStbDTbKBgYyut7LWxgb8eBtKhqGrvD3KS45mhLhK6pOAhj_piuohEY1daeNXq7rDwB

_No response_

cakeni:
# Summary

Preserve indexed-access `Any` behavior when the receiver is explicitly typed as `Any`, including `Any` propagated through attribute access and calls. Track that propagated origin separately so truly implicit or error `Any` keeps the existing post-assignment narrowing behavior.

This fixes the `replace-imports-with-any` case from #4488 without changing unannotated-container behavior.

Fixes #4488

# Test Plan

- `cargo test -p pyrefly_types test_any_style_propagation_preserves_explicit_origin`
- `cargo test -p pyrefly --lib test_subscript_assign -- --nocapture`
- `python test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema`
- `git diff --check`


rchen152 (changes requested):
Thanks for the PR! Based on the mypy primer diff, I think we likely need to distinguish between two cases: the type of the container being `Type::Any(AnyStyle::Explicit)` (preserve `Any`, respecting user intent) and the style being Implicit or Error, in which case I think we should continue narrowing as before. Otherwise behavior on unannotated containers degrades.

NathanTempest (approved):
Hey @cakeni thanks for iterating on this PR, this looks mergeable to me but I will re-run the primer

- The new match arm solves the problem without touching unrelated logic.
- Correct match ordering as the Explicit arm is placed before the ClassType arm and the wildcard, ensuring it's checked first.
- All three AnyStyle variants are tested: explicit Any (should NOT narrow), implicit Any from unannotated params (should still narrow), and error Any from missing types (should still narrow).
- Doc comment is updated to accurately reflect the new behavior.

NathanTempest:
Walking back my earlier approval on this one, I read the new primer in isolation and missed that the split removed the fix along with the noise.

The implementation is correct and does exactly what was asked, and all three new tests pass against main. The problem is scoping. #4488 has two repros: the synthetic "frame: Any = object()" one, which this does fix, and the real one the reporter actually hit geopandas via replace-imports-with-any. I ran that case: "frame = geopandas.GeoDataFrame()" gives an implicit Any, "frame["column_a"]" still narrows to list[float], and the original error still fires. Object of class list has no attribute shift. Type::Any(AnyStyle::Explicit) => false never matches, so the reported bug survives the patch untouched. That means merging with "Fixes #4488" would auto-close a live bug, and it's the one thing here that definitely has to change.

The cause is AnyStyle::propagate(), which maps Explicit to Implicit on attribute access (attr.rs:1603), calls (call.rs:1844) and subscript (expr.rs:3350). Explicit Any only survives one hop, and replace-imports-with-any makes the module itself Any, so the frame is already two hops out. Hand-written "x: Any" is about all this catches.

The primer says the same thing. The original run removed roughly 25 missing-attribute errors and every one is gone now, with pandas dropping out of the diff entirely. rotki is the control as it appears in both runs, kept its implicit-any-empty-container line and lost only the isna line, so this is the split rather than a corpus change. Also worth weighing, @rchen152: unknown-argument-type and unknown-variable-type both default to Severity::Ignore, while the missing-attribute errors the original version removed default to Error, so the regression read larger in the primer than it is for anyone on a default config.

So the decision is to either land this as noise reduction, retitle to "preserve explicitly-annotated Any", change to "Partially addresses #4488", keep the issue open, and add a bug-marked test for the geopandas shape or fix it properly by not narrowing through any Any at all. I'd lean to the latter: calculate_subscript_symmetry (class_metadata.rs:2158) already returns false on every path where symmetry can't be established, so "unproven means don't narrow" is the policy the code encodes, and Any is the unprovable case. Carrying provenance that survives propagate() is the other route, but it's a much larger change and still unsound, since a bare local can hold a DataFrame too.

Minor: the six sibling tests for subscript_assign_arm_allows_narrowing live in pyrefly/lib/test/subscript_narrow.rs, so these three belong there rather than in assign.rs.

Nice work on the PR, the gap here is in the request, not your implementation.

@rchen152 would appreciate your thoughts on this.

cakeni:
Done — switched this to all `Any` styles and moved the tests to `subscript_narrow.rs`. I also added the propagated `Any` case from #4488.

cakeni:
I think keeping track of where propagated Any came from is worth trying. That would let us distinguish it from a truly implicit Any, keep #4488 fixed, and avoid the cloud-init/comtypes regressions. It’s a bigger change, but the latest primer makes the trade-off look reasonable.

cakeni:
Implemented the origin tracking in 0c605d037. Explicit Any now propagates separately, so explicit/propagated Any skips indexed-assignment narrowing while implicit/error Any keeps the existing behavior. Targeted tests and lint pass; the new primer run is in progress.

rchen152 (approved):
Thanks! I appreciate you updating the PR multiple times as we tried to figure out the right tradeoffs here. I'd prefer to represent a propagated `AnyStyle::Explicit` via an "origin" field on `AnyStyle::Implicit` rather than a separate `AnyStyle` variant, just because it seems like it would be way too easy for us to forget to check for `AnyStyle::Propagated` in the future and introduce subtle bugs. No action needed from you; I've updated the imported code.

stroxler (approved):

PR #4510, opened 2026-08-11
Resolve `...` vs `EllipsisType` asymmetry

### LINKED ISSUE #4426: Narrowing against Ellipsis token is broken.
### Describe the Bug

The following should pass without errors:

```python
from types import EllipsisType

def check_narrow_ellipsis(arg: int | EllipsisType) -> None:
    if arg is ...:
        assert_type(arg, EllipsisType)
    else:
        assert_type(arg, int)
```

actual output:

```
ERROR sandbox.py:6:20-39: assert_type(Ellipsis, EllipsisType) failed [assert-type]
ERROR sandbox.py:8:20-30: assert_type(EllipsisType | int, int) failed [assert-type]
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZTNCsIwEIRfJdBDFCR79-7dBxCkxNAGS1Pcqq_vbJpKYqUg3krZmfzMl_mhMrMSSrUgn-9COHSdH9iz8DCjbFtnr4gVh3-eXRrYxDeBqIB0LvqEBUhgUng2xmTYZfsQq11p8pXUhQSrb5cBVuo4_VDpLlBdSL65oyCRqDn1FbrRqXYcB94TzfJwa8j1dAmWqVCQElbrR-1j7ZZ2KkTgGLb_kKMncPQaN1qwwcQqCi8

_No response_

alexander-beedie:
# Summary

Fixes #4426.

### Overview

* Python has _one_ ellipsis object that can be written _two_ ways: as literal `...`, or `types.EllipsisType`.

* `pyrefly` represents these with two different internal types, **but** there's an asymmetry in terms of which can be assigned to which (this is the root cause).

* Specifically: the value `...` is inferred as `Type::Ellipsis`, but a `types.EllipsisType` annotation is inferred as `ClassType(EllipsisType)`, and assignability between the two holds in only _one_ direction.

### Solution

* Add the missing/reverse assignability arm.

* Ensure `literal_equal` recognises both representations.

# Test Plan

* Unit test suite was run, and a new test was added.
* Was able to remove the error markers from:
  *  `test_ellipsis_is` 
  *  `test_ellipsis_eq`
  
  Updated these tests to use `assert_type` as that validates equivalence, rather than matching substrings of the printed type (as per `reveal_type`).

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -2498,9 +2498,18 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
     }
 
     fn literal_equal(&self, left: &Type, right: &Type) -> bool {
+        // `...` in value position and a `types.EllipsisType` annotation denote the same
+        // singleton, so treat the two representations as one for identity comparisons.
+        let is_ellipsis = |ty: &Type| match ty {
+            Type::Ellipsis => true,
+            Type::ClassType(cls) => cls.has_qname("types", "EllipsisType"),
+            _ => false,
+        };
+        if is_ellipsis(left) && is_ellipsis(right) {
+            return true;
+        }
         match (left, right) {
             (Type::None, Type::None) => true,
-            (Type::Ellipsis, Type::Ellipsis) => true,
             (Type::Sentinel(s1), Type::Sentinel(s2)) => s1 == s2,
             (Type::Literal(left), Type::Literal(right)) => left.value == right.value,
             _ => false,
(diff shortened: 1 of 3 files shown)


rchen152:
Thanks for the PR! In the future, would you mind first checking if the issue is already assigned and [claiming](https://github.com/facebook/pyrefly/blob/main/CONTRIBUTING.md#claiming-issues-claim--unclaim) it if it's free? (Don't worry about this PR, though; it looks like it should be pretty easy for me to reconciliate this with my in-progress work.)

alexander-beedie:
> Thanks for the PR! In the future, would you mind first checking if the issue is already assigned and [claiming](https://github.com/facebook/pyrefly/blob/main/CONTRIBUTING.md#claiming-issues-claim--unclaim) it if it's free? 

Ahh, apologies; I'm still not quite used to that... Have claimed one or two properly, but forgot here -- will make sure to do so on the next one! 😅👍

grievejia (approved):

PR #4512, opened 2026-08-11
Fix attribute writes after impossible narrowing

### LINKED ISSUE #4478: Writes to attributes should not use the narrowed type
### Describe the Bug

The following code is written defensively, with a redundant None check. The pattern was found in the sphinx project

Pyrefly narrows the attribute to Never from the None check & subsequently prevents assignments to it, which is wrong since the attribute has a type annotation and should allow assignments consistent with the annotation.

This has implications for narrowing too, since the narrowed type will no longer be valid after the assignment and therefore will need to be cleared.

```
class Container: pass
class Env:
    def __init__(self) -> None:
        self.domains: Container = Container()

    def setup(self) -> None:
        if self.domains is None:      # narrows to Never
            self.domains = Container() # false positive
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.fVRBDsMgDPtKJO7Lvdd9YA_YBa2omzQRBLT7_lygbLSiNxQUExzbvcjMVrmKjXjF-KGwVRxkl557unYZZZ0XjqyY-E8958VuiAHKc104SOUfsQo-3yqyGlx_AkXJsfzr3A_TTnBcqKJbLlDhBlEGJUwzAhMbvtytQlYaesbowsC8tYuf2Fge5RG46WBatasX_Uox3MKRJAEGwJ4oyXkDelIJVOn5HZMcznb8BQ

_No response_

cakeni:
# Summary

Fix attribute writes in impossible narrowing branches to fall back to the receiver's unnarrowed declared path when the flow receiver has become Never. Normal receiver narrowing is preserved, and the assignment continues to invalidate stale attribute facets.

The regression test verifies both sides: a Container write is accepted and an incompatible Literal[1] write is still rejected against the declared Container type.

Fixes #4478

# Test Plan

- cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib test_attr_assignment_uses_declared_type_after_impossible_narrow -- --nocapture
- python test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -448,12 +448,17 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                 let mut idx = self.bindings().key_to_idx(&key);
                 let mut gas = Gas::new(100);
                 let mut narrows = Vec::new();
+                let mut promote = false;
                 loop {
                     if gas.stop() {
                         break;
                     }
                     match self.bindings().get(idx) {
                         Binding::Forward(next) | Binding::PatternCapture(next) => idx = *next,
+                        Binding::PromoteForward(next) => {
+                            promote = true;
+                            idx = *next;
+                        }
                         Binding::Narrow(next, op, location) => {
                             narrows.push((op.as_ref(), location.range()));
                             idx = *next;
@@ -477,6 +482,10 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                             type_info = narrowed;
                         }
                     }
+                }
+                if promote {
+                    type_info =
+                        type_info.map_ty(|ty| ty.promote_shallow_implicit_literals(self.stdlib));
                 }
                 type_info
             }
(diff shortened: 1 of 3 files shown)


shobhitmehro:
Thanks for the PR! Could you review the `mypy_primer` diff above and confirm the new diagnostics are expected?

cakeni:
> Thanks for the PR! Could you review the `mypy_primer` diff above and confirm the new diagnostics are expected?

I checked the latest mypy_primer artifact. It's now +12 new errors, -46 fixed errors; the older +14/-46 result is stale, and the bokeh and steam.py issues are gone.
I also went through the remaining 12 diagnostics. They all look expected, so I don't think any further changes are needed here.

shobhitmehro (changes requested):
Thanks for checking the primer results! I found one case that still fails where it would be nice that this fix is applied to:

```
class Container:
      pass

  class Env:
      def __init__(self) -> None:
          self.container: Container = Container()

  env = Env()

  def update() -> None:
      if env.container is None:
          # This should use the declared Container type, but reports Never.
          env.container = Container()
```

Could you handle this and add a regression test in this diff? Everything else looks great!

shobhitmehro:
thank you for the fixes, this looks good to me! ill cc @rchen152 to take another look for final approval

rchen152:
Sorry for missing this, I'll take a look!

PR #4513, opened 2026-08-11
Resolve nested class forward references in annotations

### LINKED ISSUE #4410: Cannot resolve forward ref to nested class in another attr of enclosing class
### Describe the Bug

Found when checking typeshed

```

class Formatter:
    a: _Section          # false positive: Could not find name `_Section`
    class _Section: ...
    b: _Section          # OK — after the definition
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pVSxasMwEP2VAw-eYu1aC106tNC14CiOHQtcKVhSabZ8RL4wX5J3Ug4SErxkMJinu3cn3bu3aJltOyRogGciq83NmbLKPHB8Zane_fxrIqxEM5Znran9LmsvgiGqaDAQEe19sNyRpjefpi2BFTbi8AMJ0FoS10JWigisqWkaOdo8r_P5QefjicyAniiOPYuSnwhhj5Or6KsAdH0EeBZGvktwRiQ0P66CKfY0xrgPWilJ9_NO9U5tfRfUXYYiFqn5Mzb77T0d-ay0ANpXJFMXxdRLgqlZL4gAwWQ7G1fGHRjO3MD7_-e4xE-HVX438GK6s92kmEtLXHIp4Kx0clMvOWxbN-a7C7qowws

_No response_

cakeni:
# Summary

Allow postponed and quoted annotations in a class body to resolve a nested class declared later in that same body. The deferred class-body lookup now records whether same-class forward resolution is legal, while ordinary runtime reads keep the existing inherited-only behavior.

Tests cover the future-annotations regression on Python 3.13, both annotations before and after the nested class declaration, and a runtime read that must still report unknown-name.

Fixes #4410

# Test Plan

- cargo +stable-x86_64-pc-windows-gnu test -p pyrefly --lib test_nested_class_ -- --nocapture (9 passed)
- python test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/pyrefly/lib/binding/binding.rs
+++ b/pyrefly/lib/binding/binding.rs
@@ -2349,6 +2349,15 @@ impl LegacyTParamBinding {
     }
 }
 
+/// Data for a name in a class body that wasn't found in the static scope.
+#[derive(Clone, Debug)]
+pub struct ClassBodyUnknownName {
+    pub class_key: Idx<KeyClass>,
+    pub name: Identifier,
+    pub suggestion: Option<Name>,
+    pub allow_class_body_forward_reference: bool,
+}
+
 #[derive(Clone, Debug)]
 pub enum Binding {
     /// An expression, optionally with a Key saying what the type must be.
@@ -2522,7 +2531,7 @@ pub enum Binding {
     /// It could either be an unbound name or a reference to an inherited attribute
     /// We'll find out which when we solve the class. The boolean records whether a postponed or
     /// quoted annotation may also resolve an attribute declared later in the same class body.
-    ClassBodyUnknownName(Box<(Idx<KeyClass>, Identifier, Option<Name>, bool)>),
+    ClassBodyUnknownName(Box<ClassBodyUnknownName>),
     /// A match statement or if/elif chain that may be type-exhaustive.
     /// Resolves to Never if ANY narrow entry narrows to Never, None otherwise.
     Exhaustive(Box<ExhaustiveBinding>),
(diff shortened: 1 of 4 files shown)


pyrefly/lib/binding/binding.rs
```diff
@@ -2520,8 +2520,9 @@ pub enum Binding {
     Delete(Box<Expr>),
     /// A name in the class body that wasn't found in the static scope
     /// It could either be an unbound name or a reference to an inherited attribute
-    /// We'll find out which when we solve the class
-    ClassBodyUnknownName(Box<(Idx<KeyClass>, Identifier, Option<Name>)>),
+    /// We'll find out which when we solve the class. The boolean records whether a postponed or
+    /// quoted annotation may also resolve an attribute declared later in the same class body.
+    ClassBodyUnknownName(Box<(Idx<KeyClass>, Identifier, Option<Name>, bool)>),
```
fangyi-zhou: Let's make this a struct for readability, having a bare `bool` inside a 4-tuple would be a bit tricky

fangyi-zhou (changes requested):
Change looks fine apart from minor issue

pyrefly/lib/binding/binding.rs
cakeni: Updated. I replaced the tuple payload with a named ClassBodyUnknownName struct and updated all construction/access sites accordingly. Tests and formatting checks are passing.

fangyi-zhou: Did you forget to push your changes by any chance

cakeni: Thanks for the catch. The struct change is pushed now — ClassBodyUnknownName replaces the tuple payload, and the construction/access sites have been updated accordingly. Could you take another look when you have a chance?

fangyi-zhou (approved):

samwgoldman (approved):

PR #4527, opened 2026-08-12
Stop the ignore-file search at the working-tree boundary

willemkokke:

PR #4528, opened 2026-08-12
Filter hidden directories relative to every include root

### LINKED ISSUE #4526: HiddenDirFilter::RelativeTo carries only import_root, so includes outside it are dropped under a hidden-directory checkout
### Describe the Bug

Split out of #4401 at @kinto0's request — this is the hidden-directory-filter half. The ignore-file half is filed separately as #4525. This is a follow-up to #2402, which introduced `HiddenDirFilter::RelativeTo` so that projects living under a hidden directory (`~/.codex/worktrees/…`, `.claude/worktrees/…`) are not filtered out by their own ancestors.

**1. The allowance carries only `import_root`.** `ConfigFile::get_filtered_globs` builds `HiddenDirFilter::RelativeTo(vec![import_root])`. For a src-layout project `import_root` is `<project>/src`, so an include *outside* it — `tests/` — prefix-matches no root and `has_hidden_component` falls back to checking the **absolute** path's components. Under a hidden-directory checkout every such path has a hidden component, so the include is silently dropped even though nothing inside the project is hidden. The `import_root` allowance fixes `src/` and leaves every sibling include broken.

**2. The allowance is tied to `use_ignore_files`.**

```rust
let root = if self.use_ignore_files {
    self.import_root.as_deref()
} else {
    None
};
```

With `--use-ignore-files=false` the root list is empty and the filter degrades to `HiddenDirFilter::All`, so the whole project disappears. Disabling one exclusion mechanism should not silently harden a different one; the two are unrelated.

### Reproduction

```sh
mkdir -p ~/.tools/checkouts/proj/src ~/.tools/checkouts/proj/tests
cd ~/.tools/checkouts/proj
printf '[tool.pyrefly]\nproject-includes = ["src", "tests"]\n' > pyproject.toml
touch src/m.py tests/check.py
pyrefly check                          # only src/m.py is checked; tests/check.py is skipped
pyrefly check --use-ignore-files=false # nothing is checked at all
```

No git repository is involved, so this is independent of the ignore-file issue above.

### Expected behavior

Hidden ancestors above the project do not hide the project's own files, for every include root rather than just `import_root`, and regardless of `use_ignore_files`. Hidden directories *within* the project (`src/.venv/…`) stay excluded.

I have a fix and will open a PR against this issue.

willemkokke:
Fixes #4526. One of the two PRs replacing #4402, split per review feedback — the other is #4527 (`crates/pyrefly_util/src/globs.rs`). The two touch disjoint files and can land in either order.

**Problem.** Follow-up to #2402. The `HiddenDirFilter::RelativeTo` allowance carried only `import_root`, so for a src-layout project (`import_root = <project>/src`) an include outside it — `tests/check.py` — prefix-matched no root and fell back to absolute-path component checking, where a checkout under a hidden directory (`~/.codex/worktrees/…`, `.claude/worktrees/…`) has every ancestor chain hidden and the include was silently dropped. The root list was also derived from `use_ignore_files`, so `--use-ignore-files=false` degraded the filter to `HiddenDirFilter::All` and hid the project entirely.

**Approach.** Build the allowance from every include root plus the import root, independent of `use_ignore_files`.

**Why it works.** `has_hidden_component` checks components relative to the most specific matching root, so listing every root the project actually declares means each of the project's files is judged relative to a root it lives under — hidden ancestors above the project stop mattering, while hidden directories *inside* it (`src/.venv/…`) are still excluded. And the allowance describes where the project lives, which has nothing to do with whether ignore files are consulted, so disabling one mechanism no longer hardens the other.

**Tests.** New `test_hidden_dir_filter_is_relative_to_every_include_root` covers a src-layout project under `.claude/worktrees/…` with a `tests/` include, in both `use_ignore_files` states, and asserts `src/.venv/…` is still excluded; it fails on `main` (`tests/check.py` not covered) and passes with the fix. The existing `test_get_filtered_globs_coverage_scope` expectation is updated to the new `RelativeTo` construction. `cargo test -p pyrefly_config` passes.

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -798,9 +798,24 @@ impl ConfigFile {
         let hidden_dir_filter = if self.disable_project_excludes_heuristics {
             HiddenDirFilter::Disabled
         } else {
-            match root {
-                Some(r) => HiddenDirFilter::RelativeTo(vec![r.to_path_buf()]),
-                None => HiddenDirFilter::All,
+            // Hidden ancestors above the project's own roots must not hide
+            // the project's files: check components relative to every
+            // include root, plus the import root. The import root alone is
+            // not enough — a src-layout project's `import_root` is `src/`,
+            // and an include outside it (`tests/check.py`) would fall back
+            // to the absolute path, where a checkout under a hidden
+            // directory (`~/.codex/worktrees/…`, `.claude/worktrees/…`)
+            // has every component chain hidden. Deliberately independent of
+            // `use_ignore_files`: turning ignore files off must not make
+            // hidden-directory filtering stricter.
+            let mut roots = includes.roots();
+            if let Some(import_root) = self.import_root.as_deref() {
+                roots.push(import_root.to_path_buf());
+            }
+            if roots.is_empty() {
+                HiddenDirFilter::All
+            } else {
+                HiddenDirFilter::RelativeTo(roots)
             }
         };
         FilteredGlobs::new(includes, project_excludes, root, hidden_dir_filter)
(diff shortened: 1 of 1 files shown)


rchen152 (approved):

PR #4531, opened 2026-08-13
Fix first-use inference for tuple-unpacked names

### LINKED ISSUE #4522: Tuple-Unpacking defeats first-use inference of empty containers
### Describe the Bug

If I have:
```python
def single_assignment() -> None:
    xs = []

    xs.append(1)  # inferred as list[int] -- no error


def tuple_unpacking() -> None:
    xs, ys = [], []
    xs.append(1)  # same appends, but both targets error
    ys.append(2)
```

The LSP shows that `xs` is a `list[int]` with `single_assignment`, but is `list[Unknown}` in the `tuple_unpacking`). It feels the tuple unpacking shouldn't count as "first use" and `xs, ys = [], []` should be basically the same as `xs = []; ys = []`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pVTLDoIwEPyVTTggCdDo0a_wrobwqEDElkhr5O-ddtUEFS4e23R3ujOzMxeZTkIYpO5khlnxt4sEt5963t2n9piKDykCRKpqtY6IAqhzwtSycoz6NYQ6R0oSUpoZdlqy5Y3FQmZW9Xl5BugPnJhGxopn8Qa4h_gGzwtrqNCmIYPVkuYpKleO78pN9C1jQDu-QEfPCAIM-tcWMQld04MKkJCSGmP6YSvEq1xfayGVqHQ5iEmFIOfY_Ja3Pnyn7Uh72w1o-49_QrZPuOSe0JkHL9Cga8vWJLkaPQpTs2yUBw

_No response_

zfaustk:
# Summary

Fixes #4522

Tuple-unpacked name targets now retain their own first-use inference state. Empty containers remain unpinned while the shared right-hand side is unpacked, so each target can be inferred from its first semantic use. Nested unpacking follows the same path; annotated, attribute, subscript, and receiver-constrained targets keep their existing behavior.

This change was implemented with AI assistance. The commands below were run on the submitted commits.

# Test Plan

- `cargo check -p pyrefly --lib`
- `cargo test tuple_unpack_`
- `cargo test test::delayed_inference::`
- `cargo test test::lsp::inlay_hint::`
- `cargo fmt --all --check`

### DIFF
--- a/pyrefly/lib/alt/answers_solver.rs
+++ b/pyrefly/lib/alt/answers_solver.rs
@@ -1985,6 +1985,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         let binding = self.bindings().get(target);
         match binding {
             Binding::NameAssign(na) if na.def_idx.is_some() => Some(target),
+            Binding::UnpackedName(_) => Some(target),
             _ => None,
         }
     }
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -5967,7 +5967,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
                     errors,
                 ),
             Binding::UnpackedName(unpacked) => self.binding_to_type_unpacked_value(
-                unpacked.annotation,
+                None,
                 unpacked.source,
                 unpacked.range,
                 &unpacked.position,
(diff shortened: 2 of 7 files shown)


rchen152:
@zfaustk Looks like we're missing a signed CLA from you. Please follow the meta-cla bot's instructions above to sign the CLA, and we can review your PR once that's done.

PR #4535, opened 2026-08-13
Resolve generated stubs alongside editable packages

### LINKED ISSUE #2859: Generated typestubs are not found in an editable install

ting-hong-shieh:
## Summary

- preserve typed namespace roots when a later search root contains an editable regular package
- use those typed roots as overlays when resolving package submodules
- retain the regular package's `__init__.py` as the package module
- add resolver-level and end-to-end regressions for generated extension stubs

## Root cause

Generated stubs installed next to an extension module form an implicit namespace package. When an editable source directory later supplied the package's `__init__.py`, normal regular-package precedence discarded the earlier namespace root, so submodules such as `_core.pyi` were no longer visible.

The overlay is limited to namespace roots containing `py.typed`, preserving existing behavior for ordinary implicit namespace packages.

## User impact

Editable installs can expose their source package and generated extension stubs at the same time, so imported native APIs retain their declared types.

## Testing

- `cargo test module::finder::tests` (71 tests)
- `cargo test test_editable_install_generated_stub_overlay`

Fixes #2859

### DIFF
--- a/crates/pyrefly_build/src/module_resolver.rs
+++ b/crates/pyrefly_build/src/module_resolver.rs
@@ -233,6 +233,10 @@ pub enum FindResult {
     /// Found a regular package. The first field is its `__init__` file, and the
     /// second is the package directory used as the sole root for submodule lookup.
     RegularPackage(PathBuf, PathBuf),
+    /// Found a regular package after one or more typed namespace directories.
+    /// The namespace directories contain `py.typed` and provide overlays for
+    /// generated extension stubs in editable installs.
+    RegularPackageWithTypedOverlay(PathBuf, Vec1<PathBuf>),
     /// Found a legacy namespace package: a package whose `__init__` calls
     /// `pkgutil.extend_path`. The first field is the winning `__init__`; the
     /// roots accumulate same-named directories across the search path.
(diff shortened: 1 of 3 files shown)


ting-hong-shieh:
The Meta Linter check is currently reporting **Action Required**, but the internal details are not accessible to external contributors. Could someone with access share the failing diagnostic so I can address it? Thanks!

PR #4537, opened 2026-08-13
Split protected member diagnostics from no-access

### LINKED ISSUE #3659: `reportPrivateUsage` from pyright is not covered by `no-access`
[the error kind mapping](https://pyrefly.org/en/docs/migrating-from-pyright/#error-kind-mapping) says `reportPrivateUsage` is covered by `no-access` but that doesn't seem to be the case:

```py
class Foo:
    def _foo(self): ...
    def __bar(self): ...
    
foo = Foo()

foo._foo() # no error
foo.__foo() # missing-attribute
```
https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuiAOgE5vr2-2Yww9APpgWACjgwoYAJSJ6hJZ259%2BgkcOyoOk6XIVLCKnnxNjc9ALxMJsk%2BZaFRd%2BgGJ66SzA4dcXLnQLZxdccVl3egBbCDoIdABzAFpUABdUjghsAFdUmBAAGhAyDkEoUkJU3CioCg8ABVJSsHL6NCw8fHpKXCCIBOyONIhe43QPAGUYGHoAC3TiOEQAemWSsor-BOWYdGXMXEo4ZZ6%2BgaHUkb36MQ56VAA3VGhUbFhu3shz4d76XGJLr04GMyKlZr0kg8fHArtZ6GwQABmQgARgATAjuABtHx%2BDhwAC63Gy6AgUWI-jymCSmAgpUolyhcIA5P0vKVmdwvKkkqUAI7ZOkwakAaxgpBSlEo8AYNmZAHcdOhOaTyVAIJQIDyMKQWbj-CqYPhiOrNdr0Lq5fqOMyQABfIqoBkQKGMaD5RAgdo4Agkch2oA

note that confusingly, `reportPrivateUsage` actually only really applies to protected members, not private members, which are instead covered by `reportAttributeAccessIssue`. [basedpyright playground](https://basedpyright.com/?pyrightVersion=1.39.5&typeCheckingMode=all&reportUnusedParameter=false&reportUnknownMemberType=false&reportUnusedFunction=false&code=MYGwhgzhAEBiD28BcBYAUNT0AmBTAZtAPr6IAUEuI%2BAlEtAHRPpY4HFEBGYAThVbXpMGLLOlLxoAXjjka6cYgYk50AMTQeuAA7weAFwAKPAJYA3MPtwBVCGADmuRfGUr4ZGus069%2BgIL6%2BqacAK5WfsDAuFAAklAhuEA)

ting-hong-shieh:
## Summary

- add a dedicated `private-usage` error kind for protected class members
- allow single-underscore member access from the defining class and subclasses
- keep double-underscore private attributes under `no-access`
- map Pyright's `reportPrivateUsage` to `private-usage` and document the new diagnostic

## Root cause

The Pyright migration mapped `reportPrivateUsage` to `no-access`, even though `no-access` only checked name-mangled double-underscore attributes. This made the migrated setting affect an unrelated diagnostic without enforcing protected-member access.

## User impact

Projects migrating from Pyright can configure protected-member checks independently. Disabling `no-access` no longer disables `reportPrivateUsage` behavior.

## Testing

- `cargo test protected_attribute`
- `cargo test private_attribute`
- `cargo test test_report_private_usage_mapping`
- full Pyrefly library suite: 7,862 passed and 3 ignored; the sandbox-blocked IPC tests passed when rerun with IPC permissions

Fixes #3659

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -347,6 +347,8 @@ pub enum ErrorKind {
     /// guaranteed. This is a separate error code from BadKeywordArgument to allow
     /// users to opt-in to this stricter check.
     PotentialBadKeywordArgument,
+    /// Accessing a protected class member from outside its defining class or a subclass.
+    PrivateUsage,
     /// A protocol attribute was first defined inside a method instead of the class body.
     ProtocolImplicitlyDefinedAttribute,
     /// Calling `.cuda()` on a `torch.Tensor` hard-codes the target device.
--- a/crates/pyrefly_config/src/migration/pyright.rs
+++ b/crates/pyrefly_config/src/migration/pyright.rs
@@ -399,7 +399,7 @@ impl RuleOverrides {
         add(self.report_operator_issue, ErrorKind::NotCallable);
         add(self.report_return_type, ErrorKind::BadReturn);
         add(self.report_return_type, ErrorKind::InvalidYield);
-        add(self.report_private_usage, ErrorKind::NoAccess);
+        add(self.report_private_usage, ErrorKind::PrivateUsage);
         add(self.report_deprecated, ErrorKind::Deprecated);
         add(
             self.report_incompatible_method_override,
@@ -737,6 +737,16 @@ executionEnvironments = [
         let pyr = serde_jsonrc::from_str::<PyrightConfig>(raw_file)?;
         let config = pyr.convert();
         assert!(!config.project_includes.is_empty());
+        Ok(())
+    }
+
+    #[test]
+    fn test_report_private_usage_mapping() -> anyhow::Result<()> {
+        let pyr = serde_json::from_str::<PyrightConfig>(r#"{"reportPrivateUsage": "warning"}"#)?;
+        let config = pyr.convert();
+        let errors = config.root.errors.expect("expected a diagnostic override");
+        assert_eq!(errors.severity(ErrorKind::PrivateUsage), Severity::Warn);
+        assert_eq!(errors.severity(ErrorKind::NoAccess), Severity::Error);
         Ok(())
     }
 }
--- a/crates/pyrefly_python/src/ast.rs
+++ b/crates/pyrefly_python/src/ast.rs
@@ -538,6 +538,10 @@ impl Ast {
         name.starts_with("__") && !name.ends_with("__")
     }
 
+    pub fn is_protected_attr(name: &Name) -> bool {
+        name.len() > 1 && name.starts_with('_') && !name.starts_with("__")
+    }
+
     // Parameters and variables that are prefixed (but not suffixed) with a single underscore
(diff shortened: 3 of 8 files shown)


ting-hong-shieh:
The Meta **Builds & Tests** check is currently failing, but the internal failure details are not accessible to external contributors. The local formatter, Clippy, targeted regressions, and Pyrefly library suite passed. Could someone with access share the failing test or log snippet so I can address it? Thanks!

ting-hong-shieh:
Following up on the internal **Builds & Tests** failure — I think I've found a likely cause, so this may not need anyone to dig out the log.

`PrivateUsage` isn't listed in `default_severity`, so it falls through to `_ => Severity::Error`, and `parent_kind` returns `None` for it. The new diagnostic is therefore an error by default for every project, and it fires on any `obj._protected` access from outside the defining class — ordinary practice inside a library's own internals.

Measured by checking polars 1.43.2's own source (207 files) as a project, building both sides from this checkout so the version is held constant:

| | `main` (`1.3.0-dev.1`) | this branch |
|---|---|---|
| `private-usage` | 0 | **350** |
| `bad-return` | 318 | 318 |
| `missing-import` | 163 | 163 |
| `invalid-annotation` | 6 | 6 |
| **total** | **487** | **837** |

Every other category is identical, so the 350 are entirely attributable to this change. They aren't false positives either — they're helper modules reaching into a sibling class, which is what the check is designed to catch:

```
ERROR Protected attribute `_ldf` cannot be accessed outside of its defining class or a subclass [private-usage]
  --> src/polars/_utils/cloud.py:39:15
```

If a 200-file library produces 350 of these, a large internal codebase would produce far more, which would explain the failure without anything being wrong with the check itself.

If that's right, the fix is to default it to `Severity::Ignore` and let people opt in, the way `explicit-any` and the `implicit-any-*` kinds already work. That also matches the setting it maps from — Pyright has `reportPrivateUsage` at `"none"` in off, basic, and standard, and only `"error"` in strict:

```
| reportPrivateUsage | "none" | "none" | "none" | "error" |
```

The migration path would be unaffected: `test_report_private_usage_mapping` in this PR already asserts that a Pyright config with `reportPrivateUsage` set maps onto `PrivateUsage` independently of `NoAccess`, so projects arriving from Pyright with the rule enabled would still get it. That was the point of #3659.

Happy to push that change. But the default severity of a new diagnostic is a product decision rather than mine to assume, so I'll wait for a maintainer to say which way you want it — and if the internal failure turns out to be something else, a log snippet would still be welcome.

*Disclosure, per the AI Usage section of CONTRIBUTING.md: the measurements above and this comment were produced by an AI agent (Claude Code) working in my checkout. I reviewed them before posting.*

PR #4538, opened 2026-08-13
Improve SQLModel constructor hover and completion

### LINKED ISSUE #4530: SQLModel: better tooltip/completion needed
### Describe the Bug

Pyrefly is doing pretty good job checking missing fields.

```py
from sqlmodel import SQLModel

class A(SQLModel):
    a: int
    b: str

A() # Missing argument `a` in function `A.__init__`
    # Missing argument `b` in function `A.__init__`
```

However, its tooltip is not so informative.

VS Code Image

VS Code

Image

Zed

I believe this could be much better considering pyrefly shows pydantic basemodel an informative tooltip - or even this is a bug.

### Version information

VSCode - official pyrefly extension (v. 1.1.1)

Zed - pyrefly language server extension (v. 0.0.1)

Locally installed pyrefly - v. 1.2.0

### Sandbox Link

_No response_

_No response_

ting-hong-shieh:
## Summary

- prefer an overridden `__init__` signature when a class inherits an unannotated catch-all `__new__`
- use the same preference for constructor call traces so hover and signature surfaces agree
- add SQLModel-shaped Pydantic regressions for hover and keyword completion

## Root cause

SQLModel defines `__new__(cls, *args: Any, **kwargs: Any)` without a return annotation. For subclasses, Pyrefly combined that inherited catch-all with the synthesized Pydantic `__init__`; the broader signature then obscured the model fields in IDE surfaces even though constructor diagnostics still checked those fields correctly.

The new preference is intentionally narrow: `__new__` must be inherited, unannotated, and gradual, while `__init__` must be overridden. Direct construction continues to check both methods.

## User impact

SQLModel subclasses now expose their synthesized Pydantic fields in hover and keyword completion instead of only `args` and `kwargs`.

## Test plan

- `cargo test -p pyrefly test::lsp::hover::`
- `cargo test -p pyrefly test::lsp::completion::`
- `cargo test -p pyrefly constructor`

Fixes #4530

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1124,6 +1124,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // Tracks whether we've already recorded a trace for IDE features.
         // Priority: metaclass __call__ > overridden __new__ > __init__.
         let mut recorded_trace = false;
+        let prefer_init_trace = self.constructor_prefers_init_over_inherited_new(&cls);
         let errors = self.error_collector();
         if let Some(ret) = self.call_metaclass(
             &cls,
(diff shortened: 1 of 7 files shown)


stroxler (approved):

PR #4540, opened 2026-08-13
fix [Feat] better behavior for getattr() w/ literal attribute names #4477

### LINKED ISSUE #4477: [Feat] better behavior for `getattr()` w/ literal attribute names
### Describe the Bug

Right now `getattr()` returns `Any` when no default is provided and `Any | None` when a default is provided.

the arguments are `object, attr_name, default`

I think we can improve the behavior here with some special casing. If the second argument is a known literal, then we should be able to get a precise type for the attribute instead of returning `Any`.

If the literal does not correspond to a known attribute, then we probably need to return `Any` if the class is not final. If the class is final then we might want to error + return `Never`.

```
class Foo:
    x: int # x is conditionally initialized
    def __init__(self):
        pass
    
    def set_x(self, x: int):
        self.x = x

foo = Foo()
attr_name: str = ""

# right now, getattr always returns `Any` when there is no default
reveal_type(getattr(foo, "x"))  # expect `int`
reveal_type(getattr(foo, "y"))  # expect `Any`
reveal_type(getattr(foo, attr_name))  # fall back to regular call path, expect `Any | None`

# right now, getattr always returns `Any | None` when there is a default
reveal_type(getattr(foo, "x", None))  # expect `int | None`
reveal_type(getattr(foo, "y", None))  # expect `Any | None`
reveal_type(getattr(foo, attr_name))  # fall back to regular call path, expect `Any | None`
```

For now maybe we can just do this if the first argument is a `Type::ClassType`, but maybe in the future we can get fancier

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.lVXBboMwDP0Va1xAq5Z7L7vtPGlXpBG1HjChBJEwwr5-dhI2wira3kKIHef5vec7LDPxmSClF60jIu5I3bCQgWMCnrQ6t8EKu5l-0Fp27TeSS1zU3Jp2oS28-jts0L67qKhw0zrCy9ERPC6wQ9OSKmOySGsHTyvvDB5BRq9UGQxt3VhQejpAjZYPguwmOZtIfAMVDYkKpgYV2AYHLy2luSA5drZUK0DymCKn2w90iSsfigIIDnQ9i7eikqvdiHkb4UdKBY_cBQyU3Enw-9A0iX9BmuL2t0fL2UIgb0WAxiJb1RaHJe01OC6Fr6vKyQY-R8NQ8fdz8V-RGbyGDYjkZmZ-tPVIE4_Y-cRYvCFCY21vjkIs4XqoBSpx1icjkggBbD7yS7Z-jqbpQHsHMZR2xwr6AYnOfmvBkfW8J9If

_No response_

asukaminato0721:
# Summary

Fixes #4477

Literal-name getattr now reuses normal attribute inference for class instances instead of always falling back to typeshed’s Any.

It still runs the ordinary builtin call validation first, then only overrides the return type when the call is getattr(obj, "literal", default?), the receiver is a ClassType, and the attribute lookup succeeds.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/test/attributes.rs
+++ b/pyrefly/lib/test/attributes.rs
@@ -916,6 +916,27 @@ def test(foo: Foo) -> None:
 );
 
 testcase!(
+    test_builtin_getattr_literal_attribute_name,
+    r#"
+from typing import Any, assert_type
+
+class Foo:
+    x: int
+
+    def __init__(self) -> None:
+        pass
+
+def test(foo: Foo, attr_name: str) -> None:
+    assert_type(getattr(foo, "x"), int)
+    assert_type(getattr(foo, "x", None), int | None)
+    assert_type(getattr(foo, "y"), Any)
+    assert_type(getattr(foo, "y", None), Any | None)
+    assert_type(getattr(foo, attr_name), Any)
+    assert_type(getattr(foo, attr_name, None), Any | None)
+    "#,
+);
+
+testcase!(
     test_object_setattr,
     r#"
 from typing import assert_type
--- a/pyrefly/lib/test/attributes.rs
+++ b/pyrefly/lib/test/attributes.rs
@@ -937,6 +937,22 @@ def test(foo: Foo, attr_name: str) -> None:
 );
 
 testcase!(
+    test_builtin_getattr_non_special_call_shapes,
+    r#"
+from typing import Any, assert_type
+
+class Foo:
+    x: int
+
+def test(foo: Foo) -> None:
+    assert_type(getattr(*(foo, "x")), Any)
+    assert_type(getattr(*[foo], "x"), Any)
+    assert_type(getattr(foo, "x", *(None,)), Any | None)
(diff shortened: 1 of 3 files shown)


pyrefly/lib/test/attributes.rs
```diff
@@ -915,6 +915,27 @@ def test(foo: Foo) -> None:
...
+        pass
+
+def test(foo: Foo, attr_name: str) -> None:
+    assert_type(getattr(foo, "x"), int)
+    assert_type(getattr(foo, "x", None), int | None)
+    assert_type(getattr(foo, "y"), Any)
+    assert_type(getattr(foo, "y", None), Any | None)
+    assert_type(getattr(foo, attr_name), Any)
+    assert_type(getattr(foo, attr_name, None), Any | None)
+    "#,
```
maggiemoss, line 936: can we add some tests cases for the unreachable examples above? otherwise LGTM.

maggiemoss (changes requested):

PR #4542, opened 2026-08-13
Don't resolve implicit `__bool__` through `__getattr__`

### LINKED ISSUE #4467: `not-callable` false positive when `__getattr__` returns a non-callable type
### Describe the Bug

Hi, I'm not sure if this intended behavior or a regression, since I found two related issues that were supposedly fixed already (#1334 and #1155). But I observed the following situation:

`[not-callable]` is reported for `__bool__` in a Boolean context (e.g. an `or` expression) whenever a class defines `__getattr__` with a non-callable return type and no explicit `__bool__`. 

The issue was found in practice with `torch.nn.Module`, whose `__getattr__` is typed as:

```python
def __getattr__(self, name: str) -> Union[Tensor, "Module"]: ...]
```

Because `nn.Module` has no explicit `__bool__`, pyrefly resolves `__bool__` through
`__getattr__` and infers its type as `Tensor | Module` – not a callable – and raises `[not-callable]`.

## Minimal Example
```python
import torch.nn as nn

def f(x: nn.Module | None) -> nn.Module:
    return x or nn.Module()
```

```
ERROR Expected `__bool__` to be a callable, got `(self: NoneType) -> Literal[False] | Module | Tensor` [not-callable]
  --> repro.py:4:12
   |
4  |     return x or nn.Module()
   |            ^
   |
ERROR Expected `__bool__` to be a callable, got `Module | Tensor` [not-callable]
  --> repro.py:4:12
   |
4  |     return x or nn.Module()
   |            ^
   |
```

**pyrefly version:** 1.2.0

### Sandbox Link

_No response_

_No response_

devteamaegis:
# Summary

Fixes #4467.

Implicit dunder methods such as `__bool__` are looked up on the type and bypass `__getattr__` at runtime (CPython resolves them via `_PyObject_LookupSpecial`, which ignores both the instance dict and `__getattr__`/`__getattribute__`).

When evaluating truthiness in `as_bool`, Pyrefly looked up `__bool__` with the `__getattr__` fallback **enabled**. So a class whose `__getattr__` returns a non-callable type — e.g. `torch.nn.Module`, whose `__getattr__` is typed to return `Tensor | Module` — was treated as defining a non-callable `__bool__`, and using an instance in a boolean context was wrongly reported as `not-callable`:

```python
class Tensor: ...
class M:
    def __getattr__(self, name: str) -> "Tensor | M": ...

def f(x: M | None) -> M:
    return x or M()
    # before: ERROR Expected `__bool__` to be a callable, got `M | Tensor` [not-callable]
    # after:  0 errors
```

The fix disables the `__getattr__` fallback for the `__bool__` lookup in `as_bool` (`pyrefly/lib/alt/expr.rs`). The separate `check_dunder_bool_is_callable` path already disables the fallback; this brings the truthiness-evaluation path in line. The subsequent `__bool__` call is only reached when a real `__bool__` is found directly, so classes that genuinely define `__bool__` are unaffected.

# Test Plan

- Added `test_getattr_does_not_provide_dunder_bool` in `pyrefly/lib/test/attributes.rs`. It fails before the change (`expected 0 errors, but got 2`) and passes after.
- `cargo test -p pyrefly --lib`: all type-checking tests pass (the only failures on my machine are two pre-existing `lsp_interaction::diagnostic` timeout tests that also fail on unmodified `main`, unrelated to this change).
- `cargo clippy -p pyrefly --lib` and `cargo fmt --check` are clean for the changed files.

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1974,9 +1974,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
         }
         ty.as_bool().or_else(|| {
-            // If the object defines `__bool__`, we can check if it returns a statically known value
+            // If the object defines `__bool__`, we can check if it returns a statically known value.
+            // Implicit dunder lookups (like the `__bool__` used for truthiness) are resolved on the
+            // type and do not go through `__getattr__`, so we disable the `__getattr__` fallback here.
+            // Otherwise a class with a `__getattr__` returning a non-callable type would be treated as
+            // defining a non-callable `__bool__` and wrongly reported as `not-callable` (see #4467).
             if self
-                .type_of_magic_dunder_attr(ty, &dunder::BOOL, range, errors, None, "as_bool", true)?
+                .type_of_magic_dunder_attr(ty, &dunder::BOOL, range, errors, None, "as_bool", false)?
                 .is_never()
             {
                 return None;
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1975,10 +1975,8 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         }
         ty.as_bool().or_else(|| {
             // If the object defines `__bool__`, we can check if it returns a statically known value.
-            // Implicit dunder lookups (like the `__bool__` used for truthiness) are resolved on the
-            // type and do not go through `__getattr__`, so we disable the `__getattr__` fallback here.
-            // Otherwise a class with a `__getattr__` returning a non-callable type would be treated as
-            // defining a non-callable `__bool__` and wrongly reported as `not-callable` (see #4467).
+            // Implicit dunder lookups are resolved on the type and do not go through `__getattr__`,
+            // so disable the `__getattr__` fallback here.
             if self
                 .type_of_magic_dunder_attr(
                     ty,
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1980,7 +1980,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             // Otherwise a class with a `__getattr__` returning a non-callable type would be treated as
             // defining a non-callable `__bool__` and wrongly reported as `not-callable` (see #4467).
             if self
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/expr.rs
```diff
@@ -1974,9 +1974,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             }
         }
         ty.as_bool().or_else(|| {
-            // If the object defines `__bool__`, we can check if it returns a statically known value
+            // If the object defines `__bool__`, we can check if it returns a statically known value.
+            // Implicit dunder lookups (like the `__bool__` used for truthiness) are resolved on the
+            // type and do not go through `__getattr__`, so we disable the `__getattr__` fallback here.
+            // Otherwise a class with a `__getattr__` returning a non-callable type would be treated as
+            // defining a non-callable `__bool__` and wrongly reported as `not-callable` (see #4467).
```
shobhitmehro: Nice explanation here. Could we please trim this to the core lookup invariant and remove mentioning the specifc github issue `#4467`?

shobhitmehro (changes requested):
Thanks for the PR, the logic looks good to me! Could you please run `rustfmt` and add test coverage for custom `__getattribute__` and inherited `__getattr__`? Thank you!

devteamaegis:
Thanks for the review @shobhitmehro! Both addressed:

- **rustfmt**: ran it — the `type_of_magic_dunder_attr` call is now formatted (reflowed to multi-line).
- **Tests**: added `test_getattribute_does_not_provide_dunder_bool` (custom `__getattribute__` returning a non-callable type) and `test_inherited_getattr_does_not_provide_dunder_bool` (`__getattr__` inherited from a base class rather than defined directly). Both pass, alongside the original `__getattr__` case — confirming implicit `__bool__` is resolved on the type and skips both the `__getattr__` and `__getattribute__` fallbacks through inheritance too.

Let me know if you'd like any other cases covered.

shobhitmehro:
Thank you for the fixes! I believe [this](https://github.com/facebook/pyrefly/pull/4542/changes/efe668dc0b622363b0ab33f64533f4a7d037d4d1..6ff6a456a1778b9164726df2277c2010314e4715#r3787111784) is still not addressed, but once it is I can accept!

devteamaegis:
Done — trimmed the comment to just the core invariant (implicit dunders resolve on the type and skip the `__getattr__` fallback) and removed the `#4467` reference. Thanks @shobhitmehro!

shobhitmehro (approved):

PR #4544, opened 2026-08-14
fix: preserve known exports in unresolvable `__all__`

### LINKED ISSUE #4424: `from torch import Tensor` incorrectly marked as `implicit-reexport`.

dillydill123:
# Summary

Preserve statically known `__all__` names when a later dynamic append or extend cannot be resolved. This prevents explicit re-exports such as `torch.Tensor` from being reported as implicit while keeping the conservative fallback for the unresolved portion of `__all__`.

Taken from #4479, just modified to split the logic for computing partially known `__all__` entries and to make a separate test for this case.

Fixes #4424

---

There are still a bug with `implicit-reexport` that I decided not to tackle in this PR. `DunderAllKind::Specified` does not account for `DunderAllEntry::Remove`, so an `__all__` that calls `.remove()` will not be correctly identified as an implicit reexport. This is a really minor issue due to the unlikeliness of `__all__.remove`. For example:
```
# foo.py
c: int = 3

# bar.py
from foo import c
__all__ = ["c"]
__all__.remove("c")

# baz.py
from bar import c  # No error here.
``` 

I think fixing this and simplifying the code could involve modifying `get_explicit_dunder_all_names_iter` to accept `DunderAllKind::Specified | DunderAllKind::Unresolved` and provide the same entry logic + handle `DunderAllEntry::Remove`. This would remove the need for the `partially_known_dunder_all` member variable. But `get_explicit_dunder_all_names_iter` is used elsewhere and modifying this might have an unexpected blast radius. So I kept the safer, minimally local approach. Another option is to avoid using `get_explicit_dunder_all_names_iter` in the implicit reexport check and implement that same replay logic described above, only for use in that check only. I didn't feel comfortable implementing either solution and understanding its effects.

---

There was one other quirk. `DunderAllKind::Unresolvable` is not "Sticky". The following causes `__all__` to be `DunderAllKind::Specified`:
```
__all__ = []
for name in ["a"]:
    __all__.append(name)
__all__.remove("a")
```

Whereas the following causes `__all__` to be `DunderAllKind::Unresolvable`
```
__all__ = []
__all__.remove("a")
for name in ["a"]:
    __all__.append(name)
```

But this didn't bring out any bugs or unexpected behavior I could find.

# Test Plan

- `python3 test.py`
- Built Pyrefly `maturin build --release -m pyrefly/Cargo.toml` and checked `from torch import Tensor` against `torch==2.12.0` with `pyrefly check --preset all example.py`. No significant runtime increase and 0 errors now.

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -806,19 +806,15 @@ impl DefinitionsBuilder {
                                 self.inner.dunder_all.entries.extend(entries);
                             }
                             None => {
-                                self.inner.dunder_all = DunderAll {
-                                    kind: DunderAllKind::Unresolvable(arguments.args[0].range()),
-                                    entries: Vec::new(),
-                                };
+                                self.inner.dunder_all.kind =
+                                    DunderAllKind::Unresolvable(arguments.args[0].range());
                             }
                         },
                         "append" => match DunderAllEntry::as_item(&arguments.args[0]) {
                             Some(entry) => self.inner.dunder_all.entries.push(entry),
                             None => {
-                                self.inner.dunder_all = DunderAll {
-                                    kind: DunderAllKind::Unresolvable(arguments.args[0].range()),
-                                    entries: Vec::new(),
-                                };
+                                self.inner.dunder_all.kind =
+                                    DunderAllKind::Unresolvable(arguments.args[0].range());
                             }
                         },
                         "remove" => {
--- a/pyrefly/lib/export/exports.rs
+++ b/pyrefly/lib/export/exports.rs
@@ -110,6 +110,8 @@ pub struct Exports {
     /// but they take up very little space, so not worth the hassle to detect when
     /// calculation completes.
     definitions: Definitions,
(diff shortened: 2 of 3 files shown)


rchen152 (approved):

PR #4551, opened 2026-08-15
feat: support `baseline-error-level` config option, emit baseline provenance in `pyrefly check` output

### LINKED ISSUE #4543: feature request: allow `pyrefly check` to emit baselined findings at reduced severity
## Problem
`pyrefly check` currently suppresses findings that match the configured baseline. This prevents known findings from appearing in text, JSON, SARIF, and other output formats.

Baselines are useful to unblock type-checker upgrades and migrations without needing to change any Python code, but in their current state, they also hide away technical debt in an external file. Users may want to surface baselined findings as warnings to developers as a way of flagging technical debt in their code.

## Request
Make it possible to emit baselined errors in Pyrefly's CLI output (both human- and machine-readable formats), whilst noting that the findings are part of the baseline.

### New configuration option
Add a `--baseline-error-level` CLI option and `baseline-error-level` config option accepting:

- `error` (severity is never downgraded)
- `warn`
- `info`
- `ignore` (default, preserving current behaviour)

When set to `warn` or `info`, matching baseline findings should remain in the output with the severity reduced to this level. Findings not in the baseline should retain their original severity.

For example:

```sh
pyrefly check \
  --baseline=pyrefly-baseline.json \
  --baseline-error-level=warn
```

### New output information
When a baselined finding is emitted in Pyrefly's output, in addition to having its severity reduced, it should also be clear that the finding is part of the baseline. This would help separate newly introduced `warn`-level findings from baselined `error`-level findings, downgraded to `warn`.

For `--output-format=text`, this could be done with a suffix to the error's message header, like so:

```
 WARN `Literal[1]` is not assignable to `str` [bad-assignment] [baselined]
 --> matched.py:1:10
  |
1 | x: str = 1
  |          ^
```

For `--output-format=json`, we could emit an additional boolean field:

```json
{
  "errors": [
    {
      "line": 1,
      "column": 10,
      "stop_line": 1,
      "stop_column": 11,
      "path": "matched.py",
      "code": -2,
      "name": "bad-assignment",
      "description": "`Literal[1]` is not assignable to `str`",
      "concise_description": "`Literal[1]` is not assignable to `str`",
      "severity": "warn",
      "baselined": true
    },
    ...
  ]
}
```

And for `--output-format=sarif`, there is the standard `baselineState` field, which should be set to either `unchanged` (meaning the finding matches the baseline) or `new` (meaning the finding is new):

```json
{
  "version": "2.1.0",
  "runs": [
    {
      "results": [
        {
          "ruleId": "bad-assignment",
          "level": "warning",
          "message": {
            "text": "`Literal[1]` is not assignable to `str`"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "matched.py"
                }
              }
            }
          ],
          "baselineState": "unchanged"
        },
        {
          "ruleId": "bad-return",
          "level": "error",
          "message": {
            "text": "Returned type `Literal[1]` is not assignable to declared return type `str`"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "new.py"
                }
              }
            }
          ],
          "baselineState": "new"
        }
      ]
    }
  ]
}
```

## Open questions
- The Pyrefly LSP currently displays baselined findings as `HINT` diagnostics, and this is similarly not configurable.
    1. Should `baseline-error-level` effect this behaviour? For consistent and unsurprising UX, I think it probably should - i.e. `baseline-error-level = "warn"` should upgrade the baselined errors emitted by the LSP to `WARNING`.
    2. With the default `baseline-error-level = "ignore"`, should the LSP continue emitting baseline errors as `HINT`? Or should they be suppressed?

tobyh-canva:
Resolves #4543.

With the `baseline-error-level` option, baselined errors can be shown in Pyrefly's output at a possibly reduced severity. 

When this setting causes baselined results to be present in the output, most output formats (excluding `code-climate` and `junit-xml`) include a marker with each result which matches the baseline.

Default `pyrefly check` behaviour remains unchanged. Also, this setting does not affect the Pyrefly LSP (yet).

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -594,8 +594,12 @@ pub struct ConfigFile {
     pub typeshed_path: Option<PathBuf>,
 
     /// Path to baseline file for comparing type errors.
-    /// Errors matching the baseline are suppressed.
+    /// Errors matching the baseline are suppressed by default.
     pub baseline: Option<PathBuf>,
+
+    /// Severity assigned to errors that match the baseline.
+    /// Defaults to `ignore`.
+    pub baseline_error_level: Option<Severity>,
 
     /// Default error output format for CLI checks when `--output-format` is not set.
     pub output_format: Option<OutputFormat>,
@@ -710,6 +714,7 @@ impl Default for ConfigFile {
             use_ignore_files: true,
             typeshed_path: None,
             baseline: None,
+            baseline_error_level: None,
             min_severity: None,
             output_format: None,
             skip_lsp_config_indexing: false,
@@ -2047,6 +2052,7 @@ mod tests {
                 },
                 typeshed_path: None,
                 baseline: None,
+                baseline_error_level: None,
                 min_severity: None,
                 skip_lsp_config_indexing: false,
                 extra_file_extensions: Vec::new(),
@@ -2371,6 +2377,7 @@ mod tests {
             },
             typeshed_path: Some(PathBuf::from(typeshed)),
             baseline: Some(PathBuf::from("baseline.json")),
+            baseline_error_level: None,
             min_severity: None,
             skip_lsp_config_indexing: false,
             extra_file_extensions: Vec::new(),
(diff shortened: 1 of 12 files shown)


stroxler (approved):

PR #4553, opened 2026-08-15
Remove float for unnecessary-type-conversion warning to account for int

### LINKED ISSUE #4524: Potential false-positive `unnecessary-type-conversion` for `float` annotations
### Describe the Bug

Pyrefly reports `float(a)` as `"unnecessary-type-conversion"`:
```python
def f(a: float) -> None:
    float(a)

```

Python: 3.12

I think this is potentially a false positive because of PEP 484’s numeric-tower rule: an `int` is accepted where a `float` is annotated, even though the runtime value may still actually be an `int`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.w1VkghOiRqKVQhqwkMHIBGBBjURgMGOGgrJCAERAAWogMP8Dgy-9FFjKAINFLyZPGVjApCpklJQUFFvp68O05xel66fm6afkJxfro-jQVwBFeGJZYia47EI1TiEfHGvFQGMpCX51SOir4wt8dVDYA1XgDU8A

kakolla:
# Summary

As @pdhall99 found, pyrefly reports a 'unnecessary-type-conversion' warning for running a float() conversion, even though an int can be passed without runtime conversion which makes float() valid under python's numeric tower rule. 

Fixes #4524

# Test Plan
$ cargo test
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

$ cargo test test_float_to_float_ok
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out; finished in 0.00s

$ python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema
passed with 1 warning unrelated to this change

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -1496,7 +1496,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         range: TextRange,
         errors: &ErrorCollector,
     ) {
-        let builtin_names = ["str", "int", "float", "bool", "bytes"];
+        let builtin_names = ["str", "int", "bool", "bytes"];
         if !builtin_names
             .iter()
             .any(|name| cls.has_qname("builtins", name))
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #4554, opened 2026-08-15
fix Semantic auto completion fails on lines preceding comments #4552

### LINKED ISSUE #4552: Semantic auto completion fails on lines preceding comments
### Describe the Bug

In VS Code, when triggering auto completion on a line located above a comment, Pyrefly fails to offer semantic Python completions like variable and function names. Instead, it falls back to plain text based suggestions.

## Steps to Reproduce
```py
import datetime


class FooBar:
    def __init__(self):
        pass


def _no_code_complete(foo: str):
    bar = datetime.date.fromisoformat(foo)
    
    # foobar foobar foobar


def _pass(foo: str):
    bar = datetime.date.fromisoformat(foo)
    

```

On line 17, type `foo.` and press `Ctrl + Space`. Member functions are shown as expected. Do the same on line 11. Observe that only plain text suggestions appear.

Pylance does not have such issue.

### Sandbox Link

_No response_

VS Code 1.130.0
Python extension 2026.4.0
Pyrefly extension 1.2.0
Python 3.14.7

asukaminato0721:
# Summary

Fixes #4552

Corrected CRLF-aware comment offsets

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -4344,13 +4344,20 @@ impl<'a> Transaction<'a> {
         let source = module.lined_buffer().contents();
         let mut offset = TextSize::from(0);
 
-        for line in source.lines() {
+        for line_with_ending in source.split_inclusive('\n') {
+            let line_without_lf = line_with_ending
+                .strip_suffix('\n')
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #4561, opened 2026-08-16
Show bound methods without their receiver parameter

### LINKED ISSUE #4162: Confusing diagnostic when assigning a bound method to a class (unbound) method
### Describe the Bug

Related to #3810 I think, but the current open PR for the fix looked to add a new error kind. I think this is still a `bad-assignment` and just requires adjustment of the diagnostic.

```
class A:
    def f(self) -> None: pass
    def g(self) -> None: pass

def h(self: A) -> None: pass

A.f = A().g
```
`pyrefly` correctly identifies an error but does not include "boundness" in the diagnostic.
```
ERROR `(self: A) -> None` is not assignable to attribute `f` with type `(self: A) -> None` [bad-assignment]
```

mypy error for reference.
```
example.py:7: error: Incompatible types in assignment (expression has type "Callable[[], None]", variable has type "Callable[[A], None]")  [assignment]
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQCCiAOuvR-ZjGPWABRwYUMAEp6AWgB89AHK50MRPWK04bTlx70A5oOFjJM%2BYuWq6bNt14ALfSOWNx0uQqUq1l9I0K8AvEz8ooQ6bCAANCBkAE48UKSEAC64ALZQFADE9AAKpLFg8fRoWHj49JQKkDoArtGoiRAKhGxZAMowMPQ2iYnEcIgA9AMxcQm40ToDMOgDmLiUcAMV6FW19Y0zfOP0qABuqNCo2LDllRA1dQ0K9LjEV%2BhwzehkiTYKErsw0XAb9AEsIAAzIQAIwAJgBbGIsSEiT%2B9ABqCgUABIAAvpFUJQGp8AGLQGAUYo4AgkchooA

_No response_

ting-hong-shieh:
# Summary

Fixes #4162.

A bound method was displayed as its underlying function type, receiver parameter
included, so it printed identically to the unbound method it came from. That is
what makes the diagnostic in the issue unreadable — both sides render the same
text:

```
ERROR `(self: A) -> None` is not assignable to attribute `f` with type `(self: A) -> None` [bad-assignment]
```

`BoundMethodType::strip_receiver` drops the parameter that binding has already
consumed, and the standard display mode renders that view. This is the same strip
the solver's `bind_bound_method_type` performs when it checks a call, so the
displayed type is the one being checked:

```
ERROR `() -> None` is not assignable to attribute `f` with type `(self: A) -> None` [bad-assignment]
```

This stays within `bad-assignment` and adds no new error kind, per the issue's
suggestion that the kind is right and the wording is the problem.

The signature-mismatch block in override errors is untouched, since it renders
declarations rather than types and `self` belongs there:

```
Class member `B.method` overrides parent class `A` in an inconsistent manner
  `B.method` has type `() -> None`, which is not assignable to `(x: int) -> int`, the type of `A.method`
  Signature mismatch:
  expected: def method(self: B, x: int) -> int: ...
  found:    def method(self: Unknown) -> None: ...
```

# Test Plan

`cargo test` on `2c37b0a`, rebased onto `ef5ef0b`: 7882 passed, 0 failed.
`test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` is clean;
the one clippy warning it reports is pre-existing in `pyrefly_config/src/error_kind.rs`.

Before the change, 27 tests disagreed — 22 in the checker's test suite and 5 in
`display.rs`'s own unit tests — touching 27 expectation lines across 9 test files
plus 5 in `display.rs`. All of them moved in the same direction, the receiver
disappearing from a bound method, and every one is more accurate afterwards:

| Case | Before | After |
|---|---|---|
| `reveal_type(C.foo)`, a classmethod | `(cls: type[C]) -> int` | `() -> int` |
| `reveal_type(p.__iter__)` | `(self: Pair) -> Iterator[int \| str]` | `() -> Iterator[int \| str]` |
| `reveal_type(wrapper.__call__)` | `[R](self: Wrapper[[x: R], R], x: R) -> R` | `[R](x: R) -> R` |
| property setter in an override error | `(self: ChildNarrowed, value: A) -> None` | `(value: A) -> None` |
| `__post_init__` mismatch | `(self: Bad1, y: bytes, z: str) -> None` | `(y: bytes, z: str) -> None` |

The clearest one is `test_match_method_against_callable`, which now distinguishes
the two calls it was written to contrast:

```python
f1(C.f)     # `(self: C, x: int) -> None` is not assignable to `(int) -> None`
f2(C().f)   # `(x: int) -> None` is not assignable to `(C, int) -> None`
```

Previously both reported `(self: C, x: int) -> None`, so the second error read as
though the checker had rejected a matching type.

Also checked by hand against a local build: classmethods, staticmethods,
overloads, generic methods, `__init__`, and properties all display correctly, and
`reveal_type(a.g)` and `reveal_type(A.g)` are now distinguishable.

# One thing worth a reviewer's opinion

In the read-write attribute case, the two sides now use different forms:

```
`B.foo` has type `(self: Unknown) -> None`, which is not consistent with `(x: int) -> int` in `A.foo`
```

The left side is a plain callable attribute rather than a bound method, so it
keeps its parameter while the right side loses `self`. That is accurate — the
distinction is exactly what this change makes visible — but it reads as an
inconsistency if you don't know why. I left it alone rather than special-casing
the message. Happy to change it if you'd rather it were uniform.

# AI usage disclosure

Per the AI Usage section of CONTRIBUTING.md: this change, the expectation updates,
and this description were produced by an AI agent (Claude Code) working in my
checkout. I reviewed them before submitting. Any replies I make on this PR will
carry the same disclosure if an agent drafts them.


stroxler:
@ting-hong-shieh this looks pretty good to me! Could you rebase and make sure conformance tests pass? I'm seeing failures in the imported diff.

It's likely just some type errors that this commit improves

stroxler:
Cool! I see one failing test, which I think is just a stale expectation(?)

And the conformance tests are still failing. I'm not actually sure that we run them in external CI, I'm not seeing a signal, but if you just run `./test.py` from the repo root, it will run conformance tests and regenerate expectations if necessary

I'll also try to fix that gap in github CI

stroxler (approved):

PR #4562, opened 2026-08-17
docs: fix tensor shapes syntax to match shipped stubs

### LINKED ISSUE #4560: Tensor shapes docs use shape syntax that no longer checks against the shipped stubs
### Describe the Bug

The tensor shapes documentation teaches a shape syntax that the shipped stubs no
longer accept. Code written the way the docs show it does not check.

The current stubs want `IntVar`-bounded type parameters and a single shape list
argument to `Tensor`:

```python
class Linear[IN: IntVar, OUT: IntVar](Module):        # torch-stubs/nn/__init__.pyi
    weight: Tensor[[OUT, IN]]
```

The docs use unbounded parameters and a flat argument list — `class Model[D]`,
`Tensor[B, T, D]`. Across `website/docs/tensor-shapes*.mdx` there are 81
occurrences of the flat `Tensor[...]` form, 31 declarations with unbounded shape
type parameters, and no `IntVar` bound on any of the pages listed below.

**Reproduction.** pyrefly 1.2.0 and pyrefly-torch-stubs 1.2.0, both from PyPI,
Python 3.12.

The step-by-step model from
[Tutorial: the basics](https://pyrefly.org/en/docs/tensor-shapes-tutorial-basics/),
copied as the page builds it up:

```python
class BaselineActor[S, A](nn.Module):
    def __init__(self, state_size: Int[S], action_size: Int[A]) -> None:
        super().__init__()
        self.fc1 = nn.Linear(state_size, 400)
        self.fc2 = nn.Linear(400, 400)
        self.out = nn.Linear(400, action_size)

    def forward[B](self, state: Tensor[B, S]) -> Tensor[B, A]:
        h1 = F.relu(self.fc1(state))
        assert_type(h1, Tensor[B, 400])
        h2 = F.relu(self.fc2(h1))
        return torch.tanh(self.out(h2))
```

```
ERROR `S` must be an `IntVar` to be used as a shape dimension. [invalid-annotation]
ERROR `A` must be an `IntVar` to be used as a shape dimension. [invalid-annotation]
ERROR `B` is not assignable to upper bound `IntTuple` of type variable `Shape` [bad-specialization]
ERROR Expected 1 type argument for `Tensor`, got 2 [bad-specialization]
ERROR assert_type(Tensor[[*tuple[int, ...], 400]], Tensor[B]) failed [assert-type]
 INFO 8 errors
```

The `Int[X] | None` example from
[Reference](https://pyrefly.org/en/docs/tensor-shapes-reference/) fails the same
way, with `Expected 1 type argument for Tensor, got 3`. The same page's "Usage
patterns" table prescribes the old form directly:

| Pattern | Purpose |
|---------|---------|
| `class Model[D](nn.Module)` | Make a dimension a class-level type parameter |
| `def forward[B](self, x: Tensor[B, D])` | Bind a per-call dimension |

Rewriting either example to match the stubs — `[S: IntVar, A: IntVar]`,
`Tensor[[B, S]]` — checks clean, so this looks like documentation lagging an API
change rather than a checker bug. The repository's own examples, e.g.
`tensor-shapes/pyrefly-torch-stubs/examples/nanogpt.py`, already use the new
form.

**Scope.** Counts of the flat `Tensor[...]` form per page:

| Page | Occurrences |
|---|---|
| `tensor-shapes-tutorial-architectures.mdx` | 19 |
| `tensor-shapes-reference.mdx` | 17 |
| `tensor-shapes-tutorial-basics.mdx` | 12 |
| `tensor-shapes-tutorial-loops.mdx` | 12 |
| `tensor-shapes.mdx` | 11 |
| `tensor-shapes-tutorial-advanced.mdx` | 9 |
| `tensor-shapes-ai-porting.mdx` | 1 |

I verified two of these pages by running them; I have not run every example. The
examples on `tensor-shapes.mdx` are exposition rather than code to copy, but they
still show the old form.

The inlay-hint screenshots under `website/static/img/tensor-shapes/` show the old
syntax too (`def forward[B, T](self, x: Tensor[B, T, NEmbedding])`), so they need
retaking rather than editing.

**Question before anyone starts on it:** is this the same underlying change as
#4153, and is a rewrite already underway? I'm happy to do the mechanical pass
across the tutorials and the reference page if that's useful, but not if it would
collide with work in progress.

### Sandbox Link

_No response_

_No response_

*Disclosure, per the AI Usage section of CONTRIBUTING.md: the counts, the reproductions, and this report were produced by an AI agent (Claude Code) working in my checkout. I reviewed them before posting.*

Vishwaspatel2401:
# Summary

The tensor shapes docs taught a shape syntax that the shipped
pyrefly-torch-stubs no longer accept: unbounded type parameters
(`class Model[D]`) and a flat multi-argument `Tensor[D1, D2, ...]` form.
Code written the way the docs showed it did not check.

This rewrites the affected pages to match what the stubs actually require:
`IntVar`-bounded type parameters and a single shape-list argument to
`Tensor` (`Tensor[[D1, D2]]`), matching the pattern already used in
`torch-stubs/nn/__init__.pyi`.

One page needed more than a bracket rewrap: the variadic-dimension
convention. The docs taught a bare `TypeVarTuple` (`def forward[*Bs](...,
x: Tensor[*Bs, D])`), which doesn't just need brackets — it fails outright
under the current stubs (`Unpacked type in IntTuple must use Elements[...],
got Bs`). Replaced with the convention `nn.Linear.forward` actually uses:
a type parameter bound to `IntTuple`, spliced into the shape list with
`*Elements[...]` (`def forward[Bs: IntTuple](..., x: Tensor[[*Elements[Bs], D]])`).

`tensor-shapes-setup.mdx` is intentionally not touched here — it's already
being fixed by #4559.

Two things came up during this pass that are real but out of scope for this
issue, noted here rather than silently fixed:
- `tensor-shapes-tutorial-advanced.mdx`'s DCGAN section claims
  `DCGAN.ngf * 8` resolves to `Literal[512]`; it actually widens to plain
  `int`. Unrelated to the shape-syntax bug — a separate literal-arithmetic
  question.
- The recursive/exponential examples in `tensor-shapes-tutorial-architectures.mdx`
  (`UNet.recurse`/`_decode`, `Generator._chain`) hit genuine algebraic-gap
  solver limitations even with the corrected syntax — the same class of gap
  the page's own "Algebraic gaps" section already documents and prescribes
  `type: ignore` for.

Fixes #4560

# Test Plan

Environment: `pyrefly==1.2.0` and `pyrefly-torch-stubs==1.2.0` from PyPI in a
clean venv, matching the issue's own reproduction environment.

Confirmed the single-file default preset (`pyrefly check file.py` with no
`pyrefly.toml`) silently uses the lenient `basic` preset, which disables
assignment/call-shape validation and would falsely show 0 errors even on
broken code — all checks below use `pyrefly check -p default` to get real
validation.

Re-ran both of the issue's own reproductions, extracted verbatim from the
now-fixed files:
- The `BaselineActor` example from `tensor-shapes-tutorial-basics.mdx.`
  (the one that produced the issue's 8 errors) — now `0 errors`.
- The `Attention`/`Int[X] | None` example from `tensor-shapes-reference.mdx`
  (`Expected 1 type argument for Tensor, got 3`) — that error is gone; the
  only remaining diagnostic is an unrelated `missing-return` from the
  fragment's own `...` placeholder body, present in the original doc too.

For every other changed code block, reconstructed it with minimal synthetic
scaffolding (stand-in modules for fragments that reference outer context)
and ran `pyrefly check -p default`, confirming `assert_type` calls pass and
signatures unify. Swept all 7 files afterward for any remaining flat
`Tensor[...]` form — none found.

Did not run the Docusaurus build (`website/`'s `node_modules` isn't
installed in this environment); checked code-fence and bracket balance
across all 7 files as a structural proxy instead.

# AI usage disclosure

Per the AI Usage section of CONTRIBUTING.md: this fix, its verification
against the shipped stubs, and this description were produced by an AI
agent (Claude Code) working in my checkout. I reviewed them before
submitting.

### DIFF
--- a/website/docs/tensor-shapes-ai-porting.mdx
+++ b/website/docs/tensor-shapes-ai-porting.mdx
@@ -84,7 +84,7 @@ The skill produces a `verify_port.sh` report with these key metrics:
 | **`ig` (ignore)** | `type: ignore` comments. Lower is better. Each should have a category (algebraic gap, conditional equality, stub gap). |
 | **`bs` (bare sig)** | Bare `Tensor` in function signatures. Should be 0 for well-typed modules. |
 | **`bv` (bare var)** | Bare `Tensor` in local variable annotations. Lower is better. |
-| **`sh` (shaped)** | `assert_type` calls with full shapes (e.g., `Tensor[B, D]`). Higher is better. |
+| **`sh` (shaped)** | `assert_type` calls with full shapes (e.g., `Tensor[[B, D]]`). Higher is better. |
 | **`ba` (bare assert)** | `assert_type(x, Tensor)` — tracking gaps. Lower is better. Each should have a root-cause comment. |
 | **`sm` (smoke)** | Smoke test functions. At least 1-2 per model. |
 
--- a/website/docs/tensor-shapes-reference.mdx
+++ b/website/docs/tensor-shapes-reference.mdx
@@ -23,7 +23,11 @@ description: Experiment with Int operators, type-level arithmetic in shapes, and
 ## `Int[X]`
 
 `Int[X]` is a type constructor that bridges runtime integer values to
-type-level symbols. It is defined in the `shape_extensions` package.
+type-level symbols. It is defined in the `shape_extensions` package,
+alongside the other names used throughout this page: `IntVar` (the bound
+for shape type parameters), `IntTuple` (the bound for whole-shape type
+parameters), and `Elements` (splices a whole-shape type parameter into a
+`Tensor` argument list).
 
 ### Basics
 
--- a/website/docs/tensor-shapes-reference.mdx
+++ b/website/docs/tensor-shapes-reference.mdx
@@ -61,11 +61,11 @@ For optional dimensions — parameters that may or may not be present — use
 `if param is not None:` to recover `Int[X]` inside the branch:
 
 ```python
-class Attention[D, RK](nn.Module):
+class Attention[D: IntVar, RK: IntVar](nn.Module):
     def __init__(self, dim: Int[D], rank_k: Int[RK] | None = None):
         ...
 
-    def forward[B, T](self, x: Tensor[B, T, D]) -> Tensor[B, T, D]:
+    def forward[B: IntVar, T: IntVar](self, x: Tensor[[B, T, D]]) -> Tensor[[B, T, D]]:
         if self.rank_k is not None:
             # rank_k is Int[RK] here
             ...
@@ -76,11 +76,11 @@ class Attention[D, RK](nn.Module):
 | Pattern | Purpose |
 |---------|---------|
 | `def __init__(self, dim: Int[D])` | Accept a dimension as a constructor parameter |
-| `class Model[D](nn.Module)` | Make a dimension a class-level type parameter |
-| `def forward[B](self, x: Tensor[B, D])` | Bind a per-call dimension |
+| `class Model[D: IntVar](nn.Module)` | Make a dimension a class-level type parameter |
+| `def forward[B: IntVar](self, x: Tensor[[B, D]])` | Bind a per-call dimension |
 | `self.head_dim = dim // n_head` | Compute a derived dimension (`Int[D // NHead]`) |
 
-## `Tensor[D1, D2, ...]`
+## `Tensor[[D1, D2, ...]]`
 
 `Tensor` with type arguments represents a tensor with a known shape. The
 type arguments are the dimensions, in order.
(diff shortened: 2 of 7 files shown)


ting-hong-shieh:
I filed #4560, so I ran your branch against the stubs to check the examples. Environment: pyrefly 1.2.0 and pyrefly-torch-stubs 1.2.0 from PyPI, Python 3.12, Linux.

**Checks clean:**

- the `Attention` example on the reference page
- the variadic form you introduced — `[Bs: IntTuple]` with `Tensor[[*Elements[Bs], D]]`
- the four rows of the shape-form table, and the `.shape` / `.size(0)` claims below them
- the three exposition examples on `tensor-shapes.mdx` (`Linear`, `custom_rand_tensor`, `mm`)
- the basics tutorial, reassembled into a full `BaselineActor` the way the page builds it up — all three `assert_type` calls hold

Your `mm` signature also matches the real stub's shape relation (the stub names the parameters `[N, K, M]` where you have `[M, K, N]`, but the relation is the same).

**One thing I think is worth fixing: the new names have no stated origin.**

After this change, five pages use `IntVar`, `IntTuple`, and `Elements`, but none of them say where those come from. The reference page's only sourcing is line 26, which covers `Int` alone. Copying the variadic snippet as printed gives:

```
ERROR Could not find name `IntTuple` [unknown-name]
ERROR Could not find name `Elements` [unknown-name]
```

This isn't a regression you introduced by accident so much as a side effect of the correct fix: the old spellings needed no import, since `*Bs` is native PEP 646 syntax and unbounded `[D]` needs nothing. The new ones do. Extending that line 26 sentence to cover all four names, or adding an import line to the snippets, would close it.

**Two smaller notes, take or leave:**

In the concepts section of `tensor-shapes.mdx`, "takes `Tensor[[*Elements[Bs], 3]]` as input" introduces `Bs` with no antecedent — it's the first mention on the page, and the variadic mechanism isn't explained until the reference page. The old `Tensor[..., 3]` read more naturally in prose even though it wasn't a real annotation.

The two heading renames change their anchors (`#tensord1-d2-`, `#variadic-dimensions-with-bs`). Nothing in the repo links to either, so no internal breakage — just noting it for external links. The old headings were wrong, so renaming them looks right to me.

**Scope:** I didn't reassemble every incremental fragment on all seven pages. The architectures, advanced, and loops examples I only checked structurally, not by running them.

*Disclosure, per the AI Usage section of CONTRIBUTING.md: these checks and this comment were produced by an AI agent (Claude Code) working in my checkout. I reviewed them before posting.*

Vishwaspatel2401:
Thanks for running the branch against the stubs — good catch on the missing sourcing.

Fixed the import gap: extended the reference page's existing Int[X] sentence to also cover IntVar, IntTuple, and Elements, so there's one place that says where all four names come from. Didn't add per-snippet import lines since that would break from how every other fragment on these pages is shown (none of them show imports, including Int before this change) — the reference page's sourcing sentence is the one place meant to answer that question, so extending it felt like the right fix rather than adding inconsistent per-snippet imports.

Also fixed the Bs-with-no-antecedent issue on tensor-shapes.mdx — rewrote that sentence in plain words instead of using the Elements[Bs] syntax before it's ever explained, and linked to the reference page's variadic section for the actual syntax.

Agreed on the anchor renames — confirmed the same thing, nothing in the repo links to either old anchor, so leaving the new headings as-is.

Pushed as a follow-up commit on the same branch.

stroxler (approved):

PR #4568, opened 2026-08-17
Keep the declared type when Any is assigned to an annotated variable

markselby9:
**Problem to fix:** Assigning an `Any` value to a variable with a declared type erased the annotation. `x: str = "hello"; x = f()` (where `f() -> Any`) made `x` become `Any`, which broke `assert_type` and silenced later checks on `x`.

**Approach:** `name_assign_infer` already kept the declared type when the *first* assignment after a bare annotation is `Any`; extend that to later reassignments too, since `Any` carries no information to narrow with. This matches pyright/mypy and the for-loop / `with as` paths that already preserve the declared type. Fixes 8 `bug`-marked tests. (Attribute, unpacking, and walrus targets are separate paths, left for follow-up.)


rchen152 (approved):

rchen152:
Nice find, thanks!

PR #4570, opened 2026-08-17
Fix false positive on callables with non-constant parameter mappings

### LINKED ISSUE #3587: False-positive [inconsistent-overload] with non-constant parameter mapping
### Describe the Bug

```python
from typing import overload

@overload
def f(string: str, /) -> None: ...
@overload
def f(lower: int, upper: int) -> None: ...
def f(
    lower_or_string: str | int | None = None,
    /,
    lower: int | None = None,
    upper: int | None = None,
) -> None: ...
```

The implementation of `f` is compatible with all legal call signatures of the overloads:

- `f(str)`
- `f(int, int)`
- `f(int, upper=int)`
- `f(lower=int, upper=int)`

yet pyrefly errors (xref: https://github.com/python/typing/issues/2224)

```
ERROR sandbox.py:6:5-6: Implementation signature `(lower_or_string: int | str | None = None, /, lower: int | None = None, upper: int | None = None) -> None`does not accept all arguments that overload signature `(lower: int, upper: int) -> None` accepts [inconsistent-overload]
```

<details><summary>Full MWE with actual implementation:
</summary>

```python
from typing import overload

@overload
def f(string: str, /) -> None: ...
@overload
def f(lower: int, upper: int) -> None: ...
def f(
    lower_or_string: str | int | None = None,
    /,
    lower: int | None = None,
    upper: int | None = None,
) -> None:
    if lower_or_string is None:
        assert isinstance(lower, int), "illegal call"
        assert isinstance(upper, int), "illegal call"
        print(f"({lower=}, {upper=})")
    elif isinstance(lower_or_string, int):
        if lower is None:
            assert isinstance(upper, int), "illegal call"
            lower = lower_or_string
        else:
            assert upper is None, "illegal call"
            upper = lower
            lower = lower_or_string
        print(f"({lower=}, {upper=})")
    else:
        assert isinstance(lower_or_string, str), "illegal call"
        assert lower is None, "illegal call"
        assert upper is None, "illegal call"
        string = lower_or_string
        print(f"({string=})")


f("a")
f(1, 2)
f(1, upper=2)
f(lower=1, upper=2)
```

</details>

### Sandbox Link

https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN24AbjEpRcqTAB100gAJCRYidMwwwdMAAo4DSqzaI6OygBo6AegCUdALQA%2BOgDlc6GIcIe5C0eKlY1GppiAO4ihqwMZgCuxMRhnOgM1vZOLm50HoQqAVrSdPl0ISIA%2BrzFxvqGxnQAPgn8dc6udAC8qa4meQXmnegFhbihlOGJte0wreO9-TFxw-VjTRNtS73JDkvunjLoICYgZJRqUKSEDLRQFADEdAAKpEdgJ0YYOAR0AMYukGxRlKgMCAuLLoG4AZRgEwAFgwGMQ4IhzOZDsdTrw2OYYOhzJhcB84OYvugfn8AUDsRpeHRUIJUNBUNhYJ9vhBfv9AS4BMQOeg4CCyAwoS4bMJKHByZNJCAAMyEACMACYpdIANoiahigC60ii6C4PD4MEwNkwECOH0BwkmAHJWeheDBrdJ7QwbEcAI5RM1GmwAaxgpBsqA%2BH3gcBtwVQlHQTt2AF99sHLTAAGLQGAUNBYPBEMggONAA

_No response_

alexander-beedie:
# Summary

Fixes #3587.

### Overview

A caller of a function does not have to pass its positional-or-keyword parameters positionally - it can pass _any_ prefix positionally, and the rest by name. 

When checking whether `got` is assignable to `want`, only the one mapping where _every_ such parameter of `want` is passed positionally was checked (which also requires the two signatures to agree on parameter names), so callables that satisfy `want` under a _different_ mapping were rejected.

### Solution

* Treat the parameter list check as a question about _all_ of `want`'s calling conventions rather than just _one_ of them. 
 
* If the as-written mapping fails, `is_subset_param_list` retries the check once per split, moving the positional/by-name boundary one parameter to the right each time, and accepts only if every split is assignable.
 
* Because Python forbids positional arguments after keyword arguments, the splits are a prefix chain, so there are only `n+1` of them (rather than `2^n`).

* The retry only runs when the as-written mapping fails and the comparison is not gradual, so accepting comparisons cost what they did before, and the splits speculatively pin type variables, so they run inside `Solver::with_snapshot` and roll back on failure 👍 

# Test Plan

* Existing unit tests all pass.
* Added 6-7 more tests (including the example given in the original Issue).

### DIFF
--- a/crates/pyrefly_types/src/callable.rs
+++ b/crates/pyrefly_types/src/callable.rs
@@ -916,6 +916,17 @@ impl Param {
         }
     }
 
+    pub fn passed_by_name(&self) -> Param {
+        match self {
+            // `Pos` becomes `KwOnly`, the form a caller passing it by name sees.
+            Param::Pos(name, ty, required) => {
+                Param::KwOnly(name.clone(), ty.clone(), required.clone())
+            }
+            // Every other variant is returned unchanged.
+            param => param.clone(),
+        }
+    }
+
     pub fn is_required(&self) -> bool {
         match self {
             Param::PosOnly(_, _, Required::Required)
--- a/pyrefly/lib/alt/functools.rs
+++ b/pyrefly/lib/alt/functools.rs
@@ -599,8 +599,8 @@ fn partial_residual(
         // A positional can't follow a bound keyword, so demote later positionals to keyword-only.
         if matches!(remaining[idx], Param::Pos(..)) {
             for later in remaining.iter_mut().skip(idx + 1) {
-                if let Param::Pos(n, t, r) = later {
-                    *later = Param::KwOnly(n.clone(), t.clone(), r.clone());
+                if matches!(later, Param::Pos(..)) {
+                    *later = later.passed_by_name();
                 }
             }
         }
--- a/pyrefly/lib/solver/solver.rs
+++ b/pyrefly/lib/solver/solver.rs
@@ -567,6 +567,13 @@ impl SubsetWithSnapshotResult {
     pub fn is_ok(&self) -> bool {
         matches!(self, SubsetWithSnapshotResult::Ok)
     }
+
+    pub fn into_result(self) -> Result<(), SubsetError> {
+        match self {
+            SubsetWithSnapshotResult::Ok => Ok(()),
+            SubsetWithSnapshotResult::Err(e) => Err(e),
+        }
+    }
 }
 
 impl Solver {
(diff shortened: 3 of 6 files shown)


alexander-beedie:
Rebased/resolved 👍

alexander-beedie:
@samwgoldman - rebased/resolved  again :)
(and added an extra test, covering a duplicate-argument acceptance case) 👌

PR #4571, opened 2026-08-17
Fix select/with_columns losing schema for keyword/positional args

### LINKED ISSUE #4565: Pyrefly loses Polars DataFrame schemas for positional `with_columns` and keyword `select`
### Describe the Bug

  `select` and `with_columns` support opposite expression forms. `select` preserves the schema for positional expressions but not keyword expressions, while `with_columns` does the reverse.

  ```python

  import polars as pl

  df = pl.DataFrame({"units": [12]})

  reveal_type(df.select(pl.col("units").alias("qty"))) # DataFrame[qty: Int64]
  reveal_type(df.with_columns(pl.col("units").alias("qty"))) # DataFrame
  reveal_type(df.select(qty=pl.col("units"))) # DataFrame
  reveal_type(df.with_columns(qty=pl.col("units"))) # DataFrame[units: Int64, qty: Int64]
  ```

  I would expect the equivalent forms to produce the same schemas:

  ```text
  select(...)       -> DataFrame[qty: Int64]
  with_columns(...) -> DataFrame[units: Int64, qty: Int64]
  ```

  The common positional form of `with_columns` therefore loses all downstream column checking.
  This also affects conditional positional expressions:

  ```python
  reveal_type(
      df.with_columns(
          pl.when(pl.col("units") > 10)
          .then(pl.lit("high"))
          .otherwise(pl.lit("low"))
          .alias("bucket")
      )
  ) # DataFrame
  ```

 Reproduced on `pyrefly 1.3.0-dev.1` with Polars 1.35.2.

Vishwaspatel2401:
# Summary

`select` and `with_columns` supported opposite expression forms. `select`
preserved the schema for positional expressions but not keyword ones,
while `with_columns` did the reverse — each function's schema-tracking
logic only ever processed one argument category (`args.args` or
`args.keywords`) and fell back to bare `DataFrame` on the other, even
though real Polars accepts a mix of `*exprs` and `**named_exprs` in both.

This also affected conditional positional expressions
(`pl.when(...).then(...).otherwise(...)`), for a second, independent
reason: `polars_expr_has_single_output` had no case for `when`/`then`/
`otherwise` chains, so `.alias(...)` on top of one never resolved a name,
regardless of argument position.

Fixes #4565. Fixes both problems:
- `polars_select` now also resolves `args.keywords`, using the same
  per-argument dtype resolution `polars_with_columns` already used for
  keywords. Its existing duplicate-name detection is extended to catch a
  keyword colliding with a positional alias, since Polars raises on that
  too.
- `polars_with_columns` now also resolves `args.args`, using the same
  per-argument name resolution `polars_select` already used for
  positional expressions (including its opacity/fallback handling for
  expressions with no well-defined output name, like a bare `pl.Series()`).
- `polars_expr_has_single_output` recognizes `.then(...)`/`.otherwise(...)`
  as always single-output, since a conditional expression evaluates to
  exactly one value per row regardless of the predicates or branches
  involved.

Not in scope: `with_columns` doesn't get the same positional/keyword
collision detection `select` now has — real Polars raises on that too,
but there was no existing mechanism there to extend, and the fallback
(later argument silently wins) isn't dangerous, just less precise than
`select`. Also, `when`/`then`/`otherwise` branch values resolve to
`Unknown` dtype rather than a fully inferred type — consistent with how
other unevaluable expressions already degrade elsewhere in this file, and
not something the issue asserted an expected dtype for.

### DIFF
--- a/pyrefly/lib/alt/polars_specials.rs
+++ b/pyrefly/lib/alt/polars_specials.rs
@@ -2114,6 +2114,13 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         // same as with_columns, so a sibling's new column is not visible.
         for kw in &args.keywords {
             let name = kw.arg.as_ref()?.id.clone();
+            // A keyword value that itself resolves to multiple columns (e.g.
+            // `x=pl.col("a", "b")`) is a duplicate-column error in Polars, not a single
+            // named output — degrade the whole call, same as an opaque positional arg.
+            if !self.polars_expr_has_single_output(&kw.value) {
+                has_opaque = true;
+                continue;
+            }
             let resolved = match self.polars_column_arg(&kw.value) {
                 ColumnArg::Named(col) => resolve_column(schema, &col, kw.value.range(), errors),
                 ColumnArg::Opaque => None,
@@ -2327,10 +2334,11 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         match expr {
             Expr::Call(call) => {
                 if let Expr::Attribute(attr) = &*call.func {
-                    // A `when(...).then(...)` / `.otherwise(...)` chain always evaluates to
-                    // one value per row, regardless of the predicates or branch values.
+                    // A `when(...).then(...)` / `.otherwise(...)` chain evaluates to one value
+                    // per row, unless a branch value itself resolves to multiple columns (e.g.
+                    // `pl.col("a", "b")`), in which case the chain inherits that multiplicity.
                     if matches!(attr.attr.id.as_str(), "then" | "otherwise") {
-                        return true;
+                        return self.polars_when_chain_has_single_output(expr);
                     }
                     if PolarsExprMethod::parse(attr.attr.id.as_str()).is_some() {
                         return self.polars_expr_has_single_output(&attr.value);
@@ -2359,6 +2367,26 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             | Expr::BytesLiteral(_)
             | Expr::NoneLiteral(_) => true,
             _ => !self.is_polars_expr_value(expr),
+        }
+    }
+
+    /// Check that every branch value in a `when(...).then(...)[.when(...).then(...)]*
+    /// .otherwise(...)` chain is itself single-output. The chain's own `when`/`then`/
+    /// `otherwise` calls are a builder, not a data value, so only the branch *arguments*
+    /// need checking — an unrecognized receiver (e.g. the base `pl.when(...)` call) is
+    /// trivially fine since it isn't a data value either.
+    fn polars_when_chain_has_single_output(&self, expr: &Expr) -> bool {
+        let Expr::Call(call) = expr else { return true };
+        let Expr::Attribute(attr) = &*call.func else {
+            return true;
+        };
+        match attr.attr.id.as_str() {
+            "then" | "otherwise" => {
+                matches!(&call.arguments.args[..], [value] if self.polars_expr_has_single_output(value))
+                    && self.polars_when_chain_has_single_output(&attr.value)
+            }
+            "when" => self.polars_when_chain_has_single_output(&attr.value),
+            _ => true,
         }
     }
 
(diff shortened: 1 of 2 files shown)


pyrefly/lib/alt/polars_specials.rs
```diff
@@ -2294,10 +2326,15 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
             Expr::Call(call) => {
-                if let Expr::Attribute(attr) = &*call.func
-                    && PolarsExprMethod::parse(attr.attr.id.as_str()).is_some()
-                {
-                    return self.polars_expr_has_single_output(&attr.value);
+                if let Expr::Attribute(attr) = &*call.func {
+                    // A `when(...).then(...)` / `.otherwise(...)` chain always evaluates to
+                    // one value per row, regardless of the predicates or branch values.
+                    if matches!(attr.attr.id.as_str(), "then" | "otherwise") {
+                        return true;
```
shobhitmehro: could we avoid treating every then/otherwise chain as a single output here? for example:
```
df = pl.DataFrame({"a": [1], "b": [2]})
df.select(
pl.when(pl.col("a") > 0)
.then(pl.col("a", "b"))
.otherwise(0)
.alias("x"))
```
polars expands this into two outputs named x and raises a duplicate-column error, while pyrefly infers one tracked x column. pyrefly should report duplicate-column when multiple outputs are statically known,
and otherwise return plain dataframe.

pyrefly/lib/alt/polars_specials.rs
```diff
@@ -2097,12 +2110,31 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
+                ColumnArg::Named(col) => resolve_column(schema, &col, kw.value.range(), errors),
+                ColumnArg::Opaque => None,
+                ColumnArg::Expr => self
+                    .eval_polars_expr(&kw.value, schema, errors)
+                    .map(ExprValue::dtype),
+            };
+            if resolved.is_some() && !resolved_names.insert(name.clone()) {
+                report_duplicate_column(&name, kw.value.range(), errors);
+            }
+            columns.push((name, resolved.unwrap_or(PolarsDType::Unknown)));
```
shobhitmehro, line 2134: could we handle multi output keyword expressions here? `df.select(x=pl.col("a", "b"))` raises a duplicate-column error in polars right now

pyrefly/lib/alt/polars_specials.rs
```diff
@@ -2421,23 +2458,56 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
...
-            return None;
-        }
         if schema.kind != DataFrameKind::Polars {
             return None;
         }
-        // Validate names before inference so fallback does not duplicate diagnostics.
+        let positional = args
+            .args
+            .iter()
+            .flat_map(positional_elements)
```
shobhitmehro: `df.with_columns(pl.col("a").alias("x"), [1])` produces a `List(Int64)` column in polars, but pyrefly currently infers `literal: Int32`. since nested polars dtypes are unmodeled, could we preserve [1] as a single expression and infer Unknown or fall back instead of flattening it? i can handle modeling the list dtype separately.

shobhitmehro (changes requested):
thanks for the PR! I left a few comments with suggested fixes. Let me know if you have questions or need more context.

Vishwaspatel2401:
Thanks @shobhitmehro  for the thorough review — all three were real gaps, verified each against actual Polars behavior before fixing (not just reasoning about it).

The when/then/otherwise multi-output issue: added a polars_when_chain_has_single_output helper that walks the chain and checks every branch value instead of assuming true unconditionally. pl.when(...).then(pl.col("a","b")).otherwise(0).alias("x") now falls back to plain DataFrame. One thing worth flagging: I implemented the "otherwise return plain dataframe" half of your suggestion, not the "report duplicate-column when statically known" half — that would need threading errors/ErrorCollector into has_single_output/output_name, which don't currently take it, so it felt like a bigger change than this fix warranted. Happy to add the full diagnostic if you'd rather have it now instead of just the safe fallback.

The keyword version (x=pl.col("a", "b")) — same fix, select's keyword loop now checks has_single_output on the value first and degrades the whole call the same way an opaque positional arg already does.

The list literal one turned out more interesting than I expected — I tested it directly against real Polars and found the behavior is actually context-dependent: a bare list is only Polars' "sequence of exprs" shorthand when it's the sole positional arg. with_columns([1]) alone does flatten to an anonymous literal column, but with_columns(expr, [1]) treats [1] as one opaque value (a List-dtype column in real Polars). Fixed the flattening condition to only apply when there's exactly one positional arg, and extended the existing anonymous-literal naming to cover list/tuple displays so it still gets Polars' actual default name ("literal") with an honest Unknown dtype instead of the wrong flattened Int32.

Added regression tests for all three matching your exact examples. Full suite still passes (7891 tests, 0 failures), pushed to the same branch.

Vishwaspatel2401:
Hi @shobhitmehro, the Windows CI failure looks like a runner infra issue, not a real test failure — the annotation says "runner lost communication with the server," and it died mid-cargo test --release with no actual assertion failure. macOS and Ubuntu both passed. I don't have permission to re-run it myself — could you kick it off again when you get a chance?

shobhitmehro:
sorry for the delay, thanks for the fixes! this looks good to me now- ill cc @stroxler so he can take a look/accept!

stroxler (approved):

PR #4579, opened 2026-08-18
Run conformance tests in GitHub CI (#4579)

stroxler:
Summary:

The typing conformance suite only runs inside Meta, via the `conformance_output_test` Buck target. On GitHub nothing invokes it, so a PR that changes pyrefly's conformance output goes green while leaving the checked-in `conformance.exp` stale. The mismatch only surfaces later, when the change is imported internally and fails there.

Add a step to the `pyrefly` workflow that runs `conformance_output.py --mode check` against the release binary that job already built, matching what internal CI checks. The step is restricted to Linux: the generated output does not vary by platform, so one OS suffices, and the check cannot run on Windows because it keys results by splitting paths on `/`.

The check-mode failure message told the reader to re-generate the output with `buck2`, which a GitHub contributor has no way to run. It now branches on whether `--executable` was passed, since that flag is precisely the signal that the caller is outside Buck, and prints the corresponding cargo-based command.

Reviewed By: ndmitchell, rchen152

Differential Revision: D116394357

### DIFF
--- a/.github/workflows/pyrefly.yml
+++ b/.github/workflows/pyrefly.yml
@@ -46,6 +46,11 @@ jobs:
     - run: cargo clippy --release
     - run: cargo build --release
     - run: cargo test --release
+    - name: Check conformance output is up to date
+      # The generated output is platform-independent, so one OS is enough. It must not
+      # be Windows: the check keys results by splitting paths on "/".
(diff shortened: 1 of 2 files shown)


ndmitchell (approved):

rchen152 (approved):

PR #4593, opened 2026-08-19
Fix overrides of callable ClassVars

### LINKED ISSUE #4557: False positive when overriding callable `ClassVar`
### Describe the Bug

There seems to be a false positive in the following code (using `pyrefly` 1.2.0):

```python
from typing import ClassVar

def foo(s: str) -> None:
    pass

class Parent:
    val: ClassVar[Callable[[str], None]]

class Child(Parent):
    # ERROR Class member `Child.val` overrides parent class `Parent` in an inconsistent manner [bad-override]`.
    val: ClassVar[Callable[[str], None]] = foo
```

It also errors when `ClassVar` is removed from `Child.val`'s annotation, even though that should be OK too.

It does not error when the parent class sets a value, or when the type is not a `ClassVar`, or when it is a `ClassVar` wrapping a simpler type like `int`.

### Sandbox Link

_No response_

_No response_

fangyi-zhou:
# Summary

Fixes #4557.

`ClassVar[Callable]` fields use the method representation so instance access models function descriptor binding. That conversion previously discarded the `ClassVar` qualifier, which made compatible child declarations look like instance variables.

Preserve the qualifier on method fields and compare callable ClassVar overrides through their unbound class-level types, before method binding removes the first parameter. Compatible bare and parameterized callables are accepted, while the tests confirm that:

- an explicit child annotation without `ClassVar` remains an invalid instance-variable override;
- an unannotated child assignment is accepted;
- an incompatible callable signature is still rejected.

# Test Plan

- `cargo test test_override_class_var_callable`
- `cargo test test::class_overrides` (125 passed)
- `cargo test classvar` (21 passed)
- `python3 test.py --no-test`: formatting, linting, and conformance passed; JSON schema validation was rerun after installing its documented Python dependencies (56 schema tests and 3 SARIF tests passed).
- Full `python3 test.py` Rust suite: 7,940 passed and 4 ignored. Four IPC transport tests could not create local endpoints in the sandbox and failed with `Operation not permitted`.

An exploratory generated matrix also covered scalars, `Any`, containers and collection ABCs, unions, `Literal`, non-callable `Annotated`, custom and generic classes, protocols, `type`, `NewType`, aliases, callback protocols, optional callables, descriptors, properties, static methods, class methods, `None`, `Never`, `NoReturn`, and `LiteralString`. These produced no additional significant behavior. Callable aliases and `Annotated[Callable]` exercised the same binding issue addressed by this fix, so only the representative cases were retained as regression tests.

Tensor-shape tests were not run; they are opt-in and unrelated to this change.

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -337,6 +337,7 @@ enum ClassFieldInner {
         ty: Type,
         is_abstract: bool,
         is_function_without_return_annotation: bool,
+        is_classvar: bool,
     },
     /// A method whose instance attribute type is resolved from another method on the receiver.
     ProxyMethod { target: Name, ty: Type },
@@ -510,7 +511,7 @@ impl ClassField {
 
     fn new_synthesized_inner(ty: Type, is_classvar: bool) -> Self {
         // Detect if this is a property and construct the appropriate variant.
-        // Properties and methods are never ClassVars.
+        // Properties are never ClassVars.
         if ty.is_property_getter() || ty.is_property_setter_with_getter().is_some() {
             ClassField(
                 ClassFieldInner::Property {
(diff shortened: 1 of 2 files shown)


stroxler (approved):

Issue #4306, opened 2026-07-27
`pyrefly.lspPath` should detect the version from the current python environment by default


DetachHead:
by default, pyrefly uses the version bundled with vscode instead of the version installed in the current python environment. most of the time this is not what the user wants, and can lead to the IDE reporting different errors than the CLI, which can be very confusing to the user.

the bigger problem though is that `pyrefly.lspPath` is not cross-platform, meaning i cannot commit this in my `.vscode/settings.json`:

```jsonc
{
    "pyrefly.lspPath": ".venv/scripts/pyrefly.exe" // does not work for contributors that use linux or macos
}
```

most other extensions use an `importStrategy` setting to workaround this problem:

- https://docs.astral.sh/ruff/editors/settings/#importstrategy
- https://docs.basedpyright.com/latest/benefits-over-pyright/pypi-package-vscode-pinning/#ability-to-pin-the-version-used-by-vscode
- https://github.com/microsoft/vscode-pylint#usage-and-features

DetachHead:
it's worth noting that vscode's python environment settings is in a horrible state atm, which is affecting the aforementioned extensions:

- https://github.com/microsoft/vscode-pylint/issues/856
- https://github.com/astral-sh/ruff-vscode/issues/1114
- https://github.com/DetachHead/basedpyright/issues/1855

DetachHead:
the workaround i typically use for other extensions that have this problem is to create wrapper scripts in `.vscode`. for example:

`.vscode/pyrefly.bat`:
```bat
@echo off
".venv/Scripts/pyrefly.exe" %*
```

`.vscode/pyrefly`:
```bash
#!/bin/bash
./.venv/bin/pyrefly "$@"
```

`.vscode/settings.json`:
```json
{
    "pyrefly.lspPath": ".vscode/pyrefly"
}
```

that works in the dprint extension (see https://github.com/dprint/dprint/issues/859#issuecomment-3194102785) but it doesn't seem to work in pyrefly:

```
[Error - 5:09:24 PM] Pyrefly language server client: couldn't create connection to server.
Launching server using command .vscode/pyrefly failed. Error: spawn .vscode/pyrefly ENOENT
```

another workaround i've used with other extensions where the previous method doesn't work is a [polyglot script](https://en.wikipedia.org/wiki/Polyglot_(computing)):

`.vscode/pyrefly.bat`:
```cmd
: <<'BATCH'
@echo off
".venv/Scripts/pyrefly.exe" %*
[...]

PR #4610, opened 2026-08-20
Relative binary fix

connernilsen:
When Pyrefly would execute binaries passed in from VSCode through the `pyrefy.lspPath` setting, it would run the command without any modifications, even for relative paths. This means any relative binary definitions (or wrapper scripts) would be run from whatever the cwd of the VSCode extension would be, typically resulting in an ENOENT (or occasionally an EINVAL).

Instead, we should do at least a basic check to see if the path is relative, and prepend the workspace path onto it. Since VSCode allows multi-workspace roots, this means taking the first (a.k.a. the primary) workspace root, and using that as our cwd.

This should fix the main problem with #4306, but there will be other changes I'd like to make here to make Pyrefly binary selection match user expectations a bit more.

### DIFF
--- a/lsp/package.json
+++ b/lsp/package.json
@@ -144,7 +144,7 @@
                 "pyrefly.lspPath": {
                     "type": "string",
                     "default": "",
-                    "description": "The path to the binary used for the lsp",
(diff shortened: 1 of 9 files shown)


stroxler (approved):

PR #4614, opened 2026-08-21
fix maybe support parameter inlayhint switch in sandbox #4613

### LINKED ISSUE #4613: maybe support parameter inlayhint switch in sandbox
Try reproduce https://github.com/facebook/pyrefly/issues/4611 but with no luck, so sandbox may be a good way.

asukaminato0721:
# Summary

Fixes #4613

Add a switch to the website.

I don't know the proper layout so I just randomly put it on the right corner.

# Test Plan

add mock test

### DIFF
--- a/website/src/sandbox/Sandbox.tsx
+++ b/website/src/sandbox/Sandbox.tsx
@@ -60,7 +60,7 @@ export interface PyreflyState {
     autoComplete: (line: number, column: number) => any;
     gotoDefinition: (line: number, column: number) => monaco.IRange[] | null;
     hover: (line: number, column: number) => any;
-    inlayHint: () => any;
+    inlayHint: (callArgumentNames: boolean) => any;
     semanticTokens: (range: any) => any;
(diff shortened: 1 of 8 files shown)


website/src/sandbox/Sandbox.tsx
```diff
@@ -740,7 +752,7 @@ export default function Sandbox({
         );
         setInlayHintFunctionForMonaco(
             model,
-            () => pyreService?.inlayHint() || []
+            () => pyreService.inlayHint(showParameterHints) || []
```
stroxler: I don't think you meant to drop the `?` here(?)

stroxler (changes requested):
I think this looks reasonable, except for the probably-accidental guard against null `pyreService`

stroxler:
Thanks! Did you rebase? The import job failed, which *usually* means a failure to rebase against trunk

stroxler (approved):

PR #4615, opened 2026-08-21
Respect skip-interpreter-query for client-provided pythonPath in the LSP

### LINKED ISSUE #4445: LSP does not respect `skip-interpreter-query = true` in VSCode workspace.
### Describe the Bug

This one was a bit weird to figure out. It looks pretty niche, but somewhere along the way the LSP is not respecting `skip-interpreter-query = true`.

**Setup:**
```
$ tree
.
├── example.py
├── pyrefly.code-workspace
└── pyrefly.toml
```

`example.py` (contents are irrelevant):
```
x: int = 5
x = "str"
```

`pyrefly.code-workspace`
```
{
  "folders": [
    {
      "path": "."
    }
  ],
  "settings": {
    "python-envs.defaultEnvManager": "ms-python.python:pyenv"
  },
  "extensions": {
    "recommendations": [
      "ms-python.python"
    ]
  }
}
```

`pyrefly.toml`
```
preset = "all"
skip-interpreter-query = true
python-version = "3.14.2"
site-package-path = [
    "/home/dillon/.pyenv/versions/3.14.2/lib/python3.14/site-packages"
]
python-platform = "linux"
```

Important, my selected Python interpreter is a deleted venv with `pyenv`. I replicated it with the following:
First, ensure the workspace is opened. Click on `pyrefly.code-workspace` and select `Open Workspace` in the bottom.

Next, create the to-be-deleted venv.
```
pyenv virtualenv deleted-venv
```
Then `ctrl + shift + p` -> `Python: Select Interpreter` -> `deleted-venv (3.14.2)`.

Then delete the venv.

```
pyenv virtualenv-delete deleted-venv
```

Now when I open `example.py` and run `ctrl + shift + p` -> `Pyrefly: Restart Pyrefly Client`, I see the following output:
```
ERROR Failed to query interpreter at /home/dillon/.pyenv/versions/deleted-venv/bin/python, falling back to default Python environment settings
No such file or directory (os error 2)
```

This is a sign to me that `pyrefly` is still trying to query the Python interpreter even though I have `skip-interpreter-query = true` in `pyrefly.toml`. I am unsure if it is querying it all the time, but I would expect it not to if I provide `skip-interpreter-query`, `python-version`, `python-platform` and `site-package-path`.

Trace log:

[Pyrefly_language_server_trace.log](https://github.com/user-attachments/files/30721228/Pyrefly_language_server_trace.log)

I can see the following set. 
```
"pythonPath": "/home/dillon/.pyenv/versions/deleted-venv/bin/python"
```

### Sandbox Link

_No response_

Extension version `1.1.9003`. See above for workspace setup.

Version: 1.131.0 (user setup)
Commit: e4c7e7b1d6d060162f4aa7f8225271b67ce1df75
Date: 2026-07-28T10:51:25Z
Electron: 42.7.0
ElectronBuildId: 14710161
Chromium: 148.0.7778.280
Node.js: 24.18.0
V8: 14.8.178.38-electron.0
OS: Windows_NT x64 10.0.19045

rootkiller6788:
## Summary

Fixes #4445: with `skip-interpreter-query = true` in `pyrefly.toml`, the LSP still queried the interpreter the client supplied via `pythonPath` — logging e.g. `ERROR Failed to query interpreter at <deleted venv>...` and then applying that interpreter on top of the config, defeating the opt-out.

## Root cause

`PythonInfo::new` queried the interpreter eagerly in `update_pythonpath`, i.e. as soon as the client sent `pythonPath`, before any config file had been resolved. The `skip-interpreter-query` flag only lives on the `ConfigFile`, so it was never consulted. The configurer then applied the pre-queried environment unconditionally (when the config hadn't picked an interpreter).

## Fix

- `PythonInfo` now stores only the interpreter path; the query is deferred to the point where the workspace config is applied (`WorkspaceConfigConfigurer::configure`).
- The client interpreter is applied only when the resolved config has no interpreter selection **and** has not opted out of queries (`skip-interpreter-query`).
- A config with `skip-interpreter-query = true` now never performs (or pays for) an interpreter query, and the client `pythonPath` is ignored in favor of the config's own environment settings.

## Tests

- `test_skip_interpreter_query_ignores_lsp_pythonpath` (unix-gated LSP interaction test): opens a project whose config sets `skip-interpreter-query = true`, verifies an unresolved import stays an error even after the client sends a `pythonPath` that *would* resolve it.
- `test_skip_interpreter_query_blocks_client_pythonpath` (cross-platform unit test in `workspace.rs`): pins that a `skip-interpreter-query` config ignores a client interpreter while a config silent about interpreters still applies it.

Verified with `cargo +1.96.1 test`: all LSP interaction tests (1081) and the `workspace`/`configuration` modules pass.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/workspace.rs
+++ b/pyrefly/lib/lsp/non_wasm/workspace.rs
@@ -42,26 +42,16 @@ use crate::state::lsp::ImportFormat;
 use crate::state::lsp::InlayHintConfig;
 use crate::state::lsp::TypeCheckingMode;
 
-/// Information about the Python environment provided by this workspace.
+/// A Python interpreter path provided by the client (e.g. via `pythonPath`).
 #[derive(Debug, Clone)]
 pub struct PythonInfo {
-    /// The path to the interpreter used to query this `PythonInfo`'s [`PythonEnvironment`].
+    /// The path to the Python interpreter.
     interpreter: PathBuf,
-    /// The [`PythonEnvironment`] values all [`ConfigFile`]s in a given workspace should
-    /// use if no explicit [`ConfigFile::python_interpreter`] is provided, or any
-    /// `PythonEnvironment` values in that `ConfigFile` are unfiled. If the `interpreter
-    /// provided fails to execute or is invalid, this `PythonEnvironment` might instead
-    /// be a system interpreter or [`PythonEnvironment::pyrefly_default()`].
-    env: PythonEnvironment,
 }
 
 impl PythonInfo {
     pub fn new(interpreter: PathBuf) -> Self {
-        let (env, query_error) = PythonEnvironment::get_interpreter_env(&interpreter);
-        if let Some(error) = query_error {
-            error!("{error}");
-        }
-        Self { interpreter, env }
+        Self { interpreter }
     }
 }
 
@@ -172,12 +162,19 @@ impl ConfigConfigurer for WorkspaceConfigConfigurer {
                     config.fallback_search_path =
                         FallbackSearchPath::Explicit(Arc::new(new_fallback_search_path));
                 }
-                if let Some(PythonInfo {
-                    interpreter,
-                    mut env,
-                }) = w.python_info.clone()
+                // Use the client-provided interpreter (e.g. `pythonPath`) only when the
+                // config doesn't already specify an interpreter and hasn't opted out of
+                // interpreter queries entirely (`skip-interpreter-query`). The query is
+                // deferred until here so an opt-out config never pays for it.
+                if let Some(PythonInfo { interpreter }) = w.python_info.clone()
                     && config.interpreters.is_empty()
+                    && !config.interpreters.skip_interpreter_query
                 {
+                    let (mut env, query_error) =
+                        PythonEnvironment::get_interpreter_env(&interpreter);
+                    if let Some(error) = query_error {
+                        error!("{error}");
+                    }
                     let site_package_path: Option<Vec<PathBuf>> =
                         config.python_environment.site_package_path.take();
                     env.site_package_path = site_package_path;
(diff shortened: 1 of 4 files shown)


connernilsen:
Hey @rootkiller6788, thanks for working on this, I'll try to take a look in the next week or two. For future reference, please claim the issue you're working on so that we can make sure others don't accidentally start working on it.

connernilsen:
Oh wait, actually can you claim #4445?

rootkiller6788:
Thanks for the note, and sorry for not claiming the issue up front. I've just claimed #4445 and linked this PR there, so hopefully nobody else starts on it in the meantime.

For context, the fix defers the interpreter query until the workspace config is applied, so a config with `skip-interpreter-query = true` never queries the client-provided `pythonPath` at all. When you get a chance to look, could you please take another look?

pyrefly/lib/lsp/non_wasm/workspace.rs
```diff
@@ -172,12 +162,19 @@ impl ConfigConfigurer for WorkspaceConfigConfigurer {
...
-                    interpreter,
-                    mut env,
-                }) = w.python_info.clone()
+                // Use the client-provided interpreter (e.g. `pythonPath`) only when the
+                // config doesn't already specify an interpreter and hasn't opted out of
+                // interpreter queries entirely (`skip-interpreter-query`). The query is
+                // deferred until here so an opt-out config never pays for it.
+                if let Some(PythonInfo { interpreter }) = w.python_info.clone()
                     && config.interpreters.is_empty()
+                    && !config.interpreters.skip_interpreter_query
```
kinto0: this seems like the entire fix. what do the other changes do?

kinto0 (changes requested):
thanks for the fix! looks like a few formatting changes are needed

kinto0:
would you be able to test this manually too? 

I might actually expect to still see the `ERROR Failed to query interpreter at /home/dillon/.pyenv/versions/deleted-venv/bin/python, falling back to default Python environment settings` log because we might query the interpreter in the "default" workspace regardless (the one we use for files not covered by these configs)

pyrefly/lib/lsp/non_wasm/workspace.rs
```diff
@@ -42,26 +42,16 @@ use crate::state::lsp::ImportFormat;
 use crate::state::lsp::InlayHintConfig;
 use crate::state::lsp::TypeCheckingMode;
 
-/// Information about the Python environment provided by this workspace.
+/// A Python interpreter path provided by the client (e.g. via `pythonPath`).
 #[derive(Debug, Clone)]
 pub struct PythonInfo {
```
kinto0: now this struct only has one item in it. I think we can remove it and end up with just interpreter

rootkiller6788:
Thanks for the review, and thanks for digging into the default-workspace question.

I've pushed fixes for both points:

1. **Formatting** – the `cargo fmt -- --check` step was what all three `test` jobs were failing on. I ran `cargo fmt` over the two touched files and the tree is clean now under stable rustfmt (commit da7fb34).

2. **Removing `PythonInfo`** – agreed, with the interpreter query moved into `configure()` the struct was only wrapping a `PathBuf`. I dropped it and store `client_interpreter: Option<PathBuf>` on `Workspace` directly (commit e6cca7c).

3. **Re: the remaining ERROR log / default workspace** – I think that log now only fires where it's supposed to: for a config that doesn't opt out (or for a file with no `pyrefly.toml` at all), the client `pythonPath` is still queried and applied, because that's the point of `pythonPath` as a fallback. `test_interpreter_change_removes_type_errors` pins exactly that path. For a config with `skip-interpreter-query = true`, the pythonPath is neither queried nor applied even when it's served through the default workspace: the new `test_skip_interpreter_query_ignores_lsp_pythonpath` opens a file whose `pyrefly.toml` opts out, sends a `pythonPath` that *would* resolve the import, and asserts the import error persists. The workspace unit test `test_skip_interpreter_query_blocks_client_pythonpath` covers the same gate on the default workspace directly.

On manual testing: I don't have an interactive editor available here, but the interaction test drives the same sequence as the repro (open file → diagnostics → `didChangeConfiguration` with a resolving `pythonPath` → diagnostics unchanged). If you'd still like it verified against a live VSCode session with the deleted-venv setup, I can follow up once this lands.

Could you please take another look?

kinto0 (approved):

PR #4621, opened 2026-08-21
Return the SubsetError from `check_type_with_options`

rchen152:
# Summary

Returns the SubsetError so that callers can examine it.

I converted the `Result<(), SubsetError>` returned from `is_subset_eq` to an `Option` because callers are forced to consume `Result`s, but `check_type_with_options` already handles the error, so it doesn't need to be explicitly consumed in the common case.

# Test Plan

`./test.py`

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


stroxler (approved):

PR #4624, opened 2026-08-22
Fixes lint warning about `match { ... }` where `matches!` would do

rchen152:
# Summary

As title, fixes lint warning

# Test Plan

Format, build, and error_kind tests are clean


stroxler (approved):

PR #4625, opened 2026-08-22
Fix broken website link.

rchen152:
# Summary

Our website build was broken due to a race condition between a change that added a link to the mypy migration docs and a change that moved them to a new location. This fixes the now-stale link.

# Test Plan

Website build is green on this branch.


samwgoldman (approved):

PR #4627, opened 2026-08-22
fix assert callable narrows too much #911

### LINKED ISSUE #911: `assert callable` narrows too much

asukaminato0721:
# Summary

Fixes #911

callable(x) now narrows unknown values to `Callable[..., Any]`, while preserving known signatures.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_types/src/function.rs
+++ b/crates/pyrefly_types/src/function.rs
@@ -415,6 +415,7 @@ impl FuncSymbol {
 pub enum FunctionKind {
     IsInstance,
     IsSubclass,
+    Callable,
     /// The builtin `len`. Special-cased so that when the argument's `__len__`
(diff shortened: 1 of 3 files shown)


stroxler (approved):

PR #4629, opened 2026-08-22
Fix sticky unresolvable __all__ state

### LINKED ISSUE #4563: Unresolvable `__all__` is not "sticky"
### Describe the Bug

This reports `ERROR __all__ could not be statically analyzed`

```
__all__ = []
__all__.remove("a")
for name in ["a"]:
    __all__.append(name)
```

But the following does not:

```
__all__ = []
for name in ["a"]:
    __all__.append(name)
__all__.remove("a")
```

Every `__all__` mutation was resetting the state to `Specified` in definitions before checking whether the mutation was actually analyzable, so a later `remove`, `append`, or `extend` would cause it to become `DunderAllKind::Specified`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.fZQxDoMwDEWvYsFCF7z3FN1JhaJi0Q5NUJL2_P12w4AQrFbs2O9L70iZ44hzNQsasLQS13CU-OAa75p7JVHf9VCHhKnTR4C_VpO84anOOi57YC3d_gWqf0MVID1_ICQQ7F1o4SKhZylLvjKv7THNLIGn-Mi86WDSTf3Xv0xz23EULeCMsSdJLUmyFCvhBkN9xu8H

_No response_

tkim602:
# Summary

Fixes #4563

keep an unresolvable `__all__` state from being reset by later mutations and preserve `DunderAllKind::Unresolvable` across subsequent `append`, `extend`, and `remove` calls so star imports continue to use the fallback behavior.

# Test Plan

Added regression coverage for:

- keeping `DunderAllKind::Unresolvable` sticky across later `__all__` mutations
- preserving star-import fallback behavior after an unresolvable `append`

Ran:

```sh
cargo test test_all_unresolvable_is_sticky_across_mutations -- --nocapture
cargo test test_all_ -- --nocapture
cargo fmt --check
git diff --check
```

### DIFF
--- a/pyrefly/lib/export/definitions.rs
+++ b/pyrefly/lib/export/definitions.rs
@@ -798,7 +798,9 @@ impl DefinitionsBuilder {
                     && arguments.keywords.is_empty()
                     && !self.in_main_guard
                 {
-                    self.inner.dunder_all.kind = DunderAllKind::Specified;
+                    if !matches!(self.inner.dunder_all.kind, DunderAllKind::Unresolvable(_)) {
+                        self.inner.dunder_all.kind = DunderAllKind::Specified;
+                    }
                     match attr.as_str() {
                         "extend" => match DunderAllEntry::as_list(&arguments.args[0]) {
                             Some(mut entries) => {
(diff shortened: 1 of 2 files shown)


rchen152 (approved):
Looks good to me, thanks!

stroxler (approved):

PR #4630, opened 2026-08-22
fix Django REST Framework: false positives on serializer declared fields (named like Field attrs) and inner Meta #3933

### LINKED ISSUE #3933: Django REST Framework: false positives on serializer declared fields (named like Field attrs) and inner Meta
### Summary

(Follow-up to #168, which was closed with a request to file specific Django/DRF bugs as their own issues.)

Pyrefly emits false positives on standard **Django REST Framework** serializers. DRF's `SerializerMetaclass` collects declared fields into `_declared_fields` and consumes the inner `Meta` class — dynamic behavior that isn't modeled — so two very common, valid patterns are flagged.

### Repro

`pyrefly.toml`:
```toml
project-includes = ["m.py"]
```

`m.py`:
```python
from rest_framework import serializers


class FieldNameSerializer(serializers.Serializer):
    label = serializers.CharField()      # (1) bad-assignment
    source = serializers.CharField()     # (1) bad-assignment


class MetaSerializer(serializers.ModelSerializer):
    class Meta:                           # (2) bad-override
        fields = "__all__"
```

`pyrefly check` (v1.0.0, djangorestframework-stubs 3.17.0):
```
ERROR `CharField` is not assignable to attribute `label` with type `_StrPromise | str | None` [bad-assignment]
ERROR `CharField` is not assignable to attribute `source` with type `str | None` [bad-assignment]
ERROR Class member `MetaSerializer.Meta` overrides parent class `ModelSerializer` in an inconsistent manner [bad-override]
```

### Why these are false positives

1. **Declared field named like an inherited `Field` attribute** (`label`, `source`, and also `context`/`data`). DRF's metaclass moves declared fields into `_declared_fields`; `label = CharField()` is a serializer field, not an override of `rest_framework.fields.Field.label: str | None`. Fields with these names are valid and common.
2. **Inner `class Meta`.** djangorestframework-stubs declares `ModelSerializer.Meta` as a concrete nested class, but a user's `class Meta` does not (and cannot) subclass it — it is consumed by the metaclass, not a nominal override. (The stub itself carries a `# type: ignore[valid-type]` on that `Meta`.)

Both are exactly what mypy's `mypy_drf_plugin` special-cases; mypy + the plugin reports no errors here.

### Expected

No errors for either pattern.

### Environment
- pyrefly 1.0.0 (also reproduced on 1.1.1)
- djangorestframework-stubs 3.17.0, djangorestframework 3.x
- Python 3.12

asukaminato0721:
# Summary

Fixes #3933

DRF Serializer subclasses now recognize metaclass-consumed Field declarations, avoiding inherited-attribute false positives.

Nested Meta classes on serializers no longer trigger bad-override.

# Test Plan

add test

### DIFF
--- a/crates/pyrefly_python/src/module_name.rs
+++ b/crates/pyrefly_python/src/module_name.rs
@@ -309,6 +309,10 @@ impl ModuleName {
         Self::from_str("marshmallow.schema")
     }
 
+    pub fn rest_framework_fields() -> Self {
+        Self::from_str("rest_framework.fields")
+    }
+
     pub fn rest_framework_serializers() -> Self {
         Self::from_str("rest_framework.serializers")
     }
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2969,6 +2969,38 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         } else {
             None
         };
+        if !inferred_from_method
+            && direct_annotation.is_none()
+            && self
+                .get_metadata_for_class(class)
+                .is_django_rest_framework_serializer()
+            && let ExprOrBinding::Expr(e) = value
+        {
+            let inferred_ty = self.attribute_expr_infer(e, None, name, &self.error_swallower());
+            let is_declared_field = match &inferred_ty {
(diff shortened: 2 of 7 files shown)


stroxler (approved):

PR #4634, opened 2026-08-23
Guard against recursive protocol member checking in __getattr__ self validation

### LINKED ISSUE #4632: Stack overflow with `__getattr__`
### Describe the Bug

```python
from collections.abc import Awaitable

class C:
    def __getattr__(self: Awaitable):
        pass

C().a
```

```
thread '<unknown>' (149109) has overflowed its stack
```

### Version info
Reproducible with 1.3.0-dev.2 and commit 19d136cf

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.w1tk4ihFHEGxCCo8EHnKGSU3paeWJJaUFEEzlBVCA3LyggR_TJ6zhqZeImZgKisEQAQUoO4CugYYC-mlwMIK6CC9mDxlYDmVqpBRUlJQbKWvD9OeX5Sun5qnn5KfXKyPokNfAZRuEssSM8FFIKpxCvngyC8GGktJLKpDIlEdXxyqg6IQqAJvtAAA

_No response_

kavix:
## Summary

Fixes #4632

### Why
When accessing an attribute on a class instance where the attribute does not exist statically, Pyrefly invokes `__getattr__` / `__getattribute__` fallback resolution. When `__getattr__` has a protocol-annotated `self` parameter (e.g. `def __getattr__(self: Awaitable): ...`), Pyrefly validates receiver compatibility via structural protocol subtyping (`C <: Awaitable`).

During structural protocol subtyping, Pyrefly checks whether `C` provides each member of `Awaitable` (such as `__await__`). Because `C` does not statically define `__await__`, looking up `__await__` on `C` invokes `C.__getattr__` again, triggering another receiver self validation against `Awaitable`.

Because each fallback invocation initiates a new subset query across query boundaries, local cycle detection in `Subset` alone does not observe the outer in-progress protocol-member check. This leads to unbounded recursion and a stack overflow abort.

### What
- Added `protocol_member_guard_stack: RefCell<FxHashSet<(Type, Class, Name)>>` to `ThreadState` in `answers_solver.rs`, following the existing coinductive pattern used by `overload_self_filter_stack`.
- Added `enter_protocol_member_check` and `exit_protocol_member_check` helpers to `AnswersSolver`.
- Wrapped `is_protocol_subset_at_attr` in `attr.rs` with the coinductive guard keyed by `(got_type, protocol_class, attr_name)`.
- Added a regression test `test_getattr_self_awaitable_recursion` in `attributes.rs`.

### Why it works
The guard tracks active `(got_type, protocol_class, member_name)` checks across nested subset queries on the current thread. When the exact same check is re-entered during nested attribute resolution, Pyrefly recognizes the cycle and terminates the check coinductively (`Ok(())`), preventing unbounded recursion while preserving standard protocol and receiver validation semantics on all non-cyclic paths.

## Test Plan
- `cargo test test_getattr_self_awaitable_recursion` passes (reproducer no longer overflows the stack and reports the expected diagnostics).
- `cargo test attributes` (242 tests) passes.
- `cargo test protocol` (172 tests) passes.
- `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema` passes formatting and linting.
- `git diff --check` passes.

### DIFF
--- a/pyrefly/lib/alt/answers_solver.rs
+++ b/pyrefly/lib/alt/answers_solver.rs
@@ -32,6 +32,7 @@ use pyrefly_graph::calculation::ProposalResult;
 use pyrefly_graph::index::Idx;
 use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::module_path::ModulePath;
+use pyrefly_types::class::ClassType;
 use pyrefly_types::heap::TypeHeap;
 use pyrefly_types::quantified::AnchorIndex;
 use pyrefly_types::quantified::Quantified;
@@ -1769,6 +1770,13 @@ pub struct ThreadState {
     /// check is currently in progress, used as a coinductive guard against
     /// self-referential protocols. See `filter_overloads_by_self_type`.
     overload_self_filter_stack: RefCell<FxHashSet<(Type, Type)>>,
+    /// `(got_type, protocol_class_type, attr_name)` tuples whose protocol-member check is
+    /// currently in progress, used as a coinductive guard against recursive protocol checks
+    /// (e.g. `__getattr__` annotated with a protocol).
+    protocol_member_guard_stack: RefCell<FxHashSet<(Type, ClassType, Name)>>,
+    /// Tracks whether any coinductive assumption was used during solving
+    /// (e.g. recursive protocol member resolution via dynamic fallback).
+    coinductive_assumptions_used: Cell<bool>,
 }
 
 impl ThreadState {
@@ -1782,6 +1790,8 @@ impl ThreadState {
             trace_sink: RefCell::new(None),
             trace_sink_stack: RefCell::new(Vec::new()),
             overload_self_filter_stack: RefCell::new(FxHashSet::default()),
+            protocol_member_guard_stack: RefCell::new(FxHashSet::default()),
+            coinductive_assumptions_used: Cell::new(false),
         }
     }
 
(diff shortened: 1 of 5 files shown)


grievejia (changes requested):

grievejia:
Thanks for the patch! I do not think this PR is ready to merge though. 

There are two blocking correctness issues:

1. The recursion key uses `protocol.class_object()`. This makes `P[int]` and `P[str]` use the same key.

2. The guard returns `Ok(())` without setting `coinductive_assumptions_used`. Pyrefly can then store an assumption-dependent result in `protocol_cache`.

This makes the result depend on statement order:

```python
from typing import Protocol, TypeVar

T = TypeVar("T", covariant=True)

class P(Protocol[T]):
    @property
    def x(self) -> T: ...

class C:
    def __getattr__(self: "P[str]", name: str) -> int:
        return 0

def f_int(x: P[int]) -> None: ...
def f_str(x: P[str]) -> None: ...

f_int(C())  # Correctly accepted, but causes an incorrect cached result below.
f_str(C())  # Must produce an error, but is currently accepted.
```

Please fix both the key and the cache bookkeeping. Add this ordered case as a regression test.

The guard also runs once for every member of every protocol check, although this recursion occurs only through `Attribute::GetAttr`. This adds two potentially structural `Type` copies and two hash operations per member. Please scope the guard to the dynamic attribute fallback, or provide benchmark results showing that this cost is acceptable.

Please update the existing recursion test by adding the missing `name: str` parameter. Keep the protocol annotation on `self`, because that annotation triggers the recursion.

This PR also contains an unrelated change that propagates deprecation information through re-export chains. Please move that change and its tests to a separate PR.

kavix:
> Thanks for the patch! I do not think this PR is ready to merge though.
> 
> There are two blocking correctness issues:
> 
> 1. The recursion key uses `protocol.class_object()`. This makes `P[int]` and `P[str]` use the same key.
> 2. The guard returns `Ok(())` without setting `coinductive_assumptions_used`. Pyrefly can then store an assumption-dependent result in `protocol_cache`.
> 
> This makes the result depend on statement order:
> 
> ```python
> from typing import Protocol, TypeVar
> 
> T = TypeVar("T", covariant=True)
> 
> class P(Protocol[T]):
>     @property
>     def x(self) -> T: ...
> 
> class C:
>     def __getattr__(self: "P[str]", name: str) -> int:
>         return 0
> 
> def f_int(x: P[int]) -> None: ...
> def f_str(x: P[str]) -> None: ...
> 
> f_int(C())  # Correctly accepted, but causes an incorrect cached result below.
> f_str(C())  # Must produce an error, but is currently accepted.
> ```
> 
> Please fix both the key and the cache bookkeeping. Add this ordered case as a regression test.
> 
> The guard also runs once for every member of every protocol check, although this recursion occurs only through `Attribute::GetAttr`. This adds two potentially structural `Type` copies and two hash operations per member. Please scope the guard to the dynamic attribute fallback, or provide benchmark results showing that this cost is acceptable.
> 
> Please update the existing recursion test by adding the missing `name: str` parameter. Keep the protocol annotation on `self`, because that annotation triggers the recursion.
> 
> This PR also contains an unrelated change that propagates deprecation information through re-export chains. Please move that change and its tests to a separate PR.


Thanks for the review and feedback! I've updated the PR

1. **Fixed key for generic protocols**: Changed the guard key to include full class type arguments so `P[int]`
  and `P[str]` are treated separately.
2. **Fixed cache issue**: We now track `coinductive_assumptions_used` properly so speculative results are not saved into `protocol_cache`. Added the ordered test case as a regression test.
3. **Optimized guard check**: Scoped the guard only to dynamic attribute fallback (`Attribute::GetAttr`), so
  normal attribute lookups don't do any extra work.
4. **Updated test**: Added the missing `name: str` parameter to `__getattr__` in the test.
5. **Rebased & cleaned up commits**: Removed the unrelated re-export deprecation changes (will open a separate PR
  for those) and resolved merge conflicts.

kavix:
@grievejia can u re review?

stroxler (approved):

PR #4637, opened 2026-08-23
Preserve fixed tuple types for reverse slices

### LINKED ISSUE #4636: Fixed-length tuple loses length and element order after [::-1]
### Describe the Bug

A full reverse slice of a fixed-length tuple is inferred as a variable-length homogeneous tuple. This loses both the known length and, for heterogeneous tuples, the positional element types.

Minimal reproduction:

```python

triple: tuple[int, int, int] = (1, 2, 3)
assert_type(triple[::-1], tuple[int, int, int])

mixed: tuple[int, str, bool] = (1, "two", True)
assert_type(mixed[::-1], tuple[bool, str, int])
```

Running `pyrefly check repro.py` reports:

```text
assert_type(tuple[int, ...], tuple[int, int, int]) failed
assert_type(tuple[bool | int | str, ...], tuple[bool, str, int]) failed
```

Expected: both assertions pass. A statically known full reverse slice preserves tuple length and reverses its positional element types.

Reproduced with:

- Pyrefly 1.2.0
- Pyrefly `main` at `19d136cf5` (1.3.0-dev.2)
- macOS 26.4.1, arm64

This affects geometry code that uses fixed-length coordinate tuples such as `tuple[float, float, float]` and reverses axis order with `[::-1]`; callers otherwise need an unnecessary cast to restore the known arity.

ternaus:
## Summary

- preserve the arity and positional element types of concrete tuples for full `[::-1]` slices
- retain the existing fallback for other stepped slices and tuples with unknown structure
- add a regression test using a heterogeneous fixed-length tuple

Fixes #4636.

## Test plan

- `cargo test test_tuple_reverse_slice`
- `cargo test test::tuple::` (81 passed)
- original CLI reproduction on the patched binary (`0 errors`)

The lint command completes successfully with two pre-existing Clippy warnings in unchanged files.

## AI usage

This PR was implemented with Codex assistance. I reviewed the issue reproduction, implementation, regression test, and final diff, and ran the validation listed above.

### DIFF
--- a/pyrefly/lib/alt/subscript.rs
+++ b/pyrefly/lib/alt/subscript.rs
@@ -86,10 +86,23 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         slice: &ExprSlice,
         slice_ty: Option<&Type>,
     ) -> Option<Type> {
+        let slice_targs = slice_ty.and_then(Self::slice_type_args);
         if slice.step.is_some() {
+            let step = self
+                .parse_slice_literal(&slice.step, slice_targs.map(|[_, _, step]| step))
+                .ok()?;
+            if step == Some(-1)
+                && slice.lower.is_none()
+                && slice.upper.is_none()
+                && let Tuple::Concrete(elts) = tuple
+            {
+                return Some(
+                    self.heap
+                        .mk_concrete_tuple(elts.iter().rev().cloned().collect()),
+                );
+            }
             return None;
         }
-        let slice_targs = slice_ty.and_then(Self::slice_type_args);
         match tuple {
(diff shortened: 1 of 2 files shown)


stroxler:
I'm trying to get auto-import to run on this, I'm going to test close-and-reopen to see if it will trigger (I believe that PRs that don't have a CLA on file when opened may not import, which makes sense for legal reasons but now we have the CLA).

stroxler (approved):
LGTM. I'll try to get a second review (necessary because I couldn't get auto-import to run) and merge it sometime this week.

Thanks!

rchen152 (approved):

PR #4638, opened 2026-08-23
Consider __all__.remove when determining explicit exports

### LINKED ISSUE #4564: `implicit-reexport` does not consider `__all__.remove(...)`.
### Describe the Bug

This is definitely more on the lower priority of issues, but filing for completeness. If `__all__` is resolvable, mutations to it via `remove` are not correctly considered.

```
# foo.py
c: int = 3

# bar.py
from foo import c
__all__ = ["c"]
__all__.remove("c")

# baz.py
from bar import c  # No error here.
```

In this case, `c` should no longer be considered an exported name from bar after `__all__.remove("c")`. However, Pyrefly still allows `from bar import c`.

It looks like in `exports.rs`, `get_explicit_dunder_all_names_iter` is used to determine if a `implicit-reexport` occurred, but this function only checks explicit literals (or `appends`). It does not consider `DunderAllEntry::Remove`.

### Sandbox Link

No sandbox due to multiple files.

_No response_

kavix:
## Summary
Fixes #4564.

When checking for implicit re-exports, Pyrefly previously filtered only `DunderAllEntry::Name` entries from explicit `__all__` definitions, ignoring subsequent `__all__.remove(...)` mutations. As a result, symbols removed from `__all__` were still treated as explicitly exported.

## Changes
- Added `Exports::get_explicit_dunder_all_names` to compute the final set of user-defined `__all__` symbols by applying mutations in source order (inserting on `Name`, removing on `Remove`), consistent with `get_partially_known_dunder_all` and `wildcard`.
- Stored `explicit_dunder_all_names: Option<SmallSet<Name>>` on `Exports` and updated `get_explicit_dunder_all_names_iter`, `is_implicit_reexport`, and `changed_exports`.
- Added regression test `test_implicit_reexport_removed_from_all` in `pyrefly/lib/test/imports.rs`.

## Test Plan
- `CARGO_BUILD_JOBS=1 cargo test -p pyrefly --lib test::imports::test_implicit_reexport_removed_from_all -- --exact --nocapture`
- `CARGO_BUILD_JOBS=1 cargo check -p pyrefly --lib`
- `CARGO_BUILD_JOBS=1 cargo check -p pyrefly --tests`
- Direct CLI smoke test with `[tool.pyrefly.errors] implicit-reexport = "error"` verifying that removing a name triggers the diagnostic and appending it back silences it.

## AI Disclosure
This PR was developed and verified with the assistance of an AI coding assistant (Google Antigravity).

### DIFF
--- a/pyrefly/lib/export/exports.rs
+++ b/pyrefly/lib/export/exports.rs
@@ -112,6 +112,8 @@ pub struct Exports {
     definitions: Definitions,
     /// Names statically known to be in a partially resolvable explicit `__all__`.
     partially_known_dunder_all: SmallSet<Name>,
+    /// Names statically known to be in a user-defined explicit `__all__`.
+    explicit_dunder_all_names: Option<SmallSet<Name>>,
     /// Names that are available via `from <this_module> import *`
     wildcard: Calculation<Arc<SmallSet<Name>>>,
     /// Names that are available via `from <this_module> import <name>` along with their locations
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #4643, opened 2026-08-24
Scope the Require::Everything escalation to the requests that read ASTs, and add the #3636 regression test

DarkNightForge:
Follow-up to #3637 (landed as `3c288a2`), measured against `origin/main` @ `8ec5070d2`.

Two things: the regression test #3636 still lacks, and a narrowing of where the
`Require::Everything` escalation is paid.

## 1. The regression test

`3c288a2` is correct — I verified `test_incoming_call_hierarchy_unopened_caller_file` passes
against unmodified `main`. But it landed as 4 additions / 3 deletions in one file with no test,
so nothing holds the behaviour. `test_incoming_call_hierarchy_basic` opens *both* files, so it
cannot catch #3636.

The new test opens only `callee.py` and runs with `IndexingMode::LazyBlocking`, so `caller.py`
is reached through the index. Verified to fail against the code as it stood before `3c288a2`.

@d34db3ff — this is the test we discussed on #3637; the PR merged before I could add it there.

## 2. Who pays for the escalation

`3c288a2` moved the escalation into `compute_transitive_rdeps_for_definition_impl`, which is
shared by four features. Only two of them ever read an AST:

| consumer | reads an AST? |
| --- | --- |
| `callHierarchy/incomingCalls` | **yes** — `get_ast`, to attribute a call site to its enclosing function |
| `typeHierarchy` subtypes | **yes** — `get_ast` + `get_solutions` + `get_bindings` |
| `textDocument/references` | no — `local_references_from_definition` and `find_child_implementations_impl` are both index-backed |
| `find_global_implementations_from_definition` | no — `solutions_index` only |

So `references`, much the hottest of the four, now runs a full check over every transitive rdep
of the definition's module and retains their ASTs for data it never reads.

This PR moves the escalation to the two consumers that need it:

- **call hierarchy** asks the index which rdeps actually reference the target, then escalates
  only those. A file that never mentions the target costs nothing beyond the lookup. Ordering
  matters: the pre-`3c288a2` code fetched the AST *before* checking for references. The
  escalation is batched into one `run` — doing it per file from inside the walk leaves every
  file after the first still without an AST.
- **type hierarchy** escalates its own deduped candidates, so its behaviour is exactly what
  `main` does today.

`references` and `implementations` stop paying entirely.

## Measurements

This answers the perf question @yangdanny97 raised on #3637 in June.

Release builds, a 209-file Python project (the same one I profiled on #3637, since grown from
189), LSP driven programmatically opening only the file it queries. **A fresh server per cold rep**, median of 3; the timed request is the first one that
server receives. RSS delta is sampled immediately before and after that single request, so it is
what the request itself retained. Result counts in parentheses are ground truth and were
identical on both builds. The three positions are the same ones I profiled on #3637.

| | refs (82) | refs (36) | `incomingCalls` (32) |
| --- | --- | --- | --- |
| `main` @ `8ec5070d2` — cold | 65.4 ms | 31.5 ms | 31.6 ms |
| this PR — cold | **3.4 ms** | **1.9 ms** | **26.8 ms** |
| `main` — RSS retained by the one request | 81 MB | 15 MB | 16 MB |
| this PR — RSS retained | **2 MB** | **1 MB** | **7 MB** |
| `main` — server peak RSS | 396 MB | 329 MB | 316 MB |
| this PR — server peak RSS | **281 MB** | **281 MB** | **291 MB** |

The whole table was run twice on a quiet machine; every figure reproduced within 1%
(e.g. refs (82) cold 65.4 / 65.5 ms on `main`, 3.4 / 3.4 ms here).

A note on methodology, because it changes the answer: the profile I posted on #3637 used a
**median of 5 warm reps**. Escalation to `Require::Everything` is a one-time state transition
that the transaction retains, so reps 2..5 measure the already-escalated steady state and a warm
median structurally hides the cost being argued about. That profile was not evidence that the
escalation is cheap. Hence the fresh-server-per-rep harness here.

## Test suite

`cargo test --lib --no-fail-fast` on this branch: **8,093 passed / 0 failed / 4 ignored** in the
`pyrefly` crate, 8,828 passed and 0 failed across all 13 test binaries. The 4 ignored are
pre-existing. Result counts from the probes above were identical to `main` on every case
(82 / 36 / 32), so this is a cost change, not a behaviour change.

## What this does not do

Nothing tests the *cost* property. If the escalation is re-broadened later, the tests stay green
and only the numbers change. I did not find a good way to assert it at this layer; suggestions
welcome.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/call_hierarchy.rs
+++ b/pyrefly/lib/lsp/non_wasm/call_hierarchy.rs
@@ -436,13 +436,7 @@ impl CancellableTransaction<'_> {
             let mut callers_in_file = Vec::new();
 
             ast.visit(&mut |expr| {
-                collect_calls_from_expr(
-                    expr,
-                    &ref_set,
-                    handle.module(),
-                    &ast,
-                    &mut callers_in_file,
-                )
+                collect_calls_from_expr(expr, &ref_set, handle.module(), &ast, &mut callers_in_file)
             });
 
             if !callers_in_file.is_empty() {
--- a/pyrefly/lib/lsp/non_wasm/call_hierarchy.rs
+++ b/pyrefly/lib/lsp/non_wasm/call_hierarchy.rs
@@ -40,6 +40,7 @@ use crate::lsp::non_wasm::module_helpers::module_info_to_uri;
 use crate::state::lsp::DefinitionMetadata;
 use crate::state::lsp::FindPreference;
 use crate::state::lsp::ReferenceOptions;
+use crate::state::require::Require;
 use crate::state::state::CancellableTransaction;
 
 pub struct CallerInfo {
@@ -353,14 +354,13 @@ impl CancellableTransaction<'_> {
         definition_kind: DefinitionMetadata,
         target_definition: &TextRangeWithModule,
     ) -> Result<Vec<(Module, Vec<CallerInfo>)>, Cancelled> {
-        // Use process_rdeps_with_definition to find references and filter to call sites in a single pass
-        let results = self.process_rdeps_with_definition(
+        // Pass 1: work out which files reference the target at all. This is answered from
+        // the index, which every rdep carries at `Require::Indexing`, so a file that never
+        // mentions the target costs nothing beyond the lookup.
+        let files_with_refs = self.process_rdeps_with_definition(
             sys_info,
             target_definition,
             |transaction, handle, patched_definition| {
-                let module_info = transaction.as_ref().get_module_info(handle)?;
-                let ast = transaction.as_ref().get_ast(handle)?;
-
                 let references = transaction
                     .as_ref()
                     .local_references_from_definition(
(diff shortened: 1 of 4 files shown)


DarkNightForge:
One clarification, since this PR touches `type_hierarchy_candidate_handles` and I have just
filed a type-hierarchy bug: the two are unrelated.

#4644 is a separate, pre-existing defect — `typeHierarchy/subtypes` misses subtypes declared in
another module. I initially suspected it was a #3636-style `Require::Indexing` problem and built
a fixture for it, but all three checks refute that: it reproduces on unmodified `main` (so
`3c288a2`'s blanket escalation does not fix it), it reproduces on this branch, and it reproduces
with both files open (so it is not about opened-ness at all). I dropped that test rather than
ship a claim I could not support.

This PR deliberately keeps type hierarchy's behaviour byte-for-byte identical to `main` — it
only moves *where* the escalation is paid, so it neither fixes nor worsens #4644.

stroxler:
@DarkNightForge Thanks for the PR! I think the CI failures are just asking for an autoformat pass

stroxler (changes requested):
I think this looks pretty good, thanks for the work!

It already needs a rebase before we can merge it though

DarkNightForge:
Rebased onto current `main` (`e7e207365`) — thanks for the review!

One conflict, in `pyrefly/lib/state/lsp.rs`, and it was textual rather than semantic: `b9f38ab9e` ("Fix lint warning about unnecessary `let`") collapsed `let rdeps = …; rdeps` into a bare tail expression, and this change needs that binding back to place the narrowed `run_for_handles` between it and the return. Resolved in favour of this branch; nothing from that lint fix is lost.

Worth flagging what I checked rather than assumed: the *other* hunk here — removing the blanket `run_for_handles(&candidate_handles, Require::Everything)` at the end of `compute_transitive_rdeps_for_definition_impl` — merged cleanly and silently. Had it not applied, the branch would still compile and still pass, while quietly becoming a no-op. Verified the blanket call is gone and the narrowed one is in place.

`server.rs` needed no resolution despite `6ace22772` and `f8b724b90` landing there since; the type-hierarchy escalation is still inside `type_hierarchy_candidate_handles`, after the path dedupe and before `type_hierarchy_subtype_items` consumes the handles.

Targeted suite green on the rebased tip — 7/7 across `lsp_interaction::call_hierarchy` and `lsp_interaction::type_hierarchy`, including `test_incoming_call_hierarchy_unopened_caller_file` (the #3636 regression test this adds) and `test_type_hierarchy_subtypes_in_another_module`, which covers the path #4647 touched.

stroxler (approved):

PR #4646, opened 2026-08-25
Improve LSP path override behavior for windows, homedir paths, and relative paths not starting with `.[.]/`

connernilsen:
The previous logic was there just to get things working in the short term. This diff fixes that behavior to work on windows, handle `~`, and handle relative paths that don't start with a `.`. Simple tests are also added to validate behavior here.


grievejia (approved):

PR #4648, opened 2026-08-25
Update status bar on build system changes

### LINKED ISSUE #4229: LSP progress missing/confusing in VS Code while a custom build-system query is running
### Describe the Bug

Pyrefly's LSP can invoke a custom `[build-system]` query when a Python file is opened in VS Code. When that query is slow, the VS Code editor does not indicate that Pyrefly is still waiting for build-system data. During that interval, Pyrefly may also publish provisional diagnostics based on incomplete import information, such as `missing-import` errors that disappear after the query finishes.

This is especially confusing in build-system-backed monorepos, where a query may need to analyze a target graph, build source-database outputs, and materialize generated sources or stubs before it can return an accurate module map.

## Real-world motivation

I described our Bazel-backed custom query driver and measured editor latency in [this comment on the Bazel support issue](https://github.com/facebook/pyrefly/issues/1232#issuecomment-4974293916):

- A cold Bazel output user root can take several minutes.
- A warm output root querying a previously unopened file or target takes approximately 13–14 seconds.
- A repeated query for an already processed file or target takes approximately 4–5 seconds.

While the query is running, users cannot tell whether Pyrefly is working, stuck, or finished with an incomplete view of the project unless they open the Pyrefly VS Code output panel and monitor the log messages there. Meanwhile, transient import-related diagnostics look authoritative even though Pyrefly is still loading the source database.

## Minimal reproduction

This uses a custom query that deliberately waits before returning a module mapping that cannot be inferred from the filesystem layout.

Create `pyrefly.toml`:

```toml
preset = "basic"

[build-system]
type = "custom"
command = ["python3", "query_driver.py"]
```

Create `app.py`:

```python
from generated_api import VALUE

reveal_type(VALUE)
```

Create `generated/impl.py`:

```python
VALUE: int = 42
```

Create `query_driver.py`:

```python
import json
import time
from pathlib import Path

time.sleep(20)

print(
    json.dumps(
        {
            "root": str(Path.cwd()),
            "db": {
                "//:app": {
                    "srcs": {
                        "app": ["app.py"],
                        "generated_api": ["generated/impl.py"],
                    },
                    "deps": [],
                    "python_version": "3.12",
                    "python_platform": "linux",
                    "buildfile_path": "BUILD.bazel",
                }
            },
        }
    )
)
```

Open `app.py` using the Pyrefly VS Code extension.

## Current behavior

1. There is no visible indication that Pyrefly is waiting for the custom build-system query.
2. Pyrefly may temporarily report that `generated_api` is missing because the source-database result has not arrived yet.
3. After approximately 20 seconds, the source database is loaded and the provisional diagnostic disappears.
4. Existing LSP work-done progress may appear for the subsequent recheck, but it does not cover the long-running build-system query itself.

## Requested behavior

Could Pyrefly expose the source-database query through standard LSP work-done progress?

At minimum, it would be helpful for Pyrefly to:

1. Send `window/workDoneProgress/create` and a `$/progress` `begin` notification before invoking the custom build-system command.
2. Use an indeterminate message such as `Querying build system` while the command is running.
3. Send the corresponding `end` notification after the result has been parsed and installed.
4. Avoid publishing source-database-dependent diagnostics for a newly opened file until its initial build-system query completes, or otherwise identify those diagnostics as provisional.

An indeterminate progress indicator would already solve the primary discoverability problem. A percentage or ETA is not necessary because many build systems cannot report a reliable total amount of work.

If Pyrefly eventually wants richer progress, the custom query protocol could optionally gain a structured side channel for phases such as analysis, building, and materializing generated files. That does not seem necessary for an initial implementation.

## Related issues

- [#422](https://github.com/facebook/pyrefly/issues/422) added LSP `$/progress` support for recheck transactions. This request is specifically about extending progress coverage to the source-database query that happens before that recheck.
- [#1232](https://github.com/facebook/pyrefly/issues/1232) tracks Bazel support and contains discussion of using custom build-system drivers for editor integration.

### Sandbox Link

Not applicable. The reproduction depends on LSP behavior and a custom build-system subprocess, which the online sandbox does not exercise.

- IDE: VS Code using the official Pyrefly extension
- Pyrefly: 1.1.1
- Platform: Linux
- Workspace: Large Bazel monorepo with a custom `[build-system]` query driver

connernilsen:
# Summary

Fixes #4229

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -741,6 +741,18 @@ impl Default for ConfigFile {
     }
 }
 
+/// The result of requerying every live source database for a set of configs.
+/// See [`ConfigFile::query_source_db`].
+pub struct SourceDbQueryOutcome {
+    /// The source databases whose contents changed, and whose dependent caches
+    /// therefore need invalidating.
+    pub reloaded: SmallSet<ArcId<Box<dyn SourceDatabase + 'static>>>,
+    pub stats: TelemetrySourceDbRebuildStats,
+    /// The first query failure, rendered for display. `None` when every queried
+    /// source database succeeded.
+    pub error: Option<String>,
+}
+
 impl ConfigFile {
     /// Gets a ConfigFile for a project directory. `fallback` indicates whether this is a guessed
     /// project root that we're falling back to after failing to otherwise find an import.
@@ -1249,12 +1261,10 @@ impl ConfigFile {
         configs_to_files: &SmallMap<ArcId<ConfigFile>, SmallSet<ModulePath>>,
         force: bool,
         telemetry: Option<SubTaskTelemetry>,
-    ) -> (
-        SmallSet<ArcId<Box<dyn SourceDatabase + 'static>>>,
-        TelemetrySourceDbRebuildStats,
-    ) {
+    ) -> SourceDbQueryOutcome {
         let mut stats: TelemetrySourceDbRebuildStats = Default::default();
         stats.common.forced = force;
+        let mut first_error = None;
         let mut reloaded_source_dbs = SmallSet::new();
         let mut sourcedb_configs: SmallMap<_, Vec<_>> = SmallMap::new();
         for (config, files) in configs_to_files {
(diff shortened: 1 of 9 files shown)


connernilsen:
Rewrote as separate diff stack

PR #4649, opened 2026-08-25
Discover binary from environment, and update how it can be configured

connernilsen:
Part two addressing #4306

This should make it so we can get the Pyrefly binary from the environment (and do that by default), as well as update how the selections are made.

Not sure if this is going to be a popular decision, but I made `lspPath` only used when set in the `pyreflyExecutable` setting since that makes testing easier, but that behavior may differ from expectations.


stroxler (changes requested):

grievejia (approved):

PR #4654, opened 2026-08-25
Allow language server features for non-Jupyter notebooks

### LINKED ISSUE #4642: Notebooks that aren't Jupyter files on disk get no diagnostics
Hi! This is very related to https://github.com/facebook/pyrefly/issues/1835#issuecomment-3675116331 but I decided to make a new issue. 

We're working on the Quarto experience in Positron, which models the code chunks of an open `.qmd` file as a hidden notebook and syncs the cells over the notebook channel.

I noticed that Pyrefly's `initialize` response accepts any notebook with a Python cell, under any URI scheme, but then its notebook support assumes that each notebook is a Jupyter file on disk. A notebook that isn't one fails in three ways:

1. The extension client's `notebookSelector` names `jupyter-notebook` and nothing else, so any other notebook type is skipped on the notebook channel.
2. Even if it wasn't skipped, it's rejected at open, because a notebook URI that doesn't convert to a filesystem path is dropped with an error (like `Could not convert uri to filepath: quarto-cells:Untitled-1.qmd, expected a notebook`). An unsaved text document gets a synthesized path in this case, but a notebook doesn't.
3. Even if it wasn't rejected, it's not checked, because a notebook whose path isn't a `.py` file matches no `project_includes` glob.

Some cases where this might fail:

- code chunks projected out of a document, such as Quarto (`.qmd`) or R Markdown (`.Rmd`)
- a notebook addressed under a custom scheme, whether or not the path behind it exists
- an extension that registers its own notebook serializer for a file type that Pyrefly's includes don't list

As mentioned in #1835, Marimo notebooks work today, but only because their paths end in `.py`, which the default `project_includes` already match. Nothing else about the Marimo case is special, so the support is a bit narrower than it looks.

I do have a fix with tests, in three parts:

1. The client's `notebookSelector` matches any notebook type, so the cell language stays the only filter.
2. A notebook URI that doesn't convert to a path gets a synthesized path, the same way an unsaved text document does.
3. The `project_includes` check matches a synced notebook as if it were an `.ipynb` file at the same path, so the globs narrow by location but not by the notebook's extension.

The fix isn't specific to Positron or Quarto, but would allow for future strategies like this in other IDEs or extensions too. I can open the PR soon.

austin3dickey:
# Summary

Fixes #4642. This PR allows notebooks that aren't `.ipynb` files on disk to get language server features too. The three main changes are:

1. The client's `notebookSelector` matches any notebook type, so the cell language stays the only filter.
2. A notebook URI that doesn't convert to a path gets a synthesized path, the same way an unsaved text document does.
3. The `project_includes` check matches a synced notebook as if it were an `.ipynb` file at the same path, so the globs narrow by location but not extension.

`project_excludes` still apply, and a notebook outside a narrowed `project_includes` gets no diagnostics, the same as an open text document. Text documents keep the narrower list of schemes, so a document under a scheme that Pyrefly shouldn't analyze is still turned away.

# Test Plan

The new tests are in the notebook sync and notebook token suites. Each of these fails without the change:

- a notebook with a `.qmd` (Quarto markdown) path publishes its cell diagnostics
- the same notebook under a custom scheme
- a notebook projected from an unsaved document, which has no path at all
- a name from the first cell resolves in the second cell
- a module beside the source document resolves
- a cell edit republishes, and a close clears
- semantic tokens come back for the cells of a custom-scheme notebook

These pin the limits of the change:

- a notebook under an excluded path publishes nothing
- a notebook outside a narrowed `project_includes` publishes nothing, and one inside it still publishes
- a Marimo-style notebook, a `.py` file on disk, keeps working
- a text document under an unhandled scheme is still turned away
- every diagnostic arrives on a cell URI
- a custom-scheme notebook with an `.ipynb` path keeps working

I also tested this by hand against the first case it's meant to serve: in Positron, the code chunks of an open `.qmd` are modeled as a hidden notebook under the `quarto-cells` scheme. With this branch, Pyrefly serves those chunks over the notebook channel, and I read the LSP traffic directly to confirm it's working.

### DIFF
--- a/crates/pyrefly_lsp_test/src/object_model.rs
+++ b/crates/pyrefly_lsp_test/src/object_model.rs
@@ -1535,7 +1535,17 @@ impl LspInteraction {
         let root = self.client.get_root_or_panic();
         let notebook_path = root.join(file_name);
         let notebook_uri = Url::from_file_path(&notebook_path).unwrap().to_string();
+        self.open_notebook_with_uri(&notebook_uri, "jupyter-notebook", file_name, cells_spec);
+    }
 
+    /// Opens a notebook document under an explicit notebook URI and notebook type.
+    pub fn open_notebook_with_uri(
+        &self,
+        notebook_uri: &str,
+        notebook_type: &str,
+        file_name: &str,
+        cells_spec: Vec<(CellKind, &str)>,
+    ) {
         let mut cells = Vec::new();
         let mut cell_text_documents = Vec::new();
         for (i, (kind, text)) in cells_spec.iter().enumerate() {
@@ -1548,7 +1558,7 @@ impl LspInteraction {
             .send_notification::<DidOpenNotebookDocument>(json!({
                 "notebookDocument": {
                     "uri": notebook_uri,
-                    "notebookType": "jupyter-notebook",
+                    "notebookType": notebook_type,
                     "version": 1,
                     "metadata": {
                         "language_info": {
--- a/lsp/src/extension.ts
+++ b/lsp/src/extension.ts
@@ -184,12 +184,12 @@ export async function activate(context: ExtensionContext) {
       // Support for in-memory documents like the Positron Console
       {scheme: 'inmemory', language: 'python'},
     ],
-    // Support for notebooks
+    // Support for any notebook type
     // @ts-ignore
     notebookDocumentSync: {
       notebookSelector: [
         {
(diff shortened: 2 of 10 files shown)


rchen152 (approved):
This looks good to me, thank you! (I had to add a license header to the new test file, so this is now waiting for a second internal review before I can merge.)

ndmitchell (approved):

stroxler (approved):

PR #4664, opened 2026-08-26
[ez] Exclude 'epic' label from stale bot

yangdanny97:
These are tracking tasks and can be assigned to someone long term


stroxler (approved):

PR #4399, opened 2026-08-01
fix __call__ don't have parameter name #4377

asukaminato0721:
# Summary

Fixes #4377

Callable instances are now coerced to their bound `__call__` signature before generating parameter-name hints

# Test Plan

add test

PR #4667, opened 2026-08-26
Add regression tests for instance-method inlay hint alignment

### LINKED ISSUE #4611: Instance-method inlay hint shifted by one parameter (self not skipped); plain functions are fine
## Environment
- pyrefly: 1.3.0.dev2
- Editor: PyCharm 2026.2.1 #PY-262.9437.214 (pyrefly as LSP), Inlay Hints enabled
- Built-in (pyright) shows hints correctly — so this is a pyrefly LSP bug.

## Bug
For **instance methods**, the parameter-name inlay hint is **shifted by one
position** (the first real parameter `a` is dropped, labels misaligned). This
happens **regardless of whether the variable is explicitly annotated**. Plain
(non-method) functions are correct, so the issue is specific to the `self`
parameter not being skipped in the hint computation.

Image

## Minimal reproduction
```python
class _Mini:
    def method(self, a: int, b: str, c: float):
        return None

m = _Mini()
m.method(1, "x", 2.0)
# Expected: a= b= c=   Actual: b= c=   (a= missing -> shifted by 1)

m2: _Mini = _Mini()          # explicit annotation, still wrong
m2.method(1, "x", 2.0)
# Expected: a= b= c=   Actual: b= c=   (annotation does NOT fix it)

def plain_func(x: int, y: str, z: float): ...
plain_func(1, "x", 2.0)
# Correct: x= y= z=   (no self -> no shift)
```

## Expected behavior
Instance-method hints should skip `self` and show `a= b= c=`, matching
pyright / Built-in.

Image

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZRBCsIwEEWvMthFXZQEu8wdBA8gSNVaBWuLiVI9vX8yURq13ZTswsxnknn_D0WmWGWzhGrPKDHaBgtAEm2965BLhg5IJPeHIUGPT41XeFVeZa2C4CLD07r1LKNcCQ91bqQuqs-HG8Q77ZmhYQfNuzDeI4z3_IxHSilu6BV_6_0uN6GVXFD4J8QaqKhuCE9sG4IJcrOko3OtNVq_25trpcuL3jc7q6MOTcxxcS9OPpJjOWo8jBayU6hKBap0jKmUkULFKCYv

_No response_

nitishagar:
## Summary

Issue #4611 reports parameter-name inlay hints shifted by one position for instance-method calls (e.g. `m.method(1, "x", 2.0)` over `def method(self, a: int, b: str, c: float)` showing `b=`/`c=` instead of `a=`/`b=`/`c=`) on pyrefly `1.3.0-dev.2`.

We verified that the exact repro produces **correct** hints (`a= b= c=`) on current `main` through both harnesses:

- the core inlay-hint computation (`Transaction::inlay_hints`), with inferred and annotated receivers, and additionally for `self.method(...)` calls inside other methods, classmethods (`cls`), bound methods stored in variables, and plain functions as a control;
- the full LSP path (`textDocument/inlayHint`, the layer editors like PyCharm exercise).

Since main does not reproduce the shift (the regression window is between Aug 4 and Aug 18, tag `1.3.0-dev.2`; suspects are #4399/#4539-era changes or a stale build on the reporter side), this PR locks the correct behavior in with permanent regression tests at both levels rather than speculative refactors.

Fixes #4611 — we'd ask the reporter to retest against a current build; if the shift still reproduces there, a fresh trace from that build would pinpoint the delta.

AI-generated: this PR was prepared by an AI agent (with human review before submission), per the repository's AI contribution guidelines.

### DIFF
--- a/pyrefly/lib/test/lsp/inlay_hint.rs
+++ b/pyrefly/lib/test/lsp/inlay_hint.rs
@@ -504,6 +504,88 @@ fn parameter_name_hint_labels(code: &str, assert_zero_errors: bool) -> Vec<Strin
         .collect()
 }
 
+// Regression coverage for #4611: instance-method parameter-name hints must
+// align one-to-one with positional arguments (`self` never consumes a slot).
+#[test]
+fn test_issue_4611_instance_method_hint_alignment() {
+    // Inferred receiver.
+    let code = r#"
+class _Mini:
+    def method(self, a: int, b: str, c: float) -> None:
+        pass
+
+m = _Mini()
+m.method(1, "x", 2.0)
+"#;
+    assert_eq!(
+        parameter_name_hint_labels(code, true),
+        vec!["a= ", "b= ", "c= "]
+    );
+
+    // Annotated receiver.
+    let code = r#"
+class _Mini:
+    def method(self, a: int, b: str, c: float) -> None:
+        pass
+
+m2: _Mini = _Mini()
+m2.method(1, "x", 2.0)
+"#;
+    assert_eq!(
+        parameter_name_hint_labels(code, true),
+        vec!["a= ", "b= ", "c= "]
+    );
+
+    // Plain-function control (unchanged behavior).
+    let code = r#"
+def _func(a: int, b: str, c: float) -> None:
+    pass
+
+_func(1, "x", 2.0)
+"#;
+    assert_eq!(
+        parameter_name_hint_labels(code, true),
+        vec!["a= ", "b= ", "c= "]
+    );
+}
+
+#[test]
+fn test_issue_4611_hint_alignment_other_receivers() {
+    // self.method(...) inside another method.
+    let code = r#"
+class _Outer:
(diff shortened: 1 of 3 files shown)


NathanTempest (changes requested):
Thorough tests and nice to see both the core inlay_hints path and the full textDocument/inlayHint path (the one PyCharm hits), plus the classmethod/bound-method cases. Thanks for this @nitishagar 

One ask before landing: Fixes #4611 will auto-close the issue, but this adds tests confirming current behavior rather than fixing a reproduced bug and the summary itself asks the reporter to retest. Since it was only seen in PyCharm on 1.2.0 and nobody's reproduced it on a current build, please change Fixes → Refs #4611 and leave it open until the author confirms. LGTM once that's adjusted.

Nit: fixture uses class Mini but the unit tests use _Mini — worth aligning.

stroxler (approved):

PR #4668, opened 2026-08-26
Publish diagnostics for open extension-less scripts

### LINKED ISSUE #4397: lsp diagnostics does not work for scripts without an extension
### Describe the Bug

Pyrefly language server does not provide any diagnostics for python scripts using shebang instead of .py extension.

For comparison, some other language servers, such as pyright and ty, were able to provide diagnostics for such files.

### Example
In vscode when `python.pyrefly.typeCheckingMode` is set to `default`, the language server will provide diagnostics for a python file named `script.py` but not for a python file named `script`.

Image

### Sandbox Link

_No response_

_No response_

nitishagar:
## Summary

Extension-less Python scripts (e.g. a shebang script named `script` rather than `script.py`) received no LSP diagnostics: the display-time filter in `get_diag_if_shown` required open files to be covered by `project_includes`, whose default globs (`**/*.py*`, `**/*.ipynb`) can never match an extension-less path — so the checker computed the errors but they were dropped before publishing. Pyright and ty both provide diagnostics for such files.

The fix accepts an open file with no file extension even when the includes glob misses it (`project_includes.covers(path) || path.extension().is_none()`). Files with extensions keep exactly the previous gating — including customized `project_includes` scope — and `project_excludes` still suppress extension-less files. Both push (`publishDiagnostics`) and pull (`textDocument/diagnostic`) paths are fixed since both funnel through `get_diag_if_shown`.

Fixes #4397

## Test Plan

- New LSP interaction test `test_diagnostics_for_extensionless_script`: opens a shebang script without extension containing a type error and asserts the diagnostic arrives (push and pull).
- New test `test_diagnostics_extensionless_in_excludes_still_suppressed`: an extension-less script matched by `project-excludes` stays silent.
- Existing gating tests (`test_diagnostics_file_not_in_includes`, `test_diagnostics_file_in_excludes`) pass unchanged, proving custom project scoping is untouched.

AI-generated: this PR was prepared by an AI agent (with human review before submission), per the repository's AI contribution guidelines.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -2926,8 +2926,13 @@ impl Server {
             // the `typeCheckingMode` IDE setting reaches us through the
             // resolver at config synthesis time, not per-diagnostic.
 
+            // A file the editor explicitly opened as Python is in scope for
+            // diagnostics even though its name lacks a python extension —
+            // default includes are an extension heuristic, not a scope
+            // decision; only excludes scope open files down (pyright/ty
+            // parity, #4397).
             if let Some(lsp_file) = open_files.get(&path)
-                && config.project_includes.covers(&path)
+                && (config.project_includes.covers(&path) || path.extension().is_none())
                 && !config.project_excludes.covers(&path)
                 && type_error_status.is_enabled()
             {
--- a/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
+++ b/pyrefly/lib/test/lsp/lsp_interaction/diagnostic.rs
@@ -6,6 +6,7 @@
  */
 
 use lsp_server::RequestId;
+use lsp_types::DocumentDiagnosticReport;
 use lsp_types::DocumentDiagnosticReportResult;
 use lsp_types::PublishDiagnosticsParams;
 use lsp_types::Url;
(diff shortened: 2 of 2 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -2926,8 +2926,13 @@ impl Server {
             // the `typeCheckingMode` IDE setting reaches us through the
             // resolver at config synthesis time, not per-diagnostic.
 
+            // A file the editor explicitly opened as Python is in scope for
```
rchen152, line 2929: "the editor explicitly opened as Python" isn't ever actually checked. As far as I can tell, this change will cause python diagnostics to start firing on every extension-less file.

rchen152 (changes requested):
Thanks for the PR! I think the change to `get_diag_if_shown` is in the right place, but to prevent it from over-firing, we need to also save and consult the `language_id` from `didOpen`. I believe this is where you'd want to grab `language_id`: https://github.com/facebook/pyrefly/blob/2963ffff79ee987c6423f08de6c3f1fa2cf7eb84/pyrefly/lib/lsp/non_wasm/server.rs#L2065, and you'd want to save it like how we do other properties of open documents here: https://github.com/facebook/pyrefly/blob/2963ffff79ee987c6423f08de6c3f1fa2cf7eb84/pyrefly/lib/lsp/non_wasm/server.rs#L3974-L3975.

rchen152:
Please request another review from me when this PR is ready for re-review.

nitishagar:
Revised per your feedback: the extension-less bypass is now gated on the `language_id` recorded at didOpen time (`open_file_languages` map, cleared on didClose) rather than presence in the open-files list alone. Added `test_diagnostics_extensionless_non_python_suppressed` (plaintext-opened script gets no diagnostics); it failed pre-fix and passes now, alongside the existing python-opened and excludes tests. Full diagnostic LSP suite: 37 pass, 2 fail on the pre-existing `missing_source` pair (verified failing on the clean tree too).

PR #4670, opened 2026-08-26
Don't hand out an already-cancelled transaction from saved state

alexander-beedie:
# Summary

### Overview

The `test_shows_stdlib_errors_for_multiple_versions_and_paths_with_force_on` test showed some flakiness, where a `textDocument/diagnostic` response comes back `"items": []` for a just-opened file. This pointed to a real gremlin.

### Solution

Resetting the cancellation on restore upholds the invariant that every transaction handed                                                                                                                                                                                                                                                                                                                                                                                              out can perform work, while keeping the cached work that makes saving worthwhile (this is a general form of the guard found in the `ProvideType` handler).

# Test Plan

Existing tests pass; created a new/deterministic unit test (with the help of Opus 5) that covers the issue.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/transaction_manager.rs
+++ b/pyrefly/lib/lsp/non_wasm/transaction_manager.rs
@@ -61,9 +61,9 @@ impl<'a> TransactionManager<'a> {
         let previous_blocking = match self.saved_state.take() {
             Some(saved_state) => match saved_state.restore() {
                 Ok(mut tx) => {
-                    // The saved cancellation belonged to the previous consumer and has already
-                    // taken effect; clearing it upholds the invariant that a transaction
-                    // handed out here can perform work, instead of silently doing none.
+                    // The saved cancellation belonged to the previous consumer and has
+                    // taken effect; clearing upholds the invariant that a transaction
+                    // handed out here can perform work (instead of silently doing none).
                     tx.reset_cancellation();
                     return tx;
                 }
@@ -105,6 +105,8 @@ mod tests {
 
     /// A recheck cancels the in-flight reads that block its commit. That cancellation belongs
     /// to the request being aborted, so it should *not* survive into the saved transaction.
+    /// Clearing it must also keep the work the saved transaction already did, so the restored
+    /// transaction reuses the cached stdlib instead of recomputing it.
     #[test]
     fn test_restored_transaction_is_not_still_cancelled() {
         let mut test_env = TestEnv::new();
@@ -145,8 +147,16 @@ mod tests {
 
         let second = handle("second");
         let mut transaction = manager.non_committable_transaction(&state);
+        assert!(
+            !transaction.get_cancellation_handle().is_cancelled(),
+            "restored transaction should hold a fresh cancellation handle"
+        );
         transaction.set_memory(test_env.get_memory());
         transaction.run(&[second.dupe()], Require::Everything, None);
+        assert!(
+            transaction.compute_stdlib_cached(),
+            "restored transaction should reuse the stdlib computed before the save"
+        );
         assert_eq!(
             transaction
                 .get_errors([&second])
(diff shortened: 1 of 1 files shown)


NathanTempest (changes requested):
Nice find, @alexander-beedie. A successful saved_state.restore() means the saved transaction is still based on the current state epoch, while the previous transaction consumer has already been consumed by save(). Any cancellation-handle clones still refer to that previous request, so replacing the handle correctly prevents cancellation from leaking into the restored consumer. I also confirmed this is the only TransactionData::restore() call site.

One clarification: this does not make the ProvideType guard redundant. That handler operates on a live request after its cancellation handle has been registered. It removes the registered old handle and resets the transaction to detach provide_type from concurrent recheck cancellation. Since reset installs a new handle rather than clearing the old shared flag, a later cancellation of the registered old handle cannot affect the replacement.

The regression test proves the important user-visible behavior, but it would also pass if cancelled saved state were discarded and a fresh transaction created. Could we additionally assert immediately after restoration that the new cancellation handle is not cancelled, and after the second run that compute_stdlib_cached() is true? That would pin both parts of the intended fix: fresh request-local cancellation and preservation of cached work.

Was the originally flaky test_shows_stdlib_errors_for_multiple_versions_and_paths_with_force_on also stress-run after this change? One passing run would not be conclusive, but repeated success would provide useful confirmation alongside the deterministic unit test.

alexander-beedie:
> this does not make the `ProvideType` guard redundant

Indeed; just commenting that there is some precedent.

> Could we additionally assert immediately after restoration that the new cancellation handle is not cancelled, and after the second run that `compute_stdlib_cached()` is true?

Good idea; done! 👍 

> Was the originally flaky `test_shows_stdlib_errors_for_multiple_versions_and_paths_with_force_on` also stress-run after this change?

Yes; I absolutely hammered on it 🤣

NathanTempest (approved):
Thanks for extending the test. The direct cancellation assertion pins the request-local invariant, compute_stdlib_cached() distinguishes restoration from discarding the saved transaction and proves cached work survives, and the error assertion still covers the user-visible failure. I also reproduced the original 0 versus 1 failure locally and confirmed resetting the cancellation fixes it.

One open item, and it's optional now: the unit test proves the mechanism, but test_shows_stdlib_errors_for_multiple_versions_and_paths_with_force_on is a separate LSP interaction test, did you stress-run it after the change? Repeated passes would confirm this was the cause rather than a real bug found alongside it. No need to make any changes, you can just let me know if the stress test fixes this now.

kinto0 (changes requested):

alexander-beedie:
> did you stress-run it after the change?

Yes; I put together a simple stress test locally that could  reproduce, and then validate, the PR. Was able trigger the error condition reliably before the fix, but not after 👍

rchen152 (changes requested):

NathanTempest:
Hello @alexander-beedie sorry for the delay in review. Relaying the comments from the internal review:

Another maintainer found the flake in the wild and it matches your diagnosis. The test failed with an empty items array in https://github.com/facebook/pyrefly/actions/runs/32533556343/job/96930163340 and passed on re-run. 

- Should the framework clear the cancellation rather than the caller? Instead of calling reset_cancellation() at the call site, TransactionData::restore() could return an uncancelled transaction directly. There's currently exactly one caller of restore(), so that's behaviourally identical today while making the invariant structural rather than something a future caller has to remember. That seems like the preferred shape, and it would also address the concern that the current fix looks like it's patching a symptom at one site.

- Would making indexing synchronous in the test fix the flake instead? The relevant variant is IndexingMode::LazyBlocking, which populates project files inline instead of queueing a background recheck, so the window never opens. That would make the test deterministic, but LazyNonBlockingBackground is the default real users run, so it would stop the test exercising the race without fixing it for them. Worth stating explicitly whether you see these as complementary rather than alternatives.

Address these two and I think this is close, tagging @kinto0 @rchen152 for visibility.

alexander-beedie:
> Address these two and I think this is close, tagging @kinto0 @rchen152 for visibility.

Done! 👍

rchen152 (approved):
LGTM, just needs to be re-imported @NathanTempest

PR #4680, opened 2026-08-26
fix __get__ overload with self-type Concatenate[ObjT, P1] + sibling TypeVar R1 still unmatched after #4415 #4592

### LINKED ISSUE #4592: `__get__` overload with self-type `Concatenate[ObjT, P1]` + sibling TypeVar `R1` still unmatched after #4415
### Describe the Bug

Neither `P1` nor `R1` get solved, so the second overload is rejected and resolution falls through with `no-matching-overload`.

This looks related to but distinct from #4415 (which fixed decomposing `Concatenate[X, **P]` for a single-ParamSpec class like `class A[**P]`). Here the class has *two* type parameters, `JitDeco[**P, R]`, and the overload needs to simultaneously decompose the `Concatenate` in the first slot (`ObjT`/`P1`) *and* solve the plain sibling type var in the second slot (`R1`) from the same `self`-type constraint. That combination still fails even on a build after #4415 landed.

This is the standard idiom for typing a decorator's `__get__` so it supports both free-function and bound-method call shapes (`Self` for class access, `Callable[P1, R1]` for instance access) — e.g. for something like `jax.jit`.

Is this the same root cause as #4415 not fully covered, or a separate bug in resolving a sibling non-ParamSpec type argument alongside a `Concatenate`-decomposed one?

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.lZU9b4MwEIb_ykkeAghhkdFThqpDl6KoG2Ug1KWkEq6ANvn5ff0FRkioHSKcw3fch5_X_5DMQC1V39ST7PHbKp0XJj0pj9xTNz3IRpVJUqR0riLv44-bhbDBB2YIeUoJoADCRaaf-Jt8onprsStzvM6CsizzAJ6Wzy-BW7nADWiEOZJY3uQgdJFYcxNLV-DC7Qcrny_XF51SkaOgvIrW4iDmioNWOZ8ir4xLuriYnMzbwHbDBWSzW6Mf7OFubXL34yldSnNbrKpcu2lu_3svgu3a5Bvg0zbGIAJjLJzno1JucifEXbpzqQfX5rsAnVby8Nyqyp0SOlrrax_cYBFCR3GmAx3jmIhp9y1qjAprIHdqccmA0fYbVxnYQ9YMw5T0MU1fo-Dcu6uh5bLnb6oZ-cqDk1aV-qfuTFfW4UgZaRgRdofxP1B-sJAf9hg_aMSxYxfbXw

asukaminato0721:
# Summary

Fixes #4592

Fixes ParamSpec inference for generic classes when the expected type contains an unsolved `Concatenate[...]` pattern.

Because ParamSpec subtyping already follows callable-parameter direction, applying contravariant class variance again prevents the concrete parameter list from being decomposed during inference.

This allows descriptor `__get__` overloads using `Concatenate[ObjT, P]` and a sibling return TypeVar to resolve correctly, while preserving existing callable-residual behavior.

# Test Plan

add test.


stroxler (approved):

PR #4681, opened 2026-08-26
fix pyrefly takes more than 1000 seconds to type check a simple script... #4678

### LINKED ISSUE #4678: pyrefly takes more than 1000 seconds to type check a simple script...
### Describe the Bug

Hi, we recently met a "bug" (not sure) that pyrefly is taking more than 1000 second when we run type check.

Minimal `pyproject.toml`:

```toml
[project]
name = "compound-split-type-check"
version = "0.1.0"
requires-python = ">=3.13"
dependencies = [
    "compound-split==1.0.2",
    "pyrefly==1.1.1",
]

[tool.pyrefly]
# infer-with-first-use = false
output-format = "min-text"
```

minimal script to reproduce:

```python
from compound_split import char_split

def _select_decomposition(word: str, lang: str) -> None:
    word_split = char_split.split_compound(word, lang=lang)[0]

    print(word_split)
```

Steps to reproduce:

1. create a project with the above config and script
2. use uv to run the type check: `uv run pyrefly check`

It take a lot time for the type check to finish :( On my M5 macBook, it took 1032 seconds

Image

Note that if i uncomment the option `infer-with-first-use`, the type checking time reduces dramatically, but we would like to avoid that.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4678

The string-range collector now uses direct line lookup instead of repeatedly scanning huge non-ASCII lines to calculate unused columns, eliminating quadratic behavior.

with help from gpt 5.6 sol high

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/errors.rs
+++ b/pyrefly/lib/state/errors.rs
@@ -79,9 +79,11 @@ fn collect_string_ranges(expr: &Expr, module: &Module, ranges: &mut Vec<(LineNum
         _ => None,
     };
     if let Some(range) = text_range {
-        let display = module.display_range(range);
-        let start = display.start.line_within_file();
-        let end = display.end.line_within_file();
+        // Computing columns scans from the start of the line for non-ASCII source, but this pass
+        // only needs line numbers.
+        let line_index = module.lined_buffer().line_index();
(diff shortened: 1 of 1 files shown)


rchen152 (approved):

PR #4687, opened 2026-08-27
fix type() of Protocol value incorrectly triggers invalid-argument when used as an isinstance argument #4686

### LINKED ISSUE #4686: `type()` of `Protocol` value incorrectly triggers `invalid-argument` when used as an `isinstance` argument
### Describe the Bug

Pyrefly treats `type(value)` as the protocol class when `value` is statically typed as a protocol. It then
rejects the result as the second argument to `isinstance` unless the protocol is `@runtime_checkable`. At runtime,
`type(value)` is the concrete implementation class, so the protocol itself is never used for the instance check.

```
from typing import Protocol

class Worker(Protocol):
    def work(self) -> None: ...

class ConcreteWorker:
    def work(self) -> None:
        pass

def same_concrete_type(existing: Worker, candidate: Worker) -> bool:
    candidate_type = type(candidate)
    return isinstance(existing, candidate_type)

assert same_concrete_type(ConcreteWorker(), ConcreteWorker())
```

Pyrefly reports:

```
ERROR Protocol `Worker` is not decorated with @runtime_checkable and cannot be used with isinstance() [invalid-argument]
```

Mypy with `--strict` succeeds, and the example runs successfully with Python. [Sandbox](https://mypy-play.net/?mypy=latest&python=3.12&flags=strict%2Cextra-checks&gist=26f177a129f39f70562b4cb175c8f507).

Expected behavior:

Pyrefly should allow `type(candidate)` as the second argument to `isinstance`. The expression evaluates to the concrete runtime class and does not perform a runtime protocol check, making `@runtime_checkable` unnecessary.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZXNDoMgEIRfhcSDepG7175Abz1bJdY0kUbsz-N32AWBtiFpvC7s8PfN8Edk-kgLTjrp5aqWyg94eOzzPzFEFggIiKZpQu9Bzz0IU6yR7wzI8ZuxT4ylpXcyFH-VesHu2HLrtmbDch4m5AaW5xKpnpE8TnWbQAq4YxLaqiAp8kLkGb9UtAQpEKGc1792mB67qvH1fFTqb7AKceQCFOmNEKkgcrwjuEEabrVAZitxWdebaaX07XoZpZrloHsjkw4prIe6RzfRd5DKCU1GMJDdQ3TJQJc5nkuLM2ZkEX0D

_No response_

asukaminato0721:
# Summary

Fixes #4686

`type(protocol_value)` now produces `type[Protocol]`, representing the unknown concrete runtime implementation.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/solve.rs
+++ b/pyrefly/lib/alt/solve.rs
@@ -6745,6 +6745,10 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
     // we'll fall back to builtins.type. We can add more cases here as-needed.
     pub fn type_of(&self, ty: Type) -> Type {
         match ty {
+            // A protocol value's runtime class is an unknown concrete implementation.
+            Type::ClassType(cls) if cls.class_object().is_protocol() => {
+                self.heap.mk_type_of(Type::ClassType(cls))
+            }
             Type::ClassType(cls) => self.heap.mk_class_def(cls.class_object().clone()),
             Type::SelfType(_) => self.heap.mk_type(ty),
             Type::Literal(lit) => self.heap.mk_class_def(
--- a/pyrefly/lib/test/protocol.rs
+++ b/pyrefly/lib/test/protocol.rs
@@ -1260,6 +1260,21 @@ issubclass(ConcreteClass, RuntimeProtocol)
 );
 
 testcase!(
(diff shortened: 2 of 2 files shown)


rchen152 (approved):

grievejia (approved):

PR #4690, opened 2026-08-27
Fix Stop stubgen retaining full analysis for the whole import closure

### LINKED ISSUE #4650: `stubgen` on a single file uses ~14 GB where `check` on the same file uses 81 MB
We generate a `.pyi` for one module of a large Django monorepo (about 18,000 Python files) as part of a codegen step that we recently moved from pyright to pyrefly. The module imports our Django models, so its transitive import closure covers most of the codebase.

`pyrefly check` handles this fine, but `pyrefly stubgen` on that single file uses around 14 GB of memory. On our 4 vCPU Linux CI runner it does not finish within 300 seconds. Numbers from an Apple Silicon Mac (16 cores, 64 GB RAM), pyrefly 1.1.1, same project config for both commands, same 850 line file:

```
$ pyrefly check --threads 1 rules_based_configuration/services/conformity.py
0.8 s wall, 81 MB peak RSS, 0 errors

$ pyrefly stubgen --threads 1 --output-dir /tmp/stubs rules_based_configuration/services/conformity.py
31 s wall, 14.2 GB peak RSS
```

I expected stubgen on one file to cost roughly what checking that file costs, but the memory gap is over two orders of magnitude. pyrefly 1.2.0 behaves the same (13.3 GB peak on the same file).

Project config, in case it matters: `[tool.pyrefly]` in pyproject.toml with `preset = "basic"`, `search-path = [".", "./typings", "../shared"]`, `infer-with-first-use = false`.

**Workaround.** Running stubgen with import resolution disabled brings it down to about 6 s and 3.2 GB:

```
pyrefly stubgen \
  --search-path <empty dir> \
  --disable-search-path-heuristics \
  --site-package-path <empty dir> \
  --skip-interpreter-query \
  --ignore-missing-imports '*' \
  --output-dir <out> \
  rules_based_configuration/services/conformity.py
```

For our file the emitted stub was identical except one annotation: a dataclass `__init__` parameter whose type is inferred from an imported class came out as `Incomplete` instead of the class name. So almost everything stubgen emits for us comes from annotations already written in the source file, yet the default run pays for analyzing the whole import closure. Even with every import unresolved, stubgen still peaks at 3.2 GB on this file.

The codebase is private, so I cannot share a runnable repro, and the behavior depends on the size of the import closure, so a sandbox link will not show it either. Happy to run instrumented builds or try patches on our repo if that helps. I searched the tracker for stubgen memory reports and only found general check/LSP memory issues like #3466.

Pager-dot:
Fixes #4650

## Problem

`pyrefly stubgen` on a single file cost memory proportional to the project's
import closure rather than to the input. The reporter saw ~14 GB peak RSS where
`pyrefly check` on the same file peaked at 81 MB, and it did not finish inside a
300 s CI budget.

## Cause

`pyrefly/lib/commands/stubgen.rs` built its transaction with
`new_transaction(Require::Everything, None)`. That first argument is
`default_require` the retention level for every module the transaction reaches
*implicitly*, i.e. the whole transitive import closure (`get_module` /
`get_imported_module` both read it).

`Require` is what gates eviction: at `Everything`, `post.evict_ast()` is skipped,
`module_parse` keeps full token data, and `enable_trace` builds the per-binding
answers trace table for every module in the closure. `check` avoids this by
using `RequireLevels { specified: Errors, default: Exports }`, and the LSP relies
on the same contract.

Stubgen never needed more. `extract_module_stub` reads `get_bindings` /
`get_answers` / `get_ast` only for the handle it is given, and that handle
already receives `Require::Everything` from the per-call
`transaction.run(&[handle], ...)`.

## Fix

Lower the transaction's `default_require` to `Require::Exports` and leave the
per-handle level untouched.

## Results

Measured on a file importing `torch`, release build, `--threads 1`:

| | wall | peak RSS |
|---|---|---|
| `stubgen` before | 33.15 s | 6.7 GB |
| `stubgen` after | 0.16 s | 91 MB |
| `check` (reference) | 0.14 s | 91 MB |

The generated `.pyi` is byte-identical between the two builds.

## Tests

- `assert_dependency_asts_evicted` is called from the shared stubgen test runner,
  so all 28 stubgen tests now assert that no module reached only through an import
  keeps its AST alongside its bindings. Reverting the level to `Require::Everything`
  fails 28/28, so the guard is not vacuous.
- `test_stubgen_dependency_types_survive_eviction` covers a type crossing a module
  boundary, pinning that it still resolves to the real class rather than degrading
  to `Incomplete`.
- No `expected.pyi` snapshot changed.

## Note on the mechanism

While implementing this I confirmed by instrumenting the compute loop that
dependencies only ever reach `Step::Answers` cross-module lookup demands
`Step::Answers`, never `Solutions`, so the `evict_answers` branch does not fire for
them at any require level. What `Exports` actually buys is AST eviction plus
disabling the answers trace table across the closure, which is exactly the profile
`check` runs at. The comments in the change describe that mechanism rather than the
`Solutions`-based one I first assumed.

### DIFF
--- a/pyrefly/lib/commands/stubgen.rs
+++ b/pyrefly/lib/commands/stubgen.rs
@@ -63,8 +63,13 @@ impl StubgenArgs {
         let state = State::new(config_finder, thread_count);
         let holder = Forgetter::new(state, false);
         let handles = Handles::new(expanded_file_list);
+        // `Require::Exports` is the retention level for every module reached through an import.
+        // Stubgen reads the AST, bindings and answers only of the module it is stubbing, so
+        // imported modules need to retain no more than `check` retains for them. Passing
+        // `Require::Everything` here instead pins the AST and the answers trace table for the
+        // whole import closure, making peak memory scale with the project rather than the input.
         let mut forgetter = Forgetter::new(
-            holder.as_ref().new_transaction(Require::Everything, None),
+            holder.as_ref().new_transaction(Require::Exports, None),
             true,
         );
         let transaction = forgetter.as_mut();
--- a/pyrefly/lib/stubgen.rs
+++ b/pyrefly/lib/stubgen.rs
@@ -13,6 +13,7 @@ mod tests {
     use std::path::PathBuf;
 
     use dupe::Dupe;
+    use pyrefly_build::handle::Handle;
     use pyrefly_util::forgetter::Forgetter;
     use pyrefly_util::fs_anyhow;
     use pyrefly_util::globs::FilteredGlobs;
@@ -26,6 +27,7 @@ mod tests {
     use super::extract::extract_module_stub;
     use crate::state::require::Require;
     use crate::state::state::State;
+    use crate::state::state::Transaction;
     use crate::test::util::TestEnv;
 
     fn run_stubgen(input: &str) -> String {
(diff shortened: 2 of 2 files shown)


rchen152 (approved):
Sorry for the delay! This looks good to me.

samwgoldman (approved):

PR #4697, opened 2026-08-27
Use a single queue for all TSP connections

samwgoldman:
Summary:
The TSP supports secondary ("extra") connections which Pylance uses to
communicate to Pyrefly from a background thread. Before this change, Pyrefly
dispatched requests received on extra connections directly in a separate thread.
With this change, the extra connection instead inserts the request into the
`lsp_queue`, so all requests are handled by the same thread.

Because we use a single loop for requests from different clients, it's important
that responses are routed to the correct one. To make this safer, this change
also refactors the structure of the TSP. Before, the event loop was owned by
`TspMainConnection` which also owns the `Sender` object, meaning all request
handlers had access to that sender from `self`. It seemed too risky to me that
code might accidentally send responses to the wrong client.

To address the risk, the event loop is now owned by `TspServer` and both the
main and extra connections provide a `Reply` object for each event: the server
owns no channel, and a handler can only reach the channel it was given.

This change also enables a stacked change -- all request handlers now share a
single `TransactionManager` object. The transaction manager is also used in the
LSP and helps deal with concurrent request handling, so a recheck and read-only
request can run in parallel and multiple read-only requests can build on top of
one another before a recheck commits. The stacked change uses that to optimize
the TSP.

Test Plan:
Replayed a captured Pylance session (pandas, 45251 TSP messages over both the
stdio and IPC connections) against `pyrefly tsp`. All 45102 requests were
answered on the connection that sent them and total request time is
unchanged.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/queue.rs
+++ b/pyrefly/lib/lsp/non_wasm/queue.rs
@@ -31,6 +31,7 @@ use pyrefly_util::telemetry::TelemetryEventKind;
 use tracing::debug;
 use tracing::info;
 
+use crate::lsp::non_wasm::protocol::Message;
 use crate::lsp::non_wasm::protocol::Request;
 use crate::lsp::non_wasm::protocol::Response;
 use crate::lsp::non_wasm::server::Server;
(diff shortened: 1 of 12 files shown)


kinto0 (approved):

PR #4698, opened 2026-08-27
Commit the solve behind a type query on an unopened file

samwgoldman:
Summary:
TSP supports requests on files which have not yet been opened. In this case, we
would create a throw-away transaction that solved the module to
`Step::Solutions` with `Require::Everything`, ran the request handler, and
returned the result. Solving the module for every request is extremely
expensive, but all that work is thrown away between requests. Furthermore, TSP
is fairly chatty, sending hundreds (sometimes thousands) of `getComputedType`
requests.

This change offers a simple but effective solution: when handling a request for
a non-open file, we create a "possibly committable" transaction to raise the
require level for that handle. We are either able to (1) get a committable
transaction which updates the committed state, so future requests share the
earlier solve work, or (2) we get a non-committable transaction, but the
`TransactionManager` maintains a saved state, so that multiple requests can
still share solve work even when a background recheck is happening. This is also
how the Pyrefly LSP works.

To test this change I captured and replayed a Pylance session which had 43,811
`getComputedType` requests (89% of them through the extra connection). Request
costs compared to parent, in ms:

```
  request          target    n      | before p50   p90    total | after p50   p90   total
  getComputedType  non-open  22294  |     16.886  57.82  667.79s |    0.067  0.114   3.43s
  getComputedType  open      21517  |      0.046   0.09    1.25s |    0.059  0.074   1.34s
  getSnapshot      -           191  |      0.069   7.18    1.14s |    0.068  6.632   1.05s
  resolveImport    -           994  |      0.130   1.38    0.58s |    0.087  1.229   0.42s
                                        total 670.95s               total 6.41s
```

Both runs also carried an unrelated `demand` fix that is not part of this stack,
worth roughly 0.2s of the "after" total.

Test Plan:
Answers are unchanged. Response payloads carry allocation-order identifiers
that differ between runs, so responses are compared ignoring digits: 43822 of
43873 (99.88%) match. All 51 that differ are `-32802` snapshot-outdated
responses, never a differing type. Two runs of the same binary differ
structurally in 78 responses, so the gap between the two binaries is within a
single binary's run-to-run variation.

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -482,7 +482,14 @@ pub trait TspInterface: Send + Sync + 'static {
     ///
     /// Returns `None` when the URI cannot be resolved, the position is invalid,
     /// or no type information is available at that location.
-    fn type_at_position(&self, uri: &str, line: u32, character: u32) -> Option<tsp_types::Type>;
+    fn type_at_position<'a>(
+        &'a self,
+        ide_transaction_manager: &mut TransactionManager<'a>,
+        telemetry_event: &mut TelemetryEvent,
+        uri: &str,
+        line: u32,
+        character: u32,
+    ) -> Option<tsp_types::Type>;
 
     /// Return the computed (inferred) type for a node spanning the given range,
     /// converted to the TSP wire format.
@@ -501,8 +508,10 @@ pub trait TspInterface: Send + Sync + 'static {
     /// declaration locations are resolved against the same warm transaction
     /// that produced the type, so the export lookups cannot hit a cold
     /// `get_stdlib`.
-    fn computed_type_at_range(
-        &self,
+    fn computed_type_at_range<'a>(
+        &'a self,
+        ide_transaction_manager: &mut TransactionManager<'a>,
+        telemetry_event: &mut TelemetryEvent,
         uri: &str,
         start_line: u32,
         start_character: u32,
@@ -514,8 +523,10 @@ pub trait TspInterface: Send + Sync + 'static {
     /// expected type — a call argument's parameter type, an annotated target's
     /// declared type, etc. — falling back to the computed type where no
     /// expected-type context applies.
-    fn expected_type_at_position(
-        &self,
+    fn expected_type_at_position<'a>(
+        &'a self,
+        ide_transaction_manager: &mut TransactionManager<'a>,
+        telemetry_event: &mut TelemetryEvent,
         uri: &str,
         line: u32,
         character: u32,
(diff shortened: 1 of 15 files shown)


pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -6515,56 +6526,149 @@ impl Server {
...
+    /// open file is up to date, then commit that transaction if we can.
+    ///
+    /// Unlike the LSP, which receives a document sync and a request as separate
+    /// events, a TSP query arrives with everything needed to serve it, so the
+    /// validation, the solve and the answer all happen in one transaction. That
+    /// holds the committing lock across the query rather than just the solve,
+    /// which is a little longer but avoids resolving a handle against state we
+    /// then have to re-acquire.
+    ///
+    /// A file the client never opened is solved at `Require::Everything`, the
```
kinto0: I think this docstring could be a lot less verbose

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -6515,56 +6526,149 @@ impl Server {
...
+            ide_transaction_manager.get_possibly_committable_transaction(&self.state);
+        let transaction = match &mut possibly_committable {
+            Ok(committable) => committable.as_mut(),
+            Err(transaction) => transaction,
+        };
+
+        // Bring in-memory content up to date. `didChange` skips this step when
+        // another mutation is already queued, so the overlay can be behind
+        // `open_files` even though the queue has drained.
+        transaction.set_memory(
```
kinto0: this is the same as just running validate_in_memory_for_transaction, right? can we reuse that?

pyrefly/lib/tsp/server.rs
kinto0, line 236: I think we need to update this too if we want to be able to save the transactions from this call. It looks like it was missed because it already passed ide_transaction_manager.

kinto0 (changes requested):
great change, thank you!!

would love to see tests on the following:

  1. Repeated unopened-file query reuse
     Query the same unopened file twice and verify the second query performs no solve work. This is the primary purpose of the commit.

  2. Deferred didChange visibility
     Arrange didChange(v2) → type query → later mutation, causing validation of v2 to be deferred. Verify the query returns the type from v2 rather than stale shared state.

  3. Non-committable fallback
     Hold the recheck lock, issue a type query, and verify its solved transaction is saved. Release the lock and verify the next query reuses and commits it.

  4. Intervening resolveImport
     After saving a query transaction, call resolveImport, then issue another query. This would expose the transaction-loss issue from my review.

  5. Invalidation after caching
     Query an unopened file, change it, advance the snapshot, and verify the next query recomputes rather than returning the cached old type.

  6. Explicit queue-draining case
     Use a type conversion known to lazily enqueue additional work and verify the transaction commits successfully. Existing type-shape tests may exercise this accidentally, but none assert the invariant.

  7. Other unopened query kinds
     Only getComputedType tests unopened files. getDeclaredType and getExpectedType now use the same new path but lack equivalent coverage.

samwgoldman:
I tried to simplify the implementation a bit. Probably best to review the commits in order, since the first commit sets up a bit of infra used in the second.

pyrefly/lib/state/state.rs
```diff
@@ -1076,6 +1077,17 @@ impl<'a> Transaction<'a> {
         self.data.state.config_finder.errors()
     }
 
+    /// The `Require` level this transaction retains `handle` at, or `None` if
+    /// it has no such module. Only `Require::Everything` keeps the AST and
+    /// answers that a type lookup reads.
```
kinto0: Let's remove the second sentence here. It doesn't belong on this function

pyrefly/lib/state/state.rs
```diff
@@ -3451,6 +3472,19 @@ impl State {
         let start = Timer::start();
         let readable = self.state.read();
         let state_lock_blocked = start.elapsed();
+        self.transaction_with_guard(readable, default_require, subscriber, state_lock_blocked)
+    }
+
+    /// Build a transaction over the state `readable` observes. Takes the guard
+    /// rather than acquiring one, so a caller already holding the read lock
+    /// reuses it instead of dropping it and racing for another.
+    fn transaction_with_guard<'a>(
```
kinto0: nit: maybe `transaction_from_guard` makes more sense? when I see this API, I think it'll return a transaction with a guard. but as I understand it this is the way to get a transaction with a guarded state input

kinto0 (approved):
review for commit 1

pyrefly/lib/test/tsp/tsp_interaction/unopened_files.rs
```diff
@@ -0,0 +1,170 @@
...
+use crate::test::tsp::tsp_interaction::object_model::write_pyproject;
+
+/// A project whose open file imports a module the client never opens.
+/// Returns the interaction, the directory (kept alive for the test), the
+/// unopened module's URI, and the current snapshot.
+fn setup_with_unopened_import(helper_content: &str) -> (TspInteraction, TempDir, String, i32) {
+    let temp_dir = TempDir::new().unwrap();
+    write_pyproject(temp_dir.path());
+    std::fs::write(temp_dir.path().join("helper.py"), helper_content).unwrap();
+    std::fs::write(
```
kinto0: nit: I find it easier to write these to disk instead of doing it as part of the test

pyrefly/lib/test/tsp/tsp_interaction/unopened_files.rs
```diff
@@ -0,0 +1,170 @@
...
+        temp_dir.path().join("main.py"),
+        "import helper\nx = helper.value\n",
+    )
+    .unwrap();
+
+    let mut tsp = TspInteraction::new();
+    tsp.set_root(temp_dir.path().to_path_buf());
+    tsp.initialize(Default::default());
+    tsp.server.did_open("main.py");
+    tsp.client.expect_any_message();
```
kinto0: not sure if this was copied from other tests but it doesn't seem great to have in these tests. can we expect a specific message or leave this out?

pyrefly/lib/test/tsp/tsp_interaction/unopened_files.rs
```diff
@@ -0,0 +1,170 @@
...
+        "An unknown path should not be an error: {:?}",
+        resp.error
+    );
+    assert!(
+        resp.result.expect("Expected a result").is_null(),
+        "An unknown path has no type"
+    );
+
+    tsp.shutdown();
+}
```
kinto0, line 190: the most obvious test I missed last time (sorry) is an open file after it has been queried as unopened. if we open it, modify it, it should show the in-memory data

pyrefly/lib/lsp/non_wasm/server.rs
```diff
@@ -6479,49 +6490,80 @@ impl Server {
...
     }
 
-    /// Open `uri` at `(line, character)`: resolve the path, build a handle, and
-    /// start a transaction, returning it alongside the handle and the resolved
-    /// in-file position.
-    fn open_at_position<'a>(
+    /// Resolve `uri` and `(line, character)` into a transaction, the handle the
+    /// file is analyzed under, and the position within it. Disposal follows
+    /// [`Self::query_transaction_and_handle`].
+    fn position_transaction_and_handle<'a>(
```
kinto0: it looks like it's necessary to save the transaction to the manager after getting this transaction (all the callsites of this now call `ide_transaction_manager.save(transaction, telemetry_event)`). if so, let's make it obvious here that. that is somewhat awkward, so I wonder if we should instead make this function a closure and have it happen automatically

kinto0 (approved):
commit 2 looks good

Issue #4450, opened 2026-08-05
avoid storing line numbers in the baseline file


DetachHead:
from my understanding the line number is not used when matching errors against the baseline file, so it doesn't need to be stored.

in my experience, storing the line number results in very noisy diffs when updating the baseline file, and also means it needs to be updated much more often, since most changes to a file would result in most of the line numbers being changed (which would be annoying when #4449 is done)

yangdanny97:
I initially included it to make the file format more useful for humans, so that they can more precisely locate an error by reading the file.

Open to revisiting that decision though

tobyh-canva:
I think the same goes for the `description` field which is quite unstable - for missing import errors, it includes the absolute paths for search-path, site-package-path, etc. There's already a `concise_description` so I think the full description unfortunately just adds noise

tobyh-canva:
Somewhat related to this issue: since we just added support for the SARIF output format, SARIF has the concept of "fingerprints" to keep track of findings even if they move around the file. Instead of storing line / column numbers, you derive a fingerprint from information and context about the finding (e.g. surrounding AST, containing method name, etc.). How the fingerprint is derived is totally up to us, [GitHub code scanning](https://docs.github.com/en/enterprise-cloud@latest/code-security/reference/code-scanning/sarif-files/sarif-support#fingerprint-generation) has a somewhat crude generic approach.

So this is not necessarily saying we have to switch baseline files to SARIF format, but this concept of a fingerprint would be useful to both baseline files and for enriching Pyrefly's SARIF output.

yangdanny97:
I am open to a breaking change to the baseline file format/matching algorithm, since it's an experimental feature and can be easily re-generated upon upgrade.

tobyh-canva:
I attempted a PoC for fingerprinting support, and TBH it's tricky to retrofit an AST-based fingerprinting approach to Pyrefly, especially without affecting performance. And FWIW, I've changed my mind and think it would be an over-engineered approach to baseline files.

Fundamentally, a baseline just needs to prevent regression. Because it needs to be committed to source control, we want to design its content such that:

1. Churn is minimised. The file only needs to be update when the baseline meaningfully changes (e.g. a finding has been addressed, or newly introduced errors are being baselined).
2. Finding identification is stable. If a baselined finding moves around in the source code, it shouldn't be flagged as a new error. Similarly, a new finding should not be suppressed by the baseline.

[...]

PR #4700, opened 2026-08-28
feat(baseline): support configurable baseline matching behaviour and file format

tobyh-canva:
Provides users with some flexibility about how baselines are configured in Pyrefly, as discussed in https://github.com/facebook/pyrefly/issues/4450#issuecomment-5420902969.

**New config options:**

- `baseline-matching-mode = "column" (default) | "concise_description"`: The criteria used to match a finding with a baseline entry.
- `baseline-format = "full" (default) | "minimal"`: The amount of context encoded in the baseline file, with "minimal" only encoding the minimum fields necessary to match a finding with the baseline.

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -526,6 +526,40 @@ impl ImportLookupPathPart<'_> {
     }
 }
 
+/// Fields used to match diagnostics against baseline entries.
+#[derive(Debug, Clone, Copy, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum BaselineMatchingMode {
+    /// Match by path, error kind, and starting column.
+    #[default]
+    Column,
+    /// Match by path, error kind, and concise description.
+    ConciseDescription,
+}
+
+impl BaselineMatchingMode {
+    fn is_default(&self) -> bool {
+        *self == Self::default()
+    }
+}
+
+/// Amount of diagnostic information written to a baseline file.
+#[derive(Debug, Clone, Copy, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum BaselineFormat {
+    /// Write all available baseline metadata.
+    #[default]
+    Full,
+    /// Write only the fields required for matching.
+    Minimal,
+}
+
+impl BaselineFormat {
+    fn is_default(&self) -> bool {
+        *self == Self::default()
+    }
+}
+
 #[skip_serializing_none]
 #[derive(Debug, Deserialize, Serialize, Clone, Derivative)]
 #[serde(rename_all = "kebab-case")]
@@ -617,6 +651,14 @@ pub struct ConfigFile {
     /// Severity assigned to errors that match the baseline.
     /// Defaults to `ignore`.
     pub baseline_error_level: Option<Severity>,
+
+    /// Fields used to match diagnostics against baseline entries.
+    #[serde(default, skip_serializing_if = "BaselineMatchingMode::is_default")]
+    pub baseline_matching_mode: BaselineMatchingMode,
+
+    /// Amount of diagnostic information written to the baseline.
+    #[serde(default, skip_serializing_if = "BaselineFormat::is_default")]
+    pub baseline_format: BaselineFormat,
 
     /// Default error output format for CLI checks when `--output-format` is not set.
     pub output_format: Option<OutputFormat>,
@@ -734,6 +776,8 @@ impl Default for ConfigFile {
             typeshed_path: None,
             baseline: None,
             baseline_error_level: None,
+            baseline_matching_mode: BaselineMatchingMode::default(),
+            baseline_format: BaselineFormat::default(),
             min_severity: None,
             output_format: None,
             skip_lsp_config_indexing: false,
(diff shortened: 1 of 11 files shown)


crates/pyrefly_config/src/config.rs
```diff
@@ -526,6 +526,40 @@ impl ImportLookupPathPart<'_> {
     }
 }
 
+/// Fields used to match diagnostics against baseline entries.
+#[derive(Debug, Clone, Copy, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
```
rchen152, line 531: IMO we should use kebab case for `BaselineMatchingMode` and `BaselineFormat` for consistency with other config fields. Otherwise `concise_description` will be the only snake case option.

pyrefly/lib/commands/check.rs
rchen152, line 404: Do you want `baseline-format` to have a corresponding CLI flag? I feel like it could go either way - it feels sort of natural to be able to specify the format along with `--update-baseline`, but generally it's probably better to specify the format in the config file so two updates can't generate conflicts.

Up to you, just wanted to raise the question. If you do want a flag, the way to do it would be to add an `Option` field here and override the value from the config file when the flag is non-`None`.

(I think not having a flag for `baseline-matching-mode` is the right choice, since that describes a property of the checked-in baseline.)

website/docs/configuration.mdx
```diff
@@ -527,7 +527,40 @@ more details on how baseline files work.
     - `baseline` is a **project-level setting** and cannot be overridden in
       [`sub-config`](#sub-configs) sections.
     - Errors suppressed by the baseline file are still shown in the IDE.
-    - Errors are matched with the baseline by file, error code, and column number.
+
+### `baseline-matching-mode`
```
rchen152, line 531: I'd suggest adding a brief note here that there's intentionally no flag equivalent and why. Same for `baseline-format` if you decide not to add a flag.

pyrefly/lib/commands/check.rs
```diff
@@ -2007,7 +2033,8 @@ impl CheckArgs {
                 baseline_path,
                 &BaselineErrors {
                     errors: retained_baseline_entries,
-                },
+                }
+                .with_format(defaults.baseline_matching_mode, defaults.baseline_format),
```
rchen152, line 2037: I feel like this could lead to odd interactions if a user changes the baseline format and then runs `--prune-baseline`. If they changed full -> minimal, this will change the format, but only if the baseline needs pruning. minimal -> full obviously won't work. Maybe pruning should just ignore the format?

On a different note, requiring callers of `write_baseline_errors_to_file` to pre-format the error seems a little error-prone (it's easy to forget `with_format`). Might be better to remove the `write_baseline_to_file` wrapper and have `write_baseline_errors_to_file` directly take the mode and format.

pyrefly/lib/error/baseline.rs
```diff
@@ -34,50 +43,88 @@ pub(crate) fn normalize_baseline_path(path: &Path, relative_to: &Path) -> String
...
-    fn from_baseline_error(error: &BaselineError, relative_to: &Path) -> Self {
-        Self {
+    fn from_baseline_error(
+        error: &BaselineError,
+        relative_to: &Path,
+        matching_mode: BaselineMatchingMode,
+    ) -> Result<Self> {
+        let matching_field = match matching_mode {
+            BaselineMatchingMode::Column => BaselineMatchingField::Column(error.column.context(
+                "missing field `column`, required by `baseline-matching-mode = \"column\"`",
```
rchen152, line 53: It would be nice if we could print more information about the malformed error. Otherwise it could be hard to figure out what's wrong in a large baseline file.

pyrefly/lib/error/legacy.rs
```diff
@@ -123,13 +129,18 @@ impl LegacyErrors {
...
     pub path: String,
     /// The kebab-case name of the error kind.
     pub name: String,
-    concise_description: String,
-    #[serde(default = "default_legacy_severity")]
-    severity: Severity,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub concise_description: Option<String>,
+    #[serde(
+        default = "default_baseline_severity",
```
rchen152, line 140: Seems a little odd to invent a severity when there wasn't one. From what I can tell, the severity is just metadata, so if it's missing, it would be better to just default to `None`.

schemas/pyrefly.json
```diff
@@ -247,6 +247,18 @@
           "enum": ["ignore", "info", "warn", "error"],
           "default": "ignore"
         },
+        "baseline-matching-mode": {
```
rchen152, line 250: Could you also update the test files? https://github.com/facebook/pyrefly/blob/main/schemas/CONTRIBUTING.md#4-add-test-cases

rchen152 (changes requested):
Thanks for the PR! This looks really good. Back to you with some comments. The main things I'd like to resolve are the question of CLI flags (and updating the new documentation accordingly) and `--prune-baseline`'s behavior - the rest are essentially just polish suggestions.

pyrefly/lib/commands/check.rs
tobyh-canva: Yeah it's a good point, I did think about this but haven't documented the decision. I think both options similarly describe properties of the checked-in baseline file, so for that reason I'd keep both as config-only. I'll mention this in the docs.

pyrefly/lib/commands/check.rs
tobyh-canva: Yep good callouts, I'll make it so pruning just prunes instead of rewriting any existing entries.

tobyh-canva: I kept both functions but renamed them for clarity. The `--prune-baseline` path still needed a way to serialize pre-formatted baseline errors so that the existing format was preserved :)

rchen152 (approved):

rchen152:
LGTM, thanks again!

PR #4702, opened 2026-08-28
Reuse transactions across repeated TSP type queries

### LINKED ISSUE #4312: Avoid repeated Require::Everything solves for TSP type queries
### Summary

While profiling Pyrefly as Pylance's TSP backend, we found that type queries against imported, non-open modules repeatedly invoke:

```rust
transaction.run(&[handle.dupe()], Require::Everything, None);
```

in [`query_transaction_and_handle`](https://github.com/facebook/pyrefly/blob/e096a476f7c87d80173d2449d174ac62aa106252/pyrefly/lib/lsp/non_wasm/server.rs#L6520-L6555).

This dominates cold interactive latency. Reusing valid retained answers reduced a real Django hover from **2.84s to 1.33s (53%)**.

### Environment

- Pyrefly: `e096a476f7c87d80173d2449d174ac62aa106252`
- Version: `1.2.0-dev.3`
- Build: release
- Workload: hover on `django.db.models.base.Model`
- Django: `957d0cee7167757ae221ffde59d2cf0a322e89c7`
- Windows 11, AMD EPYC 7763

### Measurements

Five fresh-process measurements:

| Metric | Median |
|---|---:|
| Open/prewarm | 789ms |
| Cold hover | 2,837ms |
| Warm hover | 0.250ms |

The cold hover issued 57 distinct `getComputedType` requests:

| Component | Time |
|---|---:|
| Pyrefly/TSP | 2,502ms |
| 57 `getComputedType` calls | 2,476ms |
| Pylance parsing, binding, and provider work | 335ms |

Additional Pyrefly instrumentation captured 47 requests:

| Phase | Total | Median/request |
|---|---:|---:|
| Entire handler | 2,111.985ms | 18.110ms |
| `query_transaction_and_handle` | **2,110.246ms** | 18.085ms |
| Range type lookup | 0.371ms | 0.007ms |
| TSP conversion | 1.081ms | 0.023ms |

The type lookup and wire conversion are negligible; almost all time is spent ensuring `Require::Everything`.

### Experimental result

As a deliberately incomplete experiment, we skipped `run` when answers were already present:

```rust
if transaction.get_answers(&handle).is_none() {
    transaction.run(&[handle.dupe()], Require::Everything, None);
}
```

Results:

| Metric | Baseline | Experiment | Change |
|---|---:|---:|---:|
| Cold hover | 2,837ms | 1,334ms | **-53%** |
| Open + cold hover | 3,603ms | 2,065ms | **-43%** |
| Computed-type requests | 57 | 40 | -30% |

This guard is not proposed as the final fix because answer validity must account for edits and dependency invalidation.

### Suggested direction

Could Pyrefly provide a revision-aware operation equivalent to:

```text
ensure_current_answers(handle, Require::Everything)
```

It should:

- Reuse retained answers when they are valid for the current state generation.
- Recompute after source, dependency, configuration, or overlay changes.
- Coalesce concurrent work for the same `(handle, revision)`.
- Avoid repeating a full solve for multiple range queries against the same module.
- Potentially support a query-specific requirement below `Require::Everything` if range type lookup does not need every retained artifact.

Useful regression scenarios would include repeated queries across:

- Imported closed files.
- Source and dependency edits.
- Open/close overlay transitions.
- Configuration changes.
- Notebook cells.
- Concurrent queries for one module revision.

### Secondary finding: TSP indexing

Launching TSP with `--indexing-mode none` reduced cold hover from **2,837ms to 2,036ms (28%)**, with warm hover unchanged.

For hosts such as Pylance that provide their own indexing and language-service layer, Pyrefly's background workspace index appears redundant. It may be worth making indexing opt-in for TSP, or documenting `--indexing-mode none` as the recommended configuration for such hosts. Standalone LSP behavior should remain unchanged.

Punisheroot:
# Summary

Fixes #4312.

Reuse the existing revision-aware `TransactionManager` for TSP type queries instead of creating and discarding a fresh transaction for every request.

This applies to:

- `getDeclaredType`
- `getComputedType`
- `getExpectedType`
- `resolveImport`

Both the main TSP connection and auxiliary IPC connections now retain transaction data across requests. The retained transaction is automatically replaced when the committed `State` revision changes.

Repeated type queries against imported but unopened modules previously started from a fresh transaction each time. Consequently, distinct `getComputedType` requests for the same module and snapshot repeatedly ran the module's solve work.

With this change:

- the first query solves the imported module;
- subsequent queries on the same TSP connection and snapshot reuse its answers;
- after a file change produces a new snapshot, the module is recomputed normally.

## Performance evidence

The regression test records transaction telemetry for two different ranges in the same unopened imported module:

- first query: `step_answers_count > 0`;
- second query: `step_answers_count == 0`;
- second query: `run_todo_count == 0`.

In the synthetic test logs, the first query took approximately `0.09s`, while the second was reported as `0.00s` at the current logging precision.

This does not reproduce the full Django/Pylance benchmark from the issue, but it verifies that the repeated solve responsible for the reported regression is eliminated within the same connection and snapshot.

## Correctness

The regression test also modifies the imported file and advances the snapshot. The following query:

- performs solve work again;
- observes the updated type.

Transaction reuse is scoped per TSP connection; separate auxiliary connections may each perform their own initial solve.

# Test Plan

- Focused regression test passed
- TSP interaction suite: 95 passed
- Full `pyrefly` package: 8195 passed, 5 ignored
- Repository formatting and lint checks passed
- `git diff --check` passed

### DIFF
--- a/pyrefly/lib/lsp/non_wasm/server.rs
+++ b/pyrefly/lib/lsp/non_wasm/server.rs
@@ -482,7 +482,13 @@ pub trait TspInterface: Send + Sync + 'static {
     ///
     /// Returns `None` when the URI cannot be resolved, the position is invalid,
     /// or no type information is available at that location.
-    fn type_at_position(&self, uri: &str, line: u32, character: u32) -> Option<tsp_types::Type>;
+    fn type_at_position(
+        &self,
+        transaction: &mut Transaction,
+        uri: &str,
+        line: u32,
+        character: u32,
+    ) -> Option<tsp_types::Type>;
 
     /// Return the computed (inferred) type for a node spanning the given range,
     /// converted to the TSP wire format.
@@ -503,6 +509,7 @@ pub trait TspInterface: Send + Sync + 'static {
     /// `get_stdlib`.
     fn computed_type_at_range(
         &self,
+        transaction: &mut Transaction,
         uri: &str,
         start_line: u32,
         start_character: u32,
@@ -516,6 +523,7 @@ pub trait TspInterface: Send + Sync + 'static {
     /// expected-type context applies.
     fn expected_type_at_position(
         &self,
+        transaction: &mut Transaction,
         uri: &str,
         line: u32,
         character: u32,
@@ -6504,12 +6512,12 @@ impl Server {
         )
     }
 
-    /// Build a read transaction and the handle the type checker analyzes `path`
-    /// under, so `(uri, range)` queries resolve for any analyzable file rather
-    /// than only open documents.
+    /// Return the handle the type checker analyzes `path` under, so `(uri,
+    /// range)` queries resolve for any analyzable file rather than only open
+    /// documents.
     ///
-    /// Open files are served from their in-memory overlay (already committed by
-    /// the recheck that ran on `didOpen`). For anything else we reuse the handle
+    /// Open files are served from the in-memory overlay already incorporated in
+    /// the supplied transaction. For anything else we reuse the handle
     /// the file was already analyzed under — an imported dependency's filesystem
     /// handle, or a bundled stdlib stub's `BundledTypeshed` handle whose
     /// `SysInfo` we can't reconstruct here, hence the by-path lookup — and force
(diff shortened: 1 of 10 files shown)


kinto0:
see comment on https://github.com/facebook/pyrefly/issues/4312

PR #4704, opened 2026-08-28
fix / is not supported between Tensor and Module in pytorch #3812

### LINKED ISSUE #3812: / is not supported between Tensor and Module in pytorch
### Describe the Bug

https://github.com/open-spaced-repetition/srs-benchmark/blob/3e4ae401f19def8f3414b3d0709c33b73cc39a41/models/gru.py#L95

`/` is not supported between `Tensor` and `Module`
  Argument `Module` is not assignable to parameter `other` with type `Tensor | bool | complex | float | int` in function `torch._C.TensorBase.__truediv__`Pyrefly [unsupported-operation](https://pyrefly.org/en/docs/error-kinds/#unsupported-operation)

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3812

# Test Plan

add regression test

### DIFF
--- a/pyrefly/lib/test/nn_module.rs
+++ b/pyrefly/lib/test/nn_module.rs
@@ -9,11 +9,20 @@ use crate::test::util::TestEnv;
 use crate::testcase;
 
 fn torch_env() -> TestEnv {
-    TestEnv::one(
-        "torch.nn",
+    let mut env = TestEnv::one(
+        "torch",
         r#"
 class Tensor:
+    def __truediv__(self, other: Tensor | float) -> Tensor: ...
     def tolist(self) -> list[int]: ...
+
+def tensor(value: float) -> Tensor: ...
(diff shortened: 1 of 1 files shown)


rchen152:
This PR just seems to add a test - is that because the issue is already fixed? Did you confirm this using the actual `pytorch` package?

asukaminato0721:
image

https://github.com/facebook/pyrefly/pull/4320

stroxler:
I think if the original bug was just a stub issue (or a downstream effect of not supporting register_buffer), we don't really need a regression test. We can just close the original.

Feel free to reopen if there's something I'm not following, but we probably don't want tests in Pyrefly against bugs that never existed per-se (either because they had some other root cause or they were a stub issue), because the test harness does come with some cost as it grows

PR #4705, opened 2026-08-28
fix decorator error message is very confusing #3569

### LINKED ISSUE #3569: decorator error message is very confusing

asukaminato0721:
# Summary

Fixes #3569

When a decorated function omitted an injected parameter, such as current_tenant_id: str, Pyrefly previously reported a confusing callable type mismatch.

now detected this case heuristically and replaced it with a clearer message saying that the decorated function was missing a parameter of type str.

# Test Plan

add test


stroxler (approved):

PR #4707, opened 2026-08-28
don't enforce __slots__ on plain subclasses of slotted dataclasses #4616

### LINKED ISSUE #4616: False positive [missing-attribute] when inheriting slotted dataclass
### Describe the Bug

I think the following should type check, it works at runtime and no other type checker complains.

```python

@dataclass(slots=True)
class Foo:
    x: int
    y: int
        
class Bar(Foo):
    def __init__(self, x: int, y: int) -> None:
        super().__init__(x, y)
        self.z = x + y  # ERROR

assert "__slots__" not in Bar.__dict__  # OK
assert Bar(1, 2).__dict__ == {"z": 3}  # OK
```

```
ERROR sandbox.py:11:14-15: Object of class `Bar` has no attribute `z` (not declared in `__slots__`) [missing-attribute]
```

For non-dataclasses, `pyrefly` reports no error just like all other type checkers: [[sandbox]](https://pyrefly.org/sandbox/?project=v2.pZRBDsIgEEWvQuyiNZqy79KFaw_QpCFKqokpDVADnt4PFA0aurArkgkz82He_JxlhlU5CjG_uOvUXWjlpgMVpt3scVinJLtGpsGA4FTWn9k9MqhovmIWMfuJxc09MFlBUoT0v57TyGW1rd-JBrfBS9L_6TSRndPwO-aCnEKAzD8GgwMf_QQbxdzrdijgoJxctR5VQ2lMF7KnfKAXcVY0yaDEEc0e7ObNOS1HhMdSoewavsqAV7lEV-ngwo1FYF4)

```python
class Foo:
    __slots__ = "x", "y"

    def __init__(self, x: int, y: int) -> None:
        self.x = x
        self.y = y
        
class Bar(Foo):
    def __init__(self, x: int, y: int) -> None:
        super().__init__(x, y)
        self.z = x + y  # OK
```

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZRBC4JAEIX_ypAHjcKlugldgrp0MDx7kRQLwhVXQ43-e29clSTZS54WeTOOs997xsg02X5qSEc9ZKn2vG7sXvvsJGW_rtrDVZX63Hyd-RnUh6hwUDEAOGNL3WXdd5j1ZZUnhbN0x8Iaaswz8W2LPda0oobIomMQ-IEGlxO24677FQaPcBHMGEZDyxjxDDhR5J9HPU-9WdOWv9kLQO8rXLTglnZvLf9FxaKLfkH91hGSYCytEMVgxw0zCymc0K0sc-UJMZTLIhVJJmJ5VWJSIYhdET2jexfw03YkO7QV2v7DqK0RtU2E2gwoFEboPg

_No response_

shobhitmehro:
Pyrefly reported a false `missing-attribute` error when a plain subclass of a slotted dataclass assigned new attributes, even though this is valid at runtime. The fix stops slot enforcement from applying to subclasses that do not declare slots themselves.

Fixes #4616

# Test Plan

`python3 test.py`

### DIFF
--- a/pyrefly/lib/alt/attr.rs
+++ b/pyrefly/lib/alt/attr.rs
@@ -1168,8 +1168,9 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
     /// is `true` when `__dict__` appears in slots.
     fn extract_local_slot_names(&self, cls: &Class) -> Option<(SmallSet<Name>, bool)> {
         let metadata = self.get_metadata_for_class(cls);
-        if let Some(dc) = metadata.dataclass_metadata()
-            && dc.kws.slots
+        // An inherited `slots=True` does not synthesize `__slots__` on the subclass.
+        if metadata.has_local_dataclass_slots_request()
+            && let Some(dc) = metadata.dataclass_metadata()
         {
             let mut names = SmallSet::new();
             let has_dict = dc.fields.iter().any(|n| *n == dunder::DICT);
--- a/pyrefly/lib/test/slots.rs
+++ b/pyrefly/lib/test/slots.rs
@@ -75,6 +75,70 @@ class Child(Base):
 "#,
 );
 
+// A plain subclass of `@dataclass(slots=True)` declares no `__slots__`, so at runtime it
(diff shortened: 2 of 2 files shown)


stroxler (approved):

PR #4709, opened 2026-08-28
Check Ruff formatting in GitHub CI (#4709)

stroxler:
Summary:

External tensor-shape contributors had no documented way to reproduce the formatter used by the repository, and GitHub CI did not catch formatting drift. Document the pinned Ruff command and add a fast CI job that checks the full tensor-shapes source corpus.

The documentation-only `tensor-shapes/skills` directory is excluded because Ruff also reformats fenced examples that are not corpus source.

Differential Revision: D117895283


rchen152 (approved):

PR #4708, opened 2026-08-28
[tensor-shapes] add stubs for JAX elementwise functions

jakevdp:
### Summary

Update `tensor-shapes/pyrefly-jax-stubs` with shape annotations for elementwise functions

### Test Plan

Unit test file updated.

### AI usage

Changes in this PR were generated with a Gemini coding agent.

PR #4710, opened 2026-08-28
[tensor-shapes] add stubs for jax.numpy.linalg submodule

jakevdp:
### Summary

Update `tensor-shapes/pyrefly-jax-stubs` with shape annotations for `jax.numpy.linalg` APIs. Followup to #4708.

This also includes improving the `matmul_shape` DSL function and applying it to `jax.numpy.matmul`.

### Test Plan

Unit test file updated; new `test_linalg.py` file created. Tests pass locally:
```
$ uv tool run --from ruff==0.16.5 ruff format tensor-shapes
254 files left unchanged
Running formatting...
Finished in 2.62 seconds.
Running linting...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.65s
Finished in 0.88 seconds.
    Finished `release` profile [optimized + debuginfo] target(s) in 0.74s
PASS stubs (7 files)
PASS arithmetic (1 files)
PASS creation (1 files)
PASS fft (1 files)
PASS linalg (1 files)
PASS matmul (1 files)
PASS nn (1 files)
PASS reductions (1 files)
PASS reshape (1 files)
```

### AI usage

Changes in this PR were generated with a Gemini coding agent.

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/linalg/__init__.pyi
+++ b/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/linalg/__init__.pyi
@@ -1,0 +1,297 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Literal, overload, Sequence
+
+from jax._array import Array
+from jax._shapes import (
+    int_min,
+    matmul_shape,
+    reduce_shape,
+)
+from shape_extensions import broadcast, Elements, Flag, Int, IntTuple, IntVar
+
+type _Shape = IntTuple
+type _Axis = int | tuple[int, ...] | None
+
+def cholesky[Batch: IntTuple, N: IntVar](
+    a: Array[[*Elements[Batch], N, N]],
+    *,
+    upper: bool = False,
+    symmetrize_input: bool = True,
+) -> Array[[*Elements[Batch], N, N]]: ...
+def cond[Batch: IntTuple, M: IntVar, N: IntVar](
+    x: Array[[*Elements[Batch], M, N]],
+    p: Any = None,
+) -> Array[Batch]: ...
+def cross[Shape1: _Shape, Shape2: _Shape](
+    x1: Array[Shape1],
+    x2: Array[Shape2],
+    /,
+    *,
+    axis: int = -1,
+) -> Array[broadcast(Shape1, Shape2)]: ...
+def det[Batch: IntTuple, N: IntVar](
+    a: Array[[*Elements[Batch], N, N]],
+) -> Array[Batch]: ...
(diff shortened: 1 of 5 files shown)


tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/linalg/__init__.pyi
```diff
@@ -0,0 +1,297 @@
...
+    *,
+    keepdims: bool = False,
+    ord: Any = "fro",
+) -> Array[IntTuple]: ...
+def matrix_power[Batch: IntTuple, N: IntVar](
+    a: Array[[*Elements[Batch], N, N]],
+    n: int,
+) -> Array[[*Elements[Batch], N, N]]: ...
+def matrix_rank[Batch: IntTuple, M: IntVar, N: IntVar](
+    M: Array[[*Elements[Batch], M, N]],
```
stroxler: Is this intentional? `M` is an unusual name for a variable, plus it's shadowing the type variable `M`

stroxler (changes requested):
Back to you just to check that the `M` isn't a typo. Other than that LGTM

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/linalg/__init__.pyi
stroxler: Ok I checked and I think `M` is intentional, but maybe we should pick a different name for the type variable?

I wonder if, given jax variable naming conventions, we should actually consider using a different range like `A`, `B`, `C` as default type var names in jax stubs. Torch and numpy don't use capitalized varnames, so `M` and `N` make more sense there.

jakevdp: Hmm, yeah the argument is named `M` unfortunately: https://github.com/jax-ml/jax/blob/3d4e5458bacc38451888086254985b1f18364e53/jax/_src/numpy/linalg.py#L417

I'll update the annotations to use a different character.

jakevdp: JAX's naming here derives from NumPy, which called this argument `M` prior to v1.22: https://numpy.org/doc/1.21/reference/generated/numpy.linalg.matrix_rank.html

jakevdp: > I wonder if, given jax variable naming conventions, we should actually consider using a different range like `A`, `B`, `C` as default type var names in jax stubs.

JAX doesn't typically use capital letter variable names, except in places where NumPy does (or did) so. But even here, NumPy now uses `A`... so I think we just stick with the `N`, `M` convention, and modify it locally as needed.

stroxler (approved):

PR #4714, opened 2026-08-29
Make pyrefly pass the new TypedDict conformance tests

yangdanny97:
# Summary

https://github.com/python/typing/pull/2338 added a bunch of conformance tests and rules in the spec.

This PR adds those test cases to Pyrefly's suite and makes Pyrefly pass them.

Only the open typed dict rules get the special open-unpacking error code to be off-by-default, the rest get normal errors.

# Test Plan

added tests


rchen152 (approved):
LGTM, sorry for taking so long TT (Will need a second review, since I had to commandeer the internal diff.)

kinto0 (approved):

PR #4729, opened 2026-08-30
detect basedpyright config

### LINKED ISSUE #4382: BasedPyright config detection/migration
Similar to the behavior we currently have for Pyright and Mypy, we should detect when a BasedPyright config is present and default to All errors on/map other config options that are compatible.

DetachHead:
# Summary

Fixes #4382

# Test Plan

- added a new test

- manually tested by running `pyrefly init` on a project with an empty `[tool.basedpyright]` section in  `pyproject.toml`, and confirmed the following `[tool.pyrefly]` section was added:
  ```toml
  [tool.pyrefly]
  python-platform = "all"
  preset = "all"
  infer-with-first-use = false
  ```

### DIFF
--- a/crates/pyrefly_config/src/file_kind.rs
+++ b/crates/pyrefly_config/src/file_kind.rs
@@ -19,17 +19,19 @@ use crate::config::ConfigFile;
 pub enum ConfigFileKind {
     MyPy,
     Pyright,
+    BasedPyright,
     Pyrefly,
     Pyproject,
 }
 
 impl ConfigFileKind {
-    pub fn file_name(&self) -> &str {
+    pub fn file_name(&self) -> Option<&str> {
         match self {
-            Self::MyPy => "mypy.ini",
-            Self::Pyright => "pyrightconfig.json",
-            Self::Pyrefly => "pyrefly.toml",
-            Self::Pyproject => "pyproject.toml",
+            Self::MyPy => Some("mypy.ini"),
+            Self::Pyright => Some("pyrightconfig.json"),
+            Self::BasedPyright => None,
+            Self::Pyrefly => Some("pyrefly.toml"),
+            Self::Pyproject => Some("pyproject.toml"),
         }
     }
 
@@ -43,8 +45,10 @@ impl ConfigFileKind {
 
     pub fn check_for_existing_config(&self, path: &Path) -> anyhow::Result<bool> {
         let file_name = self.file_name();
-        if path.ends_with(file_name) && path.exists() {
-            return Ok(true);
+        if let Some(file_name) = file_name {
+            if path.ends_with(file_name) && path.exists() {
+                return Ok(true);
+            }
         }
         if path.ends_with(ConfigFile::PYPROJECT_FILE_NAME) && path.exists() {
             let raw_pyproject = fs_anyhow::read_to_string(path).with_context(|| {
@@ -56,10 +60,15 @@ impl ConfigFileKind {
             return Ok(raw_pyproject.contains(&self.toml_identifier()));
         }
         if path.is_dir() {
-            let custom_file = self.check_for_existing_config(&path.join(file_name));
+            if let Some(file_name) = file_name {
+                let custom_file = self.check_for_existing_config(&path.join(file_name));
+                if custom_file? {
+                    return Ok(true);
+                }
+            }
             let pyproject =
                 self.check_for_existing_config(&path.join(ConfigFile::PYPROJECT_FILE_NAME));
-            return Ok(custom_file? || pyproject?);
+            return Ok(pyproject?);
         }
         Ok(false)
     }
--- a/crates/pyrefly_config/src/migration/error_codes.rs
+++ b/crates/pyrefly_config/src/migration/error_codes.rs
@@ -12,7 +12,7 @@ use crate::config::ConfigFile;
 use crate::migration::config_option_migrater::ConfigOptionMigrater;
 use crate::migration::mypy::util;
 use crate::migration::mypy::util::MypyErrorConfigFlags;
-use crate::migration::pyright::PyrightConfig;
+use crate::migration::pyright::{PyrightConfig, TypeCheckingMode};
 
 /// Configuration option for error codes
 pub struct ErrorCodes;
@@ -67,7 +67,26 @@ impl ConfigOptionMigrater for ErrorCodes {
         &self,
         pyright_cfg: &PyrightConfig,
         pyrefly_cfg: &mut ConfigFile,
+        basedpyright: bool,
     ) -> anyhow::Result<()> {
+        pyrefly_cfg.preset = match pyright_cfg.type_checking_mode {
+            // TODO: "recommended" in basedpyright does enable all rules, but sets the severity to warning
+            // and turns on failOnWarnings. we could do the same here, but for now we just treat both "All"
+            // and "recommended" the same way.
+            Some(TypeCheckingMode::All) | Some(TypeCheckingMode::Recommended) | None => {
+                if basedpyright {
+                    // should also be treated the same as
+                    Some(Preset::All)
+                } else {
+                    None
+                }
+            }
+            Some(TypeCheckingMode::Off) => Some(Preset::Off),
+            // we intentionally don't match other typeCheckingModes such as "strict", because pyright's
+            // isn't necessarily the same as pyrefly's "Strict" preset
+            Some(_) => None,
+        };
+
         // In pyright, error settings are specified in various "report*" fields
         // The PyrightConfig struct already has a method to convert these to an ErrorDisplayConfig
         let error_config = pyright_cfg
@@ -253,7 +272,7 @@ mod tests {
         let mut pyrefly_cfg = ConfigFile::default();
 
         let error_codes = ErrorCodes;
-        let result = error_codes.migrate_from_pyright(&pyright_cfg, &mut pyrefly_cfg);
+        let result = error_codes.migrate_from_pyright(&pyright_cfg, &mut pyrefly_cfg, false);
 
         assert!(result.is_ok());
         assert!(pyrefly_cfg.root.errors.is_some());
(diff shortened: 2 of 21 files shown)


crates/pyrefly_config/src/migration/error_codes.rs
```diff
@@ -67,7 +68,12 @@ impl ConfigOptionMigrater for ErrorCodes {
         &self,
         pyright_cfg: &PyrightConfig,
         pyrefly_cfg: &mut ConfigFile,
+        basedpyright: bool
     ) -> anyhow::Result<()> {
+        if basedpyright {
+            pyrefly_cfg.preset = Some(Preset::All)
```
connernilsen: Talked directly with @DetachHead about whether we truly want `All` here. BasedPyright defaults to all unless a specific preset is set, so that mostly matches behavior here.

As for the presets, up to DetachHead about whether we want to support recommended (all with some error codes downgraded to warning, but still return nonzero exit code) or not

crates/pyrefly_config/src/migration/pyright.rs
```diff
@@ -83,8 +83,9 @@ impl PyrightConfig {
         Ok(serde_jsonrc::from_str::<Self>(text)?)
     }
 
-    pub fn convert(self) -> ConfigFile {
+    pub fn convert(self, basedpyright: bool) -> ConfigFile {
         let mut cfg = ConfigFile::default();
+        dbg!(basedpyright);
```
connernilsen: Please remove when you're done with this :)

crates/pyrefly_config/src/migration/pyright.rs
```diff
@@ -582,9 +584,13 @@ pub fn parse_pyproject_toml(raw_file: &str) -> anyhow::Result<ConfigFile> {
...
     toml::from_str::<PyProject>(raw_file)?
         .tool
-        .and_then(|tool| tool.pyright)
+        .map(|tool| (tool.pyright, tool.basedpyright))
         .ok_or(anyhow::anyhow!(PyrightNotFoundError {}))
-        .map(PyrightConfig::convert)
+        .map(|tool| match tool {
+            (Some(pyright), _) => PyrightConfig::convert(pyright, false),
+            (_, Some(basedpyright)) => PyrightConfig::convert(basedpyright, true),
+            (None, None) => todo!("how do i return a PyrightNotFoundError here????"),
```
connernilsen: I think here you may want to instead do:

```
let tool = toml::from_str(...)?
.tool
.ok_or(anyhow::anyhow!(PyrightNotFoundError {}))?;

// Can also do this by chaining on the previous expression if you remove the trailing `?;` and using `.and_then(...)` if you prefer
let config = match tool {
  Tool { pyright, .. } => {
    ...
  },
  Tool { basedpyright, .. } => {
    ...
  },
  _ => anyhow::anyhow!(PyrightNotFoundError {}),
}
```

crates/pyrefly_config/src/migration/python_platform.rs
```diff
@@ -18,29 +18,32 @@ pub struct PythonPlatformConfig;
...
         // In pyright, python platform is specified in the "pythonPlatform" field
         let platform = match &pyright_cfg.python_platform {
             Some(p) => p,
             None => {
-                return Err(anyhow::anyhow!(
-                    "No python_platform found in pyright config"
-                ));
+                if basedpyright {
+                    "all"
+                } else {
```
connernilsen, line 44: Can you update [configuration.mdx](https://pyrefly.org/en/docs/configuration/) and the similar [migrate from Pyright](https://pyrefly.org/en/docs/migrating-from-pyright/) (or a separate migrate from BasedPyright) docs to reflect that:
1. we handle migrations from BasedPyright now
2. no `python_platform` defaults to `all` for BasedPyright migration
3. no preset defaults to `all` for BasedPyright migrations

crates/pyrefly_config/src/migration/run.rs
```diff
@@ -124,19 +127,27 @@ pub fn find_and_migrate_in_memory(
...
+                pyright::parse_pyproject_toml(&raw_file).with_context(ctx)?,
+                MigratedFromKind::BasedPyright(MigratedConfigSource::PyprojectToml),
+            ))),
             // No tool sections at all — not a migrate-able config and not a
             // parse error. Treat as "nothing nearby."
-            (false, false) => Ok(None),
+            (false, false, false) => Ok(None),
+            // Both basedpyright and pyright config detected. This is invalid and
+            // is an error in basedpyright so we don't support it.
+            (false, true, true) => Ok(None),
```
connernilsen: In this case, we should probably return some kind of error if possible here to let the user know there's something very wrong. You should be able to find examples of how to return an error from this function up above.

pyrefly/lib/commands/check.rs
```diff
@@ -1257,6 +1019,10 @@ fn write_unconfigured_upsell<W: Write>(
                 MigratedFromKind::Pyright(MigratedConfigSource::PyprojectToml) => {
                     ("`[tool.pyright]` in your `pyproject.toml`", "default")
                 }
+                MigratedFromKind::BasedPyright(MigratedConfigSource::DedicatedFile) => todo!("not supported, how do i error here???"),
```
connernilsen: This should be an unreachable state that happens to be difficult to represent with the current type system. It's not often that I do this, but I think this is a reasonable place to panic with an `unreachable!(<reason>),` statement

pyrefly/lib/lsp/non_wasm/type_error_display_status.rs
```diff
@@ -231,6 +225,12 @@ pub fn derive_v2_response(
                     "Default",
                     "default",
                 ),
+                MigratedFromKind::BasedPyright(MigratedConfigSource::DedicatedFile) => todo!("not supported, how do i error here???"),
```
connernilsen: Similarly, we can probably just panic with unreachable here

connernilsen (commented):
Thanks for contributing this! Here are some answers to the questions I saw + a few other additional things to request

crates/pyrefly_config/src/migration/pyright.rs
DetachHead: oops, i wonder if clippy has a rule to catch this?

crates/pyrefly_config/src/migration/error_codes.rs
DetachHead: as discussed i made both `"recommended"` and `"all"` map to the `"all"` preset, but only if the user hasn't specified a different `typeCheckingMode`. this is consistent with how basedpyright works, except that `"recommended"` sets the diagnostic level for many of the rules to `"warning"` and also enables `failOnWarnings`. added a TODO for that

crates/pyrefly_config/src/migration/pyright.rs
DetachHead: #4821

DetachHead:
Hmm looks like this doesnt actually work at all lol. Will take another look tomorrow

crates/pyrefly_config/src/migration/error_codes.rs
```diff
@@ -67,7 +67,26 @@ impl ConfigOptionMigrater for ErrorCodes {
...
         pyrefly_cfg: &mut ConfigFile,
+        basedpyright: bool,
     ) -> anyhow::Result<()> {
+        pyrefly_cfg.preset = match pyright_cfg.type_checking_mode {
+            // TODO: "recommended" in basedpyright does enable all rules, but sets the severity to warning
+            // and turns on failOnWarnings. we could do the same here, but for now we just treat both "All"
+            // and "recommended" the same way.
+            Some(TypeCheckingMode::All) | Some(TypeCheckingMode::Recommended) | None => {
+                if basedpyright {
+                    // should also be treated the same as
```
connernilsen: nit: unfinished sentence? If you let me know what it is, I can just adjust it internally before landing :)

Or feel free to adjust yourself

connernilsen (approved):
This looks great, thanks for working on this! So glad to see you committing to the repo! Please feel free to claim any other issues you may be working on, or let me know about big issues you run into so I can jump into them

crates/pyrefly_config/src/migration/error_codes.rs
connernilsen: Actually I think I got what it was trying to say.

kinto0 (approved):

crates/pyrefly_config/src/migration/error_codes.rs
DetachHead: honestly i dont remember what i was trying to say here lol. i think it might've been what i ended up writing in the TODO comment above, and forgot to delete the half written comment here

connernilsen: I ended up just dropping in this:

```
// basedpyright defaults to the `recommended` type checking mode rather than
// pyright's `standard`, so an unset `typeCheckingMode` (`None`) should also be
// treated the same as `all`.
```

crates/pyrefly_config/src/migration/pyright.rs
```diff
@@ -719,7 +745,32 @@ executionEnvironments = [
...
+
+    #[test]
+    fn test_convert_from_pyproject_basedpyright() -> anyhow::Result<()> {
+        let src = r#"[tool.basedpyright]
+"#;
+        parse_pyproject_toml(src).map(|result| {
+            assert!(result.preset == Some(Preset::All));
+            ()
+        })
+    }
```
connernilsen, line 762: Tests can return unit, and just panic if there's a problem.

There's also a series of `assert_*` macros you can use for testing.

With both of those, the tests can look like this 
```
#[test] 
fn test_() {
  ...
  let result = parse_pyproject_toml(src).unwrap(); // or .expect("reason") if you want
  assert_eq!(result.preset, Some(Preset::All));
}
```

crates/pyrefly_config/src/file_kind.rs
```diff
@@ -43,8 +45,10 @@ impl ConfigFileKind {
 
     pub fn check_for_existing_config(&self, path: &Path) -> anyhow::Result<bool> {
         let file_name = self.file_name();
-        if path.ends_with(file_name) && path.exists() {
-            return Ok(true);
+        if let Some(file_name) = file_name {
+            if path.ends_with(file_name) && path.exists() {
```
connernilsen, line 49: `if let` can be chained with other if statements (and `if let`s)

```
if let Some(file_name) = file_name
  && path.ends_with(file_name)
  && path.exists() {
```

connernilsen (commented):
I fixed these internally, but a few code style conventions we tend to use.

Going to land soon :)

PR #4732, opened 2026-08-30
Issue #1483 dump config full: Print every resolved setting under dump-config --full

### LINKED ISSUE #1483: Verbose output for `pyrefly dump-config`
We should probably have it where we either (or both):
- dump a lot more by default with `dump-config`
- dump almost everything in a config after configuration with `dump-config` when `--verbose`

ravivats:
# Summary

Print every resolved setting under dump-config --full

Fixes #1483

## Problem
`dump-config` is what the docs point people at to debug configuration, but it reports only the interpreter and the paths imports resolve from. Every other resolved setting is invisible, and a config key Pyrefly did not recognize is dropped silently, which is exactly the case people reach for this command to diagnose.

## Approach
Adds `--full`, which prints the rest of the resolved configuration in the same indented, prose-labelled style the command already uses. Default output is unchanged.

As per the comment on the issue, the file-list output is untouched since `--max-files`already covers it. Settings that already appear under `Resolving imports from:` are deliberately omitted so that `--full` adds information rather than repeating it.

The rendering lives in `pyrefly_config` beside `structured_import_lookup_path`, mirroring that pattern, because several `ConfigBase` settings are crate-private and unreadable from `dump_config.rs`. Error severities are sorted on the way out
because `ErrorDisplayConfig` wraps a `HashMap`.

## On the flag name (--full instead of --verbose as suggested in the issue description)
The issue says "when `--verbose`", but there is already a global `-v/--verbose` on every subcommand (`commands/util.rs`) that raises log verbosity, and clap rejects a duplicate. I used `--full` to keep config output independent of log
noise.

## Testing
- New end-to-end test in `test/config.md`
- `cargo test` (8227 passed), `cargo clippy`, `cargo fmt` all clean
- The existing `dump-config` golden tests in `test/config.md` and
  `test/interpreter.md` pass unchanged, confirming default outp

## AI usage disclosure
I have worked on this issue for things like setting up the `pyrefly` repository as well as understanding the code and current dump-config output etc. using Claude, but I have done partially the implementation by hand, and I have reviewed and verified the implementation, the tests, and edited this description.

### DIFF
--- a/crates/pyrefly_config/src/config.rs
+++ b/crates/pyrefly_config/src/config.rs
@@ -984,6 +984,256 @@ impl ConfigFile {
         result
     }
 
+    /// Render every user-facing setting as `label: value` lines for
+    /// `dump-config --full`, indented relative to the caller's own indentation.
+    ///
+    /// This omits what `dump-config` already prints — the config source, the
+    /// interpreter, the covered files, and everything reported by
+    /// `structured_import_lookup_path` — so that `--full` adds information
+    /// rather than repeating it.
+    pub fn structured_settings(&self) -> Vec<String> {
+        fn value<T: Display>(value: &Option<T>) -> Option<String> {
+            value.as_ref().map(T::to_string)
+        }
+
+        fn path(value: &Option<PathBuf>) -> Option<String> {
+            value.as_ref().map(|p| p.display().to_string())
+        }
+
+        // Render an enum by its config-file spelling (`skip-and-infer-return-any`)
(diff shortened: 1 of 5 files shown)


ravivats:
I will send the updated code for it soon.

rchen152:
> I will send the updated code for it soon.

Sorry, you can ignore that bot message! We're a bit behind on PR reviews right now, but I still intend to review this as soon as I can.

PR #4735, opened 2026-08-30
3388 - Update analysis after uv sync

### LINKED ISSUE #3388: update analysis after uv sync
### Describe the Bug

update the version number in pyproject.toml

uv sync

ideally it should analyze again.

### Sandbox Link

_No response_

_No response_

savagecx:
# Summary

Adds the uv.lock file to the list of watched files so that running a uv sync initiates a file analysis.

Fixes #3388

# Test Plan

Added an additional test to check that uv.lock is added to the watch list.
Ran test.py successfully.


crates/pyrefly_config/src/config.rs
```diff
@@ -845,6 +845,9 @@ impl ConfigFile {
         Self::PYREFLY_HIDDEN_FILE_NAME,
         Self::PYPROJECT_FILE_NAME,
     ];
+
+    pub const UV_LOCK: &str = "uv.lock";
```
connernilsen, line 850: Can we change this to an array slice (`&[]`) and rename it to `ADDITIONAL_CRITICAL_FILES` or something similar? I'm open to other names, but I can see us adding other filetypes here in the future!

connernilsen (changes requested):
Hey @savagecx, thanks for working on this, it looks great! I just have one minor change to request around making the type a list, since I can see us adding more filetypes to watch to that in the future, and it'll make future updates a bit easier for us.

Once you're ready for another review, don't forget to request me up under the reviewers list! It's not too big of a change, so I'll probably pull it in and get it merged pretty quickly

crates/pyrefly_config/src/config.rs
savagecx: Great idea @connernilsen! I have added an array slice, updated the insertion mechanism and expanded the tests!

kinto0 (approved):

PR #4750, opened 2026-08-31
fix Final exhaustive enum identity comparison is incorrectly reported as always false #4748

### LINKED ISSUE #4748: Final exhaustive enum identity comparison is incorrectly reported as always false
## Summary

When an `if`/`elif` chain checks every member of `inspect._ParameterKind`, Pyrefly reports the final comparison with `VAR_KEYWORD` as always false. After the preceding members have been excluded, that comparison is always true.

## Reproduction

[Sandbox Link](https://pyrefly.org/sandbox/?project=v2.nZVNC4JAEIb_ypIHu-ReOnkT6iCJikUhCCIpKoYru1b075vVVTE_yo67zBcz7zMztTIFeSCRAmpspFkEFEYEC8PP0jxct0-1MVTs5qsaPSiqk2drrXBnrtOBk2JbR_2kW6Zm-JZpuEM1equCsLTeshuS3168Ga2ml6Zw_MPevVjO7lsiyufwJDRcnO6sOX6XciwPoM8WhxV1T3ZJlPtfi3jNM43JQIRdye03KZOICm1-kCEhu_5AQmRwEwCp-A6XBzqseLkERydCSVkWTMW4cSc0xlGOQ3JluOeBEV8CwSNIq3vWD4dIRTKDsONIbn9CUq6JlOeAlDmPYDHL2Bs)

```python
import inspect

def parameter_kind(parameter: inspect.Parameter) -> str:
    if parameter.kind is inspect.Parameter.POSITIONAL_ONLY:
        return "positional-only"
    elif parameter.kind is inspect.Parameter.POSITIONAL_OR_KEYWORD:
        return "positional-or-keyword"
    elif parameter.kind is inspect.Parameter.VAR_POSITIONAL:
        return "args"
    elif parameter.kind is inspect.Parameter.KEYWORD_ONLY:
        return "keyword-only"
    elif parameter.kind is inspect.Parameter.VAR_KEYWORD:
        return "kwargs"
    return "other"
```

With Pyrefly 1.2.0 and `unnecessary-comparison` enabled, `pyrefly check repro.py` reports:

```text
WARN sandbox.py:12:28-57: Identity comparison `_ParameterKind.VAR_KEYWORD is _ParameterKind.VAR_KEYWORD` is always False [unnecessary-comparison]
```

## Expected behavior

The comparison must not be described as always false. Given the enum's five members and the four preceding branches, it is always true when reached. If Pyrefly reports an unnecessary comparison, the message should say that it is always true; otherwise, it should emit no diagnostic.

The current message is also internally contradictory because it prints the comparison as `_ParameterKind.VAR_KEYWORD is _ParameterKind.VAR_KEYWORD`.

## Environment

- Pyrefly: 1.2.0
- Python target: 3.14
- `unnecessary-comparison` severity: warning

asukaminato0721:
# Summary

Fixes #4748

Identity comparisons now ignore explicit/implicit literal style metadata and correctly recognize equal enum members as singletons

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/operators.rs
+++ b/pyrefly/lib/alt/operators.rs
@@ -1069,7 +1069,8 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
         match (left, right) {
             // If both are literals/None, check for predictable results
             (Type::Literal(l1), Type::Literal(l2)) => {
-                if l1 != l2 {
+                // Explicit/implicit literal style is typing metadata, not runtime identity.
+                if l1.value != l2.value {
                     emit_literal_warning(
                         &l1.value.to_string(),
                         &l2.value.to_string(),
--- a/pyrefly/lib/test/unnecessary_comparison.rs
+++ b/pyrefly/lib/test/unnecessary_comparison.rs
@@ -43,6 +43,26 @@ if False is not False:  # E: Identity comparison `False is not False` is always 
 );
 
 testcase!(
+    test_exhaustive_enum_narrowing,
+    r#"
+import inspect
+
+def parameter_kind(parameter: inspect.Parameter) -> str:
+    if parameter.kind is inspect.Parameter.POSITIONAL_ONLY:
+        return "positional-only"
+    elif parameter.kind is inspect.Parameter.POSITIONAL_OR_KEYWORD:
+        return "positional-or-keyword"
+    elif parameter.kind is inspect.Parameter.VAR_POSITIONAL:
+        return "args"
+    elif parameter.kind is inspect.Parameter.KEYWORD_ONLY:
+        return "keyword-only"
+    elif parameter.kind is inspect.Parameter.VAR_KEYWORD:
+        return "kwargs"
+    return "other"
+"#,
+);
+
+testcase!(
     test_non_singleton_same_value_ok,
     r#"
 # Same integer/string literals - no warning (interning varies)


h3har:
Does this address https://github.com/facebook/pyrefly/issues/4748#issuecomment-5473989525 or should I open a separate issue for that?

asukaminato0721:
> Does this address [#4748 (comment)](https://github.com/facebook/pyrefly/issues/4748#issuecomment-5473989525) or should I open a separate issue for that?

I think a separate issue is ok.

h3har:
> I think a separate issue is ok.

Done: https://github.com/facebook/pyrefly/issues/4751

NathanTempest (changes requested):
Thanks as always, @asukaminato0721. Good catch that literal style is typing bookkeeping and should not affect runtime identity.

The primer results separate into seven corrected diagnostics and 42 entirely new diagnostics. Comparing l1.value produces the seven corrections. Adding Lit::Enum(_) to is_singleton_literal produces the 42 new warnings, since "always True" was previously unreachable for enum literals.

Could we keep this PR focused on the l1.value fix and move enum-tautology detection to a follow-up? That alone resolves #4748 through the reporter's accepted "emit no diagnostic" outcome. The 42 additions occur in idiomatic exhaustive if/elif chains, where retaining the final explicit condition can be useful for safely handling future enum members.

Two related follow-ups:

- Diagnostics render the narrowed operand types rather than the source expressions, producing messages such as Scope.Session is Scope.Session when the source says scope is Scope.Session. The corrected "always True" result is no longer contradictory, but the rendering may still be confusing. This is the second point in #4748's body, distinct from #4751, and isn't tracked anywhere yet, so closing #4748 would drop it.

- Duplicate-value enum aliases need coverage. CRIMSON = 1 alongside RED = 1 makes Color.CRIMSON is Color.RED true at runtime; Python collapses both to a single member named RED, with CRIMSON absent from the member list entirely. If Lit::Enum retains CRIMSON as a distinct member name, the not-equal branch will report "always False" for a comparison that is always true.

My suggested scope here is: retain the .value comparison, keep enum tautology warnings out of this fix, and track the broader enum diagnostic behavior separately.

rchen152 (approved):

PR #4755, opened 2026-08-31
Cut over to all-ruff formatting for code and imports (#4755)

stroxler:
Summary:

**High-level**

I enabled ruff formatting for open-source to try to reduce the number of failed PR imports on tensor-shapes, but I didn't realize that our internal linter configures both a "formatter" and a "sorter" (for import sorting).

I don't see any advantage to making contributors use `usort` (which is slower and requires an extra tool) so I'm cutting sorting over to ruff and making the OSS flow also sort (which is available in ruff but is a separate command from `ruff format`.

Vendored python files are excluded in both configs, I was surprised to see
that wasn't already the case. I don't think we should be reformatting vendored
stubs.

**Details**

Pyrefly was using ruff for code formatting internally (formatter=ruff-api in pyfmt config) but usort for imports, while OSS ruff.toml was formatter-only. This caused internal BLACK lint errors that were actually ruff vs usort import sorting mismatches, and OSS CI only checked tensor-shapes formatting.

This diff cuts over both sides to all-ruff. Now OSS and internal formatting should be byte-identical.

Reviewed By: kinto0

Differential Revision: D118137467

### DIFF
--- a/.github/workflows/pyrefly.yml
+++ b/.github/workflows/pyrefly.yml
@@ -41,10 +41,10 @@ jobs:
       with:
         enable-cache: true
       if: ${{ matrix.os == 'ubuntu-latest' }}
-    - name: Check tensor-shape formatting
-      run: >
-        uv tool run --from ruff==0.16.5 ruff format --check
-        tensor-shapes
+    - name: Check Python formatting (ruff)
+      run: |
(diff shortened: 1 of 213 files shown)


kinto0 (approved):

stroxler:
This landed, not sure why the PR is still open

PR #4757, opened 2026-08-31
[tensor-shapes] add stubs for jax.numpy array creation routines

jakevdp:
# Summary

Update `tensor-shapes/pyrefly-jax-stubs` with shape annotations for `jax.numpy` array creation APIs. Followup to #4708

# Test Plan

Unit test file added
```
$ uv tool run --from ruff==0.16.5 ruff format tensor-shapes
256 files left unchanged
Running formatting...
Finished in 2.37 seconds.
Running linting...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.63s
Finished in 0.83 seconds.
    Finished `release` profile [optimized + debuginfo] target(s) in 0.49s
PASS stubs (6 files)
PASS arithmetic (1 files)
PASS creation (1 files)
PASS fft (1 files)
PASS matmul (1 files)
PASS nn (1 files)
PASS reductions (1 files)
PASS reshape (1 files)
```

# AI usage

Changes in this PR were generated with a Gemini coding agent.

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
+++ b/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
@@ -3,10 +3,11 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, overload, Sequence
+from typing import Any, Callable, Literal, overload, Sequence
 
 from jax._array import Array as Array, Array as ndarray
 from jax._shapes import (
+    int_min,
     matmul_shape,
     permute_shape,
     reduce_shape,
@@ -32,6 +33,8 @@ type _NewShape = int | tuple[int, ...] | None
 # downstream shaped-array operations without degrading to unknown. The NumPy
 # stubs carry the same limitation.
 @overload
+def zeros(shape: tuple[()], dtype: Any = ..., *, device: Any = ...) -> Array[[]]: ...
+@overload
 def zeros[N: IntVar](
     shape: Int[N], dtype: Any = ..., *, device: Any = ...
 ) -> Array[[N]]: ...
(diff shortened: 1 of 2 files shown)


tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -85,50 +142,274 @@ def full[N: IntVar, M: IntVar](
 ) -> Array[[N, M]]: ...
 @overload
 def full[N: IntVar, M: IntVar, K: IntVar](
-    shape: IntTuple[N, M, K], fill_value: Any, dtype: Any = ..., *, device: Any = ...
+    shape: IntTuple[N, M, K], dtype: Any = ..., *, device: Any = ...
```
stroxler: I think we dropped fill_value here by accident?

stroxler (changes requested):
I think my initial bootstrap of jax may have already had a bad pattern, but it seems like we're making it worse here: as far as I can tell, the majority of these stubs really could be written without overloads.

I think we ought to be able to write, for example,
```
def full[Shape: IntTuple](
   shape: Shape, fill_value: Any, dtype: Any = ..., *, device: Any = ...
) -> Array[Shape]: ...
```
which would be both much simpler and more performant at type checking time than overloads.

Maybe I missed something, but I think it's worth a look before we expand the number of overloads so much

jakevdp:
> I think we ought to be able to write, for example,
> 
> ```
> def full[Shape: IntTuple](
>    shape: Shape, fill_value: Any, dtype: Any = ..., *, device: Any = ...
> ) -> Array[Shape]: ...
> ```

This works, but we also need to handle the case where `shape` is a scalar integer (that kind of polymorphism was inherited from NumPy)

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
jakevdp: Fixed!

jakevdp:
Updated – PTAL

stroxler (approved):

PR #4759, opened 2026-09-01
fix add recommend stub #4753

### LINKED ISSUE #4753: SciPy FFT functions are inferred as tuple[Dispatchable]
## Describe the Bug

`scipy.fft.rfftn` and `scipy.fft.irfftn` are inferred as `tuple[Dispatchable]`, although they return NumPy arrays at runtime. This produces false-positive NumPy overload, indexing, and mutation errors in otherwise valid code.

Minimal reproduction (inside a configured project that resolves SciPy):

```python
import numpy as np
from scipy import fft


def transform(img: np.ndarray, spikes: np.ndarray) -> np.ndarray:
    axes = tuple(range(spikes.shape[-1]))
    spectrum = fft.rfftn(np.ascontiguousarray(img), axes=axes, workers=1)
    reconstructed = fft.irfftn(
        spectrum,
        s=tuple(img.shape[axis] for axis in axes),
        axes=axes,
        workers=1,
    )
    return np.clip(reconstructed, 0.0, 1.0)
```

Expected: `spectrum` and `reconstructed` are inferred as `ndarray`, and the function type-checks.

Actual:

```text
revealed type: tuple[Dispatchable]
No matching overload found for numpy.clip called with arguments: (tuple[Dispatchable], float, float)
```

The same `tuple[Dispatchable]` type also makes valid spectrum indexing and in-place addition fail.

Environment:

- Pyrefly 1.2.0
- Python 3.10.16
- SciPy 1.15.3
- NumPy 2.2.6
- macOS 26.4.1 arm64

The relevant SciPy API returns an ndarray at runtime. A local `cast("np.ndarray", ...)` is sufficient as a workaround, but loses the checker’s actual return-type information.

## Sandbox Link

Not available: the repro requires the SciPy package and its resolved stubs in a configured local project.

## (Only applicable for extension issues) IDE Information

N/A

asukaminato0721:
# Summary

Fixes #4753

Added scipy -> scipy-stubs recommendation

# Test Plan

add test


stroxler (approved):

PR #4770, opened 2026-09-01
Model matmul with gufunc shapes

stroxler:
Summary: Express JAX `matmul` through gufunc signatures for its four vector and matrix rank combinations. This preserves existing vector behavior while adding precise leading-dimension broadcasting and core-dimension validation for batched calls.

Differential Revision: D118226441

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
+++ b/tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
@@ -4,7 +4,12 @@
 # LICENSE file in the root directory of this source tree.
 
 import shape_extensions.dsl as dsl
-from shape_extensions import Int, IntTuple, type_shape_dsl_function
+from shape_extensions import (
+    gufunc_broadcast,
+    Int,
+    IntTuple,
+    type_shape_dsl_function,
+)
 
 @type_shape_dsl_function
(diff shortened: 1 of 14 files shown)


tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
```diff
@@ -18,75 +23,20 @@ def int_min(a: Int, b: Int) -> Int:
...
 def matmul_shape(left: IntTuple, right: IntTuple) -> IntTuple:
-    # 1. Reject operands with rank 0 (scalars)
     if len(left) == 0 or len(right) == 0:
-        return dsl.Invalid("matmul requires at least 1-D operands")
-
-    # (k)(*batch,k,m) -> (*batch,m)
+        return dsl.Invalid("matmul expects at least 1-D arrays")
+    operands = dsl.IntTuples((left, right))
+    if len(left) == 1 and len(right) == 1:
+        spec = "(n),(n)->()"
```
jakevdp, line 30: I think we could generalize this to
```
if len(right) == 1:
  spec = "(n),(n)->()"
```
and then skip the `(m,n),(n)->(m)` case below, which should be covered by broadcasted batch dimensions in the above.

kinto0 (approved):

stroxler:
This landed internally, closing the PR

PR #4772, opened 2026-09-01
Treat builtin type constructor constants as implicitly typed in `pyrefly coverage`

jorenham:
# Summary

Before, `pyrefly coverage check` would report stuff like `SPAM = 1` as implicitly typed, but `MISSING = object()` as untyped, because it only exempted literal values. 

This also exempts the builtin type constructors lik `object()` and `frozenset()` as implicitly typed, so that `pyrefly coverage` will no longer require those to be annotated.

# Test Plan

Tests added.


stroxler (approved):

PR #4773, opened 2026-09-01
Fix Ruff import formatting in shape DSL

stroxler:
Summary: Normalize the whitespace after the shape DSL imports so the pinned GitHub Ruff import-order check passes.

Differential Revision: D118307281


rchen152 (approved):

PR #4776, opened 2026-09-01
[tensor-shapes] add stubs for jax.numpy einsum & tensor contractions

jakevdp:
### Summary

Add tensor shape stubs for `jax.numpy` einsum & other tensor contractions

### Test Plan

Updated existing unit tests, and tests pass locally:
```
328 files already formatted
All checks passed!
Running formatting...
Finished in 2.69 seconds.
Running linting...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.91s
Finished in 1.15 seconds.
    Finished `release` profile [optimized + debuginfo] target(s) in 0.61s
PASS stubs (9 files)
PASS arithmetic (1 files)
PASS contractions (1 files)
PASS creation (1 files)
PASS fft (1 files)
PASS lax (1 files)
PASS linalg (1 files)
PASS matmul (1 files)
PASS nn (1 files)
PASS reductions (1 files)
PASS reshape (1 files)
```

### AI disclosure

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
+++ b/tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
@@ -447,3 +447,182 @@ def tridiagonal_diag_minus_one_shape(a_shape: IntTuple) -> IntTuple:
         return dsl.Invalid("tridiagonal requires a square matrix")
     batch = a_shape[: len(a_shape) - 2]
     return dsl.concat(batch, dsl.IntTuple((n1 - 1,)))
+
+@type_shape_dsl_function
+def einsum_shape(spec: str, shapes: IntTuples) -> IntTuple:
+    return dsl.einsum(spec, shapes)
+
+@type_shape_dsl_function
+def dot_shape(left: IntTuple, right: IntTuple) -> IntTuple:
+    if len(left) == 0:
+        return right
+    if len(right) == 0:
+        return left
+    if len(left) == 1 and len(right) == 1:
+        if left[0] != right[0]:
+            return dsl.Invalid("dot dimensions must match")
+        return dsl.IntTuple(())
+    if len(right) == 1:
+        if left[len(left) - 1] != right[0]:
+            return dsl.Invalid("dot inner dimensions must match")
+        return left[: len(left) - 1]
+    if len(left) == 1:
+        if left[0] != right[len(right) - 2]:
+            return dsl.Invalid("dot inner dimensions must match")
+        return dsl.concat(
+            right[: len(right) - 2], dsl.IntTuple((right[len(right) - 1],))
+        )
+    if left[len(left) - 1] != right[len(right) - 2]:
+        return dsl.Invalid("dot inner dimensions must match")
+    return dsl.concat(
+        left[: len(left) - 1],
+        dsl.concat(right[: len(right) - 2], dsl.IntTuple((right[len(right) - 1],))),
+    )
+
+@type_shape_dsl_function
+def inner_shape(left: IntTuple, right: IntTuple) -> IntTuple:
+    if len(left) == 0:
+        return right
+    if len(right) == 0:
+        return left
+    if left[len(left) - 1] != right[len(right) - 1]:
+        return dsl.Invalid("inner dimensions must match")
+    if len(left) == 1 and len(right) == 1:
+        return dsl.IntTuple(())
+    return dsl.concat(left[: len(left) - 1], right[: len(right) - 1])
+
+@type_shape_dsl_function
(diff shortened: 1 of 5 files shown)


tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -886,9 +904,167 @@ def logaddexp2[Shape: _Shape](
...
 ) -> Array[broadcast(Shape1, Shape2)]: ...
+def cross[Shape1: _Shape, Shape2: _Shape](
+    a: Array[Shape1],
+    b: Array[Shape2],
+    /,
+    axisa: int = -1,
+    axisb: int = -1,
+    axisc: int = -1,
+    axis: int | None = None,
+) -> Array[broadcast(Shape1, Shape2)]: ...
```
stroxler: How correct is this? An agent thinks this is an approximation, but when `axis` arguments remap dimensions or there's a 2D vs 3D cross broadcast might not be right.

If that's the case, we might want to make a new DSL function and if necessary put a gradual fallback in place (you can use type vars with an upper bound of `Flag[int]` to capture arguments and pass them to the DSL, if you haven't yet used that, and the DSL allows you to explicitly return gradual shapes and sizes)

If it turns out to be hard to do well, I think a TODO is also fine (and if we discover the DSL is missing expressive power, happy to try to work on extending it! It's normal that we hit limitations in the DSL as we write more stubs)

tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
```diff
@@ -494,3 +494,182 @@ def tridiagonal_diag_minus_one_shape(a_shape: IntTuple) -> IntTuple:
...
+    if d0 == d1:
+        return dsl.concat(shape[2:], dsl.IntTuple((d0,)))
+    if dsl.is_concrete_int(d0) and dsl.is_concrete_int(d1):
+        if d0 < d1:
+            return dsl.concat(shape[2:], dsl.IntTuple((d0,)))
+        return dsl.concat(shape[2:], dsl.IntTuple((d1,)))
+    return dsl.concat(shape[2:], dsl.IntTuple((dsl.Int.gradual(),)))
+
+@type_shape_dsl_function
+def trace_shape(shape: IntTuple, offset: int, axis1: int, axis2: int) -> IntTuple:
```
stroxler, line 607: Same note as `diagonal_shape`, the extra args here probably impact the output shape but we don't seem to be using them.

tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
```diff
@@ -494,3 +494,182 @@ def tridiagonal_diag_minus_one_shape(a_shape: IntTuple) -> IntTuple:
...
+    if dims < 0:
+        return dsl.Invalid("tensordot dims must be non-negative")
+    if dims > len(left) or dims > len(right):
+        return dsl.Invalid("tensordot dims exceeds input rank")
+    if any(left[len(left) - dims + i] != right[i] for i in range(dims)):
+        return dsl.Invalid("tensordot contracted dimensions must match")
+    return dsl.concat(left[: len(left) - dims], right[dims:])
+
+@type_shape_dsl_function
+def diagonal_shape(shape: IntTuple, offset: int, axis1: int, axis2: int) -> IntTuple:
```
stroxler, line 540: An agent thinks this is probably approximate: offset and axis are ignored here, but they play a role in the correct result.

Maybe we could fall back to gradual shapes if it's hard to get the right thing, or at least put a TODO in to come back to this?

stroxler (changes requested):
Overall it looks good to me.

Back to you for a few cases where I think the stubs are lossy without being gradual, I think we should try to improve them and/or throw some TODOs in

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
jakevdp: Updated!

tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
jakevdp: Updated!

tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
jakevdp: Updated!

stroxler (approved):

PR #4780, opened 2026-09-02
fix Go to Definition resolves Flytekit @task calls to Task.__call__ instead of the decorated function #4778

### LINKED ISSUE #4778: `Go to Definition` resolves Flytekit `@task` calls to `Task.__call__` instead of the decorated function
### Describe the Bug

Pyrefly's VS Code extension resolves a call to a function decorated with Flytekit's `@task` decorator to Flytekit's `Task.__call__` implementation instead of the original function definition.

This appears related to [pyrefly#3356](https://github.com/facebook/pyrefly/issues/3356), but it occurs with Flytekit's decorator rather than `functools.cache`.

In particular, [this comment on pyrefly#3356](https://github.com/facebook/pyrefly/issues/3356#issuecomment-4525784836) reports the same behavior with Django and `django-stubs`, where Pyright navigates to the Django definition but Pyrefly navigates to the stub definition. 

I also saw a similar issue reported here that was closed without any notes: https://github.com/facebook/pyrefly/issues/3363

These broader reports suggest this may be a general go-to-definition issue for decorated callables or third-party stub-backed symbols.

In a private project, the task is defined in one module and imported into a Flyte workflow in another module.

When I invoke `Go to Definition` on the imported task name at the call site, Pyrefly opens `flytekit/core/base_task.py` at the `Task.__call__` method.

The expected destination is the `def foo(...)` definition in the user module.

### Minimal public Flytekit reproduction

The reproduction directory contains these files.

```text
pyrefly_flytekit_repro/
├── pyproject.toml
└── src/
    └── pyrefly_flytekit_repro/
        ├── __init__.py
        └── bar.py
```

`pyproject.toml`:

```toml
[project]
name = "pyrefly-flytekit-repro"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = ["flytekit==1.16.23"]
dev_dependencies = ["pyrefly==1.2.0"]

[build-system]
requires = ["uv_build>=0.12.7,<0.13.0"]
build-backend = "uv_build"
```

`src/pyrefly_flytekit_repro/bar.py`:

```python
from flytekit.core.task import task


@task(retries=2)
def foo(value: int) -> int:
    return value + 1
```

`src/pyrefly_flytekit_repro/__init__.py`:

```python
from flytekit.core.workflow import workflow

from .bar import foo


@workflow
def pipeline(value: int) -> int:
    return foo(value=value)
```

From the project directory, run `uv sync` to synchronize the pinned Flytekit and Pyrefly packages.

```console
uv sync
```

Then open the directory in VS Code with the Pyrefly extension installed (version 1.2.0) with all default settings configured.

Place the cursor on `foo` in the call inside `src/pyrefly_flytekit_repro/__init__.py` and invoke `Go to Definition`.

Expected behavior: navigation goes to the `foo` definition in `src/pyrefly_flytekit_repro/bar.py`.

Actual behavior: navigation goes to Flytekit's `Task.__call__` method in `.venv/lib/python3.12/site-packages/flytekit/core/base_task.py`.

Image

Interestingly, Pyrefly's go to definition *does* perform the expected behavior of going to the `foo` definition in `src/pyrefly_flytekit_repro/bar.py` if you instead perform it on the import of `foo` instead of the call:

Image

### Sandbox Link

N/A

VS Code Version: 1.135.0 (Universal)
Commit: 08d4889f9ec4a1685d257b9b95de036c8e1ce1e5
Date: 2026-08-25T14:26:52Z
Electron: 42.8.1
ElectronBuildId: 14906494
Chromium: 148.0.7778.280
Node.js: 24.18.1
V8: 14.8.178.38-electron.0
@github/copilot: 1.0.81-0
@github/copilot-sdk: 1.0.11
OS: Darwin arm64 25.6.0

asukaminato0721:
# Summary

Fixes #4778

Decorated functions now navigate to their original def, even when their resulting type is a callable class instance.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/lsp.rs
+++ b/pyrefly/lib/state/lsp.rs
@@ -2194,10 +2194,6 @@ impl<'a> Transaction<'a> {
                 );
                 defs
             }
-            // Workaround so functions decorated with `functools.lru_cache` go to definition on the source, not the decorator.
-            Type::ClassType(class) if class.has_qname("functools", "_lru_cache_wrapper") => {
-                vec![]
-            }
             Type::ClassType(_) => self
                 .find_attribute_definition_for_base_type(handle, preference, ty, &dunder::CALL)
                 .map(Vec1::into_vec)
@@ -2811,9 +2807,20 @@ impl<'a> Transaction<'a> {
                         }
                     }
                     ExprContext::Load | ExprContext::Del | ExprContext::Invalid => {
+                        let definition = self.find_definition_for_name_use(handle, &id, preference);
+                        // A decorator may give a function a class instance type. Preserve the
+                        // function definition instead of navigating to the instance's `__call__`.
+                        let is_function_or_method = match &definition {
+                            Ok(Some(item)) => matches!(
+                                item.metadata.symbol_kind(),
+                                Some(SymbolKind::Function | SymbolKind::Method)
+                            ),
+                            Ok(None) | Err(_) => false,
+                        };
                         // If this name is the callee of a call expression, jump
                         // to constructor or __call__ definitions when applicable.
                         if preference.resolve_call_dunders
+                            && !is_function_or_method
                             && let Some(AnyNodeRef::ExprCall(call)) = covering_nodes.get(1)
                             && call.func.range() == id.range
                             && let Some(bindings) = self.get_bindings(handle)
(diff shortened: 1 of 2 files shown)


rchen152 (approved):

PR #4782, opened 2026-09-02
Update overload selection for spec change (take 2) (#4782)

rchen152:
Summary:

embarrassed-bunny 

D118074925 moved too fast and implemented only half of https://github.com/python/typing-council/issues/65. The selected return type needs to be assignable to all candidates *and permit all operations that are supported on any candidate*. Now it should be mostly right.

The one place where I know it's still wrong is due to pyrefly's int/float/complex special case, which we implement by making int assignable to float, and int and float assignable to complex. Because of this, pyrefly will think that between `int | Any` and `float`, it can pick `int | Any`, even though `int` does not permit all operations supported on `float`.

This is preexisting subtyping weirdness (https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex explicitly says "`complex`, `float`, and `int` are not subtypes of each other"!), so overload evaluation is not the right place to fix or special-case it.

Mypy primer error delta vs. my previous fix attempt:
```
Error kind                Added  Removed      Net
----------------------  -------  -------  -------
unknown-argument-type       116      251     -135
unknown-variable-type       104      171      -67
assert-type                  76       60      +16
unused-call-result            2      126     -124
[all other error kinds have <100 added or removed errors apiece]
----------------------  -------  -------  -------
Total                       349      903     -554
```

Differential Revision: D118372348

### DIFF
--- a/pyrefly/lib/alt/overload.rs
+++ b/pyrefly/lib/alt/overload.rs
@@ -834,7 +834,7 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
                 }
             }
             if matched_overloads.len() > 1 {
-                // Step 5, part 1: for each overload, check whether it's the case that all possible
+                // Step 5: for each overload, check whether it's the case that all possible
                 // materializations of each argument are assignable to the corresponding parameter.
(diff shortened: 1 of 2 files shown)


grievejia (approved):

PR #4793, opened 2026-09-02
[tensor-shapes] add stubs for jax.numpy array structure APIs

jakevdp:
### Summary

Type stubs for jax structural ops: concatenate, stack, split, dimension operations, etc.

### Test Plan

Adds a new test file; all tests pass locally:
```
328 files already formatted
All checks passed!
Running formatting...
Finished in 2.68 seconds.
Running linting...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.87s
Finished in 1.08 seconds.
    Finished `release` profile [optimized + debuginfo] target(s) in 0.56s
PASS stubs (9 files)
PASS arithmetic (1 files)
PASS creation (1 files)
PASS fft (1 files)
PASS lax (1 files)
PASS linalg (1 files)
PASS matmul (1 files)
PASS nn (1 files)
PASS reductions (1 files)
PASS reshape (1 files)
PASS structural (1 files)
```

### AI disclosure

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
+++ b/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
@@ -7,23 +7,40 @@ from typing import Any, Callable, Literal, overload, Sequence, Unpack
 
 from jax._array import Array as Array, Array as ndarray
 from jax._shapes import (
+    broadcast_to_shape,
+    column_stack_shape,
+    concatenate_shape,
     cross_axes_shape,
     cross_axis_shape,
     diagonal_shape,
     dot_shape,
+    dstack_shape,
     einsum_shape,
+    expand_dims_shape,
+    flip_shape,
+    fliplr_shape,
+    flipud_shape,
+    hstack_shape,
     inner_shape,
     int_min,
     kron_shape,
     matmul_shape,
     matvec_shape,
+    moveaxis_shape,
     permute_shape,
+    ravel_shape,
     reduce_shape,
     reshape_shape,
     reverse_shape,
+    roll_shape,
+    rollaxis_shape,
+    squeeze_shape,
+    stack_shape,
+    swapaxes_shape,
     tensordot_shape,
     trace_shape,
     vecmat_shape,
+    vstack_shape,
 )
 from shape_extensions import (
     broadcast,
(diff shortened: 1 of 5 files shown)


tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -1056,6 +1094,179 @@ def reshape(
...
+def flip[Shape: _Shape](
+    m: Array[Shape],
+    axis: Sequence[int] | None = None,
+) -> Array[Shape]: ...
+def fliplr[Shape: _Shape](
+    m: Array[Shape],
+) -> Array[fliplr_shape(Shape)]: ...
+def flipud[Shape: _Shape](
+    m: Array[Shape],
+) -> Array[flipud_shape(Shape)]: ...
```
jakevdp: One question on these: for both of these the output shape is equal to the input shape, but there is a lower-bound on the number of dimensions (`fliplr` requires an array with at least one dimension, `flipud` requires at least two).

We could also express this without a DSL function, via something like `Array[[*Elements[Batch], N]]` and `Array[[*Elements[Batch], N, M]]`.

Which approach is preferable?

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -1056,6 +1094,179 @@ def reshape(
...
+@overload
+def roll[Shape: _Shape, Axis: Flag[_Axis] = None](
+    a: Array[Shape],
+    shift: Any,
+    axis: Axis = None,
+) -> Array[roll_shape(Shape, Axis)]: ...
```
jakevdp, line 1460: Notice that for `roll` and `flip`, the output shape always equals the input shape. The only reason to have a DSL function is to validate that `axis` is in-bounds.

Should we be doing this kind of validation, or should we just focus on shape propagation, avoiding the DSL call?

tensor-shapes/test-requirements.txt
```diff
@@ -18,9 +18,9 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 
 numpy==2.5.0
-jax==0.10.2
+jax==0.11.1
 # Pinned alongside `jax`: it is a separate distribution that would otherwise
 # resolve freely under a pinned `jax`.
-jaxlib==0.10.2
+jaxlib==0.11.1
```
jakevdp, line 24: This version bump was required for the fix in https://github.com/jax-ml/jax/pull/39075, which properly supports negative `axis` in `jnp.permute_dims`.

stroxler (approved):

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
stroxler: I'm honestly not sure - I think `Array[[*Elements[Batch], N, M]]` is probably clearer for someone actually reading the types, whereas the dsl function allows cleaner error messages

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
stroxler: I think using the DSL for validation is fine, we already have some similar examples in torch stubs (for example `cosine_similarity_shape` is entirely validation)

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
jakevdp: OK, I'll make that switch.

PR #4794, opened 2026-09-03
fix inconsistent syntax highlighting for field of type inspect._ParameterKind #4751

### LINKED ISSUE #4751: inconsistent syntax highlighting for field of type `inspect._ParameterKind`
This is a dedicated issue for https://github.com/facebook/pyrefly/issues/4748#issuecomment-5473989525 .

## Summary

Pyrefly highlights an attribute of type `inspect._ParameterKind` differently after type narrowing.

## Reproduction

[Sandbox Link](https://pyrefly.org/sandbox/?project=v2.nZVNC4JAEIb_ypIHu-ReOnkT6iCJikUhCCIpKoYru1b075vVVTE_yo67zBcz7zMztTIFeSCRAmpspFkEFEYEC8PP0jxct0-1MVTs5qsaPSiqk2drrXBnrtOBk2JbR_2kW6Zm-JZpuEM1equCsLTeshuS3168Ga2ml6Zw_MPevVjO7lsiyufwJDRcnO6sOX6XciwPoM8WhxV1T3ZJlPtfi3jNM43JQIRdye03KZOICm1-kCEhu_5AQmRwEwCp-A6XBzqseLkERydCSVkWTMW4cSc0xlGOQ3JluOeBEV8CwSNIq3vWD4dIRTKDsONIbn9CUq6JlOeAlDmPYDHL2Bs)

```python
import inspect

def parameter_kind(parameter: inspect.Parameter) -> str:
    if parameter.kind is inspect.Parameter.POSITIONAL_ONLY:
        return "positional-only"
    elif parameter.kind is inspect.Parameter.POSITIONAL_OR_KEYWORD:
        return "positional-or-keyword"
    elif parameter.kind is inspect.Parameter.VAR_POSITIONAL:
        return "args"
    elif parameter.kind is inspect.Parameter.KEYWORD_ONLY:
        return "keyword-only"
    elif parameter.kind is inspect.Parameter.VAR_KEYWORD:
        return "kwargs"
    return "other"
```

With Pyrefly 1.2.0 and `unnecessary-comparison` enabled, the GUI in the Pyrefly sandbox and VSCode show similar behaviours:

Image

Image

## Actual behaviour

In the first branch, `kind` is inferred to have type `inspect._ParameterKind`. In subsequent branches, it's narrowed to have a `Literal` type for a subset of the variants. Pyrefly's LSP seems to assign a different token type to the narrowed field.

## Expected behaviour

Even if we want to highlight `Literal` fields differently, it seems inconsistent to treat the narrowed field differently, since in all branches the field has a fixed number of variants (whether that type is spelled as `Literal` or not). That being said other language servers that I know of would just use the same color for all instance attributes, whether inferred to have a `Literal`-compatible type or not.

## Environment

- Pyrefly: 1.2.0
- Python target: 3.14
- `unnecessary-comparison` severity: warning

asukaminato0721:
# Summary

Fixes #4751

fixes incorrect semantic highlighting for narrowed enum instance attributes.

Previously, holder.kind could be classified as an enumMember after narrowing.

Now instance attributes such as holder.kind remain property, actual enum members accessed through the enum class, such as E.A, remain enumMember.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/state/semantic_tokens.rs
+++ b/pyrefly/lib/state/semantic_tokens.rs
@@ -263,24 +263,28 @@ fn range_overlaps(limit_range: Option<TextRange>, range: TextRange) -> bool {
 /// Classify an attribute's resolved type into a semantic token kind. For a union,
 /// every member must agree on the same kind; any disagreement (or a member that is
 /// a plain attribute) falls back to `PROPERTY`.
-fn attribute_semantic_token_type(ty: Type) -> SemanticTokenType {
+fn attribute_semantic_token_type(ty: Type, accessed_on_class: bool) -> SemanticTokenType {
     match ty {
         Type::Union(union) => {
             let mut members = union.members.into_iter();
             let Some(first) = members.next() else {
                 return SemanticTokenType::PROPERTY;
             };
-            let kind = attribute_semantic_token_type(first);
+            let kind = attribute_semantic_token_type(first, accessed_on_class);
             if kind == SemanticTokenType::PROPERTY {
                 return SemanticTokenType::PROPERTY;
             }
-            if members.all(|member| attribute_semantic_token_type(member) == kind) {
+            if members
+                .all(|member| attribute_semantic_token_type(member, accessed_on_class) == kind)
+            {
                 kind
             } else {
                 SemanticTokenType::PROPERTY
             }
         }
-        Type::Literal(lit) if matches!(lit.value, Lit::Enum(_)) => SemanticTokenType::ENUM_MEMBER,
+        Type::Literal(lit) if accessed_on_class && matches!(lit.value, Lit::Enum(_)) => {
+            SemanticTokenType::ENUM_MEMBER
+        }
         _ => {
             attribute_symbol_kind_from_type(&ty)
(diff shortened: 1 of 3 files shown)


pyrefly/lib/state/semantic_tokens.rs
```diff
@@ -263,24 +263,28 @@ fn range_overlaps(limit_range: Option<TextRange>, range: TextRange) -> bool {
 /// Classify an attribute's resolved type into a semantic token kind. For a union,
 /// every member must agree on the same kind; any disagreement (or a member that is
 /// a plain attribute) falls back to `PROPERTY`.
-fn attribute_semantic_token_type(ty: Type) -> SemanticTokenType {
+fn attribute_semantic_token_type(ty: Type, accessed_on_class: bool) -> SemanticTokenType {
```
kinto0: isn't every attribute "accessed on class"?

kinto0 (changes requested):
I don't love this special-case. until now, all attributes could be colored based on only their type. is this the only exception? Is it true that accessed_on_class + being an enum are the only qualifiers?

PR #4795, opened 2026-09-03
fix support type check for patch assert_called_with #3913

### LINKED ISSUE #3913: support type check for patch assert_called_with
### Describe the Bug

it use Any as type, but if we can analysis the patched string, we may able to do some extra type check.

Though a bit ad-hoc.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #3913

patch("module.callable") now retains the target signature through context managers and .start().

`assert_called_with`, `assert_called_once_with`, `assert_any_call`, and async equivalents validate arguments.

Handles sync, async, and bound methods.

new_callable= retains its declared replacement type.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/call.rs
+++ b/pyrefly/lib/alt/call.rs
@@ -2414,7 +2414,8 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
                 )
         } else {
             self.expand_mut(&mut callee_ty);
-            self.check_unittest_mock_patch_target(&callee_ty, &x.arguments, errors);
+            let unittest_mock_patch_target =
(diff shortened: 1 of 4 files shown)


pyrefly/lib/alt/call.rs
```diff
@@ -2425,6 +2426,16 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
                     );
                 }
                 match ty.callee_kind() {
+                _ if ty.toplevel_func_metadata().is_some_and(|metadata| metadata.flags.unittest_mock_type.is_some()) =>
```
kinto0, line 2450: none of the added tests cover this. could you add one?

kinto0 (changes requested):

PR #4797, opened 2026-09-03
Suppress unknown-name for global declarations the function itself defines

### LINKED ISSUE #4785: False positive unknown-name for global assign w/o declaration
### Describe the Bug

```python
def set_workload_id(value: str) -> None:
    global workload_id
    workload_id = value


set_workload_id("test")
assert globals()["workload_id"] == "test"

```

Python: 3.12

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZTBDsIgDIZfpdkO6MXefQrvYhZ0BBcXMAP19W3pljASd_HatH8o39__V2QywmhT9wnTg2OmG_rFj2yTiqsbw9WMUDRLvSiUl6J9La3ze3VD4CTrZsm4259p6UK3uYg_pD9_RIWhhZMUYN6IAoj4uRfFHHE5aN9Swlm4p_SMR8RlPEwOrcc-3CKuJhDYceZthhyeazkI2TaRZP_hrwS_2qKvGD51bAL9Ag

_No response_

nitishagar:
Fixes #4785

## Problem

A `global` declaration inside a function errored with `unknown-name` ("Could not find name `x`") whenever the module had no module-level definition of `x` — even when the declaring function assigns `x` itself:

```python
def set_workload_id(value: str) -> None:
    global workload_id  # ERROR: Could not find name `workload_id`
    workload_id = value
```

`global x; x = value` is a legal way to create the module-level name at runtime, so the declaration should not error.

## Why it happened

The mutable-capture lookup for `global` names consults module-scope statics (plus a builtin fallback) only, and a function's global-assign intentionally never publishes a module-scope static — so the lookup cannot find the very name the function is about to define.

## Change

The definitions phase already merges every definition site in a scope, so it knows whether the scope gives the name a value beyond the capture declaration:

- `Definition` records whether any definition site is a value definition (assignment or import; `del` and other capture declarations do not count).
- The flag is threaded through the `MutableCapture` static.
- `declare_mutable_capture` suppresses only the not-found error for `global` declarations in function/method scopes whose own body defines the name.

Everything else is unchanged: undefined reads (`global a; print(a)` with no definition anywhere), assignments textually before the declaration (`x = 1; global x`), module top-level and class-body declarations, and all `nonlocal` behavior keep erroring exactly as before.

## Test plan

- [x] New `testcase!`s in `pyrefly/lib/test/scope.rs`: the issue repro, undefined-read still errors, nested-if assignments (both orders), `del` doesn't define, assignment-before-declaration keeps its error, module top-level and class-body keep erroring
- [x] `cargo test -p pyrefly --lib` green (the two pre-existing `missing-source` LSP interaction failures also fail on unmodified `main`)
- [x] Formatting/linting clean

### DIFF
--- a/pyrefly/lib/binding/bindings.rs
+++ b/pyrefly/lib/binding/bindings.rs
@@ -92,6 +92,7 @@ use crate::binding::pytest::PytestBindingInfo;
 use crate::binding::pytest::is_pytest_fixture_function;
 use crate::binding::scope::Exportable;
 use crate::binding::scope::FlowStyle;
+use crate::binding::scope::MutableCaptureError;
 use crate::binding::scope::NameReadInfo;
 use crate::binding::scope::ScopeTrace;
 use crate::binding::scope::Scopes;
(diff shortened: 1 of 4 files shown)


kinto0 (approved):

PR #4798, opened 2026-09-03
Exempt base-dictated Any positions in @override methods from explicit-any

### LINKED ISSUE #4548: `explicit-any` is reported for `@override` methods that must match a base-class signature annotated with `Any`
# Describe the Bug

When a method is decorated with `@typing.override`, its signature is dictated by the base class. If the base class annotates a parameter or the return as `Any`, the override has to repeat that `Any` — narrowing it would change the signature, and removing it is not possible. `explicit-any` currently fires on those annotations anyway, so the diagnostic points at code that cannot be changed.

## Repro

`repro.py`:

```python
import json
from typing import Any
from typing import override


class CustomEncoder(json.JSONEncoder):
    @override
    def default(self, o: Any) -> Any:
        return super().default(o)
```

`pyrefly.toml`:

```toml
[errors]
explicit-any = "error"
```

Actual (pyrefly 1.2.0):

```
ERROR Explicit `Any` is not allowed [explicit-any]
 --> repro.py:8:26
ERROR Explicit `Any` is not allowed [explicit-any]
 --> repro.py:8:34
 INFO 2 errors
```

Expected: no errors. Both annotations are fixed by typeshed's `json.JSONEncoder.default`, which is declared as `def default(self, o: Any) -> Any`.

## Why this seems worth special-casing

Ruff reached the same conclusion for its equivalent lint, ANN401. It exempts `@override` methods, implemented in [astral-sh/ruff#4409](https://github.com/astral-sh/ruff/pull/4409) closing [astral-sh/ruff#4408](https://github.com/astral-sh/ruff/issues/4408). The reasoning given there:

> Where that discussion landed was that it is reasonable to suppress `ARG002` for overridden methods, since you cannot control the method signature. I think it follows that `ANN401` should be suppressed as well, since if we cannot control the method signature we also cannot control the types, and it is impossible to narrow or remove `Any` in such cases.

That earlier `ARG002` discussion is [astral-sh/ruff#3952](https://github.com/astral-sh/ruff/issues/3952).

The practical effect is that in a codebase running `preset = "all"`, `explicit-any` is actionable for first-party annotations but produces unsuppressable noise wherever the codebase subclasses a framework. In ours, every `explicit-any` site in `src/` that we could not remove was an `@override` of a Starlette/syrupy method.

This is a different mechanism from [#4088](https://github.com/facebook/pyrefly/issues/4088) (third-party type aliases that expand to `Any`), but the same underlying theme: `explicit-any` firing on `Any` that the first-party code did not choose.

## Workaround

Per-line suppression on each affected parameter and return:

```python
    @override
    def default(self, o: Any) -> Any:  # pyrefly: ignore[explicit-any]
        ...
```

One suppression covers every annotation on its line, so the cost is bounded, but it has to be repeated at every override of every framework method. For `*args: Any, **kwargs: Any` overrides it is worse than it looks: a formatter (we use `ruff format`) splits the parameters onto separate lines once the suppression comment makes the line long enough, which strands the annotations that no longer share a line with the comment and forces one suppression per parameter.

nitishagar:
Fixes #4548

## Problem

With `explicit-any = "error"`, a `@typing.override` method whose base signature is annotated with `Any` must repeat that `Any` — narrowing or dropping it would break the override — yet the rule fired on exactly those annotations:

```python
import json
from typing import Any, override

class CustomEncoder(json.JSONEncoder):
    @override
    def default(self, o: Any) -> Any:   # 2 explicit-any errors
        return super().default(o)
```

The signature is fixed by typeshed's `json.JSONEncoder.default` (`(self, o: Any) -> Any`), so the diagnostic points at code that cannot be changed. Ruff exempts `@override` methods in its equivalent lint (ANN401) for the same reason.

## Change

- The binding step already knows `is_override` (name-only decorator check) when it registers method annotations; those annotations now carry the method's name when they belong to an `@override` method.
- Parameter and return annotations report `explicit-any` from `solve_annotation` (which has the class context) instead of `expr_untype`; a position is exempt only when the annotation is exactly `Any` **and** the corresponding base position is explicitly annotated `Any`:
  - params matched by name (`*args`/`**kwargs` and return positions included),
  - async base returns unwrapped from `Coroutine[Any, Any, T]`,
  - every overload of an overloaded base considered.
- Everything else still reports: undictated positions, methods without `@override`, `@override` with no base member (bad-override still fires), base positions that merely lack annotations, and nested `Any` like `list[Any]`.

The base's *own* explicit `Any`s still report when the base is user code in the checked project (in the issue's case the base is in typeshed and not checked) — only the dictated override positions are exempt.

## Test plan

- [x] 12 new `testcase!`s in `pyrefly/lib/test/inference.rs` covering: exemption (plain/args-kwargs/overload/typing_extensions/async), still-reporting (undictated positions, no decorator, no base member, unannotated base param, nested `Any`, per-position granularity)
- [x] Issue repro checks clean with `explicit-any = "error"`
- [x] `cargo test -p pyrefly --lib -- --skip test::lsp` green (the two `missing-source` LSP interaction failures are pre-existing on `main`)
- [x] Formatting/linting clean

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -2471,7 +2471,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         if ty.is_some_and(Self::is_proxy_method_type) {
             return true;
         }
-        let BindingAnnotation::AnnotateExpr(_, expr, _) = self.bindings().get(annotation) else {
+        let BindingAnnotation::AnnotateExpr(_, expr, _, _) = self.bindings().get(annotation) else {
             return false;
         };
         Self::proxy_method_annotation_syntax_mentions_name(expr)
@@ -2483,7 +2483,7 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
         &self,
         annotation: Idx<KeyAnnotation>,
     ) -> ProxyMethodAnnotationForm {
-        let BindingAnnotation::AnnotateExpr(_, expr, _) = self.bindings().get(annotation) else {
+        let BindingAnnotation::AnnotateExpr(_, expr, _, _) = self.bindings().get(annotation) else {
             return ProxyMethodAnnotationForm::Other;
         };
         self.proxy_method_annotation_expr_form(expr)
@@ -4566,12 +4566,14 @@ impl<'a, Ans: LookupAnswer> AnswersSolver<'a, Ans> {
             })
     }
 
-    /// Return the first inherited method signature (parameters and flags) for `name`.
-    pub(crate) fn inherited_method_signature(
+    /// Return the instantiated type of the first inherited member for `name`
+    /// whose type is accepted by `accept`, walking the MRO.
+    pub(crate) fn inherited_member_where(
         &self,
         cls: &Class,
         name: &Name,
-    ) -> Option<(ParamList, FuncFlags)> {
+        accept: impl Fn(&Type) -> bool,
+    ) -> Option<Type> {
         let derived_instance = self.instantiate(cls);
         for ancestor in self.get_mro_for_class(cls).ancestors(self.stdlib) {
             let parent_cls = ancestor.class_object();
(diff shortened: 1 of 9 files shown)


nitishagar:
Rebased onto current main. Note for reviewers: upstream 5e6ecb2e32 ("Report an explicit `Any` where it is written") replaced the `check_explicit_any`/`can_report_explicit_any` mechanism this PR was built on, so the exemption is ported to the new architecture — `solve_annotation` now solves may-exempt annotations into a local collector and drops the written-position `explicit-any` error via `extend_filtered` only when the base dictates `Any` (same jaxtyping precedent). Also ported to `toplevel_callable_signatures` and the 4-field `AnnotateExpr` threading. All 18 explicit-any tests pass (PR's 10 + upstream's 8); full lib suite: 8783 pass with only the 2 pre-existing `missing_source` LSP failures (verified failing on clean upstream main too).

PR #4801, opened 2026-09-03
add regression test for from X import * in site-packages .py modules is not treated as a re-export #4130

### LINKED ISSUE #4130: from X import * in site-packages .py modules is not treated as a re-export
Hello, all. Thanks for this tool.

### Describe the Bug

Star imports inside a py.typed package installed in site-packages don't re-export names. The identical package as first-party source works.

```python
# fakestarlib/_impl.py
__all__ = ["Starred"]

class Starred:
    value: int = 0

# fakestarlib/_direct.py
class Explicit:
    value: int = 0

# fakestarlib/__init__.py  (package installed in site-packages, with py.typed)

if TYPE_CHECKING:
    from ._direct import Explicit
    from ._impl import *
```

In a project using that environment:

```python
from fakestarlib import Explicit, Starred
# Explicit: ok
# Starred: ERROR Could not import `Starred` from `fakestarlib` [missing-module-attribute]
```

Same file, same TYPE_CHECKING block: the explicit import resolves, the starred one doesn't. A plain star import (no TYPE_CHECKING) fails the same way. Copy the package into the project source tree and both resolve.

If the package also defines a module-level `__getattr__` (pydantic does, annotated `-> object`), the missing name silently becomes `object` instead of erroring, so use sites report `Expected a type form, got instance of object [not-a-type]` and friends.

That's the real-world hit: pydantic's `__init__.py` pulls `.types`, `.networks`, `.main`, `.errors` in via `import *` under TYPE_CHECKING, so `from pydantic import SecretStr / AnyHttpUrl / PositiveInt / BaseModel / create_model` all check as `object`, while `Field` (explicit import, same block) is fine. It can even produce `type[BaseModel] is not assignable to type[BaseModel]` when two packages resolve `BaseModel` through different paths. types-networkx hits it too: `networkx-stubs/__init__.pyi` does `from networkx.classes import *`, so `nx.Graph` is unresolvable with the stubs installed.

Looks related to #2852, but that fix (e5a11e34) synthesized `__all__` for wildcard imports in `.pyi` files only. This is the `.py`-in-site-packages case, still failing on 1.2.0-dev.2.

Workaround: import from the defining submodules (`from pydantic.types import SecretStr`, `from networkx.classes import Graph`, ...).

pyrefly 1.1.1 and 1.2.0-dev.2, Python 3.14, pydantic 2.13.4, types-networkx 3.6.1.20260624. Config uses `python-interpreter-path`; the package is found through that environment's site-packages.

asukaminato0721:
# Summary

Fixes #4130

# Test Plan

add regression test

### DIFF
--- a/pyrefly/lib/test/imports.rs
+++ b/pyrefly/lib/test/imports.rs
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+use std::fs;
+
 use pyrefly_build::handle::Handle;
 use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::module_path::ModulePath;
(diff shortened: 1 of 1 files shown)


pyrefly/lib/test/imports.rs
```diff
@@ -1986,6 +1988,55 @@ fn test_pkgutil_namespace_absorbs_implicit_namespace() {
         .unwrap();
 }
 
+#[test]
+fn test_site_package_star_import_reexport() {
+    let tempdir = tempfile::tempdir().unwrap();
```
kinto0, line 1993: nit: i wonder if the testenv added [here](https://github.com/Jonathangadeaharder/pyrefly/commit/b21ede8829f98afdc964fb5f97b9a204d73f1db7) is a cleaner way of writing the same test

here's AI's analysis of the test coverage we have:
test_all_augassign_unresolvable`** (`imports.rs:1683`) — the closest one. It builds a real package:



pkg/__init__.py       from .sub import *
pkg/sub/__init__.py   from ._impl import *   +  __all__
pkg/sub/_impl.py      class Base: ...


Then it asserts `assert_type(pkg.x, int)` through the chain. So it already covers: package, relative star import, `__all__`, a class, and attribute access. This is most of the shape of #4130.



**`test_implicit_reexport_wildcard_ok`** (`imports.rs:2446`) — plain modules `foo` and `bar`. No package. It tests the `implicit-reexport` diagnostic, which is a different error kind from the one in #4130.



**`test_star_import_type_checking`** (`imports.rs:1508`) — very weak. Module `a` contains only `from typing import TYPE_CHECKING`. There is no class to re-export. It tests almost nothing.



**Why this makes the PRs redundant.** The two PRs add three things on top of the landed tests: `TYPE_CHECKING`, a site-package path, and `py.typed`. I removed each one and re-ran. The result did not change. All three are inert in this harness. So the PRs add no new coverage. They only restate `test_all_augassign_unresolvable` in a longer form, and with a weaker assertion.

kinto0 (changes requested):
would you be able to test this manually? this test seems to imply that the bug doesn't exist on trunk, but I wonder if it's just an artifact of the test

asukaminato0721:
image

pyrefly/lib/test/imports.rs
asukaminato0721: more interestingly, v1.1.1 seems works on this.

kinto0:
thanks!!

PR #4804, opened 2026-09-03
fix Hover should prefer a non-empty .pyi docstring before falling back to .py #4803

### LINKED ISSUE #4803: Hover should prefer a non-empty .pyi docstring before falling back to .py
### Describe the Bug

Pyrefly currently prefers the corresponding `.py` definition when resolving hover documentation, even when the selected `.pyi` declaration has its own non-empty docstring.

Preferring `.py` is a useful fallback when a stub omits documentation, as discussed in #406 and #1090. However, some libraries deliberately ship richer generated documentation in their inline stubs. In that case Pyrefly ignores the more accurate stub documentation.

Minimal layout:

```text
project/
├── foo.py
├── foo.pyi
└── main.py
```

```python
# foo.py
def f():
    """Documentation with an unexpanded %(generated_section)s placeholder."""
    return 1
```

```python
# foo.pyi
def f() -> int:
    """Documentation with the fully expanded generated section."""
    ...
```

```python
# main.py
from foo import f

f()
# Hover f
```

Actual hover:

```text
def f() -> int

Documentation with an unexpanded %(generated_section)s placeholder.
```

Expected hover:

```text
def f() -> int

Documentation with the fully expanded generated section.
```

In other words, documentation resolution should preferably be:

1. Use the `.pyi` docstring when it is non-empty.
2. Fall back to the corresponding `.py` docstring only when the stub has no documentation.

This can remain independent from go-to-definition, which may still reasonably prefer the executable `.py` implementation.

This behavior occurs in UltraPlot, whose generated inline stubs statically expand reusable docstring snippets:

- Source declaration with placeholders: https://github.com/Ultraplot/UltraPlot/blob/fix/static-parsing/ultraplot/ui.py#L199
- Generated stub with expanded documentation: https://github.com/Ultraplot/UltraPlot/blob/fix/static-parsing/ultraplot/ui.pyi#L371

BasedPyright/Pylance display the `.pyi` documentation for this case. Pylance's documented policy is also to use a non-empty stub docstring and fall back to source only when it is absent:

https://github.com/microsoft/pylance-release/blob/main/docs/howto/docstring-resolution-order.md#rule-f--stub-vs-source

Reproduced with both:

- Pyrefly 1.2.0
- Pyrefly 1.3.0-dev.3

### Sandbox Link

Not provided; the reproduction requires paired `.py` and `.pyi` files for the same module. The complete three-file reproduction is included above.

Reproduced directly against `pyrefly lsp` using a `textDocument/hover` request, so this does not appear to be editor-extension-specific.

asukaminato0721:
# Summary

Fixes #4803

Hover now prefers non-empty .pyi documentation and falls back to .py for empty/missing stub docs 

# Test Plan

add test


kinto0 (changes requested):
great change, thank you!

I wonder if we could move docstring resolution into its own function since this diff makes it bigger? this function is getting too large

asukaminato0721:
https://github.com/facebook/pyrefly/commit/6ace9af5643e960ab25e6afbbba540bb66172469

PR #4805, opened 2026-09-03
[tensor-shapes] add stubs for jnp.array and other general JAX array creation APIs

jakevdp:
### Summary

Type stubs for general array creation apis: `jnp.array`, `jnp.asarray`, `jnp.atleast_[123]d`

### Test Plan

Adds a new test file; all tests pass locally:
```
328 files already formatted
All checks passed!
Running formatting...
Finished in 2.66 seconds.
Running linting...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.84s
Finished in 1.06 seconds.
    Finished `release` profile [optimized + debuginfo] target(s) in 0.52s
PASS stubs (9 files)
PASS arithmetic (1 files)
PASS contractions (1 files)
PASS creation (1 files)
PASS fft (1 files)
PASS lax (1 files)
PASS linalg (1 files)
PASS matmul (1 files)
PASS nn (1 files)
PASS reductions (1 files)
PASS reshape (1 files)
```

### AI disclosure

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
+++ b/tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
@@ -7,6 +7,9 @@ from typing import Any, Callable, Literal, overload, Sequence, Unpack
 
 from jax._array import Array as Array, Array as ndarray
 from jax._shapes import (
+    atleast_1d_shape,
+    atleast_2d_shape,
+    atleast_3d_shape,
     broadcast_to_shape,
     column_stack_shape,
     concatenate_shape,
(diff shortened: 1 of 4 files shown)


tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -1264,6 +1324,36 @@ def reshape(
...
+@overload
+def atleast_1d(ary: _Scalar, /) -> Array[[1]]: ...
+@overload
+def atleast_1d[Shape: _Shape](
+    ary: Array[Shape], /
+) -> Array[atleast_1d_shape(Shape)]: ...
+@overload
+def atleast_1d(ary: Any, /) -> Array[IntTuple]: ...
+@overload
+def atleast_1d(*arys: Any) -> list[Array[IntTuple]]: ...
```
jakevdp, line 1536: Unfortunately since this returns a list rather than a tuple, this is about the best we can do here.

NumPy changed these return values from lists to tuples in v2.0; I think we'll try to get the same change into JAX for v0.12.0 (mid-October) then we can have more granular types for these overloads.

stroxler: We could potentially lie about the types in pyrefly-tensor-shapes. For many purposes, that's probably fine - tuple is *mostly* more restrictive except in a few edge cases.

I've also considered trying to introduce a new type system primitive for "it's a list value, but behaves like a tuple" if we really need to because this does come up (in torch stubs as well)

stroxler (approved):

PR #4808, opened 2026-09-03
[WIP] Proposal: coercible shape_extensions.Scalar

jakevdp:
**DO NOT MERGE – draft only**

This PR is a draft/proposal for a new `shape_extensions.Scalar` type that can be used to simplify APIs in NumPy and related libraries that allow scalars to be coerced into arrays.

Currently, and API like `jax.numpy.add` requires multiple overloads to allow mixing of scalar and array values:
```python
from shape_extensions import IntTuple

type Scalar = bool | int | float | complex

@overload
def add[Shape: IntTuple](x1: Array[Shape], x2: Scalar) -> Array[Shape]: ...
@overload
def add[Shape: IntTuple](x1: Scalar, x2: Array[Shape]) -> Array[Shape]: ...
@overload
def add[Shape1: IntTuple, Shape2: IntTuple](
    x1: Array[Shape1],
    x2: Array[Shape2]
) -> Array[broadcast(Shape1, Shape2)]: ...
```

This PR contains a proposed `shape_extensions.Scalar` which allows this common pattern to be expressed much more succinctly:
```python
from shape_extensions import IntTuple, Scalar

type ArrayLike[Shape: IntTuple] = Array[Shape] | Scalar[Shape]

def add[Shape1: IntTuple, Shape2: IntTuple](
    x1: ArrayLike[Shape1],
    x2: ArrayLike[Shape2]
) -> Array[broadcast(Shape1, Shape2)]: ...
```

The `Scalar` type defined here is specially recognized by pyrefly, and is only valid when its argument is an empty shape. This would allow libraries to compactly specify APIs which allow scalar inputs to be coerced to scalar arrays without the need for many overloaded signatures.

This PR adds a proposed `Scalar` implementation, modifies `jax.numpy.add` to make use of it, and adds a test to cover its intended semantics.

A few open questions:

- how to support library-defined scalar types, e.g. `numpy.number`? Perhaps the Scalar API should be something more general, like `Scalar[Shape, int | float | complex]`?
- how to support dtype annotations like those used by NumPy?
- `Scalar[Shape]` would admit arbitrary shapes in Python's general type system. Can we solve this a different way?
- other considerations I'm not thinking of?

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -5133,8 +5133,24 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
         {
             return Type::ClassType(base_class);
         }
-        self.shaped_array_classtype_to_shaped_array_type(&base_class)
-            .to_type()
+        let shaped_array = self.shaped_array_classtype_to_shaped_array_type(&base_class);
+        if cls.has_toplevel_qname("shape_extensions", "Scalar") {
+            let is_valid = match shaped_array.shape().view() {
+                IntTupleView::Concrete(dims) => dims.is_empty(),
+                IntTupleView::Unpacked { .. } => true,
+                IntTupleView::Gradual => false,
+            };
+            if !is_valid {
+                self.error(
+                    errors,
+                    range,
+                    ErrorKind::InvalidAnnotation,
+                    "`shape_extensions.Scalar` only accepts an empty shape `[]`".to_owned(),
+                );
+                return Type::any_error();
+            }
+        }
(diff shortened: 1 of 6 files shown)


stroxler:
We decided that for now, we'll just use defaults to achieve the functionality we want here without a new type system concept: for the use case we have in mind,
```
def f[S: IntTuple = []](x: Array[S] | ndarray[S] | float): Array[S]: ...
```
will have the desired affect of binding `S` to `IntTuple[()]` whenever a float is passed.

The downside is that you do have to remember the default, otherwise `S` will silently solve to gradual `IntTuple` when we pass a float; we can potentially implement a type-aware lint to catch this.

PR #4809, opened 2026-09-03
Accept shaped arrays and scalars in broadcasting APIs

stroxler:
Summary: Introduce a reusable JAX array-like alias that preserves shapes from JAX and NumPy arrays while treating scalar inputs as empty shapes. Use it to express `maximum` and three-argument `where` without combinatorial overloads.

Differential Revision: D118557925

### DIFF
--- a/pyrefly/lib/alt.rs
+++ b/pyrefly/lib/alt.rs
@@ -23,6 +23,7 @@ pub mod operators;
 pub mod overload;
 pub mod polars_specials;
 pub mod regex;
+pub mod scalar_as_shape;
 pub mod shape_flag;
 pub mod singledispatch;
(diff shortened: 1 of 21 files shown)


stroxler:
We decided that for now, we'll just use defaults to achieve the functionality we want here without a new type system concept: for the use case we have in mind,
```
def f[S: IntTuple = []](x: Array[S] | ndarray[S] | float): Array[S]: ...
```
will have the desired affect of binding `S` to `IntTuple[()]` whenever a float is passed.

The downside is that you do have to remember the default, otherwise `S` will silently solve to gradual `IntTuple` when we pass a float; we can potentially implement a type-aware lint to catch this.

PR #4814, opened 2026-09-04
[tensor-shapes] add JAX BERT example

jakevdp:
Here I've deliberately avoided dependence on any neural network framework (such as flax or equinox), instead opting to work in terms of simple dataclasses registered as JAX pytrees. This keeps the example simple and self-contained with no hidden magic, and requires only shape-annotated JAX APIs.

This required a few updates to JAX's shape stubs, also done as part of this PR.


stroxler (approved):

PR #4816, opened 2026-09-04
narrow membership tests against named concrete containers (#4469)

### LINKED ISSUE #4469: Membership test against a named `set` does not narrow `<type> | None`.
### Describe the Bug

Pyrefly narrows an optional value after a membership test against an inline set literal, but does not perform the same narrowing when the set is assigned to a variable.

```
def f() -> str | None: ...
def h(y: str) -> None: ...
x = f()
d: set[str] = {"test", "test2"}
if x in d:
    h(x)  # ERROR: Argument `str | None` is not assignable to parameter `y` with type `str` in function `h` [bad-argument-type]
if x in {"test", "test2"}:
    h(x)  # No error
```

Both membership checks should narrow `x` from `str | None` to `str`. Since `d` has the explicit type `set[str]`, reaching the body of `if x in d` establishes that `x` is a member of a `str` `set` and therefore is a `str`.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.fZQ_D4IwEMW_ykUWTZTubA6ualh1KEpDHaQEikKM393XK0jwDxu59o7rvd-9f5bJIHr9gEUHX0RhGHpI9bzlZRsU9mcNukQeLuFY-dVD6HHkho6zJd7gvjCVJ6veOBXTbq563iyIAtrE8S6OaF1m9VXlluTQg3SAY0LOEzEy52FkDRQpAQ-sjGQr6X6x2hm-4kyk5O-tIqklHU5JusLKcfWVuwhl-mZ-9vrR4NZ4SL4hCGjvA9TNE_YHerIaJovfY0QB_FWRtraoIiH6dFNmQuUiNedKjDIEOd6TW3Jh6x6XI8PQVig7QV9RKmjBIUDM-Ewx8QI

_No response_

shobhitmehro:
named containers fell through membership narrowing since only inline literals used their element type. this narrows list, set, frozenset, and deque through the same equality-aware path as ==, so it doesn't over-narrow stuff like float against set[int]. abstract containers and custom `__contains__` stay untouched since they don't have to follow equality.

fixes #4469 

# Test Plan

`python3 test.py`

### DIFF
--- a/pyrefly/lib/alt/narrow.rs
+++ b/pyrefly/lib/alt/narrow.rs
@@ -337,6 +337,22 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
         self.unions(matches)
     }
 
+    /// Element type of a `list`, `set`, `frozenset`, or `deque`.
+    fn concrete_container_element_type(&self, ty: &Type) -> Option<Type> {
+        let Type::ClassType(cls) = ty else {
+            return None;
+        };
+        let obj = cls.class_object();
+        let is_concrete = obj == self.stdlib.list_object()
+            || obj == self.stdlib.set_object()
+            || obj == self.stdlib.frozenset_object()
+            || cls.has_qname("collections", "deque");
+        if !is_concrete {
+            return None;
+        }
+        self.unwrap_iterable(ty)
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #4819, opened 2026-09-05
set debug level to 2 in the dev profile to allow inspecting variable values

DetachHead:
# Summary

@connernilsen this is what i was talking about the other day. changing this allows the debugger to see the values of variables at runtime

https://doc.rust-lang.org/cargo/reference/profiles.html#debug

# Test Plan

clicked the "Debug" button on a test and hovered over a variable (tested with the [CodeLLDB](https://github.com/vadimcn/codelldb) extension):

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ codegen-units = 1
 strip = "debuginfo"
 
 [profile.dev]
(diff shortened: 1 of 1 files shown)


connernilsen (approved):

PR #4821, opened 2026-09-05
enable the `dbg_macro` clippy rule

DetachHead:
# Summary

after @connernilsen pointed out that i accidentally committed a `dbg!` macro, i wondered if there was a clippy rule to prevent such mistakes (see https://github.com/facebook/pyrefly/pull/4729#discussion_r3888493732). turns out there is!

note that enabling a rule globally seems to be nowhere near as straightforward as it is in most other linters, so i had to make some other changes:

- i had to set `workspace = true` in each package (https://github.com/rust-lang/rust-clippy/issues/16541#issuecomment-3881220162)
- i also had to move the existing `unexpected_cfgs` config out of the individual packages into the global config, to workaround https://github.com/rust-lang/cargo/issues/13157 

# Test Plan

intentionally inserted a `dbg!` macro and ran `cargo clippy`:

```
error: the `dbg!` macro is intended as a debugging tool
  --> crates/pyrefly_config/src/migration/error_codes.rs:72:9
   |
72 |         dbg!(&pyright_cfg.type_checking_mode);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#dbg_macro
   = note: requested on the command line with `-D clippy::dbg-macro`
help: remove the invocation before committing it to a version control system
   |
72 -         dbg!(&pyright_cfg.type_checking_mode);
72 +         &pyright_cfg.type_checking_mode;
   |

error: could not compile `pyrefly_config` (lib) due to 1 previous error
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,3 +41,9 @@ resolver = "2"
 [workspace.package]
 license = "MIT"
 repository = "https://github.com/facebook/pyrefly"
+
+[workspace.lints.clippy]
+dbg_macro = "deny"
+
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(fbcode_build)"]}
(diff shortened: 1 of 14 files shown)


stroxler (approved):
Thanks!

Don't worry about the red internal signal, I had to patch this because cargo config is derived from our internal buck build setup but the change is good. I'll try to get this in soon, merging requires a second review and it's a holiday weekend so I'm not sure how soon

samwgoldman (approved):

PR #4827, opened 2026-09-05
fix false bad-return for a generator function with Iterator[X] | T return annotation #4825

### LINKED ISSUE #4825: false `bad-return` for a generator function with `Iterator[X] | T` return annotation
### Describe the Bug

```py
from collections.abc import Iterator


def gen(x: int) -> Iterator[int] | int:
    yield from range(4)
    return 5
```

```console
ERROR sandbox.py:6:12-13: Returned type `Literal[5]` is not assignable to declared return type `None` [[bad-return](https://pyrefly.org/en/docs/error-kinds/#bad-return)]
```

All either type checkers are fine with this, but Pyrefly emits the error above

Noticed in huggingface-hub

https://github.com/huggingface/huggingface_hub/blob/deb97775d91cb1ce2a4c254b29ad30081044257d/src/huggingface_hub/_hot_reload/client.py#L70-L79

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.w1tk4ihFYCUGKC4gSTY9NU-jwgoYH5BEDlMQDRSIBSZcIAUNuEpQUaEANrwoMS89VcMEGFVIic0UM1SVFQIgAgpQBwKdBYyO9FKgFUCX6cXkKQMLrFSFjJKSgmIrfX2Y9vyidP3UPP2U_ORifRQd-gqgBJRYlpgJLgtRjVPIB6eCYqCxlESnOiQ21fFFpjooLoEq8MYPAA

_No response_

asukaminato0721:
# Summary

Fixes #4825

Iterator and Iterable annotations now allow generator return values, while explicit Generator return types remain checked.

# Test Plan

add test

### DIFF
--- a/conformance/third_party/conformance.exp
+++ b/conformance/third_party/conformance.exp
@@ -1802,17 +1802,6 @@
     },
     {
       "code": -2,
-      "column": 12,
-      "concise_description": "Returned type `Literal[True]` is not assignable to declared return type `None`",
-      "description": "Returned type `Literal[True]` is not assignable to declared return type `None`",
-      "line": 71,
-      "name": "bad-return",
-      "severity": "error",
-      "stop_column": 16,
-      "stop_line": 71
-    },
-    {
-      "code": -2,
       "column": 11,
       "concise_description": "Yielded type `B` is not assignable to declared yield type `A`",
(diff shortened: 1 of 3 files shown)


kinto0 (approved):

PR #4831, opened 2026-09-07
fix missing syntax highlighting after multi-target assignments, walrus assignments, etc. #4739

### LINKED ISSUE #4739: missing syntax highlighting after multi-target assignments, walrus assignments, etc.
### Describe the Bug

Pyrefly does not emit variable semantic tokens for references to names defined by several assignment-like constructs. The definition sites are highlighted as variables, but later uses fall back to the editor's syntax highlighting.

**Version:** `pyrefly 1.2.0`

## Reproduction

Open the following file in an editor using the Pyrefly language server:

```python
multi_a = multi_b = 1
print(multi_a, multi_b)

if (walrus := 1):
    print(walrus)
```

The assignment targets `multi_a`, `multi_b`, and `walrus` are highlighted as variables, but their later references are not. With the VS Code theme where I noticed this, the later references are white rather than blue.

Changing either construct to an ordinary single-target assignment restores the expected highlighting:

```python
multi_a = 1
multi_b = 1
walrus = 1

print(multi_a, multi_b, walrus)
```

Image

Annotation-only declarations exhibit the same semantic-token behavior:

```python
declared: int
print(declared)
```

The definition site receives a `variable` token, while the later reference does not. This case is about symbol classification independently of whether a particular runtime control flow initializes the declaration before use.

## Expected behavior

References to variables should receive `variable` semantic tokens regardless of whether the variables were defined by an ordinary assignment, a multi-target assignment, an assignment expression, or an annotation-only declaration.

## Actual behavior

Pyrefly emits `variable` tokens for the definition sites but no semantic tokens for their later load sites. Since Pyrefly advertises an augmenting, partial semantic-token stream, the editor falls back to its baseline syntax highlighting for those references.

## Scope

The behavior follows the binding variant used for the definition:

| Construct | Binding reached for the definition | Symbol kind | Later references |
| --- | --- | --- | --- |
| Ordinary single-target assignment | `NameAssign` | `Variable` | Correctly highlighted |
| Multi-target assignment such as `a = b = value` | `MultiTargetAssign(..., None)` | `None` | Missing token |
| Assignment expression such as `(a := value)` | `Expr` | `None` | Missing token |
| Annotation-only declaration such as `a: int` | `AnnotatedType` | `None` | Missing token |
| Tuple/list unpacking | `UnpackedValue` | `Variable` | Correctly highlighted |
| Loop and comprehension targets | `IterableValueLoop` / `IterableValueComprehension` | `Variable` | Correctly highlighted |
| `with ... as` and exception-handler targets | `ContextValue` / `ExceptionHandler` | `Variable` | Correctly highlighted |
| Pattern captures | `PatternCapture` | `Variable` | Correctly highlighted |
| Augmented assignments | `AugAssign` | `Variable` | Correctly highlighted |
| Names passing through branches or narrowing | `Forward`, `Phi`, `LoopPhi`, or `Narrow` | Resolution follows the underlying definition | Correctly highlighted |

The remaining binding variants with no symbol kind are synthetic, anonymous, flow-internal, error-recovery, attribute/subscript, deletion, or otherwise do not represent another ordinary variable-definition form.

## Cause

`Binding::symbol_kind()` currently groups the three affected definition-producing variants into its `None` branch:

https://github.com/facebook/pyrefly/blob/9cb2b27cea7c0e8aa4421565f23b9dd59fb05484/pyrefly/lib/binding/binding.rs#L2933-L2959

```rust
Binding::Expr(_, _)
| Binding::StmtExpr(_, _)
| Binding::MultiTargetAssign(_, _, _, None)
// ...
| Binding::AnnotatedType(_, _)
// ...
=> None,
```

Walrus targets are bound as `Binding::Expr`:

https://github.com/facebook/pyrefly/blob/9cb2b27cea7c0e8aa4421565f23b9dd59fb05484/pyrefly/lib/binding/expr.rs#L1120-L1126

Semantic-token generation resolves load-context names through a `BoundName` key and only emits a token when that resolution produces a symbol kind:

https://github.com/facebook/pyrefly/blob/9cb2b27cea7c0e8aa4421565f23b9dd59fb05484/pyrefly/lib/state/semantic_tokens.rs#L439-L459

Definition sites happen to remain highlighted through separate fallbacks. Store-context names fall back to `SemanticTokenType::VARIABLE`, and `AnnAssign` targets are explicitly emitted as variables. There is no corresponding fallback for later load sites, which exposes the missing definition metadata.

The affected variants appear to need variable symbol metadata when they are reached through a definition key. Conceptually, the classification would include:

```rust
Binding::Expr(_, _)
| Binding::MultiTargetAssign(_, _, _, None)
| Binding::AnnotatedType(_, _) => Some(SymbolKind::Variable),
```

This would also allow the existing type-aware refinement in `key_to_export` to upgrade the symbol kind when one of these constructs binds a function, class, module, or type alias. `Binding::Expr` is also used for anonymous expression bindings, so the implementation may instead need to apply the variable fallback at the definition/IDE boundary where the key is known rather than classifying every expression binding globally.

asukaminato0721:
# Summary

Fixes #4739

References after multi-target assignments, walrus assignments, and annotation-only declarations now receive semantic highlighting, including function/class refinement.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/binding/binding.rs
+++ b/pyrefly/lib/binding/binding.rs
@@ -2954,14 +2954,14 @@ impl Binding {
             Binding::MultiTargetAssign(_, _, _, Some(_)) => Some(SymbolKind::Class),
             Binding::UnpackedValue(value) if value.receiver.is_some() => Some(SymbolKind::Class),
             Binding::UnpackedValue(_) => Some(SymbolKind::Variable),
-            Binding::AugAssign(_, _) => Some(SymbolKind::Variable),
-            Binding::Expr(_, _)
-            | Binding::StmtExpr(_, _)
+            Binding::AugAssign(_, _)
+            | Binding::Expr(_, _)
             | Binding::MultiTargetAssign(_, _, _, None)
+            | Binding::AnnotatedType(_, _) => Some(SymbolKind::Variable),
+            Binding::StmtExpr(_, _)
             | Binding::ReturnExplicit(_)
             | Binding::ReturnImplicit(_)
             | Binding::ReturnType(_)
-            | Binding::AnnotatedType(_, _)
             | Binding::None
             | Binding::Any(_)
             | Binding::Forward(_)
--- a/pyrefly/lib/test/lsp/semantic_tokens.rs
+++ b/pyrefly/lib/test/lsp/semantic_tokens.rs
@@ -162,6 +162,107 @@ token-type: variable, token-modifiers: [readonly]
 }
 
 #[test]
+fn assignment_forms_test() {
+    let code = r#"
+multi_a = multi_b = 1
+multi_a, multi_b
+if (walrus := 1):  # pyrefly: ignore[redundant-condition]
+    walrus
+declared: int
+declared  # pyrefly: ignore[unbound-name]
+"#;
+    assert_full_semantic_tokens(
+        &[("main", code)],
+        r#"
+# main.py
+line: 1, column: 0, length: 7, text: multi_a
+token-type: variable
+
+line: 1, column: 10, length: 7, text: multi_b
+token-type: variable
+
+line: 2, column: 0, length: 7, text: multi_a
+token-type: variable
+
+line: 2, column: 9, length: 7, text: multi_b
+token-type: variable
+
+line: 3, column: 4, length: 6, text: walrus
+token-type: variable
+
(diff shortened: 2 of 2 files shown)


fangyi-zhou (approved):

PR #4834, opened 2026-09-07
Diagnose Polars `join` suffix collisions and defer selector/pattern keys

alexander-beedie:
# Summary

### Problem

* Polars `join` ops can produce duplicate column names when a disambiguating suffix is applied; for example, a right-hand `colx` becoming `colx_<suffix>` when that suffixed name already exists. `pyrefly` would miss this (it raises `DuplicateError` at runtime).

* Wildcard and regex join keys (such as `"*"` and `"^id$"`) get interpreted as literal column names, producing incorrect `unknown-column` diagnostics.

### Solution

* Detect suffix collisions in `join` output schemas and report `duplicate-column`.

* Use the existing column-argument classification to leave wildcard and regex `on` keys unresolved (note: I'll take a look at adding proper `selector`[^1] support to the Polars "on" parameter - not unreasonable to think that should work 🤔).

* Share `duplicate-name` detection between `join` ops, `hstack`, and CSV renaming.

## Also

* Generalised the diagnostic wording and documentation to cover both projections _and_ joins (eg: prefer use of "operation" over "projection").

# Test Plan

* All existing Polars tests pass (`cargo test --quiet test::polars`).
* Added new coverage for right-hand suffix collisions, valid `semi`/`anti` joins with otherwise-colliding schemas, and wildcard/regex key fallback.

[^1]: [Polars Selectors](https://docs.pola.rs/api/python/stable/reference/selectors.html)

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -167,7 +167,7 @@ pub enum ErrorKind {
     DirectAbstractBaseInstantiation,
     /// Division, floor division, or modulo by a literal zero value.
     DivisionByZero,
-    /// A Polars projection produces more than one column with the same name.
+    /// A Polars operation produces more than one column with the same name.
     DuplicateColumn,
     /// A function has an empty body despite declaring a non-None return type.
     EmptyBody,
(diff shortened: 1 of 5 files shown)


rchen152 (approved):

PR #4837, opened 2026-09-07
Use argument context for Polars column completions

alexander-beedie:
# Summary

### Problem

* The method-wide completion rules could offer column names for strings that actually represent data, such as `df.filter(foo="")` or `df.select(pl.lit(""))`, and also for control arguments.

* Dictionary-key completion could also intercept column-name strings in unpacked expressions such as `df.select(**{"alias": ""})`.

### Solution

* Check argument context and dataframe kind when deciding whether to offer column completions. Recognise `col` and `lit` through their resolved callables, preserving the distinction between column references and literal values.

* Keep suggestions tied to the nearest dataframe call, and allow unpacked named-expression values to reach column completion (while retaining normal dict-key handling).

# Test Plan

* All existing unit tests pass.
* Multiple new unit tests added.

### DIFF
--- a/pyrefly/lib/alt/polars_specials.rs
+++ b/pyrefly/lib/alt/polars_specials.rs
@@ -138,14 +138,6 @@ fn is_pandas_dataframe(cls: &Class) -> bool {
     RuntimeClass::PandasDataFrame.matches(cls)
 }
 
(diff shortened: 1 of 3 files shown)


rchen152 (approved):

PR #4839, opened 2026-09-07
fix Concrete type not assignable to gradual type when TypeVarTuple is involved #4609

### LINKED ISSUE #4609: Concrete type not assignable to gradual type when TypeVarTuple is involved
### Describe the Bug

In the following code example:
```py
from typing import Any, Callable

class K[*Ts]:
    func: Callable[[Any, *Ts], None]

def g[*Ts](x: int, *xs: *Ts):
    K[*Ts]().func(x, *xs)
```
The call to `func` gets rejected because "Unpacked argument `tuple[int, *Ts]` is not assignable to varargs type `tuple[Any, *Ts]`". But it should be assignable, because Any is a gradual type. Even if it wasn't, tuples are also covariant.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.I7fIRGQn72itkOJYaMCA0q4VXFF0NFgPSB5YRAKTATgXglJHOliTRoUVMA6BhZlWBTCrAkVgyQ9ipoamHsg8jQqwAmCcYga5skIAREAB6npgYQOMq_RSYJEGjAO9mDxlYGmWqpBRUlJQbKWvD9OeX5Sun5qnn5KfXKyPokNfAZS6EssSM8E-QDVOIR-cRIqBxlIS1-qQqFbHF9PqoIgGqsAbeQA

_No response_

asukaminato0721:
# Summary

Fixes #4609

removes an extra tuple wrapper around forwarded `*Ts` arguments before checking assignability.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/callable.rs
+++ b/pyrefly/lib/alt/callable.rs
@@ -16,6 +16,7 @@ use pyrefly_types::literal::Lit;
 use pyrefly_types::literal::LitStyle;
 use pyrefly_types::meta_shape_dsl::MetaShapeFunction;
 use pyrefly_types::meta_shape_dsl::ShapeTransform;
+use pyrefly_types::simplify::simplify_tuples;
 use pyrefly_types::tuple::Tuple;
 use pyrefly_types::typed_dict::ExtraItems;
 use pyrefly_types::types::TArgs;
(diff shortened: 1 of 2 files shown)


stroxler (approved):

PR #4841, opened 2026-09-07
fix(bench): resolve relative PyTorch checkout paths

### LINKED ISSUE #4828: Relative PYREFLY_PYTORCH_BENCH_PATH causes a panic in LSP benchmarks
Description

`pytorch_root()` accepts an existing relative path from `PYREFLY_PYTORCH_BENCH_PATH` and returns it unchanged.
The LSP benchmarks then pass that path to `Url::from_file_path(root).unwrap()` which requires an absolute path.
A valid relative checkout path therefore leads to a panic.

Reproduction

Set `PYREFLY_PYTORCH_BENCH_PATH` to a relative path pointing to an existing PyTorch checkout containing `torch/nn/__init__.py`.
For example:

```bash
PYREFLY_PYTORCH_BENCH_PATH=./pytorch cargo bench -p pyrefly --bench pytorch -- cold_start
```

The relevant path through the code is:

1. `pytorch_root()` checks that the sentinel exists under the override path.
2. It returns the relative `PathBuf` without making it absolute.
3. `cold_start_definition()` passes it to `Url::from_file_path(root).unwrap()`.
4. The conversion returns `Err(())` and `unwrap()` panics.

The same URL conversion is also present in `error_propagation.rs`.

Expected behavior

An existing relative checkout path should work like the equivalent absolute path.
Alternatively an unsupported relative path should produce a clear error instead of a panic.

Actual behavior

An isolated reproduction using the unchanged checkout helper and the same URL conversion produced:

```text
root=fixture
is_absolute=false
file_url=Err(())
called `Result::unwrap()` on an `Err` value: ()
```

The equivalent absolute path to the same directory succeeded.

Possible fix

Resolve the accepted override to an absolute path before returning it from `pytorch_root()`.
Handle path resolution errors explicitly.

Code locations

- https://github.com/facebook/pyrefly/blob/ce6b61ec55894912b3679846b5e604560aba7a8a/pyrefly/benches/pytorch/common.rs#L123-L130
- https://github.com/facebook/pyrefly/blob/ce6b61ec55894912b3679846b5e604560aba7a8a/pyrefly/benches/pytorch/cold_start.rs#L65
- https://github.com/facebook/pyrefly/blob/ce6b61ec55894912b3679846b5e604560aba7a8a/pyrefly/benches/pytorch/error_propagation.rs#L90

Validation scope

Checked `main` at commit `ce6b61ec55894912b3679846b5e604560aba7a8a` on macOS.
The checkout helper was exercised in an isolated harness with a small local fixture containing the sentinel file.
Only the unused LSP argument types were replaced with stand-ins.
The path conversion was tested with the `url` crate.
The full PyTorch/LSP benchmark was not run.

kavyansh18:
## Summary

Fixes a panic when `PYREFLY_PYTORCH_BENCH_PATH` is set to a valid relative PyTorch checkout path.

The benchmark path override was returned unchanged, allowing a relative path to reach `Url::from_file_path(...).unwrap()`, which rejects relative paths and causes the benchmark to panic.

This change canonicalizes a valid override path before returning it, so relative checkout paths are resolved to absolute paths while existing absolute paths continue to work unchanged.

## Linked issue

Fixes #4828

## Test plan

- `cargo fmt -- --check`
- `cargo bench -p pyrefly --bench pytorch --no-run`
- `cargo test -p pyrefly_util`
- `git diff --check`
- Verified a relative `PYREFLY_PYTORCH_BENCH_PATH` using a minimal PyTorch checkout fixture with `cargo bench -p pyrefly --bench pytorch -- --list`

All checks passed.

## Risk / breaking changes

No public API changes.

The change is limited to resolving the existing PyTorch benchmark path override to a canonical absolute path.

### DIFF
--- a/pyrefly/benches/pytorch/common.rs
+++ b/pyrefly/benches/pytorch/common.rs
@@ -127,7 +127,7 @@ fn pytorch_root() -> Option<PathBuf> {
     if let Some(root) = override_root
         && root.join(SENTINEL).exists()
     {
-        return Some(root);
+        return fs::canonicalize(&root).ok();
     }
 
     if let Some(root) = buck_resource_root() {


kavyansh18:
hey @rchen152 can you please review the PR

rchen152 (approved):

PR #4855, opened 2026-09-08
Push a notification when the build system status changes (#4855)

connernilsen:
Summary:

`typeErrorDisplayStatus` is a request, so the client only learns about
changes when it asks. Build system functionality makes it so we might need to push state from the language server, which complicates this.

Now, we create a new `pyrefly/typeErrorDisplayStatusChanged` notification, which tells the IDE to re-pull the state from the language server.

Reviewed By: kinto0

Differential Revision: D118569337


kinto0 (approved):

PR #4857, opened 2026-09-08
[tensor-shapes] annotate jax dtype arguments with jax.typing.DTypeLike

jakevdp:
# Summary

Use `dtype: jax.typing.DTypeLike` rather than `dtype: Any` across JAX stubs.


stroxler (approved):

PR #4866, opened 2026-09-09
Add experimental programmatic API for Monty

yangdanny97:
# Summary

For https://www.github.com/pydantic/monty/pull/501

Something about how Monty can't depend on a crate that's a GH revision hash.

I guess after this gets merged you guys need to publish it, but IDK how to do that

# Test Plan

I ran all the tests at some point but after the demise of my Macbook Pro I now use a Macbook Air which has no fans so I'm reluctant to risk running the test suite and overheating it

so, TLDR trust CI


pyrefly/lib/embed.rs
```diff
@@ -0,0 +1,214 @@
...
+    state: State,
+    sys_info: SysInfo,
+}
+
+impl Checker {
+    /// Build a checker for the given Python version (e.g. `"3.14"`, or the default
+    /// when `None`) over the in-memory modules named in `modules`. Only those module
+    /// names are importable between the supplied sources; everything else resolves to
+    /// the bundled typeshed. No interpreter is queried.
+    pub fn new(python_version: Option<&str>, modules: &[&str]) -> Result<Self, String> {
```
kinto0: it's unclear what modules is. is this every module this API can check? 

the "config" is just a python_version. I imagine we will need a lot more config at some point. the config options that are essential here depends on the use-case. should we use the full ConfigFile struct?

pyrefly/lib/embed.rs
```diff
@@ -0,0 +1,214 @@
...
+        Ok(Self {
+            state: State::new(config_finder, ThreadCount::default()),
+            sys_info,
+        })
+    }
+
+    /// Type check the `target` module, returning diagnostics for it only.
+    ///
+    /// `files` supplies the current source for each in-memory module (each
+    /// `(module_name, source)`); every name must have been declared in
```
kinto0: > every name must have been declared

this is a confusing API and I don't think it can land like this

kinto0 (changes requested):
somewhat confusing API for an unclear use-case. if we ever merge something like this, I think it needs to be simple and have clear documentation.

I also don't see a disclaimer that this is experimental anywhere in the code

yangdanny97:
In my view this is in the same category as glean's API or the WASM stuff, where it should be undocumented until we refine the API enough and decide to publicize it

The main purpose of merging it is so that Monty can actually release a version that uses Pyrefly

pyrefly/lib/embed.rs
yangdanny97: I don't think we should overdo it now - the full config is excessive IMO since it's just a REPL

we can add it later if we need it

yangdanny97:
Cleaned up a few things

pyrefly/lib/embed.rs
yangdanny97: modules for Monty is 2 files, 1 is the "file" containing the REPL, 2 is the stubs that Monty injects

kinto0 (approved):

stroxler (approved):

PR #4871, opened 2026-09-09
Update inference to better support nested Polars dtypes

alexander-beedie:
# Summary

### Problem

* Polars schemas containing nested dtype constructors lose column-aware inference:  specifically the `Array`,  `List`, and `Struct` types.

### Solution

* Add inference points for nested dtypes.
* Preserve existing fallback behaviour for dynamic or unsupported arguments.

# Test Plan

* All existing unit tests pass.
* New tests added for:
  * nested constructors
  * keyword arguments
  * fallback behaviour
  * constructor diagnostics
  
  ---
  
**Note**: the Windows test failure is unrelated to this PR (the same error seems to be occurring in a number of recent PRs), but I took a look and fixed it here: https://github.com/facebook/pyrefly/pull/4872 👍


rchen152 (approved):

pyrefly/lib/test/polars/nested_dtype.rs
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+use super::dataframe::env_with_polars_stubs;
```
rchen152, line 8: If this env is useful in multiple test files, could be convenient to start a polars/util.rs for it, like we have for django: https://github.com/facebook/pyrefly/blob/main/pyrefly/lib/test/django/util.rs. (For a future PR, if you're so inclined.)

PR #4873, opened 2026-09-09
fix Generics usage for a TypedDict's extra_items breaks type checking #4870

### LINKED ISSUE #4870: Generics usage for a TypedDict's extra_items breaks type checking
### Describe the Bug

The simplest example would be
```python
from typing import TypedDict

class TD[T](TypedDict, extra_items=T):
    a: int

d: TD[str] = {"a":1}
reveal_type(d2["b"])  # T  # expected str
```

but also
```python
# ERROR sandbox.py:19:28-31: `Literal['b']` is not assignable to TypedDict key with type `T` [[bad-assignment](https://pyrefly.org/en/docs/error-kinds/#bad-assignment)]
d2: TD[str] = {"a": 1, "b": "b"}
```

Needless to say that more subtle inference for `T` does not work either. See the sandbox link for more examples, but for instance

```python
class Foo[T]:
    def __init__(self, b: T):
         self.td: TD[T] = {"a":1, "b":b}

# T # expected str
reveal_type(Foo("b").td["b"])
```

Finally, unrelated to `extra_items` but maybe not unrelated to the core issue is the following

```python
class TD2[T](TypedDict):
    a: int
    b: T

d4: TD2 = {"a":1, "b":"2"}

reveal_type(d4["b"])  # Unknown  # expected str
```

(thank you for the lovely pyrefly tool ✨)

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZVRa4MwFIX_ysU-tIV1ttaHIeyt7HWjy5vKaptQyzYzaroKY_99J3E20aIw9iJi7on33pN89w_IvEDKXiW2ilk6ceglKnXMXowp96w5TFkEp1qq55asG2fciohZxYOUMQvS_iurwwP38I6ozOXpjesaxC4XO1hv14ypIMF6_bgmW3d0F4XzWbiIaMOCjb5JsELDF95oWJKStgsEZ-h8ULn5BRQbircZn9Xh76JQcLtFDMWRJToWpLDyK_EynG2a38DVrX7bfhNdKYAw065ejVUYfiB3trKhC6zzwPxWQ7C9kBTO6JjwRWw2Tae6Q0w_RIXe4aRiKL0W8twVBP0Cg2E3GA5OEq9KvCmKuui0rKtCwktTtJNsU2_i4TbpksK6LZ2aumGdfJduvr8lDWfNQ1cyXFTpVGUirwExoqf6Q3PmMBpBlv0JAxjEuE2KEWavoFypjzLy_UYuj3tfFD6Xu9JvKXzSLMw-s4MZ6-3tSBqgldj2P2Qa12AaD3FprLGEiEHU_AA

_No response_

asukaminato0721:
# Summary

Fixes #4870

Generic `extra_items` now uses the supplied type arguments, including legacy `Generic[T]` syntax.

# Test Plan

add test

### DIFF
--- a/pyrefly/lib/alt/class/class_field.rs
+++ b/pyrefly/lib/alt/class/class_field.rs
@@ -1511,6 +1511,7 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
         let metadata = self.get_metadata_for_class(class);
         if metadata.is_typed_dict() {
             return self.calculate_typed_dict_field(
+                class,
                 metadata,
                 name,
                 range,
--- a/pyrefly/lib/alt/class/class_metadata.rs
+++ b/pyrefly/lib/alt/class/class_metadata.rs
@@ -16,10 +16,10 @@ use pyrefly_python::module_name::ModuleName;
 use pyrefly_python::short_identifier::ShortIdentifier;
 use pyrefly_types::annotation::Annotation;
 use pyrefly_types::callable::Params;
+use pyrefly_types::class::PrecomputedTParams;
 use pyrefly_types::quantified::Quantified;
 use pyrefly_types::quantified::QuantifiedKind;
 use pyrefly_types::type_var::Restriction;
-use pyrefly_types::typed_dict::ExtraItem;
 use pyrefly_types::typed_dict::ExtraItems;
 use pyrefly_types::typed_dict::TypedDict;
 use pyrefly_types::types::Forallable;
@@ -74,6 +74,7 @@ use crate::binding::binding::Key;
 use crate::binding::binding::KeyAnnotation;
 use crate::binding::binding::KeyClassField;
 use crate::binding::binding::KeyDecorator;
+use crate::binding::binding::KeyTParams;
 use crate::binding::binding::ShapedArrayMetadata as ShapedArrayDecoratorMetadata;
 use crate::binding::django::DjangoFieldInfo;
 use crate::binding::pydantic::EXTRA;
@@ -213,6 +214,10 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
             .map(|x| self.parse_base_class(x, is_new_type))
             .collect::<Vec<_>>();
         let contains_base_class_any = parsed_results.iter().any(|x| x.is_any());
+        let is_typed_dict = has_typed_dict_base_class
+            || parsed_results.iter().any(|base| {
+                matches!(base, BaseClassParseResult::Parsed(base) if base.metadata.is_typed_dict())
+            });
         let protocol_metadata = self.final_protocol_metadata(
             initial_protocol_metadata,
             &decorators,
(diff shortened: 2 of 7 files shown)


stroxler (changes requested):
I think this feature is trickier than it looks at first because `extra_items` is simultaneously metadata, but also acts sort of like a class field.

As a result, there are a bunch of issues that pop up which I think may not be handled right now:
- We always have to substitute `T` suitably in inheritance, for example if we have a `Parent[T]` typed dict and we have `Child(Parent[str])` we need to be sure the extra fields behaves as `str` in `Child`. I'm not sure this is true right now.
- We really need to make sure that the behavior where `extra_items` is parsed as an expression where type vars are in scope is restricted to *only* TypedDict, I think that's not the case right now
- We should make sure to at least test what happens if you write `class TD(TypedDict, extra_items=T)` for a bare legacy `TypeVar`. Claude thinks that the current code would make `TD` generic without complaining that it should include a `Generic[T]` base, because the legacy tparam detector will traverse the class keywords. This is *arguably* okay actually, but we should at least document it and show it in a test if we're going to allow it.
- We probably should have tests for functional-form use like `TD = TypedDict("TD", {"a": int}, extra_items=T)` with a raw legacy type var `T`. This probably should just raise a type error and fall back to Any? But in any case we need to define the behavior

Overall I think this is a pretty nontrivial feature. All these complications are related to the problem - which is especially thorny due to legacy-style type vars - that we have something in the class *header* which semantically behaves a lot like a class *field*.

Note that I'm not suggesting we should implement this by making a class field, I'm just noting that for many purposes it should behave a lot *like* a field.




-------


In case it's not clear what I meant by *other* classes with `extra_items`, consider code like this:
```python
class Base:
    def __init_subclass__(cls, extra_items: str = "", **kwargs) -> None: ...
class C(Base, extra_items="hello"): ...
```
We should not be applying TypedDict-specific rules here, `extra_items` is just a vanilla class keyword.

stroxler:
Actually there are a few more things to consider, also related to the unusual semantic status of this where it's both part of the header and somewhat field-like:
- because it's in a class header, using an unquoted forward-ref will crash at runtime. But I think it may be the case that we don't validate for that because we're validating it more like a field, so we get a false negative. Not a huge problem, but a gap if it's true.
- we probably should have more tests of the methods that depend on extra_items like `pop` and such (or at least one test) to make sure everything works. I believe it does work, but a regression test might be handy
- we don't have a test showing a case where a generic typed dict is specialized and an operation *fails*, like `td: TD[str]` followed by `td[extra_key] = 5`. We should show that this correctly errors
- we don't have a test of `ReadOnly` plus generic `extra_items`; I believe it works but we should probably test it

kinto0:
@rchen152 took a stab at it with [this commit](https://github.com/facebook/pyrefly/commit/289ab08234d66a9437ade56d88d527a8e19d4d73?fbclid=IwY2xjawURXG5wZG9mA2V4dG4DYWVtAjExAGJyaWQRMVVuVlBjeFZCeldCdEJvR1VzcnRjBmFwcF9pZAEwAAEe2vKXe_GJZOvueVynJP-KEYKXcB4lf3xp2tgaUQZgZchbFYo7TCxBCd_rwmI_aem_MXWafTQzzGH5kjQdhGPsnQ) and a few follow-ups. closing this PR for now.

PR #4874, opened 2026-09-09
Track literal shapes in array constructors (#4874)

stroxler:
Summary:

Adopt `ArrayCoercible` in the NumPy, JAX, and PyTorch array constructors so ordinary nested scalar literals infer precise shapes. Preserve existing shaped inputs, integer-size `torch.Tensor` construction, and broad gradual fallbacks for unsupported or dynamic inputs.

For example:

```python
assert_type(np.array([[1, 2], [3, 4]]), np.ndarray[[2, 2], Any])
assert_type(jnp.asarray([[], []]), jnp.Array[[2, 0]])
assert_type(torch.tensor([1, 2, 3]), Tensor[[3]])
assert_type(torch.Tensor(2, 3), Tensor[[2, 3]])  # existing size constructor
```

Differential Revision: D119041654

### DIFF
--- a/tensor-shapes/pyrefly-jax-stubs/test/test_creation.py
+++ b/tensor-shapes/pyrefly-jax-stubs/test/test_creation.py
@@ -5,8 +5,10 @@
 
 from __future__ import annotations
 
+from typing import assert_type, TYPE_CHECKING
+
 import jax.numpy as jnp
-from shape_extensions import assert_shape
+from shape_extensions import assert_shape, IntTuple
 
 # A multi-argument `arange` has a length the DSL cannot compute, and a shape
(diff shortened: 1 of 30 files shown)


tensor-shapes/pyrefly-jax-stubs/test/test_creation.py
```diff
@@ -17,6 +19,34 @@
 }
 
 
+def check_array_and_asarray_list_literal_types() -> None:
+    assert_type(jnp.array(1), jnp.Array[[]])
```
jakevdp, line 23: minor (we can fix later if necessary): `jnp.Array` doesn't exist outside the typestubs. `jax.Array` would be more idiomatic.

PR #4878, opened 2026-09-10
Fix `ParamSpec` inference for generic descriptors

### LINKED ISSUE #4844: bug(typechecking): false-positive `bad-argument-type` when resolving a generic descriptor using `Concatenate`
### Describe the Bug

Pyrefly incorrectly rejects instance access through a generic descriptor whose `self` annotation removes its first parameter using `Concatenate`.

### Reproduction

```python

from typing import Any, cast, Concatenate, Protocol


class Descriptor[**P](Protocol):
    def __get__[**P2](
        self: Descriptor[Concatenate[Any, P2]],
        instance: object,
        owner: type,
    ) -> Descriptor[P2]: ...


def descriptor[**P](func: Callable[P, Any]) -> Descriptor[P]:
    return cast(Descriptor[P], func)


class Example:
    @descriptor
    def field(self, value: int) -> int:
        return value


bound = Example().field
```

#### Actual behaviour

```text
Argument `Descriptor[[self: Example, value: int]]` is not assignable to
parameter `self` with type `Descriptor[[Any, value: int]]` in function
`Descriptor.__get__` [bad-argument-type]
```

#### Expected behaviour

No error should be reported. The leading `Any` parameter should match `self: Example`, leaving `P2` as `(value: int)`.

### Weave impact

[Weave uses this pattern in its `Op` protocol](https://sourcegraph.com/github.com/wandb/weave@master/-/blob/weave/trace/op_protocol.py?L73:5-79:100):

```python
@overload
def __get__(self, instance: None, owner: type) -> Op[P, R]: ...

@overload
def __get__(
    self: Op[Concatenate[Any, P2], R],
    instance: object,
    owner: type,
) -> Op[P2, R]: ...
```

For an instance method decorated with `@weave.op`:

```python
import weave


class Example:
    @weave.op
    def traced(self, value: int) -> int:
        return value

    def call(self) -> int:
        return self.traced(1)
```

Pyrefly flags `self.traced`:

```text
No matching overload found for function
`weave.trace.op_protocol.Op.__get__`
called with arguments: (Example, type[Example]) [no-matching-overload]

Possible overloads:
  (instance: None, owner: type[Any])
    -> Op[[self: Example, value: int], int] [closest match]
  (instance: object, owner: type[Any])
    -> Op[P2, int]

Argument `Example` is not assignable to parameter `instance` with type `None`
in function `weave.trace.op_protocol.Op.__get__`
```

### Sandbox Link

https://pyrefly.org/sandbox/?code=GYJw9gtgBA+jwFcAuCQFM5QJYQA5hCSgEMA7UsJYpLMUgZwChHRIoBjMAGy7XZrr0AdMQBG7bHgJEAwsR5jeLcNCQBPXFlIBzSfkJQAgqTUAaDsXpJzMuu2ppSD8wAVwSMJy7NG7LpfooABE0enYQLFwPEABtACo4lwBdAAo3Sk9uAEoALkYoAqgAEzRgWBhtNCQ4eMSAJlT8wub6NC5gHODQ8MjomNtSeyRHBxjjMygXBqTTJuaCrSsydjROsFEAKz5rOfmwAHdSNBBO9Vw0WeasqABaAD4usIioghippM6hL58SspKn3qvBLJFKIQadOQKUS8N7mcZJa73R49F6xZJ5ZroFAgUgWKwpEIA1FvGZQMHsLI+PwBKAAUQAHsQ8LwMYUAAL-FHRXa-MlYNpFFKtdrmABu8gQq2wpCQiIeWiQrPmUCxqFx4q4kp8ojACFIRSgAF46YzmWgUlkhMB+VwikA

---

🤖  I used an AI agent to draft this issue report :)

tobyh-canva:
# Summary

Pyrefly handled some `ParamSpec` argument lists in the wrong direction. This caused Pyrefly to accept valid descriptor access (false-positive `bad-argument-type` / `no-matching-overload`) and accept unsafe callback replacements (false-negative `bad-assignment` / `bad-argument-type`).

Fixes #4844.

### Before

Pyrefly rejected valid instance access for generic descriptors that use `Concatenate`. It did not recognise that descriptor binding changes:

```python
(self, value: int)
```

to:

```python
(value: int)
```

This affected descriptor APIs such as `@weave.op()`.

Pyrefly also accepted unsafe callback replacements. A replacement could remove keyword support from a callback. A later valid keyword call could then fail at runtime.

### After

Pyrefly removes the receiver parameter when it binds these descriptors. It keeps the remaining parameter types, names, and keyword support.

Pyrefly also rejects callback replacements that would make an existing valid call unsafe.

### Typing behaviour

A `ParamSpec` describes function inputs. Input compatibility runs in the opposite direction to return-type compatibility. Pyrefly now uses this direction when it checks a `ParamSpec` inside a generic class.

Pyrefly now infers variance correctly when a generic class refers to itself. An incomplete result from the recursive reference no longer overrides direct evidence from methods and fields.

A writable callable field supports both reads and writes. Pyrefly therefore treats its type parameters as invariant. This prevents an unsafe callback replacement.

# Test Plan
I've added two regression tests covering #4844:
- One replicates the false-positive `no-matching-overload` observed with `@weave.op()`.
- The other is reduced down to a simpler false-positive `bad-argument-type`.

I also added a control test to ensure this fix doesn't over-correct and introduce false-negative errors. That's actually how I found the 2nd bug related to callback replacement 😅 

# Conformance
Resolves 13 outstanding conformance differences (reduced from 42 to 29).

---

I used Codex to help draft this PR :)


tobyh-canva:
According to ChatGPT, all the `mypy_primer` findings are expected. This is what it said:

- Home Assistant’s 19 removed errors are expected. `HassJob` holds a read-only callback, so compatible callback parameters and return types should be accepted. These were false positives. See [`HassJob`](https://github.com/home-assistant/core/blob/dev/homeassistant/core.py) and the [callable assignability rules](https://typing.python.org/en/latest/spec/callables.html).
- hydra-zen’s removed error is also expected. `Partial.__call__` is a writable protocol attribute, so keeping its type variable invariant is correct under [protocol variance rules](https://typing.python.org/en/latest/reference/protocols.html).

kinto0 (changes requested):

kinto0:
thanks! looks great at first glance, but it will take a sec to review it. I'll be splitting it into 3 diffs for review:
bug 1 Example().field triggers Descriptor.__get__: two diffs
- Variance flip (subset.rs, lines 2919–2924)
- Generic residuals are fallback-only (solver.rs, lines 1743–1756)

bug 2 false negative on unsafe callback replacement: one diff

pyrefly/lib/solver/solver.rs
```diff
@@ -555,6 +556,15 @@ enum NewBound {
     AddBound(Type),
 }
 
+/// Whether a bound is a placeholder for deferred generic callable structure.
+fn is_generic_callable_residual(ty: &Type) -> bool {
```
kinto0, line 560: nit: I think we prefer this defined in crates/pyrefly_types/src/callable_residual.rs

kinto0 (approved):
looks great to me! sorry we missed this for 1.3, but I'm sure we'll have a pre-release out soon enough

I'll get another set of eyes from someone who understands variance better than me before merging

I've already split it so aside from anything else I tell you about I'll resolve all follow-ups on our end

stroxler (approved):

kinto0:
rebasing and merging

stroxler:
Hmm. I think this got merged, but the PR linkage was somehow lost so the PR didn't auto-close via ShipIt

Thanks @tobyh-canva!

PR #4891, opened 2026-09-10
fix min(it, key=func) of iterable of list[Any] gets its annotation from func #4886

### LINKED ISSUE #4886: `min(it, key=func)` of iterable of `list[Any]` gets its annotation from `func`
### Describe the Bug

```py
from typing import reveal_type, Any


def f(found: list[Any]) -> None:
    prefix = min(found, key=len)
    reveal_type(prefix)
    print(prefix + "!")
```

```
ERROR sandbox.py:7:11-23: `+` is not supported between `Sized` and `Literal['!']` [[unsupported-operation](https://pyrefly.org/en/docs/error-kinds/#unsupported-operation)]
  Cannot find `__add__` or `__radd__`
 INFO sandbox.py:6:16-24: revealed type: Sized [[reveal-type](https://pyrefly.org/en/docs/error-kinds/#reveal-type)]
```

I find this a bit surprising - `found` is `list[Any]`, so shouldn't `prefix` just be `Any`? Or, the intersection of `Any` and `Sized`?

This is easy enough to workaround by just doing say
```
prefix: str = min(found, key=len)
```

spotted in Hypothesis

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZRNCsIwEIWvMtpFClZnL3gHD1AXwcZa0KSYVvT2vkyiWIVC6XZ-Hsx8M2-CZX75TBH8M6BIF5ufXG-r9G1IHX7Bt1hU88AEV0CU4oIw2e5iLFjFa_uo57E6JSLVJLDCChblEqn_9We0jwFKk8B4wK3uYW_gsSltBmczdO661m-Z3-3uVrOxXLmj50EHU7g0fdeNmOZQjpyci4fsHO4qYldj1FWALhVrbZ8okr9EXMSRGCX8Ag

_No response_

asukaminato0721:
# Summary

Fixes #4886

The change preserves Any from the input instead of replacing it with the callback’s parameter type. 

# Test Plan

add test


rchen152:
I don't think this is the right fix. Distinguishing between implicit and explicit `Any` leads to a couple surprising and unintuitive behaviors:
1. `list` and `list[Any]` (equivalent according to the typing spec) behave differently:
```
from typing import Any, reveal_type
def f(x1: list[Any], x2: list) -> None:
    reveal_type(min(x1, key=len))  # Any
    reveal_type(min(x2, key=len))  # Sized
```
2. An attribute accessed on `Any` behaves differently from `Any`:
```
from typing import Any, reveal_type
def f(x: Any) -> None:
    reveal_type(min(x, key=len))  # Any
    reveal_type(min(x.attr, key=len))  # Sized
```
IMO the issue's other suggestion of `Sized & Any` seems more promising.

PR #4893, opened 2026-09-10
[tensor-shapes] add stubs for jax.numpy structural & logical operations

jakevdp:
## Summary

Add stubs for structural ops (`jnp.split`/`jnp.unstack`) and logic ops (`jnp.isnan`, `jnp.allclose`)

## Test Plan
Updated existing tests; they pass when run locally:
```
343 files already formatted
All checks passed!
Running formatting...
Finished in 2.78 seconds.
Running linting...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.84s
Finished in 1.06 seconds.
$ TENSOR_SHAPES_VENV=.venv uv run python tensor-shapes/pyrefly-jax-stubs/run_pyrefly.py --release
    Finished `release` profile [optimized + debuginfo] target(s) in 0.56s
PASS stubs (9 files)
PASS arithmetic (1 files)
PASS contractions (1 files)
PASS creation (1 files)
PASS fft (1 files)
PASS indexing (1 files)
PASS lax (1 files)
PASS linalg (1 files)
PASS matmul (1 files)
PASS nn (1 files)
PASS reductions (1 files)
PASS reshape (1 files)
PASS searching-sorting (1 files)
PASS structural (1 files)
PASS examples (1 files)
```

## AI disclosure
This PR was created with the assistance of a Gemini coding agent.


tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -1506,6 +1533,64 @@ def dstack(
...
+    *,
+    axis: int = 0,
+) -> tuple[Array[IntTuple], ...]: ...
+def pad(
+    array: Array[Any] | _Scalar,
+    pad_width: Any,
+    mode: str | Callable[..., Any] = "constant",
+    **kwargs: Any,
+) -> Array[IntTuple]: ...
+def repeat(
```
jakevdp, line 1571: We could do better on this via a DSL function in some cases (e.g. integer repeats), but I decided to punt on it since the semantics are somewhat complicated.

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
```diff
@@ -1571,6 +1656,18 @@ def roll[Shape: _Shape](
     axis: Sequence[int] | None = None,
 ) -> Array[Shape]: ...
 @overload
+def rot90[Shape: _Shape, K: Flag[int] = 1](
```
stroxler: I'm not sure this stub is ideal.

I don't think we need an overload - `Flag[int]` can already bind an unknown int, which leads to gradual fallback, so I don't think we need an overload.

Also, do `axes` not affect the behavior?

stroxler (changes requested):
Back to you with one question - I'm not sure about rot90 (an agent caught it, not me)

We could probably just throw a TODO on if we aren't ready to handle `axes`, but I do think the overload is probably unnecessary

tensor-shapes/pyrefly-jax-stubs/jax-stubs/numpy/__init__.pyi
jakevdp: Good catch – I'll update the rule to properly handle `axes`.

Regarding the overload, the main reason for it is to fall back to `m: Array[Any]`

stroxler (approved):

PR #4894, opened 2026-09-10
[tensor-shapes] add stubs for jax.numpy constants, dtypes, etc.

jakevdp:
## Summary

Constants, dtypes, and APIs imported directly from NumPy.


stroxler (approved):

PR #4905, opened 2026-09-11
[pyrefly] Diagnose directory-relative imports

### LINKED ISSUE #4174: Add a diagnostic for implicit relative imports

WilliamK112:
# Summary

Add a dedicated `implicit-relative-import` diagnostic when an unqualified import resolves only through `FallbackSearchPath::DirectoryRelative`. Normal configured roots, build-system sources, typeshed, and site packages retain their existing precedence, while explicitly configured fallback roots remain silent.

The diagnostic is a non-fatal warning by default, so the module continues to resolve while users get guidance to use an explicit relative import or the full absolute package path. This also maps BasedPyright `reportImplicitRelativeImport` settings to the dedicated error kind, updates the generated preset inventory and migration documentation, and covers resolver precedence plus end-to-end reporting.

Fixes #4174

# Test Plan

- `cargo test -p pyrefly_config -- --skip test_find_interpreter_precedence_venv` (the skipped host-dependent test selects the globally installed `/opt/anaconda3` interpreter)
- `cargo test -p pyrefly_config implicit_relative`
- `cargo test -p pyrefly_config test_error_presets_json`
- `cargo test -p pyrefly test::state`
- `cargo test -p pyrefly module::finder::tests::test_`

### DIFF
--- a/crates/pyrefly_config/src/error_kind.rs
+++ b/crates/pyrefly_config/src/error_kind.rs
@@ -213,6 +213,8 @@ pub enum ErrorKind {
     /// only re-exported when redundantly aliased (`from x import y as y`),
     /// listed in `__all__`, or brought in via a wildcard import.
     ImplicitReexport,
+    /// An unqualified import that resolves only relative to the importing file.
+    ImplicitRelativeImport,
     /// An attribute was implicitly defined by assignment to `self` in a method that we
     /// do not recognize as always executing (we recognize constructors and some test setup
     /// methods).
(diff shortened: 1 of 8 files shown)


kinto0 (approved):

kinto0:
looks great to me, but I wonder if this should default to "ignore" instead of "warn". cc @connernilsen for opinions

PR #4906, opened 2026-09-11
Add method assignment diagnostic

### LINKED ISSUE #4161: Feature request: implement `method-assign` from `mypy`

WilliamK112:
## Summary

- add a configurable `method-assign` diagnostic for assignments to class and instance methods, including inherited, static, and class methods
- keep the diagnostic disabled by default and enable it in the `strict` and `all` presets
- preserve ordinary callable data-field assignment and map mypy `method-assign` and strict-mode migration to the new error kind
- document the error kind and preset/migration behavior

## Testing

- `env -u CONDA_PREFIX -u CONDA_DEFAULT_ENV -u VIRTUAL_ENV cargo test -p pyrefly_config`
- `cargo test -p pyrefly test_method_assign -- --nocapture`

Fixes #4161

## AI assistance

Codex assisted with implementation, tests, and documentation. I reviewed the final diff and verified the behavior locally.


crates/pyrefly_config/src/migration/mypy/pyproject.rs
```diff
@@ -482,8 +482,27 @@ allow_redefinition = true
 
     #[test]
     fn test_strict_checks_unannotated_defs() -> anyhow::Result<()> {
-        let cfg = parse_pyproject_config("[tool.mypy]\nstrict = true\n")?;
+        let mut cfg = parse_pyproject_config("[tool.mypy]\nstrict = true\n")?;
         assert_eq!(cfg.root.check_unannotated_defs, Some(true));
+        cfg.configure();
+        assert_eq!(
```
kinto0, line 488: nit: should we remove method assign from this test so it only tests unannotated_defs?

kinto0 (approved):

kinto0:
before merging, would you be able to look into the mypy_primer differences? why do we see such a high number of positives that aren't removing unused-type-ignore errors?

pyrefly/lib/alt/attr.rs
```diff
@@ -1043,6 +1043,18 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
         };
         let (lookup_found, lookup_not_found, lookup_error) =
             self.lookup_attr(attr_base.clone(), attr_name).decompose();
+        if lookup_found
```
kinto0, line 1046: Could we avoid reconstructing method-ness from found_on here? lookup_attr has already applied special lookup semantics, but this helper performs a second, simpler get_class_member lookup that can disagree with the member actually resolved.

For example:
- C.f can resolve a method from C's metaclass, but this checks only C.
- TypedDict methods such as items are resolved through TypedDict fallback, but this returns false for all TypedDicts.
- super() lookup starts after a specific class in the MRO, but this searches from the receiver class, causing both false positives and false negatives when subclasses shadow members.

Could we instead preserve whether the successfully resolved member is a method in the lookup result and diagnose based on that? That would reuse the existing lookup precedence rather than duplicating its AttributeBase1 case split. At minimum, I think we need tests for metaclass methods, TypedDict methods, and shadowing through super().

kinto0 (changes requested):
a few questions actually. might be the cause of the mypy_primer diffs

WilliamK112:
Thanks for the careful review. I pushed fdddaaf: attribute lookup now carries method-ness from the resolved member, including the metaclass, TypedDict, and super() paths, rather than re-looking up from found_on. I added tests for those cases (including both directions of super() shadowing) and separated the strict-mode configuration assertion from the unannotated-definitions test. Locally, the 211 attribute tests, focused configuration tests, and repository formatting/lint check pass.

I also traced the mypy_primer numbers. The refreshed mypy_primer diff has 319 added diagnostics, all [method-assign], and 121 removed diagnostics, all [unused-type-ignore] (the prior run was +318/-118). The primer wrapper (`scripts/pyrefly_primer_wrapper.sh`) explicitly passes `--preset=all` to every `check` on both revisions, so this corpus intentionally surfaces off-by-default diagnostics; the large count does not mean `method-assign` is enabled in Pyrefly’s default preset. Most additions are method monkey-patching/mocking sites. The default severity remains Ignore; strict and all enable it, as discussed on #4161. I will watch the refreshed CI/primer run for any unexpected differences.

WilliamK112:
Updated the branch to current `main` and resolved the `attr.rs` conflict while preserving both sides of the lookup design: the new `Composite`/`Combine` folding remains intact, and each resolved candidate continues carrying method-ness through the actual lookup path. This keeps the metaclass, TypedDict, and `super()` precedence cases from the review fix without reconstructing the result afterward.\n\nValidation on head `b18537596`:\n- `cargo test method_assign` — 7 targeted tests passed, including lookup-precedence and migration coverage\n- `cargo fmt -- --check`\n- `python3 test.py --no-test --no-tensor-shapes --no-conformance --no-jsonschema --no-extension`\n- `git diff --check`\n\nThe merge conflict is resolved and the branch is ready for another review when convenient.

PR #4907, opened 2026-09-11
Test extension binary selection and restart recovery (#4907)

connernilsen:
Summary:

The extension now picks its Pyrefly binary at runtime and restarts the language
server when that choice changes. Neither had a test, and review found real bugs
in both: a failed restart left the launch specification pointing at the binary
that had just failed, so no later change was ever seen as a change.

Two suites cover it. `resolveExecutable` drives the selection table through real
settings, with stubs only for the output channel and the interpreter lookup, so
the fallback reasons that end up in the output channel are asserted directly.
`language server recovery` drives a real `LanguageClient` at a binary that
cannot start, and checks that the specification is put back and a server is
still running afterwards — the second half matters because a client whose start
failed cannot be restarted, only replaced, so a running server is the only proof
the replacement path works.

Both need a real extension host rather than mocks, because the behaviour they
depend on belongs to `vscode-languageclient`. Mocking it would assert our
assumptions about it instead. `activate` therefore returns a small hook so tests
drive restarts directly rather than racing the configuration listener.

Running them per pull request is close to free: `pyrefly.yml` already builds a
release binary, so the new step copies it into `lsp/bin` and runs the suite on
one platform. Until now nothing about the extension ran before merge — not even
type checking.

Reviewed By: kinto0

Differential Revision: D119697409


kinto0 (changes requested):

kinto0 (approved):

PR #4911, opened 2026-09-12
Find SQLAlchemy Mapped model fields in parent classes

### LINKED ISSUE #4903: unexpected-keyword false positive from SQLAlchemy `update().values()` with models that use inheritance
### Describe the Bug

When an SQLAlchemy Mapped field gets added to a model through inheritance from another model, the `update().values()` check doesn't see it and considers it an unexpected keyword.

Example:
```py
from sqlalchemy import update
from sqlalchemy.orm import DeclarativeBase, Mapped

class Base(DeclarativeBase):
    pass

class TableA(Base):
    __abstract__ = True
    field_1: Mapped[int]

class TableB(TableA):
    field_2: Mapped[str]

# This should type-check without errors.
update(TableB).values(field_1=12, field_2="abc")  # E: Unexpected SQLAlchemy update field `field_1`
```

This issue is a problem with the same new check as #4902, but is not related to SQLModel, just base SQLAlchemy, so it seems different enough for its own issue.

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.fZXBboMwDIZfJWovrAeiop2QehjaA-zATmNCGYugWgcdoav29vsTO6XeKq5gO3b8_18Wkem-DubQdPbzJ7r-dIQH_XXI3-kwYhsU8mhhMxAF5xXGQVYeKzbwITrQf0_-xEXl0Vbm2NJT6iG5Dqlr8watYiYvlrDj6L-Aonqb86EvEERwvqhWJFQ01qOk7JIUcYFlhXEpvLgjy7uED9ltM2Zfne2qFTBbrSC8_7pYqyf6oPiKQUQIqj356Yc-rfo1kGtVN01Hl2sd04ex1bbX70PjtMjQylvAfJt9oLksp4agY4eytwV5z_qZ16exPi0VRNfFF1K-Cli2dmZlZEjEFB6mM15OwbwyV2mKduYKbrECc6-c6UH5F-pL5XBvCJFD3RroOSz0appzZ0fLnWya0T_seyO6r1aUBBDJMVgNnLv5gDVbvDhLqRHxrKvJb09kcH8UvWjcXw

_No response_

hrolfurgylfa:
# Summary

This solves #4903 by going through the method resolution order and finding all the Mapped fields on it or its parent classes, instead of just looking at the target class itself.

Fixes #4903

# Test Plan

Added tests with both a parent class that is an SQLAlchemy model and one that isn't, and thus has to use it as a Mixin to get its Mapped fields.

### DIFF
--- a/pyrefly/lib/alt/expr.rs
+++ b/pyrefly/lib/alt/expr.rs
@@ -1554,10 +1554,21 @@ impl<'ctx, 'answer, Ans: LookupAnswer> AnswersSolver<'ctx, 'answer, Ans> {
     }
 
     fn sqlalchemy_mapped_model_fields(&self, model: &Class) -> SmallSet<Name> {
-        self.get_class_field_map(model)
-            .into_iter()
+        let mut seen_fields = SmallSet::new();
+
+        std::iter::once(model)
+            .chain(
+                self.get_mro_for_class(model)
+                    .ancestors_no_object()
+                    .iter()
+                    .map(|ancestor| ancestor.class_object()),
+            )
+            .flat_map(|class| self.get_class_field_map(class))
             .filter_map(|(name, field)| {
-                if name.as_str().starts_with('_') || field.is_class_var() {
+                if !seen_fields.insert(name.clone())
+                    || name.as_str().starts_with('_')
+                    || field.is_class_var()
+                {
                     return None;
                 }
                 let (_, annotation, _) = field.for_variance_inference();
--- a/pyrefly/lib/test/descriptors.rs
+++ b/pyrefly/lib/test/descriptors.rs
@@ -1019,6 +1019,33 @@ update(User).values(nam="alice")
 );
 
 testcase!(
+    test_sqlalchemy_update_values_checks_inherited_mapped_fields,
+    sqlalchemy_mapped_env(),
+    r#"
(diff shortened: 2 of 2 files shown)


rchen152 (changes requested):
Thanks! This looks close, but I believe the MRO walk is backwards. I'd expect it to start with `model`, then go through `ancestors_no_object` in the regular direction, and collect only the names whose *first* appearance is a `Mapped` field (to handle attribute overrides).

hrolfurgylfa:
> Thanks! This looks close, but I believe the MRO walk is backwards. I'd expect it to start with `model`, then go through `ancestors_no_object` in the regular direction, and collect only the names whose _first_ appearance is a `Mapped` field (to handle attribute overrides).

Thanks for the review, yep, forgot to consider what happens with overwritten fields. I've now added that to the test, reversed the mro and modified the walk to ignore fields it has already seen.

I also did some cleanup, found a way to skip the loop within loop with `flat_map`, and fixed the test, I was filtering on `User.id` for the `AdminUser`. I think that should still be valid SQLAlchemy, but it is very weird.

rchen152 (approved):

rchen152:
LGTM, thanks!

PR #4912, opened 2026-09-12
fix(torch-stubs): declare torch.device as a context manager

yeetypete:
# Summary

Since PyTorch 2.0, `torch.device` can be used as a context manager to set the default device for a block. The shape stubs' `device` had no `__enter__`/`__exit__`, so every `with torch.device(...)` block reported `bad-context-manager`. We add them here, matching  the PyTorch stubs: 

https://github.com/pytorch/pytorch/blob/d6c03540dc0f040e1d8a920d19d9333de6513545/torch/_C/__init__.pyi.in#L121-L122

# Test Plan

Run `test.py`.


stroxler (approved):

PR #4914, opened 2026-09-12
fix(torch-stubs): add nn.Module.extra_repr stub

yeetypete:
# Summary

We add a stub for `nn.Module.extra_repr`, which was previously missing. Without it, overriding `extra_repr` in a subclass reported `bad-override`. See
https://github.com/pytorch/pytorch/blob/d6c03540dc0f040e1d8a920d19d9333de6513545/torch/nn/modules/module.py#L2997

# Test Plan

Run `test.py`.


stroxler (approved):

PR #4915, opened 2026-09-12
feat: let ProxyMethod resolve callable attribute targets

yeetypete:
# Summary

Let `ProxyMethod` resolve a target declared as a class attribute, in addition to an ordinary instance method. A target is accepted whenever pyrefly can call it: `Callable[...]` annotations, callback protocols, and `type[X]` constructors all work but non-callable targets such as `forward: int` are still rejected.

This lets the shape stubs declare `nn.Module.forward` as `Callable[..., Any]`, matching PyTorch:

https://github.com/pytorch/pytorch/blob/d6c03540dc0f040e1d8a920d19d9333de6513545/torch/nn/modules/module.py#L526

The stubs previously declared `forward` as a method because that was the only target `ProxyMethod` could resolve. Under `strict-callable-subtyping`, a method-form `forward(*args, **kwargs)` makes every concrete `forward(self, x)` override a `bad-override`.

# Test Plan

Run `test.py`.


stroxler (approved):

stroxler:
Thanks! I wasn't aware of this pattern using an annotation for `forward` in some modules when I limited it to methods.

I think it needs a rebase before we can merge though; I think it may have collided with another of your PRs as well as an improvement I made to the pyrefly-torch-stubs pyrefly.toml today

yeetypete:
Fyi I had to enable strict callable subtyping for the entire "torch-positive" stubs test suite. Without strict callable subtyping a narrowed override of forward() gets accepted regardless of how it's declared in the stub.

It seems that passing pyrefly cli args is preferred over using a `pyrefly.toml` file in `shape_testing.py` since other configuration is passed as cli args and the default is to pass `os.devnull` for the pyrefly config. But if I do this then any settings I pass will apply to the whole suite. Lmk if you'd like me to rework this (passing the config with a scoped enabling of strict callable subtyping instead of the arg) or have some concerns.

stroxler:
This seems good to me. I think it's expected that CLI arguments take priority, they are intended to override the pyrefly config.

I'm not too worried about the test suite organization because I'm actually planning to overhaul the entire test suite over the weekend - I want it structured much more like the numpy and jax suites, with things organized by API surface area and most tests running as dual runtime + static so that we can actually validate the behavior.

I'm fairly sure there are a lot of (mostly minor) bugs in the torch stubs right now, partly because the tests predate some improved validation patterns that I adopted for numpy and jax.

It's good to get this fix in first, and just make sure the overhaul keeps things working :)

PR #4916, opened 2026-09-12
fix(torch-stubs): return torch.return_types named tuples

yeetypete:
# Summary

Add a stub for `torch.return_types` and use it for the operations that return named tuples. Currently `sort`, `max(dim=...)`, `topk`, `aminmax`, `slogdet`, etc. return types are annotated as plain `tuple[Tensor, Tensor]` and `.values`, `.indices`, `.sign`, and the other documented fields are reported as `missing-attribute` errors.

PyTorch adds these annotations by generating its own `return_types.pyi` from the output names in `native_functions.yaml`:

https://github.com/pytorch/pytorch/blob/d6c03540dc0f040e1d8a920d19d9333de6513545/torchgen/api/python.py#L1012-L1033

For simplicity we just write the stubs manually here instead of generating them.

Note: `linalg_slogdet` is included as well although its missing in the PyTorch generated stubs.

# Test Plan

Run `test.py`.

### DIFF
--- a/tensor-shapes/pyrefly-torch-stubs/test/negative_tests/test_return_types_errors.py
+++ b/tensor-shapes/pyrefly-torch-stubs/test/negative_tests/test_return_types_errors.py
@@ -1,0 +1,19 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Test that `torch.return_types` tuples expose only their documented fields."""
+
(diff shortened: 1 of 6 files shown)


stroxler:
Thanks @yeetypete!

Something seems stuck here - auto-import actually failed, and I can't seem to import manually either. Maybe a rebase would help, but another option might be to just close this and open a fresh PR. What ought to happen is that a diff number should usually appear a few minutes after a PR is submitted

yeetypete:
Apologies for the delay. Rebased and looks like the import worked now :)

kinto0 (approved):

PR #4921, opened 2026-09-14
Use --skip-interpreter-query inside test cases

oprypin:
# Summary

Use --skip-interpreter-query inside test cases

This is done to allow the test to still work in case there isn't a Python binary available.

(There is no associated issue for this fix)

# Test Plan

Changing only test code, so N/A.

I was just running tests in my environment that doesn't have a python executable and noticed an issue here, so sending a fix accordingly.


rchen152:
Thanks!

rchen152 (approved):

Issue #4926, opened 2026-09-14
incomplete completions for django annotate


kinto0:
### Describe the Bug

reported here:
https://www.reddit.com/r/django/comments/1we6roj/comment/p9kqzpl/

```
from django.db import models

class Article(models.Model):
    title = models.CharField(max_length=100)

for article in Article.objects.annotate(extra_title=models.F("title")):
    article.
#           ^
```

PR #4927, opened 2026-09-14
Capture Django annotate completion gap

kinto0:
Summary:
Add a passing LSP characterization test showing that model fields are offered on an item from `QuerySet.annotate()`, while the annotation-defined attribute is currently omitted. This pins the reported Django IDE gap without changing production behavior.

reproduces https://github.com/facebook/pyrefly/issues/4926

Differential Revision: D119989910


rchen152 (approved):

PR #4930, opened 2026-09-14
[tensor-shapes] define and use ArrayLike in JAX APIs

jakevdp:
## Summary

Use the new IntTuple default approach to define and use `ArrayLike` for jax APIs.

This lets us cover a wider range of valid inputs for each function with fewer overloads.


stroxler (approved):

PR #4931, opened 2026-09-14
Fix CI from bad function argument name in LSP typescript file

connernilsen:
Summary:
Oops https://github.com/facebook/pyrefly/actions/runs/34889191287/job/104127549609

Added a fix, but it looks like I folded it into the diff above this one on accident.

Differential Revision: D120033428


kinto0 (approved):

PR #4934, opened 2026-09-14
Report invalid property signatures with extra required parameters

### LINKED ISSUE #4918: no error on invalid usage of `@property`
### Describe the Bug

```py
class Foo:
    @property
    def value(self, huh: str) -> int: # no error
        return 1
```
i initially suspected that this was a typeshed issue (https://github.com/python/typeshed/issues/4731), but `property`'s constructor looks like this:

```py
def __init__(
    self,
    fget: Callable[[Any], Any] | None = None,
    fset: Callable[[Any, Any], None] | None = None,
    fdel: Callable[[Any], None] | None = None,
    doc: str | None = None,
) -> None: ...
```
so it should report an error that the provided `fget` function takes the wrong number of arguments (ie. should not require any special-casing). however i think fixing the case where the parameter has a default value would require special-casing (see https://github.com/astral-sh/ty/issues/4522)

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.w1VkQrKKW36-VQw0QzkUFOUXAEuhSpgAKJbBiQOa1DNKM8D5DxzpwJiwUlBWyMuHBCZMD1K6MiQ2AE1A_i0oSi1OLQELAeMhBr-nAA

_No response_

dedsec-terminal:
# Summary

The property decorator path attaches metadata without checking the decorated function against the fget/fset/fdel shapes declared on property in typeshed, so a getter like def value(self, huh: str) passed silently.

Validated required parameters when the decorator is applied: one for getters and deleters (self), two for setters (self, value). Extra required positional or keyword-only parameters now report an error. Parameters with defaults stay allowed since they remain callable, and static/class methods plus top-level functions are untouched.

Fixes #4918

# Test Plan

4 new cases in pyrefly/lib/test/descriptors.rs (getter, setter, deleter errors, defaulted-extra clean). Ran descriptors (67 passed), decorators (62 passed), property filter (46 passed, 1 pydantic case needs PYDANTIC_TEST_PATH env), cached_property (3 passed). cargo fmt, check and clippy clean with no new warnings.


stroxler (approved):

PR #4938, opened 2026-09-15
fix [Inlay hints] Wrong call argument name for loguru.Logger #4936

### LINKED ISSUE #4936: [Inlay hints] Wrong call argument name for loguru.Logger
### Describe the Bug

Image

The first call argument should exclude self (the `Logger` instance).Is this issue caused by the non-standard naming of self (`__self`)?

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4936

For a bound method whose receiver is named `__self`, positional inlay hints now start with the first actual argument.

also dedup a fn

# Test Plan

add test


stroxler (approved):

anci3ntr0ck:
The current method for determining actual parameters based on the receiver's name seems insufficiently intelligent, as it is impossible to anticipate all non-standard naming. I would like to know whether judging based on whether the function is located within a class and the presence of annotations like `@classmethod` and `@staticmethod` can cover most cases.🤔

PR #4940, opened 2026-09-15
fix(torch-stubs): re-export torch submodules as attributes

gaborv:
# Summary

`torch/__init__.py` imports its public submodules (`from torch import (... cuda as cuda, ...)`), and `torch/nn/__init__.py` does the same for `attention`, `modules`, `parallel`, `parameter` and `utils`. That is why `torch.cuda.is_available()` works without `import torch.cuda`. The shape stubs did not re-export them, so every such use reported `implicit-import`.

This mirrors the runtime import lists:
- [`torch/__init__.py`](https://github.com/pytorch/pytorch/blob/cf30153c4c131c8164ee7798e5022d810682e2cb/torch/__init__.py#L2298)
- [`torch/nn/__init__.py`](https://github.com/pytorch/pytorch/blob/cf30153c4c131c8164ee7798e5022d810682e2cb/torch/nn/__init__.py#L9)

# Test Plan

Added `test/test_submodule_attributes.py`. Its uses report `implicit-import` before the change and nothing after. That is a warning, which the runner does not fail on, so I checked it with `--min-severity warn`. `run_pyrefly.py` passes all suites.


stroxler (approved):

stroxler:
Thanks! I had been meaning to do this but kept running out of time. This should be a good usability win

PR #4941, opened 2026-09-15
fix(torch-stubs): accept bare `@torch.no_grad`

gaborv:
# Summary

`torch.no_grad` inherits [`_NoParamDecoratorContextManager.__new__`](https://github.com/pytorch/pytorch/blob/cf30153c4c131c8164ee7798e5022d810682e2cb/torch/utils/_contextlib.py#L159-L170), which returns the function when called with one, so `@torch.no_grad` works without parentheses. The stub declared only a no-argument `__init__`. `@torch.no_grad` therefore reported `bad-argument-count`, and every call of the decorated function lost its signature (`not-iterable`, `unexpected-keyword`, …).

This declares the same two `__new__` overloads PyTorch does.

# Test Plan

Extended `test/test_no_grad_decorator.py` with a bare-decorator case. It fails before the change and passes after. `run_pyrefly.py` passes all suites.


stroxler (approved):

PR #4942, opened 2026-09-15
fix(torch-stubs): accept `torch.where(condition)` and keyword arguments

gaborv:
# Summary

PyTorch's generated `_VariableFunctions.pyi` declares these `torch.where` overloads:
- `where(condition) -> tuple[Tensor, ...]`, equivalent to `nonzero(as_tuple=True)`;
- `where(condition, input: Tensor, other: Tensor, *, out=None)`;
- `where(condition, input: Tensor, other: Number)`;
- `where(condition, self: Number, other: Tensor)` and `where(condition, self: Number, other: Number)`: a scalar value parameter is named `self`.

The stub had only the three-argument forms, with parameters named `x` and `y`. So `torch.where(mask)` and `torch.where(mask, a, other=b)` reported `no-matching-overload`.

This mirrors PyTorch's overloads. The result shape stays gradual, as before.

# Test Plan

- Extended `test/test_top_level_gaps.py` with the single-argument form, `other=` and `input=` keywords, and `out=`; these fail before the change.
- Added `test/negative_tests/test_where_keyword_errors.py` for two forms PyTorch rejects at runtime: `input=` with a scalar value, and `out=` with a scalar value. I checked both against torch 2.13.
- `run_pyrefly.py` passes all suites, including the existing positional `where` tests.


stroxler (approved):

PR #4943, opened 2026-09-15
Collect SQLAlchemy fields from SQLModel classes with or without Mapped

### LINKED ISSUE #4902: `unexpected-keyword` false positives from the SQLAlchemy `update().values()` check on SQLModel models with `Mapped[]` relationships
Since 1.3.0, the `update(Model).values(field=...)` check from #4047 reports every plain column of a SQLModel table model as "Unexpected SQLAlchemy update field" if the model has at least one relationship annotated with `sqlalchemy.orm.Mapped[...]`.

The check takes the model's columns to be the attributes annotated `Mapped[...]`, and skips the model if there are none. That works for `DeclarativeBase` classes, where every column is `Mapped[...]`. In SQLModel, columns are Pydantic fields (`name: str = Field(...)`) and `Mapped[...]` only appears on `Relationship()` attributes. So the check sees only the relationships as columns and reports every real column as unknown.

SQLModel accepts `Mapped[...]` on relationships, though its docs do not use it. `SQLModelMetaclass.__init__` (`sqlmodel/main.py`, since fastapi/sqlmodel#700) unwraps a `Mapped[...]` annotation if present, otherwise it rewrites the annotation to `Mapped[ann]` itself; either way it builds the `relationship()` from the inner type, so both spellings produce the same mapper. `selectinload(Model.rel)` only type-checks with the explicit form: `Mapped.__get__` returns `InstrumentedAttribute[_T]` for class-level access, while with a plain annotation `Model.rel` has the instance type and `selectinload` rejects it.

## Reproduction

`pyrefly==1.3.0`, `sqlmodel==0.0.42`, `sqlalchemy==2.0.52`, Python 3.13.

```python
from typing import Optional
from uuid import UUID

from sqlalchemy import update
from sqlalchemy.orm import Mapped
from sqlmodel import Field, Relationship, SQLModel


class Node(SQLModel, table=True):
    id: UUID = Field(primary_key=True)
    name: str = Field(description="x")
    parent_id: UUID | None = Field(default=None, foreign_key="node.id")
    parent: Mapped[Optional["Node"]] = Relationship()


update(Node).values(name="a")        # real column, false positive
update(Node).values(parent_id=None)  # real column, false positive
update(Node).values(nope="a")        # wrong, the check catches it
```

```
ERROR Unexpected SQLAlchemy update field `name` [unexpected-keyword]
  --> repro.py:16:21
ERROR Unexpected SQLAlchemy update field `parent_id` [unexpected-keyword]
  --> repro.py:17:21
ERROR Unexpected SQLAlchemy update field `nope` [unexpected-keyword]
  --> repro.py:18:21
```

Removing the `parent` relationship, or its `Mapped[...]` wrapper, makes all three lines pass, including the wrong one, because the model then has no `Mapped[...]` attributes and the check skips it.

1.2.0 reports nothing for this file.

## Expected behavior

Either

- skip classes whose MRO contains `sqlmodel.SQLModel`, since the check cannot read their columns from `Mapped[...]` annotations, or
- use the Pydantic fields as the columns for SQLModel classes, which keeps the `nope` error and removes the false positives.

There is no config switch for this check and it uses the generic `unexpected-keyword` kind, so staying on 1.3.0 means a suppression at every `update().values()` call on such a model (31 sites in our codebase).

# Sandbox Link

[Link here](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0B5Ygwi50qKAB100sNXpwAjlAkBjABYwapTt178ArsUyoGMWfLpKVUDVtKFe9Ljz50AsqmLEYmC7StlGlxMGChdV34AMQgwzAAaOgAlMNNRdDh1CGJEgGUARQAZdxCw6XL0VRU4ODoAOVKACgLi0qhEhlRsWABeABVKAxgASkRpOgnOTEROdH4euhi4xuJKLlRKUgB9AGsYUn7BkfHJ8RoYGbgGSjoFpahMRtC4VTXhdJ7JEHwv4ZOJ4gbGBzLYQaazfgAH3qYhgt0WsQeTxgYFQBigDB6DXQMESYF4MAgbHQu32nxA6FKhDBv3+dEBlGBDBmnm8vgA2kIRGIJOyvg1Ql8ALpC+EpFTcjJZYiNP4yeVGExmRoCkaEABuEiGcEaZxg5NQv0mxroAGI6IyJHRVLgoAYaOg8RI4HCeHAICJ1eZ0IrTDAVaVhhqtfAVkCQWCsbDhhNzZbwja7Q6nVAXfTcO7Pd7fcrVUHNXbQ5SfAajSbY3QAO7UdgdTTWzSqHbW0x2Woe6QgeIgaxqTTaAD0W1B6A9w5I5EQICqqBqdAAqsY-WN0MbQmAq5pGY0XVAwIkAFSvD0wNaoGa4bAAKxgqgYMYAtAA+OhfRdK8xIOiEH909d0AttR3MJ9zoA8Dx2SsNjYOAL2vW97zoZ9XxAd8-S+GYf0ICppH-HN-U6boLjoS8bzvR8XzQsxMN-dAux7ZQ+3sAcnCHEcxy2CcKBnOdWR8TB2T6IUVzXFE6GHNgYAYYdgL3RJWCuDBVGI0iEMSXBKxxSg4LIxDkL6GjsNXSZ-2HF1pK2WTQIUzpKhU+C70SQDiL6CiYRxQzO27axglCKA2NYDiuKnOQAiYFh2AifQ6AAQXQUgcMqapahaEo-JEkyxOHQLpLgAxsB4uAZKqOAOi6WAZjwW14SiZ0RiQl9sWIrDEv-e5HgPaDYNi+LDwgqDKBgmY4tINyRs8rAxPFNIxEybJGk6wbupGvrIK64b4rG+KJvoshGTAKAHAYWgoAoMgGHUMQHy9Sh3TEeEvgAZkIABGR6vk7ABfbtUDvCAvRiWAKDQLA8CIMgQE+oA)

hrolfurgylfa:
# Summary

Modified `sqlalchemy_mapped_model_fields` to also collect model fields supported by SQLModel. SQLModel doesn't require Mapped annotation, but does allow it, so previously it was showing up with errors on fields that weren't using Mapped if some other field used Mapped.

I ended up going a bit further then I expected at first in order to follow SQLModel's rules on DB fields. A SQLModel class only gets turned into a table if a class has a `table=True` keyword, and that `table=True` can only show up on one table, on the last one specifically [(docs)](https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models/?h=inheritance#only-inherit-from-data-models). This isn't just a recommendation, I tried creating a model with table inheritance and it fails:

```py
from sqlmodel import Field, Session, SQLModel, create_engine

class HeroBase(SQLModel):
    name: str = Field(index=True)
    secret_name: str
    age: int | None = Field(default=None, index=True)

class Hero(HeroBase, table=True):
    id: int | None = Field(default=None, primary_key=True)

class EvilHero(Hero, table=True):
    evil_amount: int

engine = create_engine("sqlite:///database.db")
SQLModel.metadata.create_all(engine)
with Session(engine) as session:
    EvilHero(name="Doctor Evil", secret_name="John", age=78, evil_amount=3)
    >>> AttributeError, from inside SQLAlchemy
```

As I mentioned on the issue:
> It still feels a bit weird to be adding this additional logic for other libraries though, maybe we want to just ignore this check entirely for SQLModel classes instead? I'm not sure

Even after doing this, I'm still not fully convinced this is the right route.

Also, I removed a now unnecessary dead code lint ignore, since I am now calling said function from this code. I think it was being called already from one other place though, but the lint had yet to be removed. If this should be done in a separate PR instead I can go ahead and remove it.

Fixes #4902

# Test Plan

Added automated tests and ran them, I also did some messing around with the Pyrefly LSP in a test SQLModel project, just to see if I could find anything else I missed with the tests.


stroxler (approved):

PR #4944, opened 2026-09-15
[tensor-shapes] allow ArrayLike in jax linalg & fft stubs

jakevdp:
- Switch from `_Array` to `_ArrayLike` for input arguments
- Consolidate fft shape rules
- Additional tests for NumPy inputs to some APIs
- Other minor cleanups


stroxler (approved):

PR #4945, opened 2026-09-15
[tensor-shapes] add JAX nanogpt example

jakevdp:
Also update JAX bert example to use matching conventions.


stroxler (approved):

PR #4947, opened 2026-09-16
Test binary selection in a multi-root workspace (#4947)

connernilsen:
Summary:

The extension picks one Pyrefly for the whole window, from the interpreter
selected for the first workspace folder. Until now that was only asserted
indirectly — the existing tests run in a single-root window, so they can check
that exactly one interpreter lookup happens and that it is scoped to folder
zero, but with one folder open "folder zero" and "the only folder" are the same
thing. A regression that resolved per folder, or against the wrong folder, would
not have been caught.

This adds a second `.vscode-test.mjs` configuration that opens a two-folder
fixture workspace, and a suite that runs only there. `workspaceFolder` is handed
straight to the VS Code CLI as the path to open, so pointing it at a
`.code-workspace` file gives a real multi-root window.

A second configuration rather than changing the existing one: every other test
is written against a single root, and moving the whole suite into a multi-root
window would change the environment for activation, infer and the restart tests
to no purpose. The two file globs are disjoint, so neither set runs twice.

Reviewed By: kinto0

Differential Revision: D120368744


kinto0 (approved):

PR #4951, opened 2026-09-16
[tensor-shapes] use ArrayLike for jax.lax and jax.nn APIs

jakevdp:
Update APIs to generally accept `ArrayLike[Shape]` rather than `Array[Shape]`.


stroxler (approved):

PR #4952, opened 2026-09-16
Prefer type parameter defaults to contextual hints

stroxler:
Summary:
**What**

Treat type parameter defaults as operational type information rather than a fallback after contextual return inference. During contextual call matching, use isolated variables for defaulted parameters so context can still solve non-defaulted parameters without overriding defaults. Explicit call arguments continue to infer the real parameters before defaults are applied.

**Why (general argument)**

I think as a rule `default = ...` values for type variables are intended to describe actual operational semantics: if I pass a set of parameters that doesn't otherwise determine the generic behavior, I'm going to get some specific instantiation out. Operationally, the values always bubble out: there's no such thing as "contextual function call" in Python.

As a result, I think using contextual typing to override a default is probably almost always the wrong thing - we're effectively using a time-traveling static "guess" to override the actual operational semantics the author indicated. This is probably a bug, in the sense that we likely guess wrong and infer the wrong type.

**Why (specific)**

This came up in tensor shapes: jax (and also numpy and torch) make heavy use of functions that automatically promote a scalar input to an empty-shape array, which has specific broadcasting properties that make this work really well.

We have modeled this using defaults, which correctly describes the operational semantics. But Pyrefly prefers the contextual hint to the default, which leads to this specific instance of the general issue I described above... in
```
def g(Array[[2, 2]]) -> None: ...

def f[Shape: IntTuple = []](x: Array[Shape] | scalar) -> Array[Shape]: ...

g(f(1))
```
trunk Pyrefly incorrectly (relative to the known right answer for this code) says that `f(1)` returns `Array[[2, 2]]`, effectively saying we think `f` magically knows it's called inside `g` whereas in reality `f(1)` is an `Array[[]]` and this should be a type error

Differential Revision: D120381571


samwgoldman (approved):

PR #4954, opened 2026-09-16
Update Github CI to run new extension tests (#4954)

connernilsen:
Summary:

I want to make sure extension tests run when the extension changes. This adds two changes:
1. makes `test_extension` run whenever someone edits anything under `lsp/`
2. adds a few quick tests to `pyrefly.yml`, which runs on each PR

Differential Revision: D120397203


kinto0 (changes requested):

PR #4957, opened 2026-09-16
Add test when Pyrefly extension fails to start (#4957)

connernilsen:
Summary:
Add a test validating behavior when the Pyrefly extension fails to start. 

Differential Revision: D120419312


kinto0 (approved):

PR #4963, opened 2026-09-17
Trigger website CI on website workflow changes

stroxler:
Summary: The website build workflow only runs on PRs touching website/**, so edits to the workflow file itself never get validated. Include the workflow file in its own paths filter.

Differential Revision: D120569608


connernilsen (approved):

PR #4970, opened 2026-09-18
fix [tensor-shapes] Update numpy stubs to use Shape: IntTuple = [] where suitable #4966

### LINKED ISSUE #4966: [tensor-shapes] Update numpy stubs to use `Shape: IntTuple = []` where suitable
### Describe the Bug

We recently both added syntax hooks and tweaked the solver so that
```
def f[Shape: IntTuple = []](x: ndarray[Shape] | float) -> ndarray[Shape]: ...
```
can now bind `Shape` to the empty `IntTuple` (rather than leaving it unsolved and falling back to a gradual shape) when `x` fails to constrain it.

This feature was specifically requested by @jakevdp because by using it, the jax stubs are able to mark a large number of functions that automatically treat plain scalar inputs as if they were empty-shape arrays (for example in broadcasting) as using this default, in the process greatly reducing the number of overloads.

This has saved us well over 1000 lines of stubs in pyrefly-jax-stubs, plus avoiding overload resolution is better for performance. See, for example https://github.com/facebook/pyrefly/pull/4930/changes which was a massive simplification.

We should do the same in numpy (and eventually in torch as well) where applicable.

asukaminato0721:
# Summary

Fixes #4966

simplify NumPy stubs：use OtherShape: IntTuple = [] to merge calculation, compare and bit calculation's scalar / array signature, reduce dup overload.

# Test Plan

add test


stroxler (approved):

PR #4974, opened 2026-09-18
fix match-case type narrowing across branches #4972

### LINKED ISSUE #4972: match-case type narrowing across branches
### Describe the Bug

We often use match-case statements to check exactly one of two values is populated. In this case, pyrefly is not able to type narrow the return types to be non-null. I've checked this on pyrefly v1.3.0

```python
def get_present_value(
    left: int | None,
    right: str | None,
) -> int | str:
    match left, right:
        case None, None:
            raise ValueError("Both values are missing")
        case value, None:
            return value
        case None, value:
            return value
        case _:
            raise ValueError("Both values are present")
``` 

### Sandbox Link

https://pyrefly.org/sandbox/?project=v2.pZTPCsIwDMZfJbDDiogB8bSn8ORJkDLL5sFVbJXu7U36B9c5iuKxbb40bX75frDMzGcWzlcJYL5auIbak6hdw-hHMy4TZVdp2x6Eo-MpeGRwCkTQTcNz7g7MpQePqci1lHFR-H6BcB-iWGhJNM5Fpy9KI-b4Q3bbWJRv3gydCvZhA2IXyDSJue5B1kwsbY5DRa6soLf2ZhrEJNf3DtWAZ90azBQIPCXyKS_e8PN0oD3qhtL-w2wdkK1LxNYMLEUUIXwB

_No response_

asukaminato0721:
# Summary

Fixes #4972

Tuple matching now preserves exclusions from earlier cases, so the reported captures narrow to int and str without None.

# Test Plan

add test


stroxler (approved):

PR #4975, opened 2026-09-18
Match JAX single-file module layout in pyrefly-jax-stubs

QEDady:
In JAX, `jax.lax.linalg` (`jax/lax/linalg.py`), `jax.numpy.fft` (`jax/numpy/fft.py`), and `jax.numpy.linalg` (`jax/numpy/linalg.py`) are single-file modules rather than `__init__.py` packages. Rename their stubs in `jax-stubs/` from `__init__.pyi` to `.pyi` to match the runtime module structure (and the convention used in `torch-stubs/fft.pyi` and `torch-stubs/linalg.pyi`).


stroxler (approved):

PR #4982, opened 2026-09-18
add regression test for < is not supported between ndarray[IntTuple, dtype[float64]] and float #4917

### LINKED ISSUE #4917: `<` is not supported between `ndarray[IntTuple, dtype[float64]]` and `float`
### Describe the Bug

The following is perfectly cromulent Python:
```
import numpy as np


rng = np.random.default_rng(7)

# ---------------- parameters ----------------
N       = 300
L       = 20.0          # world size (periodic)
R       = 1.5

pos   = rng.uniform(0, L, (N, 2))

d = pos[:, None, :] - pos[None, :, :]
d -= L * np.round(d / L)
vals = (d ** 2).sum(-1) < R * R

print(vals)
print(vals.shape)
```
Running `pyrefly check` over this produces two errors:
```
ERROR `<` is not supported between `ndarray[IntTuple, dtype[float64]]` and `float` [unsupported-operation]
  --> typing_test.py:15:8
   |
15 | vals = (d ** 2).sum(-1) < R * R
   |        ----------------^^^-----
   |        |                  |
   |        |                  has type `float`
   |        has type `ndarray[IntTuple, dtype[float64]]`
  Argument `ndarray[IntTuple, dtype[float64]]` is not assignable to parameter `value` with type `float` in function `float.__gt__`
ERROR Object of class `bool` has no attribute `shape` [missing-attribute]
  --> typing_test.py:18:7
   |
18 | print(vals.shape)
   |       ^^^^^^^^^^
 INFO 2 errors
```
Both of these seem to stem from Pyrefly not being able to correctly infer NumPy broadcasting with `<` between an `ndarray` and a `float`. Using Pyrefly 1.3.0 with `pyrefly-numpy-stubs` and Python 3.14.

### Sandbox Link

_No response_

_No response_

asukaminato0721:
# Summary

Fixes #4917

# Test Plan

add regression test


stroxler (approved):

PR #4984, opened 2026-09-18
[tensor-shapes] annotate jax device & sharding arguments

jakevdp:
This is not related to shape propagation, but will properly validate `device` and `out_sharding` arguments when these stubs are used.


stroxler (approved):

PR #4992, opened 2026-09-20
Generate `# pyrefly: ignore[code]` without a space before the bracket

### LINKED ISSUE #4991: Docs and `pyrefly suppress` use different spacing in ignore comments
The docs show `# pyrefly: ignore[bad-return]` (no space), but `pyrefly suppress` generates `# pyrefly: ignore [bad-return]` (with a space).

Both forms work, but the docs never say the space is optional, so the mismatch looks confusing.

Suggested fix: maybe we can just add a note like this after https://github.com/facebook/pyrefly/blob/ed01fb5dc390b0be43072ba9481e5d42837fc802/website/docs/error-suppressions.mdx?plain=1#L90

```
The generated comments use the form `# pyrefly: ignore [error-code]`, with a space before
the bracket. This is equivalent to the `# pyrefly: ignore[error-code]` form shown above, and
Pyrefly accepts both.
```

happy to send a PR, if this looks correct or please let me know if it needs changes. Thanks :)

Aniketsy:
# Summary

`pyrefly suppress` were adding comments like `# pyrefly: ignore [code]` (with a space), but the docs use `# pyrefly: ignore[code]` (no space). This PR makes both generate the no-space style, as discussed in the issue.

comments with a space still work, nothing changes for existing code.

Fixes #4991 

# Test Plan

- Updated the existing tests and added a few new ones to check that comments with a space are still accepted.
- Ran `cargo test -p pyrefly suppress` and `cargo test -p pyrefly code_actions`.


stroxler (approved):
LGTM, thanks!

I'll aim to get a second review and merge it soon.

grievejia (approved):

Aniketsy:
thanks :)

PR #4993, opened 2026-09-20
[tensor-shapes] improve jax.numpy.fft stubs

jakevdp:
Propagate shapes in more cases & avoid overloads in python stubs.


tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
```diff
@@ -294,44 +294,220 @@ def irfft_shape(shape: IntTuple, n: Int | None, dim: int) -> IntTuple:
...
 
 @type_shape_dsl_function
-def rfftn_default_shape(shape: IntTuple) -> IntTuple:
-    if len(shape) == 0:
+def fftn_shape(
+    shape: IntTuple,
+    s: int | tuple[int, ...] | None,
+    axes: int | tuple[int, ...] | None,
+) -> IntTuple:
+    rank = len(shape)
```
stroxler, line 303: Is it worth extracting some shared helper functions here? I think a lot of the body of these functions are identical, and the DSL does allow sharing

stroxler (changes requested):
Back to you for the question about shared helpers. If you'd rather use helpers that's fine, but my first instinct is that it would make the _shapes code easier to maintain

tensor-shapes/pyrefly-jax-stubs/jax-stubs/_shapes.pyi
jakevdp: Refactored to use shared logic

stroxler (approved):

